PR #799, opened 2020-09-10
SF-1038 Fix broken login in Safari 13.1

Nateowami:
When the app loads, it immediately redirects back to the home page. This was caused by a bug in Webkit that causes storage events to be fired in the tab where they originate, rather than only in other tabs: https://bugs.webkit.org/show_bug.cgi?id=210512

The workaround is to check that the old value for `user_id` is not nullish. I also check that the value is different from the previous value, just in case some browser would fire an event for "changes" to the value that don't actually change the value.

I've made a CodePen that demonstrates what each browser it's opened in does: https://codepen.io/Nateowami/full/GRZxYbp
In browsers that implement the storage event properly, it only fires if you open that in two tabs.

I haven't tested my changes in Safari 13.1 (loading pages from my machine to Browserstack is extremely slow and succeeds far, far less than 50% of the time), but I have tested that the bug is not present in Safari 12.1, and that these changes make Scripture Forge log out as intended in Firefox. Testing QA works much better.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/799)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
@@ -69,11 +69,19 @@ export class AuthService {
     private readonly pwaService: PwaService,
     private readonly noticeService: NoticeService
   ) {
-    // listen for changes to the auth state
-    // this indicates that a user has logged in/out on a different tab/window
-    // when localStorage is cleared event.key is null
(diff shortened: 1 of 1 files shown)


irahopkinson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/799#-:-MGvpP5SDTPXeb1VgY9l:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #801, opened 2020-09-14
SF-1055 - An error occurred during login

nigel-wells:
- Retry `checkSession` for `auth0` when a timeout occurs

This issue relates to a timeout from `auth0` but I'm not entirely sure what and when this occurs. The only way I was able to consistently cause a timeout was to edit my hosts file and direct `sil-appbuilder.auth0.com` to my `localhost`. 

The timeout happens after 1 minute of running `checkSession()` in `auth.service.ts` which then runs `checkSession()` from `auth0`.

What I've currently done is a simple retry. This could be easily changed to retry multiple times but I think if it is failing multiple times then perhaps there is a larger issue going on.

I also think we don't need to check with `auth0` everytime the app loads. When we first authenticate `auth0` gives us the relevant tokens and states when it will expire. If we know the expiry date, and are storing the tokens, then there really is no need to keep asking if we have permission. I'd see this as a separate task. I know Nathaniel has also raised questions about looking at our whole login process so this could perhaps form part of that review.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/801)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
@@ -329,12 +329,20 @@ export class AuthService {
     });
   }
 
-  private checkSession(): Promise<auth0.Auth0DecodedHash | null> {
+  private checkSession(retryTimeout: boolean = false): Promise<auth0.Auth0DecodedHash | null> {
     return new Promise<auth0.Auth0DecodedHash | null>((resolve, reject) => {
-      this.auth0.checkSession({ state: JSON.stringify({}) }, (err, authResult) => {
+      this.auth0.checkSession({ state: JSON.stringify({}) }, async (err, authResult) => {
         if (err != null) {
           if (err.code === 'login_required') {
             resolve(null);
+          } else if (!retryTimeout && err.code === 'timeout') {
+            this.checkSession(true)
+              .then(retryAuthResult => {
+                resolve(retryAuthResult);
+              })
+              .catch(retryError => {
+                reject(retryError);
+              });
           } else {
             reject(err);
           }
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
@@ -337,12 +337,8 @@ export class AuthService {
             resolve(null);
           } else if (retryUponTimeout && err.code === 'timeout') {
             this.checkSession(false)
-              .then(retryAuthResult => {
-                resolve(retryAuthResult);
-              })
-              .catch(retryError => {
-                reject(retryError);
-              });
+              .then(resolve)
+              .catch(reject);
           } else {
             reject(err);
           }
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
@@ -329,14 +329,14 @@ export class AuthService {
     });
   }
 
-  private checkSession(retryTimeout: boolean = false): Promise<auth0.Auth0DecodedHash | null> {
+  private checkSession(retryUponTimeout: boolean = true): Promise<auth0.Auth0DecodedHash | null> {
     return new Promise<auth0.Auth0DecodedHash | null>((resolve, reject) => {
-      this.auth0.checkSession({ state: JSON.stringify({}) }, async (err, authResult) => {
+      this.auth0.checkSession({ state: JSON.stringify({}) }, (err, authResult) => {
         if (err != null) {
           if (err.code === 'login_required') {
             resolve(null);
-          } else if (!retryTimeout && err.code === 'timeout') {
-            this.checkSession(true)
+          } else if (retryUponTimeout && err.code === 'timeout') {
+            this.checkSession(false)
               .then(retryAuthResult => {
                 resolve(retryAuthResult);
               })


Nateowami (changes requested):
Nice work. Just a couple notes.

It looks to me like the only change is we retry upon timeout. Looking at how a promise rejection from `checkSession` would be handled, it looks like `renewTokens` would log an error and log the user out. `tryOnlineLogIn` would show a very generic "An error occurred during login" dialog and then its caller would redirect the user to log in.

I haven't thought through all the logic, but I think sometimes we are checking the session and if it times out we can schedule to try again later and don't need to inform the user (especially if we are offline). In any situation where we need to inform the user, I think we should provide a more specific message, not just "an error occurred during login". I don't know what the message should be, but for a lot of users, telling them "your session could not be renewed" is probably only slightly more helpful than showing hieroglyphics. It might be better to let the renewal fail silently (it's probably just a network issue) until a request is attempted that requires authentication (these are intercepted by `AuthHTTPInterceptor`).

Perhaps addressing the above concerns should be reserved for a future PR.

> I also think we don't need to check with auth0 everytime the app loads. When we first authenticate auth0 gives us the relevant tokens and states when it will expire. If we know the expiry date, and are storing the tokens, then there really is no need to keep asking if we have permission. I'd see this as a separate task. I know Nathaniel has also raised questions about looking at our whole login process so this could perhaps form part of that review.

I think one reason to periodically check is so that if the user logs out without our app knowing (I think there are ways for this to happen) we can log the user out. I also think it makes sense to renew tokens a good while before they expire. But if we can't reach Auth0 and the token is still valid, it doesn't make sense to bother the user about it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/801#-:-MHBhDzY1MbpPEo1-_4c:b-hgf9nm)* status: all files reviewed, 3 unresolved discussions (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 332 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/801#-MHBZTUXD2dhJdGVlOoz:-MHBZTUXD2dhJdGVlOp-:b-432b9h) ([raw file](https://github.com/sillsdev/web-xforge/blob/458f410c87ba6009439ced08a32e9c085c768fe1/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L332)):*
> ```XML
>   }
> 
>   private checkSession(retryTimeout: boolean = false): Promise<auth0.Auth0DecodedHash | null> {
> ```

Seeing the variable `retryTimeout` I interpreted it as "retry upon timeout." However, the semantics are the exact opposite. I'm not sure how to interpret it. Is it "this is a retry, therefore fail if it times out"?

I would have expected something like
``` ts
checkSession(retryUponTimeout = true) {
  ...
    if (retryUponTimeout && err.code === 'timeout') {
      this.checkSession(false)
      ...
    }
  ...
}
```

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 334 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/801#-MHBbKZt-rKOB1gwCpXL:-MHBbKZu93NVNXOR9Dy8:bt4eiq3) ([raw file](https://github.com/sillsdev/web-xforge/blob/458f410c87ba6009439ced08a32e9c085c768fe1/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L334)):*
> ```XML
>   private checkSession(retryTimeout: boolean = false): Promise<auth0.Auth0DecodedHash | null> {
>     return new Promise<auth0.Auth0DecodedHash | null>((resolve, reject) => {
>       this.auth0.checkSession({ state: JSON.stringify({}) }, async (err, authResult) => {
> ```

There's no need for the `async` keyword unless you're going to use `await`. `async` and `await` are just syntactic sugar for working with promises. If you work with them directly using their `then()` and `catch()` methods, then they work out of the box with any version of JavaScript (I believe `await` was never a reserved word in JavaScript, so it is necessary to use `async` to declare that `await` will be used as a keyword rather than a variable name).
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 339 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/801#-MHBdRBI5536YemDmYts:-MHBdRBJE-kespcmsxEA:b-1nhhf0) ([raw file](https://github.com/sillsdev/web-xforge/blob/458f410c87ba6009439ced08a32e9c085c768fe1/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L339)):*
> ```XML
>             resolve(null);
>           } else if (!retryTimeout && err.code === 'timeout') {
>             this.checkSession(true)
> ```

I think this and the following lines can be shortened to just `this.checkSession(true).then(resolve).catch(reject);`

(Actually Prettier will convert that to three lines).

nigel-wells (commented):
Yes I agree that we could make things smarter and silently fail when we know it is safe to do so. In past applications I typically only refresh tokens when needed and have that formed as part of dealing with the response from the API i.e.  call API method, if repsonse is a bad token then renew tokens and then call the API method again.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/801#-:-MHENuW51082313GPvQr:bhjzqrs)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 332 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/801#-MHBZTUXD2dhJdGVlOoz:-MHEM3k1BMNTuxqAXrT3:babgci3) ([raw file](https://github.com/sillsdev/web-xforge/blob/458f410c87ba6009439ced08a32e9c085c768fe1/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L332)):*

Seeing the variable `retryTimeout` I interpreted it as "retry upon timeout." However, the semantics are the exact opposite. I'm not sure how to interpret it. Is it "this is a retry, therefore fail if it times out"?

I would have expected something like
``` ts
checkSession(retryUponTimeout = true) {
  ...
    if (retryUponTimeout && err.code === 'timeout') {
      this.checkSession(false)
      ...
    }
  ...
}
</blockquote></details>

As they say, there are multiple ways to skin a cat :) I'll reword it to what you've suggested to avoid any confusion.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 334 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/801#-MHBbKZt-rKOB1gwCpXL:-MHEKncMEeOd8mfnqmgV:b-yplgjr) ([raw file](https://github.com/sillsdev/web-xforge/blob/458f410c87ba6009439ced08a32e9c085c768fe1/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L334)):*

There's no need for the `async` keyword unless you're going to use `await`. `async` and `await` are just syntactic sugar for working with promises. If you work with them directly using their `then()` and `catch()` methods, then they work out of the box with any version of JavaScript (I believe `await` was never a reserved word in JavaScript, so it is necessary to use `async` to declare that `await` will be used as a keyword rather than a variable name).
</blockquote></details>

Yes my bad - the `async` was leftover from a previous attempt and I forgot to remove it after switching to `then()` and `catch()`. Have removed now.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 339 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/801#-MHBdRBI5536YemDmYts:-MHELFjCAPUDCuDajHOg:b-hdx31) ([raw file](https://github.com/sillsdev/web-xforge/blob/458f410c87ba6009439ced08a32e9c085c768fe1/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L339)):*

I think this and the following lines can be shortened to just `this.checkSession(true).then(resolve).catch(reject);`

(Actually Prettier will convert that to three lines).
</blockquote></details>

But I need to resolve and pass the actual response from `checkSession(true)` - same for the `reject`.


<!-- Sent from Reviewable.io -->

**Review by Nateowami — CHANGES_REQUESTED** [review]
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/801)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 339 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/801#-MHBdRBI5536YemDmYts:-MHEbpAr0ZTX6dCER7Ke:b-85pkr7) ([raw file](https://github.com/sillsdev/web-xforge/blob/458f410c87ba6009439ced08a32e9c085c768fe1/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L339)):*

But I need to resolve and pass the actual response from `checkSession(true)` - same for the `reject`.
</blockquote></details>

Why would it not pass the response from `checkSession(true)`? The following:
``` ts
retryAuthResult => {
  resolve(retryAuthResult);
}
```
Is an anonymous function that when called takes what is passed to it and calls another function. The only functional difference between it and the `resolve` function is that if multiple arguments are passed to it, it only passes one along (which I don't think is important in this particular instance).

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/801)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 339 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/801#-MHBdRBI5536YemDmYts:-MHJ8tRuAjhNT9Pzt_MC:b5homfa) ([raw file](https://github.com/sillsdev/web-xforge/blob/458f410c87ba6009439ced08a32e9c085c768fe1/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L339)):*

Why would it not pass the response from `checkSession(true)`? The following:
``` ts
retryAuthResult => {
  resolve(retryAuthResult);
}
```
Is an anonymous function that when called takes what is passed to it and calls another function. The only functional difference between it and the `resolve` function is that if multiple arguments are passed to it, it only passes one along (which I don't think is important in this particular instance).
</blockquote></details>

Right ok. I did not realise that simply going `.then(resolve)` would also pass the returned value from that promise. Good to know!

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/801)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #803, opened 2020-09-15
SF-1002 Allow users to edit display name offline

Nateowami:
- This is already possible when the user answers a question for the first time, since the user is prompted with a dialog.
- There isn't really any reason it needed to be disabled in the first place. I think maybe at the time I thought editing the name would require immediately saving it via an API call, but the user object is part of the realtime system. This is just reverting a change that never needed to be made in the first place.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/803)


RaymondLuong3 (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/803#-:-MHNGgs55ZODrARLuBzR:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #805, opened 2020-09-16
SF-1038b Fix tryOnlineLogIn failing when 3rd party cookies are blocked

Nateowami:
The problem was caused by attempting to get the token from the URL after it had already been removed from the URL. That lead to sending a request to Auth0 to check the session, and in browsers that block third-party cookies, the request would fail, leading to the user being logged out.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/805)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@angular/core';
 import { Router } from '@angular/router';
 import { translate } from '@ngneat/transloco';
-import { AuthorizeOptions, WebAuth } from 'auth0-js';
+import { Auth0DecodedHash, AuthorizeOptions, WebAuth } from 'auth0-js';
 import jwtDecode from 'jwt-decode';
 import { CookieService } from 'ngx-cookie-service';
 import { SystemRole } from 'realtime-server/lib/common/models/system-role';
@@ -57,6 +57,10 @@ export class AuthService {
     scope: 'openid profile email ' + environment.scope,
     audience: environment.audience
   });
+
+  private parsedHashPromise = new Promise<Auth0DecodedHash | null>((resolve, reject) =>
+    this.auth0.parseHash((err, authResult) => (err != null ? reject(err) : resolve(authResult)))
+  );
 
   constructor(
     private readonly remoteStore: SharedbRealtimeRemoteStore,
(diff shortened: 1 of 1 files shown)


nigel-wells (commented):
For the most part it looks fine. Does this resolve the issue if someone is logged in, goes offline, refreshes the page while still offline, and then goes back online? I believe the check session would get called  as the promise would have returned `null` for the `parseHash`. That said, this is still better as it at least resolves the immediate issue you're facing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/805#-:-MHOnRV99l2BjvVcDGC3:bb3voj6)* status: :shipit: complete! all files reviewed, all discussions resolved

irahopkinson (approved):
@nigel-wells the scenario you mentioned sounds like you were running locally without PWA testing. My local testing showed that scenario works fine.

I don't think we have yet fixed the scenario of being online and refreshing in a browser that doesn't allow 3rd party cookies. With this change the user will just have to log in again (since silent login is blocked by the browser).

However, this issue is not about that and it does appear to fix the issue at hand. And I want it on QA to ship with the migration.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/805#-:-MHPUIAy4kWmt2GrBLBX:b83mdl5)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 203 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/805#-MHPV15i1tM35A5q37bm:-MHPV15i1tM35A5q37bn:b5qxpbu) ([raw file](https://github.com/sillsdev/web-xforge/blob/9314e265e013e883d10c47285e20feab02c97ef2/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L203)):*
> ```
> let authResult = await this.parseHash();
> ```
It would probably be better (if possible) to pull this up into `tryLogIn` which is already a promise rather than introducing another promise. However that is not sufficient reason on its own to block this PR.

Nateowami (commented):
@nigel-wells I just tested this on qa. After coming back online it briefly redirects to Auth0 but then comes back so fast I barely notice. Changes made while offline were preserved.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/805#-:-MHR2N527XaXUCWD4kQp:bts3tq1)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 203 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/805#-MHPV15i1tM35A5q37bm:-MHR1ehh8aOJRptVRkM2:bz5opki) ([raw file](https://github.com/sillsdev/web-xforge/blob/9314e265e013e883d10c47285e20feab02c97ef2/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L203)):*

> ```
> let authResult = await this.parseHash();
> ```
It would probably be better (if possible) to pull this up into `tryLogIn` which is already a promise rather than introducing another promise. However that is not sufficient reason on its own to block this PR.
</blockquote></details>

I would prefer if it didn't create a separate promise that's sticking around like this, but I'm not quite sure how that would be done since it's used in two separate places.

PR #816, opened 2020-09-24
Comment out the output from ParatextService

RaymondLuong3:
* The output was helpful during development of ParatextService
* It can be uncommented on a local machine when needed

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/816)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -46,7 +46,6 @@ namespace SIL.XForge.Scripture.Services
     /// </summary>
     public class ParatextService : DisposableBase, IParatextService
     {
-        private readonly IWebHostEnvironment _webHostEnvironment;
         private readonly IOptions<ParatextOptions> _paratextOptions;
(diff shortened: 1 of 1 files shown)


marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/816)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 123 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/816#-MI0bgPiEZo3ysF_yGPD:-MI0bgPiEZo3ysF_yGPE:bh5ptfm) ([raw file](https://github.com/sillsdev/web-xforge/blob/a80c505d6916280932d8086be4d54ac6295ca719/src/SIL.XForge.Scripture/Services/ParatextService.cs#L123)):*
> ```Smalltalk
>         /// <summary> Prepare access to Paratext.Data library, authenticate, and prepare Mercurial. </summary>
>         public void Init()
>         {
> ```

Okay. And another way would be to only do it if an environment variable (like SF_PT_OUTPUT or something) was set.

PR #826, opened 2020-10-05
Mock versioning manager functionality in ParatextService tests

RaymondLuong3:
An exception was caught in test each time VersioningManager was used to get a new VersionedText. The tests continued and passed anyway. This change pulls the commit part out of PutNotes so it can be mocked in the tests.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/826)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/IParatextDataHelper.cs
+++ b/src/SIL.XForge.Scripture/Services/IParatextDataHelper.cs
@@ -1,0 +1,10 @@
+using Paratext.Data;
+using Paratext.Data.Repository;
+
+namespace SIL.XForge.Scripture.Services
+{
+    public interface IParatextDataHelper
+    {
+        void CommitVersionedText(ScrText scrText, string comment);
+    }
+}
--- a/src/SIL.XForge.Scripture/Services/ParatextDataHelper.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextDataHelper.cs
@@ -1,0 +1,15 @@
+using Paratext.Data;
+using Paratext.Data.Repository;
+
+namespace SIL.XForge.Scripture.Services
+{
+    /// <summary> Provides methods calls to Paratext Data. Can be mocked in tests. </summary>
+    public class ParatextDataHelper : IParatextDataHelper
+    {
+        public void CommitVersionedText(ScrText scrText, string comment)
+        {
+            VersionedText vText = VersioningManager.Get(scrText);
+            vText.Commit(comment, null, false);
+        }
+    }
+}
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -56,6 +56,7 @@ namespace SIL.XForge.Scripture.Services
         private readonly IExceptionHandler _exceptionHandler;
         private readonly ILogger _logger;
         private readonly IJwtTokenHelper _jwtTokenHelper;
+        private readonly IParatextDataHelper _paratextDataHelper;
         private string _applicationProductVersion = "SF";
         private string _registryServerUri = "https://registry.paratext.org";
         private string _sendReceiveServerUri = InternetAccess.uriProduction;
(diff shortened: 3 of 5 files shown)


marksvc (commented):
Hmm. So when the SF app is running, or when the tests are running, calling vText.Commit() results in removing the CommentManager for the ScrText.  The SF app can recover from this just fine because it can Load() prior comments. But the test doesn't have any comments to Load() when a CommentManager for ScrText is re-created for the edit-comment test.

This patch would fix the test by making the CommentManager not get removed in the first place. So although it will make the test pass, it will do so my incorrectly describing ParatextData's behaviour. 

I want to look at the problem a bit more. 

(BTW my understanding of how all this works in currently thru the lens of working with ParatextData 9.2.2-alpha, so it may be a bit different than what you may have recently been debugging.)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/826#-:-MIttRUm-b7kX7hnsQRT:b-khdygn)* status: :shipit: complete! all files reviewed, all discussions resolved

marksvc (commented):
Okay, so I think a couple ways to fix this that would be more comfortable with are:

(1) If we have ParatextService use a mock CommentManager, so when ParatextService.cs PutNotes() calls `CommentManager.Get(scrText)`, we could specify ahead of time what will be returned. And we could specify that each time it's called for the PutNotes test, it gets back a certain comment manager, that does or doesn't have a thread in it. For the edit-comment case, our CommentManager could have a thread that could be edited. 

(2) We could move the mocking a bit deeper into the PTD library. In CommentManager.cs Load() it loads up some comments that belong to a ScrText. We could set the necessary parts of a ScrText so it finds and loads files (from a fake filesystem). That's exercises more of the ParatextData DLL which is nice, but less of a unit test. I would favour option 1 above. 

Now, the difference between option 1 and the patch you have in this PR  is rather abstract. 

I think I would be okay with mocking _both_ the CommentManager and the VersionedText.Commit() method. 

If this doesn't all sound really sensible, let's chat about it. :)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/826#-:-MItw_VhBFZm5vtBfQmH:b-vroaj2)* status: :shipit: complete! all files reviewed, all discussions resolved

RaymondLuong3 (commented):
OK. So this is only to fix the exception that was being caught silently in the test. If you want to push changes to this branch, you can. Or we can get this PR through, and you can open a new one.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/826#-:-MItzNFZ3ZDm0NgYzonB:b-1j8avf)* status: :shipit: complete! all files reviewed, all discussions resolved

marksvc (commented):
Right, thank you. There is more context here than I remembered. :-)   I will switch to the master branch and look again.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/826#-:-MIuIQWF8mXz9dCC2VYB:bckemmf)* status: :shipit: complete! all files reviewed, all discussions resolved

marksvc (commented):
When examining from the master branch (with the currently-used ParatextData nuget), I observed how VersionedText starts to go astray because we didn't set a Settings.Guid. Setting that value in our tests lets the test fail without the usage of VersionedText crashing. Although that doesn't mean we shouldn't be mocking out our usage of VersionedText.Commit() anyway. What do you think of https://github.com/sillsdev/web-xforge/pull/827 ?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/826#-:-MIuZ45L45CosBfzFI7y:bnbfy6c)* status: :shipit: complete! all files reviewed, all discussions resolved

marksvc (approved):
I see now that it is crashing because in VersionedText Commit(), Hg.Default==null on your machine but not on mine. I agree to mock our usage of Versioningmanager / VersionedText.
BTW I wonder if your Hg.Default is not set because ParatextService SetupMercurial() has trouble finding the hg path?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/826#-:-MIurpHwDR6gf5nDDNPN:b-6dwf91)* status: :shipit: complete! all files reviewed, all discussions resolved

irahopkinson (approved):
:lgtm: just note this doesn't fix the SF-1089 issue for me but the chat above and in Slack doesn't indicate that it should and this change does no harm so I will merge it.

> Reviewed 4 of 5 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/826#-:-MIx6Q6eCPqTzpgBKPM1:b-98fdh1)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #827, opened 2020-10-05
Set MockScrText.Settings.Guid

marksvc:
- ParatextServiceTests.cs PutNotes test fails when VersionedText.cs
  VersionedText(ScrText) checks if ScrText Settings.Guid is empty, and
  if so calculates a new guid for it. It tries to write the info to a
  file but has trouble and crashes.
- I haven't fully compared the context of our MockScrText construction
  and how ScrText objects get constructed. But I do see that setting
  ScrText.Settings results in setting ScrText.CachedGuid to the value
  of Settings.Guid.
- Set ScrText.Settings.Guid when we set up a MockScrText with a known
  Guid so we don't crash in VersionedText.cs VersionedText(ScrText).

To clarify, this code was written to prevent the test from crashing rather than testing. It won't necessarily make the test pass.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/827)

### DIFF
--- a/test/SIL.XForge.Scripture.Tests/Services/MockScrText.cs
+++ b/test/SIL.XForge.Scripture.Tests/Services/MockScrText.cs
@@ -20,12 +20,18 @@ namespace SIL.XForge.Scripture.Services
         {
             this.projectName = projectName;
             Settings = new MockProjectSettings(this);
+            // ScrText sets its cachedGuid from the settings Guid. Here we are doing it the other way around.
+            // Some tests may need both MockScrText.Guid and MockScrText.Settings.Guid to be set.
             _language = new MockScrLanguage(this);
         }
 
         public string CachedGuid
         {
-            set { cachedGuid = value; }
+            set
+            {
+                cachedGuid = value;
+                Settings.Guid = value;
+            }
         }
 
         public Dictionary<string, string> Data = new Dictionary<string, string>();


RaymondLuong3 (commented):
Are you seeing a crash when `VersioningManager.Get(ScrText))` where it call `VersionedText.cs VersionedText(ScrText)`? That is where I crash. I can see why this would be different on linux and windows. It crashes with a "object reference not set to an instance of an object" presumably on `Hg.Default.IsRepository()`. I think we do need to mock out `VersioningManager`

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/827#-:-MIueDS62al8yIkkne3Z:b11glv2)* status: 0 of 1 files reviewed, all discussions resolved

marksvc (commented):
Yes, I agree now about mocking out VersioningManager.
I would like to take a look to see if we should still use this MockScrText patch, or not worry about it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/827#-:-MIuxU42-cpl6NaXNJQ3:be065dd)* status: 0 of 1 files reviewed, all discussions resolved

irahopkinson (changes requested):
This change doesn't fix SF-1089 for me, although on Windows I may have some additional set up required.

As I mentioned in Slack:
SF-1089 is not even published at this point. Please leave it until after SF-1088 is complete, esp since @pmachapman  is now waiting on it. Do note that although SF-1089 is only a test failure not a functional failure, i.e. the code works even if the test doesn't.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/827#-:-MIx8Hy5-cQQPJzY3E0W:bavk6d2)* status: 0 of 1 files reviewed, all discussions resolved

marksvc:
Sorry, this wasn't clear from the descriptions I included. This code was written to prevent the test from crashing rather than testing. It won't necessarily make the test pass.

RaymondLuong3 (approved):
This does not necessarily fix SF-1089, but it is a bug, and this might reveal issues in other tests. We should get it in.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/827#-:-MJ-24bY2TL5GbMj4BGe:b-99l8zf)* status: :shipit: complete! all files reviewed, all discussions resolved

marksvc:
> might reveal issues in other tests
Good point. I want to wait until I have another look at the PutNotes test before advocating this along.

marksvc:
Okay, so this code doesn't 'fix' anything broken right now. But it may prevent a problem in the future. I revised the patch's comment and commit message to better reflect the new situation, and I return to proposing its inclusion.

marksvc:
(The bot is complaining about Delta.cs.)

marksvc:
(So github says that @irahopkinson has "requested changes" on the PR. I assume that isn't significant right now, but I am not in a hurry and will leave the PR in case Ira has something to add or get changed.)

irahopkinson (approved):
:lgtm:

> Reviewed 1 of 1 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/827#-:-MJDAClACl0MWmGM71dK:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #831, opened 2020-10-06
SF-1088 Sync: Upgrade to ParatextData 9.2.3-alpha

marksvc:
- Stop using now-removed MultiUserScrText.
- Satisfy new ParatextUser argument to ScrText ctor.
- Subclass ParatextUser as SFParatextUser, to have a 'valid' user.
- RegistryServer.Initialize() now wants a ParatextUser. We aren't
  using it anyway. Stop trying to initializing it until we need it.
- Satisfy call to GetHgUri() without crashing because of a null
  RegistrationCode.

---

So far so good in testing.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/831)

### DIFF
--- a/src/SIL.XForge.Scripture/Models/SFParatextUser.cs
+++ b/src/SIL.XForge.Scripture/Models/SFParatextUser.cs
@@ -1,0 +1,13 @@
+using Paratext.Data.Users;
+
+namespace SIL.XForge.Scripture.Models
+{
+    /// <summary>
+    /// SF customization of ParatextUser
+    /// </summary>
+    public class SFParatextUser : ParatextUser
+    {
+        public SFParatextUser(string ptUsername) : base(ptUsername, true)
+        { }
+    }
+}
--- a/src/SIL.XForge.Scripture/SIL.XForge.Scripture.csproj
+++ b/src/SIL.XForge.Scripture/SIL.XForge.Scripture.csproj
@@ -18,7 +18,7 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Mvc.NewtonsoftJson" Version="3.1.2" />
-    <PackageReference Include="ParatextData" Version="9.1.10-beta" />
+    <PackageReference Include="ParatextData" Version="9.2.3-alpha" />
     <PackageReference Include="SIL.Machine.WebApi" Version="$(MachineVersion)" />
     <PackageReference Include="System.Text.Encoding.CodePages" Version="4.7.0" />
   </ItemGroup>
--- a/src/SIL.XForge.Scripture/Services/InternetSharedRepositorySourceProvider.cs
+++ b/src/SIL.XForge.Scripture/Services/InternetSharedRepositorySourceProvider.cs
@@ -1,5 +1,7 @@
 using System;
+using Paratext.Data.Users;
 using SIL.XForge.Models;
+using SIL.XForge.Scripture.Models;
 
 namespace SIL.XForge.Scripture.Services
 {
@@ -24,11 +26,13 @@ namespace SIL.XForge.Scripture.Services
                 throw new ArgumentException();
             }
 
+            string ptUsername = _jwtTokenHelper.GetParatextUsername(userSecret);
+            var ptUser = new SFParatextUser(ptUsername);
             JwtRESTClient jwtClient = GenerateParatextRegistryJwtClient(userSecret, registryServerUri,
                 applicationProductVersion);
             IInternetSharedRepositorySource source =
                 new JwtInternetSharedRepositorySource(userSecret.ParatextTokens.AccessToken,
-                    jwtClient, sendReceiveServerUri);
+                    jwtClient, ptUser, sendReceiveServerUri);
             source.RefreshToken(userSecret.ParatextTokens.AccessToken);
             return source;
         }
--- a/src/SIL.XForge.Scripture/Services/JwtInternetSharedRepositorySource.cs
+++ b/src/SIL.XForge.Scripture/Services/JwtInternetSharedRepositorySource.cs
@@ -15,8 +15,8 @@ namespace SIL.XForge.Scripture.Services
     {
         private readonly JwtRESTClient _registryClient;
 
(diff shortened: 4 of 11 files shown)


marksvc:
Still working:
SF and PT sync Scripture text back and forth.
SF sends answer-notes to PT.

ddaspit (approved):
I just have one question, otherwise :lgtm:.

> Reviewed 11 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/831#-:-MJ0Rsj04g2EU6LxqcTQ:bjim1u)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/Services/JwtInternetSharedRepositorySource.cs, line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/831#-MJ0RHxC1tFzomLf1jII:-MJ0RHxC1tFzomLf1jIJ:b-rullm3) ([raw file](https://github.com/sillsdev/web-xforge/blob/2950902894a617a6c4a7f4a96a3993e58dd20c22/src/SIL.XForge.Scripture/Services/JwtInternetSharedRepositorySource.cs#L92)):*
> ```Smalltalk
>         /// This looks like it would be important, but it doesn't seem to matter what it returns, to synchronize.
>         /// </summary>
>         public override string GetHgUri(SharedRepository sharedRepository)
> ```

Why does this need to be overridden? Is it an abstract method?

pmachapman (commented):
Works great with my DBL source code. My only real comment is that the GetHgUri() function might need be implemented (I can see why it caused problems for you).

> Reviewed 11 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/831#-:-MJ0cQ0q62WA2JmfOUTN:b-548k3s)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[src/SIL.XForge.Scripture/Services/JwtInternetSharedRepositorySource.cs, line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/831#-MJ0RHxC1tFzomLf1jII:-MJ0bYVA5xSvgBNvKzgg:b-2knjns) ([raw file](https://github.com/sillsdev/web-xforge/blob/2950902894a617a6c4a7f4a96a3993e58dd20c22/src/SIL.XForge.Scripture/Services/JwtInternetSharedRepositorySource.cs#L92)):*

Why does this need to be overridden? Is it an abstract method?
</blockquote></details>

We might need to implement this for SharingLogic::ShareChanges() to work correctly, as it looks like VersionedText::ShareChanges() calls this directly. Perhaps it is a function that will get a bit more use in future?

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/831)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit and @pmachapman)

___
*[src/SIL.XForge.Scripture/Services/JwtInternetSharedRepositorySource.cs, line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/831#-MJ0RHxC1tFzomLf1jII:-MJ2whVj4SRo7r2w8sCM:bdmxjzf) ([raw file](https://github.com/sillsdev/web-xforge/blob/2950902894a617a6c4a7f4a96a3993e58dd20c22/src/SIL.XForge.Scripture/Services/JwtInternetSharedRepositorySource.cs#L92)):*
When syncing we were getting into InternetSharedRepositorySource.cs GetHgUri(), which uses Uri.EscapeDataString(...RegistrationCode). 

I had found in another place that I need our objects for ParatextUser to have .IsValid=true, and the only way I can see to do that without asking for another modification to Paratext, is to use the ParatextUser constructor that lets me specify if the user is valid or not. 
This means then that there is no way for me to set or override the RegistrationCode, so it just ends up being null.

Calling `Uri.EscapeDataString(null)` throws an exception, because it won't tolerate a null argument. So in our JwtInternetSharedRepositorySource subclass of InternetSharedRepositorySource, I override GetHgUri so we don't end up in InternetSharedRepositorySource.GetHgUri, but in our own implementation which doesn't crash. 

There are a number of places that call GetHgUri(). I think the place where GetHgUri was being called in my testing was in fact VersionText ShareChanges(). And wherever it was, it seemed to not matter if I returned a slightly different, or a completely wrong return value. 

> Is it an abstract method?

No, it overrides another override, which overrides a virtual.

> We might need to implement this for SharingLogic::ShareChanges() to work correctly, as it looks like VersionedText::ShareChanges() calls this directly. Perhaps it is a function that will get a bit more use in future?

Right. I don't know. And if we do need this method to do something, we're going to need to figure out what to do with the RegistrationCode, since we don't use that in SF.


I just had a look to see what the impact might be for not letting GetHgUri return something meaningful for the case of calling it from VersionedText ShareChanges().

- Returning junk from GetHgUri() will mean that when Hg.GetOutgoingRevisions is called, it won't return a real set of revisions.
- So in VersionedText.cs ShareChanges(), result.RevisionsSent will not be a real set of revisions that were sent. A comment says it relates to displaying conflicts in a summary. 
- We go back to SharingLogic.cs Share1Project, which doesn't care about result.RevisionsSent.
- We go back to SharingLogic.ShareChanges(), which bundles results.
- ..
- We get back to ParatextService.cs SendReceiveAsync(), which receives the results and throws them away. 

So it's probably not an immediate problem for us. Unless we call GetHgUri from more places.

It doesn't look like RevisionsSent is used much in ParatextData.cs. But it looks like it is used by Paratext/Repository/*.cs files.

marksvc (commented):
I'm not sure if my explanation will make anyone no longer approve :). I'm going to send it in, and Peter can be building on the latest ParatextData. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/831#-:-MJ3kOTg6d2-yMjZ9OcC:b8yah38)* status: 10 of 11 files reviewed, 1 unresolved discussion (waiting on @ddaspit and @pmachapman)

PR #832, opened 2020-10-07
SF-935 Show useful messages on scripture panel when text is unavailable

RaymondLuong3:
* Show that scripture book does not exist if online
* Show that scripture text is unavailable if offline
* Show book is empty if book does not contain chapters
* Add text.component tests
* Put generic SFProject and TextData construction in test-utils

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/832)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.spec.ts
@@ -79,6 +79,11 @@ describe('CheckingTextComponent', () => {
     expect(env.segmentHasQuestion(1, 4)).toBe(true);
     expect(env.isSegmentHighlighted(1, 4)).toBe(true);
   }));
+
+  it('shows Loading message when offline and text not available', fakeAsync(() => {
+    const env = new TestEnvironment();
+    env.wait();
+  }));
 });
 
 class TestEnvironment {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.spec.ts
@@ -1,23 +1,25 @@
 import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
-import { CheckingShareLevel } from 'realtime-server/lib/scriptureforge/models/checking-config';
 import { SFProject } from 'realtime-server/lib/scriptureforge/models/sf-project';
-import { SFProjectRole } from 'realtime-server/lib/scriptureforge/models/sf-project-role';
 import { VerseRef } from 'realtime-server/lib/scriptureforge/scripture-utils/verse-ref';
 import * as RichText from 'rich-text';
+import { BehaviorSubject } from 'rxjs';
 import { anything, mock, when } from 'ts-mockito';
+import { PwaService } from 'xforge-common/pwa.service';
 import { TestRealtimeModule } from 'xforge-common/test-realtime.module';
 import { TestRealtimeService } from 'xforge-common/test-realtime.service';
 import { configureTestingModule, TestTranslocoModule } from 'xforge-common/test-utils';
 import { UICommonModule } from 'xforge-common/ui-common.module';
 import { SFProjectDoc } from '../../../core/models/sf-project-doc';
 import { SF_TYPE_REGISTRY } from '../../../core/models/sf-type-registry';
-import { Delta, TextDoc, TextDocId } from '../../../core/models/text-doc';
+import { TextDoc, TextDocId } from '../../../core/models/text-doc';
 import { SFProjectService } from '../../../core/sf-project.service';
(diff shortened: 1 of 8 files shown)


irahopkinson (changes requested):
Just a couple of minor things and then :lgtm:

> Reviewed 8 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/832#-:-MJCs97K9LCoCCwaM9GJ:bfi2obx)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.spec.ts, line 146 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/832#-MJCor6rBiSaImBiumxU:-MJCor6s8pb9ZcYyOM4z:b2fo5lw) ([raw file](https://github.com/sillsdev/web-xforge/blob/56607e249f7fc83db2dcafb660fb318ab284d10b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.spec.ts#L146)):*
> ```
> addTextDoc
> ```
I don't see any reason for this to become `public`?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 31 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/832#-MJCqEM_3ux1gmSBTAw_:-MJCqEM_3ux1gmSBTAwa:b-x0vv10) ([raw file](https://github.com/sillsdev/web-xforge/blob/56607e249f7fc83db2dcafb660fb318ab284d10b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L31)):*
> ```
> when(mockedTranslocoService.translate<string>(anything())).thenCall(
>   (translationStringKey: string) => translationStringKey
> );
> ```
Is there some reason this needs to be declared in the global namespace?

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/832)* status: all files reviewed, 2 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.spec.ts, line 146 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/832#-MJCor6rBiSaImBiumxU:-MJEBqZaDheVLFMu7hA9:byhbatj) ([raw file](https://github.com/sillsdev/web-xforge/blob/56607e249f7fc83db2dcafb660fb318ab284d10b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.spec.ts#L146)):*

> ```
> addTextDoc
> ```
I don't see any reason for this to become `public`?
</blockquote></details>

You're right. Not sure how that got changed.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 31 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/832#-MJCqEM_3ux1gmSBTAw_:-MJECxxV9SdldIVSFgkB:b-47s7xw) ([raw file](https://github.com/sillsdev/web-xforge/blob/56607e249f7fc83db2dcafb660fb318ab284d10b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L31)):*

> ```
> when(mockedTranslocoService.translate<string>(anything())).thenCall(
>   (translationStringKey: string) => translationStringKey
> );
> ```
Is there some reason this needs to be declared in the global namespace?
</blockquote></details>

Nope. Should be corrected now.

irahopkinson (approved):
:lgtm:

> Reviewed 8 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/832#-:-MJF9e6e4TGiqKgiUi16:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

Nateowami:
@irahopkinson Is it intentional that the commits were merged instead of rebased or squashed?

irahopkinson:
@nateowami yes, if there is more than one committer I do a merge commit rather than a squash to give credit to both authors. This is unusual though.

Nateowami:
@irahopkinson Ah, OK. I just assumed we'd also try to keep a linear history by rebasing.

irahopkinson:
@Nateowami we are as long as a commit on master is complete. For PR #832, both commits where needed since they were different authors but they needed to be together to make them complete, hence a merge commit. That way doing a bisect should always work since commits are complete. I had actually asked you to do a merge commit for the NG 9 work since even though it was all the same author it was only complete as one unit. But its not really a problem just the not the ideal.

Nateowami:
@irahopkinson Hmmm. I don't really understand. I guess it's to preserve the history of "these two commits go together"? That makes sense. I don't think the GitHub UI even allows that, so we probably have to do it from the CLI (The UI wouldn't allow it because we require branches to be up to date, which would result in a merge commit not being created. Unless of course GitHub offers the option of creating merge commits even when not needed).

> That way doing a bisect should always work since commits are complete.

`git bisect` will go down the two different paths, so I don't see a difference here, except that the bisect will be slightly more confusing because the history is not linear. But I think this is less important than preserving the history of what commits go together.

PR #837, opened 2020-10-13
SF-1080 - Quill box does not display in the Project after login to SF

nigel-wells:
- Increased minimum browser support of Safari to 11.1 desktop and 11.3 on mobile

After investigation and looking at browser version usage it was decided to simply not support Safari 10.x any longer.

Note that I've had some issues with testing on BrowserStack with `localhost` so it might need to wait until it reaches QA to test the dialog.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/837)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts
@@ -33,13 +33,13 @@ export function supportedBrowser(): boolean {
     chromium: '>=58',
     edge: '>=76',
     firefox: '>=51',
-    safari: '>=10.1',
+    safari: '>=11.1',
 
     mobile: {
       chrome: '>=78',
       firefox: '>=68',
       opera: '>=46',
-      safari: '>=10.3',
+      safari: '>=11.3',
       'android browser': '>=76',
       'samsung internet': '>=7.2'
     }


Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/837)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts, line 42 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/837#-MJXNV8vEOn8QXtZfbl4:-MJXORBG6i2gdrbwTbbz:b-kphyb1) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc244b0ffdf71d55c6bd25b2f88532fef5ed94d2/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts#L42)):*

Why 11.3 here, but 11.1 for desktop?
</blockquote></details>

I think the original reason for the discrepancy was a lack of complete data in [caniuse for IndexedDB 2.0](https://caniuse.com/indexeddb2):



Notice that It shows IndexedDB 2.0 is supported in Safari desktop 10.1 but not 10.0, whereas for mobile there is no support data from 10.0 through 10.2, leaving 10.3 as the first definitely supported version. Rather than digging deeper I think the supported versions were set to the first known-good version. So unless there's some other reason you know of, it wouldn't make sense to carry the discrepancy to version 11.

Probably we can just require >=11.0?

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/837)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts, line 42 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/837#-MJXNV8vEOn8QXtZfbl4:-MJYQnXH3AwTyfOSa_oQ:b8mkq8g) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc244b0ffdf71d55c6bd25b2f88532fef5ed94d2/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts#L42)):*

I think the original reason for the discrepancy was a lack of complete data in [caniuse for IndexedDB 2.0](https://caniuse.com/indexeddb2):



Notice that It shows IndexedDB 2.0 is supported in Safari desktop 10.1 but not 10.0, whereas for mobile there is no support data from 10.0 through 10.2, leaving 10.3 as the first definitely supported version. Rather than digging deeper I think the supported versions were set to the first known-good version. So unless there's some other reason you know of, it wouldn't make sense to carry the discrepancy to version 11.

Probably we can just require >=11.0?
</blockquote></details>

My thinking was based on 2 assumptions. The first being that BrowserStack specifically put versioni 11.3 for iOS and 11.1 for desktop because that is the highest version someone can automatically upgrade to on that version of the operating system i.e. we know users will at least be on that version and nothing earlier. The second is that BrowserStack only offers that version for those devices which means I can't confirm (at least via BrowserStack) if it is actually resolved in version 11.0. So, based on these assumptions, I figure given the already very low percentage of someone being on 11.x in the first place it will be an even lower percentage that someone hasn't had the automatic updates applied to Safari for the applicable OS.

Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/837#-:-MJYSW4016X0K6ewfzSn:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts, line 42 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/837#-MJXNV8vEOn8QXtZfbl4:-MJYSSDuDj-lpgU3YXts:b-bguy2t) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc244b0ffdf71d55c6bd25b2f88532fef5ed94d2/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts#L42)):*

My thinking was based on 2 assumptions. The first being that BrowserStack specifically put versioni 11.3 for iOS and 11.1 for desktop because that is the highest version someone can automatically upgrade to on that version of the operating system i.e. we know users will at least be on that version and nothing earlier. The second is that BrowserStack only offers that version for those devices which means I can't confirm (at least via BrowserStack) if it is actually resolved in version 11.0. So, based on these assumptions, I figure given the already very low percentage of someone being on 11.x in the first place it will be an even lower percentage that someone hasn't had the automatic updates applied to Safari for the applicable OS.
</blockquote></details>

Ah, OK, that makes sense then.

PR #839, opened 2020-10-14
SF-1096 Fix DebugElement toBeDefined to not.toBeNull

marksvc:
- DebugElement query() appears to be returning type
  `DebugElement | null`, and so checking if the return is
  .toBeDefined() will always be true. So instead, check whether the
  return is or isn't null.
- scripture-chooser-dialog.component.spec.ts: Methods were defined
  as returning type `DebugElement | undefined`, which is probably not
  true. Change that to the also incorrect(?), but matching the .d.ts
  declarations, type `DebugElement`.
- app.component.spec.ts: A test here was using .toBeDefined, and
  when that was changed, it revealed that the test wasn't really
  working.
  - Add more setup to test.
  - mdc-list-item elements don't seem to keep an id, and so we
    can't check for them by looking for an element with such an id. So
    look for items in the menu list that have matching text, to check
    for presence or absence of a list item.
  - Setting `component.isProjectAdmin$ = of(true or false)` didn't
    seem to be working, so keep the same observable and send out
    .next() values on it.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/839)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html
@@ -228,7 +228,6 @@
           <mdc-list-item
             [appRouterLink]="['/projects', selectedProjectId, 'sync']"
             [class.list-item-disabled]="!isAppOnline"
-            id="sync-item"
             (selectionChange)="itemSelected()"
           >
             <mdc-icon mdcListItemGraphic>sync</mdc-icon>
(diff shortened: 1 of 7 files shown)


RaymondLuong3 (approved):
:lgtm: 

> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/839#-:-MJd1T2A24CXUb6Q5p6o:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved

Nateowami (approved):
:lgtm:

> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/839#-:-MJd3C-N3dvK5_TIAW3A:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #843, opened 2020-10-19
SF-1104 Increase width of project selection list

marksvc:
- The contents of the list can start to disappear from view if wrapped
  over several lines in the vertical space given to each list item.
  This can happen if the project name is long, or if we append
  "(already connected)".
- Increase the width of the two project lists to decrease the chance
  of some of the menu item being obscured by wrapping. It only starts
  to be a problem if it wraps over more than 2 lines.
- The #projects-menu min-width uses min() to help choose between
  values suitable for different size viewports.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/843)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.scss
@@ -59,7 +59,6 @@
 
     #project-select {
       width: 100%;
-      margin-right: 32px;
     }
 
     #no-projects-msg {
@@ -114,11 +113,22 @@
   max-width: calc(100% - 44px);
 }
 
+mdc-menu {
+  // Make the MDC Menu List extend to the width of the MDC Select that it drops down from, minus some space to still
+  // show the drop down arrow.
+  width: calc(100% - 40px);
+}
+
 mdc-select {
   margin-bottom: 4px;
   ::ng-deep .mdc-select__anchor {
     width: 100%;
   }
+}
+
+mdc-list-item {
+  height: unset;
+  min-height: 48px;
 }
 
 mdc-linear-progress ::ng-deep {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss
@@ -108,3 +108,8 @@
     min-width: 150px;
   }
 }
+
+mdc-list-item {
+  height: unset;
+  min-height: 48px;
+}


marksvc:
I tried to use SCSS variables to reduce the number of places where numbers are entered, to ease maintenance. I ran into problems and eventually just went with numbers instead of variables. If that's not okay here, I can go experiment more with SCSS.

marksvc:
See 1 minute animation showing daughter project list width, mother / Based on  project list width, and a really long project name, in both wider displays and narrow displays.

connect-width

nigel-wells (commented):
Added a couple of notes if they are of any interest. Also, does the based on under setting need to be looked at as well?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/843#-:-MK35bS59ph5CiLBFa63:bfyr4oj)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.scss, line 122 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/843#-MK33qq-1sawfCXiqF5_:-MK33qq00b1xN_sgCWsr:b-3g4rpx) ([raw file](https://github.com/sillsdev/web-xforge/blob/410a6e4367273c2c8ed59aa89e063b8310f1f22b/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.scss#L122)):*
> ```SCSS
>   // be the max-width of the div (512), but pulled in by 12px to stop the right side at the same place as
>   // on small displays.
>   min-width: min((512px - 12px), calc(100vw - 32px));
> ```

I do notice that if you remove the `margin-right: 32px` from `#project-select` that the select goes full width as I believe was its initial intention? Perhaps we need to add `calc(100% - 32px)` to `mdc-menu` to get the margin back in there as I believe it was added to stop the menu appearing over the down/up arrow of the selector?

marksvc (commented):
I had a look at the Settings component today. The drop down there is more complicated. Maybe we can see if it becomes a problem. It's less likely to since it won't show the "(Already connected)" suffix. Though it's also a bit more narrow presently. I can put more time into it if you think I should. I think I could at least get it to work with hardcoded widths per viewport.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/843#-:-MKCJWp318gjX6cIaay8:btqh1cl)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.scss, line 122 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/843#-MK33qq-1sawfCXiqF5_:-MKCISPB2whqk2gf-sZ4:b276e4q) ([raw file](https://github.com/sillsdev/web-xforge/blob/410a6e4367273c2c8ed59aa89e063b8310f1f22b/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.scss#L122)):*

I do notice that if you remove the `margin-right: 32px` from `#project-select` that the select goes full width as I believe was its initial intention? Perhaps we need to add `calc(100% - 32px)` to `mdc-menu` to get the margin back in there as I believe it was added to stop the menu appearing over the down/up arrow of the selector?
</blockquote></details>

It looks like the Settings component Based on drop down does go past where the arrow is (tho doesn't actually cover it up; it's underneath). Are you sure we don't want to cover the arrow?

I made some revisions.
The change is shortened a lot. And it gives space for the arrow.

Nateowami (commented):
Making the menu wider helps, but it can only go so far. Why not make the height of list items variable?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/843#-:-MKFQDNQ3RB64I_-woTR:b-qj7qxr)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

marksvc (commented):
I think if we make the list item heights not uniform, it would not look very nice.
We could make all list item heights the same, but still big enough for their contents, but then we might have large list item heights throughout (such as where we don't need it) and look odd. 
I did try working with list item heights at first, but found it to be complex. Maybe you can show me if you have an idea. I imagine we could figure it out by carefully building up the right flex layout parameters. But I suspect that keeping the current heights may be a better aesthetic. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/843#-:-MKLqrie8qIy80QPWFg3:b-f0h8k4)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

Nateowami (commented):
> I think if we make the list item heights not uniform, it would not look very nice.

I agree it's not ideal, but it would be dramatically better than it is right now. Alternatively we could provide a decent amount of space and then use an ellipsis to truncate the text. Here's how it looks on this branch for very long titles, and how it would look with variable height:

fix/connect-list | with `height: initial \| unset \| auto` on `mdc-list-item`
-----------------|-----------------------------------------------------------
 | 

I couldn't find an easy way to use `text-overflow: ellipsis`.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/843#-:-MKLxDeu6UKQRNnwP5cV:bo0rvsf)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

Nateowami (commented):
(Actually, to make the height variable properly, it would take setting `height: unset; min-height: 48px;`).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/843#-:-MKLzs7ZDL_kvw4HK0Gj:b-qw7z3i)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

marksvc (commented):
Thanks for your work on that. I was not approaching it the right way. 

By combining it with the current code, we support most situations horizontally, and exceptional situtaions by growing the height of a list item. For the connect project component.

I applied the height-growing to the settings component as well. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/843#-:-MKMNTK_1bJ8ryiUqv5x:byuizk3)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

irahopkinson (approved):
:lgtm:

> Reviewed 2 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/843#-:-MKNSmjH3QTJqMbIFjLF:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #844, opened 2020-10-20
SF-1097 - Reload and add question error

nigel-wells:
- Ensure question offline data has finished loading before updating any new offline data

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/844)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts
@@ -68,6 +68,7 @@ export class QuestionDoc extends ProjectDataDoc<Question> {
   protected async updateOfflineData(force: boolean = false): Promise<void> {
     // Check to see if any answers have been removed by comparing with current offline data
     if (this.realtimeService.offlineStore != null && this.realtimeService.fileService != null) {
+      await this.loadOfflineData();
       const offlineData = await this.realtimeService.offlineStore.get<RealtimeOfflineData>(this.collection, this.id);
       if (offlineData != null) {
         for (const answer of offlineData.data.answers) {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts
@@ -68,7 +68,6 @@ export class QuestionDoc extends ProjectDataDoc<Question> {
   protected async updateOfflineData(force: boolean = false): Promise<void> {
     // Check to see if any answers have been removed by comparing with current offline data
     if (this.realtimeService.offlineStore != null && this.realtimeService.fileService != null) {
-      await this.loadOfflineData();
       const offlineData = await this.realtimeService.offlineStore.get<RealtimeOfflineData>(this.collection, this.id);
       if (offlineData != null) {
         for (const answer of offlineData.data.answers) {
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts
@@ -19,6 +19,7 @@ export interface RealtimeDocConstructor {
  */
 export abstract class RealtimeDoc<T = any, Ops = any> {
   private updateOfflineDataSub: Subscription;
+  private updatingOfflineData: Promise<void> = Promise.resolve();
   private onDeleteSub: Subscription;
   private offlineSnapshotVersion?: number;
   private subscribePromise?: Promise<void>;
@@ -201,10 +202,13 @@ export abstract class RealtimeDoc<T = any, Ops = any> {
       type: this.adapter.type.name,
       pendingOps
     };
-    await this.realtimeService.offlineStore.put(this.collection, offlineData);
+    this.updatingOfflineData = new Promise<void>(async resolve => {
+      await this.realtimeService.offlineStore.put(this.collection, offlineData);
+      resolve();
+    });
   }
 
-  protected async loadOfflineData(): Promise<void> {
+  private async loadOfflineData(): Promise<void> {
     if (this.loadOfflineDataPromise == null) {
       this.loadOfflineDataPromise = this.loadFromOfflineStore();
     }
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts
@@ -204,7 +204,7 @@ export abstract class RealtimeDoc<T = any, Ops = any> {
     await this.realtimeService.offlineStore.put(this.collection, offlineData);
   }
 
-  private async loadOfflineData(): Promise<void> {
+  protected async loadOfflineData(): Promise<void> {
     if (this.loadOfflineDataPromise == null) {
       this.loadOfflineDataPromise = this.loadFromOfflineStore();
     }
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts
@@ -19,7 +19,6 @@ export interface RealtimeDocConstructor {
  */
 export abstract class RealtimeDoc<T = any, Ops = any> {
   private updateOfflineDataSub: Subscription;
-  private updatingOfflineData: Promise<void> = Promise.resolve();
   private onDeleteSub: Subscription;
   private offlineSnapshotVersion?: number;
   private subscribePromise?: Promise<void>;
@@ -115,8 +114,8 @@ export abstract class RealtimeDoc<T = any, Ops = any> {
 
   async create(data: T): Promise<void> {
     this.adapter.create(data).then(() => this.updateOfflineData(true));
-    await this.updateOfflineData(true);
     this.loadOfflineDataPromise = Promise.resolve();
+    await this.updateOfflineData(true);
     await this.realtimeService.onLocalDocUpdate(this);
   }
 
@@ -202,10 +201,7 @@ export abstract class RealtimeDoc<T = any, Ops = any> {
       type: this.adapter.type.name,
       pendingOps
     };
-    this.updatingOfflineData = new Promise<void>(async resolve => {
-      await this.realtimeService.offlineStore.put(this.collection, offlineData);
-      resolve();
-    });
+    await this.realtimeService.offlineStore.put(this.collection, offlineData);
   }
 
   private async loadOfflineData(): Promise<void> {


Nateowami (commented):
Tested it out and experimentally I can confirm that this fixes the issue. I haven't reviewed the code though (I see it's a simple change, but I haven't looked into the underlying issue, and I've reached the end of the day, so I think I'll leave code review to someone else).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/844#-:-MK2o2JNBrkzU_ZUXLG8:b14tq5m)* status: 0 of 2 files reviewed, all discussions resolved

irahopkinson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/844)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts, line 71 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/844#-MK4QOX64XSDjPSbEWSD:-MK4QOX64XSDjPSbEWSE:b279a7v) ([raw file](https://github.com/sillsdev/web-xforge/blob/a65fa04b5518bee76c18ecdc9a7970102db0cced/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts#L71)):*
> ```
> await this.loadOfflineData();
> ```
Can you explain why this fixes the problem?
E.g. if the following code in the `if` causes the error then this is in the right place. However, if its the call to `await super.updateOfflineData(force);` or `await this.updateFileCache();` that causes the issue then it probably needs to be further up in the hierarchy so that it fixes it for any doc.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/844)* status: all files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts, line 71 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/844#-MK4QOX64XSDjPSbEWSD:-MK7RndQ0G6gHcYmRjvo:bveccut) ([raw file](https://github.com/sillsdev/web-xforge/blob/a65fa04b5518bee76c18ecdc9a7970102db0cced/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts#L71)):*

> ```
> await this.loadOfflineData();
> ```
Can you explain why this fixes the problem?
E.g. if the following code in the `if` causes the error then this is in the right place. However, if its the call to `await super.updateOfflineData(force);` or `await this.updateFileCache();` that causes the issue then it probably needs to be further up in the hierarchy so that it fixes it for any doc.
</blockquote></details>

In hindsight I should have copied across the discussion from Slack. The issue is a race condition and it is the `await` on `const offlineData = await this.realtimeService.offlineStore.get<RealtimeOfflineData>(this.collection, this.id);` which is causing the "issue". The purpose of the code in `question-doc.ts` is to check offline storage before it gets overwritten with anything else so it needs to run before the normal `updateOfflineData()` is run in `realtime-doc.ts`. The actual error is triggered from `loadFromOfflineStore()` in `realtime-doc.ts` where it gets a doc from offline storage and if there is data with no version then it assumes it needs to be created. We know it doesn't need to be created as that is already in the process of taking place. `updateOfflineData()` gets called under lots of different circumstances but `loadFromOfflineStore()` only gets called once. Everywhere `loadOfflineData()` is called (by subscribing directly or via a query) it always needs the promise to resolve before resolving the subscription. What I beleive I've done here in `question-doc.ts` is the same thing - wait for it to finish before querying it.

irahopkinson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/844)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts, line 71 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/844#-MK4QOX64XSDjPSbEWSD:-MK8LrAME6sjMONbUTH_:b-j7lggf) ([raw file](https://github.com/sillsdev/web-xforge/blob/a65fa04b5518bee76c18ecdc9a7970102db0cced/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts#L71)):*

In hindsight I should have copied across the discussion from Slack. The issue is a race condition and it is the `await` on `const offlineData = await this.realtimeService.offlineStore.get<RealtimeOfflineData>(this.collection, this.id);` which is causing the "issue". The purpose of the code in `question-doc.ts` is to check offline storage before it gets overwritten with anything else so it needs to run before the normal `updateOfflineData()` is run in `realtime-doc.ts`. The actual error is triggered from `loadFromOfflineStore()` in `realtime-doc.ts` where it gets a doc from offline storage and if there is data with no version then it assumes it needs to be created. We know it doesn't need to be created as that is already in the process of taking place. `updateOfflineData()` gets called under lots of different circumstances but `loadFromOfflineStore()` only gets called once. Everywhere `loadOfflineData()` is called (by subscribing directly or via a query) it always needs the promise to resolve before resolving the subscription. What I beleive I've done here in `question-doc.ts` is the same thing - wait for it to finish before querying it.
</blockquote></details>

As per our GMeet call, please dig some more until we can identify the root cause.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/844)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts, line 71 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/844#-MK4QOX64XSDjPSbEWSD:-MKDBW3X49uex_RZ69E-:b-ubkg5h) ([raw file](https://github.com/sillsdev/web-xforge/blob/a65fa04b5518bee76c18ecdc9a7970102db0cced/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts#L71)):*

As per our GMeet call, please dig some more until we can identify the root cause.
</blockquote></details>

Through much logging I can now see the race condition happening and fix the actual problem. Here is what's happening:
1. Document is submitted to be created via ShareDB
2. Realtime doc immediately creates an entry in IndexDB without waiting for ShareDB to respond - no version is added at this point and it is saved in IndexDB as `null`
3. ShareDB responds that the document has been created
4. Realtime doc proceeds to update IndexDB using an `await`. After IndexDB is updated the `loadOfflineDataPromise` is resolved as there is no need to load data for a freshly created document
5. At the same time in point 3 ShareDB emits an event to the Realtime query which triggers `onAddedToSubscribeQuery` and this immediately calls `loadOfflineData()` .  As point 4 is still in the process of updating IndexDB it goes ahead and tries to load the offline data. At this point the version number is still `null` as point 4 still isn't completed so we therefore try and create the document again in ShareDB which then errors out.

My fix for this is to create a promise in `updateOfflineData()` so that `onAddedToSubscribeQuery()` knows to hold fire if something is currently in progress. Once `updateOfflineData()` finishes then the `create()` will resolve the need to worry about loading offline data and the issue is now resolved.

ddaspit (changes requested):
Great job tracking down the root cause. I have a suggestion for a simpler fix.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/844#-:-MKDONHg7rCKM-5SaMni:b92yaw7)* status: all files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts, line 119 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/844#-MKDO8Im5ULgcZLAX4jQ:-MKDO8Im5ULgcZLAX4jR:bjuitru) ([raw file](https://github.com/sillsdev/web-xforge/blob/99ca2858f33f4806f6a87505b668f1a249c4eeeb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts#L119)):*
> ```XML
>     this.adapter.create(data).then(() => this.updateOfflineData(true));
>     await this.updateOfflineData(true);
>     this.loadOfflineDataPromise = Promise.resolve();
> ```

I think an easier fix would be to move this line to the top of the function.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/844)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit and @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts, line 119 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/844#-MKDO8Im5ULgcZLAX4jQ:-MKDPHzb114BQz3wTfgo:b-bkkqe4) ([raw file](https://github.com/sillsdev/web-xforge/blob/99ca2858f33f4806f6a87505b668f1a249c4eeeb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts#L119)):*

I think an easier fix would be to move this line to the top of the function.
</blockquote></details>

I did consider that but it wouldn't be true if we moved it up as the offline data hasn't actually finished being updated yet which I figure could potentially cause other rollon affects where there is code waiting for that promise to resolve.

ddaspit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/844)* status: all files reviewed, 3 unresolved discussions (waiting on @irahopkinson and @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts, line 119 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/844#-MKDO8Im5ULgcZLAX4jQ:-MKENs4X1PooaqzFFCyx:b-falsz3) ([raw file](https://github.com/sillsdev/web-xforge/blob/99ca2858f33f4806f6a87505b668f1a249c4eeeb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts#L119)):*

I did consider that but it wouldn't be true if we moved it up as the offline data hasn't actually finished being updated yet which I figure could potentially cause other rollon affects where there is code waiting for that promise to resolve.
</blockquote></details>

That is a good point. What about if we do the following:
```
this.loadOfflineDataPromise = this.updateOfflineData(true);
await this.loadOfflineDataPromise;
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts, line 205 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/844#-MKEOFwpBPaoXqpafBi9:-MKEOFwpBPaoXqpafBiA:bcocmsi) ([raw file](https://github.com/sillsdev/web-xforge/blob/99ca2858f33f4806f6a87505b668f1a249c4eeeb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts#L205)):*
> ```XML
>       pendingOps
>     };
>     this.updatingOfflineData = new Promise<void>(async resolve => {
> ```

Does this fix the problem because it doesn't await the call to the offline store that inserts the offline data? I feel like this could have unintended consequences elsewhere.

ddaspit (dismissed):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/844)* status: all files reviewed, 3 unresolved discussions (waiting on @irahopkinson and @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts, line 119 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/844#-MKDO8Im5ULgcZLAX4jQ:-MKEairVAT0ZJxycrm_B:bllqr2k) ([raw file](https://github.com/sillsdev/web-xforge/blob/99ca2858f33f4806f6a87505b668f1a249c4eeeb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts#L119)):*

That is a good point. What about if we do the following:
```
this.loadOfflineDataPromise = this.updateOfflineData(true);
await this.loadOfflineDataPromise;
```
</blockquote></details>

I thought about this a little more. I think all we need to do is move this assignment up one line of code, like this:
```
this.adapter.create(data).then(() => this.updateOfflineData(true));
this.loadOfflineDataPromise = Promise.resolve();
await this.updateOfflineData(true);
```
The main point of the promise is that it is resolved after the doc data has been populated. The `create` call has the effect of populating the doc data, so I think we can resolve it right after that call is made.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/844)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit and @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts, line 119 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/844#-MKDO8Im5ULgcZLAX4jQ:-MKH9s4_2Yi4BSgYgU5n:b70h7gn) ([raw file](https://github.com/sillsdev/web-xforge/blob/99ca2858f33f4806f6a87505b668f1a249c4eeeb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts#L119)):*

I thought about this a little more. I think all we need to do is move this assignment up one line of code, like this:
```
this.adapter.create(data).then(() => this.updateOfflineData(true));
this.loadOfflineDataPromise = Promise.resolve();
await this.updateOfflineData(true);
```
The main point of the promise is that it is resolved after the doc data has been populated. The `create` call has the effect of populating the doc data, so I think we can resolve it right after that call is made.

</blockquote></details>

Yes that does resolve the issue even though, in one sense, it doesn't "feel" correct i.e. the offline data hasn't been fully updated yet but, yes, the purpose of this promise is fulfilled as there is at least some offline data available by this point. I've done a bunch of additional tests to see if there was a potential flow on effect with IndexDB taking a while to update and I couldn't see any issues.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts, line 205 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/844#-MKEOFwpBPaoXqpafBi9:-MKH9d8_71GMpAaiowJA:bj5dvoe) ([raw file](https://github.com/sillsdev/web-xforge/blob/99ca2858f33f4806f6a87505b668f1a249c4eeeb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts#L205)):*

Does this fix the problem because it doesn't await the call to the offline store that inserts the offline data? I feel like this could have unintended consequences elsewhere.
</blockquote></details>

Whoops! Just before committing I reverted the `protected` back to `private` on `loadOfflineData()` which also removed my `await`. Not needed now anyway as per other discussions.

irahopkinson (approved):
:lgtm:

> Reviewed 1 of 2 files at r2, 1 of 1 files at r3.
> Dismissed @ddaspit from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/844#-:-MKIUMAvBgxOBnPoYssh:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts, line 119 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/844#-MKDO8Im5ULgcZLAX4jQ:-MKITUYv6Egnpy8QPOBr:b-jqtaga) ([raw file](https://github.com/sillsdev/web-xforge/blob/99ca2858f33f4806f6a87505b668f1a249c4eeeb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts#L119)):*

Yes that does resolve the issue even though, in one sense, it doesn't "feel" correct i.e. the offline data hasn't been fully updated yet but, yes, the purpose of this promise is fulfilled as there is at least some offline data available by this point. I've done a bunch of additional tests to see if there was a potential flow on effect with IndexDB taking a while to update and I couldn't see any issues.
</blockquote></details>

However, I think it is actually correct as the promise is labelled `loadOfflineDataPromise`. If it was labeled `updateOfflineDataPromise` then I think it needed to be in the original location and dealt with differently to correctly reflect its label. Remember that create is a special case in that there is no new data to load other than what was passsed in.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/844)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts, line 119 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/844#-MKDO8Im5ULgcZLAX4jQ:-MKIUmKO24lwun2Hav7L:b-ofsnz5) ([raw file](https://github.com/sillsdev/web-xforge/blob/99ca2858f33f4806f6a87505b668f1a249c4eeeb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts#L119)):*

However, I think it is actually correct as the promise is labelled `loadOfflineDataPromise`. If it was labeled `updateOfflineDataPromise` then I think it needed to be in the original location and dealt with differently to correctly reflect its label. Remember that create is a special case in that there is no new data to load other than what was passsed in.
</blockquote></details>

Yes good point - I "feel" happy now ;)

PR #846, opened 2020-10-20
Fix clickability of help menu list items

Nateowami:
A couple list items in the help menu were not fully clickable (some regions of the list item just closed the menu when clicked). Here's a visualization of the element structure created using Pesticide CSS:

Before | After
-------|------
 | 

I've made two changes:
1. The entire region of each of the top two list items is now clickable.
2. I changed `<div class="mdc-list-group__subheader">` to `<div mdcListGroupSubheader>` because it seemed more consistent with the way we use Angular MDC.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/846)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html
@@ -49,21 +49,17 @@
             <mdc-menu #helpMenu [anchorCorner]="'bottomStart'" [anchorElement]="helpMenuAnchor">
               <mdc-list-group>
                 <mdc-list>
-                  <mdc-list-item>
-                    <a target="_blank" [href]="helpsPage">{{ t("help") }}</a>
-                  </mdc-list-item>
-                  <mdc-list-item name="issues">
-                    <a target="_blank" [href]="issueMailTo">
-                      <mdc-list-item-text>
-                        {{ t("report_issue") }}
-                        <div>
-                          <small>{{ t("report_issue_email", { email: issueEmail }) }}</small>
-                        </div>
-                      </mdc-list-item-text>
-                    </a>
-                  </mdc-list-item>
+                  <a mdc-list-item target="_blank" [href]="helpsPage">{{ t("help") }}</a>
+                  <a mdc-list-item target="_blank" [href]="issueMailTo">
+                    <mdc-list-item-text
+                      >{{ t("report_issue") }}
+                      <mdc-list-item-secondary
+                        >{{ t("report_issue_email", { email: issueEmail }) }}
+                      </mdc-list-item-secondary>
+                    </mdc-list-item-text>
+                  </a>
                   <mdc-list-divider></mdc-list-divider>
-                  <div class="mdc-list-group__subheader">{{ t("product_version", { version: version }) }}</div>
+                  <div mdcListGroupSubheader class="version">{{ t("product_version", { version: version }) }}</div>
                 </mdc-list>
               </mdc-list-group>
             </mdc-menu>
@@ -87,11 +83,8 @@
                 <div class="mdc-list-group__subheader">
                   {{ t("logged_in_as") }}
                   <div fxLayout="row" fxLayoutAlign="start center">
-                    <span fxFlex>&nbsp;</span>
-                    <button id="edit-name-btn" mdcIconButton type="button" (click)="editName()">
-                      <mdc-icon>edit</mdc-icon>
-                    </button>
                     <strong class="user-menu-name">{{ currentUser?.displayName }}</strong>
+                    <button id="edit-name-btn" mdcIconButton icon="edit" type="button" (click)="editName()"></button>
                   </div>
                 </div>
                 <div>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
@@ -83,9 +83,13 @@
     mdc-top-app-bar-section {
       mdc-menu {
         min-width: 200px;
+        text-align: left; // Needed for some menu items in Chrome but not Firefox
         a {
           width: 100%;
           color: var(--mdc-theme-text-primary-on-background);
+        }
+        .version {
+          color: var(--mdc-theme-text-secondary-on-background);
         }
       }
     }


irahopkinson (changes requested):
Personnaly I don't see any difference in Windows Chrome. I expect you are seeing some problem in Linux Firefox. However, I don't mind this change except that in Windows Chrome you will need to add some style support. Before it looked like this:

On your branch it has become inconsistent:

Ideally, to keep it consistent with other menus, it would look like this:

Where clickable items are normal aligned (left on LTR, right on RTL) and non-clickable items are center or opposite aligned. I see that in Windows Firefox everything is normal aligned (that's fine too).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/846#-:-MK7zlAs9rLSYDuGzBci:bg9liuu)* status: :shipit: complete! all files reviewed, all discussions resolved

Nateowami (commented):
Good catch; I'll fix the alignment.

> Personnaly I don't see any difference in Windows Chrome.

Do you mean see (visually) or see (detect)? There isn't supposed to be a visible change (though as you've pointed out there is one). It was a problem of functionality (which I have reproduced in Firefox and Chromium on Linux). The screenshots I showed were only to give a visualization of the layout of elements that are not visually distinguishable.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/846#-:-MK89u6c4sOCKI2MNGHo:bl744yc)* status: :shipit: complete! all files reviewed, all discussions resolved

irahopkinson (changes requested):
Yes, as in detect. Thanks for that, a little bit more playing and I can "see" the issue in Windows also. I think I was expecting the cursor to change when it was clickable but that is the actual issue in that the cursor doesn't change but you can get different result depending on where you click.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/846#-:-MK8OSm_8i8rmlyz5TVI:b-cd0aqk)* status: :shipit: complete! all files reviewed, all discussions resolved

Nateowami (commented):
I've made a couple changes:

- The right alignment discrepancy is in Chrome but not Firefox and has existed for as long as I can remember. I've never known why, and my investigation today (removing a ton of styles expecting it to go away) failed to come up with any reason. I found it was simpler to fix by adding a single CSS rule.
- Alignment is also changed in the user menu and is now consistent with Firefox.
- I fixed a minute alignment issue caused by a space before "Report an issue"
- The product version is now a disabled looking color. This change is debatable, but I thought it helped convey that the element is not interactive.
- The styling of `email scriptureforgeissues@sil.org` is now slightly different because it uses MDC's option for list item secondary text.

Before | After
-------|------
 | .png)
.png) | .png)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/846#-:-MK9zNjZ-ieJiKu6KjWg:b27jlb2)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @irahopkinson)

irahopkinson (changes requested):
Since you have gone for normal alignment for all items it would make sense to now move the pencil to the other end of the users display name. What do you think?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/846#-:-MKA5jD27TV-dfVcJ_hW:b-kybyv)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html, line 55 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/846#-MKA4nNE8Tx2zSOAfSg3:-MKA4nNF7nDMehQWDY92:b-s2gqwe) ([raw file](https://github.com/sillsdev/web-xforge/blob/2feeafd45bdf17fcb917d085c06fa5b85117f9f5/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L55)):*
> ```
> >
> ```
nit: this should be with the line above.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html, line 57 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/846#-MKA4u_-0Rqhe9EosmML:-MKA4u_-0Rqhe9EosmMM:b-s2gqwe) ([raw file](https://github.com/sillsdev/web-xforge/blob/2feeafd45bdf17fcb917d085c06fa5b85117f9f5/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L57)):*
> ```
> >
> ```
nit: this should be with the line above.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss, line 86 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/846#-MKA53vwDvieSb6a4fjO:-MKA53vwDvieSb6a4fjP:b61uhh) ([raw file](https://github.com/sillsdev/web-xforge/blob/2feeafd45bdf17fcb917d085c06fa5b85117f9f5/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L86)):*
> ```
> left
> ```
This probably won't work for RTL interface languages. Maybe we leave that until we have one?

Nateowami (commented):
Done.


.png)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/846#-:-MKA8Nox-NqC_xW1ttwc:b49w1pu)* status: 1 of 2 files reviewed, 3 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html, line 55 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/846#-MKA4nNE8Tx2zSOAfSg3:-MKA6wR67cIzjB1VUX_H:bvpr7gh) ([raw file](https://github.com/sillsdev/web-xforge/blob/2feeafd45bdf17fcb917d085c06fa5b85117f9f5/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L55)):*

> ```
> >
> ```
nit: this should be with the line above.
</blockquote></details>

Actually this is intentional. If it were on the line above that would introduce a space at the beginning of the text, which breaks alignment with other list items. Prettier is smart enough to not insert spaces where there are none and so formats it like this.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html, line 57 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/846#-MKA4u_-0Rqhe9EosmML:-MKA7R9sDprRSyXpcHs2:b2nreh6) ([raw file](https://github.com/sillsdev/web-xforge/blob/2feeafd45bdf17fcb917d085c06fa5b85117f9f5/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L57)):*

> ```
> >
> ```
nit: this should be with the line above.
</blockquote></details>

Same as above.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss, line 86 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/846#-MKA53vwDvieSb6a4fjO:-MKA7SmP1Tmzl5PwBZNm:b-mg9hay) ([raw file](https://github.com/sillsdev/web-xforge/blob/2feeafd45bdf17fcb917d085c06fa5b85117f9f5/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L86)):*

> ```
> left
> ```
This probably won't work for RTL interface languages. Maybe we leave that until we have one?
</blockquote></details>

True, but my guess would be that it already didn't work correctly for rtl languages, so I think we should wait until we have one.

irahopkinson (approved):
:lgtm: thanks for sorting all this out.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/846#-:-MKAEsVE7qNwIPm-tgrr:bmbx8w8)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html, line 55 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/846#-MKA4nNE8Tx2zSOAfSg3:-MKAEksy9ZgklL7Tdyh-:b-ahle5g) ([raw file](https://github.com/sillsdev/web-xforge/blob/2feeafd45bdf17fcb917d085c06fa5b85117f9f5/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L55)):*

Actually this is intentional. If it were on the line above that would introduce a space at the beginning of the text, which breaks alignment with other list items. Prettier is smart enough to not insert spaces where there are none and so formats it like this.
</blockquote></details>

Good to know.

PR #848, opened 2020-10-22
SF-1107 Access answer audio through cache

RaymondLuong3:
* The answers component loads audio content through the cache
* Audio cache gets updated anytime local audio is uploaded
* When browser or add-to-homescreen is re-opened it will play cached answers

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/848)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html
@@ -66,7 +66,7 @@
           </div>
           <div [fxShow]="answerTabs.activeTabIndex === 1" class="answer-tab answer-record-upload">
             <app-checking-audio-combined
-              [source]="activeAnswer?.audioUrl"
+              [source]="answerUrls[activeAnswer?.dataId]"
               (update)="processAudio($event)"
             ></app-checking-audio-combined>
           </div>
@@ -138,7 +138,8 @@
                 }}</span>
                 <span class="answer-scripture-verse">{{ scriptureTextVerseRef(answer.verseRef) }}</span>
               </div>
-              <app-checking-audio-player *ngIf="answer.audioUrl" [source]="answer.audioUrl"></app-checking-audio-player>
+              <app-checking-audio-player *ngIf="answer.audioUrl" [source]="answerUrls[answer.dataId]">
+              </app-checking-audio-player>
               <div class="answer-footer">
                 <div class="actions">
                   <button
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html
@@ -11,8 +11,9 @@
       >
         <div class="question">
           <div class="question-text">{{ questionDoc?.data?.text }}</div>
-          <div *ngIf="questionUrl" class="question-audio">
-            <app-checking-audio-player [source]="questionUrl"></app-checking-audio-player>
+          <div *ngIf="questionDoc?.data?.audioUrl" class="question-audio">
+            <app-checking-audio-player [source]="getFileSource(questionDoc?.data?.audioUrl)">
+            </app-checking-audio-player>
           </div>
           <div class="question-footer" *ngIf="isProjectAdmin">
             <div class="actions">
@@ -66,7 +67,7 @@
           </div>
           <div [fxShow]="answerTabs.activeTabIndex === 1" class="answer-tab answer-record-upload">
             <app-checking-audio-combined
-              [source]="answerUrls[activeAnswer?.dataId]"
+              [source]="getFileSource(activeAnswer?.audioUrl)"
               (update)="processAudio($event)"
             ></app-checking-audio-combined>
           </div>
@@ -138,7 +139,7 @@
                 }}</span>
                 <span class="answer-scripture-verse">{{ scriptureTextVerseRef(answer.verseRef) }}</span>
               </div>
-              <app-checking-audio-player *ngIf="answer.audioUrl" [source]="answerUrls[answer.dataId]">
+              <app-checking-audio-player *ngIf="answer.audioUrl" [source]="getFileSource(answer.audioUrl)">
               </app-checking-audio-player>
               <div class="answer-footer">
                 <div class="actions">
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts
@@ -323,10 +323,7 @@ export class CheckingAnswersComponent extends SubscriptionDisposable implements 
   }
 
   getFileSource(url: string | undefined): string | undefined {
-    if (url == null || url === '') {
-      return undefined;
-    }
-    if (this.fileSources.has(url)) {
+    if (url != null && url !== '' && this.fileSources.has(url)) {
       return this.fileSources.get(url);
     }
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts
@@ -86,12 +86,12 @@ export class CheckingAnswersComponent extends SubscriptionDisposable implements 
     if (questionDoc == null) {
       return;
     }
-    this.updateQuestionAudioUrl();
+    this.updateQuestionDocAudioUrls();
     if (this.questionChangeSubscription != null) {
       this.questionChangeSubscription!.unsubscribe();
     }
     this.questionChangeSubscription = this.subscribe(questionDoc.remoteChanges$, ops => {
-      this.updateQuestionAudioUrl();
+      this.updateQuestionDocAudioUrls();
       // If the user hasn't added an answer yet and is able to, then
       // don't hold back any incoming answers from appearing right away
       // as soon as the user adds their answer.
@@ -131,6 +131,7 @@ export class CheckingAnswersComponent extends SubscriptionDisposable implements 
   verseRef?: VerseRef;
   answersHighlightStatus: Map<string, boolean> = new Map<string, boolean>();
   questionUrl?: string;
+  answerUrls: { [answerId: string]: string } = {};
 
   /** IDs of answers to show to user (so, excluding unshown incoming answers). */
   private _answersToShow: string[] = [];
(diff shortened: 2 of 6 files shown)


nigel-wells (changes requested):
Looking good - just a couple of thoughts below.

> Reviewed 6 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/848#-:-MKHYbSw6I5DAX_djqf_:bt9vnej)* status: all files reviewed, 4 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts, line 134 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/848#-MKHTqsA0L4rkhUimNOy:-MKHTqsA0L4rkhUimNOz:b-6z0oai) ([raw file](https://github.com/sillsdev/web-xforge/blob/b88ac0f01df9274d2dee820fae893aeae1db69b7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L134)):*
> ```XML
>   answersHighlightStatus: Map<string, boolean> = new Map<string, boolean>();
>   questionUrl?: string;
>   answerUrls: { [answerId: string]: string } = {};
> ```

Perhaps we could use a `Map` instead as that is what we tend to use throughout the rest of our codebase
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts, line 332 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/848#-MKHSsy81RHU2k6wSeXd:-MKHSsy81RHU2k6wSeXe:bcqugi5) ([raw file](https://github.com/sillsdev/web-xforge/blob/b88ac0f01df9274d2dee820fae893aeae1db69b7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L332)):*
> ```XML
>       return;
>     }
>     const blob = await this.questionDoc.getFileContents(FileType.Audio, this.questionDoc.data.dataId);
> ```

With the above check no longer looking at `data.audioUrl` for `null` then we'll want to add it in here or potentially move `this.answerUrls()` further up
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts, line 348 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/848#-MKHUW3Z1HSIGhoQ2E_7:-MKHUW3Z1HSIGhoQ2E_8:b-fd7ecf) ([raw file](https://github.com/sillsdev/web-xforge/blob/b88ac0f01df9274d2dee820fae893aeae1db69b7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L348)):*
> ```XML
>       const audio: Blob | undefined = await this.questionDoc.getFileContents(FileType.Audio, answer.dataId);
>       // default to the answer's audioUrl to show the audio unavailable icon when offline and audio is not cached
>       this.answerUrls[answer.dataId] = audio == null ? answer.audioUrl : URL.createObjectURL(audio);
> ```

What do you think of the idea of having one collection of file URLs and their source? This way it could work for audio files, question files, and potentially anything else in the future i.e. attachments. You could then have one function that does these checks which are mostly the same as the question ones above.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-player/checking-audio-player.component.ts, line 52 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/848#-MKHVKrT5ooW-ML4CsJE:-MKHVKrT5ooW-ML4CsJF:b6kng8k) ([raw file](https://github.com/sillsdev/web-xforge/blob/b88ac0f01df9274d2dee820fae893aeae1db69b7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-player/checking-audio-player.component.ts#L52)):*
> ```XML
> 
>     this.audio.addEventListener('error', () => {
>       this.localUrlError = isLocalBlobUrl(this.audio.src) ? true : false;
> ```

What are the instances where the `HTMLAudioElement` would emit an error? Couldn't an error still happen even if it happens to be locally cached?

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/848)* status: all files reviewed, 4 unresolved discussions (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts, line 134 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/848#-MKHTqsA0L4rkhUimNOy:-MKaMgwd4C6XEV89R1Qy:b-g32f8b) ([raw file](https://github.com/sillsdev/web-xforge/blob/b88ac0f01df9274d2dee820fae893aeae1db69b7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L134)):*

Perhaps we could use a `Map` instead as that is what we tend to use throughout the rest of our codebase
</blockquote></details>

Ok, I've made a revision to use a map. Good thinking.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts, line 332 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/848#-MKHSsy81RHU2k6wSeXd:-MKaMmVT2-jawYdQgQIq:b6f9fb9) ([raw file](https://github.com/sillsdev/web-xforge/blob/b88ac0f01df9274d2dee820fae893aeae1db69b7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L332)):*

With the above check no longer looking at `data.audioUrl` for `null` then we'll want to add it in here or potentially move `this.answerUrls()` further up
</blockquote></details>

It seems that we do want to call `getFileContents()` regardless of whether `audioUrl` is or is not null so that indexedDB can be updated.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts, line 348 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/848#-MKHUW3Z1HSIGhoQ2E_7:-MKaNDx05tkPEX9hFHhc:b-fgd6yg) ([raw file](https://github.com/sillsdev/web-xforge/blob/b88ac0f01df9274d2dee820fae893aeae1db69b7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L348)):*

What do you think of the idea of having one collection of file URLs and their source? This way it could work for audio files, question files, and potentially anything else in the future i.e. attachments. You could then have one function that does these checks which are mostly the same as the question ones above.
</blockquote></details>

That's a good idea. I'll make the refactor.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-player/checking-audio-player.component.ts, line 52 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/848#-MKHVKrT5ooW-ML4CsJE:-MKaNMdp0b85ZA9un9gX:bnu64ty) ([raw file](https://github.com/sillsdev/web-xforge/blob/b88ac0f01df9274d2dee820fae893aeae1db69b7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-player/checking-audio-player.component.ts#L52)):*

What are the instances where the `HTMLAudioElement` would emit an error? Couldn't an error still happen even if it happens to be locally cached?
</blockquote></details>

I can't say that I know the full list of things that will cause the error event, but this is primarily so that we can set `localUrlError`. If we care about errors coming from content on the server, that can be added later.

nigel-wells (approved):
Yes I like what you've done there - looks much cleaner now as well. Just one minor non blocking point.

> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/848#-:-MKbQ-Or5WztM0lH02yC:bnu30kc)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts, line 326 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/848#-MKbNOD05V0tGLeldbIQ:-MKbNOD17imuBx-REQGG:b-1g51pm) ([raw file](https://github.com/sillsdev/web-xforge/blob/a00a1647ac39eefa84638e249e9aa827129e64e7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L326)):*
> ```XML
> 
>   getFileSource(url: string | undefined): string | undefined {
>     if (url == null || url === '') {
> ```

Perhaps these checks can simply be part of the `if` statement below. `undefined` is already the case if nothing was found in `this.flieSources.has()` so no need to return it specifically
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-player/checking-audio-player.component.ts, line 52 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/848#-MKHVKrT5ooW-ML4CsJE:-MKbPFq60wCcnBYWlL7I:b-epykql) ([raw file](https://github.com/sillsdev/web-xforge/blob/b88ac0f01df9274d2dee820fae893aeae1db69b7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-player/checking-audio-player.component.ts#L52)):*

I can't say that I know the full list of things that will cause the error event, but this is primarily so that we can set `localUrlError`. If we care about errors coming from content on the server, that can be added later.
</blockquote></details>

Ah right, sorry, I thought you were changing an existing `error` check but this is new code that you've introduced.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/848)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts, line 326 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/848#-MKbNOD05V0tGLeldbIQ:-MKebo84BrwsugG0JvFP:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/a00a1647ac39eefa84638e249e9aa827129e64e7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L326)):*

Perhaps these checks can simply be part of the `if` statement below. `undefined` is already the case if nothing was found in `this.flieSources.has()` so no need to return it specifically
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-player/checking-audio-player.component.ts, line 52 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/848#-MKHVKrT5ooW-ML4CsJE:-MKebotK1Cmk23txm08L:b-coeb0f) ([raw file](https://github.com/sillsdev/web-xforge/blob/b88ac0f01df9274d2dee820fae893aeae1db69b7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-player/checking-audio-player.component.ts#L52)):*

Ah right, sorry, I thought you were changing an existing `error` check but this is new code that you've introduced.
</blockquote></details>

Not a problem :)

PR #849, opened 2020-10-22
Setup: Add WritingSystemRepository, enhance

marksvc:
While testing SF-1047 (server Ansible for ParatextData) on a dev machine by adding some Ansible into our existing dev machine Ansible files, the WritingSystemRepository directory was found to be something our dev machine ansible wasn't covering already. This PR adds that in. 
And also a cleanup while working in that area.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/849)

### DIFF
--- a/deploy/dependencies.yml
+++ b/deploy/dependencies.yml
@@ -110,6 +110,7 @@
         - "/var/lib/scriptureforge/audio"
         - "/var/lib/xforge"
         - "/var/lib/xforge/avatars"
+        - "{{lookup('env','HOME')}}/.local/share/SIL/WritingSystemRepository/3"
 
     - name: add localhost to dnsmasq
       lineinfile:
--- a/deploy/dependencies.yml
+++ b/deploy/dependencies.yml
@@ -63,8 +63,14 @@
         state: latest
         global: yes
 
-    - name: change node version
-      shell: "n {{node_version}}"
+    - name: check current node version
+      shell: node --version || echo error
+      register: node_version_installed
+      changed_when: False
+
+    - name: set node version
+      command: "n {{node_version}}"
+      when: "node_version_installed.stdout != 'v' + node_version"
 
     # Install mercurial 4.7+ from system or pip
     - name: install pip mercurial dependencies


irahopkinson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/849)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[deploy/dependencies.yml, line 67 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/849#-MKIoLjQ-BXGSNVuIyMK:-MKIoLjQ-BXGSNVuIyML:b-r3a0ga) ([raw file](https://github.com/sillsdev/web-xforge/blob/f0528f88fe1cf3dd6251bfbad6ae8413f0dfb0a4/deploy/dependencies.yml#L67)):*
> ```
> n which {{node_version}}
> ```
I appreciate your attempt to get this idempotent as I wasn't able to. The problem with `n which` is it is only checking that that version is installed but we want to check which one of the installed versions is being used. If another version is currently being used but the one we want is installed then this code won't change to the version we want. You could use `n prune` after `n {{node_version}}` but there is no guarantee that a user hasn't manually changed versions. It seemed safer to me to change to the version we want (even if we are already on that version) than to not change.

Given that we primarily use Ansible to document the setup with a secondary benefit that it also gives us a from-scratch automation, and that we **don't** support it changing from one setup to another, then the fact that `n {{node_version}}` is not idempotent is practically irrelevant. Typically idempotency is an important part of Ansible and worth while generally for us but it is not our use case.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/849)* status: all files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[deploy/dependencies.yml, line 67 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/849#-MKIoLjQ-BXGSNVuIyMK:-MKMDe2E-H2RdNhBrjyw:b-ezjh47) ([raw file](https://github.com/sillsdev/web-xforge/blob/f0528f88fe1cf3dd6251bfbad6ae8413f0dfb0a4/deploy/dependencies.yml#L67)):*
> which one of the installed versions is being used.

Ah ha. I was relying on querying n itself to tell me things like that. But after you mentioned this, I found that I can easily see what node version is being used by just running `node --version`, since `n NEW_VERSION` changes the /usr/local/bin/node file. 

I was confused about n only distinguishing between installed and not-installed, rather than accounting much for 'selected'. 

> Given that we primarily use Ansible to document the setup with a secondary benefit that it also gives us a from-scratch automation, and that we don't support it changing from one setup to another

I would say that our Ansible serves a significant role of configuring dev and test machines so they work. After we change our dotnet version, or node version, or add mp3 encoding support, etc, we can just instruct people to run the ansible playbook and it configures their dev or test machine. In my view, we have run the playbook to _change_ (or fix) a machine significantly more often than to initially set up a machine. 

But, I hear what you're saying, and it wasn't broken before. That said, your comment quickly landed me on a solution that I have updated the PR with. If that's good, we can go with that. But we don't need to send this part thru if you prefer not.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/849)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[deploy/dependencies.yml, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/849#-MKMHgbP4jMDrMeTxs-H:-MKMHgbP4jMDrMeTxs-I:b-hdhro8) ([raw file](https://github.com/sillsdev/web-xforge/blob/f0528f88fe1cf3dd6251bfbad6ae8413f0dfb0a4/deploy/dependencies.yml#L68)):*
> ```YAML
>     - name: check if have required node version
>       shell: n which {{node_version}} || true
>       register: required_node_version_installed
> ```

(So, if `node` isn't installed then it will complain, print 'error' to the process stdout (so far all silently to the human running ansible), and show as an 'ok' Ansible task (indeed, it successfully checked the node version and made a note of the result). 

If `node` is installed then it will print something like `v1.2.3` and we save that output.)

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/849)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[deploy/dependencies.yml, line 67 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/849#-MKIoLjQ-BXGSNVuIyMK:-MKMQ8lyClnV0J-Xslc6:bkyqngf) ([raw file](https://github.com/sillsdev/web-xforge/blob/f0528f88fe1cf3dd6251bfbad6ae8413f0dfb0a4/deploy/dependencies.yml#L67)):*

> which one of the installed versions is being used.

Ah ha. I was relying on querying n itself to tell me things like that. But after you mentioned this, I found that I can easily see what node version is being used by just running `node --version`, since `n NEW_VERSION` changes the /usr/local/bin/node file. 

I was confused about n only distinguishing between installed and not-installed, rather than accounting much for 'selected'. 

> Given that we primarily use Ansible to document the setup with a secondary benefit that it also gives us a from-scratch automation, and that we don't support it changing from one setup to another

I would say that our Ansible serves a significant role of configuring dev and test machines so they work. After we change our dotnet version, or node version, or add mp3 encoding support, etc, we can just instruct people to run the ansible playbook and it configures their dev or test machine. In my view, we have run the playbook to _change_ (or fix) a machine significantly more often than to initially set up a machine. 

But, I hear what you're saying, and it wasn't broken before. That said, your comment quickly landed me on a solution that I have updated the PR with. If that's good, we can go with that. But we don't need to send this part thru if you prefer not. 


</blockquote></details>

( Oh, I meant to say: I am aware that the Ansible isn't being used to do configuration on Windows dev machines, but I think if we modified it to work there, it could be helpful. Tho I'm not hearing SF developers on Windows complaining that they want automated machine configuration, so maybe it's not significant. (Or maybe they don't know how great it is :-) )

irahopkinson (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/849)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[deploy/dependencies.yml, line 67 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/849#-MKIoLjQ-BXGSNVuIyMK:-MKN_M0gCWHP__SuEr__:b-li8i1c) ([raw file](https://github.com/sillsdev/web-xforge/blob/f0528f88fe1cf3dd6251bfbad6ae8413f0dfb0a4/deploy/dependencies.yml#L67)):*

( Oh, I meant to say: I am aware that the Ansible isn't being used to do configuration on Windows dev machines, but I think if we modified it to work there, it could be helpful. Tho I'm not hearing SF developers on Windows complaining that they want automated machine configuration, so maybe it's not significant. (Or maybe they don't know how great it is :-) )
</blockquote></details>

Nicely done.

I recognise that Ansible does automate changing from one setup to another quite often but we don't want to spend lots of developer effort making it capable of that. If there was a significant change to another setup and a developer wanted to spend time on getting Ansible to support it I would tell them to destroy their vagrant and standup a new one instead. If we were in dev ops then this would be a different answer because there we would be using Ansible entirely for how it is intended. Just clarifying, but I think you understand that.

PR #850, opened 2020-10-23
vagrant: Add sfdev using Ubuntu 20.04

marksvc:
The current development and testing vagrant machine uses Ubuntu 16.04. This PR introduces a development & testing vagrant machine using Ubuntu 20.04. It also breaks away from the prior base box creation process, and uses cleaner naming.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,35 +110,37 @@ We use [Prettier](https://prettier.io/) with a pre-commit hook.
 
 We use [Angular Flex-Layout](https://github.com/angular/flex-layout) with [Angular MDC](https://trimox.github.io/angular-mdc-web) including the [Material Design Icons](https://google.github.io/material-design-icons/).
 
-### Recommended Development Environment
+### Development Environment
 
-Our recommended development environment for web development is Ubuntu 16.04.
-
-- [Vagrant GUI Setup](#vagrant-gui-setup). A Vagrant box with xForge already installed is downloaded and set up on your machine. This is the easiest and cleanest to setup.
-- [Local Linux Development Setup](#local-linux-development-setup). Everything is installed directly on your machine, which needs to be running Ubuntu 16.04. This is the fastest method because development is not done in a virtual machine.
+- [Vagrant GUI Setup](#vagrant-gui-setup). A Vagrant box with xForge already installed is downloaded and set up on your
+  machine. This is the easiest and cleanest to setup.
+- [Local Linux Development Setup](#local-linux-development-setup). Everything is installed directly on your machine,
+  which needs to be running Ubuntu 16.04. This is the fastest method because development is not done in a virtual machine.
 - [Manual Setup](#manual-setup) This setup is specifically written for **Windows** but the steps could be used for any OS.
 
-#### Vagrant GUI Setup
+#### Vagrant Development Machine
 
-Install [VirtualBox](https://www.virtualbox.org/), [Vagrant](https://www.vagrantup.com/), and [git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git). To do this in Linux, run
+Install [VirtualBox](https://www.virtualbox.org/), [Vagrant](https://www.vagrantup.com/), and
+[git](https://git-scm.com/book/en/v2/Getting-Started-Installing-Git). To do this in Linux, run
 
     sudo apt install vagrant virtualbox virtualbox-guest-additions-iso git
 
-Setup git. At least name and email is needed in `.gitconfig`. You can do this from a terminal by running
+Setup git. At least name and email is needed in `.gitconfig`. You can do this from a terminal in your host machine by running
 
     git config --global user.name "My Name"
     git config --global user.email "me@example.com"
 
 Hardware-assisted virtualization (VT-x or AMD-V) needs to be enabled in your BIOS.
 
-Create a directory to manage the development machine, such as `xforge`. Checkout the xforge git repository to access (and later receive updates to) the vagrant development machine configuration file:
+Clone the xforge git repository to access (and later receive updates to) the vagrant development machine configuration file:
 
     git clone https://github.com/sillsdev/web-xforge
-    cd web-xforge/deploy/vagrant_xenial_gui
+    cd web-xforge/deploy/sfdev
 
-Run `vagrant up`. This will download, initialize, and run the development machine. The machine is about 5GB, so expect the download to take a while.
+Run `vagrant up`. This will download, initialize, and run the development machine. The machine is about 5GB, so expect
+the download to take a while.
 
-In the guest development machine, do the following additional steps. If a dialog about grub appears during upgrade, use TAB, ENTER, and SPACE to specify to use `sda` or to not worry about it, depending on which dialog you get.
+In the guest development machine, do the following additional steps.
 
 ```shell
 sudo apt update
--- a/deploy/sfdev/Vagrantfile
+++ b/deploy/sfdev/Vagrantfile
@@ -1,0 +1,74 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+Vagrant.configure("2") do |config|
+  config.vm.box = "sfdev"
+  config.vm.box_url = "http://linux.lsdev.sil.org/vagrant/sfdev/sfdev.json"
+  config.vm.hostname = "sfdev"
+
+  config.vm.provider "virtualbox" do |vb|
+    vb.gui = true
+    vb.cpus = 4
+    # At least 3000 MB of RAM is needed to run tests. Using an initial value here of 6000 MB in case the machine is
+    # opened on a computer with 8 GB RAM. Most of the time, 14000 MB RAM is enough for development, using Code,
+    # Chromium, running tests, etc. at the same time. Set the RAM value in a SFDEV_RAM environment variable by running
+    # a command such as
+    #   tee -a ~/.pam_environment <<< 'SFDEV_RAM="14000"'
+    # and then logging back into your computer. Or temporarily set the value here.
+    vb.memory = ENV['SFDEV_RAM'] || "6000"
+    # Enable 3D acceleration and enough video RAM for larger displays.
+    vb.customize ["modifyvm", :id, "--accelerate3d", "on"]
+    vb.customize ["modifyvm", :id, "--vram", "128"]
+  end
+
+  config.vm.provision "shell", privileged: false, inline: <<~SHELL
+    set -eux -o pipefail
+
+    function tryharderto() { i=0; until "$@"; do ((++i > 3)) && false; echo >&2 Retrying $i; sleep 2m; done }
+
+    # Canary
+    tee ~/Desktop/warning-not-provisioned.txt <<END
+    The presence of this file means that the vagrant virtual machine did not
+    finish provisioning correctly.
+    The best way to get to a working machine state is to backup any data that
+    you want to keep from this virtual machine, and then destroy and re-create
+    this vagrant virtual machine by running these commands from your host:
+        vagrant destroy
+        vagrant up
+    END
+
+    GIT_USER="#{`git config --get user.name`.strip}"
+    GIT_EMAIL="#{`git config --get user.email`.strip}"
+    [[ -z $GIT_USER ]] || git config --global user.name "${GIT_USER}"
+    [[ -z $GIT_EMAIL ]] || git config --global user.email "${GIT_EMAIL}"
+
+    cd ~/src/web-xforge
+    # Clear away some modified files. But not the whole tree in case a
+    # user runs this against their data.
+    git checkout -- src/SIL.XForge.Scripture/ClientApp/package{,-lock}.json || true
+    git pull --ff-only --recurse-submodules
+
+    cd ~/src/web-xforge/deploy
+    tryharderto ansible-playbook playbook_bionic.yml --limit localhost
+
+    sudo n lts
+
+    cd ~/src/web-xforge/
+    dotnet clean
+    # Remove orphaned? dotnet process.
+    sleep 10s
+    pkill dotnet
+    rm -rf /tmp/NuGetScratch/lock
+
+    cd ~/src/web-xforge/src/SIL.XForge.Scripture/ClientApp
+    npm i
+    cd ~/src/web-xforge/src/RealtimeServer
+    npm i
+    cd ~/src/web-xforge/src/SIL.XForge.Scripture
+    dotnet build
+
+    # Remove canary
+    rm ~/Desktop/warning-not-provisioned.txt
+    echo Provisioning finished successfully.
+  SHELL
+end
--- a/deploy/sfdev/base/instructions.md
+++ b/deploy/sfdev/base/instructions.md
@@ -77,28 +77,38 @@
 
 ## Generate and publish product
 
-* Export VM .box file. This may take 20+ minutes. The `--base` argument is the name of the
-  base machine in virtualbox manager. Replace the VERSION string in the below before
-  running.
+* Export VM .box file and prepare to test the result. This may take 30 minutes.
+  The `--base` argument is the name of the base machine in virtualbox manager. Replace
+  the VERSION string in the below before running.
 
-  export BOX="sfdev"
-  export VERSION="1.0.0"
-  date && time vagrant package --base ${BOX}-base --output ${BOX}-${VERSION}.box \
-    && ls -lh ${BOX}-${VERSION}.box && sha256sum ${BOX}-${VERSION}.box
+export BOX="sfdev"
+export VERSION="1.0.0"
+date && vagrant package --base ${BOX}-base --output ${BOX}-${VERSION}.box &&
+  date && ls -lh ${BOX}-${VERSION}.box &&
+  sha256sum ${BOX}-${VERSION}.box | tee -a sfdev.json &&
+  date && mkdir test && tee test/Vagrantfile << END &&
+Vagrant.configure("2") do |config|
+  config.vm.box = "../${BOX}-${VERSION}.box"
+  config.vm.provider "virtualbox" do |vb|
+    vb.gui = true
+    vb.cpus = 2
+    vb.memory = "6000"
+  end
+end
+END
+cd test && vagrant up; date
 
-* Update the box .json file, including the sha.
+* Test the result. Such as by adding secrets and running the "Full App (SF)" debug target
+  in Code.
 
-* Test base box. Enable vb.gui in Vagrantfile. Give it more RAM.
+* Clean up box test machine. In the `test` directory, run the following.
+  The `vagrant box remove` command removes the internally stored copy of the base box
+  (to free disk space); it's not removing the '../foo' _file_, but internally stored data
+  with the designation of '../foo'.
 
-  mkdir test && cd test && vagrant init ../${BOX}-${VERSION}.box && vim Vagrantfile
-  vagrant up
+  vagrant destroy && vagrant box remove "../${BOX}-${VERSION}.box" && cd .. && rm test -rf
 
-* Clean up box test machine. In the `test` directory, run the following. Then delete the
-  `test` directory. The `vagrant box remove` command removes the internally stored copy of
-  the base box (to free disk space); it's not removing the '../foo' _file_, but internally
-  stored data with the designation of '../foo'.
-
-  vagrant destroy && vagrant box remove "../${BOX}-${VERSION}.box"
+* Update the box .json file, and move the sha to the right place.
 
 ## Notes
 
--- a/deploy/sfdev/base/instructions.md
+++ b/deploy/sfdev/base/instructions.md
@@ -1,0 +1,109 @@
+# Scripture Forge development base box creation instructions
+
+## Install
+
+* Use virtualbox manager to create a new VM with a temporary name (like `sfdev-base`).
+* Give 6 GB RAM and 200 GB dynamic storage. Actual RAM available will depend on a
+  Vagrantfile setting later on. Give 2 processors.
+* Boot Ubuntu .iso. Install Ubuntu. Don't install unneeded "third-party software".
+* Use computer name `sfdev`, username `vagrant` and password `vagrant`. Choose Log in
+  automatically.
+
+## Setup machine
+
+* Launch Software Updater and apply updates. If that gives any trouble, can fall back to:
+  boot to recovery mode, enable networking, and run `apt update && apt upgrade -y`.
+
+* Install guest additions.
+
+  * On your host machine, run
+
+    sudo apt install virtualbox-guest-additions-iso
+
+  * In the guest machine, run
+
+    sudo apt update && sudo apt install -y linux-headers-$(uname -r) build-essential \
+      dkms virtualbox-guest-utils virtualbox-guest-dkms
+
+    In the guest machine window, choose Devices > Insert Guest Additions CD image. In the
+    guest, a VBOXADDITIONS window appears; click Run. A terminal window will appear,
+    asking if you want to replace versions of additions. Enter yes. It will say something
+    went wrong. Right-click the CD icon on the panel or desktop and Eject the VirtualBox
+    Additions disc.
+
+  * In the guest machine window, choose Devices > Shared clipboard > Bidirectional.
+
+  * Reboot.
+
+* Set resolution to a big enough but conservative initial value of 1600x900, that will fit
+  inside someone's 1080p desktop.
+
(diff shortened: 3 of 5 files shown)


irahopkinson (changes requested):
Looks good although I'm not the right person to review this. Do you have anyone else who can review your Linux scripts? Failing that, for this, I'm happy enough with "I tested it and it works".

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/850#-:-MKNqDbY24OPg1pP2KbU:b-vrb914)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[README.md, line 138 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKNqbVm5z4w7ibOF3sS:-MKNqbVm5z4w7ibOF3sT:b-5oql8w) ([raw file](https://github.com/sillsdev/web-xforge/blob/2735d804d8e5578e66af13504e69fe8415539004/README.md#L138)):*
> ```
> sfdev
> ```
Since this is the new folder you should delete the old folder `vagrant_xenial_gui` and the corresponding `virtualbox_xenial_gui_creation`.

I would also prefer the new folder retain `vagrant` in its name to distinguish it since its in an Ansible deploy folder. Additionally there is no easy way to know what version of Ubuntu it is so I would also prefer to have `focal` in the folder name. (Or some other way to make both of these obvious.)
___
*[deploy/sfdev/base/sfdev.json, line 7 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKNnNBU6AVbHSAt_iNQ:-MKNnNBU6AVbHSAt_iNR:b-a0nmae) ([raw file](https://github.com/sillsdev/web-xforge/blob/2735d804d8e5578e66af13504e69fe8415539004/deploy/sfdev/base/sfdev.json#L7)):*
> ```
> "version": "0.0.1"
> ```
What is the difference between these two versions? Can we add a description?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/850)* status: all files reviewed, 5 unresolved discussions (waiting on @marksvc)

___
*[deploy/sfdev/Vagrantfile, line 24 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKZmGFY37nYaeo3TXm_:-MKZmGFY37nYaeo3TXma:b-fbkrp) ([raw file](https://github.com/sillsdev/web-xforge/blob/2735d804d8e5578e66af13504e69fe8415539004/deploy/sfdev/Vagrantfile#L24)):*
> ```Ruby
>   end
> 
>   config.vm.provision "shell", privileged: false, inline: <<~SHELL
> ```

Do we have any guarantee that this will be run with `/bin/bash` instead of `/bin/sh`? I don't expect problems, but nothing gives problems with shell scripts like running it with a different shell than intended.
___
*[deploy/sfdev/Vagrantfile, line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKZnM40BksVpwk7V_Iv:-MKZnM40BksVpwk7V_Iw:b-g973yh) ([raw file](https://github.com/sillsdev/web-xforge/blob/2735d804d8e5578e66af13504e69fe8415539004/deploy/sfdev/Vagrantfile#L27)):*
> ```Ruby
>     set -eux -o pipefail
> 
>     function tryharderto() { i=0; until "$@"; do ((++i > 3)) && false; echo >&2 Retrying $i; sleep 2m; done }
> ```

How does `((++i > 3)) && false` work?
___
*[deploy/sfdev/Vagrantfile, line 66 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKZkmjc5ehoHo4U9w_c:-MKZkmjc5ehoHo4U9w_d:b-u7kq6e) ([raw file](https://github.com/sillsdev/web-xforge/blob/2735d804d8e5578e66af13504e69fe8415539004/deploy/sfdev/Vagrantfile#L66)):*
> ```Ruby
>     npm i
>     cd ~/src/web-xforge/src/RealtimeServer
>     npm i
> ```

@irahopkinson Doesn't `npm i` in `RealtimeServer` potentially affect the output of `npm i` in `ClientApp`? If so, might it make sense to reverse the order here?

marksvc (commented):
Thanks. Nathaniel is a good choice for this sort of thing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/850#-:-MK_tFHS1nsnxwcPubcp:bz2lfnp)* status: all files reviewed, 5 unresolved discussions (waiting on @irahopkinson and @Nateowami)

___
*[README.md, line 138 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKNqbVm5z4w7ibOF3sS:-MK_mXhNDXijU8UL2-Vw:bo9mxo1) ([raw file](https://github.com/sillsdev/web-xforge/blob/2735d804d8e5578e66af13504e69fe8415539004/README.md#L138)):*
One of the things I wanted to get away from with the prior dev machine was the naming :-). Different things appeal to different people: I find `sfdev` to be much nicer to look at than `vagrant_xenial_gui` which I find ugly :). 

> I would also prefer the new folder retain vagrant in its name to distinguish it since its in an Ansible deploy folder. 

There is a good consideration there. I did some googling but couldn't see that "deploy" is a name for directories of Ansible files. Certainly "deploy" makes sense as a location where we store scripts and other files related to how we send out or construct our software for actual usage on machines. In that sense, having one or more Vagrant machines in deploy makes sense, without needing to assume that the directory is for Ansible. That said, it _is_ used for Ansible. And I don't want to make another top-level folder in the repository root. ~I'll put 'vagrant' in the name.~ I'm putting `vagrant` in the path, as explained later.

> Additionally there is no easy way to know what version of Ubuntu it is so I would also prefer to have focal in the folder name. 

Right. I see that the Vagrantfile doesn't say either. We could write a comment in the Vagrantfile that the sfdev machine is Ubuntu 20.04. But that said, when I made the sfdev, part of my thinking was that sfdev could be upgraded to future versions of Ubuntu. So for example, a developer or tester could `vagrant up` from `sfdev/Vagrantfile` to get a current SF development/testing machine. Eventually they may make a mess of their machine, or just want something new, and so they run `vagrant box update ; vagrant destroy ; vagrant up`. Right now sfdev uses Ubuntu 20.04, but in a couple years, we could publish an update that uses Ubuntu 22.04. And so updates to the base box could be delivering new OS versions to users. 

So from this perspective, it's _the_ dev+test vagrant, not the Ubuntu 20.04 dev+test vagrant. 
And we wouldn't have to change the folder structure (or instructions) in the event that we release a new basebox that uses Ubuntu 22.04. It would still just be the dev+test vagrant in `sfdev/Vagrantfile`, at whatever OS version was last published to the basebox repo.

With FW, it was very significant what version of Ubuntu a dev or test machine was using. And in fact, it was important to me that we still be able to use an Ubuntu 16.04 dev machine even when we can start using an Ubuntu 18.04 dev machine. But SF development and testing is a lot less related to the OS or OS version.

I wrote in the Vagrantfile that sfdev uses Ubuntu 20.04. Then interested parties can figure out what version of Ubuntu is being used without having to run `vagrant up`, but we can still have a prettier directory name that we can use across Ubuntu versions. 
Is that adequate?

> (Or some other way to make both of these obvious.)

Actually, how about this. Let's make a `vagrant` sub directory of `deploy`. Then I'll put a `sfdev` directory in `vagrant`. That will make it clear that it's vagrant related. And it will have the short and nice directory name. (So a path like `web-xforge/deploy/vagrant/sfdev/Vagrantfile`)
I'll send in a change that does that, but let me know if that satisfies your concerns.

> Since this is the new folder you should delete the old folder vagrant_xenial_gui and the corresponding virtualbox_xenial_gui_creation.

Yes. We can do that. I am concerned that if we do that now, it may trip clients who are using the vagrant_xenial_gui machine. It won't delete their virtual machine, but if it takes the Vagrantfile away, then won't be able to `vagrant up` any more. So yes, let's remove the Ubuntu 16.04 vagrant machine, but let's do it after a period of time so that we don't think we are going to unnecessarily cause headache for anyone. (Which is only a handful of people, but may as well not cause trouble if we can just delay the removal.)
___
*[deploy/sfdev/Vagrantfile, line 24 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKZmGFY37nYaeo3TXm_:-MK_c9q4-6YzX_-12zkh:bcpl0m2) ([raw file](https://github.com/sillsdev/web-xforge/blob/2735d804d8e5578e66af13504e69fe8415539004/deploy/sfdev/Vagrantfile#L24)):*

Do we have any guarantee that this will be run with `/bin/bash` instead of `/bin/sh`? I don't expect problems, but nothing gives problems with shell scripts like running it with a different shell than intended.
</blockquote></details>

Right. Good question. So far it hasn't been a problem. And this part is identical to our prior Vagrantfile. Thru the various Vagrantfiles I've worked on, I don't recall having a problem with sh vs bash for these kinds of stanzas. So I don't think we need to worry about it. 
If it did go awry, it would be going awry somehow in the guest machine, not the host. So it shouldn't matter if the host somehow doesn't have bash, because this runs in the guest.
___
*[deploy/sfdev/Vagrantfile, line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKZnM40BksVpwk7V_Iv:-MK_cjAh5rprz_qIqM7o:b-y889yz) ([raw file](https://github.com/sillsdev/web-xforge/blob/2735d804d8e5578e66af13504e69fe8415539004/deploy/sfdev/Vagrantfile#L27)):*

How does `((++i > 3)) && false` work?
</blockquote></details>

I carefully got this put together a while ago, to give a short loop to re-try a failing commands up to a max number of tries. It has something to do with the `set -e`. I just tried playing with it for a moment now, but I wouldn't be able to say why it works without spending some more time fiddling with it. The `((++i > 3)) && false` looks really weird, because it's basically starting off as `false && false`. Eventually it changes to `true && false`. But you would think that `false && false` and `true && false` would have the same meaning. 

If you put 
```
i=0; until "false" ; do ((++i > 3)) ; echo hey $i;sleep 1s;done
``` 
into a script and run it: `bash foo`, you can see that it continues forever. But if you prepend 
```
set -e
```
then it only prints 3 times. 

So, I can't say how it works off the top of my head :). And I may not have really understood it when I finally landed upon the solution in the first place. But it I can at least say it triangulates a solution that works that satisfies a few factors of:
- short
- works with set -e
- stops if the command succeeds
- stops if the retry count is exceeded (for a total of I think 1+3 tries)
___
*[deploy/sfdev/Vagrantfile, line 66 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKZkmjc5ehoHo4U9w_c:-MK_fSbVDvzx1uJRABCP:bc9tszs) ([raw file](https://github.com/sillsdev/web-xforge/blob/2735d804d8e5578e66af13504e69fe8415539004/deploy/sfdev/Vagrantfile#L66)):*

@irahopkinson Doesn't `npm i` in `RealtimeServer` potentially affect the output of `npm i` in `ClientApp`? If so, might it make sense to reverse the order here?
</blockquote></details>

Yes. Good point. I'll change that. 
Now, this unfortunate series of commands (starting at `git checkout` above) has been vetted quite a bit. But I don't anticipate this messing it up.
___
*[deploy/sfdev/base/sfdev.json, line 7 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKNnNBU6AVbHSAt_iNQ:-MK_gkQX3ixYwDDlf-F3:b-xgqci2) ([raw file](https://github.com/sillsdev/web-xforge/blob/2735d804d8e5578e66af13504e69fe8415539004/deploy/sfdev/base/sfdev.json#L7)):*

> ```
> "version": "0.0.1"
> ```
What is the difference between these two versions? Can we add a description?
</blockquote></details>

There isn't much documentation on the file format. I just checked the site and don't see a per-version description field. That said, it's not really significant to provide more information here. This is consumed by vagrant without the user seeing it, and facilitates the `vagrant box update` command working to grab the latest version that the command finds in this file. Users wouldn't come look at this and think about whether they wanted version A or B. 

The difference is that version 0.0.1 was the first version I published when I started testing the machine. After a while I applied some fixes, deemed the box ready for others to use, and so published version 1.0.0. 

Keeping old versions in the file is probably only significant for helping client computers' `vagrant box update` commands understanding where they stand and not get confused if they don't spot their current version in the file when looking for higher versions (conceivably). 

Oh, I probably never explained this, but the sfdev.json file, and the actual basebox .box files, are uploaded to a fileserver, and the .box files and the accompanying .json file are what constitute a versioned basebox repository. So this file isn't consumed locally by anything in our repository or on our systems. But rather this file is what gets uploaded to a file server that provides base box files for download. Versioning it here is a tidy place to keep track of it. 

I guess both 0.0.1 and 1.0.0 appeared at the same time from the perspective of web-xforge.git. Tho one computer (mine) did start fetching version 0.0.1 previously. (And I still have two 0.0.1 machines that I use.)

marksvc (commented):
I am okay with this series of commits being squashed to be put onto master. But I will be glad for the series of commit messages to be brought along with it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/850#-:-MKa0PQF0pwPNZ7O3THo:bvgqmbf)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @irahopkinson and @Nateowami)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/850)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @irahopkinson and @marksvc)

___
*[deploy/sfdev/Vagrantfile, line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKZnM40BksVpwk7V_Iv:-MKa5dOiERm66KytvkX9:bhhwwvf) ([raw file](https://github.com/sillsdev/web-xforge/blob/2735d804d8e5578e66af13504e69fe8415539004/deploy/sfdev/Vagrantfile#L27)):*

I carefully got this put together a while ago, to give a short loop to re-try a failing commands up to a max number of tries. It has something to do with the `set -e`. I just tried playing with it for a moment now, but I wouldn't be able to say why it works without spending some more time fiddling with it. The `((++i > 3)) && false` looks really weird, because it's basically starting off as `false && false`. Eventually it changes to `true && false`. But you would think that `false && false` and `true && false` would have the same meaning. 

If you put 
```
i=0; until "false" ; do ((++i > 3)) ; echo hey $i;sleep 1s;done
``` 
into a script and run it: `bash foo`, you can see that it continues forever. But if you prepend 
```
set -e
```
then it only prints 3 times. 

So, I can't say how it works off the top of my head :). And I may not have really understood it when I finally landed upon the solution in the first place. But it I can at least say it triangulates a solution that works that satisfies a few factors of:
- short
- works with set -e
- stops if the command succeeds
- stops if the retry count is exceeded (for a total of I think 1+3 tries)



</blockquote></details>

Huh. Weird. I would have thought `((++i < 3))` would have been sufficient.

irahopkinson (commented):
> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/850)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[README.md, line 138 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKNqbVm5z4w7ibOF3sS:-MKdcxCbAwQw2Dc1AmbV:ba8l9kh) ([raw file](https://github.com/sillsdev/web-xforge/blob/2735d804d8e5578e66af13504e69fe8415539004/README.md#L138)):*

One of the things I wanted to get away from with the prior dev machine was the naming :-). Different things appeal to different people: I find `sfdev` to be much nicer to look at than `vagrant_xenial_gui` which I find ugly :). 

> I would also prefer the new folder retain vagrant in its name to distinguish it since its in an Ansible deploy folder. 

There is a good consideration there. I did some googling but couldn't see that "deploy" is a name for directories of Ansible files. Certainly "deploy" makes sense as a location where we store scripts and other files related to how we send out or construct our software for actual usage on machines. In that sense, having one or more Vagrant machines in deploy makes sense, without needing to assume that the directory is for Ansible. That said, it _is_ used for Ansible. And I don't want to make another top-level folder in the repository root. ~I'll put 'vagrant' in the name.~ I'm putting `vagrant` in the path, as explained later.

> Additionally there is no easy way to know what version of Ubuntu it is so I would also prefer to have focal in the folder name. 

Right. I see that the Vagrantfile doesn't say either. We could write a comment in the Vagrantfile that the sfdev machine is Ubuntu 20.04. But that said, when I made the sfdev, part of my thinking was that sfdev could be upgraded to future versions of Ubuntu. So for example, a developer or tester could `vagrant up` from `sfdev/Vagrantfile` to get a current SF development/testing machine. Eventually they may make a mess of their machine, or just want something new, and so they run `vagrant box update ; vagrant destroy ; vagrant up`. Right now sfdev uses Ubuntu 20.04, but in a couple years, we could publish an update that uses Ubuntu 22.04. And so updates to the base box could be delivering new OS versions to users. 

So from this perspective, it's _the_ dev+test vagrant, not the Ubuntu 20.04 dev+test vagrant. 
And we wouldn't have to change the folder structure (or instructions) in the event that we release a new basebox that uses Ubuntu 22.04. It would still just be the dev+test vagrant in `sfdev/Vagrantfile`, at whatever OS version was last published to the basebox repo.

With FW, it was very significant what version of Ubuntu a dev or test machine was using. And in fact, it was important to me that we still be able to use an Ubuntu 16.04 dev machine even when we can start using an Ubuntu 18.04 dev machine. But SF development and testing is a lot less related to the OS or OS version.

I wrote in the Vagrantfile that sfdev uses Ubuntu 20.04. Then interested parties can figure out what version of Ubuntu is being used without having to run `vagrant up`, but we can still have a prettier directory name that we can use across Ubuntu versions. 
Is that adequate?

> (Or some other way to make both of these obvious.)

Actually, how about this. Let's make a `vagrant` sub directory of `deploy`. Then I'll put a `sfdev` directory in `vagrant`. That will make it clear that it's vagrant related. And it will have the short and nice directory name. (So a path like `web-xforge/deploy/vagrant/sfdev/Vagrantfile`)
I'll send in a change that does that, but let me know if that satisfies your concerns.

> Since this is the new folder you should delete the old folder vagrant_xenial_gui and the corresponding virtualbox_xenial_gui_creation.

Yes. We can do that. I am concerned that if we do that now, it may trip clients who are using the vagrant_xenial_gui machine. It won't delete their virtual machine, but if it takes the Vagrantfile away, then won't be able to `vagrant up` any more. So yes, let's remove the Ubuntu 16.04 vagrant machine, but let's do it after a period of time so that we don't think we are going to unnecessarily cause headache for anyone. (Which is only a handful of people, but may as well not cause trouble if we can just delay the removal.)




</blockquote></details>

OK, thanks for the expalination, that is better, and having the version comment in the `Vagrantfile` sorts the other issue.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/850)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[deploy/sfdev/Vagrantfile, line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKZnM40BksVpwk7V_Iv:-MKje-ZF-syy0yfvK-3a:b-l0tlne) ([raw file](https://github.com/sillsdev/web-xforge/blob/2735d804d8e5578e66af13504e69fe8415539004/deploy/sfdev/Vagrantfile#L27)):*

Huh. Weird. I would have thought `((++i < 3))` would have been sufficient.
</blockquote></details>

And more experimenting may show a much simpler or elegant way.

irahopkinson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/850)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[deploy/sfdev/Vagrantfile, line 66 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKZkmjc5ehoHo4U9w_c:-MKmKWv92Gtmu4Feakct:b-jqz93x) ([raw file](https://github.com/sillsdev/web-xforge/blob/2735d804d8e5578e66af13504e69fe8415539004/deploy/sfdev/Vagrantfile#L66)):*

Yes. Good point. I'll change that. 
Now, this unfortunate series of commands (starting at `git checkout` above) has been vetted quite a bit. But I don't anticipate this messing it up.
</blockquote></details>

Actually it shouldn't any more but changing it puts it in the right order anyway. That is safer in case the problem pops up again.

Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/850)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[deploy/vagrant/sfdev/base/provision, line 87 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKtdxZf7RRCAjqQq_pv:-MKtdxZf7RRCAjqQq_pw:bf00bth) ([raw file](https://github.com/sillsdev/web-xforge/blob/fa5eb9ffbe134e6a88065a61c6c4e7fdc6dd8d73/deploy/vagrant/sfdev/base/provision#L87)):*
> ```
> Vagrant base box information
> Name: sfdev
> Version:
> ```

Are some of these fields (like version) actually supposed to be blank? If so, why?
___
*[deploy/vagrant/sfdev/base/provision, line 355 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKtjGCS9Pi1Y6hCcrko:-MKtjGCS9Pi1Y6hCcrkp:bdoh89f) ([raw file](https://github.com/sillsdev/web-xforge/blob/fa5eb9ffbe134e6a88065a61c6c4e7fdc6dd8d73/deploy/vagrant/sfdev/base/provision#L355)):*
> ```
> 
> # mongodb-org isn't released for focal yet. Just install the bionic mongodb for now.
> perl -pi -e '/base_distribution_release/ and s/bionic/focal/' ~/src/web-xforge/deploy/dependencies.yml
> ```

Is this rewriting an Ansible file, running it, and then checking it out again? Is it not possible to create another config file for Ansible and specify that it should be used?

irahopkinson (approved):

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/850)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[deploy/vagrant/sfdev/base/provision, line 87 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKtdxZf7RRCAjqQq_pv:-MKu47qtBkdYdngj9Q7T:bt577jh) ([raw file](https://github.com/sillsdev/web-xforge/blob/fa5eb9ffbe134e6a88065a61c6c4e7fdc6dd8d73/deploy/vagrant/sfdev/base/provision#L87)):*

Are some of these fields (like version) actually supposed to be blank? If so, why?
</blockquote></details>

Yes. The comment above should probably say it's a template. This puts down a partially-filled in file to describe the machine a bit. So if someone has a problem on their machine, we can ask what their ~/machine-info.txt file says, and can learn if it's an old version of the vagrant box, or one that matches someone else's machine, etc. 
The Creation date is easy to fill in automatically, and is probably all that's needed. But the other fields can be optionally filled in by the person making the base box to specify any additional useful information. 
For example, I probably filled in 1.0.0 for Version when making the recent base box. (But this script doesn't know to fill that in; it just makes the initial file that can be written into further.)
___
*[deploy/vagrant/sfdev/base/provision, line 355 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/850#-MKtjGCS9Pi1Y6hCcrko:-MKu50848qdvpbUClwMj:bpp23j) ([raw file](https://github.com/sillsdev/web-xforge/blob/fa5eb9ffbe134e6a88065a61c6c4e7fdc6dd8d73/deploy/vagrant/sfdev/base/provision#L355)):*

Is this rewriting an Ansible file, running it, and then checking it out again? Is it not possible to create another config file for Ansible and specify that it should be used?
</blockquote></details>

It is, yes. It was a temporary workaround that became part of the provision script I used. 
Yes, making another file would have been possible. And in fact, there was a permanent fix made recently, so this temporary workaround shouldn't be needed any more. It will be simpler to undo the temporary fix here after merging this change, and so I am planning to remove this in a followup commit. I probably should have posted a comment here ahead of time then.

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/850)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #858, opened 2020-11-02
SF-1121 Error when back translation not in registry

marksvc:
- Give a useful message if the user tries to connect to a project that
  isn't actually listed in the registry, like back translation
  projects can be, as we don't support that right now.
- Add to already-connected error test to distinguish between kinds of
  errors.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/858)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.spec.ts
@@ -10,7 +10,7 @@ import { CheckingShareLevel } from 'realtime-server/lib/scriptureforge/models/ch
 import { SFProject } from 'realtime-server/lib/scriptureforge/models/sf-project';
 import { SFProjectRole } from 'realtime-server/lib/scriptureforge/models/sf-project-role';
 import { BehaviorSubject, defer, of } from 'rxjs';
-import { anything, deepEqual, mock, resetCalls, verify, when } from 'ts-mockito';
+import { anything, capture, deepEqual, mock, resetCalls, verify, when } from 'ts-mockito';
 import { CommandError, CommandErrorCode } from 'xforge-common/command.service';
 import { I18nService } from 'xforge-common/i18n.service';
 import { NoticeService } from 'xforge-common/notice.service';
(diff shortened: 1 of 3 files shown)


marksvc:
Selection_098

Nateowami (changes requested):
It seems like if a project cannot be connected, then it should not be in the list at all (or it should be disabled, with a note such as `(this back translation cannot be connected)`). I assume the API just doesn't make that easy to implement, so failing after the user tries to connect it was selected as the next best option?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/858#-:-MLAlU4-2KyHbNEI2CtB:b-738fra)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.spec.ts, line 346 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/858#-MLAidKi5TllPRtf2m9s:-MLAidKi5TllPRtf2m9t:byfknqt) ([raw file](https://github.com/sillsdev/web-xforge/blob/bb0852f72a99574eb119a0fbd0a2ead885c890a6/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.spec.ts#L346)):*
> ```XML
>     verify(mockedSFProjectService.onlineCreate(anything())).once();
>     verify(mockedSFProjectService.onlineAddCurrentUser(anything())).never();
>     verify(mockedRouter.navigate(deepEqual(['/projects', 'project01']))).never();
> ```

If you use `anything()` in place of `deepEqual(['/projects', 'project01'])`, would that be too generic and cause the test to fail? If not, I think it should be written more generally since it's expecting that this _didn't_ happen. Generally speaking, we want our tests to be as specific as possible, but for negative assertions, being more general actually is a stronger assertion.

Recently I had trouble with a `verify(mockedRouter(something_i_cannot_remember)).once()`. It was failing because I didn't write it correctly (maybe I needed to use `deepEqual`). I knew the test had a problem because it was failing. But in this case, if you wrote it even slightly wrong, it would silently pass.

This is similar to when we were doing `expect(someDebugNode).not.toBeUndefined()`, which was passing, because the negative assertion wasn't general enough (we should have been including `null`). Something like `.not.toBeFalsy()` would have included all values we cared about (`null` and `undefined`) and some that we don't care about but wouldn't ever be likely to occur, and if they did, would just result in a test failure (`false`, `''`, `0`, `NaN`, `-0`, `0n`).

(Same for the test below).

marksvc:
(So, going to change the solution so that projects that are not in the registry just don't show up in the list.)

Nateowami:
By the way, @marksvc, do you have pre-commit hooks working locally? I thought they prevented us from committing code that wouldn't pass ts-lint. I [force-pushed](https://github.com/sillsdev/web-xforge/compare/bc99d15e4c033892a74db5a0062941db56cf816d..bb0852f72a99574eb119a0fbd0a2ead885c890a6) this branch yesterday to make a very trivial change to make the tests pass on TeamCity.

marksvc (commented):
Thank you for asking; I brought this to chat where we can more easily dialog about it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/858#-:-MLJmeHwA45vAfqquN78:b-sosto3)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.spec.ts, line 346 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/858#-MLAidKi5TllPRtf2m9s:-MLJmQ8y6967CXf2DzwA:blvixw9) ([raw file](https://github.com/sillsdev/web-xforge/blob/bb0852f72a99574eb119a0fbd0a2ead885c890a6/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.spec.ts#L346)):*

If you use `anything()` in place of `deepEqual(['/projects', 'project01'])`, would that be too generic and cause the test to fail? If not, I think it should be written more generally since it's expecting that this _didn't_ happen. Generally speaking, we want our tests to be as specific as possible, but for negative assertions, being more general actually is a stronger assertion.

Recently I had trouble with a `verify(mockedRouter(something_i_cannot_remember)).once()`. It was failing because I didn't write it correctly (maybe I needed to use `deepEqual`). I knew the test had a problem because it was failing. But in this case, if you wrote it even slightly wrong, it would silently pass.

This is similar to when we were doing `expect(someDebugNode).not.toBeUndefined()`, which was passing, because the negative assertion wasn't general enough (we should have been including `null`). Something like `.not.toBeFalsy()` would have included all values we cared about (`null` and `undefined`) and some that we don't care about but wouldn't ever be likely to occur, and if they did, would just result in a test failure (`false`, `''`, `0`, `NaN`, `-0`, `0n`).

(Same for the test below).
</blockquote></details>

Good points. `navigate(anything())` would be good here.

marksvc (commented):
We decided to take a different direction for the fix. I sent a new PR with the new fix and am closing this one.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/858#-:-MLJoBv2A5yM85i1zBez:b-9bfj4r)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

PR #862, opened 2020-11-04
SF-1121 Don't show not-registered projects in connect list

marksvc:
- We don't currently support connecting to a project that isn't listed
  in the PT Registry, like back translation projects can be.
- Omit such not-registered projects from the list returned by
  ParatextService GetProjectsAsync().

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/862)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -212,8 +212,13 @@ namespace SIL.XForge.Scripture.Services
         public async Task<IReadOnlyList<ParatextProject>> GetProjectsAsync(UserSecret userSecret)
         {
             IInternetSharedRepositorySource ptRepoSource = await GetInternetSharedRepositorySource(userSecret);
-            IEnumerable<SharedRepository> remotePtProjects = ptRepoSource.GetRepositories();
-            return GetProjects(userSecret, remotePtProjects, ptRepoSource.GetProjectsMetaData());
+            List<SharedRepository> remotePtProjects = ptRepoSource.GetRepositories().ToList();
+            List<ProjectMetadata> projectMetadata = ptRepoSource.GetProjectsMetaData().ToList();
+
+            // Omit projects that are not in the PT Registry until we support connecting to such projects.
+            remotePtProjects.RemoveAll((SharedRepository project) =>
+                !projectMetadata.Any((ProjectMetadata metadata) => metadata.ProjectGuid == project.SendReceiveId));
+            return GetProjects(userSecret, remotePtProjects, projectMetadata);
         }
 
         public async Task<Attempt<string>> TryGetProjectRoleAsync(UserSecret userSecret, string paratextId)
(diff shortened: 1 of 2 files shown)


irahopkinson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/862)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 220 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/862#-MLMieLD8OC4QptN2ME8:-MLMieLD8OC4QptN2ME9:b-35kpbk) ([raw file](https://github.com/sillsdev/web-xforge/blob/7702eede0d5b593caf52758f6d92ab1c6e76bae3/src/SIL.XForge.Scripture/Services/ParatextService.cs#L220)):*
> ```
> metadata.ProjectGuid == project.SendReceiveId
> ```
I'm failing to understand why this means that a project is not in the registry? It also makes code in `GetProjects` redundant (which is probably ok). However, given the code that is made redundant in `GetProjects`, my concern is that this will remove more projects than we want it to.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/862)* status: all files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 220 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/862#-MLMieLD8OC4QptN2ME8:-MLUeDju9TYvc4unZ6oj:b-yloqeg) ([raw file](https://github.com/sillsdev/web-xforge/blob/7702eede0d5b593caf52758f6d92ab1c6e76bae3/src/SIL.XForge.Scripture/Services/ParatextService.cs#L220)):*

> ```
> metadata.ProjectGuid == project.SendReceiveId
> ```
I'm failing to understand why this means that a project is not in the registry? It also makes code in `GetProjects` redundant (which is probably ok). However, given the code that is made redundant in `GetProjects`, my concern is that this will remove more projects than we want it to.
</blockquote></details>

Thank you. I agree, especially now that I have thought about SendReceiveAsync() more and that we had projects during migration that didn't have a full name available, and therefore must now have been in the PT Registry. 

Why it means that a project is not in the registry: 

remotePtProjects = ptRepoSource.GetRepositories() will likely end up in ParatextData InternetSharedRepositorySource GetRepositories(licenses), which fetches project hg repositories that are available to clone.

projectMetadata = ptRepoSource.GetProjectsMetaData() is a SF (not ParatextData) method in JwtInternetSharedRepositorySource, and will fetch projects from the Paratext Registry. 

So then 

remotePtProjects.RemoveAll( projects where 
  there isn't any projectMetadata item whose ProjectGuid matches the given project's SendReceiveId)

should remove from remotePtProjects all projects not in the registry. 
(Or I'm totally not correctly thinking about how the set logic works here, which is possible).

marksvc:
I converted the PR to a draft while I look into something.

marksvc:
I investigated and posted my findings and conclusion in a comment to SF-1121. I am marking this PR as ready for review.

irahopkinson (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/862)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #863, opened 2020-11-06
Add ability to import questions from Transcelerator

Nateowami:
## Importing status dialog

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/863)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -60,6 +60,8 @@
     "sillsdev",
     "teamcity",
     "textbox",
+    "textfield",
+    "transcelerator",
     "transloco",
     "uncheck",
     "uninvite",
--- a/src/RealtimeServer/scriptureforge/models/question.ts
+++ b/src/RealtimeServer/scriptureforge/models/question.ts
@@ -19,4 +19,5 @@ export interface Question extends ProjectData {
   dateArchived?: string;
   dateModified: string;
   dateCreated: string;
+  transceleratorQuestionId?: string;
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.html
@@ -3,6 +3,9 @@
     <mdc-icon fxFlex="32px">library_books</mdc-icon>
     <h2 fxFlex>{{ t("texts_with_questions") }}</h2>
     <div fxFlexAlign="center" *ngIf="isProjectAdmin; else overallProgressChart">
+      <button mdc-button type="button" *ngIf="hasTransceleratorQuestions" (click)="importDialog()" id="import-btn">
+        {{ t("import_questions") }}
+      </button>
       <button
         *ngIf="!isLoading"
         mdc-button
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts
@@ -35,6 +35,7 @@ import { SF_TYPE_REGISTRY } from '../../core/models/sf-type-registry';
 import { TextDocId } from '../../core/models/text-doc';
 import { SFProjectService } from '../../core/sf-project.service';
 import { CheckingModule } from '../checking.module';
+import { ImportQuestionsDialogComponent } from '../import-questions-dialog/import-questions-dialog.component';
 import { QuestionAnsweredDialogComponent } from '../question-answered-dialog/question-answered-dialog.component';
 import { QuestionDialogComponent } from '../question-dialog/question-dialog.component';
 import { QuestionDialogService } from '../question-dialog/question-dialog.service';
@@ -229,6 +230,15 @@ describe('CheckingOverviewComponent', () => {
       env.clickElement(env.questionEditButtons[0]);
       verify(mockedMdcDialog.open(QuestionAnsweredDialogComponent)).twice();
       verify(mockedQuestionDialogService.questionDialog(anything())).once();
+      expect().nothing();
+    }));
+
+    it('should open a dialog to import questions', fakeAsync(() => {
+      when(mockedProjectService.hasTransceleratorQuestions('project01')).thenResolve(true);
+      const env = new TestEnvironment();
+      env.waitForQuestions();
+      env.clickElement(env.importButton);
+      verify(mockedMdcDialog.open(ImportQuestionsDialogComponent, anything())).once();
       expect().nothing();
     }));
   });
@@ -723,6 +733,10 @@ class TestEnvironment {
 
   get archivedQuestions(): DebugElement {
     return this.fixture.debugElement.query(By.css('#text-with-archived-questions'));
+  }
+
+  get importButton(): DebugElement {
+    return this.fixture.debugElement.query(By.css('#import-btn'));
   }
 
   get noQuestionsLabel(): DebugElement {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.ts
@@ -20,6 +20,10 @@ import { TextDocId } from '../../core/models/text-doc';
 import { TextsByBookId } from '../../core/models/texts-by-book-id';
 import { SFProjectService } from '../../core/sf-project.service';
 import { CheckingAccessInfo, CheckingUtils } from '../checking.utils';
+import {
+  ImportQuestionsDialogComponent,
+  ImportQuestionsDialogData
+} from '../import-questions-dialog/import-questions-dialog.component';
 import { QuestionAnsweredDialogComponent } from '../question-answered-dialog/question-answered-dialog.component';
 import { QuestionDialogData } from '../question-dialog/question-dialog.component';
 import { QuestionDialogService } from '../question-dialog/question-dialog.service';
@@ -34,6 +38,7 @@ export class CheckingOverviewComponent extends DataLoadingComponent implements O
   itemVisibleArchived: { [bookIdOrDocId: string]: boolean } = {};
   texts: TextInfo[] = [];
   projectId?: string;
+  hasTransceleratorQuestions = false;
 
   private questionDocs = new Map<string, QuestionDoc[]>();
   private textsByBookId: TextsByBookId = {};
@@ -150,6 +155,8 @@ export class CheckingOverviewComponent extends DataLoadingComponent implements O
         }
         this.questionsQuery = await this.projectService.queryQuestions(projectId);
         this.initTexts();
+        this.hasTransceleratorQuestions =
+          this.isProjectAdmin && (await this.projectService.hasTransceleratorQuestions(projectId));
       } finally {
         this.loadingFinished();
       }
@@ -330,6 +337,18 @@ export class CheckingOverviewComponent extends DataLoadingComponent implements O
     };
     await this.questionDialogService.questionDialog(data);
     this.initTextsWithLoadingIndicator();
+  }
+
+  importDialog() {
+    if (this.projectDoc == null) {
+      return;
+    }
+    const data: ImportQuestionsDialogData = {
+      projectId: this.projectDoc.id,
+      userId: this.userService.currentUserId,
+      textsByBookId: this.textsByBookId
+    };
+    this.dialog.open(ImportQuestionsDialogComponent, { autoFocus: false, data });
   }
 
   getBookName(text: TextInfo): string {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking.module.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking.module.ts
@@ -24,6 +24,9 @@ import { CheckingQuestionsComponent } from './checking/checking-questions/checki
 import { CheckingTextComponent } from './checking/checking-text/checking-text.component';
 import { CheckingComponent } from './checking/checking.component';
 import { FontSizeComponent } from './checking/font-size/font-size.component';
+import { ImportQuestionsConfirmationDialogComponent } from './import-questions-dialog/import-questions-confirmation-dialog/import-question-confirmation-dialog.component';
+import { ImportQuestionsDialogComponent } from './import-questions-dialog/import-questions-dialog.component';
+import { ImportQuestionsProgressDialogComponent } from './import-questions-dialog/import-questions-progress-dialog/import-questions-progress-dialog.component';
 import { QuestionAnsweredDialogComponent } from './question-answered-dialog/question-answered-dialog.component';
 import { QuestionDialogComponent } from './question-dialog/question-dialog.component';
 
@@ -35,6 +38,9 @@ import { QuestionDialogComponent } from './question-dialog/question-dialog.compo
     CheckingTextComponent,
     CheckingAnswersComponent,
     QuestionDialogComponent,
+    ImportQuestionsDialogComponent,
+    ImportQuestionsConfirmationDialogComponent,
+    ImportQuestionsProgressDialogComponent,
     FontSizeComponent,
     CheckingOwnerComponent,
     CheckingCommentsComponent,
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-confirmation-dialog/import-question-confirmation-dialog.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-confirmation-dialog/import-question-confirmation-dialog.component.ts
@@ -1,0 +1,47 @@
+import { MDC_DIALOG_DATA, MDCDataTableRowSelectionChangedEvent, MdcDialogRef } from '@angular-mdc/web';
+import { Component, Inject } from '@angular/core';
+
+export interface ImportQuestionsConfirmationDialogData {
+  questions: EditedQuestion[];
+}
+
+export interface ImportQuestionsConfirmationDialogResult {
+  questions: EditedQuestion[];
+}
+
+export interface EditedQuestion {
+  before: string;
+  after: string;
+  answerCount: number;
+  checked: boolean;
+}
+
+@Component({
+  templateUrl: './import-questions-confirmation-dialog.component.html',
+  styleUrls: ['./import-questions-confirmation-dialog.component.scss']
+})
+export class ImportQuestionsConfirmationDialogComponent {
+  questions: EditedQuestion[];
+
+  constructor(
+    @Inject(MDC_DIALOG_DATA) data: ImportQuestionsConfirmationDialogData,
+    private readonly dialogRef: MdcDialogRef<
+      ImportQuestionsConfirmationDialogComponent,
+      ImportQuestionsConfirmationDialogResult
+    >
+  ) {
+    this.questions = data.questions;
+  }
+
+  onSelectionChanged(event: MDCDataTableRowSelectionChangedEvent): void {
+    this.questions[event.index].checked = event.selected;
+  }
+
+  selectAll(checked: boolean): void {
+    this.questions.forEach(question => (question.checked = checked));
+  }
+
+  submit() {
+    this.dialogRef.close({ questions: this.questions });
+  }
+}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-confirmation-dialog/import-questions-confirmation-dialog.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-confirmation-dialog/import-questions-confirmation-dialog.component.html
@@ -1,0 +1,41 @@
+<ng-container *transloco="let t; read: 'import_questions_confirmation_dialog'">
+  <mdc-dialog>
+    <mdc-dialog-container>
+      <mdc-dialog-surface>
+        <mdc-dialog-title>{{ t("some_questions_edited") }}</mdc-dialog-title>
+        <mdc-dialog-content>
+          {{ t("edited_questions_explanation") }}
+          <mdc-data-table
+            (selectionChanged)="onSelectionChanged($event)"
+            (selectedAll)="selectAll(true)"
+            (unselectedAll)="selectAll(false)"
+          >
+            <table mdcDataTableTable>
+              <thead>
+                <tr mdcDataTableHeaderRow>
+                  <th mdcDataTableHeaderCell><mdc-checkbox></mdc-checkbox></th>
+                  <th>{{ t("scripture_forge") }}</th>
+                  <th>{{ t("transcelerator") }}</th>
+                  <th>{{ t("number_answers") }}</th>
+                </tr>
+              </thead>
+              <tbody mdcDataTableContent>
+                <tr mdcDataTableRow *ngFor="let listItem of questions" [selected]="listItem.checked">
+                  <td><mdc-checkbox [checked]="listItem.checked"></mdc-checkbox></td>
+                  <td>{{ listItem.before }}</td>
+                  <td>{{ listItem.after }}</td>
+                  <td>{{ listItem.answerCount }}</td>
+                </tr>
+              </tbody>
+            </table>
+          </mdc-data-table>
+        </mdc-dialog-content>
+        <mdc-dialog-actions>
+          <button mdcDialogButton mdcDialogAction="close" unelevated type="submit" (click)="submit()">
+            {{ t("done") }}
+          </button>
+        </mdc-dialog-actions>
+      </mdc-dialog-surface>
+    </mdc-dialog-container>
+  </mdc-dialog>
+</ng-container>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-confirmation-dialog/import-questions-confirmation-dialog.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-confirmation-dialog/import-questions-confirmation-dialog.component.scss
@@ -1,0 +1,11 @@
+th:first-child {
+  padding-left: 0;
+}
+
+tr td:not(:last-child) {
+  padding-right: 0.5em;
+}
+
+table {
+  white-space: initial;
+}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-confirmation-dialog/import-questions-confirmation-dialog.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-confirmation-dialog/import-questions-confirmation-dialog.component.spec.ts
@@ -1,0 +1,194 @@
+import { MdcDialog, MdcDialogRef } from '@angular-mdc/web/dialog';
+import { OverlayContainer } from '@angular/cdk/overlay';
+import { CommonModule } from '@angular/common';
+import { Component, Directive, NgModule, ViewChild, ViewContainerRef } from '@angular/core';
+import { ComponentFixture, fakeAsync, flush, TestBed, tick } from '@angular/core/testing';
+import { FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { configureTestingModule, TestTranslocoModule } from 'xforge-common/test-utils';
+import { UICommonModule } from 'xforge-common/ui-common.module';
+import {
+  EditedQuestion,
+  ImportQuestionsConfirmationDialogComponent,
+  ImportQuestionsConfirmationDialogData,
+  ImportQuestionsConfirmationDialogResult
+} from './import-question-confirmation-dialog.component';
+
+describe('ImportQuestionsConfirmationDialogComponent', () => {
+  configureTestingModule(() => ({
+    imports: [ReactiveFormsModule, FormsModule, DialogTestModule]
+  }));
+
+  it('Allows selecting and unselecting all questions', fakeAsync(() => {
+    const env = new TestEnvironment();
+    expect(env.questionRows.length).toBe(2);
+    expect(env.rowCheckboxes.length).toBe(2);
+
+    expect(env.selectAllCheckbox.checked).toBe(true);
+    env.rowCheckboxes.forEach(n => expect(n.checked).toBe(true));
+
+    env.click(env.selectAllCheckbox);
+    expect(env.selectAllCheckbox.checked).toBe(false);
+    env.rowCheckboxes.forEach(n => expect(n.checked).toBe(false));
+  }));
+
+  it('Can handle a single question', fakeAsync(() => {
+    const env = new TestEnvironment(1);
+    expect(env.questionRows.length).toBe(1);
+    expect(env.rowCheckboxes.length).toBe(1);
+
+    expect(env.selectAllCheckbox.checked).toBe(true);
+    expect(env.rowCheckboxes[0].checked).toBe(true);
+
+    env.click(env.rowCheckboxes[0]);
+    expect(env.rowCheckboxes[0].checked).toBe(false);
+    expect(env.selectAllCheckbox.checked).toBe(false);
+
+    env.click(env.selectAllCheckbox);
+    expect(env.selectAllCheckbox.checked).toBe(true);
+    expect(env.rowCheckboxes[0].checked).toBe(true);
+
+    env.closeDialog().then(value =>
+      expect(value.questions).toEqual([
+        {
+          before: 'Original question 1',
+          after: 'Edited question 1',
+          answerCount: 0,
+          checked: true
+        }
+      ])
+    );
+  }));
+
+  it('Allows selecting a subset of questions', fakeAsync(() => {
+    const env = new TestEnvironment(3);
+    expect(env.questionRows.length).toBe(3);
+    expect(env.rowCheckboxes.length).toBe(3);
+
+    expect(env.selectAllCheckbox.checked).toBe(true);
+    env.click(env.selectAllCheckbox);
+    expect(env.selectAllCheckbox.checked).toBe(false);
+
+    env.click(env.rowCheckboxes[2]);
+    expect(env.rowCheckboxes[2].checked).toBe(true);
+    expect(env.selectAllCheckbox.indeterminate).toBe(true);
+
+    env.closeDialog().then(value =>
+      expect(value.questions).toEqual([
+        {
+          before: 'Original question 1',
+          after: 'Edited question 1',
+          answerCount: 0,
+          checked: false
+        },
+        {
+          before: 'Original question 2',
+          after: 'Edited question 2',
+          answerCount: 0,
+          checked: false
+        },
+        {
+          before: 'Original question 3',
+          after: 'Edited question 3',
+          answerCount: 0,
+          checked: true
+        }
+      ])
+    );
+  }));
+});
+
+@Directive({
+  // ts lint complains that a directive should be used as an attribute
+  // tslint:disable-next-line:directive-selector
+  selector: 'viewContainerDirective'
+})
+class ViewContainerDirective {
+  constructor(public viewContainerRef: ViewContainerRef) {}
+}
+
+@Component({
+  selector: 'app-view-container',
+  template: '<viewContainerDirective></viewContainerDirective>'
+})
+class ChildViewContainerComponent {
+  @ViewChild(ViewContainerDirective, { static: true }) viewContainer!: ViewContainerDirective;
+
+  get childViewContainer(): ViewContainerRef {
+    return this.viewContainer.viewContainerRef;
+  }
+}
+
+@NgModule({
+  imports: [CommonModule, UICommonModule, TestTranslocoModule],
+  declarations: [ViewContainerDirective, ChildViewContainerComponent, ImportQuestionsConfirmationDialogComponent],
+  exports: [ViewContainerDirective, ChildViewContainerComponent, ImportQuestionsConfirmationDialogComponent],
+  entryComponents: [ChildViewContainerComponent, ImportQuestionsConfirmationDialogComponent]
+})
+class DialogTestModule {}
+
+class TestEnvironment {
+  fixture: ComponentFixture<ChildViewContainerComponent>;
+  component: ImportQuestionsConfirmationDialogComponent;
+  dialogRef: MdcDialogRef<ImportQuestionsConfirmationDialogComponent>;
+  overlayContainerElement: HTMLElement;
+
+  constructor(questionCount: number = 2) {
+    this.fixture = TestBed.createComponent(ChildViewContainerComponent);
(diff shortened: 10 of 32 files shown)


Nateowami (commented):
So, the tests in the RealtimeServer are failing, because it never occurred to me that I needed to run them. I'll work on fixing them. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/863#-:-MLQnGbc0Hpj9ssxnwLO:bdfv8fe)* status: 0 of 33 files reviewed, all discussions resolved

RaymondLuong3 (commented):
Amazing work! Most of my comments are simple fixes.

> Reviewed 33 of 33 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/863#-:-MLU611sA9SdrB5H_reb:b-ath5gv)* status: all files reviewed, 14 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLU5B7t4G4uNcRqOJ3Z:-MLU5B7t4G4uNcRqOJ3_:bu9z5d0) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge/Services/TransceleratorService.cs#L33)):*

> ```
>         public bool HasQuestions(string paratextId)
>         {
>             return QuestionFiles(paratextId).Count() > 0;
>         }
> ```
</details>

I don't see a real advantage of having this method, since it is nearly as expensive as calling `TransceleratorService.Questions()`. This could just be removed.
___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 41 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLTntUoCYW6scwBGfxJ:-MLTntUoCYW6scwBGfxK:blne0yq) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge/Services/TransceleratorService.cs#L41)):*
> ```
> new int[] { 1, 1 }
> ```
You should be able to pull this value out into `appsettings.json` and then create a `TransceleratorOptions.cs` just like how you are using `SiteOptions` above.
___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 73 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLTp_BN6PYchdkBzn4I:-MLTp_BN6PYchdkBzn4J:b-asielp) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge/Services/TransceleratorService.cs#L73)):*
> ```
> _fileService.DirectoryExists(pathToFiles) ?
> ```
Nit: I've seen the question mark more often place on the next line.
___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 91 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLTqCPyEDjDz7Cx8gfp:-MLTqCPyEDjDz7Cx8gfq:b-pdp8tm) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge/Services/TransceleratorService.cs#L91)):*
> ```
> .DefaultIfEmpty("").SingleOrDefault();
> ```
So this means there is a possibility of an empty question if it's hasn't been translated. Do we handle that?
___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 118 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLTrVK66vPCP_I7YJwX:-MLTrVK66vPCP_I7YJwY:b-cyxzo7) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge/Services/TransceleratorService.cs#L118)):*
> ```
>  return versionStrings.Length >= numbersToSatisfy.Length;
> ```
So transcelerator version 1.1.1 would satisfy 1.1. Ok
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLTst8D1WFr0GF3DSWA:-MLTst8D1WFr0GF3DSWB:b3lztpo) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L8)):*
> ```
>             <mdc-data-table>
>               <table mdcDataTableTable>
>                 <thead>
> ```
mdcDataTableTable - That's an unfortunate name.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 154 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLTwUGcEq4f2EiV0s8y:-MLTwUGcEq4f2EiV0s8z:b-v05fwb) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L154)):*
> ```
>   get selectedCount() {
>     return this.questionList.filter(question => question.checked).length;
>   }
> ```
I would group getters below the constructor.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 200 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLTz5ys5iHp030jZcCK:-MLTz5ys5iHp030jZcCL:b-e5wfn3) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L200)):*
> ```
>     for (const listItem of this.questionList) {
>       if (listItem.matchesFilter) {
> ```
could this be simplified to `for (const listItem of this.filteredList)`
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 221 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLU1KY6CgFzOj1PigbZ:-MLU1KY6CgFzOj1Pigb_:b-i3mmvi) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L221)):*
> ```
> isBetweenRefs(ref: VerseRef, from: VerseRef | null, to?: VerseRef | null): boolean {
> ```
This can be private.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 228 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLU1Mev-YUnP8qfKV4c:-MLU1Mev-YUnP8qfKV4d:bwf739k) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L228)):*
> ```
> async confirmEditsIfNecessary(changes: DialogListItem[]): Promise<void> {
> ```
This method can be private.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 273 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLU-Nb6CbIJ8UwIwAYO:-MLU-Nb6CbIJ8UwIwAYP:b-y7flky) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L273)):*
> ```
> listItem.alreadyImportedQuestion.data &&
> ```
This should include checking != null
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 277 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLU-jEODaNqJ4x7HZnL:-MLU-jEODaNqJ4x7HZnM:b-9zqs1w) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L277)):*
> ```
> op.set(q => q.text!, listItem.question.text).set(q => q.dateModified, currentDate)
> ```
Cool, I didn't realize you could chain operations like this.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-progress-dialog/import-questions-progress-dialog.component.html, line 5 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLU2ieUEC51F8zj6l3e:-MLU2ieV7sAPeM3bySfH:b-jtk4yv) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-progress-dialog/import-questions-progress-dialog.component.html#L5)):*
> ```
> <mdc-dialog-title>Importing {{ data.count }} questions...</mdc-dialog-title>
> ```
You will want to internationalize this title.
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 104 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLU3ONa5HNaMo24qZMe:-MLU3ONa5HNaMo24qZMf:b-8o5bl4) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L104)):*
> ```
> version 1.5.3."
> ```
Do we expect this version to change at all. I would pull this out into the `environments.ts` file and read it from there.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/863)* status: 30 of 33 files reviewed, 9 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLU5B7t4G4uNcRqOJ3Z:-MLU85vI8qyAvlrHM0mS:bo20dou) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge/Services/TransceleratorService.cs#L33)):*

> ```
>         public bool HasQuestions(string paratextId)
>         {
>             return QuestionFiles(paratextId).Count() > 0;
>         }
> ```
I don't see a real advantage of having this method, since it is nearly as expensive as calling `TransceleratorService.Questions()`. This could just be removed.
</blockquote></details>

It's exposed as an API method for determining whether to show the button to open the dialog. We _could_ have the front end fetch questions and check the length of returned array, but that requires sending a _lot_ more data to the front end. Additionally, I don't think this is anywhere near as expensive. All it does enumerate files in a directory and filter for those that match a fairly simple regex. `TransceleratorService.Questions()` opens those (potentially huge) files, parses the XML, and manipulates it to create a list of questions for the front end. In most cases the user will not click the button, and just knowing whether to show it will be sufficient. `TransceleratorService.Questions()` is also far more likely to throw an error, since there are far more places for things to go wrong.
___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 41 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLTntUoCYW6scwBGfxJ:-MLU9h2G9aihebMLlR4d:blafsfz) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge/Services/TransceleratorService.cs#L41)):*

> ```
> new int[] { 1, 1 }
> ```
You should be able to pull this value out into `appsettings.json` and then create a `TransceleratorOptions.cs` just like how you are using `SiteOptions` above.
</blockquote></details>

I don't really see a reason for putting this in a settings file, since it's more a property of the surrounding code than anything else. Tom Bogle added the ID attribute to each question node for the sake of Scripture Forge, and a new version of Transcelerator was released. The schema version when this change was introduce was bumped to 1.1, and so this code, before reading properties that might not exist, checks that the schema version is at least 1.1, since that's what it was written to work with. The only reason we might change this value is because we change the lines that follow to make more assumptions about what the XML document will contain.
___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 73 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLTp_BN6PYchdkBzn4I:-MLUBX5P69cmvzL5CBYq:b-qu6n94) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge/Services/TransceleratorService.cs#L73)):*

> ```
> _fileService.DirectoryExists(pathToFiles) ?
> ```
Nit: I've seen the question mark more often place on the next line.
</blockquote></details>

Good point. Done.
___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 91 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLTqCPyEDjDz7Cx8gfp:-MLUBmIO9XKj63w9qx1a:bhpfwd6) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge/Services/TransceleratorService.cs#L91)):*

> ```
> .DefaultIfEmpty("").SingleOrDefault();
> ```
So this means there is a possibility of an empty question if it's hasn't been translated. Do we handle that?
</blockquote></details>

We use this both for the question text and the answer text. The only way we could end up with the default is if `nodes.Where(node => lang.Equals(node.Attributes["xml:lang"].Value))` ended up with length 0, meaning there is no node matching a given language. I doubt Transcelerator would generate such a file, but it seemed better to default to an empty string than to throw an exception, especially since it could be the answer text, which we're not doing anything with anyway (perhaps I should just rip the example answer out of the model).

I would assume users can have empty questions (I wouldn't expect Transcelerator to throw an error dialog at a user saying "you may not leave this field until you type at least one character in it"). Therefore, the possibility of empty questions probably exists already. I suppose we could choose to hide them. Right now we treat them as any other question and let the user do with them as they want.
___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 118 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLTrVK66vPCP_I7YJwX:-MLUDgzm7zy9VNUJCMW9:bdvfqbb) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge/Services/TransceleratorService.cs#L118)):*

> ```
>  return versionStrings.Length >= numbersToSatisfy.Length;
> ```
So transcelerator version 1.1.1 would satisfy 1.1. Ok
</blockquote></details>

I'll add a comment where we check the version above, that it's a schema version, not the version of Transcelerator. (But the schema is defined by Transcelerator, so they're related).
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLTst8D1WFr0GF3DSWA:-MLUDxqwBkQWn0mgeWLv:bo6apac) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L8)):*

> ```
>             <mdc-data-table>
>               <table mdcDataTableTable>
>                 <thead>
> ```
mdcDataTableTable - That's an unfortunate name.
</blockquote></details>

Indeed.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 154 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLTwUGcEq4f2EiV0s8y:-MLUkKLRBscOKXBirSyi:b-qu6n94) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L154)):*

> ```
>   get selectedCount() {
>     return this.questionList.filter(question => question.checked).length;
>   }
> ```
I would group getters below the constructor.
</blockquote></details>

Good point. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 200 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLTz5ys5iHp030jZcCK:-MLUjbJqDU8R-KJES7C4:boj1qgt) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L200)):*

> ```
>     for (const listItem of this.questionList) {
>       if (listItem.matchesFilter) {
> ```
could this be simplified to `for (const listItem of this.filteredList)`
</blockquote></details>

Good catch. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 221 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLU1KY6CgFzOj1PigbZ:-MLUjIAI4E6FTVXFK4kR:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L221)):*

> ```
> isBetweenRefs(ref: VerseRef, from: VerseRef | null, to?: VerseRef | null): boolean {
> ```
This can be private.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 228 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLU1Mev-YUnP8qfKV4c:-MLUivJM2N3oZMyFlOdK:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L228)):*

> ```
> async confirmEditsIfNecessary(changes: DialogListItem[]): Promise<void> {
> ```
This method can be private.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 273 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLU-Nb6CbIJ8UwIwAYO:-MLUE5FZ-fxetkjV7QgO:b-s8pow8) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L273)):*

> ```
> listItem.alreadyImportedQuestion.data &&
> ```
This should include checking != null
</blockquote></details>

That actually is a null check. I suppose it would be better to be explicit. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-progress-dialog/import-questions-progress-dialog.component.html, line 5 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLU2ieUEC51F8zj6l3e:-MLUfZGO6jrj6yHcNDu9:b-k2rwtz) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-progress-dialog/import-questions-progress-dialog.component.html#L5)):*

> ```
> <mdc-dialog-title>Importing {{ data.count }} questions...</mdc-dialog-title>
> ```
You will want to internationalize this title.
</blockquote></details>

Good catch. Fixed.
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 104 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLU3ONa5HNaMo24qZMe:-MLUEYWc2YyOyTsamLGJ:b9l7vbu) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L104)):*

> ```
> version 1.5.3."
> ```
Do we expect this version to change at all. I would pull this out into the `environments.ts` file and read it from there.
</blockquote></details>

The only reason it would change is if we change our back end code to make different assumptions about where Transcelerator will store files, or what the schema of those files will be. I don't foresee it happening, but if it did, it would be associated with a change to the schema version we check for on the back end (1.1 at present).

Nateowami (commented):
@raymondluong3 I've made a number of changes relating to verse ranges, which I had missed earlier.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/863#-:-MLVjLNK8mnymlQJ5q3N:bqydzqh)* status: 28 of 33 files reviewed, 9 unresolved discussions (waiting on @RaymondLuong3)

irahopkinson (changes requested):
> Reviewed 28 of 33 files at r1, 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/863)* status: all files reviewed, 15 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 11 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLhaNB72LQB7L6Cqw9m:-MLhaNBABDPmtn2nJmzm:berkshz) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/src/SIL.XForge/Services/TransceleratorService.cs#L11)):*
> ```
> public class TransceleratorQuestion
> ```
This would typically go in its own file in `..\Models\`.
___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 23 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLhaq_42lVZXZbKaF_M:-MLhaq_59dny5Ed_aFuL:bbhtpu8) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/src/SIL.XForge/Services/TransceleratorService.cs#L23)):*
> ```
> TransceleratorService
> ```
Any reason this is here rather than `src\SIL.XForge.Scripture` (which is where I would expect it to be since it is SF specific)?
___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 41 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLhbm8wES4rMsL8gKJf:-MLhbm8x2uASxpaMcr3m:b5m5e1w) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/src/SIL.XForge/Services/TransceleratorService.cs#L41)):*
> ```
> decalred
> ```
nit: typo
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 198 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLhn1-Q9AUNGct4V4Mo:-MLhn1-R73zXHjuU1RwG:b-r6hmo0) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L198)):*
> ```
> console.error(e);
> ```
These look like debug changes. Seems like you should revert changes to this file.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts, line 70 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLhnWlC-kNGf4l9NBHy:-MLhnWlC-kNGf4l9NBHz:bnhze7t) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L70)):*
> ```
>  as T
> ```
I'm not convinced the changes you made to this file are a good idea. Can you explain?
___
*[test/SIL.XForge.Tests/Services/TransceleratorServiceTests.cs, line 19 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLhevSE-QlnistI9nrr:-MLhevSE-QlnistI9nrs:bmbquax) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/test/SIL.XForge.Tests/Services/TransceleratorServiceTests.cs#L19)):*
> ```
> env.FileSystemService.DirectoryExists(Arg.Any<string>()).Returns(true);
> ```
I don't see why this needs to be declared twice?

irahopkinson (commented):
Ensure the issue number SF-1084 is in the commit message.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/863#-:-MLhoty7CT2ThqcrhqHX:b-og9xir)* status: all files reviewed, 15 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

RaymondLuong3 (approved):
I installed Transcelerator 1.5.4 and tried to import questions with it. It looks like version 1.5.4 changes the transcelerator filename, so the option to import is not up to date. On my machine, it becomes "Translations of Checking Questions.xml". This can be addressed in a follow-up commit, but just pointing that out here. I approve once Ira's comments are addressed.

> Reviewed 4 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/863#-:-MLhyS4410ARHLoYNxAS:bj7lxs9)* status: all files reviewed, 7 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLU5B7t4G4uNcRqOJ3Z:-MLhthNy7ugLFmaVE0V8:b-l6advu) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge/Services/TransceleratorService.cs#L33)):*

It's exposed as an API method for determining whether to show the button to open the dialog. We _could_ have the front end fetch questions and check the length of returned array, but that requires sending a _lot_ more data to the front end. Additionally, I don't think this is anywhere near as expensive. All it does enumerate files in a directory and filter for those that match a fairly simple regex. `TransceleratorService.Questions()` opens those (potentially huge) files, parses the XML, and manipulates it to create a list of questions for the front end. In most cases the user will not click the button, and just knowing whether to show it will be sufficient. `TransceleratorService.Questions()` is also far more likely to throw an error, since there are far more places for things to go wrong.
</blockquote></details>

That's a good explanation. I hadn't considered that more often than not a user would not use the import button when visiting the overview page. Thanks.
___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 91 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLTqCPyEDjDz7Cx8gfp:-MLhuPEAAup45VtYADGS:bce6jfl) ([raw file](https://github.com/sillsdev/web-xforge/blob/40eff87cb474053dd14b68680260573dbcba88e3/src/SIL.XForge/Services/TransceleratorService.cs#L91)):*

We use this both for the question text and the answer text. The only way we could end up with the default is if `nodes.Where(node => lang.Equals(node.Attributes["xml:lang"].Value))` ended up with length 0, meaning there is no node matching a given language. I doubt Transcelerator would generate such a file, but it seemed better to default to an empty string than to throw an exception, especially since it could be the answer text, which we're not doing anything with anyway (perhaps I should just rip the example answer out of the model).

I would assume users can have empty questions (I wouldn't expect Transcelerator to throw an error dialog at a user saying "you may not leave this field until you type at least one character in it"). Therefore, the possibility of empty questions probably exists already. I suppose we could choose to hide them. Right now we treat them as any other question and let the user do with them as they want.
</blockquote></details>

I agree, it is better to show an empty question rather than throwing an error.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 235 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLhxLjo4Xvws5DB8LEv:-MLhxLjo4Xvws5DB8LEw:b8ahtkq) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L235)):*

> ```
>               doc.verseRef.bookNum !== verseRefData.bookNum ||
>               doc.verseRef.chapterNum !== verseRefData.chapterNum ||
>               doc.verseRef.verseNum !== verseRefData.verseNum ||
>               doc.verseRef.verse !== verseRefData.verse)
> ```
</details>

This could be more clear with something like this `toVerseRef(doc.verseRef).BBBCCCVVV !== toVerseRef(verseRefData).BBBCCCVVV`

Nateowami (commented):
@raymondluong3 You have to confirm translations and tell Transcelerator to export files for Scripture Forge. The file you're seeing is used internally by Transcelerator and ignored by Scripture Forge. I assume you didn't get the import button? In Transcelerator the option is something like File > Export questions for Scripture Forge.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/863#-:-MLi3rRmCQoc7rNZpGn-:bco8if5)* status: all files reviewed, 6 unresolved discussions (waiting on @Nateowami)

RaymondLuong3 (changes requested):
Yes you are right. I must not have click the Generate Scripture Forge Files options. I am able to import the translated questions. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/863#-:-MLiBZhx5ZbwKn7cpzgH:b-3rl5pk)* status: all files reviewed, 7 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 276 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLiCZ657TqXbKiv6fZQ:-MLiCZ66-QkmhIbjefGH:bxkllhe) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L276)):*

> ```
>       verse:
>         // TODO Right now ignoring end chapter and verse if it's in a different chapter, since we don't yet support that
>         q.startChapter === q.endChapter && q.startVerse !== q.endVerse ? q.startVerse + '-' + q.endVerse : ''
>     };
> ```
</details>

You will want to set verse to undefined instead of the empty string. When you set it to the empty string, the checking component will not correctly set question icon and highlight for the question.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/863)* status: 27 of 33 files reviewed, 7 unresolved discussions (waiting on @irahopkinson and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 235 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLhxLjo4Xvws5DB8LEv:-MLiE1EB50lS6-BI7URv:bgj31rm) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L235)):*

> ```
>               doc.verseRef.bookNum !== verseRefData.bookNum ||
>               doc.verseRef.chapterNum !== verseRefData.chapterNum ||
>               doc.verseRef.verseNum !== verseRefData.verseNum ||
>               doc.verseRef.verse !== verseRefData.verse)
> ```
This could be more clear with something like this `toVerseRef(doc.verseRef).BBBCCCVVV !== toVerseRef(verseRefData).BBBCCCVVV`
</blockquote></details>

Thanks. Done (with a slight needed change).
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 276 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLiCZ657TqXbKiv6fZQ:-MLiEU2vCmq6xZR837BB:b-i21bbo) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L276)):*

> ```
>       verse:
>         // TODO Right now ignoring end chapter and verse if it's in a different chapter, since we don't yet support that
>         q.startChapter === q.endChapter && q.startVerse !== q.endVerse ? q.startVerse + '-' + q.endVerse : ''
>     };
> ```
You will want to set verse to undefined instead of the empty string. When you set it to the empty string, the checking component will not correctly set question icon and highlight for the question.
</blockquote></details>

Thanks. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 198 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLhn1-Q9AUNGct4V4Mo:-MLiMid141PbNe_bENa5:be6wjnq) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L198)):*

> ```
> console.error(e);
> ```
These look like debug changes. Seems like you should revert changes to this file.
</blockquote></details>

I made these changes because at present the error is basically dropped and the user can't determine what the cause was, even if the user is a developer. Most errors are sent to our error handler where they are logged and then reported. I was running into this error a lot and didn't like that I couldn't see the cause. I think this should be a permanent change, but perhaps should be in a different PR.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts, line 70 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLhnWlC-kNGf4l9NBHy:-MLiJs189zV8finUPkYC:b-nwxiia) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L70)):*

> ```
>  as T
> ```
I'm not convinced the changes you made to this file are a good idea. Can you explain?
</blockquote></details>

I only partially remember the reason for this, but for some reason if I revert it now, nothing breaks (I don't think that was the case at the time I wrote it, which seems odd). Reverted.
___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 11 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLhaNB72LQB7L6Cqw9m:-MLi8r7j-ITCWGt5GV8b:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/src/SIL.XForge/Services/TransceleratorService.cs#L11)):*

> ```
> public class TransceleratorQuestion
> ```
This would typically go in its own file in `..\Models\`.
</blockquote></details>

Done.
___
*[src/SIL.XForge/Services/TransceleratorService.cs, line 23 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLhaq_42lVZXZbKaF_M:-MLiCJQS4pWuNaoY6mT8:bbmtq0m) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/src/SIL.XForge/Services/TransceleratorService.cs#L23)):*

> ```
> TransceleratorService
> ```
Any reason this is here rather than `src\SIL.XForge.Scripture` (which is where I would expect it to be since it is SF specific)?
</blockquote></details>

Thanks. Moved it.
___
*[test/SIL.XForge.Tests/Services/TransceleratorServiceTests.cs, line 19 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLhevSE-QlnistI9nrr:-MLiCcLi-g5qVQylgkqj:b9cql6i) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/test/SIL.XForge.Tests/Services/TransceleratorServiceTests.cs#L19)):*

> ```
> env.FileSystemService.DirectoryExists(Arg.Any<string>()).Returns(true);
> ```
I don't see why this needs to be declared twice?
</blockquote></details>

Thanks. Fixed.

RaymondLuong3 (approved):
> Reviewed 8 of 9 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/863)* status: all files reviewed, 5 unresolved discussions (waiting on @irahopkinson)

irahopkinson (approved):
:lgtm:

> Reviewed 7 of 9 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/863#-:-MLlQaCj5RgQsO7dvGGY:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 198 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/863#-MLhn1-Q9AUNGct4V4Mo:-MLlO8SS8pWxqLMC9_jg:b-6rssz) ([raw file](https://github.com/sillsdev/web-xforge/blob/a4a06cbe5592a5c0c68360a11acbc48a1f736963/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L198)):*

I made these changes because at present the error is basically dropped and the user can't determine what the cause was, even if the user is a developer. Most errors are sent to our error handler where they are logged and then reported. I was running into this error a lot and didn't like that I couldn't see the cause. I think this should be a permanent change, but perhaps should be in a different PR.
</blockquote></details>

Ok, and I'm fine with them being in this PR. Perhaps a comment then.

PR #864, opened 2020-11-09
SF-1089 Fixed test breaking on some computers

pmachapman:
The `PutNotes_AddEditDeleteComment_ThreadCorrectlyUpdated` test has consistently broken on my laptop - I ran a quick step through and noticed that in ParatextService.cs, line 338:
```cs
var existingComment = existingThread?.Comments.FirstOrDefault(c => c.Id == comment.Id);
```
was consistently returning null.

Looking into it, it seems that the comment id is a concatenation of the thread, user, and date attributes.

To fix this, my change sets the date in the comment XML to a consistent value.

As to what causes the date value to be consistent for most other people I am not entirely sure - it could be raw CPU speed, CPU pipelining or caching, or the maybe the JIT doing a neat trick on platforms it can do it?

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/864)


irahopkinson (approved):
:lgtm: passes now for me too.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/864#-:-MLlU6iPCqmkeelCskNi:b-u6d3xt)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #836, opened 2020-10-13
DBL Resource Support

pmachapman:
- Support for source projects to be DBL resources
- DBL resources are displayed in the "Based on" translation drop down, below other projects
- Added model changes for book permissions for both source and target projects
- Implemented book permissions for source projects when they are resources

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/836)

PR #867, opened 2020-11-10
Splitting of Source and Target Texts and Projects

pmachapman:
The following changes are made:

* The Source TextData is split from the Target Project into its own project
* The textType segment of the TextData identifier (texts:$._id) is retained, but obsolete (kept for future use)
* Adds a “ProjectRef” attribute to the TranslateSource object
* Adds a “resources” array to the Site object (users:$.sites.sf)
* At synchronization time, both source and target projects are synchronized separately, but sequentially
* Added a Migrator program to perform an offline database and filesystem upgrade

Also includes all of the changes of https://github.com/sillsdev/web-xforge/pull/836:

* Support for source projects to be DBL resources
* DBL resources are displayed in the "Based on" translation drop down, below other projects
* Added model changes for book permissions for both source and target projects
* Implemented book permissions for source projects when they are resources

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)


ddaspit (changes requested):
> Reviewed 74 of 75 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)* status: all files reviewed, 17 unresolved discussions (waiting on @pmachapman)

___
*[src/Migrations/SourceTargetSplitting/IObjectMigrator.cs, line 8 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLlB02H0Rh8jG3k9oLv:-MLlB02I2xfp3yPKjQPX:b8nvg86) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/Migrations/SourceTargetSplitting/IObjectMigrator.cs#L8)):*
> ```Smalltalk
>     /// The object migrator interface.
>     /// </summary>
>     public interface IObjectMigrator
> ```

I don't think this interface is necessary.
___
*[src/RealtimeServer/common/realtime-server.ts, line 185 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLkqUJV5WUZETl5swDy:-MLkqUJV5WUZETl5swDz:bgbtart) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/RealtimeServer/common/realtime-server.ts#L185)):*
> ```XML
>       if (user.sites != null) {
>         for (const key in user.sites) {
>           if (user.sites[key].resources != null && user.sites[key].resources.includes(resourceId)) {
> ```

You should use the correct site Id here. I don't think the realtime server has the site Id, so you will need to pass it in when the realtime server is started. You can do this in the `RealtimeService` class on the .NET side. Check out the `CreateOptions()` method. You can get the Id from the `SiteOptions`.
___
*[src/RealtimeServer/common/models/site.ts, line 3 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLl9z5m8RjJWQZj4jD8:-MLl9z5m8RjJWQZj4jD9:bnbtjgj) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/RealtimeServer/common/models/site.ts#L3)):*
> ```XML
> export interface Site {
>   projects: string[];
>   resources: string[];
> ```

Do we need to have separate `projects` and `resources` collections? Would it be possible to just have a single `projects` collection?
___
*[src/RealtimeServer/common/services/project-service.ts, line 13 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLks3HF0jCMfhp7JsKJ:-MLks3HF0jCMfhp7JsKK:bxmyoby) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/RealtimeServer/common/services/project-service.ts#L13)):*
> ```XML
>   protected readonly immutableProps = [this.pathTemplate(p => p.name), this.pathTemplate(p => p.userRoles)];
> 
>   protected async allowRead(_docId: string, doc: T, session: ConnectSession): Promise<boolean> {
> ```

You can rename `_docId` to `docId` now that is being used in the method.
___
*[src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts, line 41 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLksz9i8TmZWvvTD0dp:-MLksz9i8TmZWvvTD0dq:b-i8586w) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts#L41)):*
> ```XML
>           if (doc.data.userRoles.hasOwnProperty(userId)) {
>             if (
>               doc.data.userRoles[userId] === SFProjectRole.ParatextTranslator ||
> ```

I assume that the correct permissions are getting set elsewhere.
___
*[src/SIL.XForge.Scripture/Models/TextData.cs, line 13 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLknfDJ7-GDeAGU_Ni1:-MLknfDJ7-GDeAGU_Ni2:b6xc9e4) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Models/TextData.cs#L13)):*
> ```Smalltalk
>     {
>         public static string GetTextDocId(string projectId, int book, int chapter,
>             TextType textType = TextType.Target)
> ```

I think it would be better just to get rid of the `textType` parameter altogether.
___
*[src/SIL.XForge.Scripture/Models/TextInfo.cs, line 14 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLknyjnEDGzDNjRdWVL:-MLknyjnEDGzDNjRdWVM:b-qz3mt9) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Models/TextInfo.cs#L14)):*
> ```Smalltalk
>     public class TextTypeUtils
>     {
>         public static string DirectoryName(TextType textType)
> ```

We can probably just get rid of this method.
___
*[src/SIL.XForge.Scripture/Services/IParatextService.cs, line 17 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLktYCD2Mj_LJRn8snl:-MLktYCD2Mj_LJRn8snm:b-stc2kk) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/IParatextService.cs#L17)):*
> ```Smalltalk
>         Task<Attempt<string>> TryGetProjectRoleAsync(UserSecret userSecret, string paratextId);
>         Task<IReadOnlyDictionary<string, string>> GetProjectRolesAsync(UserSecret userSecret, string projectId);
>         bool IsProjectLanguageRightToLeft(UserSecret userSecret, string ptProjectId, TextType textType = TextType.Target);
> ```

I would just go ahead and remove the `textType` parameter everywhere.
___
*[src/SIL.XForge.Scripture/Services/IParatextService.cs, line 20 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLktnyD6CtEtLJJ97qa:-MLktnyD6CtEtLJJ97qb:bvzt1za) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/IParatextService.cs#L20)):*
> ```Smalltalk
> 
>         IReadOnlyList<ParatextResource> GetResources(UserSecret userSecret);
>         Task<string> GetResourcePermissionAsync(UserSecret _userSecret, string paratextId, string userId);
> ```

Why does `_userSecret` have an underscore?
___
*[src/SIL.XForge.Scripture/Services/JwtRestClient.cs, line 30 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLkvWKh3vqbXoiZYapQ:-MLkvWKh3vqbXoiZYapR:b4hbrbe) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/JwtRestClient.cs#L30)):*
> ```Smalltalk
>             }
> 
>             return ReflectionHelperLite.GetResult(this, "GetInternal", options, cgiCall, "HEAD", queryvars) as string;
> ```

@irahopkinson @marksvc  Do we have a plan to access this method without using reflection?
___
*[src/SIL.XForge.Scripture/Services/LazyScrTextCollection.cs, line 17 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLky4t3DzphGz2qMRbY:-MLky4t3DzphGz2qMRbZ:b-4lp21g) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/LazyScrTextCollection.cs#L17)):*
> ```Smalltalk
>         /// A simple in memory cache that lasts the until the end of the request.
>         /// </summary>
>         private readonly ConcurrentDictionary<string, ScrText> scrTextCache
> ```

I believe that we were intentionally not caching `ScrText` objects. It could potentially cause threading and memory issues. Why did you need to add the caching?
___
*[src/SIL.XForge.Scripture/Services/SFInstallableDBLResource.cs, line 29 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLl4wBN1lABvOBEF3sC:-MLl4wBN1lABvOBEF3sD:b-k354y9) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/SFInstallableDBLResource.cs#L29)):*
> ```Smalltalk
>     /// The Scripture Forge Installable DBL Resource Implementation.
>     /// </summary>
>     public class SFInstallableDBLResource : InstallableResource
> ```

I was wondering if much of this code copied from ParatextData. If so, please add a comment stating this. It will help to remind us to work with the Paratext team to get access to the code that we need.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 453 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLkzJ560rfADxgqw-W0:-MLkzJ57EfvPjxDUBYZT:b3jtig5) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L453)):*
> ```Smalltalk
>         {
>             SFProject project = await GetProjectAsync(projectId);
>             if (project != null && project.ParatextId.Length == SFInstallableDBLResource.ResourceIdentifierLength)
> ```

I think we should throw an exception if the project isn't a resource project.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 473 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLkzxOC4nSUP-hOa7fx:-MLkzxOC4nSUP-hOa7fy:b-r3ykfx) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L473)):*
> ```Smalltalk
>                                 // If the user has the resource
>                                 if (!string.IsNullOrWhiteSpace(siteId) && userDoc.Data.Sites.ContainsKey(siteId)
> ```

I don't think you need to check if `siteId` is null or empty. You can further simplify this by using `TryGetValue` to both check if the site exists and get it in one call.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 601 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLl14O28CqWvQoDe0BJ:-MLl14O28CqWvQoDe0BK:bc4hfpd) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L601)):*
> ```Smalltalk
>             using (IConnection conn = await RealtimeService.ConnectAsync(curUserId))
>             {
>                 if (this.RealtimeService.QuerySnapshots<SFProject>().Any(
> ```

I assume that we want to ensure that only one project should be created for a particular resource. Using an if statement like this could cause a race condition. Maybe we could use the Paratext Id as the project Id, since that is what uniquely identifies the resource. I believe that ShareDB will throw if a doc with that Id already exists.
___
*[src/SIL.XForge.Scripture/Services/SFTextCorpusFactory.cs, line 52 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLl3GcJ8SN3BvcXgGPE:-MLl3GcJ8SN3BvcXgGPF:b8nm5w4) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/SFTextCorpusFactory.cs#L52)):*
> ```Smalltalk
>                 string textCorpusProjectId = type switch
>                 {
>                     TextCorpusType.Source => project.TranslateConfig.Source?.ProjectRef ?? projectId,
> ```

We should throw if we cannot get the source project Id.
___
*[src/SIL.XForge.Scripture/Services/SyncService.cs, line 58 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLl459OA_xsfSsnqmS_:-MLl459OA_xsfSsnqmSa:b-jn6z8l) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/SyncService.cs#L58)):*
> ```Smalltalk
>             {
>                 // We need to sync the source first so that we can link the source texts and train the engine
>                 var parentId = _backgroundJobClient.Enqueue<ParatextSyncRunner>(r => r.RunAsync(sourceProjectId, curUserId, false));
> ```

Please specify the type unless it is explicit in the assignment.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)* status: 44 of 83 files reviewed, 17 unresolved discussions (waiting on @ddaspit and @pmachapman)

___
*[src/Migrations/SourceTargetSplitting/IObjectMigrator.cs, line 8 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLlB02H0Rh8jG3k9oLv:-MLnnWlV9ZrhcEFOMkFj:by7g087) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/Migrations/SourceTargetSplitting/IObjectMigrator.cs#L8)):*

I don't think this interface is necessary.
</blockquote></details>

No, its not - I have removed it. When I started, I thought I might need some mock unit tests, but after writing it, I realized that mock unit tests wouldn't really test anything of value.
___
*[src/RealtimeServer/common/realtime-server.ts, line 185 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLkqUJV5WUZETl5swDy:-MLnpiGm-yo8yfQKG4XG:blr9maf) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/RealtimeServer/common/realtime-server.ts#L185)):*

You should use the correct site Id here. I don't think the realtime server has the site Id, so you will need to pass it in when the realtime server is started. You can do this in the `RealtimeService` class on the .NET side. Check out the `CreateOptions()` method. You can get the Id from the `SiteOptions`.
</blockquote></details>

Thanks for pointing out how to do that - I have added it.
___
*[src/RealtimeServer/common/models/site.ts, line 3 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLl9z5m8RjJWQZj4jD8:-MLnbafO87CTKZjqZPo_:b4e4mhv) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/RealtimeServer/common/models/site.ts#L3)):*

Do we need to have separate `projects` and `resources` collections? Would it be possible to just have a single `projects` collection?
</blockquote></details>

When we add items to the projects collection, they appear in the selection dropdown in the top left corner of the screen - we probably don't want resources appearing there?

It also helps neatly divide resources from projects from a system perspective (the projects collection has a reverse mapping in projects.userRoles, which resources does not have).
___
*[src/RealtimeServer/common/services/project-service.ts, line 13 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLks3HF0jCMfhp7JsKJ:-MLnx6LK8s6QESFuqVOH:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/RealtimeServer/common/services/project-service.ts#L13)):*

You can rename `_docId` to `docId` now that is being used in the method.
</blockquote></details>

Done.
___
*[src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts, line 41 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLksz9i8TmZWvvTD0dp:-MLnyh4WErY26CWcEkWY:biilfmy) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts#L41)):*

I assume that the correct permissions are getting set elsewhere.
</blockquote></details>

This is effectively a holder for the new chapter permissions (with the migration script matching the current permissions). The code to get the chapter permissions from Paratext is not yet written (it is a separate job), but once written, will be set on sync.
___
*[src/SIL.XForge.Scripture/Models/TextData.cs, line 13 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLknfDJ7-GDeAGU_Ni1:-MLo0x2T-_Y5ISZem_RY:bss4a2b) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Models/TextData.cs#L13)):*

I think it would be better just to get rid of the `textType` parameter altogether.
</blockquote></details>

Removed.
___
*[src/SIL.XForge.Scripture/Models/TextInfo.cs, line 14 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLknyjnEDGzDNjRdWVL:-MLo0vqv1Y5yXVfm1qDN:bss4a2b) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Models/TextInfo.cs#L14)):*

We can probably just get rid of this method.
</blockquote></details>

Removed.
___
*[src/SIL.XForge.Scripture/Services/IParatextService.cs, line 17 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLktYCD2Mj_LJRn8snl:-MLo0__DEDncbE56eGm3:bwf5zwc) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/IParatextService.cs#L17)):*

I would just go ahead and remove the `textType` parameter everywhere.
</blockquote></details>

Removed. I was thinking of retaining it as I have had to keep it in the directory structure and textdata id, but it should be easy to add back in if required.
___
*[src/SIL.XForge.Scripture/Services/IParatextService.cs, line 20 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLktnyD6CtEtLJJ97qa:-MLo65Mv8Jvxpz6IcKWS:b-ze9ssz) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/IParatextService.cs#L20)):*

Why does `_userSecret` have an underscore?
</blockquote></details>

Bad copy-paste...fixed.
___
*[src/SIL.XForge.Scripture/Services/LazyScrTextCollection.cs, line 17 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLky4t3DzphGz2qMRbY:-MLo--wQCaQYBS5Y4jN4:bq75qal) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/LazyScrTextCollection.cs#L17)):*

I believe that we were intentionally not caching `ScrText` objects. It could potentially cause threading and memory issues. Why did you need to add the caching?
</blockquote></details>

During my testing, I noticed that FindById was hammering IO as it was run many times during sync. As some resources use an older SSF format instead of settings.xml, changes I made doubled the IO for those resources on sync.

I tried to mitigate the thread issues with the cache, but yeah, the more I think about it the potential for random hard to reproduce issues looms large.

I've removed it. If IO is an issue during QA, we can look at another method.
___
*[src/SIL.XForge.Scripture/Services/SFInstallableDBLResource.cs, line 29 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLl4wBN1lABvOBEF3sC:-MLoCEqB5luHjwlz4B_g:brcpjv9) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/SFInstallableDBLResource.cs#L29)):*

I was wondering if much of this code copied from ParatextData. If so, please add a comment stating this. It will help to remind us to work with the Paratext team to get access to the code that we need.
</blockquote></details>

Good idea - I have written a detailed remarks comment that states this, and outlines key differences.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 453 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLkzJ560rfADxgqw-W0:-MLoOdbC-SHohaoSDmxn:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L453)):*

I think we should throw an exception if the project isn't a resource project.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 473 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLkzxOC4nSUP-hOa7fx:-MLoPxNT9G6zN7-L30Ap:b-sm4610) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L473)):*

I don't think you need to check if `siteId` is null or empty. You can further simplify this by using `TryGetValue` to both check if the site exists and get it in one call.
</blockquote></details>

I've cleaned up this method a bit.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 601 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLl14O28CqWvQoDe0BJ:-MLoVtEe0B6GTSP5CqXy:bne718t) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L601)):*

I assume that we want to ensure that only one project should be created for a particular resource. Using an if statement like this could cause a race condition. Maybe we could use the Paratext Id as the project Id, since that is what uniquely identifies the resource. I believe that ShareDB will throw if a doc with that Id already exists.
</blockquote></details>

The risk of a race condition is about the same as CreateProjectAsync(). CreateResourceProjectInternalAsync() checks just before creation, so I think the risk is low (but by no means impossible).

It is usually not a good idea to have the id for a Mongo document as a non-generated identifier (this is why we have the problem of migrating the TextData sources). I also think that since projects have a generated identifier, it would be a bit non-obvious to have resources (which are in the same collection) have the Paratext ID as the identifier.

If you think it is a problem, we should deal with this in a separate issue, and implement a lock that covers both CreateProjectAsync() and CreateResourceProjectAsync(). I'm not too sure on the full scope of the lock (i.e. do we implement this only in C#, what is the creation/propagation latency to account for in the lock, etc), as I do not have in depth knowledge of ShareDB.
___
*[src/SIL.XForge.Scripture/Services/SFTextCorpusFactory.cs, line 52 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLl3GcJ8SN3BvcXgGPE:-MLoUVRF78sihyR76bGt:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/SFTextCorpusFactory.cs#L52)):*

We should throw if we cannot get the source project Id.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/Services/SyncService.cs, line 58 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLl459OA_xsfSsnqmS_:-MLoSxxvAFgFVjhMSG9E:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/SyncService.cs#L58)):*

Please specify the type unless it is explicit in the assignment.
</blockquote></details>

Done.

ddaspit (commented):
> Reviewed 40 of 40 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)* status: all files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @pmachapman)

___
*[src/RealtimeServer/common/models/site.ts, line 3 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLl9z5m8RjJWQZj4jD8:-MLpmjgVE_HKvePykE4_:b1hzz6g) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/RealtimeServer/common/models/site.ts#L3)):*

When we add items to the projects collection, they appear in the selection dropdown in the top left corner of the screen - we probably don't want resources appearing there?

It also helps neatly divide resources from projects from a system perspective (the projects collection has a reverse mapping in projects.userRoles, which resources does not have).
</blockquote></details>

@irahopkinson What do you think? Should we treat resources just like any other project and display it in the project dropdown? I feel like that might be what we decided on, but I can't remember.
___
*[src/SIL.XForge.Scripture/Services/LazyScrTextCollection.cs, line 17 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLky4t3DzphGz2qMRbY:-MLpiRce7KsC9k5J7KZ-:bitziop) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/LazyScrTextCollection.cs#L17)):*

During my testing, I noticed that FindById was hammering IO as it was run many times during sync. As some resources use an older SSF format instead of settings.xml, changes I made doubled the IO for those resources on sync.

I tried to mitigate the thread issues with the cache, but yeah, the more I think about it the potential for random hard to reproduce issues looms large.

I've removed it. If IO is an issue during QA, we can look at another method.
</blockquote></details>

Sounds like a good plan. If necessary, we could probably use an LRU or time-based cache. We would need to make sure that it is thread-safe.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 601 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLl14O28CqWvQoDe0BJ:-MLpkc0-9-5BkksH95bC:b-ltf494) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L601)):*

The risk of a race condition is about the same as CreateProjectAsync(). CreateResourceProjectInternalAsync() checks just before creation, so I think the risk is low (but by no means impossible).

It is usually not a good idea to have the id for a Mongo document as a non-generated identifier (this is why we have the problem of migrating the TextData sources). I also think that since projects have a generated identifier, it would be a bit non-obvious to have resources (which are in the same collection) have the Paratext ID as the identifier.

If you think it is a problem, we should deal with this in a separate issue, and implement a lock that covers both CreateProjectAsync() and CreateResourceProjectAsync(). I'm not too sure on the full scope of the lock (i.e. do we implement this only in C#, what is the creation/propagation latency to account for in the lock, etc), as I do not have in depth knowledge of ShareDB.
</blockquote></details>

You are correct. This is an issue for normal projects as well. We really only want one SF project for each PT project/resource. It could probably be fixed with a lock. We can defer this to another issue.

irahopkinson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit and @pmachapman)

___
*[src/RealtimeServer/common/models/site.ts, line 3 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLl9z5m8RjJWQZj4jD8:-MLvJcxuAhrm7o-sbPEB:behk5sm) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/RealtimeServer/common/models/site.ts#L3)):*

@irahopkinson What do you think? Should we treat resources just like any other project and display it in the project dropdown? I feel like that might be what we decided on, but I can't remember.
</blockquote></details>

Yes, we did decide we want connected resources showing in the project dropdown. However they don't have to be in the same collection for that to happen. And it is probably a good idea to list them separately in that dropdown with a separator between them, projects first, resources second.

ddaspit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)* status: all files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @pmachapman)

___
*[src/RealtimeServer/common/models/site.ts, line 3 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLl9z5m8RjJWQZj4jD8:-MM04ktG8vUHjZpv1P4z:bf4wb5c) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/RealtimeServer/common/models/site.ts#L3)):*

Yes, we did decide we want connected resources showing in the project dropdown. However they don't have to be in the same collection for that to happen. And it is probably a good idea to list them separately in that dropdown with a separator between them, projects first, resources second.
</blockquote></details>

Ira and I discussed this. We felt that it would be simpler to just have a single `projects` collection. @pmachapman Can you try refactoring to a single collection?

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)* status: 59 of 74 files reviewed, 2 unresolved discussions (waiting on @ddaspit and @pmachapman)

___
*[src/RealtimeServer/common/models/site.ts, line 3 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLl9z5m8RjJWQZj4jD8:-MMDqVfB5EewIWzSgQMS:by8j55t) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/RealtimeServer/common/models/site.ts#L3)):*

Ira and I discussed this. We felt that it would be simpler to just have a single `projects` collection. @pmachapman Can you try refactoring to a single collection?
</blockquote></details>

No problem - I have made this change, and updated the branch.

irahopkinson (commented):
> Reviewed 42 of 75 files at r1, 1 of 1 files at r2, 40 of 40 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)* status: 59 of 74 files reviewed, 5 unresolved discussions (waiting on @ddaspit, @irahopkinson, and @pmachapman)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 850 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MMEjWVg5jtRHmyzSQYt:-MMEjWVg5jtRHmyzSQYu:b2loox2) ([raw file](https://github.com/sillsdev/web-xforge/blob/5233da46ce79eeb0b674db62e58e9b919335ee90/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L850)):*
> ```
>  it('user has no resource access', fakeAsync(() => {
> ```
I don't think this needs to be tested twice, does it?
___
*[src/SIL.XForge.Scripture/Services/ISFRESTClient.cs, line 7 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MMEotJTEe0uHcs4tQhY:-MMEotJTEe0uHcs4tQhZ:bpbp7vi) ([raw file](https://github.com/sillsdev/web-xforge/blob/8d47b1dcceec43aa13332f51d525471f117ad342/src/SIL.XForge.Scripture/Services/ISFRESTClient.cs#L7)):*
> ```
> IRESTClient
> ```
nit: we honour captializing of 1 and 2 letter acronyms (e.g SF) but longer ones get Capital case (e.g. Dbl, Rest). This file name and elsewhere in this branch. I think this follows an MS style convention.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 603 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MMEtEX02_fzY1adRvAk:-MMEtEX13shXZyKY5yff:bsua8ip) ([raw file](https://github.com/sillsdev/web-xforge/blob/5233da46ce79eeb0b674db62e58e9b919335ee90/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L603)):*
> ```
> ScriuptureForge
> ```
nit: typo in "Scripture" and "Scripture Forge"NN is 2 words.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)* status: 59 of 74 files reviewed, 5 unresolved discussions (waiting on @ddaspit, @irahopkinson, and @pmachapman)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 850 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MMEjWVg5jtRHmyzSQYt:-MMF1GbD383WIRz1Orvt:blgy4g) ([raw file](https://github.com/sillsdev/web-xforge/blob/5233da46ce79eeb0b674db62e58e9b919335ee90/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L850)):*

> ```
>  it('user has no resource access', fakeAsync(() => {
> ```
I don't think this needs to be tested twice, does it?
</blockquote></details>

There two groups of tests, one with translationSuggestionsEnabled set to true, the other false. I noticed all of the other tests were duplicated in these two groups, so I duplicated mine (matching the mutation). This flag seems to rely on the source project/resource in code that uses it in editor.component.ts, so I have kept it, unless you disagree?
___
*[src/SIL.XForge.Scripture/Services/ISFRESTClient.cs, line 7 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MMEotJTEe0uHcs4tQhY:-MMEzARO7dTdwJJLEBT4:b9tx95i) ([raw file](https://github.com/sillsdev/web-xforge/blob/8d47b1dcceec43aa13332f51d525471f117ad342/src/SIL.XForge.Scripture/Services/ISFRESTClient.cs#L7)):*

> ```
> IRESTClient
> ```
nit: we honour captializing of 1 and 2 letter acronyms (e.g SF) but longer ones get Capital case (e.g. Dbl, Rest). This file name and elsewhere in this branch. I think this follows an MS style convention.
</blockquote></details>

No problem - I only kept the over-capitalized casing to match Paratext. I've updated the casing for DBL and REST - it looks nicer now :-)
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 603 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MMEtEX02_fzY1adRvAk:-MMF0gnS0i0MQ8cJmV9Z:b9fy35v) ([raw file](https://github.com/sillsdev/web-xforge/blob/5233da46ce79eeb0b674db62e58e9b919335ee90/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L603)):*

> ```
> ScriuptureForge
> ```
nit: typo in "Scripture" and "Scripture Forge"NN is 2 words.
</blockquote></details>

Noted. Have updated 1 or 2 other places I had it as one word as well.

irahopkinson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)* status: 51 of 76 files reviewed, 2 unresolved discussions (waiting on @ddaspit, @irahopkinson, and @pmachapman)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 850 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MMEjWVg5jtRHmyzSQYt:-MMF8wxiAYQCdokt-lJI:b3tbmhl) ([raw file](https://github.com/sillsdev/web-xforge/blob/5233da46ce79eeb0b674db62e58e9b919335ee90/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L850)):*

There two groups of tests, one with translationSuggestionsEnabled set to true, the other false. I noticed all of the other tests were duplicated in these two groups, so I duplicated mine (matching the mutation). This flag seems to rely on the source project/resource in code that uses it in editor.component.ts, so I have kept it, unless you disagree?
</blockquote></details>

Yes you are right, I should have looked closer.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)* status: 51 of 76 files reviewed, 2 unresolved discussions (waiting on @ddaspit and @irahopkinson)

___
*[src/SIL.XForge.Scripture/Services/JwtRestClient.cs, line 30 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLkvWKh3vqbXoiZYapQ:-MMJExuIE5jrgTbrxg4F:bkmenlc) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/JwtRestClient.cs#L30)):*

@irahopkinson @marksvc  Do we have a plan to access this method without using reflection?
</blockquote></details>

Sorry I missed this mention. The answer was no :). But we had a discussion about it today and we are going to proceed by modifying ParatextData RESTClient.cs to have another public method that exposes the functionality we are looking for, make a PT PR and start a conversation with Tim to see what we can do.

ddaspit (approved):
:lgtm:

> Reviewed 12 of 24 files at r4, 14 of 14 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867#-:-MMJW4WY7cLJ-iqVDPPT:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[src/SIL.XForge.Scripture/Services/JwtRestClient.cs, line 30 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MLkvWKh3vqbXoiZYapQ:-MMJVG337wMr37EpqcgG:bfe3fcr) ([raw file](https://github.com/sillsdev/web-xforge/blob/69df4b69f62294ab95ca287da5ffc2ba405a43e8/src/SIL.XForge.Scripture/Services/JwtRestClient.cs#L30)):*

Sorry I missed this mention. The answer was no :). But we had a discussion about it today and we are going to proceed by modifying ParatextData RESTClient.cs to have another public method that exposes the functionality we are looking for, make a PT PR and start a conversation with Tim to see what we can do. 
</blockquote></details>

Sound good.

ddaspit (commented):
> Reviewed 3 of 79 files at r6, 3 of 5 files at r7, 12 of 19 files at r8, 1 of 2 files at r9, 1 of 1 files at r10, 6 of 6 files at r11, 2 of 2 files at r12, 77 of 77 files at r13.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

ddaspit (commented):
> Reviewed 13 of 97 files at r14, 2 of 3 files at r15, 78 of 78 files at r16.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

irahopkinson:
___
*[src/SIL.XForge.Scripture/Services/SFInstallableDblResource.cs, line 220 at r16](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MNaw3j7EkDNO__2Ga22:-MNaw3j7EkDNO__2Ga23:bttn4s6) ([raw file](https://github.com/sillsdev/web-xforge/blob/6f1838b73229762bce73564a8369f249425145a8/src/SIL.XForge.Scripture/Services/SFInstallableDblResource.cs#L220)):*
> ```
> InternetAccess.ParatextDBLServer
> ```
Do we know what this value will be on development, staging and production?

irahopkinson:
___
*[src/SIL.XForge.Scripture/Services/SFInstallableDblResource.cs, line 220 at r16](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MNaw3j7EkDNO__2Ga22:-MNgajJo2bSpnbA14w6e:bl4gv1) ([raw file](https://github.com/sillsdev/web-xforge/blob/6f1838b73229762bce73564a8369f249425145a8/src/SIL.XForge.Scripture/Services/SFInstallableDblResource.cs#L220)):*

> ```
> InternetAccess.ParatextDBLServer
> ```
Do we know what this value will be on development, staging and production?
</blockquote></details>

For it to work with the appropriate PT Registry JWT, our (SF) Development should point to QA/development DBL since PT Registry is registry-dev and our (SF) Staging and Production should point to production DBL since PT Registry is production for those 2.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)* status: all files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/Services/SFInstallableDblResource.cs, line 220 at r16](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MNaw3j7EkDNO__2Ga22:-MNgjo165gagFCxGRfZT:biz401o) ([raw file](https://github.com/sillsdev/web-xforge/blob/6f1838b73229762bce73564a8369f249425145a8/src/SIL.XForge.Scripture/Services/SFInstallableDblResource.cs#L220)):*

For it to work with the appropriate PT Registry JWT, our (SF) Development should point to QA/development DBL since PT Registry is registry-dev and our (SF) Staging and Production should point to production DBL since PT Registry is production for those 2.
</blockquote></details>

`InternetAccess.ParatextDBLServer` will return the production DBL server URL by default. The production URL is `https://paratext.thedigitalbiblelibrary.org/` and the development URL is `https://paratext-qa.thedigitalbiblelibrary.org/`. You can follow the pattern for the other PT servers. The base URLs are set in the constructor of `ParatextService`.

irahopkinson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)* status: all files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @pmachapman)

___
*[src/Migrations/SourceTargetSplitting/Startup.cs, line 88 at r16](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MNgq4gG11rPeLq6r1JF:-MNgq4gG11rPeLq6r1JG:bq97bda) ([raw file](https://github.com/sillsdev/web-xforge/blob/6f1838b73229762bce73564a8369f249425145a8/src/Migrations/SourceTargetSplitting/Startup.cs#L88)):*
> ```
> services.AddSFRealtimeServer(LoggerFactory, Configuration, IsDevelopment);
> ```
Can we add a flag here (or somewhere similar) so that if the migrator is in test mode it won't run the RealtimeServer migrations?

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867)* status: 77 of 95 files reviewed, 2 unresolved discussions (waiting on @ddaspit and @irahopkinson)

___
*[src/Migrations/SourceTargetSplitting/Startup.cs, line 88 at r16](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MNgq4gG11rPeLq6r1JF:-MNuEHmC8YrhXMSHRFai:b-y8jd3u) ([raw file](https://github.com/sillsdev/web-xforge/blob/6f1838b73229762bce73564a8369f249425145a8/src/Migrations/SourceTargetSplitting/Startup.cs#L88)):*

> ```
> services.AddSFRealtimeServer(LoggerFactory, Configuration, IsDevelopment);
> ```
Can we add a flag here (or somewhere similar) so that if the migrator is in test mode it won't run the RealtimeServer migrations?
</blockquote></details>

I have added a flag, which is passed through to the real time server, updated the migration program to use it, and have set up a test for the real time server to ensure that it works.
___
*[src/SIL.XForge.Scripture/Services/SFInstallableDblResource.cs, line 220 at r16](https://reviewable.io/reviews/sillsdev/web-xforge/867#-MNaw3j7EkDNO__2Ga22:-MNtbOP35NJ8ONQ1RRvS:b-mmg0wy) ([raw file](https://github.com/sillsdev/web-xforge/blob/6f1838b73229762bce73564a8369f249425145a8/src/SIL.XForge.Scripture/Services/SFInstallableDblResource.cs#L220)):*

`InternetAccess.ParatextDBLServer` will return the production DBL server URL by default. The production URL is `https://paratext.thedigitalbiblelibrary.org/` and the development URL is `https://paratext-qa.thedigitalbiblelibrary.org/`. You can follow the pattern for the other PT servers. The base URLs are set in the constructor of `ParatextService`.
</blockquote></details>

I have updated the DBL server access to use a base URL specified in the Paratext service (similar to the registry URL)

irahopkinson (approved):
:lgtm: 

> Reviewed 18 of 18 files at r17.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/867#-:-MNwRJ_X6h8Qf_us9DCB:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #870, opened 2020-11-11
Log authentication errors to console

Nateowami:
When we show the login error dialog (which I get a _lot_ locally) it's not possible to tell what the actual error was unless it's logged. Currently most errors get logged right before being reported to Bugsnag. However, since we don't handle these errors the same, they don't get logged. This change gives developers the option of opening the console to see what the actual error is.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
@@ -11,6 +11,7 @@ import { filter, mergeMap } from 'rxjs/operators';
 import { PwaService } from 'xforge-common/pwa.service';
 import { environment } from '../environments/environment';
 import { CommandService } from './command.service';
+import { ErrorReportingService } from './error-reporting.service';
 import { LocalSettingsService } from './local-settings.service';
 import { LocationService } from './location.service';
 import { NoticeService } from './notice.service';
@@ -72,7 +73,8 @@ export class AuthService {
     private readonly router: Router,
     private readonly localSettings: LocalSettingsService,
     private readonly pwaService: PwaService,
-    private readonly noticeService: NoticeService
+    private readonly noticeService: NoticeService,
+    private readonly reportingService: ErrorReportingService
   ) {
     // Listen for changes to the auth state. If the user logs out in another tab/window, redirect to the home page.
     // When localStorage is cleared event.key is null. The logic below may be more specific than necessary, but we can't
@@ -195,8 +197,8 @@ export class AuthService {
         return { loggedIn: false, newlyLoggedIn: false };
       }
       return { loggedIn: true, newlyLoggedIn: false };
-    } catch {
-      await this.showLoginErrorDialog();
+    } catch (error) {
+      await this.handleLoginError('tryLogIn', error);
       return { loggedIn: false, newlyLoggedIn: false };
     }
   }
@@ -219,8 +221,8 @@ export class AuthService {
         }
       }
       return { loggedIn: true, newlyLoggedIn: false };
-    } catch {
-      await this.showLoginErrorDialog();
+    } catch (error) {
+      await this.handleLoginError('tryOnlineLogIn', error);
       return { loggedIn: false, newlyLoggedIn: false };
     }
   }
@@ -289,7 +291,9 @@ export class AuthService {
     });
   }
 
-  private async showLoginErrorDialog(): Promise<void> {
+  private async handleLoginError(method: string, error: object): Promise<void> {
+    console.error(error);
+    this.reportingService.silentError(`Error occurred in ${method}`, error);
     await this.noticeService.showMessageDialog(
       () => translate('error_messages.error_occurred_login'),
       () => translate('error_messages.try_again')


nigel-wells (commented):
Do you think we should also have these go through to Bugsnag - might give some insight in to how widespread this issue actually is. I've also seen an `invalid_token` `Nonce does not match.` as well.  It would also be good to know the context of when it failed i.e. was it from within `tryLogIn` or from `tryOnlineLogIn`. Does this mean the `retryUponTimeout` is not enough of a bandaid - is it even `checkSession` that is failing?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870#-:-MLtlY-vEStLAy-oqeIt:bcd8bl0)* status: 0 of 1 files reviewed, all discussions resolved

Nateowami (commented):
That seems like a good idea, but I'm having trouble figuring out how to implement it. If I import ErrorHandler I can call `handleError`, but it would show the generic error dialog (which I want to avoid, and just show the custom dialog we have for this scenario). If I add an optional argument to `handleError` to indicate whether to show the dialog, then I have to import ExceptionHandlingService and I get a circular dependency error.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870#-:-MLu-aG7Ex_X77L_zVHb:bvo6cxl)* status: 0 of 1 files reviewed, all discussions resolved

nigel-wells (commented):
Can you just import the `ErrorReportingService` which then has a public `notify` function to send it direct to Bugsnag and bypass our main handler?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870#-:-MLu71393tTlsuxySpe9:b-hdhzaw)* status: 0 of 1 files reviewed, all discussions resolved

Nateowami (commented):
There's a significant amount of pre-processing we do to make sure we have a valid error object to send to Bugsnag (I kinda think the object we get may not be accepted by Bugsnag, but I'm not completely sure. It would depend on whether it's an actual error object). Also the `ExceptionHandlingService` logs automatically, creates an errorId, and includes the locale (it also adds the user object, but that wouldn't be relevant in this case). I'm going to see if I can come up with a way to get rid of some of the dependencies between these files.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870#-:-MLu8MIm7TI05zRHs4mf:b-ur2u3h)* status: 0 of 1 files reviewed, all discussions resolved

Nateowami (commented):
@nigel-wells I plan to pick this up again once your PR goes through. I think it's likely the situation will have changed enough that sending the error becomes much easier.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870#-:-MM_c0q-0WSIs9x3tqz_:b-lb4r6z)* status: 0 of 1 files reviewed, all discussions resolved

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 295 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/870#-MN188tL6O-Whcnk-ULU:-MN188tL6O-Whcnk-ULV:b4cjtbz) ([raw file](https://github.com/sillsdev/web-xforge/blob/15f9f667a42031028d79d5383bc54d58e16dc4e7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L295)):*
> ```XML
> 
>   private async handleLoginError(error: any): Promise<void> {
>     this.reportingService.silentError(error.message);
> ```

I definitely don't like the fact that here I am assuming the error has a property `message`, and only that message (with no stack trace) will be reported.

Actually, I've suddenly remembered I think the object is not an error object, but a regular object with two properties that I can't remember. I'll need to reproduce the issue to check. But theoretically a different error could occur, which could have different properties. Hence it would be advantageous if it were carefully handled as we handle other errors.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 295 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/870#-MN188tL6O-Whcnk-ULU:-MN1BXEe5qkRBXWVt1fo:b-hlwqk2) ([raw file](https://github.com/sillsdev/web-xforge/blob/15f9f667a42031028d79d5383bc54d58e16dc4e7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L295)):*

I definitely don't like the fact that here I am assuming the error has a property `message`, and only that message (with no stack trace) will be reported.

Actually, I've suddenly remembered I think the object is not an error object, but a regular object with two properties that I can't remember. I'll need to reproduce the issue to check. But theoretically a different error could occur, which could have different properties. Hence it would be advantageous if it were carefully handled as we handle other errors.
</blockquote></details>

Assume you want to sit on this a little longer before merging anything? We can pass extra data in to `silentError` to help support the error - gets added to the `custom` tab in Bugsnag. I don't think that would help for a stack trace though. What if we simply added a `stack` property inside `silentError` that got sent along to Bugsnag as part of the error - might be enough to have it carried through? I also see I left `silently` inside the exception handler which doesn't need to be there any more.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 295 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/870#-MN188tL6O-Whcnk-ULU:-MN1H09k08vKJMJbNcJI:bewcfrl) ([raw file](https://github.com/sillsdev/web-xforge/blob/15f9f667a42031028d79d5383bc54d58e16dc4e7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L295)):*

Assume you want to sit on this a little longer before merging anything? We can pass extra data in to `silentError` to help support the error - gets added to the `custom` tab in Bugsnag. I don't think that would help for a stack trace though. What if we simply added a `stack` property inside `silentError` that got sent along to Bugsnag as part of the error - might be enough to have it carried through? I also see I left `silently` inside the exception handler which doesn't need to be there any more.
</blockquote></details>

So, I just managed to get the dialog triggered and the actual error logged in the console, carefully took a screenshot, and then for some reason the screenshot that got saved is defective. I don't recall what the actual property name I should have used was, but I did see that it was not `message`.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 295 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/870#-MN188tL6O-Whcnk-ULU:-MN1IGij1J86TP5IbR3I:b-7ekxy8) ([raw file](https://github.com/sillsdev/web-xforge/blob/15f9f667a42031028d79d5383bc54d58e16dc4e7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L295)):*

So, I just managed to get the dialog triggered and the actual error logged in the console, carefully took a screenshot, and then for some reason the screenshot that got saved is defective. I don't recall what the actual property name I should have used was, but I did see that it was not `message`.
</blockquote></details>

Doh! For what its worth I took some screenshots when I was looking in to the timeouts issue on https://jira.sil.org/browse/SF-1055. They are cropped to the actual error from the stack trace so not overly useful for what you're currently trying to figure out.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 295 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/870#-MN188tL6O-Whcnk-ULU:-MN1djjM4tujoIXtKDUt:bimuiyy) ([raw file](https://github.com/sillsdev/web-xforge/blob/15f9f667a42031028d79d5383bc54d58e16dc4e7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L295)):*

Doh! For what its worth I took some screenshots when I was looking in to the timeouts issue on https://jira.sil.org/browse/SF-1055. They are cropped to the actual error from the stack trace so not overly useful for what you're currently trying to figure out.
</blockquote></details>

I managed to trigger a `login_required` error which looks like this: 

Apparently you can also go `Error().stack` to get the current stack trace - perhaps this could be added in to the `silentError()` method?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870)* status: 0 of 1 files reviewed, all discussions resolved

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 295 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/870#-MN188tL6O-Whcnk-ULU:-MNPQdAcA-Be0ydIUJau:b9085xq) ([raw file](https://github.com/sillsdev/web-xforge/blob/15f9f667a42031028d79d5383bc54d58e16dc4e7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L295)):*

I managed to trigger a `login_required` error which looks like this: 

Apparently you can also go `Error().stack` to get the current stack trace - perhaps this could be added in to the `silentError()` method?
</blockquote></details>

I think Bugsnag groups errors in part by looking at the stack trace. If we got the stack trace in the `silentError()` method, it's likely all those errors will be grouped together. It would be ideal if we could avoid that.

nigel-wells (approved):
Just one note if you feel its relevant otherwise happy for it to merge

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870#-:-MNQG_va7SIHYlLlCPQi:b-fsggy3)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 295 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/870#-MNQGgrA9ev9IFAE6n4R:-MNQGgrA9ev9IFAE6n4S:b-cfa2iv) ([raw file](https://github.com/sillsdev/web-xforge/blob/b4f1ed521baca73c71a138381e42b721eae7869d/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L295)):*
> ```XML
> 
>   private async handleLoginError(error: any): Promise<void> {
>     this.reportingService.silentError(error.error_description);
> ```

Would it be useful to know which method triggered it for debugging i.e. was it `tryLogIn()` or `tryOnlineLogIn()`

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 295 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/870#-MNQGgrA9ev9IFAE6n4R:-MNQzYGDBzEWgW8qQVm6:b-qu6n94) ([raw file](https://github.com/sillsdev/web-xforge/blob/b4f1ed521baca73c71a138381e42b721eae7869d/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L295)):*

Would it be useful to know which method triggered it for debugging i.e. was it `tryLogIn()` or `tryOnlineLogIn()`
</blockquote></details>

Good point. Done.

nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 295 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/870#-MNQGgrA9ev9IFAE6n4R:-MNR1EDj1fhDAR7KD6wT:b7j2dmp) ([raw file](https://github.com/sillsdev/web-xforge/blob/b4f1ed521baca73c71a138381e42b721eae7869d/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L295)):*

Good point. Done.
</blockquote></details>

Thanks. I'm being a pain now. I know one of the errors I got the `error` was quite different from the `error_description` - can you include both as I imagine the `error` would be what you specifically need to check with auth0. Perhaps include them as additional metadata in `silentError()`.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 295 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/870#-MNQGgrA9ev9IFAE6n4R:-MNR2piLA56xbDGIs335:b-gk5vqq) ([raw file](https://github.com/sillsdev/web-xforge/blob/b4f1ed521baca73c71a138381e42b721eae7869d/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L295)):*

Thanks. I'm being a pain now. I know one of the errors I got the `error` was quite different from the `error_description` - can you include both as I imagine the `error` would be what you specifically need to check with auth0. Perhaps include them as additional metadata in `silentError()`.


</blockquote></details>

Done, I think.

nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 295 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/870#-MNQGgrA9ev9IFAE6n4R:-MNR3xEWCVEQzKbKzFlQ:bkw739s) ([raw file](https://github.com/sillsdev/web-xforge/blob/b4f1ed521baca73c71a138381e42b721eae7869d/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L295)):*

Done, I think.
</blockquote></details>

How about:
```
private async handleLoginError(method: string, error: object): Promise<void> {
    this.reportingService.silentError(`Error occurred in {$method}`, error);
    ...
}
```
That way we capture the entire error in to Bugsnag. You could probably have a more specific error type if the reject from auth0 gives you one? I've not tried this code either but I'm pretty sure that will add it as metadata to the `custom` tab.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 295 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/870#-MNQGgrA9ev9IFAE6n4R:-MNRJ6WB5nodZvaCozjD:b-lt4cwd) ([raw file](https://github.com/sillsdev/web-xforge/blob/b4f1ed521baca73c71a138381e42b721eae7869d/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L295)):*
Good idea. Done, I think.

> You could probably have a more specific error type if the reject from auth0 gives you one?

Unfortunately, neither TypeScript nor I am amble to infer anything about the error type (TypeScript marks it as any),

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 295 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/870#-MNQGgrA9ev9IFAE6n4R:-MNU0lJd7soSMfMuCdA6:b44jhuc) ([raw file](https://github.com/sillsdev/web-xforge/blob/b4f1ed521baca73c71a138381e42b721eae7869d/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L295)):*

Good idea. Done, I think.

> You could probably have a more specific error type if the reject from auth0 gives you one?

Unfortunately, neither TypeScript nor I am amble to infer anything about the error type (TypeScript marks it as any),
</blockquote></details>

Actually, I don't think I like this, because the actual error object will not get logged, thus failing to satisfy the reason for the change in the first place:
``` ts
silentError(message: string, metadata?: object) {
  if (metadata != null) {
    this.addMeta(metadata);
  }
  this.notify({ name: 'Silent Error', message: message });
  console.error(message);
}
```
Secondly, I'm not sure adding the error object as metadata is really the right place to put it.

I think fundamentally the problem is something like:
1. Bugsnag expects a NotifiableError, which has the type definition:
``` ts
export type NotifiableError = Error
| { errorClass: string, errorMessage: string }
| { name: string, message: string }
| string
```
2. We want to report an arbitrary object.
3. There's no assurance that the object actually has the necessary properties.

In the exception handler we deal with this by doing a lot of checking and manipulation of the object (I wrote it with the intention that literally any JavaScript value would be handled properly, whether it's null/undefined, a function, a symbol, object with circular references, etc.). Perhaps we should reuse that logic to normalize whatever gets sent to `silentError()`? I'm not sure it will be trivial to do, but I don't see a better option at the moment.

nigel-wells (commented):
Interesting observation on this. If I've logged from auth0 then I get the dialog. I think in that instance I shouldn't get an error dialog as it was deliberate and a simple redirect to auth0 would be fine. My scenario was being logged in as the same user on localhost and beta.localhost. I logged out from beta.localhost which then brought up the dialog on localhost.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870#-:-MOZ82w15xXPzzX19KZu:b3nvptp)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

irahopkinson (approved):
:lgtm:

> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/870#-:-MQ_yxGv6RfdpaQHmen7:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #875, opened 2020-11-18
SF-1149 Include end verse when importing questions from Transcelerator

Nateowami:
The previous implementation mistakenly assumed that when a question is within a single chapter, endChapter will be defined and equal to startChapter. In reality it is undefined in this case, but with this fix no assumption is made as to whether it will be undefined or equal to startChapter.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/875)


RaymondLuong3 (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/875#-:-MMS0o3E4Fqv6mWVS7js:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #877, opened 2020-11-19
SF-1125 - Improve Bugsnag logging

nigel-wells:
- Initiate Bugsnag inside app.module.ts
- Added breadcrumb feedback for JSON RPC posts
- Set user data when the user is logged in
- Set basic user id when first processing through the auth service in case an errors throws shortly after
- Created new `AppError` to allow for throwing errors with passing additional feedback to Bugsnag
- Throw an error if a document requested was invalid

The biggest change in this is separating out Bugsnag from the error handler so that it starts with the app itself. This allows for a more complete breadcrumb as it is recording data from the start and not just when an error is triggered.

I also created a new `AppError` class to extend the main `Error`. This class lets you pass a key|string object which will then be passed along with the error in to Bugsnag. I've added one example here (dealing with documents that doesn't exist) and this can be expanded if we find other useful parts of the app where we want to log more information.

I audited all the JSON RPC POST requests we make and there is nothing of concern in here as the backend does all the heavy lifting. I decided to add this to the breadcrumb so we know what the frontend is asking the backend to do which is useful if the backend ever comes back with an error.

Huge dramas if Bugsnag ends up being initiated inside of tests - it essentially breaks a heap of tests following whenever it gets started. Ended up putting `Bugsnag.start()` inside of `app.module.ts` to get around this.

To test anything locally and have it flow through to Bugsnag you need to add `dev` into the `enabledReleaseStages` array in `app.module.ts`.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877)


RaymondLuong3 (changes requested):
Looking good. Just 1 blocking comment.

> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877#-:-MMWWWXWAAO1KnKwQXS7:b-aypwns)* status: all files reviewed, 6 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 46 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWO3Vn80UuL4Gflt3G:-MMWO3Vn80UuL4Gflt3H:ba84vqy) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L46)):*
> ```
> type UserForReport = User & { id: string };
> ```
Could you add a comment for this? I've never seen this before.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 529 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWT9yn4mNPlq4RPLNH:-MMWT9yn4mNPlq4RPLNI:b-s0cwym) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L529)):*
> ```
> (within three seconds of the service being constructed), it will
> ```
This comment is not accurate anymore.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 532 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWOyBR7_bduBEVubMW:-MMWOyBR7_bduBEVubMX:b-9m7fkp) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L532)):*
<details><summary><i alt="summary">Quoted 5 lines of code…</i></summary>

> ```
>   private async getUserForReporting(): Promise<UserForReport | undefined> {
>     const currentUser = await this.userService.getCurrentUser();
>     if (currentUser != null) {
>       return cloneDeep(currentUser.data) as UserForReport;
>     }
> ```
</details>

Did you mean to use this method in OnInit()? If not, it can be removed.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 364 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWQENZAG_tKWuTshTm:-MMWQENZAG_tKWuTshTn:bn008yb) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L364)):*
> ```
> ...(userId !== prevUserId && { prevUserId: prevUserId }),
> ```
Just checking, the double `&&` does not mean logical AND in this case?
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts, line 57 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWQd8r9wVSYQFW0lTK:-MMWQd8r9wVSYQFW0lTL:bbri96d) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L57)):*

> ```
>     Bugsnag.leaveBreadcrumb(
>       'JSON-RPC',
>       {
>         request: url,
>         method: method,
>         params: params
>       },
>       'request'
>     );
> ```
</details>

Nice! I look forward to seeing this put in use.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts, line 36 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWSRwK9SUrjv46Af3T:-MMWSRwK9SUrjv46Af3U:bsqhrua) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts#L36)):*
> ```
>   Bugsnag.notify(error, event => ErrorReportingService.beforeSend(options, event), callback)
> ```
It's nice to be able to reference `Bugsnag` in a global context.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877)* status: 12 of 13 files reviewed, 5 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 46 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWO3Vn80UuL4Gflt3G:-MMX-p-Y5qe8LKpHDcpv:b-l6f7qy) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L46)):*

> ```
> type UserForReport = User & { id: string };
> ```
Could you add a comment for this? I've never seen this before.
</blockquote></details>

This is an intersection type - https://www.probeauty.co.nz/shop/skincare/?filter_skin-concerns=sensitive-skin - it combines one type with another or, in this case, creates a new type with an additional property.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 529 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWT9yn4mNPlq4RPLNH:-MMX08mi8_OxX0UfAGYD:bgkpuki) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L529)):*

> ```
> (within three seconds of the service being constructed), it will
> ```
This comment is not accurate anymore.
</blockquote></details>

Ah - this whole section I completely refactored out and can now be deleted.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 532 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWOyBR7_bduBEVubMW:-MMX0DLl19t_x7P1QTIh:bvih7sc) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L532)):*

> ```
>   private async getUserForReporting(): Promise<UserForReport | undefined> {
>     const currentUser = await this.userService.getCurrentUser();
>     if (currentUser != null) {
>       return cloneDeep(currentUser.data) as UserForReport;
>     }
> ```
Did you mean to use this method in OnInit()? If not, it can be removed.
</blockquote></details>

As above - refectored and no longer needed. Have removed
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 364 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWQENZAG_tKWuTshTm:-MMX4sUcDQfhfYGZqZ1u:bed4m9m) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L364)):*

> ```
> ...(userId !== prevUserId && { prevUserId: prevUserId }),
> ```
Just checking, the double `&&` does not mean logical AND in this case?
</blockquote></details>

I believe it does mean AND in the normal sense. The spread operator `...` will return whatever is supplied inside. In this case we're checking a condition. If the condition is `false` then the `&&` is ignored. If the condition is `true` then the the following is returned and added to the object. I've never used it until now - there is a good readup here to help explain it more: https://medium.com/@slamflipstrom/conditional-object-properties-using-spread-in-javascript-714e0a12f496
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts, line 36 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWSRwK9SUrjv46Af3T:-MMX0QoyAI1uOD9FIgtz:bgenrz3) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts#L36)):*

> ```
>   Bugsnag.notify(error, event => ErrorReportingService.beforeSend(options, event), callback)
> ```
It's nice to be able to reference `Bugsnag` in a global context.
</blockquote></details>

Yes it is. Though I am wondering if I should be wrapping it with out own functions just in case their API changes at some point in the future?

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877)* status: 12 of 13 files reviewed, 5 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 46 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWO3Vn80UuL4Gflt3G:-MMYmaBY77HYFz9_3YFZ:b-yku35a) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L46)):*

This is an intersection type - https://www.probeauty.co.nz/shop/skincare/?filter_skin-concerns=sensitive-skin - it combines one type with another or, in this case, creates a new type with an additional property.
</blockquote></details>

Hmmm, try this URL :O https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#intersection-types

Nateowami (changes requested):
Nice work. Just a few comments.

> Reviewed 12 of 13 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877#-:-MM_aAAV8fJWQZZ_VXmz:bmpy934)* status: all files reviewed, 11 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 46 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWO3Vn80UuL4Gflt3G:-MM_JJmfDq26q0JNa7BE:b-wb69b6) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L46)):*

Hmmm, try this URL :O https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#intersection-types
</blockquote></details>

I'm not sure I understand the purpose of preserving this type. The only place I see it used is lower in the file:
``` ts
const userData = cloneDeep(this.currentUserDoc.data) as UserForReport;
if (userData != null) {
  Bugsnag.addMetadata('user', userData);
}
```
I think we could just as easily drop `as UserForReport` and `userData` would then have type `Readonly<User> | undefined`.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 265 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MM_Nbj4DmrnVpyOTbGa:-MM_Nbj5BtD99CHdXmi5:bj3doyg) ([raw file](https://github.com/sillsdev/web-xforge/blob/a8525384f5c6475980f8cc08f75d4740fd4e8e09/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L265)):*
> ```XML
>       const isNewlyLoggedIn = await this.authService.isNewlyLoggedIn;
>       if (isNewlyLoggedIn && !supportedBrowser()) {
>         this.dialog.open(SupportedBrowsersDialogComponent, { autoFocus: false });
> ```

Perhaps consider setting a piece of metadata stating that the browser is detected to be unsupported (regardless of `isNewlyLoggedIn`)? I know we should be able to determine it from the data Bugsnag gives us, but that requires manually checking what the rules we have set are.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts, line 271 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MM_VjJZ28jwl1I1kjw0:-MM_VjJ_8ymIldXNZhy9:b6dletr) ([raw file](https://github.com/sillsdev/web-xforge/blob/a8525384f5c6475980f8cc08f75d4740fd4e8e09/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts#L271)):*
> ```XML
> 
>   it('user data is set for Bugsnag', fakeAsync(() => {
>     // Ideally we'd spy on Bugsnag itself but if Bugsnag is initiated it breaks a lot of tests with outstanding timers
> ```

Is it not mockable either?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.module.ts, line 33 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MM_OUNwCrwZBwZb5Zqt:-MM_OUNwCrwZBwZb5Zqu:bf02qls) ([raw file](https://github.com/sillsdev/web-xforge/blob/a8525384f5c6475980f8cc08f75d4740fd4e8e09/src/SIL.XForge.Scripture/ClientApp/src/app/app.module.ts#L33)):*
> ```XML
> import { UsersModule } from './users/users.module';
> 
> const config: BrowserConfig = {
> ```

Does this actually need to be in this file to get it to this point in the loading process? Could it not be imported and invoked (e.g. `ExceptionHandlingService.initBugsnag()`, which could be a static method? For that matter, I wonder whether it should be invoked from `main.ts`. `app.module.ts` is more of a place for declaring dependencies, and is loaded by `main.ts`.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 364 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWQENZAG_tKWuTshTm:-MM_FxEZ4TUlJqgZ1-vz:bnfde0f) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L364)):*

I believe it does mean AND in the normal sense. The spread operator `...` will return whatever is supplied inside. In this case we're checking a condition. If the condition is `false` then the `&&` is ignored. If the condition is `true` then the the following is returned and added to the object. I've never used it until now - there is a good readup here to help explain it more: https://medium.com/@slamflipstrom/conditional-object-properties-using-spread-in-javascript-714e0a12f496
</blockquote></details>

Well, it took me a bit, but I actually figured out what this line does without reading the docs. Do I win? As confusing as this is to begin with, it might as well be shortened to:
``` ts
{
  userId,
  ...(userId !== prevUserId && { prevUserId }),
  params: claims[XF_ROLE_CLAIM]
}
```

While it could be argued this is too clever, needing to conditionally add a property is a common enough use case that we should probably all just get used to reading this. In the past I've always had to just add the property after the fact, which is clear if you read a method through, but also isn't the cleanest way to do it, and would be more likely to be missed than this more declarative method.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts, line 57 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MM_QIDb4VkiBAwYuKkt:-MM_QIDcB7MpfFsIJN5W:b-sodzg5) ([raw file](https://github.com/sillsdev/web-xforge/blob/a8525384f5c6475980f8cc08f75d4740fd4e8e09/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L57)):*
> ```XML
> 
>   async onlineInvoke<T>(url: string, method: string, params: any = {}): Promise<T | undefined> {
>     Bugsnag.leaveBreadcrumb(
> ```

Nice. I wonder if adding the id may be helpful in matching it with server logs or responses? If you move this down a couple lines it can be
``` ts
Bugsnag.leaveBreadcrumb('JSON-RPC', { request: url, method, params, id: request.id }, 'request');
```
(Prettier is happy to let it be one line).
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts, line 20 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MM_Rai91xmwD9bIsoXu:-MM_Rai91xmwD9bIsoXv:b-x3gz7t) ([raw file](https://github.com/sillsdev/web-xforge/blob/a8525384f5c6475980f8cc08f75d4740fd4e8e09/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts#L20)):*
> ```XML
>       if (breadcrumb.type === 'navigation' && breadcrumb.metadata && typeof breadcrumb.metadata.from === 'string') {
>         breadcrumb.metadata.from = ErrorReportingService.redactAccessToken(breadcrumb.metadata.from);
>         breadcrumb.metadata.to = ErrorReportingService.redactAccessToken(breadcrumb.metadata.to);
> ```

Thanks for catching that.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts, line 14 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MM_W_aZ9tnjCvaGGk4q:-MM_W_aZ9tnjCvaGGk4r:bsfsadw) ([raw file](https://github.com/sillsdev/web-xforge/blob/cb99ca30797d2fcf0a4ab477287754cb00f9d310/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L14)):*
> ```XML
> import { NoticeService } from './notice.service';
> import { UserService } from './user.service';
> import { objectId, promiseTimeout } from './utils';
> ```

This was the only use of `promiseTimeout`, which can now be removed from utils. It's such a shame though; this was the perfect example use case for the `never` type. I fear we shall never use it again.

``` ts
export function promiseTimeout<T>(promise: Promise<T>, timeout: number) {
  return Promise.race([
    promise,
    new Promise<never>((_resolve, reject) => {
      setTimeout(reject, timeout);
    })
  ]);
}
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts, line 52 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MM_Ta81C8CwgLUDrRpT:-MM_Ta81C8CwgLUDrRpU:b-8wdma2) ([raw file](https://github.com/sillsdev/web-xforge/blob/a8525384f5c6475980f8cc08f75d4740fd4e8e09/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts#L52)):*
> ```XML
>         throw new AppError('Document could not be created.', {
>           collection: collection,
>           id: id != null ? id : 'undefined'
> ```

This could be simplified to `throw new AppError('Document could not be created.', { collection, id: id ?? 'undefined' });`

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877)* status: 3 of 14 files reviewed, 9 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 46 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWO3Vn80UuL4Gflt3G:-MMmQiQO4D4Hk8ZYjxkL:bulm6eb) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L46)):*

I'm not sure I understand the purpose of preserving this type. The only place I see it used is lower in the file:
``` ts
const userData = cloneDeep(this.currentUserDoc.data) as UserForReport;
if (userData != null) {
  Bugsnag.addMetadata('user', userData);
}
```
I think we could just as easily drop `as UserForReport` and `userData` would then have type `Readonly<User> | undefined`.
</blockquote></details>

Good point - we're not adding the `id` any more so there is no need for it.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 265 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MM_Nbj4DmrnVpyOTbGa:-MMmRdzi-8Bc1uD3dbTQ:bhrslx6) ([raw file](https://github.com/sillsdev/web-xforge/blob/a8525384f5c6475980f8cc08f75d4740fd4e8e09/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L265)):*

Perhaps consider setting a piece of metadata stating that the browser is detected to be unsupported (regardless of `isNewlyLoggedIn`)? I know we should be able to determine it from the data Bugsnag gives us, but that requires manually checking what the rules we have set are.
</blockquote></details>

Have added. I also created a new function to easily add meta data which will also make it easier to test on.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts, line 271 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MM_VjJZ28jwl1I1kjw0:-MMmqKGQ-SrF2G7f0TG4:ba0tayf) ([raw file](https://github.com/sillsdev/web-xforge/blob/a8525384f5c6475980f8cc08f75d4740fd4e8e09/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts#L271)):*

Is it not mockable either?
</blockquote></details>

Not for me. I tried spying on it which works but can check if anything has been called unless `Bugsnag.start()` has been run and then is when everything falls apart for any tests that follow. I tried a few more times now and still didn't have any success. With the new `addMeta` function I've added it means we can at least test for what I wanted to know.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.module.ts, line 33 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MM_OUNwCrwZBwZb5Zqt:-MMm_KIf-69eDCy7MwxA:bo2dt60) ([raw file](https://github.com/sillsdev/web-xforge/blob/a8525384f5c6475980f8cc08f75d4740fd4e8e09/src/SIL.XForge.Scripture/ClientApp/src/app/app.module.ts#L33)):*

Does this actually need to be in this file to get it to this point in the loading process? Could it not be imported and invoked (e.g. `ExceptionHandlingService.initBugsnag()`, which could be a static method? For that matter, I wonder whether it should be invoked from `main.ts`. `app.module.ts` is more of a place for declaring dependencies, and is loaded by `main.ts`.
</blockquote></details>

Much cleaner idea - done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts, line 57 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MM_QIDb4VkiBAwYuKkt:-MMmAnYX8haPdOLmJSd4:b-ovwaw7) ([raw file](https://github.com/sillsdev/web-xforge/blob/a8525384f5c6475980f8cc08f75d4740fd4e8e09/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L57)):*

Nice. I wonder if adding the id may be helpful in matching it with server logs or responses? If you move this down a couple lines it can be
``` ts
Bugsnag.leaveBreadcrumb('JSON-RPC', { request: url, method, params, id: request.id }, 'request');
```
(Prettier is happy to let it be one line).
</blockquote></details>

Good idea.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts, line 52 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MM_Ta81C8CwgLUDrRpT:-MMmAVEEEtJC_BOabhMs:b-83sjoh) ([raw file](https://github.com/sillsdev/web-xforge/blob/a8525384f5c6475980f8cc08f75d4740fd4e8e09/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts#L52)):*

This could be simplified to `throw new AppError('Document could not be created.', { collection, id: id ?? 'undefined' });`

See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing
</blockquote></details>

Nice!

Nateowami (changes requested):
> Reviewed 12 of 12 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877)* status: all files reviewed, 4 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 46 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWO3Vn80UuL4Gflt3G:-MMpG0Pb15Hfkj06qh8N:b-s4h3sg) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L46)):*

Good point - we're not adding the `id` any more so there is no need for it.
</blockquote></details>

@raymondluong3 It's like we do `Blah | null`, which means "satisfies the Blah type, or satisfies the null type". Except instead of or it's and. So it would have to satisfy the User type and additionally satisfy the `{ id: string }` type. Which is like saying "the user type, but with an id as well."
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 265 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MM_Nbj4DmrnVpyOTbGa:-MMpFmQwBqb3HSxtPVet:b-ucfkn4) ([raw file](https://github.com/sillsdev/web-xforge/blob/a8525384f5c6475980f8cc08f75d4740fd4e8e09/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L265)):*

Have added. I also created a new function to easily add meta data which will also make it easier to test on.
</blockquote></details>

Thanks.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 261 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMpEVFuDqIKKL_kUjm_:-MMpEVFuDqIKKL_kUjma:b-46486o) ([raw file](https://github.com/sillsdev/web-xforge/blob/e4d845946309e34ea636222ec54ffc8c6ef72fae/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L261)):*
> ```XML
>       if (languageTag != null) {
>         this.i18n.trySetLocale(languageTag, false);
>         this.reportingService.addMeta({ localId: this.i18n ? this.i18n.localeCode : 'unknown' });
> ```

Unfortunately, this is not the only way for a locale to be set. Also, there's no need to check whether `this.i18n` is truthy; that was only a concern in the exception handling service, because it fetched the service via an injector, and it has to be extremely careful not to throw exceptions (because if our exception handler has errors, they won't get reported and could hide other errors).

I would put this in `I18nService.trySeLocale()`. My guess is that in that context the existence of the locale will have been definitely established. It would probably also be good to log breadcrumbs (or report) the failure to set a locale.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts, line 24 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMpIlLy1YBA_QN2a4Z8:-MMpIlLy1YBA_QN2a4Z9:bohe750) ([raw file](https://github.com/sillsdev/web-xforge/blob/e4d845946309e34ea636222ec54ffc8c6ef72fae/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts#L24)):*
> ```XML
>     });
> 
>     for (const tab in metaData) {
> ```

This looks to me as though it will add tabs with numbers, rather than names. Is that the intention? Also, is there a reason this couldn't use `for..of` rather than `for...in` with `hasOwnProperty()`?

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877)* status: all files reviewed, 2 unresolved discussions (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 46 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWO3Vn80UuL4Gflt3G:-MMpx5TDDcqjrgs4sZhO:bo9m3i4) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L46)):*

@raymondluong3 It's like we do `Blah | null`, which means "satisfies the Blah type, or satisfies the null type". Except instead of or it's and. So it would have to satisfy the User type and additionally satisfy the `{ id: string }` type. Which is like saying "the user type, but with an id as well."
</blockquote></details>

Ah, I see now.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 364 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWQENZAG_tKWuTshTm:-MMpxPhe7nJbbbutcg7F:bwj7abw) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L364)):*

Well, it took me a bit, but I actually figured out what this line does without reading the docs. Do I win? As confusing as this is to begin with, it might as well be shortened to:
``` ts
{
  userId,
  ...(userId !== prevUserId && { prevUserId }),
  params: claims[XF_ROLE_CLAIM]
}
```

While it could be argued this is too clever, needing to conditionally add a property is a common enough use case that we should probably all just get used to reading this. In the past I've always had to just add the property after the fact, which is clear if you read a method through, but also isn't the cleanest way to do it, and would be more likely to be missed than this more declarative method.
</blockquote></details>

Neat! Now that I understand what's going on, I agree we should get used to seeing this kind of syntax.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts, line 36 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMWSRwK9SUrjv46Af3T:-MMpxoD63CRw9aB6kA-P:bwmu9jm) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4280f971149ba8de1ac6c81f50e518c4b2cccd4/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts#L36)):*

Yes it is. Though I am wondering if I should be wrapping it with out own functions just in case their API changes at some point in the future?
</blockquote></details>

Hmm, that's a good point. I'm ok with either way, but seems like wrapping this would be the more robust option.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877)* status: 13 of 16 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 261 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMpEVFuDqIKKL_kUjm_:-MMr-aQbCstBbf8Y6oxG:b-bwllza) ([raw file](https://github.com/sillsdev/web-xforge/blob/e4d845946309e34ea636222ec54ffc8c6ef72fae/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L261)):*

Unfortunately, this is not the only way for a locale to be set. Also, there's no need to check whether `this.i18n` is truthy; that was only a concern in the exception handling service, because it fetched the service via an injector, and it has to be extremely careful not to throw exceptions (because if our exception handler has errors, they won't get reported and could hide other errors).

I would put this in `I18nService.trySeLocale()`. My guess is that in that context the existence of the locale will have been definitely established. It would probably also be good to log breadcrumbs (or report) the failure to set a locale.
</blockquote></details>

Have moved. I also refactored `trySetLocale()` as mentioned on Slack - let me know if you want this reversed.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts, line 24 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMpIlLy1YBA_QN2a4Z8:-MMqjm_f8tl0o8B-A1rT:b-by3hpl) ([raw file](https://github.com/sillsdev/web-xforge/blob/e4d845946309e34ea636222ec54ffc8c6ef72fae/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts#L24)):*

This looks to me as though it will add tabs with numbers, rather than names. Is that the intention? Also, is there a reason this couldn't use `for..of` rather than `for...in` with `hasOwnProperty()`?
</blockquote></details>

I used `for...in` because I wanted the value of index to be used as the tab name. You'll see in the `addMeta()` function that the index is actually a string and is used as the tab name in Bugsnag. Because we're using `for...in` we then need to add the `hasOwnProperty()` to make tslint happy.

Nateowami (changes requested):
> Reviewed 3 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877)* status: all files reviewed, 3 unresolved discussions (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts, line 24 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMpIlLy1YBA_QN2a4Z8:-MMrC7Za-0zPP-qcJDsa:b-q8x1db) ([raw file](https://github.com/sillsdev/web-xforge/blob/e4d845946309e34ea636222ec54ffc8c6ef72fae/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts#L24)):*

I used `for...in` because I wanted the value of index to be used as the tab name. You'll see in the `addMeta()` function that the index is actually a string and is used as the tab name in Bugsnag. Because we're using `for...in` we then need to add the `hasOwnProperty()` to make tslint happy.
</blockquote></details>

Here's my thinking:
1. `metaData` is an array, therefore when we do `for..in`, `tab` will be an index starting at 0.
2. `event.addMetadata(tab, metaData[tab])` will be doing something like `event.addMetadata(0, metaData[0])`.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts, line 5 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMrCvolAz5iB4BB6cei:-MMrCvolAz5iB4BB6cej:b-hrs7un) ([raw file](https://github.com/sillsdev/web-xforge/blob/211c13ce916c5104d68307e498f6d4036f14dc3a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts#L5)):*
> ```XML
> 
> export interface EventMetadata {
>   tab: object;
> ```

Is this supposed to be `[key:string]: object;`?
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts, line 114 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMr8tHY-wOz6sm6OJei:-MMr8tHY-wOz6sm6OJej:btu0ip2) ([raw file](https://github.com/sillsdev/web-xforge/blob/211c13ce916c5104d68307e498f6d4036f14dc3a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L114)):*
> ```XML
>     const locale = I18nService.getLocale(tag);
>     if (locale == null) {
>       throw new Error(`Cannot set locale to non-existent locale ${tag}`);
> ```

There are three places from which a call to `setLocale()` can originate:
- The cookie from the home page (and other pages served by the back end)
- The Auth0 interface language value
- The language menu

The language menu only lists locales that we know about, so it should never cause a failure. However, it's perfectly plausible that Auth0 may have the language set to something we don't support. Right now we're planning to do away with the language selector, but in the future we're planning to add the ability to edit the user profile on languagetechnology.org. So we can't be completely certain this is one of the values our front end knows about. Furthermore, each system seems to use a different language tag for each language, so there's the possibility a supported language is specified, but with a language tag that isn't understood by the front end.

All of this is a little improbable if we do our jobs properly, but it's likely to happen from time to time, and I thought it better to not crash the current task/thread when the locale can't be found. Reporting it to Bugsnag would be a good idea, but I'm reluctant to throw an exception, as it could potentially break the startup process.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877)* status: 11 of 17 files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts, line 24 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMpIlLy1YBA_QN2a4Z8:-MMrM7GW1YHR0zwcceOe:b-kxz8v6) ([raw file](https://github.com/sillsdev/web-xforge/blob/e4d845946309e34ea636222ec54ffc8c6ef72fae/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts#L24)):*

Here's my thinking:
1. `metaData` is an array, therefore when we do `for..in`, `tab` will be an index starting at 0.
2. `event.addMetadata(tab, metaData[tab])` will be doing something like `event.addMetadata(0, metaData[0])`.

</blockquote></details>

Interesting that the lint didn't complain about me incorrectly adding `[]` instead of `{}` and everything still worked fine! I've fixed this up so it reads correctly now - still does the same thing though.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts, line 5 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMrCvolAz5iB4BB6cei:-MMrNAP2C7rKh70d0BAE:b-qhk6xe) ([raw file](https://github.com/sillsdev/web-xforge/blob/211c13ce916c5104d68307e498f6d4036f14dc3a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts#L5)):*

Is this supposed to be `[key:string]: object;`?
</blockquote></details>

Have tided up.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts, line 114 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMr8tHY-wOz6sm6OJei:-MMrUVwiC9a325Fosem6:brdqst1) ([raw file](https://github.com/sillsdev/web-xforge/blob/211c13ce916c5104d68307e498f6d4036f14dc3a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L114)):*

There are three places from which a call to `setLocale()` can originate:
- The cookie from the home page (and other pages served by the back end)
- The Auth0 interface language value
- The language menu

The language menu only lists locales that we know about, so it should never cause a failure. However, it's perfectly plausible that Auth0 may have the language set to something we don't support. Right now we're planning to do away with the language selector, but in the future we're planning to add the ability to edit the user profile on languagetechnology.org. So we can't be completely certain this is one of the values our front end knows about. Furthermore, each system seems to use a different language tag for each language, so there's the possibility a supported language is specified, but with a language tag that isn't understood by the front end.

All of this is a little improbable if we do our jobs properly, but it's likely to happen from time to time, and I thought it better to not crash the current task/thread when the locale can't be found. Reporting it to Bugsnag would be a good idea, but I'm reluctant to throw an exception, as it could potentially break the startup process.
</blockquote></details>

Fair enough. I've reverted and done a first attempt at a `silentError` method. Initially I had it in `exception-handling-service.ts` but I received a heap of `NullInjectorError`s across most test files. I also did a quick test (which I didn't include) with using this in `auth.service.ts` for the login error notification and it seemed to work quite well.

RaymondLuong3 (approved):
Things look good on my end.

> Reviewed 7 of 12 files at r4, 1 of 3 files at r5, 6 of 6 files at r6.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877#-:-MMvXbrW9h0AF4BJyXHg:b-8mcgma)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami)

Nateowami (approved):
:lgtm:

> Reviewed 6 of 6 files at r6.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877#-:-MMvpO9NCWemomTiIVsg:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts, line 24 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/877#-MMpIlLy1YBA_QN2a4Z8:-MMv_xAqCG-f1Fwa47jq:b-a481ud) ([raw file](https://github.com/sillsdev/web-xforge/blob/e4d845946309e34ea636222ec54ffc8c6ef72fae/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-reporting.service.ts#L24)):*

Interesting that the lint didn't complain about me incorrectly adding `[]` instead of `{}` and everything still worked fine! I've fixed this up so it reads correctly now - still does the same thing though.
</blockquote></details>

I think TS allows using `object[key]` notation without type checking (but not `object.key`).

I would suggest writing this as:
``` ts
for (const tabName of Object.keys(metaData)) {
  event.addMetadata(tabName, metaData[tabName]);
}
```
The difference between the keys yielded by `for..in` and `Object.keys()` differ only in that `Object.keys()` only returns own enumerable property names.

But that's optional. I'm marking this as approved.

nigel-wells (commented):
@nateowami I just noticed by chance in the terminal that I've accidently created a circular warning between `error-reporting.service.ts` and `exception-handling-service.ts` as the are both wanting to use methods from one another. This has happened with the new `silentError()` method that calls `handleError()`. I know we initialled wanted to run through `handleError()` because of the extra processes that takes place but do you think this isn't really required any more? The silent errors are really just our own errors/messages we want to pass on to Bugsnag so there actually isn't anything to process. Or would you see this being useful to notify Bugsnag in some instances after catching an error that you'd still want to be handled via `handleError()`?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877#-:-MMwxSx10W0f3N74vHCQ:b6ytm2d)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

Nateowami (approved):
I think it probably doesn't need to go through `handleError()`. Looking over it, it appears most of what it does is work out the type of error and what should be done with it (e.g. determine if it's a network error and a snackbar should be shown, or a promise rejection error that should be unwrapped, etc.). The only thing I think would be missing is an eventId, but since the error is not shown to the user or sent anywhere else, I don't think there's anything to cross-reference it against.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877#-:-MN-BEMy8p2S23VFQNxG:buj0y1f)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

nigel-wells (commented):
Thanks - that is what I was thinking as well. I've pushed through an update along with a couple of other minor enhancements.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877#-:-MN0QqtjCMrdqKdg5AZj:b9uso92)* status: 14 of 16 files reviewed, all discussions resolved (waiting on @Nateowami and @RaymondLuong3)

Nateowami (approved):
:lgtm:

> Reviewed 2 of 3 files at r7.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/877#-:-MN0RgOsE44tqOPZYaCU:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #879, opened 2020-11-20
SF-1144 Overlapping text on connect project page in Safari

nigel-wells:
- Removed unneeded flex layout options

This appears to have been a Safari specific issue rather than Chrome on iOS as it also happened on a Mac. It appears Safari can struggle with flex and calculating heights correctly when inside a `flex-direction: column`. It is likely there is a good combination of CSS that can be added if that property is really needed but, this instance, it simply didn't need to be there in the first place.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/879)


Nateowami (approved):
:lgtm_strong: :shipit: Always nice to see bugs fixed or features added by deleting code.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/879#-:-MM_fmM80zwTn0JvGG4u:b-nd8c7p)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #880, opened 2020-11-20
Fix importing of questions on a single verse

Nateowami:
After importing a question from Transcelerator I went to edit the question and found this:

While the mistake was in the original import logic, it was unveiled by the recent fix in #875 (SF-1149 Include end verse when importing questions from Transcelerator). Questions with no end verse are getting saved with a range of e.g. "1-undefined".

This fix basically consists of two changes:
- The setting of `verse` on `verseRefData` is now conditional. Previously it would sometimes get set to `undefined`, but that's not the same as not setting it.
- Ensure the end verse actually exists before setting the verse range. Previously it just checked that it differs from the start verse, which is not sufficient.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/880)


irahopkinson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/880#-:-MMp3Wpv3FHRwKeLCCxA:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #886, opened 2020-11-30
Improve Bugsnag Breadcrumbs

nigel-wells:
- Added new static method handleBreadcrumb

This initial commit is targeted at buttons but it can be expanded to other elements as needed.

Aim of this is to resolve common struggles with Bugsnag trying to identify the label of buttons but not getting anything useful.

**BEFORE**
image

**AFTER**
image

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/886)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.spec.ts
@@ -144,6 +144,11 @@ describe('ExceptionHandlingService', () => {
           selector: 'DIV.mdc-button__ripple',
           expectedText: 'Ripple text',
           expectedSelector: 'DIV.mdc-button__ripple'
+        },
+        {
+          selector: 'BUTTON.mdc-button.child-element > i',
+          expectedText: 'Child',
+          expectedSelector: 'BUTTON.mdc-button.child-element span'
         }
       ];
       for (const test of tests) {
@@ -166,6 +171,7 @@ interface BreadcrumbTests {
   template: `
     <button class="mdc-button plain-text">Plain text</button>
     <button class="mdc-button include-icon"><i>icon_name</i><span>Inside span</span></button>
+    <button class="mdc-button child-element"><i>icon_name</i><span>Child</span></button>
     <button
       id="activated_button"
       class="mdc-button mdc-ripple-upgraded mdc-ripple-upgraded--background-focused mdc-ripple-upgraded--foreground-activation"
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.spec.ts
@@ -1,6 +1,8 @@
 import { MdcDialog, MdcDialogRef } from '@angular-mdc/web/dialog';
-import { TestBed } from '@angular/core/testing';
+import { Component, ElementRef, ViewChild } from '@angular/core';
+import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { NotifiableError } from '@bugsnag/js';
+import { Breadcrumb } from '@bugsnag/js';
 import { CookieService } from 'ngx-cookie-service';
 import { User } from 'realtime-server/lib/common/models/user';
 import { Observable } from 'rxjs';
@@ -115,10 +117,72 @@ describe('ExceptionHandlingService', () => {
     verify(mockedNoticeService.showError(anything())).once();
     expect().nothing();
   });
+
+  describe('Bugsnag', () => {
+    it('should extract text from button', async () => {
+      const env = new TestEnvironment();
+      const tests: BreadcrumbTests[] = [
+        {
+          selector: 'BUTTON.mdc-button.plain-text',
+          expectedText: 'Plain text',
+          expectedSelector: 'BUTTON.mdc-button.plain-text'
+        },
+        {
+          selector: 'BUTTON.mdc-button.include-icon',
+          expectedText: 'Inside span',
+          expectedSelector: 'BUTTON.mdc-button.include-icon span'
+        },
+        {
+          selector:
+            'BUTTON#activated_button.mdc-ripple-upgraded.mdc-ripple-upgraded--background-focused.mdc-ripple-upgraded--foreground-activation',
+          expectedText: 'Button with ID',
+          expectedSelector: 'BUTTON#activated_button span'
+        },
+        {
+          selector: 'DIV.mdc-button__ripple',
+          expectedText: 'Ripple text',
+          expectedSelector: 'DIV.mdc-button__ripple'
+        }
+      ];
+      for (const test of tests) {
+        const breadcrumb = env.addBreadcrumb(test.selector);
+        expect(breadcrumb.metadata.targetText).toBe(test.expectedText);
+        expect(breadcrumb.metadata.targetSelector).toBe(test.expectedSelector);
+      }
+    });
+  });
 });
+
+interface BreadcrumbTests {
+  selector: string;
+  expectedText: string;
+  expectedSelector: string;
+}
+
+@Component({
+  selector: 'app-host',
+  template: `
+    <button class="mdc-button plain-text">Plain text</button>
+    <button class="mdc-button include-icon"><i>icon_name</i><span>Inside span</span></button>
+    <button
+      id="activated_button"
+      class="mdc-button mdc-ripple-upgraded mdc-ripple-upgraded--background-focused mdc-ripple-upgraded--foreground-activation"
+    >
+      <span>Button with ID</span>
+    </button>
+    <button id="ripple_button">
+      <div class="mdc-button__ripple"></div>
+      <span>Ripple text</span>
+    </button>
+  `
+})
+class HostComponent {
+  @ViewChild('container', { static: true }) container!: ElementRef;
+}
 
 class TestEnvironment {
   readonly errorReports: { error: any }[] = [];
+  readonly fixture: ComponentFixture<HostComponent>;
   readonly service: ExceptionHandlingService;
   rejectUser = false;
   timeoutUser = false;
@@ -134,7 +198,12 @@ class TestEnvironment {
   } as UserDoc;
 
   constructor() {
-    this.service = TestBed.inject(ExceptionHandlingService);
+    TestBed.configureTestingModule({
+      declarations: [HostComponent]
+    });
+    this.service = TestBed.get(ExceptionHandlingService);
+    this.fixture = TestBed.createComponent(HostComponent);
+    this.fixture.detectChanges();
 
     when(mockedMdcDialog.open(anything(), anything())).thenReturn({
       afterClosed: () => {
@@ -154,5 +223,19 @@ class TestEnvironment {
   get oneAndOnlyReport() {
     expect(this.errorReports.length).toEqual(1);
     return this.errorReports[this.errorReports.length - 1];
+  }
+
+  addBreadcrumb(targetSelector: string): Breadcrumb {
+    const breadcrumb: Breadcrumb = {
+      message: 'UI click',
+      timestamp: new Date(),
+      type: 'log',
+      metadata: {
+        targetSelector,
+        targetText: '(...)'
+      }
+    };
+    ExceptionHandlingService.handleBreadcrumb(breadcrumb);
+    return breadcrumb;
   }
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.spec.ts
@@ -1,6 +1,6 @@
 import { MdcDialog, MdcDialogRef } from '@angular-mdc/web/dialog';
-import { Component, ElementRef, ViewChild } from '@angular/core';
-import { ComponentFixture, TestBed } from '@angular/core/testing';
+import { Component } from '@angular/core';
+import { ComponentFixture, fakeAsync, flush, TestBed } from '@angular/core/testing';
 import { NotifiableError } from '@bugsnag/js';
 import { Breadcrumb } from '@bugsnag/js';
 import { CookieService } from 'ngx-cookie-service';
@@ -43,6 +43,7 @@ class MockConsole {
 
 describe('ExceptionHandlingService', () => {
   configureTestingModule(() => ({
+    declarations: [HostComponent],
     providers: [
       ExceptionHandlingService,
       { provide: AuthService, useMock: mockedAuthService },
@@ -56,7 +57,7 @@ describe('ExceptionHandlingService', () => {
     imports: [TestTranslocoModule]
   }));
 
-  it('should not crash on anything', async () => {
+  it('should not crash on anything', fakeAsync(() => {
     const env = new TestEnvironment();
     const values = [
       undefined,
@@ -75,13 +76,13 @@ describe('ExceptionHandlingService', () => {
       () => {},
       BigInt(3)
     ];
-    await Promise.all(values.map(value => env.service.handleError(value)));
+    Promise.all(values.map(value => env.handleError(value)));
     expect(env.errorReports.length).toBe(values.length);
-  });
+  }));
 
-  it('should unwrap a rejection from a promise', async () => {
+  it('should unwrap a rejection from a promise', fakeAsync(() => {
     const env = new TestEnvironment();
-    await env.service.handleError({
+    env.handleError({
       message: 'This is the outer message',
       stack: 'Stack trace trace to promise implementation',
       rejection: {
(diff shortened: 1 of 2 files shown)


marksvc (changes requested):
That's great. Thanks for your work on this. I have some quibbles and suggestions. I won't be around to follow up, so when you think you've done enough you can have my approval to merge it :) 

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/886#-:-MOqxY5y0rJyU4dMboEo:b-8mqcds)* status: all files reviewed, 5 unresolved discussions (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts, line 58 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MOqqGjK5cMYCQv2H2Wx:-MOqqGjK5cMYCQv2H2Wy:b-meocg7) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eddfaddf1ba6b0495796806dc610c4d748b2273/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L58)):*
> ```
>     let targetSelector = breadcrumb.metadata.targetSelector;
>     let targetText = breadcrumb.metadata.targetText;
> ```
I don't remember if our style requires us to do this or not, but if you declare the types here it will make the code more self-explaining. (i.e. `let targetSelector: Something = breadcrumb....`).
And the same for other `let` and `const` items that follow in the method. So I'll suggest that.

It's looking like these are both strings.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts, line 64 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MOqpDvb9okw28tU2bpW:-MOqpDvb9okw28tU2bpX:bq65c9j) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eddfaddf1ba6b0495796806dc610c4d748b2273/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L64)):*
> ```
> [0]
> ```
I would have thought that if targetSelector doesn't start with any selectors[].element, that [0] here might crash, but in a browser console it looks like `[][0]` produces undefined. Hmm, so okay :)
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts, line 68 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MOqqjv3EkijNGftt99q:-MOqqjv3EkijNGftt99r:b-i5h37u) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eddfaddf1ba6b0495796806dc610c4d748b2273/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L68)):*
> ```XML
>       return;
>     }
>     let query;
> ```

This can be improved by a type declaration.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts, line 84 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MOqr21gCM4beFlhb_7O:-MOqr21gCM4beFlhb_7P:btlivbz) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eddfaddf1ba6b0495796806dc610c4d748b2273/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L84)):*
> ```
>         if (!targetSelector.includes(cls)) {
>           continue;
>         }
> ```
Because `targetSelector.replace(match, replacementText)` will do nothing if `match` is not found in `targetSelector`, I think you can omit this check that continues and doesn't call .replace() in the event that the class is not in the text. Unless I'm not thinking about this correctly.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts, line 89 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MOqrba4-HiYbyqJhWAr:-MOqrba4-HiYbyqJhWAs:brdw4lt) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eddfaddf1ba6b0495796806dc610c4d748b2273/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L89)):*
> ```XML
>       // Remove CSS specific selectors that Bugsnag added as they aren't compatible with the querySelector
>       if (targetSelector.includes('>')) {
>         targetSelector = targetSelector.substr(0, targetSelector.indexOf('>'));
> ```

I am uncertain what this would look like, or why it removes from the beginning  until just before '>'. Perhaps an example of one of these could be added to the list of tests.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts, line 105 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MOqsQAw35QbxTTPgTI5:-MOqsQAw35QbxTTPgTI6:b-f6frgl) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eddfaddf1ba6b0495796806dc610c4d748b2273/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L105)):*
<details><summary><i alt="summary">Quoted 7 lines of code…</i></summary>

> ```
>     for (const node of query.childNodes) {
>       // Only want text nodes or the specific node element
>       if (node.nodeType !== 3 && node.nodeName.toLowerCase() !== specificElement) {
>         continue;
>       }
>       targetText = node.textContent.trim();
>     }
> ```
</details>


That's interesting, so we loop over the children and set `targetText` each time we have a text node or a node with a matching name. 
Have you considered instead doing something like

const possibleTexts = query.childNodes.where((node: Node) => node.nodeType === 3 || node.nodeName.toLowerCase() === specificElement)
if (possibleTexts.length > 0)
{
  targetText = possibleTexts[0].textContent.trim()
}

In other words, rather than loop over the list and set `targetText` every time there is a match, perhaps just ask the list for _the thing_ you want, and then use that (or a specific occurrence of it) for setting `targetText`. As it seems odd for the code to be written such that multiple hits will all take turns setting `targetText`. (Even if that's not how the code will work in practice.)
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.spec.ts, line 126 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MOquRhA9ThfWjQzAi3g:-MOquRhA9ThfWjQzAi3h:ba8xql3) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eddfaddf1ba6b0495796806dc610c4d748b2273/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.spec.ts#L126)):*
> ```XML
>     it('should extract text from button', fakeAsync(() => {
>       const env = new TestEnvironment();
>       const tests: BreadcrumbTests[] = [
> ```

I like how you wrote out this unit test.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/886)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts, line 68 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MOqqjv3EkijNGftt99q:-MP1bMrs5HUwamlTjQAu:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eddfaddf1ba6b0495796806dc610c4d748b2273/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L68)):*

This can be improved by a type declaration.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts, line 84 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MOqr21gCM4beFlhb_7O:-MP1b_ce-7zX91nNXn_r:bv2vqx1) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eddfaddf1ba6b0495796806dc610c4d748b2273/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L84)):*

> ```
>         if (!targetSelector.includes(cls)) {
>           continue;
>         }
> ```
Because `targetSelector.replace(match, replacementText)` will do nothing if `match` is not found in `targetSelector`, I think you can omit this check that continues and doesn't call .replace() in the event that the class is not in the text. Unless I'm not thinking about this correctly.
</blockquote></details>

Good point - have removed.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts, line 89 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MOqrba4-HiYbyqJhWAr:-MP1c1F-19xVKx8saCvJ:benhap0) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eddfaddf1ba6b0495796806dc610c4d748b2273/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L89)):*

I am uncertain what this would look like, or why it removes from the beginning  until just before '>'. Perhaps an example of one of these could be added to the list of tests.
</blockquote></details>

`substr` goes from `x` to `y`. This line will keep all the text from the start of the string until the first ">" is found and then ignore the rest. In CSS you can be specific with the `>` to state that the first next child element must be whatever follows it rather than just matching a child/grandchild/great grandchild element somewhere in the DOM. I've added a test.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts, line 105 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MOqsQAw35QbxTTPgTI5:-MP1ahku9dy1LPUk-IjF:bdcf5f0) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eddfaddf1ba6b0495796806dc610c4d748b2273/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L105)):*

> ```
>     for (const node of query.childNodes) {
>       // Only want text nodes or the specific node element
>       if (node.nodeType !== 3 && node.nodeName.toLowerCase() !== specificElement) {
>         continue;
>       }
>       targetText = node.textContent.trim();
>     }
> ```

That's interesting, so we loop over the children and set `targetText` each time we have a text node or a node with a matching name. 
Have you considered instead doing something like

const possibleTexts = query.childNodes.where((node: Node) => node.nodeType === 3 || node.nodeName.toLowerCase() === specificElement)
if (possibleTexts.length > 0)
{
  targetText = possibleTexts[0].textContent.trim()
}

In other words, rather than loop over the list and set `targetText` every time there is a match, perhaps just ask the list for _the thing_ you want, and then use that (or a specific occurrence of it) for setting `targetText`. As it seems odd for the code to be written such that multiple hits will all take turns setting `targetText`. (Even if that's not how the code will work in practice.)



</blockquote></details>

Fair point. I've checked what's available and `childNodes` is a type `NodeListOf` so I can't make use of `where`. I see it does has its own `forEach` method implemented which is a better approach than what I originally done so I've switched to that.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.spec.ts, line 126 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MOquRhA9ThfWjQzAi3g:-MP1bEAx0le9EdimEz2T:b-ucfknh) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eddfaddf1ba6b0495796806dc610c4d748b2273/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.spec.ts#L126)):*

I like how you wrote out this unit test.
</blockquote></details>

Thanks!

marksvc (changes requested):
> Reviewed 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/886)* status: all files reviewed, 3 unresolved discussions (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts, line 68 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MOqqjv3EkijNGftt99q:-MQJPz--9mcY5kn6p4ek:btqsnf8) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eddfaddf1ba6b0495796806dc610c4d748b2273/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L68)):*

Done.
</blockquote></details>

Great. So I notice that some non-null assertion operators have come in with this. That makes me nervous, and exception-handling-service.ts is an unfortunate place to have an exception. Here are a couple changes that we could make to not let `!` into the code. What do you think of these?

First: 
```diff
-    let query: Node | null;
+    let query: Node | null | undefined;
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts, line 105 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MOqsQAw35QbxTTPgTI5:-MQJEZCH3McRpMGZbNky:b-793dys) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eddfaddf1ba6b0495796806dc610c4d748b2273/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L105)):*

Fair point. I've checked what's available and `childNodes` is a type `NodeListOf` so I can't make use of `where`. I see it does has its own `forEach` method implemented which is a better approach than what I originally done so I've switched to that.
</blockquote></details>

So this still has the odd look of setting the targetText each time we have a match, rather than finding the thing we want and using it to set targetText. I had a go at trying to refine it more but after several attempts, couldn't come up with better than

```ts
    const childNodes: Node[] = Array.from(query!.childNodes).filter(
      (node: Node) =>
        (node.nodeType === 3 || node.nodeName.toLowerCase() === specificElement) && node.textContent != null
    );
    if (childNodes.length > 0 && childNodes[0].textContent != null) {
      targetText = childNodes[0].textContent.trim();
    }
```

But then I realized that all we need is a `break` in the original code, to make the logic and intent match! Then the code shows: Look thru these things for a match, set it to the first one, and stop looking.

How about changing:
```diff
-    if (!query!.childNodes.length) {
+    if (query == null) {
       return;
     }
-    query!.childNodes.forEach((node: Node) => {
+    for (const node of Array.from(query.childNodes)) {
       // Only want text nodes or the specific node element
       if ((node.nodeType === 3 || node.nodeName.toLowerCase() === specificElement) && node.textContent != null) {
         targetText = node.textContent.trim();
+        break;
       }
-    });
+    }
```

What do you think about that?  (Or maybe this will reveal that I don't really understand what you're wanting this to do :)

My `query.childNodes` complains that it doesn't provide an iterator. Some googling led me to use Array.from(), seen in the above.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts, line 72 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MQJSB-LCc_HgYhaSU04:-MQJSB-LCc_HgYhaSU05:b-y6bx72) ([raw file](https://github.com/sillsdev/web-xforge/blob/8fedca55f47d8419eb56e5cf8414849f39e2ee97/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L72)):*
> ```XML
>     // Sometimes Bugsnag narrows it down to a nested element so we need to use the parent node
>     if (selector.useParent != null && selector.useParent) {
>       query = document.querySelector(targetSelector)!.parentNode;
> ```

Second:
```diff
-      query = document.querySelector(targetSelector)!.parentNode;
+      query = document.querySelector(targetSelector)?.parentNode;
```

irahopkinson (commented):
@nateowami can you pick this one up and get it across the line?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/886#-:-MRJ68ow7gYKeV1rIFYL:b-hiwm40)* status: all files reviewed, 3 unresolved discussions (waiting on @nigel-wells)

Nateowami (commented):
@irahopkinson Sorry, I thought I had replied to your request in the affirmative. In the mean time I've worked on other things I thought were of greater urgency. I'm picking this up now.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/886#-:-MRQs9Tj1TFjIszsV-ak:b-926yv2)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @marksvc and @nigel-wells)

marksvc (approved):
> Reviewed 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/886)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts, line 68 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/886#-MOqqjv3EkijNGftt99q:-MRzlaWYCDehSCIITNT8:bv328r3) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eddfaddf1ba6b0495796806dc610c4d748b2273/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L68)):*

@marksvc I've made some changes along that line.
</blockquote></details>

That's great; thank you.

PR #896, opened 2020-12-16
SF-798 Remove user docs and project associations

marksvc:
This PR is broken into two commits [for code review. We can squash to one commit for merging.]. The first commit adds some tests and makes some changes to allow for testing. Its commit message is:

SF-798 First add tests and prepare for delete-user implementation

- Add comments.
- Introduce IRealtimeServer to allow mocking.
- Add early-crash argument checking to some methods. This should be
the commit's only behavioural change.
- Expose RemoveUserFromProjectAsync() as internal for unit testing.
- Add tests for existing functionality.
- ProjectServiceTests Sites are expanded to be consistent with
TestProject documents, for these or follow-up tests.

The second commit makes the actual changes for SF-798. Some items to point out are:

- ProjectService RemoveUserAsync() can silently fail if there are
  simultaneous requests. So wrap in a mutex.
- When a sysadmin deletes a user, SF runs UserService
  DeleteAsync(), which calls:
  - ProjectService.RemoveUserFromAllProjectsAsync()
    which calls RemoveUserCoreAsync() for each user doc Sites.Projects
      which calls RemoveUserFromProjectAsync().
  - RealtimeService.DeleteUserAsync(), removing docs from collections
    users, user_secrets, sf_project_user_configs,
    o_users, and o_sf_project_user_configs.
  - And no longer userDoc.DeleteAsync(), which shouldn't be needed
    now.
- RealtimeService DeleteUserAsync() removes docs that are matched by
  '_id' or 'd'. Some of these matches are for the whole field. Some
  matches are just for the ending part of the field (for example a
  sf_project_user_configs, which uses an _id of the format
  project_id:user_id. Further, some collections have corresponding o_
  OT collections, and some do not. How to match the '_id' and 'd', and
  whether there is a corresponding o_ OT collection, is collected in
  an anonymous type that makes use of a Location enum.
  Location.Beginning is included in the enum for completeness.
- The knowledge for what collections to look in, and how to match
  documents to ids, can be defined per-application, and here is done
  in SFRealtimeServiceCollectionExtensions.cs.
- RealtimeService.cs: `foreach.. AddDocConfig(userDataDoc)` may not
  have been necessary, but looked like it could be helpful.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-owner/checking-owner.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-owner/checking-owner.component.spec.ts
@@ -5,7 +5,7 @@ import { TranslocoService } from '@ngneat/transloco';
 import { AvatarService } from 'ngx-avatar';
 import { CookieService } from 'ngx-cookie-service';
 import { UserProfile } from 'realtime-server/lib/common/models/user';
-import { instance, mock, when } from 'ts-mockito';
+import { anything, instance, mock, when } from 'ts-mockito';
 import { AuthService } from 'xforge-common/auth.service';
 import { AvatarTestingModule } from 'xforge-common/avatar/avatar-testing.module';
 import { UserProfileDoc } from 'xforge-common/models/user-profile-doc';
@@ -29,6 +29,15 @@ describe('CheckingOwnerComponent', () => {
     tick();
     env.fixture.detectChanges();
     expect(env.userName).toBe('User 01');
+  }));
+
+  it('displays Unknown owner name', fakeAsync(() => {
+    // A user may be removed from the database, and so an ownerRef may refer to a user we can't find.
+    const template = '<app-checking-owner ownerRef="no-longer-known-user-id"></app-checking-owner>';
+    const env = new TestEnvironment(template);
+    tick();
+    env.fixture.detectChanges();
+    expect(env.userName).toBe('checking.unknown_author');
   }));
 
   it('displays avatar', () => {
@@ -77,7 +86,7 @@ class TestEnvironment {
   readonly mockedAuthService = mock(AuthService);
   readonly mockedAvatarService = mock(AvatarService);
   readonly mockedCookieService = mock(CookieService);
-  readonly mockedTransloco = mock(TranslocoService);
+  readonly mockedTranslocoService = mock(TranslocoService);
   readonly mockedUserService = mock(UserService);
 
   private readonly realtimeService: TestRealtimeService;
@@ -90,7 +99,7 @@ class TestEnvironment {
         { provide: AuthService, useFactory: () => instance(this.mockedAuthService) },
         { provide: AvatarService, useFactory: () => instance(this.mockedAvatarService) },
         { provide: CookieService, useFactory: () => instance(this.mockedCookieService) },
-        { provide: TranslocoService, useFactory: () => instance(this.mockedTransloco) },
+        { provide: TranslocoService, useFactory: () => instance(this.mockedTranslocoService) },
         { provide: UserService, useFactory: () => instance(this.mockedUserService) }
       ]
     });
@@ -106,6 +115,9 @@ class TestEnvironment {
     });
     when(this.mockedUserService.getProfile('user01')).thenCall(() =>
       this.realtimeService.subscribe(UserProfileDoc.COLLECTION, 'user01')
+    );
+    when(this.mockedTranslocoService.translate<string>(anything())).thenCall(
+      (translationStringKey: string) => translationStringKey
     );
     this.fixture = TestBed.createComponent(HostComponent);
     this.fixture.detectChanges();
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-owner/checking-owner.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-owner/checking-owner.component.ts
@@ -1,5 +1,5 @@
 import { Component, Input, OnInit } from '@angular/core';
-import { translate } from '@ngneat/transloco';
+import { TranslocoService } from '@ngneat/transloco';
 import { UserProfile } from 'realtime-server/lib/common/models/user';
 import { I18nService } from 'xforge-common/i18n.service';
 import { UserProfileDoc } from 'xforge-common/models/user-profile-doc';
@@ -17,7 +17,11 @@ export class CheckingOwnerComponent implements OnInit {
   @Input() layoutStacked: boolean = false;
   private ownerDoc?: UserProfileDoc;
 
-  constructor(private readonly userService: UserService, readonly i18n: I18nService) {}
+  constructor(
+    private readonly userService: UserService,
+    readonly i18n: I18nService,
+    private readonly translocoService: TranslocoService
+  ) {}
 
   get date(): Date {
     return new Date(this.dateTime);
@@ -25,10 +29,10 @@ export class CheckingOwnerComponent implements OnInit {
 
   get name(): string {
     if (this.ownerDoc == null || this.ownerDoc.data == null) {
-      return '';
+      return this.translocoService.translate('checking.unknown_author');
     }
     return this.userService.currentUserId === this.ownerDoc.id
-      ? translate('checking.me')
+      ? this.translocoService.translate('checking.me')
       : this.ownerDoc.data.displayName;
   }
 
@@ -36,9 +40,9 @@ export class CheckingOwnerComponent implements OnInit {
     return this.ownerDoc == null ? undefined : this.ownerDoc.data;
   }
 
-  ngOnInit(): void {
+  async ngOnInit(): Promise<void> {
     if (this.ownerRef != null) {
-      this.userService.getProfile(this.ownerRef).then(u => (this.ownerDoc = u));
+      this.ownerDoc = await this.userService.getProfile(this.ownerRef);
     }
   }
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json
+++ b/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json
@@ -151,6 +151,7 @@
   },
   "checking": {
     "me": "Me",
+    "unknown_author": "Unknown author",
     "next": "Next",
     "previous": "Previous",
     "questions": "Questions ({{ count }})",
--- a/src/SIL.XForge.Scripture/Models/SyncUser.cs
+++ b/src/SIL.XForge.Scripture/Models/SyncUser.cs
@@ -5,6 +5,9 @@ namespace SIL.XForge.Scripture.Models
     /// </summary>
     public class SyncUser
     {
+        /// <summary>
+        /// Unique id of a SyncUser. Not intended to match a user id.
+        /// </summary>
         public string Id { get; set; }
         public string ParatextUsername { get; set; }
     }
--- a/src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs
+++ b/src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs
@@ -1,6 +1,8 @@
+using System.Collections.Generic;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using SIL.XForge.Configuration;
+using SIL.XForge.Models;
 using SIL.XForge.Realtime;
 using SIL.XForge.Scripture.Models;
 
@@ -25,6 +27,10 @@ namespace Microsoft.Extensions.DependencyInjection
                         new DocConfig("sf_project_user_configs", typeof(SFProjectUserConfig)),
                         new DocConfig("texts", typeof(TextData), OTType.RichText),
                         new DocConfig("questions", typeof(Question))
+                    });
+                    o.UserDataDocs.AddRange(new[]
+                    {
+                        new DocConfig("sf_project_user_configs", typeof(SFProjectUserConfig))
                     });
                 }, launchWithDebugging);
             return services;
--- a/src/SIL.XForge.Scripture/Services/SFServiceCollectionExtensions.cs
+++ b/src/SIL.XForge.Scripture/Services/SFServiceCollectionExtensions.cs
@@ -1,6 +1,7 @@
 using Paratext.Data;
 using Paratext.Data.Archiving;
 using SIL.XForge.Scripture.Services;
+using SIL.XForge.Services;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
@@ -18,6 +19,8 @@ namespace Microsoft.Extensions.DependencyInjection
             services.AddTransient<IParatextNotesMapper, ParatextNotesMapper>();
             services.AddSingleton<IGuidService, GuidService>();
             services.AddSingleton<ISFProjectService, SFProjectService>();
+            // When UserService requests an IProjectService, satisfy it with the SFProjectService.
+            services.AddSingleton<IProjectService, SFProjectService>();
             services.AddSingleton<IJwtTokenHelper, JwtTokenHelper>();
             services.AddSingleton<IParatextDataHelper, ParatextDataHelper>();
             services.AddSingleton<IInternetSharedRepositorySourceProvider, InternetSharedRepositorySourceProvider>();
--- a/src/SIL.XForge.Scripture/Services/SFServiceCollectionExtensions.cs
+++ b/src/SIL.XForge.Scripture/Services/SFServiceCollectionExtensions.cs
@@ -19,7 +19,6 @@ namespace Microsoft.Extensions.DependencyInjection
             services.AddTransient<IParatextNotesMapper, ParatextNotesMapper>();
             services.AddSingleton<IGuidService, GuidService>();
             services.AddSingleton<ISFProjectService, SFProjectService>();
-            // When UserService requests an IProjectService, satisfy it with the SFProjectService.
             services.AddSingleton<IProjectService, SFProjectService>();
             services.AddSingleton<IJwtTokenHelper, JwtTokenHelper>();
             services.AddSingleton<IParatextDataHelper, ParatextDataHelper>();
--- a/src/SIL.XForge/Configuration/DocConfig.cs
+++ b/src/SIL.XForge/Configuration/DocConfig.cs
@@ -17,6 +17,10 @@ namespace SIL.XForge.Configuration
 
         public string CollectionName { get; }
         public Type Type { get; }
+
+        /// <summary>
+        /// Or `null` if there is no corresponding "o_..." collection for this document type.
+        /// </summary>
         public string OTTypeName { get; }
     }
 }
--- a/src/SIL.XForge/Configuration/RealtimeOptions.cs
+++ b/src/SIL.XForge/Configuration/RealtimeOptions.cs
@@ -13,6 +13,18 @@ namespace SIL.XForge.Configuration
         public bool MigrationsDisabled = false;
         public DocConfig UserDoc { get; set; } = new DocConfig("users", typeof(User));
         public DocConfig ProjectDoc { get; set; }
+
+        /// <summary>
+        /// Additional document types (importantly, collection names) that have project related data. Defining this
+        /// helps identify and delete project data when removing a project.
+        /// </summary>
         public List<DocConfig> ProjectDataDocs { get; set; } = new List<DocConfig>();
+
+        /// <summary>
+        /// Document types (importantly, collection names) that have user information, from which to delete records
+        /// when removing a user from the database.
+        /// </summary>
+        public List<DocConfig> UserDataDocs { get; set; } = new List<DocConfig>();
     }
+
 }
--- a/src/SIL.XForge/Realtime/Document.cs
+++ b/src/SIL.XForge/Realtime/Document.cs
@@ -10,11 +10,11 @@ namespace SIL.XForge.Realtime
     /// </summary>
     public class Document<T> : IDocument<T> where T : IIdentifiable
     {
-        private readonly RealtimeServer _server;
+        private readonly IRealtimeServer _server;
         private readonly int _connHandle;
         private readonly SemaphoreSlim _lock = new SemaphoreSlim(1, 1);
 
-        internal Document(RealtimeServer server, int connHandle, string otTypeName, string collection, string id)
+        internal Document(IRealtimeServer server, int connHandle, string otTypeName, string collection, string id)
         {
             _server = server;
             _connHandle = connHandle;
--- a/src/SIL.XForge/Realtime/IRealtimeServer.cs
+++ b/src/SIL.XForge/Realtime/IRealtimeServer.cs
@@ -1,0 +1,17 @@
+using System.Threading.Tasks;
+
+namespace SIL.XForge.Realtime
+{
+    public interface IRealtimeServer
+    {
+        Task<T> ApplyOpAsync<T>(string otTypeName, T data, object op);
+        Task<int> ConnectAsync(string userId = null);
+        Task<Snapshot<T>> CreateDocAsync<T>(int handle, string collection, string id, T data, string otTypeName);
+        Task DeleteDocAsync(int handle, string collection, string id);
+        void Disconnect(int handle);
+        Task<Snapshot<T>> FetchDocAsync<T>(int handle, string collection, string id);
+        void Start(object options);
+        void Stop();
+        Task<Snapshot<T>> SubmitOpAsync<T>(int handle, string collection, string id, object op);
+    }
+}
--- a/src/SIL.XForge/Realtime/IRealtimeService.cs
+++ b/src/SIL.XForge/Realtime/IRealtimeService.cs
@@ -16,5 +16,6 @@ namespace SIL.XForge.Realtime
         IQueryable<T> QuerySnapshots<T>() where T : IIdentifiable;
 
         Task DeleteProjectAsync(string projectId);
+        Task DeleteUserAsync(string userId);
     }
 }
--- a/src/SIL.XForge/Realtime/MemoryRealtimeService.cs
+++ b/src/SIL.XForge/Realtime/MemoryRealtimeService.cs
@@ -41,6 +41,11 @@ namespace SIL.XForge.Realtime
             _docConfigs = new Dictionary<Type, DocConfig>();
         }
 
+        /// <summary>
+        /// Count of calls to DeleteUserAsync(), for tests.
+        /// </summary>
+        internal int CallCountDeleteUserAsync { get; set; } = 0;
+
         public void StartServer()
         {
         }
@@ -56,6 +61,12 @@ namespace SIL.XForge.Realtime
 
         public virtual Task DeleteProjectAsync(string projectId)
         {
+            return Task.CompletedTask;
+        }
+
+        public Task DeleteUserAsync(string userId)
+        {
+            CallCountDeleteUserAsync++;
             return Task.CompletedTask;
         }
 
--- a/src/SIL.XForge/Realtime/RealtimeServer.cs
+++ b/src/SIL.XForge/Realtime/RealtimeServer.cs
@@ -4,7 +4,7 @@ using Jering.Javascript.NodeJS;
 
 namespace SIL.XForge.Realtime
 {
-    public class RealtimeServer
+    public class RealtimeServer : IRealtimeServer
     {
         private readonly INodeJSService _nodeJSService;
         private readonly string _modulePath;
--- a/src/SIL.XForge/Realtime/RealtimeService.cs
+++ b/src/SIL.XForge/Realtime/RealtimeService.cs
@@ -43,7 +43,13 @@ namespace SIL.XForge.Realtime
             AddDocConfig(options.UserDoc);
             AddDocConfig(options.ProjectDoc);
             foreach (DocConfig projectDataDoc in options.ProjectDataDocs)
+            {
                 AddDocConfig(projectDataDoc);
+            }
+            foreach (DocConfig userDataDoc in options.UserDataDocs)
+            {
+                AddDocConfig(userDataDoc);
+            }
         }
 
         internal IRealtimeServer Server { get; }
@@ -107,6 +113,32 @@ namespace SIL.XForge.Realtime
             await opsCollection.DeleteManyAsync(dFilter);
         }
 
+        /// <summary>
+        /// Delete user-related docs from various collections.
+        /// </summary>
+        public async Task DeleteUserAsync(string userId)
+        {
+            if (string.IsNullOrEmpty(userId))
+            {
+                throw new ArgumentException("", nameof(userId));
+            }
+
+            RealtimeOptions options = _realtimeOptions.Value;
+            IEnumerable<DocConfig> collectionsToProcess = options.UserDataDocs.Append(options.UserDoc);
+            FilterDefinition<BsonDocument> idFilter = Builders<BsonDocument>.Filter.Regex("_id", $"{userId}$");
+            FilterDefinition<BsonDocument> dFilter = Builders<BsonDocument>.Filter.Regex("d", $"{userId}$");
+            foreach (var collection in collectionsToProcess)
+            {
+                IMongoCollection<BsonDocument> snapshotCollection = _database.GetCollection<BsonDocument>(
+                    collection.CollectionName);
+                await snapshotCollection.DeleteManyAsync(idFilter);
+
+                IMongoCollection<BsonDocument> opsCollection = _database.GetCollection<BsonDocument>(
+                $"o_{collection.CollectionName}");
+                await opsCollection.DeleteManyAsync(dFilter);
+            }
+        }
+
         public IQueryable<T> QuerySnapshots<T>() where T : IIdentifiable
         {
             string collectionName = GetCollectionName<T>();
@@ -155,7 +187,8 @@ namespace SIL.XForge.Realtime
                 ReleaseStage = this._configuration.GetValue<string>("Bugsnag:ReleaseStage"),
                 MigrationsDisabled = this._realtimeOptions.Value.MigrationsDisabled,
                 SiteId = this._siteOptions.Value.Id,
-                Version = System.Diagnostics.FileVersionInfo.GetVersionInfo(@System.Reflection.Assembly.GetEntryAssembly().Location).ProductVersion
+                Version = System.Diagnostics.FileVersionInfo.GetVersionInfo(
+                    @System.Reflection.Assembly.GetEntryAssembly().Location).ProductVersion
             };
         }
     }
--- a/src/SIL.XForge/Realtime/RealtimeService.cs
+++ b/src/SIL.XForge/Realtime/RealtimeService.cs
@@ -26,7 +26,7 @@ namespace SIL.XForge.Realtime
         private readonly Dictionary<Type, DocConfig> _docConfigs;
         private readonly IConfiguration _configuration;
 
-        public RealtimeService(RealtimeServer server, IOptions<SiteOptions> siteOptions,
+        public RealtimeService(IRealtimeServer server, IOptions<SiteOptions> siteOptions,
             IOptions<DataAccessOptions> dataAccessOptions, IOptions<RealtimeOptions> realtimeOptions,
             IOptions<AuthOptions> authOptions, IMongoClient mongoClient, IConfiguration configuration)
         {
@@ -46,7 +46,7 @@ namespace SIL.XForge.Realtime
                 AddDocConfig(projectDataDoc);
         }
 
-        internal RealtimeServer Server { get; }
+        internal IRealtimeServer Server { get; }
 
         public void StartServer()
         {
@@ -80,8 +80,16 @@ namespace SIL.XForge.Realtime
             return docConfig.CollectionName;
         }
 
+        /// <summary>
+        /// Delete project-related docs from various collections.
+        /// </summary>
         public async Task DeleteProjectAsync(string projectId)
         {
+            if (string.IsNullOrEmpty(projectId))
+            {
+                throw new ArgumentException("", nameof(projectId));
+            }
+
             RealtimeOptions options = _realtimeOptions.Value;
             var tasks = new List<Task>();
             foreach (DocConfig docConfig in options.ProjectDataDocs)
--- a/src/SIL.XForge/Realtime/RealtimeServiceCollectionExtensions.cs
+++ b/src/SIL.XForge/Realtime/RealtimeServiceCollectionExtensions.cs
@@ -26,7 +26,7 @@ namespace Microsoft.Extensions.DependencyInjection
             services.AddSingleton<IJsonService, RealtimeJsonService>();
 
             services.Configure(configureOptions);
-            services.AddSingleton<RealtimeServer>();
+            services.AddSingleton<IRealtimeServer, RealtimeServer>();
             services.AddSingleton<IRealtimeService, RealtimeService>();
             return services;
         }
--- a/src/SIL.XForge/Services/IProjectService.cs
+++ b/src/SIL.XForge/Services/IProjectService.cs
@@ -13,5 +13,6 @@ namespace SIL.XForge.Services
             Stream inputStream);
         Task DeleteAudioAsync(string curUserId, string projectId, string ownerId, string dataId);
         Task SetSyncDisabledAsync(string curUserId, string systemRole, string projectId, bool isDisabled);
+        Task RemoveUserFromAllProjectsAsync(string curUserId, string projectUserId);
     }
 }
--- a/src/SIL.XForge/Services/ProjectService.cs
+++ b/src/SIL.XForge/Services/ProjectService.cs
@@ -2,7 +2,6 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
-using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Options;
 using SIL.XForge.Configuration;
@@ -72,10 +71,46 @@ namespace SIL.XForge.Services
 
                 if (curUserId != projectUserId && !IsProjectAdmin(projectDoc.Data, curUserId))
                     throw new ForbiddenException();
+                await RemoveUserCoreAsync(conn, curUserId, projectId, projectUserId);
+            }
+        }
 
+
+        /// <summary>
+        /// Disassociate user projectUserId from project projectId, without checking permissions.
+        /// </summary>
+        private async Task RemoveUserCoreAsync(IConnection conn, string curUserId, string projectId, string projectUserId)
+        {
+            if (curUserId == null || projectId == null || projectUserId == null)
+            {
+                throw new ArgumentNullException();
+            }
+            IDocument<TModel> projectDoc = await GetProjectDocAsync(projectId, conn);
+            IDocument<User> userDoc = await GetUserDocAsync(projectUserId, conn);
+            await RemoveUserFromProjectAsync(conn, projectDoc, userDoc);
+        }
+
+        /// <summary>
+        /// Disassociate user projectUserId from all projects on the site that this ProjectService is operating on.
+        /// As requested by curUserId. Permissions to do so are not checked.
+        /// </summary>
+        public async Task RemoveUserFromAllProjectsAsync(string curUserId, string projectUserId)
+        {
+            if (curUserId == null || projectUserId == null)
+            {
+                throw new ArgumentNullException();
+            }
+            using (IConnection conn = await RealtimeService.ConnectAsync(curUserId))
+            {
                 IDocument<User> userDoc = await GetUserDocAsync(projectUserId, conn);
-
-                await RemoveUserFromProjectAsync(conn, projectDoc, userDoc);
+                IEnumerable<Task> removalTasks = userDoc.Data.Sites[SiteOptions.Value.Id].Projects.Select(
+                    (string projectId) => RemoveUserCoreAsync(conn, curUserId, projectId, projectUserId));
+                // The removals can be processed in parallel in production, but for unit tests, MemoryRealtimeService
+                // does not fully implement concurrent editing of docs, so run them in a sequence.
+                foreach (Task task in removalTasks)
+                {
+                    await task;
+                }
             }
         }
 
--- a/src/SIL.XForge/Services/ProjectService.cs
+++ b/src/SIL.XForge/Services/ProjectService.cs
@@ -1,5 +1,8 @@
 using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Linq;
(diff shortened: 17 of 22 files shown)


irahopkinson (changes requested):
Now that this branch is testng without errors, it also needs to be moved before review.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896#-:-MOt5riRAY6sSxF0NMwv:bs9iuv1)* status: 0 of 42 files reviewed, all discussions resolved

marksvc:
I rebased the branch onto the adjusted master branch. It's ready for review.

irahopkinson (changes requested):
This looks fine generally - have only one blocking comment. However I would like @ddaspit to look at it as it does cut into some architectual considerations.

> Reviewed 7 of 13 files at r1, 15 of 15 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896#-:-MQaKC5p8rf84_v1xh_w:bdug2vj)* status: all files reviewed, 5 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge/Realtime/RealtimeService.cs, line 136 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQa44Dr96WQVgo4mjYa:-MQa44Dr96WQVgo4mjYb:b-efewvu) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Realtime/RealtimeService.cs#L136)):*
> ```
> HasOt
> ```
nit: according to our MS style guide, in C# we honor 2 letter acronyms but not 3+ letter acronyms.
___
*[src/SIL.XForge/Realtime/RealtimeService.cs, line 144 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQa5BDi80egkIiBtxx8:-MQa5BDi80egkIiBtxx9:b-i6f2lz) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Realtime/RealtimeService.cs#L144)):*
> ```
> if (collection.TokenLocationInId == Location.Whole)
> ```
Shouldn't this be a case structure?
___
*[src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs, line 32 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQaChp_6AGuBgSR5UE2:-MQaChpa5Z1EbwfxFaoK:b-z7igdx) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs#L32)):*
> ```
> new DocConfig("user_secrets", typeof(UserSecret), null),
> ```
I thought we had talked about not removing from this collection? Isn't it needed for sync? @ddaspit can you confirm also?
___
*[src/SIL.XForge.Scripture/Services/SFServiceCollectionExtensions.cs, line 20 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQaDAny3MEDVigNKQPE:-MQaDAny3MEDVigNKQPF:bbqo145) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge.Scripture/Services/SFServiceCollectionExtensions.cs#L20)):*
> ```
> // When UserService requests an IProjectService, satisfy it with the SFProjectService.
> ```
This comment seems unecessary.
___
*[test/SIL.XForge.Tests/Services/UserServiceTests.cs, line 198 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQaF59rCU6CVBZvwG67:-MQaF59rCU6CVBZvwG68:bgsh505) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/test/SIL.XForge.Tests/Services/UserServiceTests.cs#L198)):*
> ```
> MockAuthService
> ```
nit: I'm not sure it is necessary to add `Mock` to this name. The above 2 are also mocks without it in the name. In fact everything is a mock apart from the service being tested, as should be expected.

irahopkinson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896)* status: all files reviewed, 6 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-owner/checking-owner.component.ts, line 32 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQaM8rf4_XZ9I3zE2mF:-MQaM8riCgBW-esAi3u0:bz93mob) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-owner/checking-owner.component.ts#L32)):*
> ```
> return this.translocoService.translate('checking.unknown_author');
> ```
Does this cover everywhere another users name could show up? Question, answers, comments?

ddaspit (changes requested):
> Reviewed 7 of 13 files at r1, 15 of 15 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896)* status: all files reviewed, 11 unresolved discussions (waiting on @irahopkinson and @marksvc)

___
*[src/SIL.XForge/Realtime/MemoryRealtimeService.cs, line 41 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQjf_BvDmB_VEgSebbz:-MQjf_BvDmB_VEgSebc-:b-cv8v8r) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Realtime/MemoryRealtimeService.cs#L41)):*
> ```Smalltalk
>         /// Count of calls to DeleteUserAsync(), for tests.
>         /// </summary>
>         internal int callCountDeleteUserAsync = 0;
> ```

Since this is externally accessible, I would make this a property.
___
*[src/SIL.XForge/Services/ProjectService.cs, line 97 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQjbEmr4I9TUFPnATnR:-MQjbEmr4I9TUFPnATnS:bstp5ui) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Services/ProjectService.cs#L97)):*
> ```Smalltalk
>                 // Use a mutex, since disassociating a user and a project can be incomplete if there are two
>                 // simultaneous requests regarding two different projects and the same user.
>                 await _userRemoveProjectMutex.WaitAsync();
> ```

This doesn't seem like it should be necessary. Can you give me more details about why a lock is needed?
___
*[src/SIL.XForge/Services/UserService.cs, line 155 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQjf2jwAdU0FyHNVmnp:-MQjf2jwAdU0FyHNVmnq:buged16) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Services/UserService.cs#L155)):*
> ```Smalltalk
>             }
> 
>             await _projectService.RemoveUserFromAllProjectsAsync(curUserId, userId);
> ```

You never submit a delete op for the user doc, i.e. call `IDocument.DeleteAsync()`.
___
*[src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs, line 31 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQjdVmg0oqumWbQ_zi4:-MQjdVmh-VzxHgdnGT_K:bo2qpfb) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs#L31)):*
> ```Smalltalk
>                     o.UserDataDocs.AddRange(new[]
>                     {
>                         new DocConfig("users", typeof(User)),
> ```

It looks like you are following the same pattern as project docs, which is a good idea. If you look at the `ProjectDataDocs` property, it includes real-time collections that are associated with a project, but it does not include the collection for the actual project docs. That is defined by the `ProjectDoc` property. You should do the same for the `UserDataDocs`, which means that you should not include the `users` doc config. The `users` doc config is defined by the `UserDoc` property.
___
*[src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs, line 32 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQaChp_6AGuBgSR5UE2:-MQjcDikDH9igKGmvN59:bxexgh5) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs#L32)):*

> ```
> new DocConfig("user_secrets", typeof(UserSecret), null),
> ```
I thought we had talked about not removing from this collection? Isn't it needed for sync? @ddaspit can you confirm also?
</blockquote></details>

@irahopkinson, you are thinking of the `SyncUsers` property in the `SFProjectSecret` class. In any case, I would not include this as a user data doc. These are specifically referring to real-time (ShareDB) docs. `UserSecret` docs are not real-time. I would just delete the `UserSecret` doc in `UserService.DeleteAsync()`. 
___
*[src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs, line 35 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQjeist4pMOVJcs-jqp:-MQjeisu-2rHVWnmVbtQ:b-n57010) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs#L35)):*
> ```Smalltalk
>                         new DocConfig("sf_project_user_configs", typeof(SFProjectUserConfig))
>                     });
>                     o.UserDataDocsIdLocation = new Dictionary<string, Location>
> ```

With the changes suggested above, I don't think `UserDataDocsIdLocation` is necessary anymore.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896)* status: all files reviewed, 13 unresolved discussions (waiting on @irahopkinson and @marksvc)

___
*[src/SIL.XForge/Services/ProjectService.cs, line 86 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQkJVi1DaA_ftHP2wiu:-MQkJVi22Ct2g2C_lmVO:b-ont05a) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Services/ProjectService.cs#L86)):*
> ```Smalltalk
>         /// Disassociate user projectUserId from project projectId, without checking permissions.
>         /// </summary>
>         private async Task RemoveUserCoreAsync(string curUserId, string projectId, string projectUserId)
> ```

You should pass in the connection object from the calling method instead of creating a new one.
___
*[src/SIL.XForge/Services/UserService.cs, line 140 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQkJDEj-lG48Zd4EnlL:-MQkJDEj-lG48Zd4EnlM:b-1u61uj) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Services/UserService.cs#L140)):*
> ```Smalltalk
>             if (string.IsNullOrEmpty(curUserId))
>             {
>                 throw new ArgumentException("", nameof(curUserId));
> ```

These should probably be `ArgumentNullException`s.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896)* status: all files reviewed, 13 unresolved discussions (waiting on @irahopkinson and @marksvc)

___
*[src/SIL.XForge/Services/ProjectService.cs, line 97 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQjbEmr4I9TUFPnATnR:-MQkiNji3KnOIeQj73Zf:b-ibm6n) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Services/ProjectService.cs#L97)):*

This doesn't seem like it should be necessary. Can you give me more details about why a lock is needed?
</blockquote></details>

I have tried to reproduce the problem that necessitated the lock, but I haven't be able to. I removed the lock and tried to delete a user that is a member of multiple projects and it seemed to work fine. It would be great if you could outline the steps to reproduce the issue.

marksvc (commented):
Thanks guys. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896#-:-MRMfncT2v4WCrn1WwIC:bnx2ess)* status: 10 of 22 files reviewed, 12 unresolved discussions (waiting on @ddaspit and @irahopkinson)

___
*[src/SIL.XForge/Realtime/MemoryRealtimeService.cs, line 41 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQjf_BvDmB_VEgSebbz:-MR6eDNB5L6PXcOUGefH:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Realtime/MemoryRealtimeService.cs#L41)):*

Since this is externally accessible, I would make this a property.
</blockquote></details>

Done.
___
*[src/SIL.XForge/Realtime/RealtimeService.cs, line 51 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MRMExH67MllK1sXTcwi:-MRMExH67MllK1sXTcwj:b-obb81s) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Realtime/RealtimeService.cs#L51)):*
> ```Smalltalk
>             foreach (DocConfig userDataDoc in options.UserDataDocs)
>             {
>                 AddDocConfig(userDataDoc);
> ```

I put this in, based on how these objects were being noted and made lookup-able. But is there any use to doing this? I see that it could conceivably get used by other parts of the code. But it obviously wasn't necessary so far. 
___
*[src/SIL.XForge/Realtime/RealtimeService.cs, line 136 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQa44Dr96WQVgo4mjYa:-MRLqMVFDPCqpn_9HsZB:bx7pexv) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Realtime/RealtimeService.cs#L136)):*

> ```
> HasOt
> ```
nit: according to our MS style guide, in C# we honor 2 letter acronyms but not 3+ letter acronyms.
</blockquote></details>

Done. I am having trouble finding that style guide, or that specific point in the style guides that I was looking at. Can you point me to this resource?
___
*[src/SIL.XForge/Realtime/RealtimeService.cs, line 144 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQa5BDi80egkIiBtxx8:-MRMBbQe182QuioFLC91:bcub70c) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Realtime/RealtimeService.cs#L144)):*

> ```
> if (collection.TokenLocationInId == Location.Whole)
> ```
Shouldn't this be a case structure?
</blockquote></details>

Yes, thank you. But now removed.
___
*[src/SIL.XForge/Services/ProjectService.cs, line 86 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQkJVi1DaA_ftHP2wiu:-MRMCBfj3FXpqj7vFqCk:b-wtkc5h) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Services/ProjectService.cs#L86)):*

You should pass in the connection object from the calling method instead of creating a new one.
</blockquote></details>

Yes, thank you.
___
*[src/SIL.XForge/Services/ProjectService.cs, line 97 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQjbEmr4I9TUFPnATnR:-MRMaRU-AxMDZYLaS2K0:b-cqhoph) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Services/ProjectService.cs#L97)):*

I have tried to reproduce the problem that necessitated the lock, but I haven't be able to. I removed the lock and tried to delete a user that is a member of multiple projects and it seemed to work fine. It would be great if you could outline the steps to reproduce the issue. 
</blockquote></details>

Sure. And it's not regarding deleting a user who is a member of multiple projects. It's about simultaneous requests to disassociate a user from different projects. (In other words, PT-Admin-Bob requests to remove user Charlie from ProjectAlpha. At the exact same time, PT-Admin-Julie requests to remove user Charlie from ProjectBravo.)

In RemoveUserFromProject(), a userDoc is examined and its Data.Sites[].Projects is manipulated by the index of where a project is in the Projects list.
If there are two simultaneous requests that both do
```cs
IDocument<User> userDoc = await GetUserDocAsync(projectUserId, conn);
await RemoveUserFromProjectAsync(conn, projectDoc, userDoc);
```
then the `userDoc` in each case could contain the original list of Projects before either is passed to `RemoveUserFromProjectAsync`. 

So let's say, from the example given earlier in this message, that after each PT Admin makes the removal request, each gets as far as assigning to userDoc but doesn't yet call `RemoveUserFromProjectAsync()`. 
Then PT-Admin-Bob's request runs `RemoveUserFromProjectAsync()` to remove Charlie from ProjectAlpha.
Then PT-Admin-Julie's request runs `RemoveUserFromProjectAsync()`. When this happens, the userDoc passed to `RemoveUserFromProjectAsync()` still contains ProjectAlpha in the user doc Data.Sites[].Projects list. This causes the IndexOf() call to give potentially the wrong result:
```cs
int index = userDoc.Data.Sites[siteId].Projects.IndexOf(projectDoc.Id);
```
And then we attempt to remove an item from the Projects list using the wrong location:
```cs
op.Remove(u => u.Sites[siteId].Projects, index);
```

I wrote a unit test that sets up the problem and included it. If you comment out 
```cs
await _userRemoveProjectMutex.WaitAsync();
``` 
and 
```cs
_userRemoveProjectMutex.Release();
```
you can see test RemoveUserAsync_NoRaceFailure begin to fail. (Or, it is on my computer. The very nature of the race could mean it won't necessarily fail each time or for everyone.)

I tried reproducing this in a couple web browser windows, where two project user lists were shown and I attempted to quickly request that the user be disassociated from the first item in the user's sites.sf.projects list, and immediately request, in the other browser window, that the user be disassociated from the second item in the list. But I can't click fast enough. Maybe two people could try reproducing it with two machines. 

It seems that IConnection.FetchAsync<>() (called from GetUserDocAsync) needs to be called each time a fresh, up-to-date doc is required. 
___
*[src/SIL.XForge/Services/UserService.cs, line 140 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQkJDEj-lG48Zd4EnlL:-MRMBMwq7sKtru8kf98y:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Services/UserService.cs#L140)):*

These should probably be `ArgumentNullException`s.
</blockquote></details>

Done.
___
*[src/SIL.XForge/Services/UserService.cs, line 155 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQjf2jwAdU0FyHNVmnp:-MRLqoqf1SRYxoEfYgbE:beaeom7) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Services/UserService.cs#L155)):*

You never submit a delete op for the user doc, i.e. call `IDocument.DeleteAsync()`.
</blockquote></details>

Right. That's on purpose because I am going to remove the user doc and its OTs in RealtimeService.DeleteUserAsync(). So any trace of whether we do or don't also call `IDocument.DeleteAsync()` on the user doc, will presumably be lost, making it meaningless to call DeleteAsync on the user doc. But maybe I don't have an accurate understanding. How will it help to first do 
```cs
                IDocument<User> userDoc = await conn.FetchAsync<User>(userId);
                await userDoc.DeleteAsync();
```
before 
```cs
await _realtimeService.DeleteUserAsync(userId);
```
?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-owner/checking-owner.component.ts, line 32 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQaM8rf4_XZ9I3zE2mF:-MRMQy_J53OR1DLMemz8:b-usolpk) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-owner/checking-owner.component.ts#L32)):*

> ```
> return this.translocoService.translate('checking.unknown_author');
> ```
Does this cover everywhere another users name could show up? Question, answers, comments?
</blockquote></details>

Answers - yes
Comments - yes

I'm not sure how to see the author name of a Question. If that's possible, maybe you can give me some pointers.

It looks like a user's display name can also be used in the filename of a recording (displayName + '.webm' in checking-audio-recorder.component.ts). I tried making a recording and playing with that a bit but it wasn't clear how this might be exposed. The resulting file name in /var/lib/scriptureforge/audio/foo/ is an .mp3 file that doesn't contain the user display name anyway, so I'm not sure if I did the right thing actually.

It looks like other usages of a user's display name, in our project's code, should be things that would only work for an existing user.
___
*[src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs, line 31 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQjdVmg0oqumWbQ_zi4:-MRLzcfs8afhFrZXF-Ab:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs#L31)):*

It looks like you are following the same pattern as project docs, which is a good idea. If you look at the `ProjectDataDocs` property, it includes real-time collections that are associated with a project, but it does not include the collection for the actual project docs. That is defined by the `ProjectDoc` property. You should do the same for the `UserDataDocs`, which means that you should not include the `users` doc config. The `users` doc config is defined by the `UserDoc` property.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs, line 32 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQaChp_6AGuBgSR5UE2:-MRM-12l-2j5pWnarLHw:bfp0cz) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs#L32)):*

@irahopkinson, you are thinking of the `SyncUsers` property in the `SFProjectSecret` class. In any case, I would not include this as a user data doc. These are specifically referring to real-time (ShareDB) docs. `UserSecret` docs are not real-time. I would just delete the `UserSecret` doc in `UserService.DeleteAsync()`. 
</blockquote></details>

Thank you. I think am/was confused as to what extent OTs and ShareDB cover our data storage, so this is helpful to make this differentiation. Done.
___
*[src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs, line 35 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQjeist4pMOVJcs-jqp:-MRM99wICyTrSwWM_cLy:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge.Scripture/Realtime/SFRealtimeServiceCollectionExtensions.cs#L35)):*

With the changes suggested above, I don't think `UserDataDocsIdLocation` is necessary anymore.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/Services/SFServiceCollectionExtensions.cs, line 20 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQaDAny3MEDVigNKQPE:-MRLsslf22VKQXajcHhI:bkdbhmn) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge.Scripture/Services/SFServiceCollectionExtensions.cs#L20)):*

> ```
> // When UserService requests an IProjectService, satisfy it with the SFProjectService.
> ```
This comment seems unecessary.
</blockquote></details>

Thank you. I wonder if I had just figured out something that seemed tricky and so I figured a comment was warranted. After the fact it doesn't seem significant. Done.
___
*[test/SIL.XForge.Tests/Services/UserServiceTests.cs, line 198 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQaF59rCU6CVBZvwG67:-MRLtQzcE0XzDCEsQAZt:bxw34gd) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/test/SIL.XForge.Tests/Services/UserServiceTests.cs#L198)):*

> ```
> MockAuthService
> ```
nit: I'm not sure it is necessary to add `Mock` to this name. The above 2 are also mocks without it in the name. In fact everything is a mock apart from the service being tested, as should be expected.
</blockquote></details>

Ah ha, here we get into terminology over different substitutions. So in my thinking, our MemoryRepository<UserSecret> and MemoryRealtimeService are functional things, useful for tests, but that aren't the real-production things. And our objects for IAuthService and IProjectService are just stand-ins that are pretty devoid of any functionality. So with that distinction, the IAuthService and IProjectService won a `Mock` prefix on their name. 

But it's not really important to me right here. :)  Done.

irahopkinson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896)* status: 10 of 22 files reviewed, 9 unresolved discussions (waiting on @ddaspit, @irahopkinson, and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-owner/checking-owner.component.ts, line 32 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQaM8rf4_XZ9I3zE2mF:-MRNUnsn-GLl9h-hz15A:b6brjxt) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-owner/checking-owner.component.ts#L32)):*

Answers - yes
Comments - yes

I'm not sure how to see the author name of a Question. If that's possible, maybe you can give me some pointers.

It looks like a user's display name can also be used in the filename of a recording (displayName + '.webm' in checking-audio-recorder.component.ts). I tried making a recording and playing with that a bit but it wasn't clear how this might be exposed. The resulting file name in /var/lib/scriptureforge/audio/foo/ is an .mp3 file that doesn't contain the user display name anyway, so I'm not sure if I did the right thing actually.

It looks like other usages of a user's display name, in our project's code, should be things that would only work for an existing user.

</blockquote></details>

Questions, answers and comments all inherit from `OwnedData` which contains and `ownerRef` string.
It sounds like the audio filename shouldn't be a problem but you could check with @raymondluong3.

ddaspit (changes requested):
> Reviewed 12 of 12 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896)* status: all files reviewed, 3 unresolved discussions (waiting on @marksvc and @RaymondLuong3)

___
*[src/SIL.XForge/Realtime/RealtimeService.cs, line 51 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MRMExH67MllK1sXTcwi:-MRNkbyt0ZyUveYOX_xC:bv65r9g) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Realtime/RealtimeService.cs#L51)):*

I put this in, based on how these objects were being noted and made lookup-able. But is there any use to doing this? I see that it could conceivably get used by other parts of the code. But it obviously wasn't necessary so far. 
</blockquote></details>

This is the correct. `UserDataDocs` is used so that SIL.XForge doesn't have to know about any of the SF real-time docs.
___
*[src/SIL.XForge/Services/ProjectService.cs, line 97 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQjbEmr4I9TUFPnATnR:-MRNdKNDAqjzUElTKJuh:blrz9nr) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Services/ProjectService.cs#L97)):*

Sure. And it's not regarding deleting a user who is a member of multiple projects. It's about simultaneous requests to disassociate a user from different projects. (In other words, PT-Admin-Bob requests to remove user Charlie from ProjectAlpha. At the exact same time, PT-Admin-Julie requests to remove user Charlie from ProjectBravo.)

In RemoveUserFromProject(), a userDoc is examined and its Data.Sites[].Projects is manipulated by the index of where a project is in the Projects list.
If there are two simultaneous requests that both do
```cs
IDocument<User> userDoc = await GetUserDocAsync(projectUserId, conn);
await RemoveUserFromProjectAsync(conn, projectDoc, userDoc);
```
then the `userDoc` in each case could contain the original list of Projects before either is passed to `RemoveUserFromProjectAsync`. 

So let's say, from the example given earlier in this message, that after each PT Admin makes the removal request, each gets as far as assigning to userDoc but doesn't yet call `RemoveUserFromProjectAsync()`. 
Then PT-Admin-Bob's request runs `RemoveUserFromProjectAsync()` to remove Charlie from ProjectAlpha.
Then PT-Admin-Julie's request runs `RemoveUserFromProjectAsync()`. When this happens, the userDoc passed to `RemoveUserFromProjectAsync()` still contains ProjectAlpha in the user doc Data.Sites[].Projects list. This causes the IndexOf() call to give potentially the wrong result:
```cs
int index = userDoc.Data.Sites[siteId].Projects.IndexOf(projectDoc.Id);
```
And then we attempt to remove an item from the Projects list using the wrong location:
```cs
op.Remove(u => u.Sites[siteId].Projects, index);
```

I wrote a unit test that sets up the problem and included it. If you comment out 
```cs
await _userRemoveProjectMutex.WaitAsync();
``` 
and 
```cs
_userRemoveProjectMutex.Release();
```
you can see test RemoveUserAsync_NoRaceFailure begin to fail. (Or, it is on my computer. The very nature of the race could mean it won't necessarily fail each time or for everyone.)

I tried reproducing this in a couple web browser windows, where two project user lists were shown and I attempted to quickly request that the user be disassociated from the first item in the user's sites.sf.projects list, and immediately request, in the other browser window, that the user be disassociated from the second item in the list. But I can't click fast enough. Maybe two people could try reproducing it with two machines. 

It seems that IConnection.FetchAsync<>() (called from GetUserDocAsync) needs to be called each time a fresh, up-to-date doc is required. 
</blockquote></details>

The problem you describe should not be possible with ShareDB. ShareDB is specifically designed to handle concurrent edits to the same document. Here is how this scenario works:

1. PT-Admin-Bob requests to remove user Charlie from ProjectAlpha and PT-Admin-Julie requests to remove user Charlie from ProjectBravo at the same time. They both attempt to update the same version of the user doc (version 1).
2. Arbitrarily, PT-Admin-Bob's request is received first.
3. The project id is properly removed from the user doc.
4. The doc version is incremented to 2.
5. PT-Admin-Julie's request is received second.
6. ShareDB realizes that PT-Admin-Julie's op is on an older version of the doc (version 1).
7. ShareDB transforms the PT-Admin-Julie's op according to all of the ops that have occurred since the doc version that PT-Admin-Julie is working on. This means that the index in PT-Admin-Julie's op is updated to reflect that a project id was removed by PT-Admin-Bob's op.
8. The transformed op is applied to the latest version (2) of the doc and the correct project id is removed.
9 The doc version is incremented to 3.

You are only able to reproduce this problem in unit tests, because the mock real-time service does not fully implement concurrent editing of docs. Ultimately, this means that the lock is not necessary. I hope all of that makes sense.
___
*[src/SIL.XForge/Services/UserService.cs, line 155 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQjf2jwAdU0FyHNVmnp:-MRNWDtVCCs-Q2JZRYTI:bcfihxw) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Services/UserService.cs#L155)):*

Right. That's on purpose because I am going to remove the user doc and its OTs in RealtimeService.DeleteUserAsync(). So any trace of whether we do or don't also call `IDocument.DeleteAsync()` on the user doc, will presumably be lost, making it meaningless to call DeleteAsync on the user doc. But maybe I don't have an accurate understanding. How will it help to first do 
```cs
                IDocument<User> userDoc = await conn.FetchAsync<User>(userId);
                await userDoc.DeleteAsync();
```
before 
```cs
await _realtimeService.DeleteUserAsync(userId);
```
?
</blockquote></details>

If you submit a delete op, then the client will be notified that the user was deleted and can properly handle the situation in real-time.  For example, triggering a redirect back to the landing page.

irahopkinson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896)* status: all files reviewed, 3 unresolved discussions (waiting on @marksvc and @RaymondLuong3)

___
*[src/SIL.XForge/Realtime/RealtimeService.cs, line 136 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQa44Dr96WQVgo4mjYa:-MROFI8n7TgO_RJi3PUt:b-6i0vyu) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Realtime/RealtimeService.cs#L136)):*

Done. I am having trouble finding that style guide, or that specific point in the style guides that I was looking at. Can you point me to this resource?
</blockquote></details>

Many that I found don't specifically point it out either but you can see it inherent in their example code. I got Damien to point me to this one that he/we have been using that goes into this level of detail https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit and @RaymondLuong3)

___
*[src/SIL.XForge/Realtime/RealtimeService.cs, line 136 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQa44Dr96WQVgo4mjYa:-MRRhNEZ-cPzICbQl8_M:b-ucfknh) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Realtime/RealtimeService.cs#L136)):*

Many that I found don't specifically point it out either but you can see it inherent in their example code. I got Damien to point me to this one that he/we have been using that goes into this level of detail https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions
</blockquote></details>

Thanks!
___
*[src/SIL.XForge/Services/ProjectService.cs, line 97 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQjbEmr4I9TUFPnATnR:-MRRhoTR6GjWh8ozJt2s:byv4k5l) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Services/ProjectService.cs#L97)):*

The problem you describe should not be possible with ShareDB. ShareDB is specifically designed to handle concurrent edits to the same document. Here is how this scenario works:

1. PT-Admin-Bob requests to remove user Charlie from ProjectAlpha and PT-Admin-Julie requests to remove user Charlie from ProjectBravo at the same time. They both attempt to update the same version of the user doc (version 1).
2. Arbitrarily, PT-Admin-Bob's request is received first.
3. The project id is properly removed from the user doc.
4. The doc version is incremented to 2.
5. PT-Admin-Julie's request is received second.
6. ShareDB realizes that PT-Admin-Julie's op is on an older version of the doc (version 1).
7. ShareDB transforms the PT-Admin-Julie's op according to all of the ops that have occurred since the doc version that PT-Admin-Julie is working on. This means that the index in PT-Admin-Julie's op is updated to reflect that a project id was removed by PT-Admin-Bob's op.
8. The transformed op is applied to the latest version (2) of the doc and the correct project id is removed.
9 The doc version is incremented to 3.

You are only able to reproduce this problem in unit tests, because the mock real-time service does not fully implement concurrent editing of docs. Ultimately, this means that the lock is not necessary. I hope all of that makes sense.
</blockquote></details>

Great. Thank you for the explanation. I removed the mutex.
___
*[src/SIL.XForge/Services/UserService.cs, line 155 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQjf2jwAdU0FyHNVmnp:-MRRjgPv7whNOSvjHAAL:bkdqlgp) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge/Services/UserService.cs#L155)):*

If you submit a delete op, then the client will be notified that the user was deleted and can properly handle the situation in real-time.  For example, triggering a redirect back to the landing page.
</blockquote></details>

Ah ha, yes. Thank you. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-owner/checking-owner.component.ts, line 32 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MQaM8rf4_XZ9I3zE2mF:-MRRpd0pEaAeHbT8xUbV:b4homow) ([raw file](https://github.com/sillsdev/web-xforge/blob/d37a96819bb363995c5dd0c9377b412890621e8e/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-owner/checking-owner.component.ts#L32)):*

Questions, answers and comments all inherit from `OwnedData` which contains and `ownerRef` string.
It sounds like the audio filename shouldn't be a problem but you could check with @raymondluong3.
</blockquote></details>

Looking thru some TS and C# code makes me think we drop the original filename and create a new filename based on things like user ID when saving it on the server. I'm asking Raymond what he thinks.

ddaspit (changes requested):
> Reviewed 4 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc and @RaymondLuong3)

___
*[src/SIL.XForge/Services/ProjectService.cs, line 24 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MRUBb5t6-Kv-8KB7u8n:-MRUBb5t6-Kv-8KB7u8o:buua2r0) ([raw file](https://github.com/sillsdev/web-xforge/blob/f85fa905d3f070fa2b20384e147cf7eaef995a14/src/SIL.XForge/Services/ProjectService.cs#L24)):*
> ```Smalltalk
>     {
>         private readonly IAudioService _audioService;
>         private readonly SemaphoreSlim _userRemoveProjectMutex = new SemaphoreSlim(1, 1);
> ```

You forgot to remove the mutex.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit and @RaymondLuong3)

___
*[src/SIL.XForge/Services/ProjectService.cs, line 24 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MRUBb5t6-Kv-8KB7u8n:-MRW8RQQ9UKri_q3TNOz:bb6un1j) ([raw file](https://github.com/sillsdev/web-xforge/blob/f85fa905d3f070fa2b20384e147cf7eaef995a14/src/SIL.XForge/Services/ProjectService.cs#L24)):*

You forgot to remove the mutex.
</blockquote></details>

Done

ddaspit (approved):
:lgtm:

> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896#-:-MRXoPWICMM2s7zZyKVB:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

irahopkinson (changes requested):
Thanks for this, however a TC test is failing. You'll also want to rebase since one of the test files has changed on master.

> Reviewed 8 of 12 files at r3, 3 of 4 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896#-:-MRYtZUO87bkJFEeoTbL:b-ishu6f)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[test/SIL.XForge.Tests/Services/ProjectServiceTests.cs, line 258 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MRYqmk1ACNNnnFX0M_g:-MRYqmk1ACNNnnFX0M_h:bxncl6o) ([raw file](https://github.com/sillsdev/web-xforge/blob/9148e16fde6390b6176a174c2276d1d5357b0d17/test/SIL.XForge.Tests/Services/ProjectServiceTests.cs#L258)):*
> ```Smalltalk
>             Assert.That(env.GetUser(userToDisassociate).Sites[SiteId].Projects, Does.Not.Contain(Project01));
>             Assert.That(env.GetProject(Project02).UserRoles, Does.Not.ContainKey(userToDisassociate));
>             Assert.That(env.GetUser(userToDisassociate).Sites[SiteId].Projects, Does.Not.Contain(Project02));
> ```

This line is failing on TC but it seems to pass locally. You'll likely need to login to the build agent and diagnose there. Let me know if you need help diagnosing this.

marksvc (commented):
I am able to get the same test failure on my workstation. Unfortunately, it's failing because of the race condition. If I re-introduce a mutex, it no longer fails on my machine. A simple way forward may be to just make the calls to RemoveUserCoreAsync() run in sequence rather than in parallel. I applied this in a revision. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896#-:-MRa8K6G47OHBb4oaupO:b-dai2oh)* status: 21 of 22 files reviewed, 1 unresolved discussion (waiting on @ddaspit and @irahopkinson)

irahopkinson (approved):
> Reviewed 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge/Services/ProjectService.cs, line 109 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MRbkOv_2yy7TY9Xa12L:-MRbkOv_2yy7TY9Xa12M:buyi4vu) ([raw file](https://github.com/sillsdev/web-xforge/blob/99d4cde4d65c770ba82f2257e8bc4460507e7efd/src/SIL.XForge/Services/ProjectService.cs#L109)):*
> ```
>                 // The removals can be processed in parallel in production, but for unit tests, MemoryRealtimeService
>                 // does not fully implement concurrent editing of docs, so run them in a sequence.
> ```
It's usually a bad idea to compromise production in order to enable tests. Why do you think its justified here? How can you change the tests to work around the limitations of the MemoryRealtimeService? Can something be done to fix this in MemoryRealtimeService?

Since we probably don't want to just remove the test line in order to not compromise production code, and we probably don't want to take the time to fix MemoryRealtimeService right now (which we could do by running ShareDB with a memory Mongo), and given the time difference between parallel and serial here is probably not user notice-able, and this isn't even user-facing (just SysAdmins), then this is the best solution. Well done, lets merge it.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/896)* status: all files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[src/SIL.XForge/Services/ProjectService.cs, line 109 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/896#-MRbkOv_2yy7TY9Xa12L:-MRugVjH3XNhB5IxQuWr:b-uw71vt) ([raw file](https://github.com/sillsdev/web-xforge/blob/99d4cde4d65c770ba82f2257e8bc4460507e7efd/src/SIL.XForge/Services/ProjectService.cs#L109)):*
> then this is the best solution. Well done, lets merge it.

Thank you. I wasn't sure if the earlier questions were something you wanted me to think thru and/or discuss. I haven't implemented anything but I did have a look around to see what I might be able to do as I considered the different questions, which I answer below:

Justified - Because it would be an unrecognizable performance degradation, and with behaviour that was either the same, or potentially better (if our C# or ShareDB isn't as concurrent as we thought). And we gain the unit test.

> How can you change the tests to work around the limitations of the MemoryRealtimeService? 

Right. Maybe we could use do something like mock IConnection, have it return a specific project doc and  user  doc that take note of the calls to IDocument.SubmitJson0OpAsync() and assert that appropriate operations were given by ProjectService such that the right things would have been done.  (Which would be unfortunate since we already have MemoryRealtimeService and MemoryRepository to let us check outcomes without needing to do this kind of thing.)
There may also be a Task-related setting where we can specify in the unit test that tasks should be executed in sequence rather than in parallel.
We could mock Task (and probably make a mess).

> Can something be done to fix this in MemoryRealtimeService?

Yes. I had a look around in some of the files in Realtime today. I think the right place to change the behaviour would be after the ops are fully built. (So we would let `op.Remove(u => u.Sites[siteId].Projects, index);` continue to use the same `index` value that is presently not concurrently compatible.) I see that MemoryDocument SubmitOpAsync() calls `MemoryRealtimeService.Server.ApplyOpAsync(OTTypeName, Data, op)` but then it's a bit confusing since it looks like it starts to use a real RealtimeService, so I would need to study it more to understand what it's doing.

Damien had described some of  ShareDB's relevant behaviour as:
> ShareDB realizes that PT-Admin-Julie's op is on an older version of the doc (version 1).
> ShareDB transforms the PT-Admin-Julie's op according to all of the ops that have occurred since the doc version that PT-Admin-Julie is working on. This means that the index in PT-Admin-Julie's op is updated to reflect that a project id was removed by PT-Admin-Bob's op.
> The transformed op is applied to the latest version (2) of the doc and the correct project id is removed.

There may be simpler ways to go about it for unit tests. But we'd want to talk about whether we should put in the effort.

And I see you noted it could be done "by running ShareDB with a memory Mongo", so a solution may be to just arrange different pieces together.

PR #898, opened 2020-12-17
SF-1155 Use same realtime server for beta

marksvc:
- Realtime updates will only work if beta and non-beta sites share the
same realtime server. So use the same realtime server. It will be the
one started by the non-beta app.
- So changing environment.prodBeta.ts and environment.stagingBeta.ts
`realtimePort` to match the non-beta setting.
- The beta apps will end up starting a realtime server, listening on
port 9003, that is not used, as specified in appsettings.*Beta.json.
- So to use the beta app, first start the non-beta app before starting
the beta app, so that the beta app can use the non-beta app's realtime
server. For example, on a workstation with two terminals in
`src/SIL.XForge.Scripture`, run the following two commands. Don't run
the second one until the first one has completely started, which may
take a few minutes.

    dotnet run --start-ng-serve=yes
    dotnet run --start-ng-serve=yes --environment=DevelopmentBeta

(Or note that `ng serve` can use `--configuration=developmentBeta`).
- If you just run the Beta app, then the web browser console will get
errors saying

> WebSocket connection to 'ws://beta.localhost:5003/?access_token=...'
> failed: Error in connection establishment:
> net::ERR_CONNECTION_REFUSED

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/898)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/angular.json
+++ b/src/SIL.XForge.Scripture/ClientApp/angular.json
@@ -70,6 +70,30 @@
               "buildOptimizer": true,
               "serviceWorker": true
             },
+            "productionBeta": {
+              "budgets": [
+                {
+                  "type": "anyComponentStyle",
+                  "maximumWarning": "6kb"
+                }
+              ],
+              "fileReplacements": [
+                {
+                  "replace": "src/environments/environment.ts",
+                  "with": "src/environments/environment.prodBeta.ts"
+                }
+              ],
+              "optimization": true,
+              "outputHashing": "all",
+              "sourceMap": true,
+              "extractCss": true,
+              "namedChunks": false,
+              "aot": true,
+              "extractLicenses": true,
+              "vendorChunk": false,
+              "buildOptimizer": true,
+              "serviceWorker": true
+            },
             "staging": {
               "budgets": [
                 {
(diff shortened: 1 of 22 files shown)


irahopkinson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/898)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @irahopkinson and @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/environments/environment.developmentBeta.ts, line 21 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/898#-MOnLwPI6P3P_ujqNU2Q:-MOnLwPI6P3P_ujqNU2R:b-fqdum8) ([raw file](https://github.com/sillsdev/web-xforge/blob/949abbe184a38056d532fca04313698109fcdb3f/src/SIL.XForge.Scripture/ClientApp/src/environments/environment.developmentBeta.ts#L21)):*
> ```
> 5003
> ```
This is wrong because if you run the beta independently it won't work. However, SF-1156 needs to make this change so that that code works. But in the meantime beta should still work. I've added a commit to fix this.

irahopkinson (commented):
> Reviewed 20 of 20 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/898)* status: 22 of 24 files reviewed, 1 unresolved discussion (waiting on @irahopkinson and @marksvc)

marksvc:
I don't think there is a shared understanding of what we are trying to do, so I will need to learn more. I am marking this as draft.

irahopkinson (approved):
> Reviewed 5 of 7 files at r4, 2 of 2 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/898)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #912, opened 2021-01-12
SF-1143 Wrap down front page button if text too long

marksvc:
- If the text on a Sign up or Log in button is long in a localization,
it can display only part of the text before truncating the rest.
- If we don't have enough room to show both buttons on the same line,
and show all of the button's text, then break the buttons up onto
different lines.
- Do this for both the top-of-page Sign up and Log in buttons, and for
the end of the Index page buttons Learn and Sign up.
- Wrapping the buttons was stacking them right on top of one another.
This commit adds some margin to the buttons so they have some space
between. But then the buttons have too much space under them, and
whether they are stacked or all on a horizontal line. So space for the
button margin was borrowed from surrounding elements' padding-bottom.

---

It seems to me that I should have been able to do some sort of 
```css
align-content: space-between;
```
instead of borrowing from one padding/margin and putting it in another. But I didn't have success with that. 

The top bar starts to draw over top of some of the page content. You can see that the SF logo and the languages combo box starts to be drawn over a bit when the view gets narrower. This is only an issue when the button text is long and the buttons get wrapped. So it's a bit of a trade off, but I don't particularly like how that's working. 

See animation for behaviour when the Sign up button has a long text string.

sf-1143b

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/912)


RaymondLuong3 (approved):
I played around with it and it looks good. Approved.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/912#-:-MQxZfKv63DmUesPRqft:ba1abm5)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #918, opened 2021-01-14
Return empty resource list if lookup error

marksvc:
- Querying for resources can get an ephemeral error, such as 401
Unauthorized. Trying again can be successful.
- Give the user an empty resource list rather than a 500 internal
server error popup dialog.
- There is still an underlying problem that has been around for maybe
a long time. Log the occurrence to bugsnag.

----
There is some more discussion about this in chat.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/918)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -299,6 +299,7 @@ namespace SIL.XForge.Scripture.Services
                         this._restClientFactory,
                         this._fileSystemService,
                         this._jwtTokenHelper,
+                        _exceptionHandler,
                         this._dblServerUri);
                 }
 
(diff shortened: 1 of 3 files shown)


jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/918)* status: :shipit: complete! all files reviewed, all discussions resolved

jasonleenaylor (approved):
I think this a good change, logging this issue but not crashing the user. This way we'll get a good grasp of how common the problem is. :LGTM:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/918#-:-MR24ADD40EM6pSxvtAL:b6ex38c)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #919, opened 2021-01-14
Add unit test for resource cloning fix

jasonleenaylor:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/919)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -313,17 +313,18 @@ namespace SIL.XForge.Scripture.Services
         }
 
         /// <summary>
-        /// Gets the Paratext username mapping.
+        /// Queries the ParatextRegistry for the project and builds a dictionary of SF user id
+        /// to paratext user names for members of the project.
         /// </summary>
         /// <param name="userSecret">The user secret.</param>
         /// <param name="paratextId">The project ParatextId.</param>
         /// <returns>
-        /// A dictionary where the key is user ID and the value is Paratext username.
+        /// A dictionary where the key is the SF user ID and the value is Paratext username. (May be empty)
         /// </returns>
         public async Task<IReadOnlyDictionary<string, string>> GetParatextUsernameMappingAsync(UserSecret userSecret,
             string paratextId)
         {
-            // See if the project is a resource
+            // Skip all the work if the project is a resource. Resources don't have project members
             if (paratextId.Length == SFInstallableDblResource.ResourceIdentifierLength)
             {
                 return new Dictionary<string, string>();
--- a/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs
+++ b/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs
@@ -693,6 +693,17 @@ namespace SIL.XForge.Scripture.Services
             await env.Service.SendReceiveAsync(user01Secret, resourceId);
         }
 
+
+        [Test]
+        public async Task GetParatextUsernameMappingAsync_ReturnsEmptyMappingForResourceProject()
+        {
+            var env = new TestEnvironment();
+            const string resourceId = "1234567890abcdef";
+            Assert.That(resourceId.Length, Is.EqualTo(SFInstallableDblResource.ResourceIdentifierLength));
+            var mapping = await env.Service.GetParatextUsernameMappingAsync(env.MakeUserSecret(env.User01, env.Username01), resourceId);
+            Assert.That(mapping.Count, Is.EqualTo(0));
+        }
+
         private class TestEnvironment
         {
             public readonly string Project01 = "project01";


marksvc:
Reviewing

marksvc (changes requested):
Sorry to make this complicated :). And maybe you needn't be the one to do what I discuss here.

What do you think of this?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/919#-:-MR5xzCX2vUE4ht-Z4pC:b-lqlh7l)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @jasonleenaylor)

___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 780 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/919#-MR5bCGP2L2P5PFXsXy8:-MR5bCGP2L2P5PFXsXy9:bu66288) ([raw file](https://github.com/sillsdev/web-xforge/blob/98a41f7f80790a8b8ef518f2c21c544332a70db8/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L780)):*
> ```
> Arg.Any<UserSecret>()
> ```
Oops; I think there was a mixup with how this was used. I just had a double check with https://nsubstitute.github.io/help/argument-matchers/ . 
I think you can probably use `new UserSecret()`, or you could get one of the user secrets from the memory repository that is set up. 
Actually, this test should be in ParatextServiceTests.cs anyway, right?  And there you can follow a pattern and do
```cs
UserSecret user01Secret = env.MakeUserSecret(env.User01, env.Username01);
```
to have a real user secret.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 781 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/919#-MR5dYlf6_ydtTLz1wTf:-MR5dYlf6_ydtTLz1wTg:bym7d7z) ([raw file](https://github.com/sillsdev/web-xforge/blob/98a41f7f80790a8b8ef518f2c21c544332a70db8/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L781)):*
> ```Smalltalk
>             Assert.That(resourceId.Length, Is.EqualTo(SFInstallableDblResource.ResourceIdentifierLength));
>             var mapping = await env.ParatextService.GetParatextUsernameMappingAsync(Arg.Any<UserSecret>(), resourceId);
>             Assert.That(mapping.Count, Is.EqualTo(0));
> ```

I had a closer look at GetParatextUsernameMappingAsync(), and where the method is being called. 

After looking over the method, I think we should clarify the method comment (if not method name) to explain that we are fetching the users on a PT project. So maybe the method comment could say "Get members of project, as authorized to query by userSecret.". (The current comment is not very helpful to understand what the method is actually meaning to do, and what significance the userSecret or paratextId has to its operation.)

I think that if it doesn't make sense to call 
```cs
CallApiAsync(_registryClient, userSecret, HttpMethod.Get,
                $"projects/{paratextId}/members")
```
where `paratextId` is a resource, then we should disallow that. And so GetParatextUsernameMappingAsync should throw ArgumentException if `paratextId` is not the kind of ID that we can use to look up a list of what people are members.

The method is only called in SF from ParatextSyncRunner. Let's modify  ParatextSyncRunner so we  only call the method iff we have an id that is the kind of thing we can/should query membership of. 

That makes this code review a mess. Because your unit test isn't terrible :). But it reveals that we should change the code being tested. 
The last few lines  here at the end then could be 
```cs
UserSecret user01Secret = env.MakeUserSecret(env.User01, env.Username01);
 // This is to make Tokens.ValidateLifetime() return false
user01Secret.ParatextTokens.AccessToken = null;

// SUT
ArgumentException thrown = Assert.ThrowsAsync<ArgumentException>(() => env.ParatxtService.GetParatextUsernameMappingAsync(user01Secret, resourceId));
Assert.That(thrown.Message, Does.Contain("must be a paratext project with members"));
// Should not have proceeded to talk with server
env.MockJwtTokenHelper.Received(0).RefreshAccessTokenAsync(Arg.Any<ParatextOptions>(), Arg.Any<Tokens>(), Arg.Any<HttpClient>());
```

jasonleenaylor (commented):
Here are my thoughts.

Your evaluation of the 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/919#-:-MR6-oRrC9JzmRHMR8Gp:bltfuay)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 780 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/919#-MR5bCGP2L2P5PFXsXy8:-MR60Q0v4s3x-X9fuDcl:b-2u6m65) ([raw file](https://github.com/sillsdev/web-xforge/blob/98a41f7f80790a8b8ef518f2c21c544332a70db8/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L780)):*

> ```
> Arg.Any<UserSecret>()
> ```
Oops; I think there was a mixup with how this was used. I just had a double check with https://nsubstitute.github.io/help/argument-matchers/ . 
I think you can probably use `new UserSecret()`, or you could get one of the user secrets from the memory repository that is set up. 
Actually, this test should be in ParatextServiceTests.cs anyway, right?  And there you can follow a pattern and do
```cs
UserSecret user01Secret = env.MakeUserSecret(env.User01, env.Username01);
```
to have a real user secret.
</blockquote></details>

That makes sense, but there isn't any such method in the environment currently. The Arg.Any<UserSecret> was used in several other unit tests in this file. So if there is a more appropriate way to create a user secret that we don't need for the SUT we should change it throughout.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 781 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/919#-MR5dYlf6_ydtTLz1wTf:-MR60c8_BjIi5YQMELTj:b-ft5nr2) ([raw file](https://github.com/sillsdev/web-xforge/blob/98a41f7f80790a8b8ef518f2c21c544332a70db8/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L781)):*

I had a closer look at GetParatextUsernameMappingAsync(), and where the method is being called. 

After looking over the method, I think we should clarify the method comment (if not method name) to explain that we are fetching the users on a PT project. So maybe the method comment could say "Get members of project, as authorized to query by userSecret.". (The current comment is not very helpful to understand what the method is actually meaning to do, and what significance the userSecret or paratextId has to its operation.)

I think that if it doesn't make sense to call 
```cs
CallApiAsync(_registryClient, userSecret, HttpMethod.Get,
                $"projects/{paratextId}/members")
```
where `paratextId` is a resource, then we should disallow that. And so GetParatextUsernameMappingAsync should throw ArgumentException if `paratextId` is not the kind of ID that we can use to look up a list of what people are members.

The method is only called in SF from ParatextSyncRunner. Let's modify  ParatextSyncRunner so we  only call the method iff we have an id that is the kind of thing we can/should query membership of. 

That makes this code review a mess. Because your unit test isn't terrible :). But it reveals that we should change the code being tested. 
The last few lines  here at the end then could be 
```cs
UserSecret user01Secret = env.MakeUserSecret(env.User01, env.Username01);
 // This is to make Tokens.ValidateLifetime() return false
user01Secret.ParatextTokens.AccessToken = null;

// SUT
ArgumentException thrown = Assert.ThrowsAsync<ArgumentException>(() => env.ParatxtService.GetParatextUsernameMappingAsync(user01Secret, resourceId));
Assert.That(thrown.Message, Does.Contain("must be a paratext project with members"));
// Should not have proceeded to talk with server
env.MockJwtTokenHelper.Received(0).RefreshAccessTokenAsync(Arg.Any<ParatextOptions>(), Arg.Any<Tokens>(), Arg.Any<HttpClient>());
```
</blockquote></details>

I can include a method comment clarification, however, your suggested one is more confusing to me then what is already there. I'll give it a shot.

In my second reading of the code the returning of an empty dictionary from this method is reasonable and also a possibility even for non-resource projects. The resource test could be moved outside of the method as you suggest, but doesn't provide any additional value that I can see.

A useful refactor might be to have separate functions for fetching the permissions for resource and non-resource paratext projects. That might result in more code duplication while reducing the number of places where the code has to test if the target project is a resource. If we find another bug in this area then the person working on it should consider that refactor.

marksvc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/919)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @jasonleenaylor)

___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 780 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/919#-MR5bCGP2L2P5PFXsXy8:-MR6UXkvA8de-t2sCgH6:b6lfl7d) ([raw file](https://github.com/sillsdev/web-xforge/blob/98a41f7f80790a8b8ef518f2c21c544332a70db8/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L780)):*

That makes sense, but there isn't any such method in the environment currently. The Arg.Any<UserSecret> was used in several other unit tests in this file. So if there is a more appropriate way to create a user secret that we don't need for the SUT we should change it throughout.
</blockquote></details>

So when we write an expectation, like 

```cs
env.ParatextService.DidNotReceive().PutNotes(Arg.Any<UserSecret>(), "target", Arg.Any<string>());
```

or if we define a mock's behaviour, like

```cs
env.ParatextService.GetProjectRolesAsync(Arg.Any<UserSecret>(), "target")
    .Returns(Task.FromResult<IReadOnlyDictionary<string, string>>(ptUserRoles));
```

Arg.Any<T>() is helpful in being a typed wildcard where we can ask that the expectation, or behaviour response, apply for any object of the specified type. 
I understand these to be different than actually _calling_ and running a method, where we pass arguments to the method for it to use in performing an operation.

While it's possible that `Arg.Any<T>()` could some day be used when calling a method, and  allow us to write  a test where the test runner _expands_ it to multiple values for different variable permutations and bounds checking, or where a test runner might substitute it with something like `default(T)`, the docs (in the URL earlier in the thread) indicate that that's not the way to use this at present. So unless I'm unaware of a recent advancement in Arg.Any<T>() usage, and assuming this line is the SUT, I think this was a misapplication of Arg.Any. But if we still don't see eye to eye about it, let's chat on the phone and I can learn more from you about it.

Also, what's the reason this test was placed in ParatextSyncRunnerTests.cs instead of in ParatextServiceTests.cs? Can we place it there instead since the test is regarding a method in ParatextService.cs?
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 781 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/919#-MR5dYlf6_ydtTLz1wTf:-MR6WprqCmB3PTvVKdV8:bndnpob) ([raw file](https://github.com/sillsdev/web-xforge/blob/98a41f7f80790a8b8ef518f2c21c544332a70db8/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L781)):*

I can include a method comment clarification, however, your suggested one is more confusing to me then what is already there. I'll give it a shot.

In my second reading of the code the returning of an empty dictionary from this method is reasonable and also a possibility even for non-resource projects. The resource test could be moved outside of the method as you suggest, but doesn't provide any additional value that I can see.

A useful refactor might be to have separate functions for fetching the permissions for resource and non-resource paratext projects. That might result in more code duplication while reducing the number of places where the code has to test if the target project is a resource. If we find another bug in this area then the person working on it should consider that refactor.
</blockquote></details>

Hmmmm. Okay.

jasonleenaylor (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/919)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 780 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/919#-MR5bCGP2L2P5PFXsXy8:-MR6Zc8oAzyGzJ-9eU8y:b-26cqsi) ([raw file](https://github.com/sillsdev/web-xforge/blob/98a41f7f80790a8b8ef518f2c21c544332a70db8/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L780)):*

So when we write an expectation, like 

```cs
env.ParatextService.DidNotReceive().PutNotes(Arg.Any<UserSecret>(), "target", Arg.Any<string>());
```

or if we define a mock's behaviour, like

```cs
env.ParatextService.GetProjectRolesAsync(Arg.Any<UserSecret>(), "target")
    .Returns(Task.FromResult<IReadOnlyDictionary<string, string>>(ptUserRoles));
```

Arg.Any<T>() is helpful in being a typed wildcard where we can ask that the expectation, or behaviour response, apply for any object of the specified type. 
I understand these to be different than actually _calling_ and running a method, where we pass arguments to the method for it to use in performing an operation.

While it's possible that `Arg.Any<T>()` could some day be used when calling a method, and  allow us to write  a test where the test runner _expands_ it to multiple values for different variable permutations and bounds checking, or where a test runner might substitute it with something like `default(T)`, the docs (in the URL earlier in the thread) indicate that that's not the way to use this at present. So unless I'm unaware of a recent advancement in Arg.Any<T>() usage, and assuming this line is the SUT, I think this was a misapplication of Arg.Any. But if we still don't see eye to eye about it, let's chat on the phone and I can learn more from you about it.

Also, what's the reason this test was placed in ParatextSyncRunnerTests.cs instead of in ParatextServiceTests.cs? Can we place it there instead since the test is regarding a method in ParatextService.cs?
</blockquote></details>

Oh, I'm not saying we don't see eye to eye. I'm saying I copied previous potentially bad behavior and if it is really all that bad it should be fixed everywhere. You can either see this as an opportunistic use of the library or an abuse of the library. You are correct that there would be a problem if the arg was actually used in the SUT.

irahopkinson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/919)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 780 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/919#-MR5bCGP2L2P5PFXsXy8:-MRJ4gZN7ioUU1q9wRmq:b-5jz3xp) ([raw file](https://github.com/sillsdev/web-xforge/blob/98a41f7f80790a8b8ef518f2c21c544332a70db8/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L780)):*

Oh, I'm not saying we don't see eye to eye. I'm saying I copied previous potentially bad behavior and if it is really all that bad it should be fixed everywhere. You can either see this as an opportunistic use of the library or an abuse of the library. You are correct that there would be a problem if the arg was actually used in the SUT.
</blockquote></details>

@jasonleenaylor Err... I'm no C# expert but your response seems to confirm what @marksvc is saying since this line is the SUT. I'm not seeing any previous bad behavior except here. The documentation link above clearly says don't do what you have done here. If you move this test to the correct test file then you have the environment to generate a secret to use. Since I should have added the test in the first place, would you like me to fix this up?

jasonleenaylor (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/919)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 780 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/919#-MR5bCGP2L2P5PFXsXy8:-MRVoXYX8s8-tYXoNEk-:b-sxz2vw) ([raw file](https://github.com/sillsdev/web-xforge/blob/98a41f7f80790a8b8ef518f2c21c544332a70db8/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L780)):*

@jasonleenaylor Err... I'm no C# expert but your response seems to confirm what @marksvc is saying since this line is the SUT. I'm not seeing any previous bad behavior except here. The documentation link above clearly says don't do what you have done here. If you move this test to the correct test file then you have the environment to generate a secret to use. Since I should have added the test in the first place, would you like me to fix this up?
</blockquote></details>

@irahopkinson My next push will move this test to another class, but my point was that the argument passed to this method is not used by any of the code which is being tested.  It is passed to the method but not used in the code which was changed.

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/919)* status: :shipit: complete! all files reviewed, all discussions resolved

irahopkinson:
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 780 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/919#-MR5bCGP2L2P5PFXsXy8:-MRXhDT12r9S_3-WgDvl:bowgzl3) ([raw file](https://github.com/sillsdev/web-xforge/blob/98a41f7f80790a8b8ef518f2c21c544332a70db8/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L780)):*

@irahopkinson My next push will move this test to another class, but my point was that the argument passed to this method is not used by any of the code which is being tested.  It is passed to the method but not used in the code which was changed.
</blockquote></details>

Yes, you are right, I see what you mean, its not used.

PR #932, opened 2021-01-27
SF-1178 Access source project only if user has permission

RaymondLuong3:
* Add an endpoint to return a user's role on a project

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/932)

### DIFF
--- a/src/RealtimeServer/common/models/project-role.ts
+++ b/src/RealtimeServer/common/models/project-role.ts
@@ -1,0 +1,3 @@
+export enum ProjectRole {
+  None = 'none'
+}
--- a/src/RealtimeServer/scriptureforge/models/sf-project-role.ts
+++ b/src/RealtimeServer/scriptureforge/models/sf-project-role.ts
@@ -1,3 +1,15 @@
+export function hasParatextRole(role: string | undefined): boolean {
+  switch (role) {
+    case SFProjectRole.ParatextAdministrator:
+    case SFProjectRole.ParatextTranslator:
+    case SFProjectRole.ParatextConsultant:
+    case SFProjectRole.ParatextObserver:
+      return true;
+    default:
+      return false;
+  }
+}
+
 export enum SFProjectRole {
   ParatextAdministrator = 'pt_administrator',
   ParatextTranslator = 'pt_translator',
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -740,6 +740,7 @@ describe('EditorComponent', () => {
 
     it('user has no resource access', fakeAsync(() => {
       const env = new TestEnvironment();
+      when(mockedSFProjectService.onlineGetProjectRole('resource01')).thenResolve('None');
       env.setupProject({
         translateConfig: {
           translationSuggestionsEnabled: true,
@@ -758,6 +759,8 @@ describe('EditorComponent', () => {
       env.setProjectUserConfig();
       env.updateParams({ projectId: 'project01', bookId: 'ACT' });
       env.wait();
+      verify(mockedSFProjectService.onlineGetProjectRole('resource01')).once();
+      verify(mockedSFProjectService.get('resource01')).never();
       expect(env.bookName).toEqual('Acts');
       expect(env.component.chapter).toBe(1);
       expect(env.component.sourceLabel).toEqual('SRC');
@@ -955,6 +958,7 @@ describe('EditorComponent', () => {
 
     it('user has no resource access', fakeAsync(() => {
       const env = new TestEnvironment();
+      when(mockedSFProjectService.onlineGetProjectRole('resource01')).thenResolve('None');
       env.setupProject({
         translateConfig: {
           translationSuggestionsEnabled: false,
@@ -973,6 +977,8 @@ describe('EditorComponent', () => {
       env.setProjectUserConfig();
       env.updateParams({ projectId: 'project01', bookId: 'ACT' });
       env.wait();
+      verify(mockedSFProjectService.onlineGetProjectRole('resource01')).once();
+      verify(mockedSFProjectService.get('resource01')).never();
       expect(env.bookName).toEqual('Acts');
       expect(env.component.chapter).toBe(1);
       expect(env.component.sourceLabel).toEqual('SRC');
@@ -1084,6 +1090,11 @@ class TestEnvironment {
 
   lastApprovedPrefix: string[] = [];
 
+  private userRolesOnProject = {
+    user01: SFProjectRole.ParatextTranslator,
+    user02: SFProjectRole.ParatextConsultant,
+    user03: SFProjectRole.ParatextTranslator
+  };
   private readonly realtimeService: TestRealtimeService = TestBed.inject<TestRealtimeService>(TestRealtimeService);
   private readonly params$: BehaviorSubject<Params>;
   private trainingProgress$ = new Subject<ProgressStatus>();
@@ -1093,11 +1104,7 @@ class TestEnvironment {
     paratextId: 'target01',
     shortName: 'TRG',
     isRightToLeft: false,
-    userRoles: {
-      user01: SFProjectRole.ParatextTranslator,
-      user02: SFProjectRole.ParatextConsultant,
-      user03: SFProjectRole.ParatextTranslator
-    },
+    userRoles: this.userRolesOnProject,
     writingSystem: { tag: 'qaa' },
     translateConfig: {
       translationSuggestionsEnabled: true,
@@ -1293,6 +1300,8 @@ class TestEnvironment {
   setCurrentUser(userId: string): void {
     when(mockedUserService.currentUserId).thenReturn(userId);
     when(mockedUserService.getCurrentUser()).thenCall(() => this.realtimeService.subscribe(UserDoc.COLLECTION, userId));
+    when(mockedSFProjectService.onlineGetProjectRole('project01')).thenResolve(this.userRolesOnProject[userId]);
+    when(mockedSFProjectService.onlineGetProjectRole('project02')).thenResolve(this.userRolesOnProject[userId]);
   }
 
   setupUsers(): void {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
@@ -17,6 +17,7 @@ import Quill, { DeltaStatic, RangeStatic } from 'quill';
 import { Operation } from 'realtime-server/lib/common/models/project-rights';
 import { User } from 'realtime-server/lib/common/models/user';
 import { SFProjectDomain, SF_PROJECT_RIGHTS } from 'realtime-server/lib/scriptureforge/models/sf-project-rights';
+import { hasParatextRole } from 'realtime-server/lib/scriptureforge/models/sf-project-role';
 import { TextType } from 'realtime-server/lib/scriptureforge/models/text-data';
 import { TextInfo } from 'realtime-server/lib/scriptureforge/models/text-info';
 import { TextInfoPermission } from 'realtime-server/lib/scriptureforge/models/text-info-permission';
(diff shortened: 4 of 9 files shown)


marksvc:
:eyes:

marksvc (changes requested):
Thanks for your work on this. I have some quibbles which I have commented about below.

> Reviewed 7 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/932#-:-MS54fiN7OiwuTTTRwmo:b7e5cmu)* status: all files reviewed, 5 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge/Services/ProjectService.cs, line 81 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/932#-MS5-pXe-faQxJswUwZp:-MS5-pXf0MUZG8GHs1fk:b-a99b6j) ([raw file](https://github.com/sillsdev/web-xforge/blob/89802ca2c9dde4248d0eee0a89afa0d38ba0df62/src/SIL.XForge/Services/ProjectService.cs#L81)):*
> ```
> var attempt 
> ```
Let's use type `Attempt<string>` instead, to conform to our usually-dont-use-var style.
___
*[src/SIL.XForge/Services/ProjectService.cs, line 82 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/932#-MS52Z0-BnnvIw9ev1Be:-MS52Z0-BnnvIw9ev1Bf:bv5gpf6) ([raw file](https://github.com/sillsdev/web-xforge/blob/89802ca2c9dde4248d0eee0a89afa0d38ba0df62/src/SIL.XForge/Services/ProjectService.cs#L82)):*
> ```Smalltalk
>             TModel project = await GetProjectAsync(projectId);
>             var attempt = await TryGetProjectRoleAsync(project, curUserId);
>             return attempt.Result;
> ```

I see that in class ProjectRole.cs, we define a `None` (as well as in sf-project-role.ts). What do you think about not returning null, but returning ProjectRole.None if we find that the user doesn't have a role on the project?

Hmm, I see tho that in the spec tests, you set the mock up to return 'None' sometimes instead of null. Do we sometimes return "None" from this method and sometimes null? Or what's going on?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 294 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/932#-MS53DWu88kHNgYwISOB:-MS53DWu88kHNgYwISOC:bamfgw4) ([raw file](https://github.com/sillsdev/web-xforge/blob/89802ca2c9dde4248d0eee0a89afa0d38ba0df62/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L294)):*
> ```
> : string | undefined 
> ```
(Based on another comment regarding using 'None') So we could condense this type to just `string`.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts, line 77 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/932#-MS50BJX9iLu9CF6CsvN:-MS50BJX9iLu9CF6CsvO:bi2oksn) ([raw file](https://github.com/sillsdev/web-xforge/blob/89802ca2c9dde4248d0eee0a89afa0d38ba0df62/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts#L77)):*
> ```XML
>   }
> 
>   onlineGetProjectRole(id: string): Promise<string | undefined> {
> ```

I comment elsewhere about returning 'None' instead of null. If we do that, then we can hopefully change this return type to just `Promise<string>`.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts, line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/932#-MS50M0-0zjpu21j3HNj:-MS50M0-0zjpu21j3HNk:bv6ej5l) ([raw file](https://github.com/sillsdev/web-xforge/blob/89802ca2c9dde4248d0eee0a89afa0d38ba0df62/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts#L78)):*
> ```
> 'projectRole'
> ```
I notice that the API has some inconsistent naming for the introduced endpoint.

project.service.ts exposes `onlineGetProjectRole`.
SFProjectsRpcController.cs exposes `ProjectRole`.
IProjectService provides `RoleOnProjectAsync`.
Could we make these all use a common base name  so it's easier to quickly jump to some back-end functionality when looking at a call to it on the front-end? For example, maybe they could be something like the set of:

```
onlineGetProjectRole()
GetProjectRole()
GetProjectRoleAsync()
```

RaymondLuong3 (commented):
Thanks for the review.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/932#-:-MS5BlPf3-fDp97itXy7:b-9qdoyq)* status: 3 of 8 files reviewed, 5 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge/Services/ProjectService.cs, line 81 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/932#-MS5-pXe-faQxJswUwZp:-MS5Bo5C17wjjj6tYGg4:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/89802ca2c9dde4248d0eee0a89afa0d38ba0df62/src/SIL.XForge/Services/ProjectService.cs#L81)):*

> ```
> var attempt 
> ```
Let's use type `Attempt<string>` instead, to conform to our usually-dont-use-var style.

</blockquote></details>

Done.
___
*[src/SIL.XForge/Services/ProjectService.cs, line 82 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/932#-MS52Z0-BnnvIw9ev1Be:-MS5Bpm2BFi1eJZPD7mk:b-6pcbn0) ([raw file](https://github.com/sillsdev/web-xforge/blob/89802ca2c9dde4248d0eee0a89afa0d38ba0df62/src/SIL.XForge/Services/ProjectService.cs#L82)):*

I see that in class ProjectRole.cs, we define a `None` (as well as in sf-project-role.ts). What do you think about not returning null, but returning ProjectRole.None if we find that the user doesn't have a role on the project?

Hmm, I see tho that in the spec tests, you set the mock up to return 'None' sometimes instead of null. Do we sometimes return "None" from this method and sometimes null? Or what's going on?
</blockquote></details>

I'm happy with not returning null. It returned none now if the user does not have a role.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 294 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/932#-MS53DWu88kHNgYwISOB:-MS5CW9Q5dnUecm92Z_5:bv5mqc2) ([raw file](https://github.com/sillsdev/web-xforge/blob/89802ca2c9dde4248d0eee0a89afa0d38ba0df62/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L294)):*

> ```
> : string | undefined 
> ```
(Based on another comment regarding using 'None') So we could condense this type to just `string`.
</blockquote></details>

Unfortunately there is an error in the editor if we limit the type to string only.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts, line 77 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/932#-MS50BJX9iLu9CF6CsvN:-MS5Cj1J6jYLzl55iWs4:bvgz6vt) ([raw file](https://github.com/sillsdev/web-xforge/blob/89802ca2c9dde4248d0eee0a89afa0d38ba0df62/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts#L77)):*

I comment elsewhere about returning 'None' instead of null. If we do that, then we can hopefully change this return type to just `Promise<string>`.
</blockquote></details>

onlineInvoke will return a `string` or `undefined` according to the my editor. I don't think this change can be made.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts, line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/932#-MS50M0-0zjpu21j3HNj:-MS5CwyC5WdDFQhXBJOm:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/89802ca2c9dde4248d0eee0a89afa0d38ba0df62/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts#L78)):*

> ```
> 'projectRole'
> ```
I notice that the API has some inconsistent naming for the introduced endpoint.

project.service.ts exposes `onlineGetProjectRole`.
SFProjectsRpcController.cs exposes `ProjectRole`.
IProjectService provides `RoleOnProjectAsync`.
Could we make these all use a common base name  so it's easier to quickly jump to some back-end functionality when looking at a call to it on the front-end? For example, maybe they could be something like the set of:

```
onlineGetProjectRole()
GetProjectRole()
GetProjectRoleAsync()
```


</blockquote></details>

Done.

marksvc (changes requested):
> Reviewed 5 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/932)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts, line 77 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/932#-MS50BJX9iLu9CF6CsvN:-MS9Xbh11hkoZ6Zvxa-A:b-zaubtf) ([raw file](https://github.com/sillsdev/web-xforge/blob/89802ca2c9dde4248d0eee0a89afa0d38ba0df62/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts#L77)):*

onlineInvoke will return a `string` or `undefined` according to the my editor. I don't think this change can be made.
</blockquote></details>

Right, okay. What if we change this method to do something like

```cs
return this.onlineInvoke<string>('getProjectRole', { projectId: id }) ?? SFProjectRole.None;
```

I think that would let us always return Promise<string>.

marksvc (approved):
Thanks for your work on this.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/932#-:-MSE0t-L3hd2frRU_LXQ:b7tt6m2)* status: :shipit: complete! all files reviewed, all discussions resolved

irahopkinson:
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 295 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/932#-MS7-Je10mPrYbl3uTqO:-MS7-Je2BiGTfo_7tq7g:bz7dhq9) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ddfbf39bc0581c565ed1dd85629fb927225c22d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L295)):*
> ```
> await this.projectService.get(sourceId)
> ```
Is there some reason why this call doesn't just return `undefined` if the user doesn't have a source role rather than making 2 calls to the backend? Also now with book/chapter permissions it might be we have to check that rather than their role? Since previously your role gave you permission whereas now it is deermined properly by your book/chapter permissions.

RaymondLuong3:
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 295 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/932#-MS7-Je10mPrYbl3uTqO:-MSTGbFoDAhPokVVXyBa:b-b1gn6m) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ddfbf39bc0581c565ed1dd85629fb927225c22d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L295)):*

> ```
> await this.projectService.get(sourceId)
> ```
Is there some reason why this call doesn't just return `undefined` if the user doesn't have a source role rather than making 2 calls to the backend? Also now with book/chapter permissions it might be we have to check that rather than their role? Since previously your role gave you permission whereas now it is deermined properly by your book/chapter permissions.
</blockquote></details>

Right, I hadn't considered returning `undefined` as an option. It seems that our system was setup assuming that `projectService.get()` would only be used if it was determined that a user has access (that's why it throws an error if a user did not). We may need to review our permission structures.
I chose to check the role on projects because our realtime server uses that information specifically to determine if a user has access to a project. Again this may require that we review our permission structure.

PR #934, opened 2021-01-29
SF-1181 Refresh access token if needed before fetching resources

Nateowami:
- On the settings and connect project pages, we send two HTTP requests, one for projects and one for resources.
- Both requests cause the back end to send HTTP requests to Paratext.
- The access token for Paratext expires after 20 minutes and needs to be refreshed if it is expired.
- Previously only the projects endpoint refreshed the token (the request for resources did not, so if the token was expired, that request would fail).
- Now both requests cause the token to be refreshed, but if a refresh is in progress that the other request started, then a new refresh is not started (otherwise it would result in an error from Paratext).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)

### DIFF
--- a/src/SIL.XForge.Scripture/Controllers/ParatextController.cs
+++ b/src/SIL.XForge.Scripture/Controllers/ParatextController.cs
@@ -61,13 +61,9 @@ namespace SIL.XForge.Scripture.Controllers
         [HttpGet("resources")]
         public async Task<ActionResult<Dictionary<string, string>>> ResourcesAsync()
         {
-            Attempt<UserSecret> attempt = await _userSecrets.TryGetAsync(_userAccessor.UserId);
-            if (!attempt.TryResult(out UserSecret userSecret))
-                return NoContent();
-
             try
             {
-                var resources = _paratextService.GetResources(userSecret);
+                var resources = await _paratextService.GetResourcesAsync(_userAccessor.UserId);
                 return Ok(resources.ToDictionary(r => r.ParatextId, r => r.Name));
             }
             catch (SecurityException)
--- a/src/SIL.XForge.Scripture/Services/IParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/IParatextService.cs
@@ -16,8 +16,8 @@ namespace SIL.XForge.Scripture.Services
         Task<IReadOnlyDictionary<string, string>> GetProjectRolesAsync(UserSecret userSecret, string projectId);
         bool IsProjectLanguageRightToLeft(UserSecret userSecret, string ptProjectId);
 
-        IReadOnlyList<ParatextResource> GetResources(UserSecret userSecret);
-        Task<string> GetResourcePermissionAsync(UserSecret userSecret, string paratextId, string userId);
+        Task<IReadOnlyList<ParatextResource>> GetResourcesAsync(string userId);
+        Task<string> GetResourcePermissionAsync(string paratextId, string userId);
         Task<IReadOnlyDictionary<string, string>> GetParatextUsernameMappingAsync(UserSecret userSecret,
             string paratextId);
         Task<Dictionary<string, string>> GetPermissionsAsync(UserSecret userSecret, SFProject project,
--- a/src/SIL.XForge.Scripture/Services/JwtInternetSharedRepositorySource.cs
+++ b/src/SIL.XForge.Scripture/Services/JwtInternetSharedRepositorySource.cs
@@ -3,7 +3,6 @@ using System.Collections.Generic;
 using System.Linq;
 using Newtonsoft.Json.Linq;
 using Paratext;
-using Paratext.Data;
 using Paratext.Data.Repository;
 using Paratext.Data.Users;
 using Paratext.Data.RegistryServerAccess;
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -1,16 +1,16 @@
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.IdentityModel.Tokens.Jwt;
 using System.IO;
 using System.Linq;
-using System.Net;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Reflection;
 using System.Runtime.InteropServices;
-using System.Security;
 using System.Security.Claims;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using System.Xml;
 using System.Xml.XPath;
@@ -62,6 +62,8 @@ namespace SIL.XForge.Scripture.Services
         private string _sendReceiveServerUri = InternetAccess.uriProduction;
         private readonly IInternetSharedRepositorySourceProvider _internetSharedRepositorySourceProvider;
         private readonly ISFRestClientFactory _restClientFactory;
+        /// <summary> Map user IDs to semaphores </summary>
+        private readonly ConcurrentDictionary<string, SemaphoreSlim> _tokenRefreshSemaphores = new ConcurrentDictionary<string, SemaphoreSlim>();
 
         public ParatextService(IWebHostEnvironment env, IOptions<ParatextOptions> paratextOptions,
             IRepository<UserSecret> userSecretRepository, IRealtimeService realtimeService,
@@ -155,7 +157,7 @@ namespace SIL.XForge.Scripture.Services
         {
             if (userSecret == null || ptTargetId == null) { throw new ArgumentNullException(); }
 
-            IInternetSharedRepositorySource source = await GetInternetSharedRepositorySource(userSecret);
+            IInternetSharedRepositorySource source = await GetInternetSharedRepositorySource(userSecret.Id);
             IEnumerable<SharedRepository> repositories = source.GetRepositories();
             IEnumerable<ProjectMetadata> projectsMetadata = source.GetProjectsMetaData();
             IEnumerable<string> projectGuids = projectsMetadata.Select(pmd => pmd.ProjectGuid.Id);
@@ -164,7 +166,7 @@ namespace SIL.XForge.Scripture.Services
             if (!projectGuids.Contains(ptTargetId))
             {
                 // See if this is a resource
-                IReadOnlyList<ParatextResource> resources = this.GetResourcesInternal(userSecret, true);
+                IReadOnlyList<ParatextResource> resources = await this.GetResourcesInternalAsync(userSecret.Id, true);
                 ParatextResource resource = resources.SingleOrDefault(r => r.ParatextId == ptTargetId);
                 if (resource != null)
                 {
@@ -208,7 +210,7 @@ namespace SIL.XForge.Scripture.Services
         /// <summary> Get Paratext projects that a user has access to. </summary>
         public async Task<IReadOnlyList<ParatextProject>> GetProjectsAsync(UserSecret userSecret)
         {
-            IInternetSharedRepositorySource ptRepoSource = await GetInternetSharedRepositorySource(userSecret);
+            IInternetSharedRepositorySource ptRepoSource = await GetInternetSharedRepositorySource(userSecret.Id);
             List<SharedRepository> remotePtProjects = ptRepoSource.GetRepositories().ToList();
             List<ProjectMetadata> projectMetadata = ptRepoSource.GetProjectsMetaData().ToList();
 
@@ -219,9 +221,9 @@ namespace SIL.XForge.Scripture.Services
         }
 
         /// <summary>Get Paratext resources that a user has access to. </summary>
-        public IReadOnlyList<ParatextResource> GetResources(UserSecret userSecret)
+        public async Task<IReadOnlyList<ParatextResource>> GetResourcesAsync(string userId)
         {
-            return this.GetResourcesInternal(userSecret, false);
+            return await this.GetResourcesInternalAsync(userId, false);
         }
 
         public async Task<Attempt<string>> TryGetProjectRoleAsync(UserSecret userSecret, string paratextId)
@@ -234,7 +236,7 @@ namespace SIL.XForge.Scripture.Services
                 Claim subClaim = accessToken.Claims.FirstOrDefault(c => c.Type == JwtClaimTypes.Subject);
                 // Paratext RegistryServer has methods to do this, but it is unreliable to use it in a multi-user
                 // environment so instead we call the registry API.
-                string response = await CallApiAsync(_registryClient, userSecret, HttpMethod.Get,
+                string response = await CallApiAsync(userSecret, HttpMethod.Get,
                     $"projects/{paratextId}/members/{subClaim.Value}");
                 var memberObj = JObject.Parse(response);
                 return Attempt.Success((string)memberObj["role"]);
@@ -254,7 +256,6 @@ namespace SIL.XForge.Scripture.Services
         /// <summary>
         /// Gets the permission a user has to access a resource.
         /// </summary>
-        /// <param name="userSecret">The user secret.</param>
         /// <param name="paratextId">The paratext resource identifier.</param>
         /// <param name="userId">The user identifier.</param>
         /// <returns>
@@ -263,52 +264,26 @@ namespace SIL.XForge.Scripture.Services
         /// <remarks>
         /// See <see cref="TextInfoPermission" /> for permission values.
         /// </remarks>
-        public async Task<string> GetResourcePermissionAsync(UserSecret userSecret, string paratextId, string userId)
+        public async Task<string> GetResourcePermissionAsync(string paratextId, string userId)
         {
             // See if the source is a resource
-            if (paratextId.Length == SFInstallableDblResource.ResourceIdentifierLength)
-            {
-                // The resource id is a 41 character project id truncated to 16 characters
-                UserSecret thisUserSecret;
-                if (userId == userSecret.Id)
-                {
-                    thisUserSecret = userSecret;
-                }
-                else
-                {
-                    // Get the user secret
-                    Attempt<UserSecret> userSecretAttempt = await this._userSecretRepository.TryGetAsync(userId);
-                    if (!userSecretAttempt.TryResult(out thisUserSecret))
-                    {
-                        thisUserSecret = null;
-                    }
-                }
-
-                bool canRead = false;
-                if (thisUserSecret != null)
-                {
-                    if (!(thisUserSecret.ParatextTokens?.ValidateLifetime() ?? false))
-                    {
-                        await this.RefreshAccessTokenAsync(thisUserSecret);
-                    }
-
-                    canRead = SFInstallableDblResource.CheckResourcePermission(
-                        paratextId,
-                        thisUserSecret,
-                        this._paratextOptions.Value,
-                        this._restClientFactory,
-                        this._fileSystemService,
-                        this._jwtTokenHelper,
-                        _exceptionHandler,
-                        this._dblServerUri);
-                }
-
-                return canRead ? TextInfoPermission.Read : TextInfoPermission.None;
-            }
-            else
+            if (paratextId.Length != SFInstallableDblResource.ResourceIdentifierLength)
             {
                 // Default to no permissions for projects used as sources
                 return TextInfoPermission.None;
+            }
+            using (ParatextAccessLock accessLock = await GetParatextAccessLock(userId))
+            {
+                bool canRead = SFInstallableDblResource.CheckResourcePermission(
+                        paratextId,
+                        accessLock.UserSecret,
+                        _paratextOptions.Value,
+                        _restClientFactory,
+                        _fileSystemService,
+                        _jwtTokenHelper,
+                        _exceptionHandler,
+                        _dblServerUri);
+                return canRead ? TextInfoPermission.Read : TextInfoPermission.None;
             }
         }
 
@@ -331,7 +306,7 @@ namespace SIL.XForge.Scripture.Services
             }
 
             // Get the mapping for paratext users ids to usernames from the registry
-            string response = await CallApiAsync(_registryClient, userSecret, HttpMethod.Get,
+            string response = await CallApiAsync(userSecret, HttpMethod.Get,
                 $"projects/{paratextId}/members");
             Dictionary<string, string> paratextMapping = JArray.Parse(response).OfType<JObject>()
                 .Where(m => !string.IsNullOrEmpty((string)m["userId"])
@@ -369,7 +344,7 @@ namespace SIL.XForge.Scripture.Services
             {
                 foreach (string uid in project.UserRoles.Keys)
                 {
-                    permissions.Add(uid, await this.GetResourcePermissionAsync(userSecret, project.ParatextId, uid));
+                    permissions.Add(uid, await this.GetResourcePermissionAsync(project.ParatextId, uid));
                 }
             }
             else
@@ -446,7 +421,7 @@ namespace SIL.XForge.Scripture.Services
             {
                 // Paratext RegistryServer has methods to do this, but it is unreliable to use it in a multi-user
                 // environment so instead we call the registry API.
-                string response = await CallApiAsync(_registryClient, userSecret, HttpMethod.Get,
+                string response = await CallApiAsync(userSecret, HttpMethod.Get,
                     $"projects/{projectId}/members");
                 var members = JArray.Parse(response);
                 return members.OfType<JObject>()
@@ -781,68 +756,40 @@ namespace SIL.XForge.Scripture.Services
             HgWrapper.Update(clonePath);
         }
 
-        private async Task RefreshAccessTokenAsync(UserSecret userSecret)
-        {
-            ParatextOptions options = _paratextOptions.Value;
-
-            userSecret.ParatextTokens = await _jwtTokenHelper.RefreshAccessTokenAsync(options,
-                userSecret.ParatextTokens, _registryClient);
-
-            await _userSecretRepository.UpdateAsync(userSecret, b =>
-                b.Set(u => u.ParatextTokens, userSecret.ParatextTokens));
-        }
-
-        private async Task<string> CallApiAsync(HttpClient client, UserSecret userSecret, HttpMethod method,
+        private async Task<string> CallApiAsync(UserSecret userSecret, HttpMethod method,
             string url, string content = null)
         {
-            if (userSecret == null)
-                throw new SecurityException("The current user is not signed into Paratext.");
-
-            bool expired = !userSecret.ParatextTokens.ValidateLifetime();
-            bool refreshed = false;
-            while (!refreshed)
+            using (ParatextAccessLock accessLock = await GetParatextAccessLock(userSecret.Id))
+            using (var request = new HttpRequestMessage(method, $"api8/{url}"))
             {
-                if (expired)
+                request.Headers.Authorization = new AuthenticationHeaderValue("Bearer",
+                    accessLock.UserSecret.ParatextTokens.AccessToken);
+                if (content != null)
                 {
-                    await RefreshAccessTokenAsync(userSecret);
-                    refreshed = true;
+                    request.Content = new StringContent(content);
                 }
-
-                using (var request = new HttpRequestMessage(method, $"api8/{url}"))
+                HttpResponseMessage response = await _registryClient.SendAsync(request);
+                if (response.IsSuccessStatusCode)
                 {
-                    request.Headers.Authorization = new AuthenticationHeaderValue("Bearer",
-                        userSecret.ParatextTokens.AccessToken);
-                    if (content != null)
-                        request.Content = new StringContent(content);
-                    HttpResponseMessage response = await client.SendAsync(request);
-                    if (response.IsSuccessStatusCode)
-                    {
-                        return await response.Content.ReadAsStringAsync();
-                    }
-                    else if (response.StatusCode == HttpStatusCode.Unauthorized)
-                    {
-                        expired = true;
-                    }
-                    else
-                    {
-                        throw new HttpRequestException(await ExceptionHandler.CreateHttpRequestErrorMessage(response));
-                    }
+                    return await response.Content.ReadAsStringAsync();
+                }
+                else
+                {
+                    throw new HttpRequestException(await ExceptionHandler.CreateHttpRequestErrorMessage(response));
                 }
             }
-
-            throw new SecurityException("The current user's Paratext access token is invalid.");
         }
 
         /// <summary>
         /// Get access to a source for PT project repositories, based on user secret.
         ///</summary>
-        private async Task<IInternetSharedRepositorySource> GetInternetSharedRepositorySource(UserSecret userSecret)
+        private async Task<IInternetSharedRepositorySource> GetInternetSharedRepositorySource(string userId)
         {
-            if (userSecret == null) throw new ArgumentNullException();
-            await RefreshAccessTokenAsync(userSecret);
-            IInternetSharedRepositorySource source = _internetSharedRepositorySourceProvider.GetSource(userSecret,
-                _sendReceiveServerUri, _registryServerUri, _applicationProductVersion);
-            return source;
+            using (ParatextAccessLock accessLock = await GetParatextAccessLock(userId))
+            {
+                return _internetSharedRepositorySourceProvider.GetSource(accessLock.UserSecret,
+                        _sendReceiveServerUri, _registryServerUri, _applicationProductVersion);
+            }
         }
 
         /// <summary>
@@ -853,16 +800,20 @@ namespace SIL.XForge.Scripture.Services
         /// <returns>
         /// The available resources.
         /// </returns>
-        private IReadOnlyList<ParatextResource> GetResourcesInternal(UserSecret userSecret, bool includeInstallableResource)
+        private async Task<IReadOnlyList<ParatextResource>> GetResourcesInternalAsync(string userId, bool includeInstallableResource)
         {
-            IEnumerable<SFInstallableDblResource> resources = SFInstallableDblResource.GetInstallableDblResources(
-                userSecret,
-                this._paratextOptions.Value,
-                this._restClientFactory,
-                this._fileSystemService,
-                this._jwtTokenHelper,
-                _exceptionHandler,
-                this._dblServerUri);
+            IEnumerable<SFInstallableDblResource> resources;
+            using (ParatextAccessLock accessLock = await GetParatextAccessLock(userId))
+            {
+                resources = SFInstallableDblResource.GetInstallableDblResources(
+                    accessLock.UserSecret,
+                    this._paratextOptions.Value,
+                    this._restClientFactory,
+                    this._fileSystemService,
+                    this._jwtTokenHelper,
+                    _exceptionHandler,
+                    this._dblServerUri);
+            }
             IReadOnlyDictionary<string, int> resourceRevisions =
                 SFInstallableDblResource.GetInstalledResourceRevisions();
             return resources.OrderBy(r => r.FullName).Select(r => new ParatextResource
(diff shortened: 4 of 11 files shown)


RaymondLuong3 (approved):
Good job!. A few non-blocking comments.

> Reviewed 9 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934#-:-MSTVZjZBK34DCQz3X3Z:b-6gtnjq)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 21 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSTS_hX74PtxBf-JUe2:-MSTS_hY30qS_MY5LIBf:bazrabv) ([raw file](https://github.com/sillsdev/web-xforge/blob/5adb7a68b1f4cc8d797e7888953398db88fea867/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L21)):*
> ```
> private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1, 1);
> ```
Could a more useful name be used? PTTokenRefreshLock or something similar?
___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 22 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSTSQ5h4jDWyJyilmxE:-MSTSQ5i1AEb0jdV8NV0:b-borgfc) ([raw file](https://github.com/sillsdev/web-xforge/blob/5adb7a68b1f4cc8d797e7888953398db88fea867/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L22)):*
> ```
> private readonly ConcurrentDictionary<string, Task<Tokens>> _userRefreshTasks = new ConcurrentDictionary<string, Task<Tokens>>();
> ```
Maybe `_tokenRefreshTasks` is a better name?
___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSTTUejC6xhPzWb_GOM:-MSTTUejC6xhPzWb_GON:b-3i1bb6) ([raw file](https://github.com/sillsdev/web-xforge/blob/5adb7a68b1f4cc8d797e7888953398db88fea867/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L92)):*

> ```
>             finally
>             {
>                 this._semaphore.Release();
>             }
> ```
</details>

I would have expected `Release()` to be called only after the task is completed. But I guess it is safe because the concurrent dictionary would already contain the refresh task, so the race condition is still avoided.

marksvc:
:eyes:

marksvc (dismissed):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: all files reviewed, 7 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 22 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSTSQ5h4jDWyJyilmxE:-MSThX2XCUepSK6EQt-E:b32v57) ([raw file](https://github.com/sillsdev/web-xforge/blob/5adb7a68b1f4cc8d797e7888953398db88fea867/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L22)):*

> ```
> private readonly ConcurrentDictionary<string, Task<Tokens>> _userRefreshTasks = new ConcurrentDictionary<string, Task<Tokens>>();
> ```
Maybe `_tokenRefreshTasks` is a better name?
</blockquote></details>

+1
___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSTfj1t7iBcLVbH1ylB:-MSTfj1t7iBcLVbH1ylC:b-npa5um) ([raw file](https://github.com/sillsdev/web-xforge/blob/5adb7a68b1f4cc8d797e7888953398db88fea867/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L54)):*
> ```Smalltalk
>             HttpClient client)
>         {
>             if (userSecret.ParatextTokens.ValidateLifetime())
> ```

So, an interesting aspect of how this situation operates is that calling `RefreshAccessTokenAsync()` doesn't actually refresh our access token. It does get a refreshed access token and return it to the caller, but we (1) don't update the userSecret.ParatextTokens with the refreshed tokens (here), (2) nor do we write our refreshed tokens to the DB (here). 

But the successful usage of this method happens in ParatextService.cs in a method of the same name where we do (1) and (2) above. 

My recollection of our app tells me that if data in the DB is updated, our model object won't necessarily contain that updated data. So if we have 5 threads that all fetch user Bob's UserSecret from the DB, and one of those threads gets a refreshed token, updates the userSecret model object in that thread, and updates the DB with the refreshed token, the other 4 threads will still have the old data in their UserSecret model object. 
I may be wrong about this. Have you seen in testing that the `userSecret` object in a given thread stays up-to-date with changed content in the DB?
___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 60 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSTbtSi5NqMMWlbWZMW:-MSTbtSi5NqMMWlbWZMX:bj0qak1) ([raw file](https://github.com/sillsdev/web-xforge/blob/5adb7a68b1f4cc8d797e7888953398db88fea867/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L60)):*
> ```
>             // Ensure only one refresh occurs for a given user at a time
>             await this._semaphore.WaitAsync();
> ```


Hmm. This comment on a semaphore sounds like the semaphore is making the refreshing sequential. But it's really making the task enqueing atomic such that we don't enqueue two tasks for the same user. It is true that this means we "Ensure only one refresh occurs for a given user at a time" but I think we can improve the explanation by saying something like 

// Make sure we don't create more than one refresh task for a given user at a time. 

or 

// Ensure only one refresh occurs for a given user at a time, by only adding a refresh task for a user if one isn't already running.

Hmm.  What do you think?
___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 61 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSTeJK2AmvPM5HY4_bl:-MSTeJK2AmvPM5HY4_bm:b-65u8vc) ([raw file](https://github.com/sillsdev/web-xforge/blob/5adb7a68b1f4cc8d797e7888953398db88fea867/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L61)):*
> ```Smalltalk
>             Task<Tokens> task;
>             // Ensure only one refresh occurs for a given user at a time
>             await this._semaphore.WaitAsync();
> ```

Because there is no ValidateLifetime() after the _semaphore.WaitAsync(), doesn't that mean that if 10 threads call RefreshAccessTokenAsync() for the same user, that we will still do an HTTP request to refresh the token 10 times?
___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 65 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSTeoI9-EiAwZcEcy2u:-MSTeoI9-EiAwZcEcy2v:b-7b06x9) ([raw file](https://github.com/sillsdev/web-xforge/blob/5adb7a68b1f4cc8d797e7888953398db88fea867/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L65)):*
> ```
> this.
> ```

I think in our C# our style has been to not prefix members with `this.`. There are several `this.`s in this patch to check on.
___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSTTUejC6xhPzWb_GOM:-MSThIbABE1gDnb_c5Iv:bce5ndd) ([raw file](https://github.com/sillsdev/web-xforge/blob/5adb7a68b1f4cc8d797e7888953398db88fea867/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L92)):*

> ```
>             finally
>             {
>                 this._semaphore.Release();
>             }
> ```
I would have expected `Release()` to be called only after the task is completed. But I guess it is safe because the concurrent dictionary would already contain the refresh task, so the race condition is still avoided.
</blockquote></details>

Yeah, so this mean we can _add_ tasks to the set quickly, before any of them finish, but we wont add more than one task to the task set for a given user.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: 8 of 9 files reviewed, 4 unresolved discussions (waiting on @marksvc, @Nateowami, and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 65 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSTeoI9-EiAwZcEcy2u:-MSTjQC_1UB9Haq4GPHb:b-i21bbo) ([raw file](https://github.com/sillsdev/web-xforge/blob/5adb7a68b1f4cc8d797e7888953398db88fea867/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L65)):*

> ```
> this.
> ```

I think in our C# our style has been to not prefix members with `this.`. There are several `this.`s in this patch to check on.
</blockquote></details>

Thanks. Done.
___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSTTUejC6xhPzWb_GOM:-MSTjcP3AnWKwW75-rz-:btqqu5l) ([raw file](https://github.com/sillsdev/web-xforge/blob/5adb7a68b1f4cc8d797e7888953398db88fea867/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L92)):*

Yeah, so this mean we can _add_ tasks to the set quickly, before any of them finish, but we wont add more than one task to the task set for a given user.
 
</blockquote></details>

Correct.

jasonleenaylor (commented):
> Reviewed 7 of 8 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: 12 of 13 files reviewed, 5 unresolved discussions (waiting on @marksvc, @Nateowami, and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 20 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSYxN_Z2Hx3UxXga8mA:-MSYxN_Z2Hx3UxXga8mB:b-1b4kwj) ([raw file](https://github.com/sillsdev/web-xforge/blob/077206c97c539da8223e60449507dfef15a11966/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L20)):*
> ```
> JwtTokenHelper
> ```

What is the lifecycle of the JwtTokenHelper?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: 12 of 13 files reviewed, 5 unresolved discussions (waiting on @jasonleenaylor, @marksvc, @Nateowami, and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 20 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSYxN_Z2Hx3UxXga8mA:-MSYy5jeEzg-IGvu4lya:b-ksmgjl) ([raw file](https://github.com/sillsdev/web-xforge/blob/077206c97c539da8223e60449507dfef15a11966/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L20)):*

> ```
> JwtTokenHelper
> ```

What is the lifecycle of the JwtTokenHelper?
</blockquote></details>

`SFServiceCollectionExtensions` registers it has a singleton service.

jasonleenaylor (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: 12 of 13 files reviewed, 4 unresolved discussions (waiting on @marksvc, @Nateowami, and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 20 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSYxN_Z2Hx3UxXga8mA:-MSYyMBl3CM4dT5XAhpY:b-8fx2u7) ([raw file](https://github.com/sillsdev/web-xforge/blob/077206c97c539da8223e60449507dfef15a11966/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L20)):*

`SFServiceCollectionExtensions` registers it has a singleton service.
</blockquote></details>

:+1:

jasonleenaylor (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: 12 of 13 files reviewed, 5 unresolved discussions (waiting on @marksvc, @Nateowami, and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 68 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSYyqJF61Va6xkzKxz7:-MSYyqJF61Va6xkzKxz8:bc88v6s) ([raw file](https://github.com/sillsdev/web-xforge/blob/077206c97c539da8223e60449507dfef15a11966/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L68)):*
> ```
> _tokenRefreshSemaphores.GetOrAdd(userId, (string key) => new SemaphoreSlim(1, 1));
> ```
Can't you just use `GetOrAdd(userId, new SemaphoreSlim(1,1))` instead? The docs on ConcurrentDictionary mention that code in the functions isn't guaranteed to execute atomically. So there is a very small race condition living here.

jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: 12 of 14 files reviewed, 4 unresolved discussions (waiting on @marksvc, @Nateowami, and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 68 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSYyqJF61Va6xkzKxz7:-MSZ8HBI9IbkJGy35lCB:byn7zzb) ([raw file](https://github.com/sillsdev/web-xforge/blob/077206c97c539da8223e60449507dfef15a11966/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L68)):*

Yes, we can do that (and I've changed it to that). My guess is that there wasn't an actual race condition, because most likely while the factory function might be called more than once, it would only be registered in the dictionary once. However, the only reason for a factory function would be to prevent constructing the semaphore more often than needed. But I don't have any reason to think a lambda is lighter weight than a SemaphoreSlim.
</blockquote></details>

It could be but the documentation does leave a question. What would happen when two threads reached this line and the first Add completes, but the function passed in has not executed, and then execution switches to the other thread. Since the function for creating isn't atomic with the Add what will the Get return?
If the factory function is called more than once then you have two different semaphores and the threads wouldn't block each other, anyhow, best to go with the change and avoid having to answer all these mind bending questions. :smile:

jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: all files reviewed, 4 unresolved discussions (waiting on @marksvc and @Nateowami)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: all files reviewed, 4 unresolved discussions (waiting on @marksvc and @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 68 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSYyqJF61Va6xkzKxz7:-MSZAQPsDRKUV2T4acGN:bo33h5w) ([raw file](https://github.com/sillsdev/web-xforge/blob/077206c97c539da8223e60449507dfef15a11966/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L68)):*
> What would happen when two threads reached this line and the first Add completes, but the function passed in has not executed

How could the getOrAdd method complete (it needs to return a semaphore) if it hadn't first called the factory method that creates the semaphore?

From [the docs](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd):
> Since a key/value can be inserted by another thread while valueFactory is generating a value, you cannot trust that just because valueFactory executed, its produced value will be inserted into the dictionary and returned. If you call GetOrAdd simultaneously on different threads, valueFactory may be called multiple times, but only one key/value pair will be added to the dictionary.

In other words, as I understand it, the factory method could be called multiple times, but only one of them would actually be inserted and returned to both calls of the method.

This change has caused a new alert in LGTM "Disposable 'SemaphoreSlim' is created here but is not disposed." I'm still trying to think through why (and whether just reverting it would actually get to the root).

jasonleenaylor (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: 13 of 14 files reviewed, 4 unresolved discussions (waiting on @jasonleenaylor, @marksvc, and @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 68 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSYyqJF61Va6xkzKxz7:-MSZnIopAPeiW5ZxuWOp:b-o7tkzq) ([raw file](https://github.com/sillsdev/web-xforge/blob/077206c97c539da8223e60449507dfef15a11966/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L68)):*

> What would happen when two threads reached this line and the first Add completes, but the function passed in has not executed

How could the getOrAdd method complete (it needs to return a semaphore) if it hadn't first called the factory method that creates the semaphore?

From [the docs](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentdictionary-2.getoradd):
> Since a key/value can be inserted by another thread while valueFactory is generating a value, you cannot trust that just because valueFactory executed, its produced value will be inserted into the dictionary and returned. If you call GetOrAdd simultaneously on different threads, valueFactory may be called multiple times, but only one key/value pair will be added to the dictionary.

In other words, as I understand it, the factory method could be called multiple times, but only one of them would actually be inserted and returned to both calls of the method.

This change has caused a new alert in LGTM "Disposable 'SemaphoreSlim' is created here but is not disposed." I'm still trying to think through why (and whether just reverting it would actually get to the root).
</blockquote></details>

Oops, sorry, I hadn't read far enough down the docs. So what it says is that this isn't atomic with *other* operations, but they implemented the logic for GetOrAdd to handle this situation. I withdraw my objection. You are correct, it is safe, please change it back. :smile:

jasonleenaylor (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: 13 of 14 files reviewed, 4 unresolved discussions (waiting on @jasonleenaylor, @marksvc, and @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 68 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSYyqJF61Va6xkzKxz7:-MSZnkuN-zYTN2FP7QPG:binfi6q) ([raw file](https://github.com/sillsdev/web-xforge/blob/077206c97c539da8223e60449507dfef15a11966/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L68)):*

Oops, sorry, I hadn't read far enough down the docs. So what it says is that this isn't atomic with *other* operations, but they implemented the logic for GetOrAdd to handle this situation. I withdraw my objection. You are correct, it is safe, please change it back. :smile:
</blockquote></details>

The LGTM error indicates that we technically be disposing the values of the dictionary when the JwtTokenHelper goes out of scope. Since it is a singleton and the dictionary is a member there is no practical advantage.

ddaspit (changes requested):
> Reviewed 8 of 9 files at r1, 4 of 8 files at r2, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: all files reviewed, 5 unresolved discussions (waiting on @marksvc and @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 59 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MS_yTay-kOwiWlRDROe:-MS_yTazBuJ058jxXg2F:bnu4ezv) ([raw file](https://github.com/sillsdev/web-xforge/blob/124544ad4d6ba8f853376d817f592ce8e75a799d/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L59)):*
> ```Smalltalk
>         /// cause multiple token refresh attempts for a given user.
>         /// </summary>
>         public async Task<Tokens> GetValidAccessTokenAsync(ParatextOptions options, UserSecret initialUserSecret,
> ```

Although this more recent change is better and reduces the chances of bugs, there are still potential race conditions that can occur. Here is an example:
1. Thread 1 gets a valid access token for user 1.
2. Thread 2 gets an invalid access token for user 1 (the access token has now expired).
3. Thread 2 refreshes the token for user 1.
4. Thread 1 makes a call to a PT API using the access token for user 1 and it fails because it is no longer valid.

In my experience, these kinds of race conditions will result in infrequent and inconsistent bugs that are very difficult to track down. In order to completely solve this problem, the current solution should go a bit further. In addition to synchronizing getting and updating the access token, we should also synchronize access to the validation and use of the tokens. Use the same approach implemented here, just move it up a level to the `ParatextService` and include validation and calls to the PT API in the lock. `UserSecret`s should no longer be passed into the `ParatextService` methods instead a user id should be passed in. Within a user lock, the `ParatextService` method should retrieve the `UserSecret`, check if it the token is valid, refresh the token if it is invalid, update the `UserSecret` in Mongo, and call the PT API that requires the access token.

marksvc (commented):
I approve, excepting that I agree with further enhancing it as Damien pointed out. It will be good to get this worked out in a way that is as sealed up as we can. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934#-:-MSiZ_hm22sIXX9HN0i7:bz7qsqc)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 59 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MS_yTay-kOwiWlRDROe:-MSiUIX74e7YgSpPNfVm:bkf6t0g) ([raw file](https://github.com/sillsdev/web-xforge/blob/124544ad4d6ba8f853376d817f592ce8e75a799d/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L59)):*

Although this more recent change is better and reduces the chances of bugs, there are still potential race conditions that can occur. Here is an example:
1. Thread 1 gets a valid access token for user 1.
2. Thread 2 gets an invalid access token for user 1 (the access token has now expired).
3. Thread 2 refreshes the token for user 1.
4. Thread 1 makes a call to a PT API using the access token for user 1 and it fails because it is no longer valid.

In my experience, these kinds of race conditions will result in infrequent and inconsistent bugs that are very difficult to track down. In order to completely solve this problem, the current solution should go a bit further. In addition to synchronizing getting and updating the access token, we should also synchronize access to the validation and use of the tokens. Use the same approach implemented here, just move it up a level to the `ParatextService` and include validation and calls to the PT API in the lock. `UserSecret`s should no longer be passed into the `ParatextService` methods instead a user id should be passed in. Within a user lock, the `ParatextService` method should retrieve the `UserSecret`, check if it the token is valid, refresh the token if it is invalid, update the `UserSecret` in Mongo, and call the PT API that requires the access token.
</blockquote></details>

Yeah, okay. Thanks for that. I was hoping we would keep the network request out of the lock, but I see what you mean. 
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 240 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MSiYhLQ1q-dGroWgPkO:-MSiYhLRDdNo4h2f8IRW:b-adedv2) ([raw file](https://github.com/sillsdev/web-xforge/blob/124544ad4d6ba8f853376d817f592ce8e75a799d/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L240)):*
> ```Smalltalk
>         {
>             Assert.ThrowsAsync<NullReferenceException>(async () => await env.Service.GetResourcesAsync(null));
> ```

Good job adapting this :)

Nateowami (commented):
Note: tests are broken and I still need to work on fixing them, but the other changes can be reviewed.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934#-:-MURyYi68KXSfwl9Wzgq:byfn2g0)* status: 3 of 13 files reviewed, 1 unresolved discussion (waiting on @ddaspit, @jasonleenaylor, and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 93 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MURyh6h3CoxbMgOsk6_:-MURyh6h3CoxbMgOsk6a:bpmlob4) ([raw file](https://github.com/sillsdev/web-xforge/blob/d73af9514904bad5e013675fd2ffeb59f7bc824d/src/SIL.XForge.Scripture/Services/ParatextService.cs#L93)):*
> ```Smalltalk
>                 _httpClientHandler.ServerCertificateCustomValidationCallback
>                     = HttpClientHandler.DangerousAcceptAnyServerCertificateValidator;
>                 _dblServerUri = "https://qa.thedigitalbiblelibrary.org/";
> ```

This change doesn't really belong in this PR, but it was necessary to actually make testing these changes work (since the other server is currently down).

ddaspit (changes requested):
> Reviewed 1 of 8 files at r2, 8 of 12 files at r5, 3 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/ParatextAccessLockService.cs, line 15 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MUgklAIB5ClBkZkDSs7:-MUgklAJ4N4FijT1TlkV:b-h7qopv) ([raw file](https://github.com/sillsdev/web-xforge/blob/de4da9e3bc8c4f3691290b5505c8228065fc79b2/src/SIL.XForge.Scripture/Services/ParatextAccessLockService.cs#L15)):*
> ```Smalltalk
> namespace SIL.XForge.Scripture.Services
> {
>     public class ParatextAccessLockService : IParatextAccessLockService
> ```

This service feels unnecessary to me. It will probably only be used by the `ParatextService`. I would just move this logic to the `ParatextService` class.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 232 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MUgiUwb7x-uYBoFemgo:-MUgiUwb7x-uYBoFemgp:b1g91wu) ([raw file](https://github.com/sillsdev/web-xforge/blob/de4da9e3bc8c4f3691290b5505c8228065fc79b2/src/SIL.XForge.Scripture/Services/ParatextService.cs#L232)):*
> ```Smalltalk
>         }
> 
>         public async Task<Attempt<string>> TryGetProjectRoleAsync(UserSecret userSecret, string paratextId)
> ```

This method is using the PT access token, but is not using the `ParatextAccessLockService` to get it. There are many other methods in this class that are doing the same. All of these methods have the same problem as `GetResourcesAsync` and `GetProjectsAsync` and should be fixed to work the same way. In order to avoid this problem entirely, I would update all methods in this class to not accept the `UserSecret` class as a parameter. For methods that need the access token, you would pass in the user id and get the `UserSecret` internally. For methods that only need the PT username, you could either pass in the user id and get the `UserSecret` or just pass in the PT username.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 271 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MUgiNNQ8O-yKaxaxU3F:-MUgiNNQ8O-yKaxaxU3G:bktpk0l) ([raw file](https://github.com/sillsdev/web-xforge/blob/de4da9e3bc8c4f3691290b5505c8228065fc79b2/src/SIL.XForge.Scripture/Services/ParatextService.cs#L271)):*
> ```Smalltalk
>         /// See <see cref="TextInfoPermission" /> for permission values.
>         /// </remarks>
>         public async Task<string> GetResourcePermissionAsync(UserSecret userSecret, string paratextId, string userId)
> ```

I don't think that `userSecret` is being used anymore.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/ParatextAccessLockService.cs, line 15 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MUgklAIB5ClBkZkDSs7:-MUlK5HdAMB7J4puuvEQ:bb7spe2) ([raw file](https://github.com/sillsdev/web-xforge/blob/de4da9e3bc8c4f3691290b5505c8228065fc79b2/src/SIL.XForge.Scripture/Services/ParatextAccessLockService.cs#L15)):*

This service feels unnecessary to me. It will probably only be used by the `ParatextService`. I would just move this logic to the `ParatextService` class.
</blockquote></details>

You could also keep the class, but not inject it as a service. If you go this route, I would rename it to `ParatextAccessLock` and get rid of the interface.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: all files reviewed, 3 unresolved discussions (waiting on @ddaspit and @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs, line 59 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MS_yTay-kOwiWlRDROe:-MUVZDeF-L1pys2S8biQ:bun02xs) ([raw file](https://github.com/sillsdev/web-xforge/blob/124544ad4d6ba8f853376d817f592ce8e75a799d/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L59)):*

Yeah, okay. Thanks for that. I was hoping we would keep the network request out of the lock, but I see what you mean. 
</blockquote></details>

Done. I've implemented this a little differently since there didn't appear to be a uniform way in which the API is used (some places use access tokens to communicate with Paratext via the DLL). Instead, there's a service that provides a lock on access to a given user.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 232 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MUgiUwb7x-uYBoFemgo:-MUlOCfc42GpiROhxsjH:b-9mgtb9) ([raw file](https://github.com/sillsdev/web-xforge/blob/de4da9e3bc8c4f3691290b5505c8228065fc79b2/src/SIL.XForge.Scripture/Services/ParatextService.cs#L232)):*

This method is using the PT access token, but is not using the `ParatextAccessLockService` to get it. There are many other methods in this class that are doing the same. All of these methods have the same problem as `GetResourcesAsync` and `GetProjectsAsync` and should be fixed to work the same way. In order to avoid this problem entirely, I would update all methods in this class to not accept the `UserSecret` class as a parameter. For methods that need the access token, you would pass in the user id and get the `UserSecret` internally. For methods that only need the PT username, you could either pass in the user id and get the `UserSecret` or just pass in the PT username.
</blockquote></details>

I'm not sure that this is actually the case. This method for example uses the user secret to get a claim. Without the token this method would not work. When the API is actually called it's within the lock. Now, it's possible the token is out of date when getting the claim, so perhaps part of the logic in this method should be contained within a lock, and then the API should be called. That's more complex but is doable.

I think it's probably worth changing a number of these methods to not deal with the user secret directly, but I don't think they actually have the race condition issue. I'd have to look more closely though. It's possible I just haven't seen one.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 232 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MUgiUwb7x-uYBoFemgo:-MUmLinyDDDB3J_awNeB:bc366w3) ([raw file](https://github.com/sillsdev/web-xforge/blob/de4da9e3bc8c4f3691290b5505c8228065fc79b2/src/SIL.XForge.Scripture/Services/ParatextService.cs#L232)):*

I'm not sure that this is actually the case. This method for example uses the user secret to get a claim. Without the token this method would not work. When the API is actually called it's within the lock. Now, it's possible the token is out of date when getting the claim, so perhaps part of the logic in this method should be contained within a lock, and then the API should be called. That's more complex but is doable.

I think it's probably worth changing a number of these methods to not deal with the user secret directly, but I don't think they actually have the race condition issue. I'd have to look more closely though. It's possible I just haven't seen one.
</blockquote></details>

You are correct. I didn't notice that the lock was obtained in the `CallApiAsync` method. I'm still not in love with `UserSecret`s being passed into `ParatextService`, because it is so tempting to use the access token without obtaining the lock. For example, the `SendReceiveAsync` method has a call to `GetResourcesInternal` without obtaining the lock.

ddaspit (commented):
> Reviewed 5 of 7 files at r7.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 232 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MUgiUwb7x-uYBoFemgo:-MUqTu3Z0DXNAQfR_qJw:b1m5zwk) ([raw file](https://github.com/sillsdev/web-xforge/blob/de4da9e3bc8c4f3691290b5505c8228065fc79b2/src/SIL.XForge.Scripture/Services/ParatextService.cs#L232)):*

You are correct. I didn't notice that the lock was obtained in the `CallApiAsync` method. I'm still not in love with `UserSecret`s being passed into `ParatextService`, because it is so tempting to use the access token without obtaining the lock. For example, the `SendReceiveAsync` method has a call to `GetResourcesInternal` without obtaining the lock.
</blockquote></details>

If you don't want to deal with refactoring all of the methods to not pass in a `UserSecret`, that is fine. It can always be refactored later.  Just make sure that you fix the call to `GetResourcesInternal` in `SendReceiveAsync`.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 845 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MUqTCVg1AbpeveFHH1p:-MUqTCVg1AbpeveFHH1q:b-slwd75) ([raw file](https://github.com/sillsdev/web-xforge/blob/8a09d257c89ac4e9a88b4e747c346844937cff53/src/SIL.XForge.Scripture/Services/ParatextService.cs#L845)):*
> ```Smalltalk
>         }
> 
>         private async Task<ParatextAccessLock> GetParatextAccessLock(string userId, HttpClient client)
> ```

I don't think it is necessary to pass in `HttpClient client`. It will always be `_registryClient`.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: 5 of 11 files reviewed, 3 unresolved discussions (waiting on @ddaspit)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 232 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MUgiUwb7x-uYBoFemgo:-MUtKFmq23M9kY2mTRPi:b-ft6zp5) ([raw file](https://github.com/sillsdev/web-xforge/blob/de4da9e3bc8c4f3691290b5505c8228065fc79b2/src/SIL.XForge.Scripture/Services/ParatextService.cs#L232)):*

If you don't want to deal with refactoring all of the methods to not pass in a `UserSecret`, that is fine. It can always be refactored later.  Just make sure that you fix the call to `GetResourcesInternal` in `SendReceiveAsync`.
</blockquote></details>

Yesterday after fixing the tests I started refactoring to remove uses of `UserSecret` where possible. That ended up requiring vastly more changes that I originally anticipated, so I decided to push what I had, then go back and fix `SendReceiveAsync`.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 271 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MUgiNNQ8O-yKaxaxU3F:-MUtD7Y64tCMz5e3ahrW:b-i21bbo) ([raw file](https://github.com/sillsdev/web-xforge/blob/de4da9e3bc8c4f3691290b5505c8228065fc79b2/src/SIL.XForge.Scripture/Services/ParatextService.cs#L271)):*

I don't think that `userSecret` is being used anymore.
</blockquote></details>

Thanks. Done.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 845 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/934#-MUqTCVg1AbpeveFHH1p:-MUtCux6BkmKiag371yh:b-i21bbo) ([raw file](https://github.com/sillsdev/web-xforge/blob/8a09d257c89ac4e9a88b4e747c346844937cff53/src/SIL.XForge.Scripture/Services/ParatextService.cs#L845)):*

I don't think it is necessary to pass in `HttpClient client`. It will always be `_registryClient`.
</blockquote></details>

Thanks. Done.

ddaspit (approved):
Thanks for your persistence, @nateowami.

> Reviewed 6 of 6 files at r8.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934#-:-MUwkBnl6MjvYd15-QBz:badosvi)* status: :shipit: complete! all files reviewed, all discussions resolved

irahopkinson (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/934)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #940, opened 2021-02-03
Update a number of npm dependencies

Nateowami:
Dependabot has complained about a vulnerability in the package `socket.io` and has opened #928 (Bump socket.io from 2.3.0 to 2.4.1 in /src/SIL.XForge.Scripture/ClientApp). The package is used by Karma, so realistically there's probably no actual security risk, but typically we try to update packages that are known to have vulnerabilities rather than ignoring warnings.

We can't merge PRs that Dependabot makes (they don't build on TeamCity), so this PR is intended to replace the PR Dependabot opened (Also I started work on this branch before Dependabot opened the PR. I think it waits until there's a package published that will allow it to patch the issue without altering package.json, only altering the lock file).

While I was at it I updated a number of other packages that were out of date. This does not update all dependencies to the newest version (far from it).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/940)


irahopkinson (approved):
:lgtm: with one nit. Feel free to merge yourself with or without it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/940#-:-MShZVb61FmUuzjTQeND:b-fgab2k)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/package.json, line 38 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/940#-MShZ-jzD529dXLlihW2:-MShZ-jzD529dXLlihW3:bn1f3xq) ([raw file](https://github.com/sillsdev/web-xforge/blob/d620ef34ed554caad717cc18be41d9b803326a5c/src/SIL.XForge.Scripture/ClientApp/package.json#L38)):*
> ```
> "@bugsnag/js": "^7.6.0"
> ```
nit: we should also update this in the RealtimeServer dependencies to keep them on the same version.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/940)* status: 2 of 4 files reviewed, all discussions resolved

___
*[src/SIL.XForge.Scripture/ClientApp/package.json, line 38 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/940#-MShZ-jzD529dXLlihW2:-MShiJNFB6ZO8eKjqLLe:b-pia7w) ([raw file](https://github.com/sillsdev/web-xforge/blob/d620ef34ed554caad717cc18be41d9b803326a5c/src/SIL.XForge.Scripture/ClientApp/package.json#L38)):*

> ```
> "@bugsnag/js": "^7.6.0"
> ```
nit: we should also update this in the RealtimeServer dependencies to keep them on the same version.
</blockquote></details>

Good point. Thanks for catching that. Done.

PR #943, opened 2021-02-05
SF-1162 Fetch ldml for resource if not present

marksvc:
- ParatextData ScrLanguage.LoadWsDef() can try to read an ldml file
    that isn't there. Back in ScrText.CreateLayoutEngine() we catch this
    and try to find or make it in other ways with
    ScrTextCollection.GetLanguage().
    - One possible way is via SFScrTextCollection.CreateWsDef(). The code
    being introduced here was copied (and modified) from ParatextBase.
    - The problem of an ldml file not being present when it's asked for
    can happen when using certain DBL resources, such as those that were
    made with Paratext 7, prior to ldml files being included.
    - A better solution for the problem would be to _migrate_ the
    resource to a newer version/format. But fetching from SLDR will let us
    proceed for now.
    - Rename SFScrTextCollection.cs to correspond to class name.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/943)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/SFScrTextCollection.cs
+++ b/src/SIL.XForge.Scripture/Services/SFScrTextCollection.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using Paratext.Data;
 using SIL.WritingSystems;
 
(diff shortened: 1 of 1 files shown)


marksvc:
Note that this code presumably does not distinguish between a resource Source missing an ldml file, and a PT7 Target project that is missing an ldml file. Gracefully failing to allow using an unmigrated PT project as a Target will need to be implemented separately.

irahopkinson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/943#-:-MTKICFZ4uwFCGXeHDFq:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #946, opened 2021-02-08
Turn on noImplicitReturns

marksvc:
---

I'm looking at more strict or lint checks we can turn on. I like that noImplicitReturns makes us be more explicit. But the particular places it required a change in didn't seem very significant. And in import-questions-dialog.component.ts it seems conceptually wrong to return undefined in places where we don't want to add to the array.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/946)


irahopkinson (approved):
:lgtm: thanks for adding this.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/946#-:-MT40VEI7TVYPReRPrD8:b-dlwgr4)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #959, opened 2021-02-18
SF-1153 Store activity time, not use event.timeStamp 

marksvc:
- Event timeStamp gives huge numbers in iOS Safari, but numbers of ms
on other platforms.
- So don't use Event timeStamp, and instead store the activity start
time another way, by writing to an activity whenStarted the ms
reported from `Date.now()`.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/959)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/translate-metrics-session.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/translate-metrics-session.spec.ts
@@ -134,7 +134,23 @@ describe('TranslateMetricsSession', () => {
       verify(mockedSFProjectService.onlineAddTranslateMetrics('project01', anything())).never();
       expect(env.session.metrics.type).toBe('edit');
       expect(env.session.metrics.mouseClickCount).toBe(1);
+      env.session.dispose();
+    }));
 
+    it('editing increases activity timespan', fakeAsync(() => {
+      const env = new TestEnvironment();
+      // Milliseconds delay between key down and key up, to make sure the timespan between them is not rounded down
+      // to 0. But not greater than the timeout.
+      const momentaryDelay = ACTIVE_EDIT_TIMEOUT * 0.5;
+
+      expect(env.session.metrics.timeEditActive).toBeUndefined();
+      env.keyPress('a', momentaryDelay);
+      tick(ACTIVE_EDIT_TIMEOUT);
+      expect(env.session.metrics.type).toBe('edit');
+      expect(env.session.metrics.timeEditActive).toEqual(momentaryDelay);
+      env.keyPress('b', momentaryDelay);
+      tick(ACTIVE_EDIT_TIMEOUT);
+      expect(env.session.metrics.timeEditActive).toEqual(2 * momentaryDelay);
       env.session.dispose();
     }));
 
@@ -387,7 +403,8 @@ class TestEnvironment {
     tick();
   }
 
-  keyPress(key: string): void {
+  /** @param timeSpanMs Optional milliseconds delay between pressing and releasing key. */
+  keyPress(key: string, timeSpanMs: number = 0): void {
     const keydownEvent: any = document.createEvent('CustomEvent');
     keydownEvent.key = key;
     keydownEvent.ctrlKey = false;
@@ -395,6 +412,7 @@ class TestEnvironment {
     keydownEvent.initEvent('keydown', true, true);
     this.target.editor!.root.dispatchEvent(keydownEvent);
 
+    tick(timeSpanMs);
     const keyupEvent: any = document.createEvent('CustomEvent');
     keyupEvent.key = key;
     keyupEvent.ctrlKey = false;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/translate-metrics-session.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/translate-metrics-session.ts
@@ -48,7 +48,7 @@ function getKeyActivityType(event: KeyboardEvent): ActivityType {
 }
 
 function createKeyActivity(event: KeyboardEvent): Activity {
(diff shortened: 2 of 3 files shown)


irahopkinson (changes requested):
Its really good that you found that a really large number was being put into this property as the first cause of this problem. I understand that took quite a bit of work. However if we merge this first (assuming my comments below can be satisfied) then that will just mask this problem. I would prefer we get to the root of this problem rather than patch it with this PR. Keep asking 'why' until you get to the root cause. To that end, SF-1210 is not really needed because we need to solve it here.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/959#-:-MTuto_HChSYb_mv2NFQ:b9oi3pe)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/Models/TranslateMetrics.cs, line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/959#-MTusspeAKI2au6OGOTm:-MTusspeAKI2au6OGOTn:b2klpqa) ([raw file](https://github.com/sillsdev/web-xforge/blob/42f4d51a1afd1cc07f1b3b084369c10575557260/src/SIL.XForge.Scripture/Models/TranslateMetrics.cs#L27)):*
> ```
> int
> ```
So how long is a 32-bit ms? I calculate it as about 3.2 years. Do you think a user can actively edit non-stop for anywhere near that long?
___
*[src/SIL.XForge.Scripture/Models/TranslateMetrics.cs, line 28 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/959#-MTutFHR7LfvfCXEmYBm:-MTutFHR7LfvfCXEmYBn:bgrkq9n) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0d75424aa55bf07c1c47eaeb5ae140fc5fe0f33/src/SIL.XForge.Scripture/Models/TranslateMetrics.cs#L28)):*
> ```
> Int64
> ```
My first concern here is will this go into Mongo? How does Mongo store this? Does our C# to Mongo mapper handle `Int64`?

marksvc (commented):
Yeah, I wasn't wanting to just make the problem be ignored, but it seemed that we really needed to increase the capacity of the TimeEditActive field. But I longer think so, which I discuss below.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/959#-:-MTwJ4A70ZqvmhqHEf1A:b-2r69hs)* status: all files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @marksvc)

___
*[src/SIL.XForge.Scripture/Models/TranslateMetrics.cs, line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/959#-MTusspeAKI2au6OGOTm:-MTwA6G4BBSPEzCqYwFi:b-7v9b23) ([raw file](https://github.com/sillsdev/web-xforge/blob/42f4d51a1afd1cc07f1b3b084369c10575557260/src/SIL.XForge.Scripture/Models/TranslateMetrics.cs#L27)):*

> ```
> int
> ```
So how long is a 32-bit ms? I calculate it as about 3.2 years. Do you think a user can actively edit non-stop for anywhere near that long?
</blockquote></details>

So if I do (2^32)/1000/60/60 I get 1193 hours. So if a user edits SF for 8 hours a day they can hit the limit after 149 days. If there are 4 people collaborating, they can do it faster. That would be a lot of usage. However, as I look at the rest of the object I can see that since it specifies SessionId, UserRef, and ChapterNum, I am thinking this model object may only be recording usage data for a particular short period of time, rather than the aggregate usage over a longer period of time. 

I'm looking over translate-metric-session.ts to see how it's using TranslateMetrics objects. It looks like it may be common to create new metrics and send them. And for the same SessionId as well. So I withdraw my concern. 

I will write a couple class comments from what I have learned so far.

irahopkinson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/959)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/Models/TranslateMetrics.cs, line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/959#-MTusspeAKI2au6OGOTm:-MTxM-jT8KBwFQThMGPl:b-f6j62r) ([raw file](https://github.com/sillsdev/web-xforge/blob/42f4d51a1afd1cc07f1b3b084369c10575557260/src/SIL.XForge.Scripture/Models/TranslateMetrics.cs#L27)):*

So if I do (2^32)/1000/60/60 I get 1193 hours. So if a user edits SF for 8 hours a day they can hit the limit after 149 days. If there are 4 people collaborating, they can do it faster. That would be a lot of usage. However, as I look at the rest of the object I can see that since it specifies SessionId, UserRef, and ChapterNum, I am thinking this model object may only be recording usage data for a particular short period of time, rather than the aggregate usage over a longer period of time. 

I'm looking over translate-metric-session.ts to see how it's using TranslateMetrics objects. It looks like it may be common to create new metrics and send them. And for the same SessionId as well. So I withdraw my concern. 

I will write a couple class comments from what I have learned so far. 
</blockquote></details>

Yep your calculation is right at 1193 hours (it was late last night when I tried). And this is for active edit time, i.e. if you stop typing for a few seconds (depending on its timeout value that I think is adjustable) then it considers your active edit to be over. I don't think anyone could actively edit for 24/7 for 49 days.

marksvc:
(I am working on a fix that takes into account Mobile Safari)

marksvc:
This PR was re-worked.

irahopkinson (approved):
:lgtm: thanks for persisting with this.

> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/959#-:-MUrSgiVAA7sCciE7rRy:bikzmdl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #962, opened 2021-02-23
SF-1212 Show message when source project cannot be deleted

RaymondLuong3:
* prevent users from deleting projects used as a source translation
* ignore source project during sync if not found
* skip adding user to a source project if not found

settings_page_cannot_delete_source

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/962)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts
@@ -134,6 +134,10 @@ export class SFProjectService extends ProjectService<SFProject, SFProjectDoc> {
     return (await this.onlineInvoke<string>('uninviteUser', { projectId, emailToUninvite }))!;
   }
 
+  async onlineIsSourceProject(projectId: string): Promise<boolean> {
+    return (await this.onlineInvoke<boolean>('isSourceProject', { projectId }))!;
+  }
+
   async transceleratorQuestions(projectId: string): Promise<TransceleratorQuestion[]> {
     return (await this.onlineInvoke<TransceleratorQuestion[]>('transceleratorQuestions', { projectId }))!;
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/settings/delete-project-dialog/delete-project-dialog.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/settings/delete-project-dialog/delete-project-dialog.component.html
@@ -12,19 +12,14 @@
             "
           ></p>
           <br />
-          <p *ngIf="isSourceProject; else deleteForm" class="source-project-msg">
-            {{ t("source_projects_cannot_be_deleted") }}
-          </p>
-          <ng-template #deleteForm>
-            <p>{{ t("please_type_project_name") }}</p>
-            <mdc-form-field>
-              <mdc-text-field
-                id="project-entry"
-                label="{{ t('project_name') }}"
-                [formControl]="projectNameEntry"
-              ></mdc-text-field>
-            </mdc-form-field>
-          </ng-template>
+          <p>{{ t("please_type_project_name") }}</p>
+          <mdc-form-field>
+            <mdc-text-field
+              id="project-entry"
+              label="{{ t('project_name') }}"
+              [formControl]="projectNameEntry"
+            ></mdc-text-field>
+          </mdc-form-field>
         </mdc-dialog-content>
         <mdc-dialog-actions>
           <button mdcDialogButton type="button" id="cancel-btn" mdcDialogAction="cancel">{{ t("cancel") }}</button>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/settings/delete-project-dialog/delete-project-dialog.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/settings/delete-project-dialog/delete-project-dialog.component.html
@@ -12,14 +12,19 @@
             "
           ></p>
           <br />
-          <p>{{ t("please_type_project_name") }}</p>
-          <mdc-form-field>
-            <mdc-text-field
-              id="project-entry"
-              label="{{ t('project_name') }}"
-              [formControl]="projectNameEntry"
-            ></mdc-text-field>
-          </mdc-form-field>
+          <p *ngIf="isSourceProject; else deleteForm" class="source-project-msg">
+            {{ t("source_projects_cannot_be_deleted") }}
+          </p>
+          <ng-template #deleteForm>
+            <p>{{ t("please_type_project_name") }}</p>
+            <mdc-form-field>
+              <mdc-text-field
+                id="project-entry"
+                label="{{ t('project_name') }}"
+                [formControl]="projectNameEntry"
+              ></mdc-text-field>
+            </mdc-form-field>
+          </ng-template>
         </mdc-dialog-content>
         <mdc-dialog-actions>
           <button mdcDialogButton type="button" id="cancel-btn" mdcDialogAction="cancel">{{ t("cancel") }}</button>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/settings/delete-project-dialog/delete-project-dialog.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/settings/delete-project-dialog/delete-project-dialog.component.scss
@@ -1,3 +1,0 @@
-.source-project-msg {
-  color: var(--mdc-theme-error);
-}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/settings/delete-project-dialog/delete-project-dialog.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/settings/delete-project-dialog/delete-project-dialog.component.scss
@@ -1,0 +1,3 @@
+.source-project-msg {
+  color: var(--mdc-theme-error);
+}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/settings/delete-project-dialog/delete-project-dialog.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/settings/delete-project-dialog/delete-project-dialog.component.spec.ts
@@ -1,19 +1,24 @@
 import { MdcDialog, MdcDialogConfig, MdcDialogRef } from '@angular-mdc/web/dialog';
+import { CommonModule } from '@angular/common';
 import { Component, Directive, NgModule, ViewChild, ViewContainerRef } from '@angular/core';
 import { ComponentFixture, fakeAsync, flush, inject, TestBed, tick } from '@angular/core/testing';
 import { CookieService } from 'ngx-cookie-service';
-import { mock } from 'ts-mockito';
+import { anything, mock, when } from 'ts-mockito';
 import { AuthService } from 'xforge-common/auth.service';
 import { configureTestingModule, TestTranslocoModule } from 'xforge-common/test-utils';
 import { UICommonModule } from 'xforge-common/ui-common.module';
+import { SFProjectService } from '../../core/sf-project.service';
 import { DeleteProjectDialogComponent } from './delete-project-dialog.component';
+
+const mockedProjectService = mock(SFProjectService);
 
 describe('DeleteProjectDialogComponent', () => {
   configureTestingModule(() => ({
     imports: [DialogTestModule, UICommonModule],
     providers: [
       { provide: AuthService, useMock: mock(AuthService) },
-      { provide: CookieService, useMock: mock(CookieService) }
+      { provide: CookieService, useMock: mock(CookieService) },
+      { provide: SFProjectService, useMock: mockedProjectService }
     ]
   }));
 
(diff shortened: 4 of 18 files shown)


irahopkinson (changes requested):
Thanks for working on this. I have some questions about the approach.

> Reviewed 14 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/962#-:-MUNYtMP3mp2Yr1unehl:brd04b1)* status: all files reviewed, 5 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/settings/delete-project-dialog/delete-project-dialog.component.html, line 17 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/962#-MUNSQeVBbNFVryezdFI:-MUNSQeVBbNFVryezdFJ:b-1ndrlh) ([raw file](https://github.com/sillsdev/web-xforge/blob/efddbc5ba4a355f23f8049b83b071a5254104312/src/SIL.XForge.Scripture/ClientApp/src/app/settings/delete-project-dialog/delete-project-dialog.component.html#L17)):*
> ```
>           <p *ngIf="isSourceProject; else deleteForm" class="source-project-msg">
>             {{ t("source_projects_cannot_be_deleted") }}
>           </p>
> ```
Is there some reason this is displayed in this dialog rather than directly on the **Settings** page? I would prefer we do it there rather than here as its a better UX if we display this as soon as possible to the user and disable the delete button there.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/962#-MUNURSXAOqyS425oyd3:-MUNURSXAOqyS425oyd4:b6d8w47) ([raw file](https://github.com/sillsdev/web-xforge/blob/efddbc5ba4a355f23f8049b83b071a5254104312/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts#L68)):*
<details><summary><i alt="summary">Quoted 5 lines of code…</i></summary>

> ```
>         } catch (err) {
>           if (!(err instanceof CommandError && err.code === CommandErrorCode.NotFound)) {
>             throw err;
>           }
>         }
> ```
</details>

Is this getting uglier because we are still asking about the role rather than rights or permissions? Which call is causing the exception, `this.projectService.onlineGetProjectRole` or `this.projectService.get`?
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 77 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/962#-MUNVbNRAI8pGuNX6te2:-MUNVbNRAI8pGuNX6te3:b7dxa8b) ([raw file](https://github.com/sillsdev/web-xforge/blob/efddbc5ba4a355f23f8049b83b071a5254104312/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L77)):*
> ```
> source
> ```
As developers we talk about source and target but the UI does not. How about something like "Another project is based on this project and so it cannot be deleted." Also **either** don't use "Scripture Forge" directly since its redundant since its the software we are talking about **or** if you do then use the environment variable that doesn't get translated and can be renamed.
___
*[src/SIL.XForge.Scripture/Controllers/SFProjectsRpcController.cs, line 243 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/962#-MUNX26G0LJX88dfzPd7:-MUNX26G0LJX88dfzPd8:baqweii) ([raw file](https://github.com/sillsdev/web-xforge/blob/efddbc5ba4a355f23f8049b83b071a5254104312/src/SIL.XForge.Scripture/Controllers/SFProjectsRpcController.cs#L243)):*
> ```
> public IRpcMethodResult IsSourceProject(string projectId)
> ```
I'm wondering if we need to add another endpoint (especially if you move the message to the **Settings** page) since there may be some other appropriate endpoint that is already being called that we can include this info with? Round trips to the server take time. Doing several one after the other isn't ideal if we can combine info.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 526 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/962#-MUNYDen32Uhnw4BYdj6:-MUNYDen32Uhnw4BYdj7:b-b6vcl3) ([raw file](https://github.com/sillsdev/web-xforge/blob/efddbc5ba4a355f23f8049b83b071a5254104312/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L526)):*
> ```
> DataNotFoundException
> ```
Will you only get this exception if the source has been deleted? I'm concerned this might ignore some legitimate exceptions. How else could you do this?

RaymondLuong3 (commented):
Thanks for the feedback. I've made some adjustments.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/962#-:-MUPz5_Q5xRDM0u8iWtg:bcddl4d)* status: 4 of 13 files reviewed, 5 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/settings/delete-project-dialog/delete-project-dialog.component.html, line 17 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/962#-MUNSQeVBbNFVryezdFI:-MUPzBCm02cTnmYJv26A:b8kf59r) ([raw file](https://github.com/sillsdev/web-xforge/blob/efddbc5ba4a355f23f8049b83b071a5254104312/src/SIL.XForge.Scripture/ClientApp/src/app/settings/delete-project-dialog/delete-project-dialog.component.html#L17)):*

> ```
>           <p *ngIf="isSourceProject; else deleteForm" class="source-project-msg">
>             {{ t("source_projects_cannot_be_deleted") }}
>           </p>
> ```
Is there some reason this is displayed in this dialog rather than directly on the **Settings** page? I would prefer we do it there rather than here as its a better UX if we display this as soon as possible to the user and disable the delete button there.
</blockquote></details>

Initially I thought that the settings pages appears to have a lot going on already, and I wasn't convinced that continually showing a 'warning' to the user was helpful (I expect only a handful of people will delete their project in a given year) so maybe that warning was better off in the delete dialog. But seeing it on the settings page now, it lots just fine.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/962#-MUNURSXAOqyS425oyd3:-MUPzrbwEYi-0PSUNRRv:b-gpybv8) ([raw file](https://github.com/sillsdev/web-xforge/blob/efddbc5ba4a355f23f8049b83b071a5254104312/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts#L68)):*

> ```
>         } catch (err) {
>           if (!(err instanceof CommandError && err.code === CommandErrorCode.NotFound)) {
>             throw err;
>           }
>         }
> ```
Is this getting uglier because we are still asking about the role rather than rights or permissions? Which call is causing the exception, `this.projectService.onlineGetProjectRole` or `this.projectService.get`?
</blockquote></details>

It comes from `onlineGetProjectRole`. It seems that it is unfortunate that this scenario comes up - that a project doesn't exist, but as demonstrated it can happen.
I wonder how we would approach using rights and permissions. I see permissions are stored in the `SFProject` so there isn't a way to access them unless we are on the project. And project rights in `sf-project-rights` does not apply before we know a user's role.
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 77 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/962#-MUNVbNRAI8pGuNX6te2:-MUQ94K-65nSAWyyDcI7:btyeq5o) ([raw file](https://github.com/sillsdev/web-xforge/blob/efddbc5ba4a355f23f8049b83b071a5254104312/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L77)):*

> ```
> source
> ```
As developers we talk about source and target but the UI does not. How about something like "Another project is based on this project and so it cannot be deleted." Also **either** don't use "Scripture Forge" directly since its redundant since its the software we are talking about **or** if you do then use the environment variable that doesn't get translated and can be renamed.
</blockquote></details>

Good thinking. Done.
___
*[src/SIL.XForge.Scripture/Controllers/SFProjectsRpcController.cs, line 243 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/962#-MUNX26G0LJX88dfzPd7:-MUQ98S59q2kpJcpmO3d:brkbxc1) ([raw file](https://github.com/sillsdev/web-xforge/blob/efddbc5ba4a355f23f8049b83b071a5254104312/src/SIL.XForge.Scripture/Controllers/SFProjectsRpcController.cs#L243)):*

> ```
> public IRpcMethodResult IsSourceProject(string projectId)
> ```
I'm wondering if we need to add another endpoint (especially if you move the message to the **Settings** page) since there may be some other appropriate endpoint that is already being called that we can include this info with? Round trips to the server take time. Doing several one after the other isn't ideal if we can combine info.
</blockquote></details>

I checked the other endpoints that the Settings page uses. It uses `paratext.service` on initialization, but not the `project.service` so I'm not sure we can avoid a network call. But since we call it while the settings page is loading, it should be ok if it is a little slow.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 526 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/962#-MUNYDen32Uhnw4BYdj6:-MUQ9m4a8xW4zqRztGq8:bxoxcz4) ([raw file](https://github.com/sillsdev/web-xforge/blob/efddbc5ba4a355f23f8049b83b071a5254104312/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L526)):*

> ```
> DataNotFoundException
> ```
Will you only get this exception if the source has been deleted? I'm concerned this might ignore some legitimate exceptions. How else could you do this?
</blockquote></details>

I've moved this out into a `TryGetProjectDocAsync` method that returns null if the projectDoc don't exist.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/962)* status: 3 of 14 files reviewed, 5 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/962#-MUNURSXAOqyS425oyd3:-MUUEs3Z1WyvzDz5COJv:b-jknozi) ([raw file](https://github.com/sillsdev/web-xforge/blob/efddbc5ba4a355f23f8049b83b071a5254104312/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts#L68)):*

It comes from `onlineGetProjectRole`. It seems that it is unfortunate that this scenario comes up - that a project doesn't exist, but as demonstrated it can happen.
I wonder how we would approach using rights and permissions. I see permissions are stored in the `SFProject` so there isn't a way to access them unless we are on the project. And project rights in `sf-project-rights` does not apply before we know a user's role.
</blockquote></details>

Ok, I've made a change to the endpoint to return null if we are requesting the roles for a project that is invalid. Then the role returned here will be `none`.

irahopkinson (approved):
> Reviewed 13 of 15 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/962)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.html, line 167 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/962#-MUm3wZnAEBR0FMX2MyR:-MUm3wZnAEBR0FMX2MyS:bsx7oqo) ([raw file](https://github.com/sillsdev/web-xforge/blob/5eddacc730de634249a23a750aaf310bfa6a2a2b/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.html#L167)):*
> ```
> </ng-template
> ```
nit: this should be on the next line with its closing angle bracket `>`.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/962#-MUNURSXAOqyS425oyd3:-MUm2ayTD2yn3vLMQ_Vm:bu0eg4q) ([raw file](https://github.com/sillsdev/web-xforge/blob/efddbc5ba4a355f23f8049b83b071a5254104312/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts#L68)):*

Ok, I've made a change to the endpoint to return null if we are requesting the roles for a project that is invalid. Then the role returned here will be `none`.
</blockquote></details>

Nice solution!

irahopkinson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/962)* status: 13 of 14 files reviewed, all discussions resolved (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.html, line 167 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/962#-MUm3wZnAEBR0FMX2MyR:-MUmA0m83ImTna7LgQ25:bb6un1j) ([raw file](https://github.com/sillsdev/web-xforge/blob/5eddacc730de634249a23a750aaf310bfa6a2a2b/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.html#L167)):*

> ```
> </ng-template
> ```
nit: this should be on the next line with its closing angle bracket `>`.
</blockquote></details>

Done

irahopkinson (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/962)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #963, opened 2021-02-25
Use alternate DBL qa server

RaymondLuong3:
I am unsure if there is a real difference between the two urls. It appears that the older url tries to get to a server that is offline at the moment (at least for me and for testers). This may be needed to get testers unblocked.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/963)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -88,7 +88,7 @@ namespace SIL.XForge.Scripture.Services
             {
                 _httpClientHandler.ServerCertificateCustomValidationCallback
                     = HttpClientHandler.DangerousAcceptAnyServerCertificateValidator;
-                _dblServerUri = "https://paratext-qa.thedigitalbiblelibrary.org/";
+                _dblServerUri = "https://qa.thedigitalbiblelibrary.org/";
                 _registryServerUri = "https://registry-dev.paratext.org";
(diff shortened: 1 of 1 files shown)


irahopkinson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/963)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 91 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/963#-MUTrulU1TUnVTSL0Foa:-MUTrulU1TUnVTSL0Fob:baspqx3) ([raw file](https://github.com/sillsdev/web-xforge/blob/be1a54e355062522996b3034bc862ebf82de7831/src/SIL.XForge.Scripture/Services/ParatextService.cs#L91)):*
> ```
> https://qa.thedigitalbiblelibrary.org/
> ```
If you want to merge this it would be better to add this as a comment of alternative. Unless we hear that this is a better change.

RaymondLuong3 (commented):
It seems that the original DBL server is the one that we should keep. I'll go ahead to move this to be a draft so that testers can reference to it if they need help getting unstuck.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/963#-:-MUV6RuCDn9DIYIJAOwl:bp47ps7)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

PR #965, opened 2021-03-03
Fix date localization test so it can pass in both Chromium 87 and 88

Nateowami:
Chromium stopped including a space in the date format for zh-CN, which broke a test.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/965)


nigel-wells (approved):
Passes for me now as well. Do we need to adjust anything for how it is displayed in the app i.e. Is that locale not supposed to have a space in the first place which is why Chrome is now removing it? Would it look odd to someone in that language without a space? Probably overthinking this...

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/965#-:-MUqGYlB1AzTgTpGGZ12:b-mok4ki)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #967, opened 2021-03-03
SF-1199 Include project role in sharekey

RaymondLuong3:
* link sharing now uses a sharekey with a corresponding role
* the backend now expects a role whenever inviting a user
* the share control displays the share link with the sharekey

In order to test these changes, you must first run this command in the mongo shell:
```
use xforge
db.sf_project_secrets.updateMany(
	{ },
	{ $set: { "shareKeys.$[elem].projectRole": "sf_community_checker" } },
	{ arrayFilters: [ { "elem.projectRole": { $exists: false } } ] }
)
db.sf_project_secrets.updateMany(
	{ },
	{ $set: { "shareKeys.$[elem].expirationTime": new Date(2021, 11) } },
	{ arrayFilters: [ { "elem.expirationTime": { $exists: false } } ] }
)
```

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/967)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts
@@ -2,6 +2,7 @@ import { Injectable } from '@angular/core';
 import { obj } from 'realtime-server/lib/common/utils/obj-path';
 import { getQuestionDocId, Question } from 'realtime-server/lib/scriptureforge/models/question';
 import { SFProject, SF_PROJECTS_COLLECTION } from 'realtime-server/lib/scriptureforge/models/sf-project';
+import { SFProjectRole } from 'realtime-server/lib/scriptureforge/models/sf-project-role';
 import { getSFProjectUserConfigDocId } from 'realtime-server/lib/scriptureforge/models/sf-project-user-config';
 import { CommandService } from 'xforge-common/command.service';
 import { FileService } from 'xforge-common/file.service';
@@ -126,8 +127,8 @@ export class SFProjectService extends ProjectService<SFProject, SFProjectDoc> {
     return this.onlineInvoke('checkLinkSharing', { projectId: id, shareKey });
   }
 
-  onlineInvite(id: string, email: string, locale: string): Promise<string | undefined> {
-    return this.onlineInvoke('invite', { projectId: id, email, locale });
+  onlineInvite(id: string, email: string, locale: string, role: string): Promise<string | undefined> {
+    return this.onlineInvoke('invite', { projectId: id, email, locale, role });
   }
 
   async onlineUninviteUser(projectId: string, emailToUninvite: string): Promise<string> {
@@ -136,6 +137,10 @@ export class SFProjectService extends ProjectService<SFProject, SFProjectDoc> {
 
   async onlineIsSourceProject(projectId: string): Promise<boolean> {
     return (await this.onlineInvoke<boolean>('isSourceProject', { projectId }))!;
+  }
+
+  async onlineGetLinkSharingKey(projectId: string, role: SFProjectRole): Promise<string> {
+    return (await this.onlineInvoke<string>('linkSharingKey', { projectId, role })) ?? '';
   }
 
   async transceleratorQuestions(projectId: string): Promise<TransceleratorQuestion[]> {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts
@@ -1,6 +1,6 @@
 import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { Component, DebugElement, NgModule, ViewChild } from '@angular/core';
-import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { ComponentFixture, fakeAsync, TestBed } from '@angular/core/testing';
 import { flush } from '@angular/core/testing';
 import { BrowserModule, By } from '@angular/platform-browser';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
@@ -93,7 +93,7 @@ describe('ShareControlComponent', () => {
     env.setInvitationLanguage('en');
     env.click(env.sendButton);
     verify(mockedNoticeService.show(anything())).once();
-    verify(mockedProjectService.onlineInvite(anything(), anything(), anything())).once();
+    verify(mockedProjectService.onlineInvite(anything(), anything(), anything(), anything())).once();
     expect(capture(mockedNoticeService.show).last()[0]).toContain('email has been sent');
   }));
 
@@ -103,14 +103,16 @@ describe('ShareControlComponent', () => {
     env.setInvitationLanguage('en');
     env.click(env.sendButton);
     verify(mockedNoticeService.show(anything())).once();
-    verify(mockedProjectService.onlineInvite(anything(), anything(), anything())).once();
+    verify(mockedProjectService.onlineInvite(anything(), anything(), anything(), anything())).once();
 
     expect(capture(mockedNoticeService.show).last()[0]).toContain('is already');
   }));
 
-  it('shareLink is for projectId', fakeAsync(() => {
+  it('shareLink is for projectId and has specific key', fakeAsync(() => {
     const env = new TestEnvironment(true, 'myProject1');
-    expect(env.component.shareLink).toContain('myProject1');
+    env.wait();
+    verify(mockedProjectService.onlineGetLinkSharingKey('myProject1', anything())).once();
+    expect(env.hostComponent.component.shareLink).toContain('myProject1?sharing=true&shareKey=linkSharing01');
     flush();
   }));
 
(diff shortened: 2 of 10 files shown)


Nateowami (commented):
So far this looks good (I'm still in the process of reviewing it).

Shouldn't the migration be run automatically? Do we not have a mechanism for migrating the DB?

> Reviewed 7 of 10 files at r1, 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/967#-:-MUtfMdj5gOx6z2gA-xk:b-j13zmm)* status: 7 of 10 files reviewed, all discussions resolved

Nateowami (changes requested):
> Reviewed 1 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/967)* status: 8 of 10 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[src/SIL.XForge/Models/ShareKey.cs, line 10 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/967#-MUu5ZpU5aBCMmRfq1i-:-MUu5ZpU5aBCMmRfq1i0:b-h0fj0p) ([raw file](https://github.com/sillsdev/web-xforge/blob/974728d8b66a7b0b2706f8896553f8b0df5caac8/src/SIL.XForge/Models/ShareKey.cs#L10)):*
> ```Smalltalk
>         public string Key { get; set; }
>         // Use a default expiration time for sharekeys without an expiration (sharekeys created prior to SF-753)
>         public DateTime? ExpirationTime { get; set; } = new DateTime(2021, 12, 1, 0, 0, 0, DateTimeKind.Utc);
> ```

We now have three types of share keys:
- User-specific share keys without expiration dates, that should expire come December.
- User-specific share keys that expire 14 days after they are created (if I recall correctly)
- Generic/reusable share keys that should not expire (unless they are deliberately revoked or something, but we don't have a way to do that yet)

My  questions are basically:
1. Do we handle this correctly? Won't the default expiration date be applied to the generic/reusable share keys? Or do we just ignore the expiration date if the email is null (that sounds like confusion waiting to happen)?
2. Might it be better to just migrate the DB to add an expiration to any share key that lacks one? This is easier done now before we start adding generic/reusable share keys? Also it means going forward our data in the DB is more consistent, meaning fewer edge cases to deal with.
3. Depending on what the answers to 1 and 2 above are, I think this may call for some comments explaining things, especially explaining the fact that `Email` may be null, and what that indicates.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/967)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 484 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/967#-MUu_1SgECln2hRRK1IH:-MUu_1SgECln2hRRK1II:bipo665) ([raw file](https://github.com/sillsdev/web-xforge/blob/974728d8b66a7b0b2706f8896553f8b0df5caac8/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L484)):*
> ```Smalltalk
>                 if (linkSharing)
>                 {
>                     await AddUserToProjectAsync(conn, projectDoc, userDoc, projectRole, true);
> ```

I'm confused regarding how the logic in this method works. It looks like we're fetching the user's role on the project (if it exists), and then adding the user again (even if they're already on the project). Wouldn't it make more sense to just not add the user if they're already on the project? Maybe the operation is idempotent, but I can't tell that from a quick glance at it, and I have no assurances that it will continue to be idempotent in the future.

Or maybe I'm reading this incorrectly.

RaymondLuong3 (commented):
I am not sure if there is a mechanism to do the migration already in place. @irahopkinson may know.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/967#-:-MUyLe9xB7tdExWWfJY4:b-s5xsn3)* status: 7 of 10 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge/Models/ShareKey.cs, line 10 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/967#-MUu5ZpU5aBCMmRfq1i-:-MUyLuwX90up_UB8KKNV:b-un9ef0) ([raw file](https://github.com/sillsdev/web-xforge/blob/974728d8b66a7b0b2706f8896553f8b0df5caac8/src/SIL.XForge/Models/ShareKey.cs#L10)):*

We now have three types of share keys:
- User-specific share keys without expiration dates, that should expire come December.
- User-specific share keys that expire 14 days after they are created (if I recall correctly)
- Generic/reusable share keys that should not expire (unless they are deliberately revoked or something, but we don't have a way to do that yet)

My  questions are basically:
1. Do we handle this correctly? Won't the default expiration date be applied to the generic/reusable share keys? Or do we just ignore the expiration date if the email is null (that sounds like confusion waiting to happen)?
2. Might it be better to just migrate the DB to add an expiration to any share key that lacks one? This is easier done now before we start adding generic/reusable share keys? Also it means going forward our data in the DB is more consistent, meaning fewer edge cases to deal with.
3. Depending on what the answers to 1 and 2 above are, I think this may call for some comments explaining things, especially explaining the fact that `Email` may be null, and what that indicates.
</blockquote></details>

Ok, I've updated the mongo shell update command to include adding an expiration time to the share keys. There will only be two kinds of shareKeys now.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 484 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/967#-MUu_1SgECln2hRRK1IH:-MUySQkj6hMMQ2iH0HRe:b-h5eo1r) ([raw file](https://github.com/sillsdev/web-xforge/blob/974728d8b66a7b0b2706f8896553f8b0df5caac8/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L484)):*

I'm confused regarding how the logic in this method works. It looks like we're fetching the user's role on the project (if it exists), and then adding the user again (even if they're already on the project). Wouldn't it make more sense to just not add the user if they're already on the project? Maybe the operation is idempotent, but I can't tell that from a quick glance at it, and I have no assurances that it will continue to be idempotent in the future.

Or maybe I'm reading this incorrectly.
</blockquote></details>

I added a comment that will help. When we try to get the project role, we try to access the role in the paratext registry if the user is not already on the SF project. Then that role will override what is stored in the share key. This is useful so that a translator or consultant joins the SF project using a share link, they are given the role they have on the PT project.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/967)* status: 7 of 10 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge/Models/ShareKey.cs, line 10 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/967#-MUu5ZpU5aBCMmRfq1i-:-MUzFkpa-kSGoxh4tQGg:b-vz58lp) ([raw file](https://github.com/sillsdev/web-xforge/blob/974728d8b66a7b0b2706f8896553f8b0df5caac8/src/SIL.XForge/Models/ShareKey.cs#L10)):*

Ok, I've updated the mongo shell update command to include adding an expiration time to the share keys. There will only be two kinds of shareKeys now.
</blockquote></details>

Thanks. I'm not super familiar with Mongo queries, but isn't this going to overwrite `expirationTime`? Maybe this needs to be done as two separate queries?
``` js
use xforge
db.sf_project_secrets.updateMany(
	{ },
	{ $set: { "shareKeys.$[elem].projectRole": "sf_community_checker" , "shareKeys.$[elem].expirationTime": new Date(2021, 11) } },
	{ arrayFilters: [ { "elem.projectRole": { $exists: false } } ] }
)
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/967)* status: 7 of 10 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge/Models/ShareKey.cs, line 10 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/967#-MUu5ZpU5aBCMmRfq1i-:-MUzLPYh-z61MZeayjos:b-nr7m1w) ([raw file](https://github.com/sillsdev/web-xforge/blob/974728d8b66a7b0b2706f8896553f8b0df5caac8/src/SIL.XForge/Models/ShareKey.cs#L10)):*

Thanks. I'm not super familiar with Mongo queries, but isn't this going to overwrite `expirationTime`? Maybe this needs to be done as two separate queries?
``` js
use xforge
db.sf_project_secrets.updateMany(
	{ },
	{ $set: { "shareKeys.$[elem].projectRole": "sf_community_checker" , "shareKeys.$[elem].expirationTime": new Date(2021, 11) } },
	{ arrayFilters: [ { "elem.projectRole": { $exists: false } } ] }
)
```
</blockquote></details>

Thanks for the catch! It is true that we have already released to QA adding an expiration date, so we need to handle updating the DB separately, once to update the DB with the project role, and once to add an expiration time to documents that do not have one.

Nateowami (commented):
:lgtm: I think we probably need to figure out how to do the migrations first before merging this though. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/967#-:-MUzPU-C1ms84bX2MJeC:b-fcttdh)* status: :shipit: complete! all files reviewed, all discussions resolved

Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/967#-:-MV3ofxCD5d0TeD5A3wC:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #969, opened 2021-03-08
Deploy: Install 503 error page

marksvc:
- 503.html was being installed manually. Automate it. Any error pages
in the ErrorPages directory will be included.
- `-progzlt` was quite close to `-a`, which is basically `progltD`. I
include `v` so we can see the transfer in the TC log.
- I did not include `--chmod`. The 503.html on scriptureforge-live has
permissions of 644. Testing to scriptureforge-qa shows it still uses
mod 644.
--- 

This relates to a discussion with Ira regarding our ansible, so probably don't merge this PR without  @irahopkinson 's approving first.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/969)

### DIFF
--- a/scripts/build-and-ship
+++ b/scripts/build-and-ship
@@ -11,6 +11,8 @@ SERVICE_UNIT_SUFFIX=""
 if grep "beta" <<< "${APP_SUFFIX}"; then
   SERVICE_UNIT_SUFFIX="_beta"
 fi
+# Relative to repo root
+ERROR_PAGES_PATH="src/SIL.XForge.Scripture/ErrorPages"
 
 cd ..
 
@@ -44,5 +46,8 @@ sudo chown -R :www-data "${BUILD_OUTPUT}/app"
 
 rsync -progzlt --chmod=Dug=rwx,Fug=rwx,o-rwx --delete-during --stats --rsync-path="sudo rsync" \
   --rsh="ssh -v -i ${DEPLOY_CREDENTIALS}" artifacts/app/ root@"${DEPLOY_DESTINATION}:${DEPLOY_PATH}/app"
+
+rsync -vaz --rsync-path="sudo rsync" --rsh="ssh -v -i ${DEPLOY_CREDENTIALS}" \
+  "${ERROR_PAGES_PATH}/" root@"${DEPLOY_DESTINATION}:${DEPLOY_PATH}/htdocs"
 
 ssh root@"${DEPLOY_DESTINATION}" "systemctl restart ${APP_NAME}-web-app${SERVICE_UNIT_SUFFIX}"


irahopkinson (changes requested):
Thanks for adding this. Only one blocking comment for authprogs.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/969#-:-MVM2iWi7Cg2k4feo2hU:b-kpvxew)* status: all files reviewed, 3 unresolved discussions (waiting on @marksvc)

___
*[scripts/build-and-upload-beta.sh, line 9 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/969#-MVM0T5L4tEeXpsfIYde:-MVM0T5L4tEeXpsfIYdf:bwghk6g) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b3eb0bcee3d1e2c4909b991f9d8b12cce186df3/scripts/build-and-upload-beta.sh#L9)):*
> ```
> ERROR_PAGES_PATH=src/SIL.XForge.Scripture/ErrorPages
> ```
You could check that the path is good by running it from the command build server manually.
___
*[scripts/build-and-upload-beta.sh, line 42 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/969#-MVM0AJ_-271VFS9AL6l:-MVM0AJaDuDK8jiA9bZB:bu95tgh) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b3eb0bcee3d1e2c4909b991f9d8b12cce186df3/scripts/build-and-upload-beta.sh#L42)):*
> ```
> ${DEPLOY_CREDENTIALS}
> ```
Showing my shell script ignorance here. What is the difference between this and the line above that uses `$DEPLOY_CREDENTIALS` (without {})?
___
*[scripts/build-and-upload-beta.sh, line 42 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/969#-MVM0fGUB9pGPuz8GV3J:-MVM0fGUB9pGPuz8GV3K:b8juwxo) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b3eb0bcee3d1e2c4909b991f9d8b12cce186df3/scripts/build-and-upload-beta.sh#L42)):*
> ```Shell
> rsync -progzlt --chmod=Dug=rwx,Fug=rwx,o-rwx --delete-during --stats --rsync-path="sudo rsync" --rsh="ssh -v -i $DEPLOY_CREDENTIALS" artifacts/app/ root@$DEPLOY_DESTINATION:$DEPLOY_PATH/app || exit 1
> 
> rsync -vaz --rsync-path="sudo rsync" --rsh="ssh -v -i ${DEPLOY_CREDENTIALS}" \
> ```

This ought to do what we want but you will have to make the appropriate changes in AuthProgs first to allow this command through. You are needing to do that anyway for non-beta.
___
*[scripts/build-and-upload-beta.sh, line 42 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/969#-MVM1CYqA6N7JvvL-k2X:-MVM1CYqA6N7JvvL-k2Y:bko3od4) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b3eb0bcee3d1e2c4909b991f9d8b12cce186df3/scripts/build-and-upload-beta.sh#L42)):*
> ```Shell
> rsync -progzlt --chmod=Dug=rwx,Fug=rwx,o-rwx --delete-during --stats --rsync-path="sudo rsync" --rsh="ssh -v -i $DEPLOY_CREDENTIALS" artifacts/app/ root@$DEPLOY_DESTINATION:$DEPLOY_PATH/app || exit 1
> 
> rsync -vaz --rsync-path="sudo rsync" --rsh="ssh -v -i ${DEPLOY_CREDENTIALS}" \
> ```

FYI only: This is however moving this in the wrong direction by adding this to this file (thats a lot usae of 'this'). The issue with having this in here rather than as a build step in TC is that you need to change and commit code inorder to update a build step. While it does have it under source control it is a bit of pain to change. However given we have the other rsync in here they probably should both stay together for now otherwise the scope of this issue is getting too large.

irahopkinson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/969)* status: all files reviewed, 3 unresolved discussions (waiting on @marksvc)

___
*[scripts/build-and-upload-beta.sh, line 9 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/969#-MVM0T5L4tEeXpsfIYde:-MVPwuwh-VHEmA8rVVCs:b-r3dj1l) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b3eb0bcee3d1e2c4909b991f9d8b12cce186df3/scripts/build-and-upload-beta.sh#L9)):*

> ```
> ERROR_PAGES_PATH=src/SIL.XForge.Scripture/ErrorPages
> ```
You could check that the path is good by running it from the command build server manually.
</blockquote></details>

Actually this should probable just be used from an environment variable directly (from TC) and should probably be the whole file path (eg `src/SIL.XForge.Scripture/ErrorPages/503.html`). That would make it more configurable in TC.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/969)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @irahopkinson)

___
*[scripts/build-and-upload-beta.sh, line 9 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/969#-MVM0T5L4tEeXpsfIYde:-MVTBGaZ4N2Vq81IKJUl:bjugyi9) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b3eb0bcee3d1e2c4909b991f9d8b12cce186df3/scripts/build-and-upload-beta.sh#L9)):*

Actually this should probable just be used from an environment variable directly (from TC) and should probably be the whole file path (eg `src/SIL.XForge.Scripture/ErrorPages/503.html`). That would make it more configurable in TC.
</blockquote></details>

Yeah, we could do that. So my thinking was that although we only have a 503.html right now, if we use an environment variable for just the directory holding the error pages, that is already starting to make provision for adding more error pages and uploading them. I think I would be inclined  for the set of error pages to be maintained in the script rather than in TC since what error pages do or don't exist is directly tied to what is in the git repository. But certainly we could do it in TC.

Okay, actually, after modifying authprogs to allow a certain rsync command, I think we should just rsync the whole directory of error pages :) I wouldn't want to have to modify authprogs if we add another error page. 
I suspect that my authprogs change is not going to be right, since it looks like we are needing to specify exact parameters, but I tried to make it right by modeling another command for now. 
What do you think?
___
*[scripts/build-and-upload-beta.sh, line 42 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/969#-MVM0AJ_-271VFS9AL6l:-MVTCPTGDWnVs5Nh1WBo:bwsk1uu) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b3eb0bcee3d1e2c4909b991f9d8b12cce186df3/scripts/build-and-upload-beta.sh#L42)):*

> ```
> ${DEPLOY_CREDENTIALS}
> ```
Showing my shell script ignorance here. What is the difference between this and the line above that uses `$DEPLOY_CREDENTIALS` (without {})?
</blockquote></details>

In this situation, there is no behavioural difference. But using braces is recommended practice ( eg https://google.github.io/styleguide/shellguide.html#variable-expansion ). It more cleanly delineates what is a variable from what is not a variable. Consider for example the following:

```bash
SITE_NAME="ScriptureForge"
SUFFIX="Beta"

echo "Welcome to $SITE_NAME$SUFFIX"
echo "Welcome to ${SITE_NAME}${SUFFIX}"
echo "Welcome to ${SITE_NAME}Beta"
```

The second `echo` is a bit more clean. And in the 3rd echo, it's significant whether braces are used, since "Beta" is appended.
___
*[scripts/build-and-upload-beta.sh, line 42 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/969#-MVM0fGUB9pGPuz8GV3J:-MVTISvMDUE6uBf_R3ub:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b3eb0bcee3d1e2c4909b991f9d8b12cce186df3/scripts/build-and-upload-beta.sh#L42)):*

This ought to do what we want but you will have to make the appropriate changes in AuthProgs first to allow this command through. You are needing to do that anyway for non-beta.
</blockquote></details>

Done.

irahopkinson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/969)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[scripts/build-and-upload-beta.sh, line 9 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/969#-MVM0T5L4tEeXpsfIYde:-MWD-2V51Tk8E2mUbMG7:b-swas5u) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b3eb0bcee3d1e2c4909b991f9d8b12cce186df3/scripts/build-and-upload-beta.sh#L9)):*

Yeah, we could do that. So my thinking was that although we only have a 503.html right now, if we use an environment variable for just the directory holding the error pages, that is already starting to make provision for adding more error pages and uploading them. I think I would be inclined  for the set of error pages to be maintained in the script rather than in TC since what error pages do or don't exist is directly tied to what is in the git repository. But certainly we could do it in TC.

Okay, actually, after modifying authprogs to allow a certain rsync command, I think we should just rsync the whole directory of error pages :) I wouldn't want to have to modify authprogs if we add another error page. 
I suspect that my authprogs change is not going to be right, since it looks like we are needing to specify exact parameters, but I tried to make it right by modeling another command for now. 
What do you think?



</blockquote></details>

I'm good with rsyncing the whole folder. I still think we need to check it by manually running the command from the build server first until AuthProgs is good.

marksvc:
Today I tested using the CI, targeting QA Beta, and it worked after an adjustment to authprogs. I pushed a corresponding authsprogs adjustment elsewhere.
For future reference, `rsync --dry-run` appeared adequate to preview the server-side command (tho with the `-n` dry-run argument that we would omit). I also tested deploying 503.html not as a dry-run, and it succeeded and had desirable permissions.

marksvc:
Marking as Draft so #985 can go in first, for easier resolution.

irahopkinson (approved):
👍 Ok, LGTM.

marksvc (commented):
Rebased over script rename. Slightly cleaned up quotation, for linter.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/969#-:-MW_vf52E8l-tn4RM2go:b-4q8voe)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[scripts/build-and-upload-beta.sh, line 9 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/969#-MVM0T5L4tEeXpsfIYde:-MW_vbeaEjmUIDnNonzq:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b3eb0bcee3d1e2c4909b991f9d8b12cce186df3/scripts/build-and-upload-beta.sh#L9)):*

I'm good with rsyncing the whole folder. I still think we need to check it by manually running the command from the build server first until AuthProgs is good.
</blockquote></details>

Done.

irahopkinson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/969#-:-MWcYt3i7n8-6K7xk4Tj:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #973, opened 2021-03-10
SF-1181b Fix 401 error on connect project page

Nateowami:
This isn't a very elegant solution, but I'm really not sure which abstraction layer should be responsible for checking whether we should proceed. One option is to allow an exception to be thrown, then catch it high up the stack. That doesn't seem very elegant either.

I would welcome arguments/reasons for why this should be done differently.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/973)

### DIFF
--- a/src/SIL.XForge.Scripture/Controllers/ParatextController.cs
+++ b/src/SIL.XForge.Scripture/Controllers/ParatextController.cs
@@ -66,6 +66,10 @@ namespace SIL.XForge.Scripture.Controllers
                 var resources = await _paratextService.GetResourcesAsync(_userAccessor.UserId);
                 return Ok(resources.ToDictionary(r => r.ParatextId, r => r.Name));
             }
+            catch (DataNotFoundException)
+            {
+                return NoContent();
+            }
             catch (SecurityException)
             {
                 return NoContent();
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -846,18 +846,28 @@ namespace SIL.XForge.Scripture.Services
             SemaphoreSlim semaphore = _tokenRefreshSemaphores.GetOrAdd(userId, (string key) => new SemaphoreSlim(1, 1));
             await semaphore.WaitAsync();
 
-            Attempt<UserSecret> attempt = await _userSecretRepository.TryGetAsync(userId);
-            if (!attempt.TryResult(out UserSecret userSecret))
+            try
             {
-                throw new DataNotFoundException("Could not find user secrets for " + userId);
-            }
+                Attempt<UserSecret> attempt = await _userSecretRepository.TryGetAsync(userId);
+                if (!attempt.TryResult(out UserSecret userSecret))
+                {
+                    throw new DataNotFoundException("Could not find user secrets for " + userId);
+                }
 
-            if (!userSecret.ParatextTokens.ValidateLifetime())
-            {
-                Tokens refreshedUserTokens = await _jwtTokenHelper.RefreshAccessTokenAsync(_paratextOptions.Value, userSecret.ParatextTokens, _registryClient);
-                userSecret = await _userSecretRepository.UpdateAsync(userId, b => b.Set(u => u.ParatextTokens, refreshedUserTokens));
+                if (!userSecret.ParatextTokens.ValidateLifetime())
+                {
+                    Tokens refreshedUserTokens = await _jwtTokenHelper.RefreshAccessTokenAsync(_paratextOptions.Value, userSecret.ParatextTokens, _registryClient);
+                    userSecret = await _userSecretRepository.UpdateAsync(userId, b => b.Set(u => u.ParatextTokens, refreshedUserTokens));
+                }
+                return new ParatextAccessLock(semaphore, userSecret);
             }
-            return new ParatextAccessLock(semaphore, userSecret);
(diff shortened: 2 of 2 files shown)


Nateowami (commented):
@irahopkinson When a user visits the connect project page it tries to fetch resources. If the user does not have user secrets, it causes an error. This is the issue the testers were running into.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/973#-:-MVRm0dyA9fBi3J1K0dN:bsexnzx)* status: 0 of 1 files reviewed, all discussions resolved

ddaspit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/973)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/Controllers/ParatextController.cs, line 64 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/973#-MVUU1-WBfcuXdB6p73b:-MVUU1-WBfcuXdB6p73c:b-g5trpt) ([raw file](https://github.com/sillsdev/web-xforge/blob/93780f8bb56914bf0179ec4c56e672c450d5dc7d/src/SIL.XForge.Scripture/Controllers/ParatextController.cs#L64)):*
> ```Smalltalk
>         public async Task<ActionResult<Dictionary<string, string>>> ResourcesAsync()
>         {
>             Attempt<UserSecret> attempt = await _userSecrets.TryGetAsync(_userAccessor.UserId);
> ```

We do have a `DataNotFoundException` that would be appropriate. There are examples where the `ParatextService` throws that exception, the controller catches it, and returns the appropriate status code.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/973)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[src/SIL.XForge.Scripture/Controllers/ParatextController.cs, line 64 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/973#-MVUU1-WBfcuXdB6p73b:-MVYHckKEQkWvVosdR6r:b60801s) ([raw file](https://github.com/sillsdev/web-xforge/blob/93780f8bb56914bf0179ec4c56e672c450d5dc7d/src/SIL.XForge.Scripture/Controllers/ParatextController.cs#L64)):*

We do have a `DataNotFoundException` that would be appropriate. There are examples where the `ParatextService` throws that exception, the controller catches it, and returns the appropriate status code.
</blockquote></details>

Thanks. I think this is much better. I also fixed a related issue where the throwing of an exception caused the semaphore to not be released.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/973)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 864 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/973#-MVt0UFeENmkUFjMmTWc:-MVt0UFeENmkUFjMmTWd:b6ha71l) ([raw file](https://github.com/sillsdev/web-xforge/blob/8132317cdad70aaf541853e2aa9691d8ab8d34c3/src/SIL.XForge.Scripture/Services/ParatextService.cs#L864)):*
> ```Smalltalk
>                 return new ParatextAccessLock(semaphore, userSecret);
>             }
>             catch
> ```

It would be better to use a `finally` block here instead of `catch`.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/973)* status: all files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 864 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/973#-MVt0UFeENmkUFjMmTWc:-MVt3kKgAJP69U9HVoA_:b-h1ihdd) ([raw file](https://github.com/sillsdev/web-xforge/blob/8132317cdad70aaf541853e2aa9691d8ab8d34c3/src/SIL.XForge.Scripture/Services/ParatextService.cs#L864)):*

It would be better to use a `finally` block here instead of `catch`.
</blockquote></details>

A `finally` block would always execute. This needs to execute only when an exception is thrown and the user will not have the opportunity to release the semaphore.

ddaspit (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/973)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 864 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/973#-MVt0UFeENmkUFjMmTWc:-MVu0PEVE3QcWD8hvQHv:b9c7emw) ([raw file](https://github.com/sillsdev/web-xforge/blob/8132317cdad70aaf541853e2aa9691d8ab8d34c3/src/SIL.XForge.Scripture/Services/ParatextService.cs#L864)):*

A `finally` block would always execute. This needs to execute only when an exception is thrown and the user will not have the opportunity to release the semaphore.
</blockquote></details>

I see. The semaphore is supposed to released when `ParatextAccessLock` is disposed. This code ensures that it is released if an exception is thrown before the `ParatextAccessLock` is created.

PR #985, opened 2021-03-18
Deploy: Rename and comb deploy script

marksvc:
**This PR involves both a rename and broad changes to a file, so it may be easier to review the two commits individually.**

## Deploy: Comb deploy script

- Disallow using unassigned variables (`set -u`).
- Fail on errors (`set -eo pipefail`). There were already a couple
usages of `|| exit`, but it appears that any error in the script would
be a time to stop rather than continue.
- Style: Surround variables in braces.
- Use double quotes around variables. This is clearly helpful for
paths, which can sometimes have spaces. But the linter was complaining
that any un-quoted variable use could unintentionally result in
globbing or word splitting.
- Break lines within column 120.
- We were `popd`ing without use. I removed that, and changed the
somewhat more elaborate `pushd` to just `cd`, which is less noisy.
- This was a good time to make broad changes to the file when about to
rename it anyway (and potentially cause problems or merge upsets).

## Deploy: Rename deployment script

- Remove 'beta' from script name since it's used for more than just
beta releases.
- Style: Don't use .sh for executable script filenames.
- For that matter, chmod +x the script. I don't know how the CI was
executing it. Maybe it was starting it as an argument to bash.
- 'ship' is somewhat more clear what we are doing than 'upload', as
it's not just being uploaded, it's being uploaded to a place that will
cause it to be immediately made available for use. 'build-and-ship' is
more clear that it's being made available to users than
'build-and-deploy' (which could be local). Since this script is also
used to publish software to QA, 'ship' may or may not be the right
term for that; tho QA mimics production and so using a shipping script
but to QA may be sensible.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/985)


scripts/build-and-upload-beta.sh
```diff
@@ -1,46 +1,48 @@
...
 fi
 
-pushd .. > /dev/null
+cd ..
 
-# must be before `ng build` because the constant value is inluded during optimization
+# Must be before `ng build` because the constant value is inluded during optimization
 cat <<EOF > src/SIL.XForge.Scripture/version.json
 {
   "version": "$BUILD_NUMBER"
```
rmunn: Do you want to add braces around this one too?

rmunn (approved):
> "Style: Don't use .sh for executable script filenames"

Is this personal preference, or a *de facto* standard that I wasn't aware of? What I'm used to seeing is that scripts in `/usr/bin` or `/usr/local/bin` will have no extension, but scripts that live in a repository will have a `.sh` extension (or `.py`, or `.cmd` on Windows...). Feels to me like this script should keep a `.sh` extension. I won't block the PR on that, though. LGTM overall.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/985#-:-MWReXwU3Mss_ET8JwKk:b32s7wd)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

___
*[deploy/build-and-ship, line 17 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/985#-MWRf7Fm37QF1ehgz_QE:-MWRf7Fm37QF1ehgz_QF:b5uuxv5) ([raw file](https://github.com/sillsdev/web-xforge/blob/50939ec1bf1f502440bc41e5314bb700edd3bb10/deploy/build-and-ship#L17)):*
> ```
> cd ..
> 
> # Must be before `ng build` because the constant value is inluded during optimization
> ```

This is the "inluded" typo.
___
*[deploy/build-and-ship, line 20 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/985#-MWReQJt6XKKm06gN7qa:-MWReQJt6XKKm06gN7qb:b-jbt6s7) ([raw file](https://github.com/sillsdev/web-xforge/blob/50939ec1bf1f502440bc41e5314bb700edd3bb10/deploy/build-and-ship#L20)):*
> ```
> cat <<EOF > src/SIL.XForge.Scripture/version.json
> {
>   "version": "$BUILD_NUMBER"
> ```

This is the spot where I was suggesting braces.

irahopkinson (changes requested):
I just noticed the move into the `deploy` folder. Please don't do that as they aren't related enough. The `deploy` folder is probably poorly named but is historical and contains Ansible scripts to 'deploy' the setup to a machine. Not to deploy/release the built app to a server (which would first need an Ansible 'deploy' to set it up).

marksvc (commented):
Robin:

>> Style: Don't use .sh for executable script filenames
> preference, or .. standard..?

Usage is divided. I think the strongest reason not to use .sh for bash program filenames is that it's _ugly_. We could have (for example in FieldWorks) a script  /usr/bin/fieldworks-flex.sh run  /usr/lib/fieldworks/run-app.sh which runs /usr/lib/fieldworks/setup-user.sh , but it's all prettier without `.sh`. 
I'm not as accustomed to Python and Perl, but I would think user-facing scripts in those languages would also be prettier named `process-stuff` rather than `process-stuff.py` or `process-stuff.pl`.
I wrote a Xonsh script recently, but just named it foo; someone would probably be scared to type `./foo.xonsh` :-)
This 'prettier' reason is easily argued against.

Also, I find /bin/sh to be rather limiting and much prefer using bash. But I can't imagine naming my files foo.bash. It seems strange to append '.sh' to the filename of a script that isn't actually written in sh.

I looked for other style guides today, but the only one I found that prescribes for or against using .sh is Google's Bash style guide, which I think has some sensible advice that could be similarly applied to Xonsh, Python, and Perl as well:

> Executables should have no extension (strongly preferred) or a .sh extension. Libraries must have a .sh extension and should not be executable.
> 
> It is not necessary to know what language a program is written in when executing it and shell doesn’t require an extension so we prefer not to use one for executables.
> 
> However, for libraries it’s important to know what language it is

From https://google.github.io/styleguide/shellguide.html#file-extensions .

I also don't agree with following every point of that style guide (nor have I learned enough bash to do all the right ways), but it's a useful resource.

My idealizations don't have to win out here if others disagree. Have I persuaded you? I think I'd like to keep the .sh extension here if that's what you think it should have. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/985#-:-MWUp5fJ6ZYhmrApl281:bvc57sa)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

___
*[deploy/build-and-ship, line 17 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/985#-MWRf7Fm37QF1ehgz_QE:-MWUmPJLA1ljQoneZIvn:bb6un1j) ([raw file](https://github.com/sillsdev/web-xforge/blob/50939ec1bf1f502440bc41e5314bb700edd3bb10/deploy/build-and-ship#L17)):*
<details><summary><i>Previously, rmunn (Robin Munn) wrote…</i></summary><blockquote>

This is the "inluded" typo.
</blockquote></details>

Done
___
*[deploy/build-and-ship, line 20 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/985#-MWReQJt6XKKm06gN7qa:-MWUmtH-AQ5YPF8fqsS5:b-jrhlhs) ([raw file](https://github.com/sillsdev/web-xforge/blob/50939ec1bf1f502440bc41e5314bb700edd3bb10/deploy/build-and-ship#L20)):*
<details><summary><i>Previously, rmunn (Robin Munn) wrote…</i></summary><blockquote>

This is the spot where I was suggesting braces.
</blockquote></details>

Well spotted, thank you!

marksvc (commented):
**Ira** writes 
> I just noticed the move into the deploy folder. Please don't do that as they aren't related enough. The deploy folder is probably poorly named

**Ira**: Will do. So do take note that this PR also renames the script, so this is a good time to disagree with the new name if you don't like it .
I was glad to move the only file out of /scripts so another top-level item could disappear :-).

I would advocate that we re-purpose the `deploy` folder for our various deploy operations. We deploy to servers. We provision and deploy to workstations. We have a vagrant machine that can be spun up. All these setup and install things could live together under `deploy`. 
I will move the script back into the `/scripts` directory.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/985#-:-MWUy1qV2WZ6QfikfL6w:bgi5uuf)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

irahopkinson (approved):
:lgtm: the filename much better. Thanks for taking the initiative on these changes.

If you wanted to clean up then this would be OK: move existing `deploy/` into `deploy/ansible/`; move `scripts/` into `deploy/scripts/`. But if you do, put that in a separate PR to this one.

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/985#-:-MWXQArNBNpI-RK-6fzt:b-rdp5xi)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #993, opened 2021-03-24
SF-1237 Wait for user to come online before checking if needs migration

marksvc:
Checking if the user needs migration, while offline, results in an error dialog. Don't do it.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/993)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -1,3 +1,4 @@
+import { MdcDialog } from '@angular-mdc/web';
 import { MdcList, MdcListItem } from '@angular-mdc/web/list';
 import { CommonModule, Location } from '@angular/common';
 import { HttpClientTestingModule } from '@angular/common/http/testing';
@@ -49,7 +50,8 @@ const mockedLocationService = mock(LocationService);
 const mockedNoticeService = mock(NoticeService);
 const mockedPwaService = mock(PwaService);
 const mockedFileService = mock(FileService);
-const mockErrorReportingService = mock(ErrorReportingService);
+const mockedErrorReportingService = mock(ErrorReportingService);
+const mockedMdcDialog = mock(MdcDialog);
 
 @Component({
   template: `<div>Mock</div>`
@@ -90,7 +92,8 @@ describe('AppComponent', () => {
       { provide: NoticeService, useMock: mockedNoticeService },
       { provide: PwaService, useMock: mockedPwaService },
       { provide: FileService, useMock: mockedFileService },
-      { provide: ErrorReportingService, useMock: mockErrorReportingService }
+      { provide: ErrorReportingService, useMock: mockedErrorReportingService },
+      { provide: MdcDialog, useMock: mockedMdcDialog }
     ]
   }));
 
@@ -287,8 +290,13 @@ describe('AppComponent', () => {
     const env = new TestEnvironment();
     env.init();
 
-    verify(mockErrorReportingService.addMeta(anything(), 'user')).once();
+    verify(mockedErrorReportingService.addMeta(anything(), 'user')).once();
     expect().nothing();
+  }));
+
+  it('waits for the user to be online and then migrates data', fakeAsync(() => {
+    const env = new TestEnvironment();
+    env.init();
   }));
 
   describe('Community Checking', () => {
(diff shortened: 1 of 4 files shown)


irahopkinson (approved):
Looks good, just 2 non-blocking comments. Merge as you see fit.

> Reviewed 2 of 4 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/993#-:-MWcLZAMBPQW13MS9vqU:b-wttadc)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 70 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/993#-MWcFaLg3H6v8VoYhTuX:-MWcFaLg3H6v8VoYhTuY:b-edd4vy) ([raw file](https://github.com/sillsdev/web-xforge/blob/19043641bb9e9243ec69b302745fd417bc85f26a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L70)):*
> ```
> private projectDeletedDialogRef: MdcDialogRef<ProjectDeletedDialogComponent> | null = null;
> ```
What is the reason this needs to be set to `null` rather than `undefined`? The code later sets it to `null` if there is no `projectId` but then it is defined but set to `null`. This seems like a potentially useful distinction to keep.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts, line 222 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/993#-MWcIyVo7Abo_3VRq1sI:-MWcIlSL20Q7Yv2ClR3l:b-jfpfsx) ([raw file](https://github.com/sillsdev/web-xforge/blob/19043641bb9e9243ec69b302745fd417bc85f26a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts#L222)):*
> ```
> projectDeletedDialogRefAfterClosed$.next('close')
> ```
It might be better to retain the `confirmDialog` method an put this inside. More readable here and only one place to refactor if something changes about the dialog.

marksvc (commented):
Thanks. I'll merge now to get this in and send a followup as a new PR.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/993#-:-MWeu4j7BivMAKkif6O6:b3xo6u1)* status: all files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 70 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/993#-MWcFaLg3H6v8VoYhTuX:-MWeqioFDvkj5ZWRwd8V:bbpqylm) ([raw file](https://github.com/sillsdev/web-xforge/blob/19043641bb9e9243ec69b302745fd417bc85f26a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L70)):*
My motivation was that I don't like leaving the type specified as `any`. (Because then we don't benefit from static type checking, and also sometimes I don't get intellisense on something, and `any` might be why sometimes.)

If I just set the type to `MdcDialogRef<ProjectDeletedDialogComponent>` (and don't define it when declared), then I get error

> Property 'projectDeletedDialogRef' has no initializer and is not definitely assigned in the constructor.ts(2564)

If I set the type to `MdcDialogRef<ProjectDeletedDialogComponent> | undefined`, then 

later with line `this.projectDeletedDialogRef = null;` I get error 

> Type 'null' is not assignable to type 'MdcDialogRef<ProjectDeletedDialogComponent, any> | undefined'.ts(2322)

I could set the type to `MdcDialogRef<ProjectDeletedDialogComponent> | undefined | null` and not assign it, letting it start off as undefined, and later be set to null. 

But I went with 

```
projectDeletedDialogRef: MdcDialogRef<ProjectDeletedDialogComponent>  | null=null;
```

which is more narrow.

What would you do here? Do you think 
```
private projectDeletedDialogRef?: MdcDialogRef<ProjectDeletedDialogComponent>  | null;
```
(i.e. all three of undefined | null | MdcDialogRef<..>)
is preferable? 
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts, line 222 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/993#-MWcIyVo7Abo_3VRq1sI:-MWeu1mODMOTj2ppNqC-:b-19jje4) ([raw file](https://github.com/sillsdev/web-xforge/blob/19043641bb9e9243ec69b302745fd417bc85f26a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts#L222)):*

> ```
> projectDeletedDialogRefAfterClosed$.next('close')
> ```
It might be better to retain the `confirmDialog` method an put this inside. More readable here and only one place to refactor if something changes about the dialog.
</blockquote></details>

Thanks. Doing in a followup commit.

PR #995, opened 2021-03-25
SF-1237 Refactor test method

marksvc:
----
This is a followup from a prior PR's comments.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/995)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -219,7 +219,7 @@ describe('AppComponent', () => {
     env.deleteProject('project01', false);
     verify(mockedMdcDialog.open(ProjectDeletedDialogComponent)).once();
     verify(mockedUserService.setCurrentProjectId()).once();
-    env.projectDeletedDialogRefAfterClosed$.next('close');
+    env.confirmProjectDeletedDialog();
     // Get past setTimeout to navigation
     tick();
     env.fixture.detectChanges();
@@ -247,7 +247,7 @@ describe('AppComponent', () => {
     expect(env.selectedProjectId).toEqual('project01');
     env.removesUserFromProject('project01');
     verify(mockedMdcDialog.open(ProjectDeletedDialogComponent)).once();
-    env.projectDeletedDialogRefAfterClosed$.next('close');
+    env.confirmProjectDeletedDialog();
     // Get past setTimeout to navigation
     tick();
     expect(env.location.path()).toEqual('/projects');
@@ -734,6 +734,10 @@ class TestEnvironment {
     const projectDoc = this.realtimeService.get<SFProjectDoc>(SFProjectDoc.COLLECTION, projectId);
     projectDoc.submitJson0Op(op => op.set<string>(p => p.userRoles['user01'], SFProjectRole.CommunityChecker), false);
     this.currentUserDoc.submitJson0Op(op => op.add<string>(u => u.sites['sf'].projects, 'project04'), false);
+  }
+
+  confirmProjectDeletedDialog() {
+    this.projectDeletedDialogRefAfterClosed$.next('close');
   }
 
   private addProject(projectId: string, userRoles: { [userRef: string]: string }, texts: TextInfo[]): void {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/project-deleted-dialog/project-deleted-dialog.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/project-deleted-dialog/project-deleted-dialog.component.html
@@ -4,7 +4,7 @@
       <mdc-dialog-surface>
         <mdc-dialog-title>{{ t("project_has_been_deleted") }}</mdc-dialog-title>
         <mdc-dialog-actions>
-          <button id="ok-button" mdcDialogButton type="button" mdcDialogAction="accept">{{ t("ok") }}</button>
+          <button id="ok-button" mdcDialogButton type="button" mdcDialogAction="close">{{ t("ok") }}</button>
         </mdc-dialog-actions>
       </mdc-dialog-surface>
     </mdc-dialog-container>


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/995)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts, line 740 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/995#-MWgBVIZ0MKBC8DOMvrp:-MWgBVIZ0MKBC8DOMvrq:bjxb83e) ([raw file](https://github.com/sillsdev/web-xforge/blob/c30e5fd0dd115e283f2e091edd9082ab7fef63d0/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts#L740)):*
> ```XML
> 
>   confirmProjectDeletedDialog() {
>     this.projectDeletedDialogRefAfterClosed$.next('accept');
> ```

Before the string was `'close'`, now it is `'accept'`. These seem to be "magic strings" that would be referenced elsewhere, but they aren't. Reading the source for the app component, this value isn't used. Based on testing, `this.projectDeletedDialogRefAfterClosed$.next()` should be sufficient.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/995)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts, line 740 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/995#-MWgBVIZ0MKBC8DOMvrp:-MWgE61UAeBc1rolMrms:b-n7r04l) ([raw file](https://github.com/sillsdev/web-xforge/blob/c30e5fd0dd115e283f2e091edd9082ab7fef63d0/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts#L740)):*

Before the string was `'close'`, now it is `'accept'`. These seem to be "magic strings" that would be referenced elsewhere, but they aren't. Reading the source for the app component, this value isn't used. Based on testing, `this.projectDeletedDialogRefAfterClosed$.next()` should be sufficient.
</blockquote></details>

Let's try this again:
``` suggestion
    this.projectDeletedDialogRefAfterClosed$.next();
```

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/995)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts, line 740 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/995#-MWgBVIZ0MKBC8DOMvrp:-MWkYTGc3AGp3nZNXveQ:b-60izdg) ([raw file](https://github.com/sillsdev/web-xforge/blob/c30e5fd0dd115e283f2e091edd9082ab7fef63d0/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts#L740)):*

Let's try this again:
``` suggestion
    this.projectDeletedDialogRefAfterClosed$.next();
```
</blockquote></details>

I think 'accept' more closely follows what the actual app is doing. In src/SIL.XForge.Scripture/ClientApp/src/app/project-deleted-dialog/project-deleted-dialog.component.html there is 
```html
        <mdc-dialog-actions>
          <button id="ok-button" mdcDialogButton type="button" mdcDialogAction="accept">{{ t("ok") }}</button>
        </mdc-dialog-actions>
```
I haven't attached a debugger to verify that for sure it's returning 'accept', but I assume it is based on similar situations with other template code in our repo.
I agree that `.next()` is sufficient for performing the test. But it seems to be a better mock if the mock project deleted dialog returns 'accept' when closed. 
Yeah it is a magic string. I suppose I could prepend a line 
```ts
const closeDialogAction='accept';
```
and return `closeDialogAction` instead of `'accept`'.

Hmm. 

Does this change your opinion?

Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/995#-:-MWz0x5e8CeBnB-g4uKl:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts, line 740 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/995#-MWgBVIZ0MKBC8DOMvrp:-MWywhsY2EoBS_6fTUXB:bvendix) ([raw file](https://github.com/sillsdev/web-xforge/blob/c30e5fd0dd115e283f2e091edd9082ab7fef63d0/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts#L740)):*

I think 'accept' more closely follows what the actual app is doing. In src/SIL.XForge.Scripture/ClientApp/src/app/project-deleted-dialog/project-deleted-dialog.component.html there is 
```html
        <mdc-dialog-actions>
          <button id="ok-button" mdcDialogButton type="button" mdcDialogAction="accept">{{ t("ok") }}</button>
        </mdc-dialog-actions>
```
I haven't attached a debugger to verify that for sure it's returning 'accept', but I assume it is based on similar situations with other template code in our repo.
I agree that `.next()` is sufficient for performing the test. But it seems to be a better mock if the mock project deleted dialog returns 'accept' when closed. 
Yeah it is a magic string. I suppose I could prepend a line 
```ts
const closeDialogAction='accept';
```
and return `closeDialogAction` instead of `'accept`'.

Hmm. 

Does this change your opinion?

</blockquote></details>

Ah, I see. Well then I guess the question would be why `project-deleted-dialog.component.html` specifies an action string that is not actually used elsewhere (I think the entire purpose of the `mdcDialogAction` directive is to return a value from the dialog).

My preference would be to remove it from the dialog template altogether since it's not used, but :lgtm: either way.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/995)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts, line 740 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/995#-MWgBVIZ0MKBC8DOMvrp:-MWzX88sB_csEJPdO9Hd:b-fuxo4a) ([raw file](https://github.com/sillsdev/web-xforge/blob/c30e5fd0dd115e283f2e091edd9082ab7fef63d0/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts#L740)):*
So I had a look in some documentation. It looks like 'accept' and 'close' are special, as seen around the end of https://material.src.zone/components/dialog , triggering events for accept and cancel. From https://material.io/components/dialogs/web I unfortuntaly didn't see an example of what to do if you have a dialog with only one button, but I did observe that data-mdc-dialog-action attributes were all having settings. The API there does show that the action value can be a string or null. So not having any specific action specified is accounted for in the API.

I also observed 
> Any action buttons within the dialog which equate strictly to a dismissal with no further action should also use the close action; this will make it easy to handle all such interactions consistently, while separately handling other actions.

which tells me that 'close' was correct and I shouldn't have changed it to 'accept'. Instead, I think I should change the html template as 
```diff
-mdcDialogAction="accept"
+mdcDialogAction="close"
```

I think after looking over some documentation today I don't really want to remove the action string, even though it's not specifically looked for in the component code. What do you think? Maybe we should phone about it.

marksvc:
(Talked offline. Merging)

PR #999, opened 2021-04-01
Require tests to clean up overlay

marksvc:
- Tests leave dialog and other content in the HTML overlay. This can
cause problems to following tests as they look for the presence or
properties of elements. Whether it's a problem can vary by the order
the tests are run in.
- test-utils.ts: Require that tests clean up the overlay by at least
removing added content (if not the whole overlay). Throw an error for
any test that does not clean up. Note that this only applies for tests
that use configureTestingModule.
- Adjust existing tests so that they clean up the overlay, such as by
clicking a dialog Close button.
- delete-project-dialog.component.spec.ts: Add an additional check
that the correct dialog result occurred, since more than one dialog
button was clicked.
- Add some expect().nothing().
- import-questions-confirmation-dialog.component.spec.ts: Use async
and await rather than just .then to ensure the closing processes.

Co-authored-by: MarkS <marksvc@gmail.com>

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/999)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -356,6 +356,7 @@ describe('AppComponent', () => {
     // SUT2 is in ngOnInit()
     tick();
     verify(mockedMdcDialog.open(BetaMigrationDialogComponent, anything())).once();
+    expect().nothing();
   }));
 
   describe('Community Checking', () => {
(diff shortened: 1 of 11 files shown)


Nateowami (approved):
:lgtm:

> Reviewed 11 of 11 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/999#-:-MXDsxi_C0uDL_Mvlhon:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1000, opened 2021-04-01
More accurately mock onlineGetLinkSharingKey

Nateowami:
Fixes test error using null linkSharingKey. The tests were not properly mocking the project service.

I also changed how we check for an empty string:
``` diff
- if (this.linkSharingKey.length < 1) {
+ if (this.linkSharingKey === ''") {
```
As far as I can tell, these are nearly identical. The main difference is that the latter will not crash when `linkSharingKey` is null or undefined.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1000)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts
@@ -76,7 +76,7 @@ export class ShareControlComponent extends SubscriptionDisposable implements Aft
   }
 
   get shareLink(): string {
-    if (this.linkSharingKey.length < 1) {
+    if (this.linkSharingKey === '') {
       return '';
     }
     return (
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share.component.spec.ts
@@ -113,6 +113,7 @@ class TestEnvironment {
     when(mockedProjectService.get('project01')).thenCall(() =>
       this.realtimeService.subscribe(SFProjectDoc.COLLECTION, 'project01')
     );
+    when(mockedProjectService.onlineGetLinkSharingKey(anything(), anything())).thenResolve('shareKey');
     when(mockedUserService.currentUserId).thenReturn('user01');
     when(mockedPwaService.isOnline).thenReturn(true);
     when(mockedPwaService.onlineStatus).thenReturn(of(true));


irahopkinson (changes requested):
Just one question.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1000#-:-MXGdhSs1BrZ2pMakWsO:b-w58e4h)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts, line 79 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1000#-MXGdViA67R-G1BtDgv0:-MXGdViA67R-G1BtDgv1:b-hfusrj) ([raw file](https://github.com/sillsdev/web-xforge/blob/19a55953e37b937e48f4328bce8c313fcdead02a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts#L79)):*
> ```
> if (this.linkSharingKey === '') {
> ```
Ok, good find. Shouldn't you also return an empty string if the `this.linkSharingKey` is null or undefined?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1000)* status: all files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts, line 79 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1000#-MXGdViA67R-G1BtDgv0:-MXI6UqQ9O88ovyyeWb7:b-kdje9y) ([raw file](https://github.com/sillsdev/web-xforge/blob/19a55953e37b937e48f4328bce8c313fcdead02a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts#L79)):*

> ```
> if (this.linkSharingKey === '') {
> ```
Ok, good find. Shouldn't you also return an empty string if the `this.linkSharingKey` is null or undefined?
</blockquote></details>

According to the type definition higher in the file, `this.linkSharingKey` *must* be a string. The value comes from `this.projectService.onlineGetLinkSharingKey(this._projectId, role)`, which shouldn't be able to return null (if it can't get an actual value that should be an exception, not a silent return of an incorrect value). However, it returned null in the tests because the method hadn't been mocked.

Granted, it might make more sense to use null rather than the empty string for when this value isn't set, and then checking for the empty string would be unnecessary.

irahopkinson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1000#-:-MXIHG4b1Icag4NPJNTS:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts, line 79 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1000#-MXGdViA67R-G1BtDgv0:-MXIH5Q54pMuvpPXtFsm:b-7aitaz) ([raw file](https://github.com/sillsdev/web-xforge/blob/19a55953e37b937e48f4328bce8c313fcdead02a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts#L79)):*

According to the type definition higher in the file, `this.linkSharingKey` *must* be a string. The value comes from `this.projectService.onlineGetLinkSharingKey(this._projectId, role)`, which shouldn't be able to return null (if it can't get an actual value that should be an exception, not a silent return of an incorrect value). However, it returned null in the tests because the method hadn't been mocked.

Granted, it might make more sense to use null rather than the empty string for when this value isn't set, and then checking for the empty string would be unnecessary.
</blockquote></details>

Thanks for clarifying.

PR #1002, opened 2021-04-05
Allow route index.html

marksvc:
- The pwaTest frontend configuration results in requests to
index.html. This produces error:
> Accessing System.InvalidOperationException: The SPA default page
> middleware could not return the default page '/index.html' because
> it was not found, and no other middleware handled the request.

- I added this only to the dev route set since we haven't seen this be
a problem on the server.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1002)

### DIFF
--- a/src/SIL.XForge.Scripture/Startup.cs
+++ b/src/SIL.XForge.Scripture/Startup.cs
@@ -37,7 +37,8 @@ namespace SIL.XForge.Scripture
             "vendor.js", "vendor.js.map",
             "main.js", "main.js.map",
             "manifest.json",
-            "sockjs-node"
+            "sockjs-node",
+            "index.html"
         };
         // examples of filenames are "main-es5.4e5295b95e4b6c37b696.js", "styles.a2f070be0b37085d72ba.css"
         private static readonly HashSet<string> ProductionSpaGetRoutes = new HashSet<string>


RaymondLuong3 (commented):
What are the instructions to replicate this. I tried building the frontend using `ng build --configuration=pwaTest` but I didn't see any issues when I opened it from the installed app or in the browser.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1002#-:-MXZ-_WT1YYUW4AUt8xm:b-u66qvu)* status: 0 of 1 files reviewed, all discussions resolved

Nateowami (commented):
Why is it requesting `/index.html`? Is it not caused by #1003? It doesn't seem like a safe assumption to assume it only happens on dev, just because it's only been seen there.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1002#-:-MXZ-tUS4KuCJP92t3n6:bmw2upg)* status: 0 of 1 files reviewed, all discussions resolved

marksvc:
@RaymondLuong3  Yeah, that's right. And then run the backend with `dotnet watch run --start-ng-serve=no`. Going to localhost:5000/projects then gives me the error. Interesting that it isn't happening for you tho. Unless you didn't do --start-ng-server=no ?

@Nateowami No, it's not caused by #1003. I was wondering if these two PRs would look related :). But they aren't. I did this one before applying #1003. 
With some more investigating, I see that it's probably needed because ngsw-config.json specifies it in two places:
```
"index": "/index.html",
  "assetGroups": [
    {
      "name": "app",
      "installMode": "prefetch",
      "resources": {
        "files": ["/favicon.ico", "/index.html", "/*.css", "/*.js", "/manifest.json"]
```
Compiling the pwaTest makes a dist/index.html file. I suspect it is how the SPA is delivered. Although.. then wouldn't it be broken in production?

marksvc:
> It doesn't seem like a safe assumption to assume it only happens on dev, just because it's only been seen there.

@Nateowami   Oh and what I mean is, we don't immediately get an ASP.NET error page when going to qa.scriptureforge.org/projects. It's true that there could be some case of this happening that has slipped by. But what I saw was a really significant blocking error.

Nateowami (commented):
I've run `ng build --configuration=pwaTest` followed by `dotnet watch run --start-ng-serve=no`. I'm not seeing an error when going to `/projects`, even after deleting the `dist` folder.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1002#-:-MXZqBzF2irHG_QAMiY6:b-rg2bj1)* status: 0 of 1 files reviewed, all discussions resolved

nigel-wells (commented):
@marksvc the error you got, was that in the context of the SPA or the razor pages? We technically have 2 service works running and the index will be different for both. The razor pages aren't cached by the service worker either except for the offline.html file.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1002#-:-MXZr15R19co9sKX638b:b89dmhn)* status: 0 of 1 files reviewed, all discussions resolved

marksvc:
Thanks everyone. The problem was I ran the backend and must have tried to use the web page before the frontend had fully built. Letting the frontend build all the way, and then running the backend, results in not having this PR's problem. I am closing the PR.

@nigel-wells I think I may have had it happen for both /projects and / , so that may mean both SPA and razor pages (tho I'm still growing in my understanding of the razor pages).

PR #1006, opened 2021-04-06
Show paratext notes in translate app

RaymondLuong3:
This is the ground work to support Paratext notes in the translate app.
*  notes are stored in mongo as a thread
*  a note thread doc contains some info about the context of the thread
*  Paratext comments are compared to note threads and changes applied to the note thread docs
*  the quill editor is registered with additional blots to style the display of notes
*  hovering over notes will show the content of the first note

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1006)


RaymondLuong3:
Made on the wrong branch.

PR #1019, opened 2021-04-14
SF-1232 Inform user when sharing link is unavailable offline

Nateowami:
The user should be informed when the link sharing URL is not available due to being offline. Here are some screenshots showing two scenarios, one where the page is loaded and then the user goes offline, and the other where the user is already offline when the page is loaded.

Online, then going offline (share link still available) | Offline at the point of visiting the page
-------------------------------------------------------------------|----------------------------------------------------
 | 

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1019)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.html
@@ -1,8 +1,15 @@
 <ng-container *transloco="let t; read: 'share_control'">
   <div fxLayout="column" fxLayoutGap="20px">
-    <div *ngIf="isLinkSharingEnabled">
+    <div *ngIf="isLinkSharingEnabled" class="invite-by-link">
       <div mdcSubtitle1>{{ t("link_sharing") }}</div>
-      <mdc-text-field id="share-link" #shareLinkField outlined [value]="shareLink" [readonly]="true">
+      <mdc-text-field
+        id="share-link"
+        #shareLinkField
+        outlined
+        [value]="shareLink"
+        [disabled]="!shareLink"
+        [readonly]="true"
+      >
         <mdc-icon
           id="share-link-copy-icon"
           mdcTextFieldIcon
@@ -13,8 +20,9 @@
           >file_copy
         </mdc-icon>
       </mdc-text-field>
+      <div *ngIf="showLinkSharingUnavailable" class="offline-text">{{ t("link_sharing_not_available_offline") }}</div>
     </div>
-    <div>
+    <div class="invite-by-email">
       <div mdcSubtitle1>{{ t("invite_people") }}</div>
       <div fxLayout="row" fxLayoutAlign="start center" fxLayoutGap="10px">
         <form fxFlex="100%" fxLayout.gt-xs="row" fxLayout.xs="column" id="email-form" [formGroup]="sendInviteForm">
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.scss
@@ -11,3 +11,7 @@ mdc-text-field {
   // add margin that is same height as the text field helper text, so the button is centered properly
   margin-bottom: 19px;
 }
+
+.invite-by-link .offline-text {
+  margin-top: 20px;
+}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts
@@ -162,11 +162,11 @@ describe('ShareControlComponent', () => {
     env.wait();
     expect(env.component.sendInviteForm.enabled).toEqual(true);
     expect((env.inputElement.nativeElement as HTMLInputElement).disabled).toEqual(false);
-    expect(env.offlineMessage).toBeNull();
+    expect(env.emailSharingOfflineMessage).toBeNull();
     env.onlineStatus = false;
     expect(env.component.sendInviteForm.enabled).toEqual(false);
     expect((env.inputElement.nativeElement as HTMLInputElement).disabled).toEqual(true);
-    expect(env.offlineMessage).not.toBeNull();
+    expect(env.emailSharingOfflineMessage).not.toBeNull();
   }));
 
   it('share link should be hidden if link sharing is turned off', fakeAsync(() => {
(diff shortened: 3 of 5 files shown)


RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1019)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts, line 180 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1019#-MYL5ngz0o0SVA1--rdT:-MYL5ngz0o0SVA1--rdU:b9ymlm1) ([raw file](https://github.com/sillsdev/web-xforge/blob/6fd220f87072c8143f7014e728335c6060b27e8e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts#L180)):*
> ```
>   it('share link should not be shown when offline', fakeAsync(() => {
> ```
You could add a new test where going offline from an online state keeps the link on display usable.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts, line 185 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1019#-MYL58TK0dUD_A-jpM1N:-MYL58TK0dUD_A-jpM1O:b-3fyvi4) ([raw file](https://github.com/sillsdev/web-xforge/blob/6fd220f87072c8143f7014e728335c6060b27e8e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts#L185)):*
> ```
>     expect(env.shareLink.nativeElement.value).toEqual('');
> ```
What is the reason for testing the value is equal to ''? I generally see we test the element is not null.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts, line 186 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1019#-MYL5M485b1AgCGlnJjJ:-MYL5M485b1AgCGlnJjK:bmhij0r) ([raw file](https://github.com/sillsdev/web-xforge/blob/6fd220f87072c8143f7014e728335c6060b27e8e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts#L186)):*
> ```
> expect(env.linkSharingOfflineMessageText).toContain('Inviting users by link sharing is unavailable while offline.');
> ```
Alternatively this could also test that the innerText of the element contains the message. Just one step down the chain.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1019)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts, line 180 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1019#-MYL5ngz0o0SVA1--rdT:-MYLo1ME8dSoyAZZMAUT:b84oz8w) ([raw file](https://github.com/sillsdev/web-xforge/blob/6fd220f87072c8143f7014e728335c6060b27e8e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts#L180)):*

> ```
>   it('share link should not be shown when offline', fakeAsync(() => {
> ```
You could add a new test where going offline from an online state keeps the link on display usable.
</blockquote></details>

I thought about it, but it didn't really seem worth it. This feature exists just because by default that's what it would do, not because we want to go out of our way to make it work. If it broke for some reason it would likely be because doing it this way was no longer trivially simple, and if that's the case, that would be sufficient reason to stop doing it anyway.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts, line 185 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1019#-MYL58TK0dUD_A-jpM1N:-MYLkCdZ-He5RfGV3Zr7:b-nf74rz) ([raw file](https://github.com/sillsdev/web-xforge/blob/6fd220f87072c8143f7014e728335c6060b27e8e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts#L185)):*

> ```
>     expect(env.shareLink.nativeElement.value).toEqual('');
> ```
What is the reason for testing the value is equal to ''? I generally see we test the element is not null.
</blockquote></details>

Because this is the input element that displays the share link. The element is always present; I just want to ensure the input field is blank.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts, line 186 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1019#-MYL5M485b1AgCGlnJjJ:-MYLk7rK-StFY3Wr1U8-:buqtep0) ([raw file](https://github.com/sillsdev/web-xforge/blob/6fd220f87072c8143f7014e728335c6060b27e8e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts#L186)):*

> ```
> expect(env.linkSharingOfflineMessageText).toContain('Inviting users by link sharing is unavailable while offline.');
> ```
Alternatively this could also test that the innerText of the element contains the message. Just one step down the chain.
</blockquote></details>

Done. Can't remember the exact reason for this.

RaymondLuong3 (approved):
Ok, looks good. Thanks.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1019#-:-MYM02q19FruXd_Ta7_5:b8wp49h)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts, line 180 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1019#-MYL5ngz0o0SVA1--rdT:-MYM-qN0CKwYoeC_JosZ:b76jmrb) ([raw file](https://github.com/sillsdev/web-xforge/blob/6fd220f87072c8143f7014e728335c6060b27e8e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts#L180)):*

I thought about it, but it didn't really seem worth it. This feature exists just because by default that's what it would do, not because we want to go out of our way to make it work. If it broke for some reason it would likely be because doing it this way was no longer trivially simple, and if that's the case, that would be sufficient reason to stop doing it anyway.
</blockquote></details>

Ok, it just happens to work that way, but is not necessarily the intended use case. That is fine.

PR #1023, opened 2021-04-22
- Only allow the migration component to run on beta

marksvc:
- Removed reliance on document referrer and instead did a frame check along with the migration path

This isn't full proof and it will stop being able to browse the site in beta mode but I believe we're beyond the need to do that now.

----
I tested this on my workstation and showed that it fixes the migration dialog loop in Firefox, and that Chromium still works ok.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1023)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
@@ -108,8 +108,8 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
       // Check authentication when coming back online
       // This is also run on first load when the websocket connects for the first time
       if (this.isAppOnline && !this.isAppLoading) {
-        // Redirect to the master site unless the referrer was the master site
-        if (environment.beta && !document.referrer.includes(environment.masterUrl)) {
+        // Redirect to the master site unless operating in the migration iframe
+        if (environment.beta && (window.self === window.top || this.locationService.pathname !== '/migration')) {
           this.locationService.go(environment.masterUrl + window.location.pathname);
         }
         this.authService.checkOnlineAuth();
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/beta-migration/beta-migration.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/beta-migration/beta-migration.component.ts
@@ -34,7 +34,7 @@ export class BetaMigrationComponent {
     const documentPromises: Promise<RealtimeDoc>[] = [];
 
     // Only run if we're in an iframe from master
-    if (window.self === window.top) {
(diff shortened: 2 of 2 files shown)


RaymondLuong3 (approved):
That's too bad the referrer isn't doing its job. :shipit: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1023#-:-MYuGESj848qSB35026U:bd6go6j)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/beta-migration/beta-migration.component.ts, line 37 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1023#-MYuFz1W4hRe2bbeG8p6:-MYuFz1W4hRe2bbeG8p7:bf3bglv) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc2b92f07c34c348baaa82bc49e9ce2e5e8b7449/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/beta-migration/beta-migration.component.ts#L37)):*
> ```
>    if (!environment.beta || window.self === window.top) {
> ```
I wouldn't have expected that we need to check the environment since it would have already been checked in app component, but if it works it works.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1023)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/beta-migration/beta-migration.component.ts, line 37 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1023#-MYuFz1W4hRe2bbeG8p6:-MYuK3FL1TEVZhJm_X-x:bxbvj5a) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc2b92f07c34c348baaa82bc49e9ce2e5e8b7449/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/beta-migration/beta-migration.component.ts#L37)):*

> ```
>    if (!environment.beta || window.self === window.top) {
> ```
I wouldn't have expected that we need to check the environment since it would have already been checked in app component, but if it works it works.
</blockquote></details>

I think this is regarding a situation where someone goes to stable-site/migration. This will make sure that that connection gets sent off to stable-site/projects. 

But a connection to beta-site/migration will be allowed at this place, unless window.self===window.top, which presumably means "is not in an iframe". 

So this check then means that someone can only go to /migration, and stay there, if it is on the beta-site, and the beta-site is loaded in an iframe.

Whereas the app.component check is doing.. let's see.
If (the site is beta, and (not in an iframe, or not at /migration) Then send to stable-site.

I _think_ there isn't an overlap. But it's complex :). And more importantly, it seems to work in practice.

PR #1025, opened 2021-04-27
SF-1276 - Auth0 timeout deletes offline data

nigel-wells:
- Ensure `checkSession` is only run once to avoid overlapping requests
- On error for auth0 trigger a `logIn` rather than `logOut`
- Report error to bugsnag for when renewing token fails

Although this task was primarily around triggering a login rather than a logout (which deletes data) I needed to find a way to trigger auth0 errors in order to test that it works. After understanding the errors I was able to resolve them at the same time.

One of the underlying issues with auth0 errors is when local storage `expires_at` is triggered at the same time the `tryOnlineLogIn` is triggered. Both call auth0 `checkSession` which generates a cookie and nonce in local storage and the end up stomping on each other. This occurrence can happen if someone has been logged in, shuts the browser down while still being logged in, and loads the site up again after the expiration time. This fix ensures only one query to auth0 is run at a time.

This should resolve errors such as:
```
Nonce (nonce) claim value mismatch in the ID token; expected \"F4X0t4y1drrAVByzl0ygHUrqbHO3b7XG\", found \"basQbzZv08EvMODEYCJ~SYDwX6_UOBj0\"
```
and hopefully 
```
`state` does not match.
```
as they are both stored in the same value in local storage.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1025)

### Acceptance tests
1. Login
2. Close the browser down
3. Wait more than 2 hours
4. Navigate to localhost:5000/projects
5. Previously you would either be logged out or you would get a login error dialog. Now, you will either notice nothing OR if it is one of the other auth0 errors not mentioned above then you'll be taken to auth0 to login (which may also send you back automatically if it says you are logged in).

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
@@ -54,6 +54,7 @@ export class AuthService {
   private tryLogInPromise: Promise<LoginResult>;
   private refreshSubscription?: Subscription;
   private renewTokenPromise?: Promise<void>;
+  private checkSessionPromise?: Promise<Auth0DecodedHash | null>;
 
   private readonly auth0 = new WebAuth({
     clientID: environment.authClientId,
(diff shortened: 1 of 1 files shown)


marksvc (approved):
I approve, tho I have a question about whether it will really be atomic.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1025#-:-MZK9D1G2qx1EwKaVPhX:b-eaxhmo)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 381 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1025#-MZK7ROaA74127md78Nf:-MZK7ROaA74127md78Ng:b-1bxf63) ([raw file](https://github.com/sillsdev/web-xforge/blob/25c90003fae014f2dee3ec8283f4244358cb8e00/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L381)):*
> ```XML
> 
>   private async checkSession(retryUponTimeout: boolean = true): Promise<Auth0DecodedHash | null> {
>     if (this.checkSessionPromise == null) {
> ```

This doesn't look to be atomic to guard against two items using the method at the same time.
It looks like `renewTokens()` is written the same way. 

So, I don't think this will eliminate two things running checkSession() at a time, but it will greatly reduce the likelihood that it will happen. 

Hmm, maybe that's not going to be a problem with JavaScript though?

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1025)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 381 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1025#-MZK7ROaA74127md78Nf:-MZKPmcZ3xtH1ocsq-MK:b-xfh4rs) ([raw file](https://github.com/sillsdev/web-xforge/blob/25c90003fae014f2dee3ec8283f4244358cb8e00/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L381)):*

This doesn't look to be atomic to guard against two items using the method at the same time.
It looks like `renewTokens()` is written the same way. 

So, I don't think this will eliminate two things running checkSession() at a time, but it will greatly reduce the likelihood that it will happen. 

Hmm, maybe that's not going to be a problem with JavaScript though?
</blockquote></details>

Yes a little confusing in the naming. The `checkSession` we don't want to be run at the same time is the `auth0.checkSession` one below. The local `checkSession` gets called from `tryOnlineLogin` and `renewTokens` which can occur in quick succession before auth0 has a chance to respond. I set it so the local `checkSession` can be called as many times as required and just `await` any promise that is currently in progress with auth0.

irahopkinson (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1025)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

PR #1037, opened 2021-05-10
Feature/realtime server esm

irahopkinson:
This is to remove a bunch of warnings during frontend compile about using CommonJS modues in Angular that "can cause optimization bailouts". To fix this we need to use ESM (ES Modules):
- It removes all of the warnings from our RealtimeServer by adding an ESM build to use from Angular. We also move the CommonJS build to its own 'lib' output folder so Node can continue to use it from our backend.
- It removes one external dependency warning by upgrading to a version that has ESM builds. There are still 4 external dependency warnings.


marksvc (approved):
I have reviewed the changes. I am not knowledgeable enough to say they are all what should be done, but they look like the right kinds of things. Good job changing the build to have two formats of module output. I understand that when we start using node 14 on the server, we should be able to use the ESM modules with node as well. 

I approve the change, but someone with more knowledge about this may want to have a look thru as well. I left the files as red/unreviewed to not make it harder for the next person.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1037#-:-M_M4zuo-PTQSO5Dk0dN:b-bxjzsn)* status: 0 of 109 files reviewed, all discussions resolved

irahopkinson (commented):
It will need a little bit more work for an ESM node. The top 2 level files are excluded from the ESM build currently because they are not used in Angular and it was taking too long to get them built in ESM and therefore out of scope for this fix. But definitely possible.

Apart from what you have already done for review, for this kind of infrastructure change, one useful review is to run it up and check that everything still works in case I missed committing a file or similar. I guess our TC tests are good enough for that so I will merge anyway.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1037#-:-M_P4tLGAVh8LXUDRUTw:bc8200g)* status: 0 of 109 files reviewed, all discussions resolved

PR #1042, opened 2021-05-17
SF-1297 Allow granting question permission to non admins

Nateowami:
---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1042)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -34,6 +34,7 @@
     "ldml",
     "libuv",
     "mailto",
+    "mingo",
     "mixins",
     "mockito",
     "nameof",
--- a/src/RealtimeServer/common/models/project-rights.ts
+++ b/src/RealtimeServer/common/models/project-rights.ts
@@ -1,23 +1,24 @@
 import { OwnedData } from './owned-data';
+import { Project } from './project';
 
 export enum Operation {
-  Create = 1,
-  Edit = 2,
-  Delete = 3,
-  View = 4,
+  Create = 'create',
+  Edit = 'edit',
+  Delete = 'delete',
+  View = 'view',
 
-  EditOwn = 5,
-  DeleteOwn = 6,
-  ViewOwn = 7
+  EditOwn = 'edit_own',
+  DeleteOwn = 'delete_own',
+  ViewOwn = 'view_own'
 }
 
 export interface ProjectRight {
-  projectDomain: number;
+  projectDomain: string;
   operation: Operation;
 }
 
 export class ProjectRights {
-  private readonly rights = new Map<string, Set<number>>();
+  private readonly rights = new Map<string, string[]>();
 
   constructor(rights: { [role: string]: ProjectRight[] } = {}) {
     for (const role in rights) {
@@ -27,21 +28,15 @@ export class ProjectRights {
     }
   }
 
-  hasRight(role: string, right: ProjectRight, userId?: string, data?: OwnedData): boolean {
-    const rights = this.rights.get(role);
-    if (rights == null) {
-      return false;
-    }
+  hasRight(project: Project, userId: string, projectDomain: string, operation: Operation, data?: OwnedData): boolean {
+    const rights = (this.rights.get(project.userRoles[userId]) || []).concat(project.userPermissions[userId] || []);
 
-    if (rights.has(right.projectDomain + right.operation)) {
-      if (right.operation === Operation.Create && userId != null && data != null && userId !== data.ownerRef) {
-        return false;
-      }
-      return true;
+    if (rights.includes(this.joinRight(projectDomain, operation))) {
+      return operation === Operation.Create && userId != null && data != null ? userId === data.ownerRef : true;
     }
 
     let ownOperation: Operation;
-    switch (right.operation) {
+    switch (operation) {
       case Operation.Edit:
         ownOperation = Operation.EditOwn;
         break;
@@ -55,10 +50,18 @@ export class ProjectRights {
         return false;
     }
 
-    return userId != null && data != null && rights.has(right.projectDomain + ownOperation) && data.ownerRef === userId;
+    return userId != null && data?.ownerRef === userId && rights.includes(this.joinRight(projectDomain, ownOperation));
+  }
+
+  roleHasRight(role: string, projectDomain: string, operation: Operation): boolean {
+    return (this.rights.get(role) || []).includes(this.joinRight(projectDomain, operation));
+  }
+
+  joinRight(domain: string, operation: string) {
+    return domain + '.' + operation;
   }
 
   protected addRights(role: string, rights: ProjectRight[]): void {
-    this.rights.set(role, new Set<number>(rights.map(r => r.projectDomain + r.operation)));
+    this.rights.set(role, Array.from(new Set<string>(rights.map(r => this.joinRight(r.projectDomain, r.operation)))));
   }
 }
--- a/src/RealtimeServer/common/models/project.ts
+++ b/src/RealtimeServer/common/models/project.ts
@@ -1,6 +1,7 @@
 export interface Project {
   name: string;
   userRoles: { [userRef: string]: string };
+  userPermissions: { [userRef: string]: string[] };
   /** Whether the project has its capability to synchronize project data turned off. */
   syncDisabled?: boolean;
 }
--- a/src/RealtimeServer/common/realtime-server.spec.ts
+++ b/src/RealtimeServer/common/realtime-server.spec.ts
@@ -83,7 +83,7 @@ describe('RealtimeServer', () => {
     expect(ops.length).toEqual(3);
   });
 
-  it('gets correct project role', async () => {
+  it('gets correct project', async () => {
     const env = new TestEnvironment();
     await env.createData();
     let session: ConnectSession;
@@ -95,10 +95,11 @@ describe('RealtimeServer', () => {
     const userConn = clientConnect(env.server, 'user01');
     await submitOp(userConn, PROJECTS_COLLECTION, 'project01', []);
     expect(session!.userId).toEqual('user01');
-    expect(env.server.getUserProjectRole(session!, 'project01')).resolves.toEqual('admin');
+    const project = await env.server.getProject('project01');
+    expect(project?.name).toEqual('Project 01');
   });
 
-  it('gets correct project role when new project added', async () => {
+  it('gets correct project when new project added', async () => {
     const env = new TestEnvironment();
     await env.createData();
     let session: ConnectSession;
@@ -112,29 +113,13 @@ describe('RealtimeServer', () => {
       name: 'Project 02',
       userRoles: {
         user01: 'user'
-      }
+      },
+      userPermissions: {}
     });
     await submitOp(userConn, PROJECTS_COLLECTION, 'project02', []);
     expect(session!.userId).toEqual('user01');
-    expect(env.server.getUserProjectRole(session!, 'project02')).resolves.toEqual('user');
-  });
-
-  it('gets correct project role when role changed', async () => {
-    const env = new TestEnvironment();
-    await env.createData();
-    let session: ConnectSession;
-    env.server.use('submit', (context, callback) => {
-      session = context.agent.connectSession as ConnectSession;
-      callback();
-    });
-
-    const userConn = clientConnect(env.server, 'user01');
-    await env.submitJson0Op<Project>(PROJECTS_COLLECTION, 'project01', ops =>
-      ops.set<string>(p => p.userRoles['user01'], 'user')
-    );
-    await submitOp(userConn, PROJECTS_COLLECTION, 'project01', []);
-    expect(session!.userId).toEqual('user01');
-    expect(env.server.getUserProjectRole(session!, 'project01')).resolves.toEqual('user');
+    const project = await env.server.getProject('project02');
+    expect(project?.name).toEqual('Project 02');
   });
 });
 
@@ -182,7 +167,8 @@ class TestEnvironment {
       name: 'Project 01',
       userRoles: {
         user01: 'admin'
-      }
+      },
+      userPermissions: {}
     });
   }
 
--- a/src/RealtimeServer/common/realtime-server.ts
+++ b/src/RealtimeServer/common/realtime-server.ts
@@ -201,15 +201,11 @@ export class RealtimeServer extends ShareDB {
     return agent;
   }
 
-  async getUserProjectRole(session: ConnectSession, projectId: string): Promise<string | undefined> {
+  async getProject(projectId: string): Promise<Project | undefined> {
     const conn = this.connect();
     const projectDoc = conn.get(this.projectsCollection, projectId);
     await docFetch(projectDoc);
-    if (projectDoc.data != null) {
-      const project: Project = projectDoc.data;
-      return project.userRoles[session.userId];
-    }
-    return undefined;
+    return projectDoc.data as Project | undefined;
   }
 
   async migrateIfNecessary(): Promise<void> {
--- a/src/RealtimeServer/common/services/project-data-service.spec.ts
+++ b/src/RealtimeServer/common/services/project-data-service.spec.ts
@@ -248,9 +248,9 @@ describe('ProjectDataService', () => {
 });
 
 enum TestProjectDomain {
-  TestData = 1000,
-  SubData = 2000,
-  SubSubData = 3000
+  TestData = 'test_data',
+  SubData = 'sub_data',
+  SubSubData = 'sub_sub_data'
 }
 
 interface TestSubSubData extends OwnedData {
@@ -441,7 +441,8 @@ class TestEnvironment {
         user: 'user',
         userOwn: 'userOwn',
         observer: 'observer'
-      }
+      },
+      userPermissions: {}
     });
 
     await createDoc<TestData>(conn, TEST_DATA_COLLECTION, 'test01', {
--- a/src/RealtimeServer/common/services/project-data-service.ts
+++ b/src/RealtimeServer/common/services/project-data-service.ts
@@ -2,6 +2,7 @@ import ShareDB from 'sharedb';
 import { ConnectSession } from '../connect-session';
 import { MigrationConstructor } from '../migration';
 import { OwnedData } from '../models/owned-data';
+import { Project } from '../models/project';
 import { ProjectData } from '../models/project-data';
 import { Operation, ProjectRights } from '../models/project-rights';
 import { RealtimeServer } from '../realtime-server';
@@ -13,7 +14,7 @@ import { JsonDocService } from './json-doc-service';
  * entity type stored in a JSON0 doc.
  */
 export interface ProjectDomainConfig {
-  projectDomain: number;
+  projectDomain: string;
   pathTemplate: ObjPathTemplate;
 }
 
@@ -69,16 +70,9 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
     if (this.server == null) {
       throw new Error('The doc service has not been initialized.');
     }
-    const role = await this.server.getUserProjectRole(session, doc.projectRef);
-    if (role == null) {
-      return false;
-    }
-
+    const project = await this.server.getProject(doc.projectRef);
     const domain = this.getUpdatedDomain([]);
-    if (domain == null) {
-      return false;
-    }
-    return this.hasRight(role, domain, Operation.Create, session.userId, doc);
+    return project != null && domain != null && this.hasRight(project, domain, Operation.Create, session.userId, doc);
   }
 
   protected async allowDelete(_docId: string, doc: T, session: ConnectSession): Promise<boolean> {
@@ -89,16 +83,9 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
     if (this.server == null) {
       throw new Error('The doc service has not been initialized.');
     }
-    const role = await this.server.getUserProjectRole(session, doc.projectRef);
-    if (role == null) {
-      return false;
-    }
-
+    const project = await this.server.getProject(doc.projectRef);
     const domain = this.getUpdatedDomain([]);
-    if (domain == null) {
-      return false;
-    }
-    return this.hasRight(role, domain, Operation.Delete, session.userId, doc);
+    return project != null && domain != null && this.hasRight(project, domain, Operation.Delete, session.userId, doc);
   }
 
   protected async allowRead(_docId: string, doc: T, session: ConnectSession): Promise<boolean> {
@@ -109,13 +96,13 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
     if (this.server == null) {
       throw new Error('The doc service has not been initialized.');
     }
-    const role = await this.server.getUserProjectRole(session, doc.projectRef);
-    if (role == null) {
+    const project = await this.server.getProject(doc.projectRef);
+    if (project == null) {
       return false;
     }
 
     for (const domain of this.domains) {
-      if (!this.hasRight(role, domain, Operation.View, session.userId, doc)) {
+      if (!this.hasRight(project, domain, Operation.View, session.userId, doc)) {
         return false;
       }
     }
@@ -136,8 +123,8 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
     if (this.server == null) {
       throw new Error('The doc service has not been initialized.');
     }
-    const role = await this.server.getUserProjectRole(session, oldDoc.projectRef);
-    if (role == null) {
+    const project = await this.server.getProject(oldDoc.projectRef);
+    if (project == null) {
       return false;
     }
 
@@ -154,25 +141,25 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
         const oldEntity = this.deepGet(entityPath, oldDoc);
         // if the entity doesn't exist in the old doc, then it must be inserted by a previous op that the user has a
         // right to perform, so we don't need to check this edit right
-        if (oldEntity != null && !this.hasRight(role, domain, Operation.Edit, session.userId, oldEntity)) {
+        if (oldEntity != null && !this.hasRight(project, domain, Operation.Edit, session.userId, oldEntity)) {
           return false;
         }
       } else {
         const listOp = op as ShareDB.ListReplaceOp;
         if (listOp.li != null && listOp.ld != null) {
           // replace
-          if (!this.hasRight(role, domain, Operation.Edit, session.userId, listOp.ld)) {
+          if (!this.hasRight(project, domain, Operation.Edit, session.userId, listOp.ld)) {
             return false;
           }
         } else if (listOp.li != null) {
           // create
-          if (!this.hasRight(role, domain, Operation.Create, session.userId, listOp.li)) {
+          if (!this.hasRight(project, domain, Operation.Create, session.userId, listOp.li)) {
             return false;
           }
           checkImmutableProps = false;
         } else if (listOp.ld != null) {
           // delete
-          if (!this.hasRight(role, domain, Operation.Delete, session.userId, listOp.ld)) {
+          if (!this.hasRight(project, domain, Operation.Delete, session.userId, listOp.ld)) {
             return false;
           }
           checkImmutableProps = false;
@@ -202,10 +189,10 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
    *
    * @param {string} _userId The user id.
    * @param {string} _docId The doc id.
-   * @param {number} _projectDomain The project domain of the inserted entity.
+   * @param {string} _projectDomain The project domain of the inserted entity.
    * @param {OwnedData} _entity The inserted entity.
    */
-  protected onInsert(_userId: string, _docId: string, _projectDomain: number, _entity: OwnedData): Promise<void> {
+  protected onInsert(_userId: string, _docId: string, _projectDomain: string, _entity: OwnedData): Promise<void> {
     return Promise.resolve();
   }
 
@@ -215,10 +202,10 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
    *
    * @param {string} _userId The user id.
    * @param {string} _docId The doc id.
-   * @param {number} _projectDomain The project domain of the updated entity.
+   * @param {string} _projectDomain The project domain of the updated entity.
    * @param {OwnedData} _entity The updated entity.
    */
-  protected onUpdate(_userId: string, _docId: string, _projectDomain: number, _entity: OwnedData): Promise<void> {
+  protected onUpdate(_userId: string, _docId: string, _projectDomain: string, _entity: OwnedData): Promise<void> {
     return Promise.resolve();
   }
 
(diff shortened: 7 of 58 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1042)* status: 0 of 60 files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 549 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_vXDut7i5jex8rSZz9:-M_vXDut7i5jex8rSZzA:b-yr3ebl) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L549)):*
> ```Smalltalk
>                 if (!projectDoc.IsLoaded)
>                     throw new DataNotFoundException("The project does not exist.");
>                 // TODO Checking whether the permissions contains a particular string is not a very robust way to check
> ```

I was originally panning on doing this in this PR, but determined that it would be more complicated than I thought, and there's no critical reason why it needs to be included in this PR. It can easily be added afterwards and this line can be updated with zero change in functionality.

RaymondLuong3 (changes requested):
Good work! I have a few comments. I also found that the migration was a little clukey, an error dialog popped up after my browser hung for a good 20 seconds. I think I had the refresh the page after that, but it worked then.

> Reviewed 60 of 60 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-:-M_wPFBp98k-DcElnTeQ:bgngx04)* status: all files reviewed, 8 unresolved discussions (waiting on @Nateowami)

___
*[src/RealtimeServer/common/models/project-rights.ts, line 11 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_vyf8N3bJMy3RUQmF2:-M_vyf8N3bJMy3RUQmF3:bzd25fk) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/RealtimeServer/common/models/project-rights.ts#L11)):*

> ```
> 
>   Create = 'create',
>   Edit = 'edit',
>   Delete = 'delete',
>   View = 'view',
>   EditOwn = 'edit_own',
>   DeleteOwn = 'delete_own',
>   ViewOwn = 'view_own'
> }
> ```
</details>

Is this just so that it is clearer, or is there an actual case where this appears useful to the user (say it can be used and localized for the user?
___
*[src/RealtimeServer/common/models/project-rights.ts, line 58 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_wBbsW8RbN8TxFnjR6:-M_wBbsX6x1VKfyktSZj:b-v5qbok) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/RealtimeServer/common/models/project-rights.ts#L58)):*
> ```
>  joinRight(domain: string, operation: string) {
> ```
A better name for this is "formatRight".
___
*[src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts, line 110 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_w-nj33CdtFFDEQ0Dh:-M_w-nj33CdtFFDEQ0Di:b-ihi1nd) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts#L110)):*
> ```
>  await env.server.migrateIfNecessary();
> ```
This could check userPermissions to be undefined before this line.
___
*[src/RealtimeServer/scriptureforge/services/text-service.ts, line 47 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_w2n4N6MiYb0jZxzc7:-M_w2n4N6MiYb0jZxzc8:b-35vwmz) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/RealtimeServer/scriptureforge/services/text-service.ts#L47)):*
> ```
> const role = project.userRoles[userId] || ProjectRole.None;
> ```
Could we just return early if the userRole is not listed in the project?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.ts, line 98 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_w9IM9C_P-IsBJPCle:-M_w9IM9C_P-IsBJPClf:b-ppf2zx) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.ts#L98)):*
> ```
> return (
> ```
Nit: The open and closing parentheses are not needed.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/rights.service.ts, line 7 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_wEsKxCgmYF00CkoeJ:-M_wEsKxCgmYF00CkoeK:b-p4fzzc) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/SIL.XForge.Scripture/ClientApp/src/app/core/rights.service.ts#L7)):*
> ```
> export class RightsService {
> ```
Could this service be moved into xforge-common so that each xForge app does not need to implement its own rights service?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts, line 219 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_wHa7yCqKslPclcOGv:-M_wHa7yCqKslPclcOGw:bokyduh) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts#L219)):*
> ```
>       SF_PROJECT_RIGHTS.joinRight(SFProjectDomain.Questions, Operation.Create),
>       SF_PROJECT_RIGHTS.joinRight(SFProjectDomain.Questions, Operation.Edit)
> ```
Since rights service is being used to expose rights, maybe rights service should also handle this join rights operation.
___
*[test/SIL.XForge.Tests/Realtime/RealtimeServiceTests.cs, line 59 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_wKLDM7k65Y-St5WF6:-M_wKLDM7k65Y-St5WF7:b-wqtme7) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/test/SIL.XForge.Tests/Realtime/RealtimeServiceTests.cs#L59)):*
> ```
> await collectionToPrune.Received(1).DeleteManyAsync
> ```
I wonder if this needs to be awaited. Did you see a reason?

Nateowami (commented):
Hmm. I have not seen the migration issue. Is it reproducible? If so we need to do something about it. I wouldn't be surprised if it's an issue with how we do migrations rather than specific to this migration.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-:-M_wTNBoBX1RzvmlfIN0:b4kq8sg)* status: 45 of 60 files reviewed, 7 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/RealtimeServer/common/models/project-rights.ts, line 11 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_vyf8N3bJMy3RUQmF2:-M_wPxzfBY8GG34vWbqG:b-285vdk) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/RealtimeServer/common/models/project-rights.ts#L11)):*

> ```
> 
>   Create = 'create',
>   Edit = 'edit',
>   Delete = 'delete',
>   View = 'view',
>   EditOwn = 'edit_own',
>   DeleteOwn = 'delete_own',
>   ViewOwn = 'view_own'
> }
> ```
Is this just so that it is clearer, or is there an actual case where this appears useful to the user (say it can be used and localized for the user?
</blockquote></details>

It's just a little easier to deal with, and now that permissions are persisted in the DB, it will be more readable. So instead of `userPermissions: [ 1201, 1202 ]` it will be stored as `userPermissions: [ "question.create", "question.edit" ]`.
___
*[src/RealtimeServer/common/models/project-rights.ts, line 58 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_wBbsW8RbN8TxFnjR6:-M_xTbZJ-C4LxOzmS8Qy:b-8ijwwn) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/RealtimeServer/common/models/project-rights.ts#L58)):*

> ```
>  joinRight(domain: string, operation: string) {
> ```
A better name for this is "formatRight".
</blockquote></details>

Hmm. Right now it's stored as a string but theoretically we could change it at any time to be an object (`{ domain: blah, operation: blah }`), or a number, or a tuple if they existed in TypeScript, etc. Previously it was stored as a number, and different places throughout our code we would add them together. The problem with that is that now that it's a string they can't just be added. So this helper method was created to produce a value that could be stored and compared. In fact I almost went with objects, but equivalent objects don't `===` each other, so a string with a separator was the simplest way to do this. Before domains and operations were joined via addition (`SFProjectDomain.Questions + Operation.Create` = `1200 + 1` = `1201`). So being a string is more of an implementation detail.
___
*[src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts, line 110 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_w-nj33CdtFFDEQ0Dh:-M_wRqoP0r3GrwiszHiX:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts#L110)):*

> ```
>  await env.server.migrateIfNecessary();
> ```
This could check userPermissions to be undefined before this line.
</blockquote></details>

Done.
___
*[src/RealtimeServer/scriptureforge/services/text-service.ts, line 47 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_w2n4N6MiYb0jZxzc7:-M_xRSLU1PTxRQbU4vut:b9c7x3b) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/RealtimeServer/scriptureforge/services/text-service.ts#L47)):*

> ```
> const role = project.userRoles[userId] || ProjectRole.None;
> ```
Could we just return early if the userRole is not listed in the project?
</blockquote></details>

I've rewritten this in a way that makes this irrelevant.
___
*[src/SIL.XForge/Services/ProjectService.cs, line 256 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_xVEjt2lZGj9bzzeUa:-M_xVEjt2lZGj9bzzeUb:betd0uq) ([raw file](https://github.com/sillsdev/web-xforge/blob/20032fac501fffb165c55b0c75c637c0b2352c3a/src/SIL.XForge/Services/ProjectService.cs#L256)):*
> ```Smalltalk
>             if (projectDoc.IsLoaded)
>                 await projectDoc.SubmitJson0OpAsync(op => op.Unset(p => p.UserRoles[userDoc.Id]));
>             // TODO XXX remove user permissions if they exist
> ```

I think this will be fairly simple and will also call for updating the tests. I forgot about it until one of your comments on another file reminded me of the need to make sure we were removing permissions when users are removed from the project. There may be other places as well. But there's no need to hold up review of everything else while this is being worked on.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.ts, line 98 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_w9IM9C_P-IsBJPCle:-M_wSSLQ2Qc35Zyk2spC:b-5jbl10) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.ts#L98)):*

> ```
> return (
> ```
Nit: The open and closing parentheses are not needed.
</blockquote></details>

There's no way to satisfy our Prettier formatting rules regarding line length and line wrapping without using parentheses. And due to automatic semicolon insertion something like
``` typescript
return
  blah &&
  blah
```
Is equivalent to:
``` typescript
return;
// the next two lines are dead code that are incorrectly indented
  blah &&
  blah
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/rights.service.ts, line 7 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_wEsKxCgmYF00CkoeJ:-M_wZdo03AeM8sInFFZ6:bh0ggp9) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/SIL.XForge.Scripture/ClientApp/src/app/core/rights.service.ts#L7)):*

> ```
> export class RightsService {
> ```
Could this service be moved into xforge-common so that each xForge app does not need to implement its own rights service?
</blockquote></details>

~Looking at this, it's clear it's specific to `SF_PROJECT_RIGHTS`. Perhaps the service should be generic and `SF_PROJECT_RIGHTS` should be injected. Or be provided by the project itself.~

Never mind all that (though I suppose it could be relevant in the future). This was essentially a helper method, and then I discovered everywhere the underlying method was used had been swapped out to use this helper service or another helper method, so it made more sense to update the project rights service itself.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts, line 219 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_wHa7yCqKslPclcOGv:-M_wktqcBsOLg5Lv4y6q:b-ue1cpl) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts#L219)):*

> ```
>       SF_PROJECT_RIGHTS.joinRight(SFProjectDomain.Questions, Operation.Create),
>       SF_PROJECT_RIGHTS.joinRight(SFProjectDomain.Questions, Operation.Edit)
> ```
Since rights service is being used to expose rights, maybe rights service should also handle this join rights operation.
</blockquote></details>

~The rights service is only on the front end as a helper method.~

Actually now it's deleted.
___
*[test/SIL.XForge.Tests/Realtime/RealtimeServiceTests.cs, line 59 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_wKLDM7k65Y-St5WF6:-M_x0-nh71fcdwerrbYi:bcz5367) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/test/SIL.XForge.Tests/Realtime/RealtimeServiceTests.cs#L59)):*

> ```
> await collectionToPrune.Received(1).DeleteManyAsync
> ```
I wonder if this needs to be awaited. Did you see a reason?
</blockquote></details>

I added this to fix warnings about it not being awaited. I'm guessing if we don't await something within a test, there's no guarantee it's finished before another test starts, and that errors can be properly attributed to tests. We had some problems like that in the realtime server tests.

irahopkinson (changes requested):
Good change. This is a lot bigger than I was initially imagining. Thanks for working it through.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-:-M_y2W50-jxU92laIssS:b-7pu5gs)* status: 45 of 60 files reviewed, 11 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[src/RealtimeServer/common/realtime-server.ts, line 208 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_xxaohAob1WHxvMdDG:-M_xxaohAob1WHxvMdDH:buzvw9m) ([raw file](https://github.com/sillsdev/web-xforge/blob/20032fac501fffb165c55b0c75c637c0b2352c3a/src/RealtimeServer/common/realtime-server.ts#L208)):*
> ```
> null
> ```
My understanding is that generally `null` was a bad invention and so should use `undefined` where we can. Is there some reason we want to use `null` instead?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts, line 218 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_y-pEBE168pKWU9UHM:-M_y-pEBE168pKWU9UHN:b-i5gcyl) ([raw file](https://github.com/sillsdev/web-xforge/blob/20032fac501fffb165c55b0c75c637c0b2352c3a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L218)):*
> ```
> canEditQuestion
> ```
Doesn't this still need to be a getter due to the way its called in the HTML?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.html, line 2 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_y-CwA8noZ2y0jLWSP:-M_y-CwA8noZ2y0jLWSQ:b-ds5347) ([raw file](https://github.com/sillsdev/web-xforge/blob/20032fac501fffb165c55b0c75c637c0b2352c3a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.html#L2)):*
> ```
> hasQuestionCreateRight
> ```
Since you are introducing these, I'd be happier with moving to the convention `can<operation><domain>`, i.e. `canCreateQuestion`. Here and below.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/sf-project-settings.ts, line 15 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_y0rYX41qGY1LlJ0Z5:-M_y0rYY31Rb9z4AQkYU:b-34boji) ([raw file](https://github.com/sillsdev/web-xforge/blob/20032fac501fffb165c55b0c75c637c0b2352c3a/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/sf-project-settings.ts#L15)):*
> ```
> number
> ```
Should this be `string`?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1042)* status: 43 of 60 files reviewed, 11 unresolved discussions (waiting on @irahopkinson and @RaymondLuong3)

___
*[src/RealtimeServer/common/realtime-server.ts, line 208 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_xxaohAob1WHxvMdDG:-Ma-FIAl3xSFOJmEJ6Ge:b41hq6i) ([raw file](https://github.com/sillsdev/web-xforge/blob/20032fac501fffb165c55b0c75c637c0b2352c3a/src/RealtimeServer/common/realtime-server.ts#L208)):*
If I'm not mistaken, the criticism of `null` as an invention would apply just as much to `undefined` (perhaps more so actually). I think the context of those criticisms is in a language like C where you can declare a pointer to a string object but then set the pointer value to null (which is actually just 0), and then it becomes very easy to mistakenly de-reference the pointer without realizing that it's necessary to do a null check. It's a method of making a value optional without making clear via the type system that that value is optional and checking for `null` is necessary. At least, that's what I think the criticism is.

From MDN, regarding `undefined`:
> A variable that has not been assigned a value is of type undefined. A method or statement also returns undefined if the variable that is being evaluated does not have an assigned value. A function returns undefined if a value was not returned.

From MDN, regarding `null`:
> The value null represents the intentional absence of any object value.

I view `undefined` as "the absence of any declared value" and `null` as "this value is intentionally marked blank." In practice that means I would use `null` whenever possible, and leave it to the JavaScript engine to make something `undefined`.

How we use them is mostly just a matter of convention, and it makes sense to pick a convention that is consistent with how the JavaScript engine uses them.

With strict `null` checking and algebraic types, I think the fundamental issues with `null` values have been solved, as we can now declare that a value is either required (no `null` checking needed) or optional, and it's enforced by the type system. It's better to avoid optional values when possible, but sometimes they're fundamental to the problem at hand. In this particular case the alternative to returning `null` or `undefined` would be to throw an exception, and looking at the users of this method, I think it would be less helpful.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts, line 218 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_y-pEBE168pKWU9UHM:-Ma-Kp6BEj7DQiV8RCqw:boj1qgt) ([raw file](https://github.com/sillsdev/web-xforge/blob/20032fac501fffb165c55b0c75c637c0b2352c3a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L218)):*

> ```
> canEditQuestion
> ```
Doesn't this still need to be a getter due to the way its called in the HTML?
</blockquote></details>

Good catch. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.html, line 2 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_y-CwA8noZ2y0jLWSP:-Ma-LuMe2Ub_QyDu480_:byah2ip) ([raw file](https://github.com/sillsdev/web-xforge/blob/20032fac501fffb165c55b0c75c637c0b2352c3a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.html#L2)):*

> ```
> hasQuestionCreateRight
> ```
Since you are introducing these, I'd be happier with moving to the convention `can<operation><domain>`, i.e. `canCreateQuestion`. Here and below.
</blockquote></details>

I concur. Done.

I opted not to do this in the editor component though, since it has `hasEditRight` and `canEdit`, which returns `this.isValid && this.hasEditRight && this.dataInSync`. If I renamed `hasEditRight` to `canEditText` we would have `canEdit` and `canEditText` which would be confusing.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/sf-project-settings.ts, line 15 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_y0rYX41qGY1LlJ0Z5:-Ma-cddcA0wnAHGqPrvh:b-we2zz0) ([raw file](https://github.com/sillsdev/web-xforge/blob/20032fac501fffb165c55b0c75c637c0b2352c3a/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/sf-project-settings.ts#L15)):*

> ```
> number
> ```
Should this be `string`?
</blockquote></details>

Actually this should't be here at all. It's left over from when I was trying to add the permissions to the `SFProject` model rather than the `Project` model. Afterwards I realized permissions were a more general concept, so I moved them to the `Project` model, which is where the `userRoles` property is. I thought I'd reverted this, but apparently I missed this part.

RaymondLuong3 (approved):
Ira can give the final approval, but looks good from my end.

> Reviewed 13 of 15 files at r2, 4 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-:-Ma0OhxO8gvtZzsoz6FU:bckhyg7)* status: all files reviewed, 6 unresolved discussions (waiting on @irahopkinson and @RaymondLuong3)

___
*[src/RealtimeServer/common/models/project-rights.ts, line 11 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_vyf8N3bJMy3RUQmF2:-Ma0JmkJCPaLLR1Qorzb:bg1iqnt) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/RealtimeServer/common/models/project-rights.ts#L11)):*

It's just a little easier to deal with, and now that permissions are persisted in the DB, it will be more readable. So instead of `userPermissions: [ 1201, 1202 ]` it will be stored as `userPermissions: [ "question.create", "question.edit" ]`.
</blockquote></details>

Ok, that is true. Now we are committed to it since it is recorded in the database now!
___
*[src/RealtimeServer/common/models/project-rights.ts, line 58 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_wBbsW8RbN8TxFnjR6:-Ma0KszpBxR1JM9QNEpM:b-e4sfn) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/RealtimeServer/common/models/project-rights.ts#L58)):*

Hmm. Right now it's stored as a string but theoretically we could change it at any time to be an object (`{ domain: blah, operation: blah }`), or a number, or a tuple if they existed in TypeScript, etc. Previously it was stored as a number, and different places throughout our code we would add them together. The problem with that is that now that it's a string they can't just be added. So this helper method was created to produce a value that could be stored and compared. In fact I almost went with objects, but equivalent objects don't `===` each other, so a string with a separator was the simplest way to do this. Before domains and operations were joined via addition (`SFProjectDomain.Questions + Operation.Create` = `1200 + 1` = `1201`). So being a string is more of an implementation detail.
</blockquote></details>

I wonder if joinRight is a good name for that, it almost sounds like it is some sort of database operation (e.g. join right, join left). But I'll leave that for you to decide.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/rights.service.ts, line 7 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_wEsKxCgmYF00CkoeJ:-Ma0NombE6LjB-xLbGcW:bptq0wa) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/SIL.XForge.Scripture/ClientApp/src/app/core/rights.service.ts#L7)):*

~Looking at this, it's clear it's specific to `SF_PROJECT_RIGHTS`. Perhaps the service should be generic and `SF_PROJECT_RIGHTS` should be injected. Or be provided by the project itself.~

Never mind all that (though I suppose it could be relevant in the future). This was essentially a helper method, and then I discovered everywhere the underlying method was used had been swapped out to use this helper service or another helper method, so it made more sense to update the project rights service itself.
</blockquote></details>

I liked the idea of a rights service, however I see that it is not quite the same as our other services that aim to achieve statelessness.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1042)* status: 58 of 60 files reviewed, 5 unresolved discussions (waiting on @irahopkinson and @RaymondLuong3)

___
*[src/RealtimeServer/common/models/project-rights.ts, line 58 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_wBbsW8RbN8TxFnjR6:-Ma0cpp639Zuj32qjW1i:b-a8hldf) ([raw file](https://github.com/sillsdev/web-xforge/blob/54c0ecfebf80d771bd876a179badeddc9f648439/src/RealtimeServer/common/models/project-rights.ts#L58)):*

I wonder if joinRight is a good name for that, it almost sounds like it is some sort of database operation (e.g. join right, join left). But I'll leave that for you to decide.
</blockquote></details>

Hmm. Good point. What do you think about "buildRight" or something of that sort?

irahopkinson (approved):
:lgtm: 

> Reviewed 5 of 60 files at r1, 1 of 15 files at r2, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-:-Ma4Ws3F4NMYT5bQVWYs:bg7igkv)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[src/RealtimeServer/common/realtime-server.ts, line 208 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_xxaohAob1WHxvMdDG:-Ma4TVIXAikk4pEw0DCA:b-heqhs6) ([raw file](https://github.com/sillsdev/web-xforge/blob/20032fac501fffb165c55b0c75c637c0b2352c3a/src/RealtimeServer/common/realtime-server.ts#L208)):*

If I'm not mistaken, the criticism of `null` as an invention would apply just as much to `undefined` (perhaps more so actually). I think the context of those criticisms is in a language like C where you can declare a pointer to a string object but then set the pointer value to null (which is actually just 0), and then it becomes very easy to mistakenly de-reference the pointer without realizing that it's necessary to do a null check. It's a method of making a value optional without making clear via the type system that that value is optional and checking for `null` is necessary. At least, that's what I think the criticism is.

From MDN, regarding `undefined`:
> A variable that has not been assigned a value is of type undefined. A method or statement also returns undefined if the variable that is being evaluated does not have an assigned value. A function returns undefined if a value was not returned.

From MDN, regarding `null`:
> The value null represents the intentional absence of any object value.

I view `undefined` as "the absence of any declared value" and `null` as "this value is intentionally marked blank." In practice that means I would use `null` whenever possible, and leave it to the JavaScript engine to make something `undefined`.

How we use them is mostly just a matter of convention, and it makes sense to pick a convention that is consistent with how the JavaScript engine uses them.

With strict `null` checking and algebraic types, I think the fundamental issues with `null` values have been solved, as we can now declare that a value is either required (no `null` checking needed) or optional, and it's enforced by the type system. It's better to avoid optional values when possible, but sometimes they're fundamental to the problem at hand. In this particular case the alternative to returning `null` or `undefined` would be to throw an exception, and looking at the users of this method, I think it would be less helpful.
</blockquote></details>

This doesn't need to hold up this PR since we already have a mixture, I do see your points but:
TypeScript team doesn't use `null` : [TypeScript coding guidelines](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined) and it hasn't caused any problems. Douglas Crockford thinks `null` [is a bad idea](https://www.youtube.com/watch?v=PSGEjv3Tqo0&feature=youtu.be&t=9m21s) and we should all just use `undefined`.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1042)* status: 58 of 60 files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @RaymondLuong3)

___
*[src/RealtimeServer/common/realtime-server.ts, line 208 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1042#-M_xxaohAob1WHxvMdDG:-Ma4ZmSTD6sH0fPGbXs4:b-n6lqqz) ([raw file](https://github.com/sillsdev/web-xforge/blob/20032fac501fffb165c55b0c75c637c0b2352c3a/src/RealtimeServer/common/realtime-server.ts#L208)):*

This doesn't need to hold up this PR since we already have a mixture, I do see your points but:
TypeScript team doesn't use `null` : [TypeScript coding guidelines](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#null-and-undefined) and it hasn't caused any problems. Douglas Crockford thinks `null` [is a bad idea](https://www.youtube.com/watch?v=PSGEjv3Tqo0&feature=youtu.be&t=9m21s) and we should all just use `undefined`.
</blockquote></details>

Ah, I see. I thought you were referring to [Tony Hoare's statement](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare/) that his invention of the null reference in 1965 was a billion-dollar mistake (which I've seen repeated elsewhere).

Let's stick to `undefined` then.

It looks like the type is `any`, so I've added a type assertion to convert it to `Project | undefined`.

PR #1050, opened 2021-05-25
Temporarily fix userPermissions undefined error

Nateowami:
The data migration adds the property, but there are ways to end up with updated code and non-migrated data which causes errors. This is a temporary fix so we can ship. Long term the way we handle migrations will need to be improved.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1050)


irahopkinson (approved):
:lgtm: thanks for the quick turn around on this fix.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1050#-:-MaWe7A4AM9V0e6ZNO99:b-1nistx)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1051, opened 2021-05-26
SF-1048: Added Cancel ability when synchronizing

pmachapman:
This adds support for cancelling a sync that is queued or in progress.

Cancellation is handled via a `CancellationToken`, with this passed to logic that will accept it with the exception of ParatextData, which uses a static system-wide cancellation method (calling this would be catastrophic for any calls being made to ParatextData in other threads concurrently). The `ParatextSyncRunner.RunAsync()` method polls this value at specific places so that cancellation can be handled without causing corruption. If the job was cancelled after the data was downloaded from Paratext, but before it could fully be loaded into the database, the logic utilizing the `Sync.DataInSync` property from SF-1320a will ensure that the user cannot make modifications to the data until they complete a successful resync.

A `jobIds` array is added to the Project Sync object, storing the job ids from Hangfire for any jobs that have been created.

Cancelling a sync for a project cancels all jobs associated with that project - the reason for this is that it is not exactly possible to know which sync task the user actually wants to cancel (if there are multiple jobs associate with a project), and multiple jobs queued for a project are either accidental or caused by concurrent users. Either way, the user(s) will be notified that the job has failed, and can reschedule a sync if they require.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -65,6 +65,7 @@
     "transcelerator",
     "transloco",
     "uncheck",
+    "unelevated",
     "uninvite",
     "unsynchronized",
     "usfm",
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts
@@ -116,6 +116,10 @@ export class SFProjectService extends ProjectService<SFProject, SFProjectDoc> {
     return this.onlineInvoke('sync', { projectId: id });
   }
 
+  onlineCancelSync(id: string): Promise<void> {
+    return this.onlineInvoke('cancelSync', { projectId: id });
+  }
+
   onlineUpdateSettings(id: string, settings: SFProjectSettings): Promise<void> {
     return this.onlineInvoke('updateSettings', { projectId: id, settings });
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html
@@ -40,6 +40,17 @@
             [projectDoc]="projectDoc"
             (inProgress)="syncActive = $event"
           ></app-sync-progress>
+          <button
+            *ngIf="syncActive"
+            class="action-button"
+            mdc-button
+            type="button"
+            (click)="cancelSync()"
+            [disabled]="isLoading || !isAppOnline || syncDisabled"
+            id="btn-cancel-sync"
+          >
+            {{ t("cancel") }}
+          </button>
           <span *ngIf="!syncActive" [title]="lastSyncDate" id="date-last-sync" class="sync-info">
             {{ lastSyncNotice }}
           </span>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss
@@ -20,7 +20,8 @@ h2 {
   align-self: center;
 }
 
-#btn-sync {
+#btn-sync,
+.progress-bar {
   margin-bottom: 1rem;
 }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.spec.ts
@@ -106,7 +106,7 @@ describe('SyncComponent', () => {
     expect(env.component.syncActive).toBe(true);
     expect(env.progressBar).not.toBeNull();
     // Simulate sync in progress
-    env.emitSyncProgress(0, 'testProject01');
+    env.emitSyncProgress(0.25, 'testProject01');
     // Simulate sync error
     env.emitSyncComplete(false, 'testProject01');
     expect(env.component.syncActive).toBe(false);
@@ -131,6 +131,21 @@ describe('SyncComponent', () => {
     const env = new TestEnvironment(true, false, true, false);
     expect(env.syncButton.nativeElement.disabled).toBe(false);
     expect(env.syncDisabledMessage).toBeNull();
+  }));
+
+  it('should report cancelled if sync was cancelled', fakeAsync(() => {
+    const env = new TestEnvironment(true);
+    verify(mockedProjectService.get('testProject01')).once();
+    env.clickElement(env.syncButton);
+    verify(mockedProjectService.onlineSync('testProject01')).once();
+    expect(env.component.syncActive).toBe(true);
+    expect(env.progressBar).not.toBeNull();
+
+    env.clickElement(env.cancelButton);
+    env.emitSyncComplete(false, 'testProject01');
+
+    expect(env.component.syncActive).toBe(false);
+    verify(mockedNoticeService.show('Synchronize for Sync Test Project was cancelled.')).once();
   }));
 });
 
@@ -208,6 +223,10 @@ class TestEnvironment {
 
   get syncButton(): DebugElement {
     return this.fixture.debugElement.query(By.css('#btn-sync'));
+  }
+
+  get cancelButton(): DebugElement {
+    return this.fixture.debugElement.query(By.css('#btn-cancel-sync'));
   }
 
   get progressBar(): DebugElement {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts
@@ -22,8 +22,10 @@ export class SyncComponent extends DataLoadingComponent implements OnInit {
   syncDisabled: boolean = false;
   projectDoc?: SFProjectDoc;
 
-  private paratextUsername?: string;
   private _syncActive: boolean = false;
+  private isSyncCancelled = false;
+  private paratextUsername?: string;
+  private previousLastSyncDate?: Date;
 
   constructor(
     private readonly route: ActivatedRoute,
@@ -53,7 +55,7 @@ export class SyncComponent extends DataLoadingComponent implements OnInit {
     }
   }
 
-  get lastSyncDate() {
+  get lastSyncDate(): string {
     if (
       this.projectDoc == null ||
       this.projectDoc.data == null ||
@@ -66,7 +68,7 @@ export class SyncComponent extends DataLoadingComponent implements OnInit {
   }
 
   get projectName(): string {
-    return this.projectDoc == null || this.projectDoc.data == null ? '' : this.projectDoc.data.name;
+    return this.projectDoc?.data == null ? '' : this.projectDoc.data.name;
   }
 
   get syncActive(): boolean {
@@ -75,17 +77,21 @@ export class SyncComponent extends DataLoadingComponent implements OnInit {
 
   set syncActive(isActive: boolean) {
     if (this._syncActive && !isActive && this.projectDoc?.data != null) {
-      if (this.projectDoc.data.sync.lastSyncSuccessful) {
-        this.noticeService.show(
-          translate('sync.successfully_synchronized_with_paratext', { projectName: this.projectDoc.data.name })
-        );
+      const projectName: string = this.projectDoc.data.name;
+      if (this.isSyncCancelled && new Date(this.lastSyncDate) <= this.previousLastSyncDate!) {
+        this.noticeService.show(translate('sync.synchronize_was_cancelled', { projectName }));
       } else {
-        const name: string = this.projectDoc.data.name;
-        this.noticeService.showMessageDialog(() =>
-          translate('sync.something_went_wrong_synchronizing_this_project', { projectName: name })
-        );
+        if (this.projectDoc.data.sync.lastSyncSuccessful) {
+          this.noticeService.show(translate('sync.successfully_synchronized_with_paratext', { projectName }));
+          this.previousLastSyncDate = new Date(this.lastSyncDate);
+        } else {
+          this.noticeService.showMessageDialog(() =>
+            translate('sync.something_went_wrong_synchronizing_this_project', { projectName })
+          );
+        }
       }
     }
+    this.isSyncCancelled = false;
     this._syncActive = isActive;
   }
 
@@ -140,6 +146,14 @@ export class SyncComponent extends DataLoadingComponent implements OnInit {
     this.projectService.onlineSync(this.projectDoc.id);
   }
 
+  cancelSync(): void {
+    if (this.projectDoc == null) {
+      return;
+    }
+    this.projectService.onlineCancelSync(this.projectDoc.id);
+    this.isSyncCancelled = true;
+  }
+
   private checkSyncStatus(): void {
     if (this.projectDoc?.data == null) {
       return;
@@ -148,5 +162,6 @@ export class SyncComponent extends DataLoadingComponent implements OnInit {
       this.syncDisabled = this.projectDoc.data.syncDisabled;
     }
     this._syncActive = this.projectDoc.data.sync.queuedCount > 0;
+    this.previousLastSyncDate = new Date(this.lastSyncDate);
   }
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json
+++ b/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json
@@ -194,6 +194,7 @@
     "translation_suggestions_settings": "Translation suggestions settings"
   },
   "sync": {
+    "cancel": "Cancel",
     "connect_network_to_synchronize": "Please connect to the internet to synchronize this project",
     "last_synced_time_stamp": "Last synced on {{ timeStamp }}",
     "log_in_to_paratext": "Log in to Paratext",
@@ -201,6 +202,7 @@
     "something_went_wrong_synchronizing_this_project": "Something went wrong while synchronizing {{ projectName }} with Paratext. Please try again.",
     "successfully_synchronized_with_paratext": "Successfully synchronized {{ projectName }} with Paratext.",
     "synchronize": "Synchronize",
+    "synchronize_was_cancelled": "Synchronize for {{ projectName }} was cancelled.",
     "synchronize_project": "Synchronize {{ projectName }} with Paratext",
     "your_project_is_being_synchronized": "Your project is being synchronized...",
     "sync_is_disabled": "Synchronization for this project needed to be disabled, so it is not available at this time. Please contact {{ email }} to discuss what needs to be done so project synchronization can be enabled again."
--- a/src/SIL.XForge.Scripture/Controllers/SFProjectsRpcController.cs
+++ b/src/SIL.XForge.Scripture/Controllers/SFProjectsRpcController.cs
@@ -291,6 +291,23 @@ namespace SIL.XForge.Scripture.Controllers
             }
         }
 
+        public async Task<IRpcMethodResult> CancelSync(string projectId)
+        {
+            try
+            {
+                await _projectService.CancelSyncAsync(UserId, projectId);
+                return Ok();
+            }
+            catch (ForbiddenException)
+            {
+                return ForbiddenError();
+            }
+            catch (DataNotFoundException dnfe)
+            {
+                return NotFoundError(dnfe.Message);
+            }
+        }
+
         public async Task<IRpcMethodResult> DeleteAudio(string projectId, string ownerId, string dataId)
         {
             try
--- a/src/SIL.XForge.Scripture/Models/SFProjectSecret.cs
+++ b/src/SIL.XForge.Scripture/Models/SFProjectSecret.cs
@@ -6,6 +6,17 @@ namespace SIL.XForge.Scripture.Models
     public class SFProjectSecret : ProjectSecret
     {
         /// <summary>
+        /// The queued or active Hangfire Job Ids for the project.
+        /// </summary>
+        /// <value>
+        /// The Hangfire Job Ids.
+        /// </value>
+        /// <remarks>
+        /// The <see cref="List{T}.Count">Count</see> should correspond to <see cref="QueuedCount" />.
+        /// </remarks>
+        public List<string> JobIds { get; set; } = new List<string>();
+
+        /// <summary>
         /// Keeps track of all Paratext usernames that have been used to sync notes with Paratext
         /// </summary>
         public List<SyncUser> SyncUsers { get; set; } = new List<SyncUser>();
--- a/src/SIL.XForge.Scripture/Services/IJwtTokenHelper.cs
+++ b/src/SIL.XForge.Scripture/Services/IJwtTokenHelper.cs
@@ -1,4 +1,5 @@
 using System.Net.Http;
+using System.Threading;
 using System.Threading.Tasks;
 using SIL.XForge.Configuration;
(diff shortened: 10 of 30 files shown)


irahopkinson (changes requested):
> Reviewed 1 of 35 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051)* status: 1 of 45 files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[src/RealtimeServer/scriptureforge/models/sync.ts, line 8 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-MacC-C95PKqDW_nk_0w:-MacC-CA1KbgitPs8OYL:bbo49ga) ([raw file](https://github.com/sillsdev/web-xforge/blob/2ffb9be30a222bd49b5c0ada2af0520570828be4/src/RealtimeServer/scriptureforge/models/sync.ts#L8)):*
> ```
> jobIds: string[];
> ```
This is not a full review as I am on holiday today but I noticed this so you might like to get to it sooner rather than later. I don't think we want to put this in here. This information is not needed on the frontend and shouldn't be revealed to it. Additionally this is a OT doc and this data doesn't need OT support. Use an standard Mongo collection rather than an OT doc that is only used on the backend.

pmachapman (commented):
The LGTM alerts seem to be old alerts from other code I haven't touched?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-:-Mb-aAeN8LxRm_Ns3FSl:b-5rjsff)* status: 0 of 50 files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[src/RealtimeServer/scriptureforge/models/sync.ts, line 8 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-MacC-C95PKqDW_nk_0w:-Mb-aL_07pvFYeSXU-Xw:bl1dn5r) ([raw file](https://github.com/sillsdev/web-xforge/blob/2ffb9be30a222bd49b5c0ada2af0520570828be4/src/RealtimeServer/scriptureforge/models/sync.ts#L8)):*

> ```
> jobIds: string[];
> ```
This is not a full review as I am on holiday today but I noticed this so you might like to get to it sooner rather than later. I don't think we want to put this in here. This information is not needed on the frontend and shouldn't be revealed to it. Additionally this is a OT doc and this data doesn't need OT support. Use an standard Mongo collection rather than an OT doc that is only used on the backend.
</blockquote></details>

I have updated it to use `project_secrets`. As I needed to add an extra method to `IUpdateBuilder` (I was unable to use `RemoveAll` to remove individual elements from a string array), I took the opportunity to write unit tests for `MemoryUpdateBuilder`, as there were none and I felt I should write a test for the new method (`Remove`).

ddaspit (changes requested):
@irahopkinson we should probably add an issue to investigate adding proper cancellation token support in the ParatextData DLL.

> Reviewed 10 of 14 files at r1, 20 of 35 files at r3, 20 of 20 files at r4, 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-:-Mb-mQRp3ANgP36Mg6Go:blqx36a)* status: all files reviewed, 5 unresolved discussions (waiting on @irahopkinson and @pmachapman)

___
*[src/SIL.XForge.Scripture/Models/Sync.cs, line 2 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-Mb-cIzD4tWuM7qwx1eA:-Mb-cIzECzHUOTsCLODM:b-64d865) ([raw file](https://github.com/sillsdev/web-xforge/blob/3382d8dffdedc195be904da3ea534dce8154aeb0/src/SIL.XForge.Scripture/Models/Sync.cs#L2)):*
> ```Smalltalk
> using System;
> using System.Collections.Generic;
> ```

I think this can be removed.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 216 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-Mb-dOWmAaQcE1vSfC38:-Mb-dOWmAaQcE1vSfC39:bfsszy1) ([raw file](https://github.com/sillsdev/web-xforge/blob/3382d8dffdedc195be904da3ea534dce8154aeb0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L216)):*
> ```Smalltalk
>         {
>             IInternetSharedRepositorySource ptRepoSource = await GetInternetSharedRepositorySource(userSecret.Id,
>                 CancellationToken.None);
> ```

Is there are reason why you aren't passing in a cancellation token here?
___
*[src/SIL.XForge.Scripture/Services/SyncService.cs, line 73 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-Mb-gLtR0FfL2HvemFjx:-Mb-gLtR0FfL2HvemFjy:bcbw9xj) ([raw file](https://github.com/sillsdev/web-xforge/blob/3382d8dffdedc195be904da3ea534dce8154aeb0/src/SIL.XForge.Scripture/Services/SyncService.cs#L73)):*
> ```Smalltalk
>                         // the job unless the queued count and job ids have been incremented appropriately.
>                         // We need to sync the source first so that we can link the source texts and train the engine.
>                         string sourceJobId = _backgroundJobClient.Schedule<ParatextSyncRunner>(
> ```

I assume that we are doing this, because we need the job id before the job is actually started so that we can add it to the `JobIds` collection. If we add the job id to the `JobIds` collection inside of the sync, would it avoid having to do this hack? I believe it is possible to get the job id inside of a job.
___
*[src/SIL.XForge.Scripture/Services/SyncService.cs, line 201 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-Mb-lmQm2IKXA7UCIh5x:-Mb-lmQm2IKXA7UCIh5y:b-u7bzvg) ([raw file](https://github.com/sillsdev/web-xforge/blob/3382d8dffdedc195be904da3ea534dce8154aeb0/src/SIL.XForge.Scripture/Services/SyncService.cs#L201)):*
> ```Smalltalk
>                                 // Load the source project secrets, so we can get any job ids
>                                 if ((await _projectSecrets.TryGetAsync(sourceProjectId)).TryResult(out SFProjectSecret sourceProjectSecret))
> ```

This code seems very similar to the code for the main project. I would refactor this into a separate method that you can call for the source and target project.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051)* status: all files reviewed, 5 unresolved discussions (waiting on @ddaspit and @irahopkinson)

___
*[src/SIL.XForge.Scripture/Models/Sync.cs, line 2 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-Mb-cIzD4tWuM7qwx1eA:-Mb-qc1V11cJiAcnae1g:b-k69bgi) ([raw file](https://github.com/sillsdev/web-xforge/blob/3382d8dffdedc195be904da3ea534dce8154aeb0/src/SIL.XForge.Scripture/Models/Sync.cs#L2)):*

I think this can be removed.
</blockquote></details>

Yes it can - thanks!
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 216 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-Mb-dOWmAaQcE1vSfC38:-Mb-qgGRD2lklHohjOmc:b-eo2ser) ([raw file](https://github.com/sillsdev/web-xforge/blob/3382d8dffdedc195be904da3ea534dce8154aeb0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L216)):*

Is there are reason why you aren't passing in a cancellation token here?
</blockquote></details>

Only because none of the code that calls that method supports a cancellation token. I can expand the use of cancellation tokens, if you think its beneficial?

The main reason I limited the scope of the implementation is that the rest of the code, particularly the front end isn't really set up to support it, but it is something I can implement much more widely. Will be quite a wide (but potentially useful down the track) change.
___
*[src/SIL.XForge.Scripture/Services/SyncService.cs, line 73 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-Mb-gLtR0FfL2HvemFjx:-Mb-rmm80cvHan3eGCa1:bs7dm66) ([raw file](https://github.com/sillsdev/web-xforge/blob/3382d8dffdedc195be904da3ea534dce8154aeb0/src/SIL.XForge.Scripture/Services/SyncService.cs#L73)):*

I assume that we are doing this, because we need the job id before the job is actually started so that we can add it to the `JobIds` collection. If we add the job id to the `JobIds` collection inside of the sync, would it avoid having to do this hack? I believe it is possible to get the job id inside of a job.
</blockquote></details>

Yes, you can (see https://discuss.hangfire.io/t/how-to-get-jobid-within-job/851/5), but a problem would occur if more than job was queued for a project (i.e. if it is used as a source project for multiple projects, etc), where the second or third jobid would not appear in the array due to the Mutex restricting one sync action at a time for a project. This would mean that cancelling would not stop the second or third queued job from running, as only the first jobid would be in the array.

I considered a way of making sure that only one sync can ever be scheduled at a time for a project, but that failed an edge case where the project was a target (with another source), and another project had it as a source project.

You are right it is a hack - the best supported way of resolving it I could find is to use the batch feature of Hangfire Pro (requires a subscription).
___
*[src/SIL.XForge.Scripture/Services/SyncService.cs, line 201 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-Mb-lmQm2IKXA7UCIh5x:-Mb-rfxm-mSf-lsYqlQY:bpxyi9g) ([raw file](https://github.com/sillsdev/web-xforge/blob/3382d8dffdedc195be904da3ea534dce8154aeb0/src/SIL.XForge.Scripture/Services/SyncService.cs#L201)):*

This code seems very similar to the code for the main project. I would refactor this into a separate method that you can call for the source and target project.
</blockquote></details>

Yes, it is an obvious target for refactor - thanks for spotting it!

ddaspit (changes requested):
> Reviewed 2 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051)* status: all files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @pmachapman)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 216 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-Mb-dOWmAaQcE1vSfC38:-Mb09bGQ6m0zg_zsOsOa:bjl5phc) ([raw file](https://github.com/sillsdev/web-xforge/blob/3382d8dffdedc195be904da3ea534dce8154aeb0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L216)):*

Only because none of the code that calls that method supports a cancellation token. I can expand the use of cancellation tokens, if you think its beneficial?

The main reason I limited the scope of the implementation is that the rest of the code, particularly the front end isn't really set up to support it, but it is something I can implement much more widely. Will be quite a wide (but potentially useful down the track) change.
</blockquote></details>

Ok, I see why you aren't passing it here. The `GetProjectsAsync` method is not used when syncing. We don't need to expand the use of cancellation tokens outside of sync right now, so this is fine.
___
*[src/SIL.XForge.Scripture/Services/SyncService.cs, line 73 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-Mb-gLtR0FfL2HvemFjx:-Mb09Q8N0X_HBavggvj_:b-tvcq44) ([raw file](https://github.com/sillsdev/web-xforge/blob/3382d8dffdedc195be904da3ea534dce8154aeb0/src/SIL.XForge.Scripture/Services/SyncService.cs#L73)):*

Yes, you can (see https://discuss.hangfire.io/t/how-to-get-jobid-within-job/851/5), but a problem would occur if more than job was queued for a project (i.e. if it is used as a source project for multiple projects, etc), where the second or third jobid would not appear in the array due to the Mutex restricting one sync action at a time for a project. This would mean that cancelling would not stop the second or third queued job from running, as only the first jobid would be in the array.

I considered a way of making sure that only one sync can ever be scheduled at a time for a project, but that failed an edge case where the project was a target (with another source), and another project had it as a source project.

You are right it is a hack - the best supported way of resolving it I could find is to use the batch feature of Hangfire Pro (requires a subscription).
</blockquote></details>

That makes sense. I would use `TimeSpan.MaxValue` just to be absolutely sure that the job doesn't start before we have incremented the queue count and added the job id.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051)* status: all files reviewed, 2 unresolved discussions (waiting on @ddaspit and @irahopkinson)

___
*[src/SIL.XForge.Scripture/Services/SyncService.cs, line 73 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-Mb-gLtR0FfL2HvemFjx:-Mb0Hs-W-rFDyXJB9Gys:b-b6zi7q) ([raw file](https://github.com/sillsdev/web-xforge/blob/3382d8dffdedc195be904da3ea534dce8154aeb0/src/SIL.XForge.Scripture/Services/SyncService.cs#L73)):*

That makes sense. I would use `TimeSpan.MaxValue` just to be absolutely sure that the job doesn't start before we have incremented the queue count and added the job id.
</blockquote></details>

I originally had that, but `TimeSpan.MaxValue` causes the request to hangfire to crash due to their code's handling of the value.

I chose 30 seconds, as it will cover any reasonable delays, and any unreasonable delays (like extreme latency to the RealtimeServer or MongoDB) will result in the job running within a comparatively quick timeframe (and most likely failing from the cause of the latency original problem - I am picturing a server shutdown or unexpected outage here). I was afraid if I made it too long, it would start running too far in the future, unconnected to the original request/environment of the user, and might run successfully and send the queuecount into negative, maybe cause other issues or confusion, etc.

I can make it longer if you like? 2 minutes? 5 minutes? An alternative would be to make it super long in the future (maybe a year or two), but then hangfire most definitely would run it then if we forgot to purge the jobGraph regularly.

ddaspit (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051)* status: all files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @pmachapman)

___
*[src/SIL.XForge.Scripture/Services/SyncService.cs, line 73 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-Mb-gLtR0FfL2HvemFjx:-MbKCnLd0Y7GpHxVePKE:bpawke2) ([raw file](https://github.com/sillsdev/web-xforge/blob/3382d8dffdedc195be904da3ea534dce8154aeb0/src/SIL.XForge.Scripture/Services/SyncService.cs#L73)):*

I originally had that, but `TimeSpan.MaxValue` causes the request to hangfire to crash due to their code's handling of the value.

I chose 30 seconds, as it will cover any reasonable delays, and any unreasonable delays (like extreme latency to the RealtimeServer or MongoDB) will result in the job running within a comparatively quick timeframe (and most likely failing from the cause of the latency original problem - I am picturing a server shutdown or unexpected outage here). I was afraid if I made it too long, it would start running too far in the future, unconnected to the original request/environment of the user, and might run successfully and send the queuecount into negative, maybe cause other issues or confusion, etc.

I can make it longer if you like? 2 minutes? 5 minutes? An alternative would be to make it super long in the future (maybe a year or two), but then hangfire most definitely would run it then if we forgot to purge the jobGraph regularly.
</blockquote></details>

Sounds reasonable. I would make it a little longer, 30 seconds feels too short to me. Maybe somewhere in the range of 2-5 minutes.

pmachapman (commented):
There was a lot of warnings after the rebase, so I sorted them out.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-:-MbYVajo3N7fYQesqbfm:b8uq50r)* status: 31 of 50 files reviewed, 2 unresolved discussions (waiting on @ddaspit and @irahopkinson)

___
*[src/SIL.XForge.Scripture/Services/SyncService.cs, line 73 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-Mb-gLtR0FfL2HvemFjx:-MbYViRY0rToHa1OfSXD:bw2w8tv) ([raw file](https://github.com/sillsdev/web-xforge/blob/3382d8dffdedc195be904da3ea534dce8154aeb0/src/SIL.XForge.Scripture/Services/SyncService.cs#L73)):*

Sounds reasonable. I would make it a little longer, 30 seconds feels too short to me. Maybe somewhere in the range of 2-5 minutes.
</blockquote></details>

No problem - it is now 5 minutes.

irahopkinson (changes requested):
> Reviewed 9 of 14 files at r1, 15 of 35 files at r3, 1 of 2 files at r6.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051)* status: 31 of 50 files reviewed, 4 unresolved discussions (waiting on @ddaspit and @pmachapman)

___
*[src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs, line 678 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-MbZqC3w0DBWeynpK_hp:-MbZqC3w0DBWeynpK_hq:b-smhf0s) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4b25b8a79250b63a1dea42739173f7c0fc723fc/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L678)):*
> ```
> else
> ```
We don't need to do an operation if it equals 0. Perhaps change `else` to `else if (_projectDoc.Data.Sync.QueuedCount < 0)`? Although I have a feeling that it doesn't actually result in an actual operation if the value is already 0 so perhaps what you have is fine?
___
*[src/SIL.XForge.Scripture/Services/SyncService.cs, line 48 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-MbZuyOhDzjXhgwqS_21:-MbZuyOhDzjXhgwqS_22:b9o16xi) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4b25b8a79250b63a1dea42739173f7c0fc723fc/src/SIL.XForge.Scripture/Services/SyncService.cs#L48)):*
> ```
> projectSecret = null;
> ```
Isn't this a problem to proceed past this point?
___
*[src/SIL.XForge.Scripture/Services/SyncService.cs, line 160 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-MbZvQPF2-KZFc2QzSOS:-MbZvQPF2-KZFc2QzSOT:b-odrlv3) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4b25b8a79250b63a1dea42739173f7c0fc723fc/src/SIL.XForge.Scripture/Services/SyncService.cs#L160)):*
> ```
> projectDoc.Data.SyncDisabled
> ```
Hmm... if the sync is already started presumably it wasn't disabled when it was started. Shouldn't we allow the user to still cancel if the sync has been disabled since?

irahopkinson (commented):
Another question: when a project gets deleted I'm pretty sure the JobIds will get deleted because I think the projectSecret gets deleted but do any jobs get cancelled first?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-:-Mb_9ZuqCxiH2d969Cst:bfqmjgm)* status: 31 of 50 files reviewed, 4 unresolved discussions (waiting on @ddaspit and @pmachapman)

pmachapman (commented):
I had not thought of that scenario! I've made it so the jobs are cancelled, and updated the appropriate unit test to check for it. Thanks!!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-:-Mb_Bubc8pU0u3jjckor:b-cxuwhh)* status: 31 of 50 files reviewed, 4 unresolved discussions (waiting on @ddaspit and @irahopkinson)

___
*[src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs, line 678 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-MbZqC3w0DBWeynpK_hp:-Mb_75nICDxnCp56ur5X:bxa3qd6) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4b25b8a79250b63a1dea42739173f7c0fc723fc/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L678)):*

> ```
> else
> ```
We don't need to do an operation if it equals 0. Perhaps change `else` to `else if (_projectDoc.Data.Sync.QueuedCount < 0)`? Although I have a feeling that it doesn't actually result in an actual operation if the value is already 0 so perhaps what you have is fine?
</blockquote></details>

An op isn't created when it is already zero - I added this "reset" block to clean up any erroneous negatives (like I created before I wrote the cancel code on my dev environment), assuming that some projects might have negative counts due to the process terminating abnormally before cancel support is added. I created my erroneous negatives by ending the process strangely, then hangfire restarting the job automatically.

It is no problem to remove this else block though - the cancel job code does not require it, as process exits are now handled correctly - if you think it is overkill.
___
*[src/SIL.XForge.Scripture/Services/SyncService.cs, line 48 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-MbZuyOhDzjXhgwqS_21:-Mb_AX9y-6449ilPulsD:b-9k9pi1) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4b25b8a79250b63a1dea42739173f7c0fc723fc/src/SIL.XForge.Scripture/Services/SyncService.cs#L48)):*

> ```
> projectSecret = null;
> ```
Isn't this a problem to proceed past this point?
</blockquote></details>

Yes - good spotting - thanks!
___
*[src/SIL.XForge.Scripture/Services/SyncService.cs, line 160 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-MbZvQPF2-KZFc2QzSOS:-Mb_ApaF5beKZrySAB2z:bz35sap) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4b25b8a79250b63a1dea42739173f7c0fc723fc/src/SIL.XForge.Scripture/Services/SyncService.cs#L160)):*

> ```
> projectDoc.Data.SyncDisabled
> ```
Hmm... if the sync is already started presumably it wasn't disabled when it was started. Shouldn't we allow the user to still cancel if the sync has been disabled since?
</blockquote></details>

Good spotting...removed!

irahopkinson (commented):
Great. I've converted the PR to a draft so no one else merges it without the rollback code if @ddaspit approves the rest of this PR.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051#-:-Mb_iGsvBrVUqAT2jAj7:b20ut7b)* status: 31 of 50 files reviewed, 1 unresolved discussion (waiting on @ddaspit)

ddaspit (commented):
> Reviewed 19 of 19 files at r10, 3 of 3 files at r11.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

ddaspit (commented):
> Reviewed 1 of 24 files at r12, 13 of 19 files at r13, 3 of 3 files at r14, 7 of 7 files at r15.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

irahopkinson (commented):
> Reviewed 1 of 14 files at r1, 1 of 24 files at r12, 3 of 3 files at r14, 20 of 20 files at r17.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1051)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #1057, opened 2021-06-03
Sync: Consider data to be in sync if no SyncedToRepositoryVersion or …

marksvc:
…is Res

- Projects that were last synced before we started recording
SyncedToRepositoryVersion will not have SyncedToRepositoryVersion set.
Currently that results in isDataInSync being false, and so we don't
write out SF data to the hg repo.
- Set isDataInSync = true if SyncedToRepositoryVersion == null.
- We might instead here to just set isDataInSync = true, and only
set it to false on certain conditions. But for now I'm spelling out
the
situations in code.

ParatextService:
- Make GetLatestShareVersion return null if the project id is for a
resource. There is no hg repository to examine for a commit id.
- Pass in IHgWrapper so we can mock it.

HgWrapper:
- GetLastPublicRevision() can't be easily mocked as a static method.
- Change it to a non-static method.
- Add method to IHgWrapper interface.
- JwtInternetSharedRepositorySource now has an instance of HgWrapper
so it can call GetLastPublicRevision().

===

I am still working on testing the result in a running SF, which has been difficult today because of network trouble getting to the PT servers.

I suggest that we review this here, and if approved and merged to sf-live, then we then cherry-pick to master.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1057)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/HgWrapper.cs
+++ b/src/SIL.XForge.Scripture/Services/HgWrapper.cs
@@ -31,9 +31,9 @@ namespace SIL.XForge.Scripture.Services
         /// <summary>
         /// Get the most recent revision id of the commit from the last push or pull with the PT send/receive server.
         /// </summary>
-        public static string GetLastPublicRevision(string repository)
+        public string GetLastPublicRevision(string repository)
         {
-            string ids = RunCommand(repository, "log --rev \"public()\" --template \"{node}\n\"");
+            string ids = HgWrapper.RunCommand(repository, "log --rev \"public()\" --template \"{node}\n\"");
             return ids.Split(new[] { "\n" }, StringSplitOptions.RemoveEmptyEntries).LastOrDefault();
         }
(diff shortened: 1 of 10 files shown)


marksvc:
Testing showed that things are working well, as described in chat. 
The special case of a sync _failure_, where we then have `dateInSync: false` but no `syncedToRepositoryVersion` resulted in losing the SF text edit, so I will want to think more about that case. But it appears that this commit is an improvement to the current sf-live.

ddaspit (approved):
:lgtm:

> Reviewed 8 of 8 files at r1, 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1057#-:-MbK5lALCLubwdkUvw2P:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1063, opened 2021-06-07
SF-1287 Update text permissions on user join

marksvc:
- Permissions were being updated in SF, but on Synchronization. This
means a PT user who joins a SF project won't have their expected PT
permissions until the next Synchronization.
- Separate out permissions updating in ParatextSyncRunner, leaving a
separate UpdateDocsAsync() and UpdatePermissionsAsync(). Move it into
SFProjectService, where it is now called both by
ParatextSyncRunner.RunAsync() and by
SFProjectService.AddUserToProjectAsync().
- Users can join a project in a number of ways. This commit handles a
user joining by connecting a project to SF for the first time
(SFProjectService.CreateProjectAsync()), joining a project by a link
(SFProjectService.CheckLinkSharingAsync()), or "connecting" an
already-connected project using the Connect Project page
(ProjectService.AddUserAsync()).
- Users may not only have a PT Role on a PT project, but they may or
may not have permissions to access a DBL resource. There are separate
sources to determine who has what PT project Role or who has DBL
resource access permission.
- SFProjectService.UpdatePermissionsAsync() uses the joining user's PT
username and PT access tokens to fetch permission information. It then
updates permissions in SF for all SF users of a given project. The
joining user must therefore have adequate access to fetch this
information, or AddUserToProjectAsync() will not call
UpdatePermissionsAsync() when the user joins.
- The change of s/paratextId/projectPTId/ and s/projectId/projectPTId/
in ParatextService.cs was done to reduce ambiguity and provide
clarity.

Because this change involved some cut-and-pasting, like extract method and moving chunks of code from ParatextSyncRunner/Tests to SFProjectService/Tests, the work was done as a series of commits to ease parts of code review. If desired to help in code review, these commits are available for now in branch `task/new-user-permissions`.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1063)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/IParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/IParatextService.cs
@@ -13,24 +13,25 @@ namespace SIL.XForge.Scripture.Services
         Task<IReadOnlyList<ParatextProject>> GetProjectsAsync(UserSecret userSecret);
         string GetParatextUsername(UserSecret userSecret);
         Task<Attempt<string>> TryGetProjectRoleAsync(UserSecret userSecret, string paratextId);
-        Task<IReadOnlyDictionary<string, string>> GetProjectRolesAsync(UserSecret userSecret, string projectId);
-        bool IsProjectLanguageRightToLeft(UserSecret userSecret, string ptProjectId);
+        Task<IReadOnlyDictionary<string, string>> GetProjectRolesAsync(UserSecret userSecret, string paratextId);
+        bool IsProjectLanguageRightToLeft(UserSecret userSecret, string paratextId);
 
         Task<IReadOnlyList<ParatextResource>> GetResourcesAsync(string userId);
+        bool IsResource(string paratextId);
         Task<string> GetResourcePermissionAsync(string paratextId, string userId);
         Task<IReadOnlyDictionary<string, string>> GetParatextUsernameMappingAsync(UserSecret userSecret,
             string paratextId);
         Task<Dictionary<string, string>> GetPermissionsAsync(UserSecret userSecret, SFProject project,
             IReadOnlyDictionary<string, string> ptUsernameMapping, int book = 0, int chapter = 0);
 
-        IReadOnlyList<int> GetBookList(UserSecret userSecret, string ptProjectId);
-        string GetBookText(UserSecret userSecret, string ptProjectId, int bookNum);
-        Task PutBookText(UserSecret userSecret, string ptProjectId, int bookNum, string usx,
+        IReadOnlyList<int> GetBookList(UserSecret userSecret, string paratextId);
+        string GetBookText(UserSecret userSecret, string paratextId, int bookNum);
+        Task PutBookText(UserSecret userSecret, string paratextId, int bookNum, string usx,
             Dictionary<int, string> chapterAuthors = null);
-        string GetNotes(UserSecret userSecret, string ptProjectId, int bookNum);
-        void PutNotes(UserSecret userSecret, string ptProjectId, string notesText);
-        string GetLatestSharedVersion(UserSecret userSecret, string ptProjectId);
+        string GetNotes(UserSecret userSecret, string paratextId, int bookNum);
+        void PutNotes(UserSecret userSecret, string paratextId, string notesText);
+        string GetLatestSharedVersion(UserSecret userSecret, string paratextId);
 
-        Task SendReceiveAsync(UserSecret userSecret, string ptTargetId, IProgress<ProgressState> progress = null);
+        Task SendReceiveAsync(UserSecret userSecret, string paratextId, IProgress<ProgressState> progress = null);
     }
 }
--- a/src/SIL.XForge.Scripture/Services/ISFProjectService.cs
+++ b/src/SIL.XForge.Scripture/Services/ISFProjectService.cs
@@ -2,6 +2,7 @@ using System.Threading.Tasks;
 using System.Collections.Generic;
 using SIL.XForge.Scripture.Models;
 using SIL.XForge.Services;
+using SIL.XForge.Realtime;
 
 namespace SIL.XForge.Scripture.Services
 {
@@ -22,5 +23,6 @@ namespace SIL.XForge.Scripture.Services
         bool IsSourceProject(string projectId);
         Task<IEnumerable<TransceleratorQuestion>> TransceleratorQuestions(string curUserId, string projectId);
         Task<bool> HasTransceleratorQuestions(string curUserId, string projectId);
+        Task UpdatePermissionsAsync(string curUserId, IDocument<SFProject> projectDoc);
     }
 }
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -151,12 +151,14 @@ namespace SIL.XForge.Scripture.Services
         }
 
         /// <summary>
-        /// Synchronizes the text and notes data on the SF server with the data on the Paratext server.
+        /// Synchronizes the text and notes data on the SF server with the data on the Paratext server, for the PT
+        /// project referred to by paratextId. Or if paratextId refers to a DBL Resource, update the local copy of the
+        /// resource if needed.
         /// </summary>
-        public async Task SendReceiveAsync(UserSecret userSecret, string ptTargetId,
+        public async Task SendReceiveAsync(UserSecret userSecret, string paratextId,
             IProgress<ProgressState> progress = null)
         {
-            if (userSecret == null || ptTargetId == null) { throw new ArgumentNullException(); }
+            if (userSecret == null || paratextId == null) { throw new ArgumentNullException(); }
 
             IInternetSharedRepositorySource source = await GetInternetSharedRepositorySource(userSecret.Id);
             IEnumerable<SharedRepository> repositories = source.GetRepositories();
@@ -164,37 +166,37 @@ namespace SIL.XForge.Scripture.Services
             IEnumerable<string> projectGuids = projectsMetadata.Select(pmd => pmd.ProjectGuid.Id);
             Dictionary<string, ParatextProject> ptProjectsAvailable =
                 GetProjects(userSecret, repositories, projectsMetadata).ToDictionary(ptProject => ptProject.ParatextId);
-            if (!projectGuids.Contains(ptTargetId))
+            if (!projectGuids.Contains(paratextId))
             {
                 // See if this is a resource
                 IReadOnlyList<ParatextResource> resources = await this.GetResourcesInternalAsync(userSecret.Id, true);
-                ParatextResource resource = resources.SingleOrDefault(r => r.ParatextId == ptTargetId);
+                ParatextResource resource = resources.SingleOrDefault(r => r.ParatextId == paratextId);
                 if (resource != null)
                 {
                     ptProjectsAvailable.Add(resource.ParatextId, resource);
                 }
                 else
                 {
-                    _logger.LogWarning($"The target project did not have a full name available {ptTargetId}");
+                    _logger.LogWarning($"The project with PT ID {paratextId} did not have a full name available.");
                 }
             }
-            if (!ptProjectsAvailable.TryGetValue(ptTargetId, out ParatextProject targetPtProject))
+            if (!ptProjectsAvailable.TryGetValue(paratextId, out ParatextProject ptProject))
             {
                 throw new ArgumentException(
-                    $"PT projects with the following PT ids were requested but without access or they don't exist: {ptTargetId}");
+                    $"PT projects with the following PT ids were requested but without access or they don't exist: {paratextId}");
             }
 
-            EnsureProjectReposExists(userSecret, targetPtProject, source);
+            EnsureProjectReposExists(userSecret, ptProject, source);
             StartProgressReporting(progress);
-            if (!(targetPtProject is ParatextResource))
+            if (!(ptProject is ParatextResource))
             {
-                SharedProject targetSharedProj = SharingLogicWrapper.CreateSharedProject(ptTargetId,
-                    targetPtProject.ShortName, source.AsInternetSharedRepositorySource(), repositories);
+                SharedProject sharedProj = SharingLogicWrapper.CreateSharedProject(paratextId,
+                    ptProject.ShortName, source.AsInternetSharedRepositorySource(), repositories);
                 string username = GetParatextUsername(userSecret);
                 // Specifically set the ScrText property of the SharedProject to indicate the project is available locally
-                targetSharedProj.ScrText = ScrTextCollection.FindById(username, ptTargetId);
-                targetSharedProj.Permissions = targetSharedProj.ScrText.Permissions;
-                List<SharedProject> sharedPtProjectsToSr = new List<SharedProject> { targetSharedProj };
+                sharedProj.ScrText = ScrTextCollection.FindById(username, paratextId);
+                sharedProj.Permissions = sharedProj.ScrText.Permissions;
+                List<SharedProject> sharedPtProjectsToSr = new List<SharedProject> { sharedProj };
 
                 // TODO report results
                 List<SendReceiveResult> results = Enumerable.Empty<SendReceiveResult>().ToList();
@@ -227,6 +229,17 @@ namespace SIL.XForge.Scripture.Services
             return await this.GetResourcesInternalAsync(userId, false);
         }
 
+        /// <summary>
+        /// Is the PT project referred to by `paratextId` a DBL resource?
+        /// </summary>
+        public bool IsResource(string paratextId)
+        {
+            return paratextId?.Length == SFInstallableDblResource.ResourceIdentifierLength;
+        }
+
+        /// <summary>
+        /// Returns `userSecret`'s role on a PT project according to the PT Registry.
+        /// </summary>
         public async Task<Attempt<string>> TryGetProjectRoleAsync(UserSecret userSecret, string paratextId)
         {
             if (userSecret.ParatextTokens == null)
@@ -255,20 +268,20 @@ namespace SIL.XForge.Scripture.Services
         }
 
         /// <summary>
-        /// Gets the permission a user has to access a resource.
+        /// Gets the permission a user has to access a resource, according to a DBL server.
         /// </summary>
         /// <param name="paratextId">The paratext resource identifier.</param>
         /// <param name="userId">The user identifier.</param>
         /// <returns>
-        /// A dictionary of permissions where the key is the user ID and the value is the permission
+        /// Read or None.
         /// </returns>
         /// <remarks>
         /// See <see cref="TextInfoPermission" /> for permission values.
         /// </remarks>
         public async Task<string> GetResourcePermissionAsync(string paratextId, string userId)
         {
-            // See if the source is a resource
-            if (paratextId.Length != SFInstallableDblResource.ResourceIdentifierLength)
+            // See if the source is even a resource
+            if (!IsResource(paratextId))
             {
                 // Default to no permissions for projects used as sources
                 return TextInfoPermission.None;
@@ -301,7 +314,7 @@ namespace SIL.XForge.Scripture.Services
             string paratextId)
         {
             // Skip all the work if the project is a resource. Resources don't have project members
-            if (paratextId.Length == SFInstallableDblResource.ResourceIdentifierLength)
+            if (IsResource(paratextId))
             {
                 return new Dictionary<string, string>();
             }
@@ -340,8 +353,7 @@ namespace SIL.XForge.Scripture.Services
         {
             var permissions = new Dictionary<string, string>();
 
-            // See if the source is a resource
-            if (project.ParatextId.Length == SFInstallableDblResource.ResourceIdentifierLength)
+            if (IsResource(project.ParatextId))
             {
                 foreach (string uid in project.UserRoles.Keys)
                 {
@@ -411,9 +423,9 @@ namespace SIL.XForge.Scripture.Services
         }
 
         public async Task<IReadOnlyDictionary<string, string>> GetProjectRolesAsync(UserSecret userSecret,
-            string projectId)
+            string paratextId)
         {
-            if (projectId.Length == SFInstallableDblResource.ResourceIdentifierLength)
+            if (IsResource(paratextId))
             {
                 // Resources do not have roles
                 return new Dictionary<string, string>();
@@ -423,7 +435,7 @@ namespace SIL.XForge.Scripture.Services
                 // Paratext RegistryServer has methods to do this, but it is unreliable to use it in a multi-user
                 // environment so instead we call the registry API.
                 string response = await CallApiAsync(userSecret, HttpMethod.Get,
-                    $"projects/{projectId}/members");
+                    $"projects/{paratextId}/members");
                 var members = JArray.Parse(response);
                 return members.OfType<JObject>()
                     .Where(m => !string.IsNullOrEmpty((string)m["userId"]) && !string.IsNullOrEmpty((string)m["role"]))
@@ -432,25 +444,25 @@ namespace SIL.XForge.Scripture.Services
         }
 
         /// <summary> Determine if a specific project is in a right to left language. </summary>
-        public bool IsProjectLanguageRightToLeft(UserSecret userSecret, string ptProjectId)
+        public bool IsProjectLanguageRightToLeft(UserSecret userSecret, string paratextId)
         {
-            ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), ptProjectId);
+            ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
             return scrText == null ? false : scrText.RightToLeft;
         }
 
         /// <summary> Get list of book numbers in PT project. </summary>
-        public IReadOnlyList<int> GetBookList(UserSecret userSecret, string ptProjectId)
+        public IReadOnlyList<int> GetBookList(UserSecret userSecret, string paratextId)
         {
-            ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), ptProjectId);
+            ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
             if (scrText == null)
                 return Array.Empty<int>();
             return scrText.Settings.BooksPresentSet.SelectedBookNumbers.ToArray();
         }
 
         /// <summary> Get PT book text in USX, or throw if can't. </summary>
-        public string GetBookText(UserSecret userSecret, string ptProjectId, int bookNum)
+        public string GetBookText(UserSecret userSecret, string paratextId, int bookNum)
         {
-            ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), ptProjectId);
+            ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
             if (scrText == null)
                 throw new DataNotFoundException("Can't get access to cloned project.");
             string usfm = scrText.GetText(bookNum);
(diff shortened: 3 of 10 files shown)


irahopkinson (changes requested):
Thanks for your work on this. I'm looking forward to getting this merged and released. I only found minor things, the functionality looks good.

> Reviewed 9 of 9 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-:-Mbp8itrBxOdne3XLPeE:bx6cxq)* status: all files reviewed, 6 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 233 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-MboXgA31BSPWA25FQNo:-MboXgA31BSPWA25FQNp:bp3qzid) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/src/SIL.XForge.Scripture/Services/ParatextService.cs#L233)):*
> ```
> string projectPTId
> ```
Thanks for noticing the inconsistency with these PT Id arguments and fixing them. However, I think you might like to rethink this rename in this file. The `PT` acronym disappears to mostly insignificant with this name. Given that we are always in the context of a project, adding that to the name doesn't add anything other than more I have to read and discard as irrelevant. Also given that the property value that is ultimately passed down here is `SFProject.ParatextId` then I would change these all to `paratextId`. You should also change them in `IParatextService` to keep things consistent there too.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 601 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-Mbot7736lWIuY0CBZto:-Mbot7736lWIuY0CBZtp:bcqzxpj) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L601)):*
> ```
> projectPTIdIsResourceId
> ```
Thanks for pulling this out into its own method. It's a better idea to name a method for what it does rather than how its implemented. We should be free to change how its implemented without needing to change its name. How about simply `projectIsResource`? There are other places in the code where we would benefit from access to this method and keep our code DRY. Is there a better place to house this? I think ParatextService is common to all.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 601 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-MbouQkWBSsPV9W9z1op:-MbouQkWBSsPV9W9z1oq:b-856ngr) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L601)):*
> ```
> string projectPDId
> ```
I don't know what a PDId is. Should this be `paratextId`?
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 622 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-Mboxk2l6IuwZcuo-fjn:-Mboxk2l6IuwZcuo-fjo:b-r10y4i) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L622)):*
> ```
> targetParatextId, targetBooks, targetIsResource, etc
> ```
From where this was extracted we needed to make the distinction of source or target. In this method's new location that is no longer necessary. Remove references to `target`.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 665 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-MbovyM8CfdHx2TITKNR:-MbovyM8CfdHx2TITKNS:buwykg1) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L665)):*
> ```
> int chapterIndex = chapters.FindIndex(c => c.Number == chapter.Number);
> ```
Isn't there a better way in C# to access the index in a `foreach` loop without having to 'find' it in each iteration? Here and below.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 861 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-Mbp0G117a3UVxL6dK9Y:-Mbp0G117a3UVxL6dK9Z:b31kxgq) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L861)):*
> ```
> string targetprojectSFId, string sourceprojectSFId
> ```
Both these arguments aren't following the capitalization convention. You have changed `ID` to `Id` but `project` should be capitalized. Perhaps this was a search and replace faux par.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1063)* status: all files reviewed, 6 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 233 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-MboXgA31BSPWA25FQNo:-Mbwedmq6yDsGPNcNicA:b-5zexg8) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/src/SIL.XForge.Scripture/Services/ParatextService.cs#L233)):*
Thanks for your thoughts on this. IIRC, the reason I started on doing some renames here was because I wanted to call a method in ParatextService, but it was unclear whether the `paratextId` I was supposed to pass in was for a project or a user. So I disambiguated it by writing 'project' in the variable name. Imagine, for example, if you were calling 
```cs 
GetParatextUsernameMappingAsync(UserSecret userSecret, string paratextId)
```

Are you providing the user secret and user Paratext ID to get back a username? No. And it's not impossible to figure out what one is supposed to do. That method even has a method comment, and we can read the method source code if needed.

I am guessing the method I was using was `TryGetProjectRoleAsync(UserSecret userSecret, string paratextId)`.  With some pause to consider, "If I want to get the user's role on a PT project, and there are 2 arguments, and they are userSecret and some paratext-related ID, then the paratext-related ID must be about the _project_, not the _user_." But making this method's arguments immediately clear that the paratext-related ID is a paratext _project_ ID, would be a welcome change.

I also disagree that "Given that we are always in the context of a project, adding that to the name doesn't add anything" is adequate. Although many methods in this class do regard Paratext projects, the class itself is ParatextService, which is a class providing interaction with Paratext DLLs and APIs. Not "ProjectService" or "PTProjectService". 

>  Also given that the property value that is ultimately passed down here is SFProject.ParatextId then I would change these all to paratextId

A couple things to consider: 
1. When you call this method, you don't know that other clients are passing SFProject.ParatextId unless you go look at other references.
2. The `ParatextId` in `SFProject.ParatextId` has the very significant context of being the ParatextId _of an SF project_. So I don't think that means "ParatextId" can always mean a PT project ID. There is also a User.ParatextId, as well. 

So, I disagree, but one person's explanatory comments, variable names, and method names, are another person's irrelevant hindrance in the way. I will remove these in my revision, but can set them back if I have convinced you of my point of view :).


BTW, I have found it confusing working in other parts of SF, where we pass around IDs, as to whether those IDs are SF IDs or PT IDs. Usually that's probably been in the sync area, but consider this snippet from ParatextService:

```cs
        /// <summary>Get Paratext resources that a user has access to. </summary>
        public async Task<IReadOnlyList<ParatextResource>> GetResourcesAsync(string userId)
        {
            return await this.GetResourcesInternalAsync(userId, false);
        }
```

Is `userId` an SF user ID or a PT user ID? It's not clear. It would make sense for it to be a PT user ID, since this is in class _ParatextService_. Further, since resource permissions pertain to PT users, it would make sense to be a PT user ID. But my experience has been that it's quite rare for us to pass around a PT user ID, so maybe it's a SF user ID, which we use frequently. 

I dug down a few layers, and found that it's using `userId` to look in an IRepository<UserSecret>, which means: It's an SF user ID. I won't remember that :), and if I ever call `GetResourcesAsync(string userId)` (or perhaps another of the methods in ParatextService that have an argument of `string userId`), I may need to look it up again to understand what I need to pass to the method. So similarly, my position here would be to name the argument `userSFId`, so that a caller can apply their problem solving abilities to other things. 
So that's some of my thinking and motivation regarding this.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 601 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-Mbot7736lWIuY0CBZto:-Mbwt_Eq0_Zqnca-FAj0:bx21wsu) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L601)):*

> ```
> projectPTIdIsResourceId
> ```
Thanks for pulling this out into its own method. It's a better idea to name a method for what it does rather than how its implemented. We should be free to change how its implemented without needing to change its name. How about simply `projectIsResource`? There are other places in the code where we would benefit from access to this method and keep our code DRY. Is there a better place to house this? I think ParatextService is common to all.
</blockquote></details>

Thank you. Done. How about `IsResource(string paratextId)`?
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 601 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-MbouQkWBSsPV9W9z1op:-Mbwtp_65lr4QWLdWca5:b28ec48) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L601)):*

> ```
> string projectPDId
> ```
I don't know what a PDId is. Should this be `paratextId`?
</blockquote></details>

Ha; thank you. 
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 622 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-Mboxk2l6IuwZcuo-fjn:-MbrovqJ3GCzaoBtwsoc:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L622)):*

> ```
> targetParatextId, targetBooks, targetIsResource, etc
> ```
From where this was extracted we needed to make the distinction of source or target. In this method's new location that is no longer necessary. Remove references to `target`.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 665 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-MbovyM8CfdHx2TITKNR:-MbrtAzT7EFBi-q7aJPX:bvindoa) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L665)):*

> ```
> int chapterIndex = chapters.FindIndex(c => c.Number == chapter.Number);
> ```
Isn't there a better way in C# to access the index in a `foreach` loop without having to 'find' it in each iteration? Here and below.
</blockquote></details>

Thank you. I was focusing more on the context of the change, but there are some improvements we can make here. 

It looks like we could instead write

```cs
for (int chapterIndex = 0; chapterIndex < chapters.Count; chapterIndex++) 
{
  chapterPermissionOperations.Add((textIndex, chapterIndex, permissions));
}
```

or even better (because it doesnt involve a `for` loop (which IMO opens the door for mistakes, compared to something like `foreach` or Linq):

```cs
chapterPermissionOperations.AddRange(
  chapters.Select((Chapter chapter, int chapterIndex) => (textIndex, chapterIndex, permissions) ));
```

I made a couple changes like that .I also am modifying the UpdatePermissionsAsync() method so that it compiles a set of book permissions, and chapter permissions, and then submits them all as one op at the end of the method. Putting all the changes together in one op at the end may reduce a chance of only doing half of something and then crashing, and increasing the chance of doing all or nothing. It may also improve speed since it may reduce the number of connections to the DB, and the amount of change requests the DB receives. (Maybe.)
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 861 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-Mbp0G117a3UVxL6dK9Y:-Mbwe5wf8maXzVEr4ldX:bvs55d0) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L861)):*

> ```
> string targetprojectSFId, string sourceprojectSFId
> ```
Both these arguments aren't following the capitalization convention. You have changed `ID` to `Id` but `project` should be capitalized. Perhaps this was a search and replace faux par.
</blockquote></details>

Thank you

irahopkinson (changes requested):
> Reviewed 5 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1063)* status: 9 of 10 files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @marksvc)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 233 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-MboXgA31BSPWA25FQNo:-MbxwwJr07xgmUWpfcBw:bajus5a) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/src/SIL.XForge.Scripture/Services/ParatextService.cs#L233)):*

Oops. I'm also updating IParatextService.cs. And I notice that SendReceiveAsync() was taking a ptTargetId. I have studied SendReceiveAsync() and I think it is used for both source and target projects, not just targets. I have changed argument name ptTargetId to paratextId,  added to the method comment about what I learned, and  renamed a couple local variables to not mention 'target'.
</blockquote></details>

Good job spotting that extra argument that needed changing.

You make a good case. I'm much happier with using the the property name `ParatextId` than `PTId` though in the argument name. You could change them all from `paratextId` to `projectParatextId` but that makes the argument name long but I'd be Ok with that. My inclination is to leave them as `paratextId` and make sure they all have param comments, e.g. `<param name="paratextId">The project ParatextId.</param>`. But the verbose name `projectParatextId` is probably the easiest to read and work with and you can work it out more easily from the method call just by hovering without having to go read the method comments.

In the case of `GetParatextUsernameMappingAsync` you have to ask what it is mapping to. So perhaps the method would be better named `GetParatextUsernameProjectMappingAsync` then we would at least expect at a minimum to supply something for the user and something for the project. At the moment its saving grace is that it has good method comments (probably thanks to you).
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 666 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-Mby5_N3355p9wpRBIOg:-Mby5_N3355p9wpRBIOh:b-8t3ufh) ([raw file](https://github.com/sillsdev/web-xforge/blob/3390364831866b9ba01e449195b5b8d0d8f2f932/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L666)):*
> ```
> await Task.WhenAll
> ```
@pmachapman I recall you had trouble with running these in parallel and it was better done serially. Of course I may be remembering about something else. Thoughts?
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 861 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-Mbp0G117a3UVxL6dK9Y:-Mby-GiF8Jiu4ZO6NAp5:by4ghr6) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L861)):*

Thank you
</blockquote></details>

Good but you still need to fix this in `SetupPTData` and `SetupPTDataForProjectIds`.

irahopkinson (commented):
> Reviewed 1 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1063)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc)

pmachapman (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1063)* status: all files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @marksvc)

___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 666 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-Mby5_N3355p9wpRBIOg:-Mc6GOzi8ONErA0n5Zt1:b-4gm1ss) ([raw file](https://github.com/sillsdev/web-xforge/blob/3390364831866b9ba01e449195b5b8d0d8f2f932/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L666)):*

> ```
> await Task.WhenAll
> ```
@pmachapman I recall you had trouble with running these in parallel and it was better done serially. Of course I may be remembering about something else. Thoughts?
</blockquote></details>

I don't really remember, but this may have been regarding resource permission retrieval - `ParatextService.GetResourcePermissionAsync()` has a lock to stop concurrent access, so this code will just run in serial if called in this way (it isn't in this method).

I did a quick review of the code in `ParatextService` and Paratext, and it *should* work concurrently for the same project - the way `ScrText` is used in Paratext though, it isn't used in this way, so we may be in random/rare bug territory.

The `PermissionManager` code in Paratext looks to be very suitable to run concurrently (the code has locks in all the right places), so if concurrent execution is preferred for speed, we could just load the `ScrText` for the project once via `LazyScrTextCollection` in `SFProject.UpdatePermissionsAsync()`, and refactor `GetPermissionsAsync` so there is an internal method we pass the `ScrText` object to and calculate permissions concurrently for all chapters via the `Task.WhenAll`. This will improve speed and memory usage for this code anyway, as `ParatextService.GetPermissions()` loads an `ScrText` object via the file system every time it is called.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1063)* status: all files reviewed, 2 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 233 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-MboXgA31BSPWA25FQNo:-McB8ISq08rNzA3D2WdJ:b-xfzzwp) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/src/SIL.XForge.Scripture/Services/ParatextService.cs#L233)):*

Good job spotting that extra argument that needed changing.

You make a good case. I'm much happier with using the the property name `ParatextId` than `PTId` though in the argument name. You could change them all from `paratextId` to `projectParatextId` but that makes the argument name long but I'd be Ok with that. My inclination is to leave them as `paratextId` and make sure they all have param comments, e.g. `<param name="paratextId">The project ParatextId.</param>`. But the verbose name `projectParatextId` is probably the easiest to read and work with and you can work it out more easily from the method call just by hovering without having to go read the method comments.

In the case of `GetParatextUsernameMappingAsync` you have to ask what it is mapping to. So perhaps the method would be better named `GetParatextUsernameProjectMappingAsync` then we would at least expect at a minimum to supply something for the user and something for the project. At the moment its saving grace is that it has good method comments (probably thanks to you).
</blockquote></details>

Haha, here you and Jason win the prize for making the method comment :)

Okay. I think what I will do is to leave the argument names as is for now, but will keep in mind potentially renaming them if ambiguity strikes me again here. 
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 601 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-Mbot7736lWIuY0CBZto:-McB5gka-OfDhQ3zREjQ:bj4iwi6) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L601)):*

Thank you. Done. How about `IsResource(string paratextId)`?
</blockquote></details>

Oops; I didn't change some other code to use this method. Done.
___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 666 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-Mby5_N3355p9wpRBIOg:-McBAKGg8OIJj8nymDWi:bprxarr) ([raw file](https://github.com/sillsdev/web-xforge/blob/3390364831866b9ba01e449195b5b8d0d8f2f932/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L666)):*
Thanks guys.

>  so if concurrent execution is preferred for speed, 

:) So my motivation to adjust the code to run this way was so it didn't use a .FindIndex() lookup in each iteration of a foreach loop, as in:

```cs
foreach (Chapter chapter in chapters) 
{
  ...
  int chapterIndex = chapters.FindIndex(...)
  ...
}
```

(And instead I benefit from .Select() giving me the index as an argument to the lambda I pass it.)

I think my default inclination would be to not speed it up without understanding that it's an application bottleneck that needs improvement. But @irahopkinson what is your preference here? This SFProjectService.UpdatePermissionsAsync() method and the ParatextService.GetPermissionsAsync() methods are not using one another very sensibly, as GetPermissionsAsync() fetches info on blocks of data, returns the _one thing we asked for_, and then we iterate to a next thing and _call GetPermissionsAsync()_ again, making it fetch the same block of data again and return the next one thing we ask for. 
So an improvement would be to instead ask GetPermissionsAsync to give us a block of data (or similar do-it-once implementation, wherever that falls in the methods), thus reducing the processing by an order of magnitude. I didn't notice this until today. 
I think I'm still inclined to not try to make it optimized, without knowing that it's a bottle neck. But what would be your preference for this kind of thing? 

(I think maybe I also didn't entirely understand what Peter is meaning, so can chat more with Peter if you prefer more work in this area at this time.)
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 861 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-Mbp0G117a3UVxL6dK9Y:-McB6jGe3A08pMce182m:bfp30y3) ([raw file](https://github.com/sillsdev/web-xforge/blob/191fc1dd66160bee0b2cd8a7fdc25244ee03be07/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L861)):*

Good but you still need to fix this in `SetupPTData` and `SetupPTDataForProjectIds`.
</blockquote></details>

Thank you. I thought I had looked for more instances, but may not.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1063)* status: 7 of 10 files reviewed, 2 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 666 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-Mby5_N3355p9wpRBIOg:-McBYHO96KpID_qPJJQH:b5rkx7h) ([raw file](https://github.com/sillsdev/web-xforge/blob/3390364831866b9ba01e449195b5b8d0d8f2f932/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L666)):*

Thanks guys.

>  so if concurrent execution is preferred for speed, 

:) So my motivation to adjust the code to run this way was so it didn't use a .FindIndex() lookup in each iteration of a foreach loop, as in:

```cs
foreach (Chapter chapter in chapters) 
{
  ...
  int chapterIndex = chapters.FindIndex(...)
  ...
}
```

(And instead I benefit from .Select() giving me the index as an argument to the lambda I pass it.)

I think my default inclination would be to not speed it up without understanding that it's an application bottleneck that needs improvement. But @irahopkinson what is your preference here? This SFProjectService.UpdatePermissionsAsync() method and the ParatextService.GetPermissionsAsync() methods are not using one another very sensibly, as GetPermissionsAsync() fetches info on blocks of data, returns the _one thing we asked for_, and then we iterate to a next thing and _call GetPermissionsAsync()_ again, making it fetch the same block of data again and return the next one thing we ask for. 
So an improvement would be to instead ask GetPermissionsAsync to give us a block of data (or similar do-it-once implementation, wherever that falls in the methods), thus reducing the processing by an order of magnitude. I didn't notice this until today. 
I think I'm still inclined to not try to make it optimized, without knowing that it's a bottle neck. But what would be your preference for this kind of thing? 

(I think maybe I also didn't entirely understand what Peter is meaning, so can chat more with Peter if you prefer more work in this area at this time.)

</blockquote></details>

Peter and I talked on the phone and he showed me what he means. Thanks Peter. 
I'm working on that change.
It sounds like it may not be so significant to optimize how we do the fetching of the blocks of permission data. 
The method structure can instead be changed such that we have

public async GetPermissionsAsync() 
{
  var scrText = fetch_it()
  return GetPermissionsCore()
}

internal GetPermissionsCore(scrText) 
{
  // most of the current GetPermissionsAsync() here. 
}

And then the UpdatePermissionsAsync() code can switch to, sometimes instead of calling GetPermissionsAsync() over and over, it could manually do `var scrText = fetch_it()`, and then call `GetPermissionsCore(scrText)` over and over, and gain performance by reducing the iterations of `var scrText = fetch_it()` by an order of magnitude. 

So I'm working on that.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1063)* status: 7 of 10 files reviewed, 2 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 666 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-Mby5_N3355p9wpRBIOg:-McGdkps8FBrPbxGfmI-:b-z76dpj) ([raw file](https://github.com/sillsdev/web-xforge/blob/3390364831866b9ba01e449195b5b8d0d8f2f932/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L666)):*

Peter and I talked on the phone and he showed me what he means. Thanks Peter. 
I'm working on that change.
It sounds like it may not be so significant to optimize how we do the fetching of the blocks of permission data. 
The method structure can instead be changed such that we have

public async GetPermissionsAsync() 
{
  var scrText = fetch_it()
  return GetPermissionsCore()
}

internal GetPermissionsCore(scrText) 
{
  // most of the current GetPermissionsAsync() here. 
}

And then the UpdatePermissionsAsync() code can switch to, sometimes instead of calling GetPermissionsAsync() over and over, it could manually do `var scrText = fetch_it()`, and then call `GetPermissionsCore(scrText)` over and over, and gain performance by reducing the iterations of `var scrText = fetch_it()` by an order of magnitude. 

So I'm working on that. 
</blockquote></details>

Improving performance in the way I was working has had difficulties amidst exposing ScrText or ScrTextCollection outside of ParatextService (like to SFProjectService) either via return types, argument types, or DI; whether I can or should specify an `internal` method on the IParatextService interface (and GetPermsCore() needs to be on the interface because we use the interface); fiddling with async + mocks; etc. I gave up on that direction for now and am implementing a ParatextService.GetAllPermissionsAsync() method that will return a _set_ of all project permissions for a user (project-level, book-level, chapter-level). 

Because this is a performance improvement over the implemented functionality, and because it didn't all just come together today, I suggest that we continue to move this PR thru the system and I can work on the performance change as a followup.

irahopkinson (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1063#-:-McHvHFhCa11xBr-c-h6:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1064, opened 2021-06-09
SF-1320b: Sync rollback support

pmachapman:
This change implements a "transaction styled" commit/rollback feature for the Realtime Server, and backup/restore support for Paratext repositories (based on Paratext's Backup/Restore). These two features are used by `ParatextSyncRunner` to allow rollback of Realtimeserver and Mercurial changes when a sync is cancelled.

The transaction implementation is accessed via the `Connection`, and properties can be excluded from the transaction (and so committed immediately), meaning that the sync progress is still updated in real time for the user.

**Other Fixes Included**
 - `HgWrapper.GetLastPublicRevision()` now supports repositories without public commits (i.e. restored from backup)
 - Source projects are no longer synced if the target project has translation suggestions disabled (see `SyncService.SyncAsync()`).
 - `ParatextSyncRunner.CompleteSync()` no longer crashes if the source could not be set, due to an issue contacting the DBL (the sync will rollback however)
 - `SFInstallableDblResource.CheckResourcePermission()` checks for the HTTP status codes in a more robust manner that allows for server message differences
 - Permissions were being incorrectly overwritten on sync, so the `ChapterEqualityComparer` no longer checks permissions (as permissions are handled via `SFProjectService`), and `SFProjectService.UpdatePermissionsAsync()` now uses the permission dictionary comparer to stop unnecessary writes to the project permissions collections.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1064)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -65,6 +65,7 @@
     "transcelerator",
     "transloco",
     "uncheck",
+    "unelevated",
     "uninvite",
     "unsynchronized",
     "usfm",
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts
@@ -116,6 +116,10 @@ export class SFProjectService extends ProjectService<SFProject, SFProjectDoc> {
     return this.onlineInvoke('sync', { projectId: id });
   }
 
+  onlineCancelSync(id: string): Promise<void> {
+    return this.onlineInvoke('cancelSync', { projectId: id });
+  }
+
   onlineUpdateSettings(id: string, settings: SFProjectSettings): Promise<void> {
     return this.onlineInvoke('updateSettings', { projectId: id, settings });
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html
@@ -40,6 +40,17 @@
             [projectDoc]="projectDoc"
             (inProgress)="syncActive = $event"
           ></app-sync-progress>
+          <button
+            *ngIf="syncActive"
+            class="action-button"
+            mdc-button
+            type="button"
+            (click)="cancelSync()"
+            [disabled]="isLoading || !isAppOnline || syncDisabled"
+            id="btn-cancel-sync"
+          >
+            {{ t("cancel") }}
+          </button>
           <span *ngIf="!syncActive" [title]="lastSyncDate" id="date-last-sync" class="sync-info">
             {{ lastSyncNotice }}
           </span>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss
@@ -20,7 +20,8 @@ h2 {
   align-self: center;
 }
 
-#btn-sync {
+#btn-sync,
+.progress-bar {
   margin-bottom: 1rem;
 }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.spec.ts
@@ -106,7 +106,7 @@ describe('SyncComponent', () => {
     expect(env.component.syncActive).toBe(true);
     expect(env.progressBar).not.toBeNull();
     // Simulate sync in progress
-    env.emitSyncProgress(0, 'testProject01');
+    env.emitSyncProgress(0.25, 'testProject01');
     // Simulate sync error
     env.emitSyncComplete(false, 'testProject01');
     expect(env.component.syncActive).toBe(false);
@@ -131,6 +131,21 @@ describe('SyncComponent', () => {
     const env = new TestEnvironment(true, false, true, false);
     expect(env.syncButton.nativeElement.disabled).toBe(false);
     expect(env.syncDisabledMessage).toBeNull();
+  }));
+
+  it('should report cancelled if sync was cancelled', fakeAsync(() => {
+    const env = new TestEnvironment(true);
+    verify(mockedProjectService.get('testProject01')).once();
+    env.clickElement(env.syncButton);
+    verify(mockedProjectService.onlineSync('testProject01')).once();
+    expect(env.component.syncActive).toBe(true);
+    expect(env.progressBar).not.toBeNull();
+
+    env.clickElement(env.cancelButton);
+    env.emitSyncComplete(false, 'testProject01');
+
+    expect(env.component.syncActive).toBe(false);
+    verify(mockedNoticeService.show('Synchronize for Sync Test Project was cancelled.')).once();
   }));
 });
 
@@ -208,6 +223,10 @@ class TestEnvironment {
 
   get syncButton(): DebugElement {
     return this.fixture.debugElement.query(By.css('#btn-sync'));
+  }
+
+  get cancelButton(): DebugElement {
+    return this.fixture.debugElement.query(By.css('#btn-cancel-sync'));
   }
 
   get progressBar(): DebugElement {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts
@@ -22,8 +22,10 @@ export class SyncComponent extends DataLoadingComponent implements OnInit {
   syncDisabled: boolean = false;
   projectDoc?: SFProjectDoc;
 
-  private paratextUsername?: string;
   private _syncActive: boolean = false;
+  private isSyncCancelled = false;
+  private paratextUsername?: string;
+  private previousLastSyncDate?: Date;
 
   constructor(
     private readonly route: ActivatedRoute,
@@ -53,7 +55,7 @@ export class SyncComponent extends DataLoadingComponent implements OnInit {
     }
   }
 
-  get lastSyncDate() {
+  get lastSyncDate(): string {
     if (
       this.projectDoc == null ||
       this.projectDoc.data == null ||
@@ -66,7 +68,7 @@ export class SyncComponent extends DataLoadingComponent implements OnInit {
   }
 
   get projectName(): string {
-    return this.projectDoc == null || this.projectDoc.data == null ? '' : this.projectDoc.data.name;
+    return this.projectDoc?.data == null ? '' : this.projectDoc.data.name;
   }
 
   get syncActive(): boolean {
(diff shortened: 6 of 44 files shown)


ddaspit (changes requested):
> Reviewed 9 of 19 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1064)* status: 9 of 19 files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[src/SIL.XForge/Realtime/Connection.cs, line 69 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1064#-McJS-zV--EyBTsi1dUm:-McJS-zV--EyBTsi1dUn:b-omoizo) ([raw file](https://github.com/sillsdev/web-xforge/blob/197812df1d77515ad359ae6a8f5b03883306a1c2/src/SIL.XForge/Realtime/Connection.cs#L69)):*
> ```Smalltalk
>                 // We must clear the documents cache so each document has the correct server
>                 _documents.Clear();
>                 _realtimeServer = new QueuedRealtimeServer(_realtimeService.Server);
> ```

Rather than using this decorator for `RealtimeServer`, I would have the `Document` class call the `Connection` class to see if a transaction is in process and queue the operation. That way you don't have to update the `_realtimeServer` field and clear the `_documents` collection, which might cause unexpected behavior.

ddaspit (changes requested):
> Reviewed 10 of 19 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1064)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[src/SIL.XForge.Scripture/Services/JwtInternetSharedRepositorySource.cs, line 156 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1064#-McTXG_L-NahervpUPfD:-McTXG_L-NahervpUPfE:b-q2y3c8) ([raw file](https://github.com/sillsdev/web-xforge/blob/197812df1d77515ad359ae6a8f5b03883306a1c2/src/SIL.XForge.Scripture/Services/JwtInternetSharedRepositorySource.cs#L156)):*
> ```Smalltalk
>         private string GetBaseRevision(string repository)
>         {
>             return _hgWrapper.GetLastPublicRevision(repository, true);
> ```

I find it helpful to specify the explicit parameter name when passing a boolean constant to a method.

```
return _hgWrapper.GetLastPublicRevision(repository, allowEmptyIfRestoredFromBackup: true);
```

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1064)* status: 10 of 22 files reviewed, 2 unresolved discussions (waiting on @ddaspit)

___
*[src/SIL.XForge/Realtime/Connection.cs, line 69 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1064#-McJS-zV--EyBTsi1dUm:-McgHcddDgLEEp68BYgl:b-og0x04) ([raw file](https://github.com/sillsdev/web-xforge/blob/a27c0cbaefa4f638c5832ad913c22cc3c4c01548/src/SIL.XForge/Realtime/Connection.cs#L69)):*

Rather than using this decorator for `RealtimeServer`, I would have the `Document` class call the `Connection` class to see if a transaction is in process and queue the operation. That way you don't have to update the `_realtimeServer` field and clear the `_documents` collection, which might cause unexpected behavior.
</blockquote></details>

I have updated the Document to use the Connection instead of the RealtimeServer (see 96adfe3), to make the logic simpler and achieve the same effect without the collection clearing. Please let me know if this is not satisfactory.
___
*[src/SIL.XForge.Scripture/Services/JwtInternetSharedRepositorySource.cs, line 156 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1064#-McTXG_L-NahervpUPfD:-McgI7De7-1Y_Xp6iVyD:bwnely1) ([raw file](https://github.com/sillsdev/web-xforge/blob/a27c0cbaefa4f638c5832ad913c22cc3c4c01548/src/SIL.XForge.Scripture/Services/JwtInternetSharedRepositorySource.cs#L156)):*

I find it helpful to specify the explicit parameter name when passing a boolean constant to a method.

```
return _hgWrapper.GetLastPublicRevision(repository, allowEmptyIfRestoredFromBackup: true);
```
</blockquote></details>

No problem - fixed in 854b44a.

ddaspit (commented):
> Reviewed 1 of 12 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1064)* status: 11 of 22 files reviewed, 2 unresolved discussions (waiting on @ddaspit and @pmachapman)

___
*[src/SIL.XForge/Realtime/Connection.cs, line 69 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1064#-McJS-zV--EyBTsi1dUm:-McnFpZNEDbCrBLdpmLD:b-rgj72w) ([raw file](https://github.com/sillsdev/web-xforge/blob/a27c0cbaefa4f638c5832ad913c22cc3c4c01548/src/SIL.XForge/Realtime/Connection.cs#L69)):*

I have updated the Document to use the Connection instead of the RealtimeServer (see 96adfe3), to make the logic simpler and achieve the same effect without the collection clearing. Please let me know if this is not satisfactory.
</blockquote></details>

That is fine, but now the `QueuedRealtimeServer` seems unnecessary. Could you move the queueing code to this class and just get rid of the `QueuedRealtimeServer` altogether?

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1064)* status: 10 of 23 files reviewed, 2 unresolved discussions (waiting on @ddaspit)

___
*[src/SIL.XForge/Realtime/Connection.cs, line 69 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1064#-McJS-zV--EyBTsi1dUm:-McqB6Aa3QIe9nqEjfZ4:bjz24wc) ([raw file](https://github.com/sillsdev/web-xforge/blob/a27c0cbaefa4f638c5832ad913c22cc3c4c01548/src/SIL.XForge/Realtime/Connection.cs#L69)):*

That is fine, but now the `QueuedRealtimeServer` seems unnecessary. Could you move the queueing code to this class and just get rid of the `QueuedRealtimeServer` altogether?
</blockquote></details>

No problem - I have merged it into the `Connection` class, and updated/created unit tests to match the change.

ddaspit (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1064)* status: 10 of 23 files reviewed, 2 unresolved discussions (waiting on @ddaspit and @pmachapman)

___
*[src/SIL.XForge/Realtime/Connection.cs, line 48 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1064#-McrStzd4rOKUfyJAOuG:-McrStzd4rOKUfyJAOuH:b-qwbcm5) ([raw file](https://github.com/sillsdev/web-xforge/blob/b01cb4c9ed187705ec24c7962b37746f100dd326/src/SIL.XForge/Realtime/Connection.cs#L48)):*
> ```Smalltalk
>         /// The lock to ensure that the queue is not modified during commit or out of sequence.
>         /// </summary>
>         private readonly SemaphoreSlim _lock = new SemaphoreSlim(1, 1);
> ```

 If we make `_queuedOperations` a `ConcurrentQueue`, then we can get rid of this lock. The main synchronization issue that we need to worry about is multiple operations being submitted in parallel. A `ConcurrentQueue` should be able to handle that case.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1064)* status: 10 of 23 files reviewed, 2 unresolved discussions (waiting on @ddaspit)

___
*[src/SIL.XForge/Realtime/Connection.cs, line 48 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1064#-McrStzd4rOKUfyJAOuG:-McrWx8wAu8T2f9p9O69:b-o3tzva) ([raw file](https://github.com/sillsdev/web-xforge/blob/b01cb4c9ed187705ec24c7962b37746f100dd326/src/SIL.XForge/Realtime/Connection.cs#L48)):*

 If we make `_queuedOperations` a `ConcurrentQueue`, then we can get rid of this lock. The main synchronization issue that we need to worry about is multiple operations being submitted in parallel. A `ConcurrentQueue` should be able to handle that case.
</blockquote></details>

I always forget about `System.Collections.Concurrent` :-) I thought there might be a problem with `SubmitOpAsync`, when it decides whether to submit directly to MongoDB or queue, but I think that the lock that the `Document` uses will cover that. Have updated the `Connection` to use `ConcurrentQueue`.

ddaspit (commented):
> Reviewed 6 of 12 files at r2, 2 of 5 files at r3, 5 of 5 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1064)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[src/SIL.XForge/Realtime/Connection.cs, line 343 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1064#-McxTTYp3t4PyjGhHi2P:-McxTTYqEp9Bj6lrA5D5:bv1l4eo) ([raw file](https://github.com/sillsdev/web-xforge/blob/2c5f7dd371a18591e497c3442958b6ddd4510be5/src/SIL.XForge/Realtime/Connection.cs#L343)):*
> ```Smalltalk
>                     snapshot.Version++;
>                 }
>                 catch (InvocationException)
> ```

I just noticed this. Why are we eating this exception? `applyOp` does not touch the database.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1064)* status: 18 of 23 files reviewed, 1 unresolved discussion (waiting on @ddaspit)

___
*[src/SIL.XForge/Realtime/Connection.cs, line 343 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1064#-McxTTYp3t4PyjGhHi2P:-Mcxe3u6Cr8kpKhRK1_q:b-8idsd6) ([raw file](https://github.com/sillsdev/web-xforge/blob/2c5f7dd371a18591e497c3442958b6ddd4510be5/src/SIL.XForge/Realtime/Connection.cs#L343)):*

I just noticed this. Why are we eating this exception? `applyOp` does not touch the database.
</blockquote></details>

I was writing something about the error, etc, but I had an idea on how to do it better. I've made a commit that sorts this out. Thanks for pointing out the code smell!

The error was because the part of the object being modified by the op was not in the database (such as a new chapter or book updating a project object). I now just pass the current state from the Document, and apply the op to that.

ddaspit (approved):
Good job on this PR.

> Reviewed 5 of 5 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1064#-:-Md01Bop7ODFvp28j3sy:b-72k8eu)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #1068, opened 2021-06-14
Show invitee role on users page

Nateowami:
While considering whether SF-1198 is complete (Allow inviting users to view translation) this seemed like something that was missing, as it's not possible to tell what role a user was invited to.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1068)

PR #1069, opened 2021-06-15
Update invitee role when user is re-invited

Nateowami:
When a user is re-invited with a different role the invitation role should be updated.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1069)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/SFProjectService.cs
+++ b/src/SIL.XForge.Scripture/Services/SFProjectService.cs
@@ -373,6 +373,7 @@ namespace SIL.XForge.Scripture.Services
                 await ProjectSecrets.UpdateAsync(
                     p => p.Id == projectId && p.ShareKeys.Any(sk => sk.Email == email),
                     update => update.Set(p => p.ShareKeys[index].ExpirationTime, expTime)
+                                    .Set(p => p.ShareKeys[index].ProjectRole, role)
                 );
             }
             string key = projectSecret.ShareKeys.Single(sk => sk.Email == email).Key;
--- a/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs
+++ b/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs
@@ -81,7 +81,8 @@ namespace SIL.XForge.Scripture.Services
         {
             var env = new TestEnvironment();
             const string email = "bob@example.com";
-            const string role = SFProjectRole.CommunityChecker;
+            const string initialRole = SFProjectRole.CommunityChecker;
+            const string endingRole = SFProjectRole.Observer;
 
             SFProjectSecret projectSecret = env.ProjectSecrets.Get(Project03);
             Assert.That(projectSecret.ShareKeys.Single(sk => sk.Email == email).ExpirationTime,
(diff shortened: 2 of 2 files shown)


Nateowami (commented):
This is failing because it depends on #1068. Once that gets merged and this is rebased I think the tests will pass.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1069#-:-McGb_-Y9JL0YWjU0_Yt:bit4lku)* status: 0 of 2 files reviewed, all discussions resolved

irahopkinson (changes requested):
FYI, I took a punt that the tests would pass and rebased your branch before reviewing it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1069#-:-McIQNU13GNqKZpciAK3:bboiyr0)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 376 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1069#-McIQ32J67stNS3sv4X_:-McIQ32J67stNS3sv4Xa:bij8mv9) ([raw file](https://github.com/sillsdev/web-xforge/blob/34a2c730a8682473766cc6faa2a6123a8a7c27e1/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L376)):*
> ```
> .Set(p => p.ShareKeys[index].ProjectRole, role)
> ```
This will only apply to new invitations and so the role won't show for any that are already invited? Do we need to migrate existing data?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1069)* status: all files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 376 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1069#-McIQ32J67stNS3sv4X_:-McJskPR2JcFh_o9CM4g:b-l2d1pp) ([raw file](https://github.com/sillsdev/web-xforge/blob/34a2c730a8682473766cc6faa2a6123a8a7c27e1/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L376)):*

> ```
> .Set(p => p.ShareKeys[index].ProjectRole, role)
> ```
This will only apply to new invitations and so the role won't show for any that are already invited? Do we need to migrate existing data?
</blockquote></details>

This role information is what is used when the user joins in order to determine what role the user was invited to. Previously if a user was invited to be a community checker, and then re-invited to be a translation observer (presumably because the original invitation was in error) the expiration time would be updated, but the invitation would still be for a community checker. The role was already being set for new share keys (see about 13 rows up), but it wasn't being updated when a new user was invited.

irahopkinson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1069#-:-McK4HHP8-09l24b2BLC:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 376 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1069#-McIQ32J67stNS3sv4X_:-McK3PAD4lxcycsO2uG4:byw5vb) ([raw file](https://github.com/sillsdev/web-xforge/blob/34a2c730a8682473766cc6faa2a6123a8a7c27e1/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L376)):*

This role information is what is used when the user joins in order to determine what role the user was invited to. Previously if a user was invited to be a community checker, and then re-invited to be a translation observer (presumably because the original invitation was in error) the expiration time would be updated, but the invitation would still be for a community checker. The role was already being set for new share keys (see about 13 rows up), but it wasn't being updated when a new user was invited.
</blockquote></details>

Yep, that makes more sense as I look at the code again. After I reviewed this I checked out the commit before your previous related PR, added an invite as _View Translation_ which doesn't show the role. Then I checked out your branch and it showed the role so it seems no migration of existing projects is needed.

PR #1073, opened 2021-06-21
SF-1219b: Fix remaining memory leaks

pmachapman:
This pull request should help reduce out of memory issues by fixing the following memory leaks:
 - `ScrText` objects are now disposed after they have been created (commit 6d2f1385138df1dc7ad4a8d9341b9ef35d2e4689).
 - The `RealtimeJsonService` now uses `JsonSerializer` instead of `JToken` to reduce memory use and object creation (commit d47f155c646e711aeb259b2ac18c3738aebccebb).
- Added `using` statements to tests that were missing them (commit a040d350fc581d7dc98a5f2414061c50c3eb4d8e)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1073)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -558,7 +558,7 @@ namespace SIL.XForge.Scripture.Services
                 // Dispose the ScrText objects
                 foreach (ScrText scrText in scrTexts.Values)
                 {
-                    scrText.Dispose();
+                    scrText?.Dispose();
                 }
 
                 // Clear the collection to release the references to the ScrTexts for GC
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -195,7 +195,8 @@ namespace SIL.XForge.Scripture.Services
                     ptProject.ShortName, source.AsInternetSharedRepositorySource(), repositories);
                 string username = GetParatextUsername(userSecret);
                 // Specifically set the ScrText property of the SharedProject to indicate the project is available locally
-                sharedProj.ScrText = ScrTextCollection.FindById(username, paratextId);
+                using ScrText scrText = ScrTextCollection.FindById(username, paratextId);
+                sharedProj.ScrText = scrText;
                 sharedProj.Permissions = sharedProj.ScrText.Permissions;
                 List<SharedProject> sharedPtProjectsToSr = new List<SharedProject> { sharedProj };
 
@@ -368,7 +369,7 @@ namespace SIL.XForge.Scripture.Services
             else
             {
                 // Get the scripture text so we can retrieve the permissions from the XML
-                ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), project.ParatextId);
+                using ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), project.ParatextId);
 
                 // Calculate the project and resource permissions
                 foreach (string uid in project.UserRoles.Keys)
@@ -451,14 +452,14 @@ namespace SIL.XForge.Scripture.Services
         /// <summary> Determine if a specific project is in a right to left language. </summary>
         public bool IsProjectLanguageRightToLeft(UserSecret userSecret, string paratextId)
         {
-            ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
+            using ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
             return scrText == null ? false : scrText.RightToLeft;
         }
 
         /// <summary> Get list of book numbers in PT project. </summary>
         public IReadOnlyList<int> GetBookList(UserSecret userSecret, string paratextId)
         {
-            ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
+            using ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
             if (scrText == null)
                 return Array.Empty<int>();
             return scrText.Settings.BooksPresentSet.SelectedBookNumbers.ToArray();
@@ -467,7 +468,7 @@ namespace SIL.XForge.Scripture.Services
         /// <summary> Get PT book text in USX, or throw if can't. </summary>
         public string GetBookText(UserSecret userSecret, string paratextId, int bookNum)
         {
-            ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
+            using ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
             if (scrText == null)
                 throw new DataNotFoundException("Can't get access to cloned project.");
             string usfm = scrText.GetText(bookNum);
@@ -478,77 +479,90 @@ namespace SIL.XForge.Scripture.Services
         public async Task PutBookText(UserSecret userSecret, string projectId, int bookNum, string usx,
             Dictionary<int, string> chapterAuthors = null)
         {
-            string username = GetParatextUsername(userSecret);
-            ScrText scrText = ScrTextCollection.FindById(username, projectId);
-            var doc = new XmlDocument
+            Dictionary<string, ScrText> scrTexts = new Dictionary<string, ScrText>();
+            try
             {
-                PreserveWhitespace = true
-            };
-            doc.LoadXml(usx);
-            UsxFragmenter.FindFragments(scrText.ScrStylesheet(bookNum), doc.CreateNavigator(),
-                XPathExpression.Compile("*[false()]"), out string usfm);
-            usfm = UsfmToken.NormalizeUsfm(scrText.ScrStylesheet(bookNum), usfm, false, scrText.RightToLeft, scrText);
+                string username = GetParatextUsername(userSecret);
+                ScrText scrText = ScrTextCollection.FindById(username, projectId);
 
-            if (chapterAuthors == null || chapterAuthors.Count == 0)
-            {
-                // If we don't have chapter authors, update book as current user
-                if (scrText.Permissions.AmAdministrator)
+                // We add this here so we can dispose in the finally
+                scrTexts.Add(userSecret.Id, scrText);
+                var doc = new XmlDocument
                 {
-                    // if the current user is an administrator, then always allow editing the book text even if the user
-                    // doesn't have permission. This will ensure that a sync by an administrator never fails.
-                    scrText.Permissions.RunWithEditPermision(bookNum,
-                        () => scrText.PutText(bookNum, 0, false, usfm, null));
-                }
-                else
+                    PreserveWhitespace = true
+                };
+                doc.LoadXml(usx);
+                UsxFragmenter.FindFragments(scrText.ScrStylesheet(bookNum), doc.CreateNavigator(),
+                    XPathExpression.Compile("*[false()]"), out string usfm);
+                usfm = UsfmToken.NormalizeUsfm(scrText.ScrStylesheet(bookNum), usfm, false, scrText.RightToLeft, scrText);
+
+                if (chapterAuthors == null || chapterAuthors.Count == 0)
                 {
-                    scrText.PutText(bookNum, 0, false, usfm, null);
-                }
-                _logger.LogInformation("{0} updated {1} in {2}.", userSecret.Id,
-                    Canon.BookNumberToEnglishName(bookNum), scrText.Name);
-            }
-            else
-            {
-                // As we have a list of chapter authors, build a dictionary of ScrTexts for each of them
-                Dictionary<string, ScrText> scrTexts = new Dictionary<string, ScrText>();
-                foreach (string userId in chapterAuthors.Values.Distinct())
-                {
-                    if (userId == userSecret.Id)
+                    // If we don't have chapter authors, update book as current user
+                    if (scrText.Permissions.AmAdministrator)
                     {
-                        scrTexts.Add(userId, scrText);
+                        // if the current user is an administrator, then always allow editing the book text even if the user
+                        // doesn't have permission. This will ensure that a sync by an administrator never fails.
+                        scrText.Permissions.RunWithEditPermision(bookNum,
+                            () => scrText.PutText(bookNum, 0, false, usfm, null));
                     }
                     else
                     {
-                        // Get their user secret, so we can get their username, and create their ScrText
-                        UserSecret authorUserSecret = await _userSecretRepository.GetAsync(userId);
-                        string authorUserName = GetParatextUsername(authorUserSecret);
-                        scrTexts.Add(userId, ScrTextCollection.FindById(authorUserName, projectId));
+                        scrText.PutText(bookNum, 0, false, usfm, null);
                     }
-                }
-
-                // If there is only one author, just write the book
-                if (scrTexts.Count == 1)
-                {
-                    scrTexts.Values.First().PutText(bookNum, 0, false, usfm, null);
-                    _logger.LogInformation("{0} updated {1} in {2}.", scrTexts.Keys.First(),
+                    _logger.LogInformation("{0} updated {1} in {2}.", userSecret.Id,
                         Canon.BookNumberToEnglishName(bookNum), scrText.Name);
                 }
                 else
                 {
-                    // Split the usfm into chapters
-                    List<string> chapters = ScrText.SplitIntoChapters(scrText.Name, bookNum, usfm);
-
-                    // Put the individual chapters
-                    foreach ((int chapterNum, string authorUserId) in chapterAuthors)
+                    // As we have a list of chapter authors, build a dictionary of ScrTexts for each of them
+                    foreach (string userId in chapterAuthors.Values.Distinct())
                     {
-                        if ((chapterNum - 1) < chapters.Count)
+                        if (userId != userSecret.Id)
                         {
-                            // The ScrText permissions will be the same as the last sync's permissions, so no need to check
-                            scrTexts[authorUserId].PutText(bookNum, chapterNum, false, chapters[chapterNum - 1], null);
-                            _logger.LogInformation("{0} updated chapter {1} of {2} in {3}.", authorUserId,
-                                chapterNum, Canon.BookNumberToEnglishName(bookNum), scrText.Name);
+                            // Get their user secret, so we can get their username, and create their ScrText
+                            UserSecret authorUserSecret = await _userSecretRepository.GetAsync(userId);
+                            string authorUserName = GetParatextUsername(authorUserSecret);
+                            scrTexts.Add(userId, ScrTextCollection.FindById(authorUserName, projectId));
+                        }
+                    }
+
+                    // If there is only one author, just write the book
+                    if (scrTexts.Count == 1)
+                    {
+                        scrTexts.Values.First().PutText(bookNum, 0, false, usfm, null);
+                        _logger.LogInformation("{0} updated {1} in {2}.", scrTexts.Keys.First(),
+                            Canon.BookNumberToEnglishName(bookNum), scrText.Name);
+                    }
+                    else
+                    {
+                        // Split the usfm into chapters
+                        List<string> chapters = ScrText.SplitIntoChapters(scrText.Name, bookNum, usfm);
+
+                        // Put the individual chapters
+                        foreach ((int chapterNum, string authorUserId) in chapterAuthors)
+                        {
+                            if ((chapterNum - 1) < chapters.Count)
+                            {
+                                // The ScrText permissions will be the same as the last sync's permissions, so no need to check
+                                scrTexts[authorUserId].PutText(bookNum, chapterNum, false, chapters[chapterNum - 1], null);
+                                _logger.LogInformation("{0} updated chapter {1} of {2} in {3}.", authorUserId,
+                                    chapterNum, Canon.BookNumberToEnglishName(bookNum), scrText.Name);
+                            }
                         }
                     }
                 }
+            }
+            finally
+            {
+                // Dispose the ScrText objects
+                foreach (ScrText scrText in scrTexts.Values)
+                {
+                    scrText.Dispose();
+                }
+
+                // Clear the collection to release the references to the ScrTexts for GC
+                scrTexts.Clear();
             }
         }
 
@@ -556,7 +570,7 @@ namespace SIL.XForge.Scripture.Services
         public string GetNotes(UserSecret userSecret, string projectId, int bookNum)
         {
             // TODO: should return some data structure instead of XML
-            ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), projectId);
+            using ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), projectId);
             if (scrText == null)
                 return null;
 
@@ -573,7 +587,7 @@ namespace SIL.XForge.Scripture.Services
             string username = GetParatextUsername(userSecret);
             List<string> users = new List<string>();
             int nbrAddedComments = 0, nbrDeletedComments = 0, nbrUpdatedComments = 0;
-            ScrText scrText = ScrTextCollection.FindById(username, projectId);
+            using ScrText scrText = ScrTextCollection.FindById(username, projectId);
             if (scrText == null)
                 throw new DataNotFoundException("Can't get access to cloned project.");
             CommentManager manager = CommentManager.Get(scrText);
@@ -636,7 +650,7 @@ namespace SIL.XForge.Scripture.Services
                 return null;
             }
 
-            ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
+            using ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
             if (scrText != null)
             {
                 return _hgHelper.GetLastPublicRevision(scrText.Directory, allowEmptyIfRestoredFromBackup: false);
@@ -664,7 +678,7 @@ namespace SIL.XForge.Scripture.Services
             }
 
             // Get the scripture text
-            ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
+            using ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
 
             // If we do not have a scripture text, do not back up
             if (scrText == null)
@@ -685,7 +699,7 @@ namespace SIL.XForge.Scripture.Services
             }
 
             // Get the scripture text
-            ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
+            using ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
 
             // If we do not have a scripture text, do not back up
             if (scrText == null)
@@ -731,7 +745,7 @@ namespace SIL.XForge.Scripture.Services
             }
 
             // Get the scripture text
-            ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
+            using ScrText scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), paratextId);
 
             // If we do not have a scripture text, do not back up
             if (scrText == null)
@@ -896,8 +910,8 @@ namespace SIL.XForge.Scripture.Services
             IInternetSharedRepositorySource repositorySource)
         {
             string username = GetParatextUsername(userSecret);
-            bool targetNeedsCloned =
-                ScrTextCollection.FindById(username, target.ParatextId) == null;
+            using ScrText scrText = ScrTextCollection.FindById(username, target.ParatextId);
+            bool targetNeedsCloned = scrText == null;
             if (target is ParatextResource resource)
             {
                 // If the target is a resource, install it
--- a/src/SIL.XForge/Realtime/RealtimeJsonService.cs
+++ b/src/SIL.XForge/Realtime/RealtimeJsonService.cs
@@ -3,7 +3,6 @@ using System.Threading;
 using System.Threading.Tasks;
 using Jering.Javascript.NodeJS;
 using Newtonsoft.Json;
-using Newtonsoft.Json.Linq;
 using Newtonsoft.Json.Serialization;
 
 namespace SIL.XForge.Realtime
@@ -22,25 +21,23 @@ namespace SIL.XForge.Realtime
         };
 
         /// <inheritdoc />
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
         public async ValueTask<T> DeserializeAsync<T>(Stream stream, CancellationToken cancellationToken = default)
+#pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
         {
-            using (StreamReader sr = new StreamReader(stream, System.Text.Encoding.Default, true, 1024, true))
-            using (JsonTextReader reader = new JsonTextReader(sr))
-            {
-                JToken json = await JToken.LoadAsync(reader, cancellationToken);
-                return json.ToObject<T>(_serializer);
-            }
+            using StreamReader sr = new StreamReader(stream, System.Text.Encoding.Default, true, 1024, true);
+            using JsonTextReader reader = new JsonTextReader(sr);
+            return _serializer.Deserialize<T>(reader);
         }
 
         /// <inheritdoc />
+#pragma warning disable CS1998 // Async method lacks 'await' operators and will run synchronously
         public async Task SerializeAsync<T>(Stream stream, T value, CancellationToken cancellationToken = default)
+#pragma warning restore CS1998 // Async method lacks 'await' operators and will run synchronously
         {
-            using (StreamWriter sw = new StreamWriter(stream, System.Text.Encoding.Default, 1024, true))
-            using (JsonWriter writer = new JsonTextWriter(sw))
-            {
-                JToken json = JToken.FromObject(value, _serializer);
-                await json.WriteToAsync(writer, cancellationToken);
-            }
+            using StreamWriter sw = new StreamWriter(stream, System.Text.Encoding.Default, 1024, true);
+            using JsonWriter writer = new JsonTextWriter(sw);
+            _serializer.Serialize(writer, value);
         }
     }
 }
--- a/test/SIL.XForge.Scripture.Tests/Services/ComparableProjectPermissionManager.cs
+++ b/test/SIL.XForge.Scripture.Tests/Services/ComparableProjectPermissionManager.cs
@@ -1,0 +1,32 @@
+using System;
+using Paratext.Data;
+using Paratext.Data.Users;
+using PtxUtils;
+
+namespace SIL.XForge.Scripture.Services
+{
+    /// <summary>
+    /// This is an implementation of the ProjectPermissionManager that allows comparison of permission managers.
+    /// </summary>
+    /// <seealso cref="ProjectPermissionManager" />
+    /// <seealso cref="IEquatable{ComparableProjectPermissionManager}" />
+    public class ComparableProjectPermissionManager
+        : ProjectPermissionManager, IEquatable<ComparableProjectPermissionManager>
+    {
+        public ComparableProjectPermissionManager(ScrText scrText)
+            : base(scrText)
+        {
+        }
+
+        /// <summary>
+        /// Gets the XML data.
+        /// </summary>
+        /// <remarks>
+        /// This is based on the code for <see cref="PermissionManager.Clone" />.
+        /// </remarks>
+        public string XmlData => Memento.ToXmlString(Data, false, true);
+
+        public bool Equals(ComparableProjectPermissionManager other)
+            => this.XmlData == other?.XmlData;
+    }
+}
--- a/test/SIL.XForge.Scripture.Tests/Services/ComparableProjectPermissionManager.cs
+++ b/test/SIL.XForge.Scripture.Tests/Services/ComparableProjectPermissionManager.cs
@@ -24,7 +24,7 @@ namespace SIL.XForge.Scripture.Services
         /// <remarks>
         /// This is based on the code for <see cref="PermissionManager.Clone" />.
         /// </remarks>
-        public string XmlData => Memento.ToXmlString(Data, false, true);
+        private string XmlData => Memento.ToXmlString(Data, false, true);
 
         public bool Equals(ComparableProjectPermissionManager other)
             => this.XmlData == other?.XmlData;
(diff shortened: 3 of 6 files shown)


marksvc (changes requested):
I reviewed two of the commits separately, which was nice for isolation in review; thank you. 

Thanks for your work on this. 

> Reviewed 13 of 13 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-:-Mck8DGZ6w2XHrJ9Mmgk:be2c0dz)* status: all files reviewed, 10 unresolved discussions (waiting on @pmachapman)

___
*[src/SIL.XForge/Realtime/RealtimeJsonService.cs, line 40 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-McjU9Xo2blT7fjgARwf:-McjU9Xp5330VM2szr88:b-2137yt) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge/Realtime/RealtimeJsonService.cs#L40)):*
> ```Smalltalk
>             using StreamWriter sw = new StreamWriter(stream, System.Text.Encoding.Default, 1024, true);
>             using JsonWriter writer = new JsonTextWriter(sw);
>             _serializer.Serialize(writer, value);
> ```

Yeah okay. @irahopkinson , is there a reason why (git log says) you  used JToken to read+write? I'm assuming not because it was doing something that .Deserialize and .Serialize weren't handling, and that this way is fine.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 478 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-MckA05J5OhxUTMw2Asm:-MckA05J5OhxUTMw2Asn:b-3pwlzn) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ParatextService.cs#L478)):*
> ```Smalltalk
>         /// <summary> Write up-to-date book text from mongo database to Paratext project folder. </summary>
>         public async Task PutBookText(UserSecret userSecret, string projectId, int bookNum, string usx,
>             IScrTextCollection scrTextCollection, Dictionary<int, string> chapterAuthors = null)
> ```

I don't like that `class ParatextService` is now pushing responsibility for maintaining a ScrTextCollection to the clients of ParatextService. A ScrText and a ScrTextCollection are Paratext classes* and it would be nice to contain them within the ParatextService. 

* Well, IScrTextCollection isn't technically, but it's trying to be like a Paratext class :). 

Hmm. But what to do. Maybe this would be good to talk about in our meeting.
___
*[src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs, line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-Mck3CIJCOOkFK-iBG7s:-Mck3CIJCOOkFK-iBG7t:b-atf1la) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L92)):*
> ```Smalltalk
>         public async Task RunAsync(string projectId, string userId, bool trainEngine)
>         {
>             using ScopedScrTextCollection scrTextCollection = new ScopedScrTextCollection(_fileSystemService);
> ```

It may be that every sync job makes a new ParatextSyncRunner object and then calls RunAsync() on it. Notice that there are _projectDoc and _projectSecret and _userSecret objects, so a given ParatextSyncRunner object is only operating on one project, as requested by one user, at a time. 

Because of this, you could put the scrTextCollection in a private class field rather than pass it around. 

An option if desired.
___
*[src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs, line 166 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-McjY2UR7SaCvvQAuoPF:-McjY2US60DKeG3o9KC-:b-9t4e8l) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L166)):*
> ```Smalltalk
> 
>                 // Create the handler
>                 progress.ProgressUpdated += handler;
> ```

I assume that placing the code above inline here, rather than in the UseNewProgress() method, is helping with GC because when  the handler is removed from ProgressUpdated, and RunAsync() is finished,  the `handler` itself goes out of scope. Is that what's helping GC here?

It's too bad to inline this since ParatextSyncRunner has generally suffered from too-long methods. I wonder if there's a clever way to get  some GC-friendly functionality like this but keep this progress code out in its own method. We could modify UseNewProgress() so it was something like

```cs
void UseNewProgressWhlieRunning(lambda) 
{
  handler = new handler();
  progress += handler;
  try {
    lambda();
  }
  finally {
    progress -= handler;
  }
}
```

and then call it here with

```cs
UseNewProgressWhileRunning(() => {
  // perform Paratext send/receive
  await _paratextService.SendReceiveAsync(_userSecret, targetParatextId, scrTextCollection, progres
});
```

(with appropriate async/await sprinkled in).

This doesn't need to be done as part of the PR at hand. But do you have any thoughts about how to  avoid inlining UseNewProgress(), yet still have the GC benefit?
___
*[src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs, line 172 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-Mck5hhVCeTgM5QZxtix:-Mck5hhVCeTgM5QZxtiy:b-re0fhn) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L172)):*
> ```Smalltalk
> 
>                 // Deregister the handler
>                 progress.ProgressUpdated -= handler;
> ```

Come to think of it, this isn't protected by a `finally`. And so when we  get a crash in SR, we will still be keeping these in memory. What do you think about doing the `-= handler`  down in `CompleteSync()`?

I suppose that would mean making at least `progress` into a field, if not also `handler` (tho we can probably iterate thru progress.ProgressUpdated and `-=` all its contents). 

I see that SyncProgress is ours. Maybe SyncProgress could be made to implement IDisposable, on dispose it could dump its ProgressUpdated handlers, and in ParatextSyncRunner we could `using progress = new SyncProgress();`. Then we wouldn't need to add `progress` or `handler` as class fields.

So, class SyncProgress could maybe have

```cs
public class SyncProgress ... 
{
  private _listOfListenersIfOtherwiseAPainToFigureOut = new List();
  public AddProgressUpdatedListener(listener) 
  {
    _listOfListeners += listener;
    ProgressUpdated += listener;
  }
  public override ManagedDispose()
  {
    ...
    for each listener in _listOfListeners
    {
      ProgressUpdated -= listener;
    }
  }
}
```
___
*[src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs, line 15 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-McjiAQM-ydD3yzVgwJm:-McjiAQN8U0qtrfuUdYk:b-co8nai) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs#L15)):*
> ```
>  using a simple cache.
> ```

Irony. So it might be good to add some more info to maintain a bit of knowledge about what is going on here. 

IIRC: 

The LazyScrTextCollection allowed us to specify our own FindById() method that would fetch a ScrText _for a specific user_, to help us in the multi-user environment. Probably the Paratext ScrTextCollection.FindById() was fetching a ScrText, caching it, and then different users in SF who tried to fetch a ScrText for the same project would then get the same ScrText but now they are using a ScrText that doesn't have the right understanding of _who_ is using it. So writing, as the old class comment said, "A class that can be used to get a ScrText without using a cache" had let us provide ScrText objects that were for a specific user, as they weren't grabbed from a cached that had made them with respect to the wrong user.  (A brief glance at ScrText.cs makes me wonder if that's the right explanation tho :). 
The _cache_ isn't so much the problem, as trouble when multi-users use the same cached objects. 

Let's record some of this into the class comment. Perhaps we could use a class comment like the following:

> A class providing a multi-user-capable replacement for some of ParatextData.ScrTextCollection, specifically FindById(), which does not allow specifying which user to associate with the ScrText and is not for use in a multi-user system. 
> This class provides a cache and should be disposed of when the <see cref="ScrText"/> objects retrieved by it are no longer in use. The ScrText objects are cached by the scope of particular users.
> This class is not related by type to ScrTextCollection. 

I tried to include in the above what it means that the class is "Scoped". It's scoped in that the ScrText objects are with respect to a particular user.

It would be interesting, but perhaps not significantly helpful, to pass in a Paratext Username as a constructor parameter, and then FindById() could omit the Paratext Username parameter and only fetch (and cache) things that pertained to the particular user. An idea that occurred to me. 
___
*[src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs, line 32 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-Mcjo2oX5AMY4Ztdtnbo:-Mcjo2oX5AMY4Ztdtnbp:ba642hc) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs#L32)):*
> ```
>  the project id and Paratext user id 
> ```

Could you specify in the code comment if this is "the ScriptureForge project id" or "the Paratext project id"? I find that we sometimes have that ambiguity.
___
*[src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs, line 53 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-Mck1xz_9Ipxnp9JL-Z4:-Mck1xza6LJxCQ9EcTT0:b6fdcrg) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs#L53)):*
> ```Smalltalk
>         /// </summary>
>         /// <param name="ptUsername"> The username of the user retrieving the ScrText. </param>
>         /// <param name="projectId"> The ID of the target project. </param>
> ```

This may be obvious now, but can we specify that it's the _Paratext_ project ID in the comment? 
___
*[src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs, line 94 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-McjljEQAfO0vc0eZT54:-McjljEQAfO0vc0eZT55:b8g192f) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs#L94)):*
> ```
> TryAdd
> ```

Why TryAdd here instead of just add? If the value is already in the dictionary, then the `TryGetValue` up above should have succeeded and we wouldn't get here, unless our mutex wasn't working, right? What do you think about using just Add here, and letting it crash to alert of an unexpected mutex or logic problem rather than silently not-add when it's already in the dictionary? Or maybe there is more to it than I'm seeing.
___
*[src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs, line 120 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-McjmkXd7enjLw6Q-XwW:-McjmkXd7enjLw6Q-XwX:b887ef) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs#L120)):*
> ```Smalltalk
>                     if (_scrTextsByProjectIdAndUserId.TryGetValue(key, out ScrText scrText))
>                     {
>                         scrText.Dispose();
> ```

I think this is a possible NullReferenceException, right? 
Hmm. Maybe here we should do something like 
```cs
if (scrText!=null) 
{
  scrText.Dispose();
}
```
I don't see that we would have this situation. But we don't need to crash when disposing. This goes against my earlier comment that pertains to not silently hiding a problem :)
Letting it crash here would be more consistent with not silently hiding a problem. Maybe that's a good idea to not check if scrText is null before calling .Dispose() on it. Do you have an opinion?

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1073)* status: 11 of 13 files reviewed, 10 unresolved discussions (waiting on @marksvc and @pmachapman)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 478 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-MckA05J5OhxUTMw2Asm:-MckQdPiDthjq2tq1SA2:bble3ee) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ParatextService.cs#L478)):*

I don't like that `class ParatextService` is now pushing responsibility for maintaining a ScrTextCollection to the clients of ParatextService. A ScrText and a ScrTextCollection are Paratext classes* and it would be nice to contain them within the ParatextService. 

* Well, IScrTextCollection isn't technically, but it's trying to be like a Paratext class :). 

Hmm. But what to do. Maybe this would be good to talk about in our meeting.
</blockquote></details>

Yes, totally. The other option is to create/dispose the `ScrText` objects in each method that requires them. I tried this first off the bat, and it worked...mostly...some methods passed on parts of the object that were lost when the dispose was hit, but I'm sure I can sort this out properly if that is the way we want to go.

Using the `ScopedScrTextCollection` and effectively passing around the same `ScrText` object only saved a couple of seconds for the entire execution time.
___
*[src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs, line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-Mck3CIJCOOkFK-iBG7s:-MckRE7G8aF1cbqIhb2F:bpd4xkt) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L92)):*

It may be that every sync job makes a new ParatextSyncRunner object and then calls RunAsync() on it. Notice that there are _projectDoc and _projectSecret and _userSecret objects, so a given ParatextSyncRunner object is only operating on one project, as requested by one user, at a time. 

Because of this, you could put the scrTextCollection in a private class field rather than pass it around. 

An option if desired.


</blockquote></details>

Yes, that is more logical, especially since the permissions code moved to `SFProjectService` creates its own `ScopedScrTextCollection` there. If we choose to keep the `ScopedScrTextCollection`, I'll do that.
___
*[src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs, line 166 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-McjY2UR7SaCvvQAuoPF:-MckRWr7Djbzl_TbauD1:bk543e) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L166)):*

I assume that placing the code above inline here, rather than in the UseNewProgress() method, is helping with GC because when  the handler is removed from ProgressUpdated, and RunAsync() is finished,  the `handler` itself goes out of scope. Is that what's helping GC here?

It's too bad to inline this since ParatextSyncRunner has generally suffered from too-long methods. I wonder if there's a clever way to get  some GC-friendly functionality like this but keep this progress code out in its own method. We could modify UseNewProgress() so it was something like

```cs
void UseNewProgressWhlieRunning(lambda) 
{
  handler = new handler();
  progress += handler;
  try {
    lambda();
  }
  finally {
    progress -= handler;
  }
}
```

and then call it here with

```cs
UseNewProgressWhileRunning(() => {
  // perform Paratext send/receive
  await _paratextService.SendReceiveAsync(_userSecret, targetParatextId, scrTextCollection, progres
});
```

(with appropriate async/await sprinkled in).

This doesn't need to be done as part of the PR at hand. But do you have any thoughts about how to  avoid inlining UseNewProgress(), yet still have the GC benefit?

</blockquote></details>

I've cleaned it up a bit, but haven't gone with the IDisposable pattern you mentioned below, as everything stayed in memory (even with appropriate dereferencing) when I implemented it.
___
*[src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs, line 172 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-Mck5hhVCeTgM5QZxtix:-MckRk-I1JRMYhDxFVIl:b30pqgs) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L172)):*

Come to think of it, this isn't protected by a `finally`. And so when we  get a crash in SR, we will still be keeping these in memory. What do you think about doing the `-= handler`  down in `CompleteSync()`?

I suppose that would mean making at least `progress` into a field, if not also `handler` (tho we can probably iterate thru progress.ProgressUpdated and `-=` all its contents). 

I see that SyncProgress is ours. Maybe SyncProgress could be made to implement IDisposable, on dispose it could dump its ProgressUpdated handlers, and in ParatextSyncRunner we could `using progress = new SyncProgress();`. Then we wouldn't need to add `progress` or `handler` as class fields.

So, class SyncProgress could maybe have

```cs
public class SyncProgress ... 
{
  private _listOfListenersIfOtherwiseAPainToFigureOut = new List();
  public AddProgressUpdatedListener(listener) 
  {
    _listOfListeners += listener;
    ProgressUpdated += listener;
  }
  public override ManagedDispose()
  {
    ...
    for each listener in _listOfListeners
    {
      ProgressUpdated -= listener;
    }
  }
}
```
</blockquote></details>

Yes, I remembered last night while try to sleep that I forgot the finally... :-)

The `IDisposable` pattern still kept all of the objects in memory, so I have just gone with the try/finally pattern, but cleaned it up a bit.
___
*[src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs, line 15 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-McjiAQM-ydD3yzVgwJm:-MckUvnT02wPzDRcbV9t:b-rofdzz) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs#L15)):*

> ```
>  using a simple cache.
> ```

Irony. So it might be good to add some more info to maintain a bit of knowledge about what is going on here. 

IIRC: 

The LazyScrTextCollection allowed us to specify our own FindById() method that would fetch a ScrText _for a specific user_, to help us in the multi-user environment. Probably the Paratext ScrTextCollection.FindById() was fetching a ScrText, caching it, and then different users in SF who tried to fetch a ScrText for the same project would then get the same ScrText but now they are using a ScrText that doesn't have the right understanding of _who_ is using it. So writing, as the old class comment said, "A class that can be used to get a ScrText without using a cache" had let us provide ScrText objects that were for a specific user, as they weren't grabbed from a cached that had made them with respect to the wrong user.  (A brief glance at ScrText.cs makes me wonder if that's the right explanation tho :). 
The _cache_ isn't so much the problem, as trouble when multi-users use the same cached objects. 

Let's record some of this into the class comment. Perhaps we could use a class comment like the following:

> A class providing a multi-user-capable replacement for some of ParatextData.ScrTextCollection, specifically FindById(), which does not allow specifying which user to associate with the ScrText and is not for use in a multi-user system. 
> This class provides a cache and should be disposed of when the <see cref="ScrText"/> objects retrieved by it are no longer in use. The ScrText objects are cached by the scope of particular users.
> This class is not related by type to ScrTextCollection. 

I tried to include in the above what it means that the class is "Scoped". It's scoped in that the ScrText objects are with respect to a particular user.

It would be interesting, but perhaps not significantly helpful, to pass in a Paratext Username as a constructor parameter, and then FindById() could omit the Paratext Username parameter and only fetch (and cache) things that pertained to the particular user. An idea that occurred to me. 


</blockquote></details>

I couldn't think of a good name for this class. It is meant to mean that the `ScrText` objects are scoped to the lifetime of this class.

I didn't do this per user, as the code for pushing chapters to Paratext will `sudo` as another user if required, hence why the collection has the user id.

It is meant to avoid the issues that Paratext's ScrTextCellection has by instead of scoping to app lifetime via a static, we can scope them to a service action (i.e. syncing or updating permissions).

I'm not really happy with how I did this from a architecture point of view, but it does stop any leaks. Something to discuss at the meeting?

I've updated the comments anyway.
___
*[src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs, line 32 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-Mcjo2oX5AMY4Ztdtnbo:-MckXr6q3xFtjSKl1tUy:bt2vr1b) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs#L32)):*

> ```
>  the project id and Paratext user id 
> ```

Could you specify in the code comment if this is "the ScriptureForge project id" or "the Paratext project id"? I find that we sometimes have that ambiguity.
</blockquote></details>

Fixed.
___
*[src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs, line 53 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-Mck1xz_9Ipxnp9JL-Z4:-MckXsg32hkllnPHP8an:bt2vr1b) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs#L53)):*

This may be obvious now, but can we specify that it's the _Paratext_ project ID in the comment? 
</blockquote></details>

Fixed.
___
*[src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs, line 94 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-McjljEQAfO0vc0eZT54:-MckXwV8DCpo8UbrZ_hS:bbfeav0) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs#L94)):*

> ```
> TryAdd
> ```

Why TryAdd here instead of just add? If the value is already in the dictionary, then the `TryGetValue` up above should have succeeded and we wouldn't get here, unless our mutex wasn't working, right? What do you think about using just Add here, and letting it crash to alert of an unexpected mutex or logic problem rather than silently not-add when it's already in the dictionary? Or maybe there is more to it than I'm seeing.


</blockquote></details>

Sorry, a case of muscle memory...fixed.
___
*[src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs, line 120 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-McjmkXd7enjLw6Q-XwW:-MckYF8JCmSxC6GdQSS9:b1l6egv) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs#L120)):*

I think this is a possible NullReferenceException, right? 
Hmm. Maybe here we should do something like 
```cs
if (scrText!=null) 
{
  scrText.Dispose();
}
```
I don't see that we would have this situation. But we don't need to crash when disposing. This goes against my earlier comment that pertains to not silently hiding a problem :)
Letting it crash here would be more consistent with not silently hiding a problem. Maybe that's a good idea to not check if scrText is null before calling .Dispose() on it. Do you have an opinion?
</blockquote></details>

I've added a "?" to ignore a null reference error if it occurs.

marksvc (commented):
Thanks for your work on this. So basically I approve, except for some API reservations regarding ParatextService. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-:-McklrJ04jhD_87Hi2hY:bo949l6)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 478 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-MckA05J5OhxUTMw2Asm:-Mcklm7p4KVU_vOTnecR:b-xiegov) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ParatextService.cs#L478)):*

Yes, totally. The other option is to create/dispose the `ScrText` objects in each method that requires them. I tried this first off the bat, and it worked...mostly...some methods passed on parts of the object that were lost when the dispose was hit, but I'm sure I can sort this out properly if that is the way we want to go.

Using the `ScopedScrTextCollection` and effectively passing around the same `ScrText` object only saved a couple of seconds for the entire execution time.
</blockquote></details>

Hm, ok. Yeah let's talk more about it. 
___
*[src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs, line 15 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-McjiAQM-ydD3yzVgwJm:-MckhunZ-CTQOmnMq9W5:b-mkphzh) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ScopedScrTextCollection.cs#L15)):*
> It is meant to mean that the ScrText objects are scoped to the lifetime of this class.
> ...
> It is meant to avoid the issues that Paratext's ScrTextCellection has by instead of scoping to app lifetime via a static, we can scope them to a service action (i.e. syncing or updating permissions).

Ahh, right. 
Hmm. "ShortLivedScrTextCollection"? "EphemeralScrTextCollection"?
I don't think the class name actually needs to reveal that it's not expected to hang around for the life of the application run*. If we just used the equivalent of calling it "ScrTextCollection", then it's a collection of ScrTexts, and we construct it, use it, and maybe destroy it, in which case another ScrTextCollection that is later constructed isn't going to have all of the ScrText objects in it. 

`*` Although I guess if ScrTextCollection is the name of something in ParatextData that _is_ expected to hang around for the life of the application, it could be confusing if our similarly named one isn't. 


> I didn't do this per user, as the code for pushing chapters to Paratext will sudo as another user if required, hence why the collection has the user id.

So, the dictionary is keyed in part by user ID. And when we construct a ScrText we specify the user to associate it with, to ParatextData. I'm not sure I understand what you mean.



The class comment looks good.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1073)* status: 1 of 15 files reviewed, 2 unresolved discussions (waiting on @marksvc and @pmachapman)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 478 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-MckA05J5OhxUTMw2Asm:-Md59nkQ7kQP-vFXyfO0:b-y204fx) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ParatextService.cs#L478)):*

Hm, ok. Yeah let's talk more about it. 
</blockquote></details>

Making the `ParatextService` a scoped service failed quite badly due to how initialisation and how it interacts with the file system. I couldn't get it working without a really drastic and inefficient refactor.

So, I have wrapped the `ScrText`'s in using statements, and updated the unit tests to handle this. As the `ScrText` is not passed outside of the `ParatextService` (with the exception of one minor helper service), it works well. Although it doesn't have the object reuse of `ScopedScrTextCollection`, it is no more inefficient that the current implementation.

marksvc (changes requested):
I am posting my comments so far. 

> Reviewed 1 of 14 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-:-MdOzaYS0leAoXNjDvSg:b-ftbv3h)* status: 2 of 15 files reviewed, 3 unresolved discussions (waiting on @marksvc and @pmachapman)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 478 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-MckA05J5OhxUTMw2Asm:-MdOvtES9HRYve7yF5MW:b-9hpahz) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ParatextService.cs#L478)):*
And I see that previously we observed that 

> The other option is to create/dispose the ScrText objects in each method that requires them. I tried this first off the bat, and it worked...mostly...some methods passed on parts of the object that were lost when the dispose was hit, but I'm sure I can sort this out properly if that is the way we want to go.

Is this something we need to look into more?
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 489 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-MdOwwZx2oMh3hb67Vj7:-MdOwwZx2oMh3hb67Vj8:bd79m9q) ([raw file](https://github.com/sillsdev/web-xforge/blob/6d2f1385138df1dc7ad4a8d9341b9ef35d2e4689/src/SIL.XForge.Scripture/Services/ParatextService.cs#L489)):*
> ```Smalltalk
> 
>                 // We add this here so we can dispose in the finally
>                 scrTexts.Add(userSecret.Id, scrText);
> ```

So, this makes _sure_ that the scrText is in the dictionary, for us to dispose of it later, right. 
userSecret.Id might not exist in chapterAuthors, I would think, and so we might not do this Add() with the current/older code.
It looks like later lines in the method then consider cases of whether scrTexts.Count == 1 or not, which this would influence. 
I need to send my comments-in-progress now without more time to think more about what's happening here or if that's a problem.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1073)* status: 1 of 15 files reviewed, 3 unresolved discussions (waiting on @marksvc and @pmachapman)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 478 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-MckA05J5OhxUTMw2Asm:-MdP-FNr3O3_V6oP5AhP:bxnrnhl) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge.Scripture/Services/ParatextService.cs#L478)):*

And I see that previously we observed that 

> The other option is to create/dispose the ScrText objects in each method that requires them. I tried this first off the bat, and it worked...mostly...some methods passed on parts of the object that were lost when the dispose was hit, but I'm sure I can sort this out properly if that is the way we want to go.

Is this something we need to look into more?
</blockquote></details>

No - I dug into it, and my problems were solely with the unit tests - I have sorted out the unit tests to function correctly.

I fixed the broken tests by keeping the FileManager in the scope of the test, and implementing the ComparableProjectPermissionManager to provide the permission object comparison (rather than the simple object reference comparison that was used previously).
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 489 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-MdOwwZx2oMh3hb67Vj7:-MdP1rpVDsTcJ3EJhGEC:b-qoxqcr) ([raw file](https://github.com/sillsdev/web-xforge/blob/6d2f1385138df1dc7ad4a8d9341b9ef35d2e4689/src/SIL.XForge.Scripture/Services/ParatextService.cs#L489)):*

So, this makes _sure_ that the scrText is in the dictionary, for us to dispose of it later, right. 
userSecret.Id might not exist in chapterAuthors, I would think, and so we might not do this Add() with the current/older code.
It looks like later lines in the method then consider cases of whether scrTexts.Count == 1 or not, which this would influence. 
I need to send my comments-in-progress now without more time to think more about what's happening here or if that's a problem.
</blockquote></details>

I sought to minimize the number of changes to the existing logic as far as possible - the line right after - `UsxFragmenter.FindFragments(scrText.ScrStylesheet(bookNum)` - will cause a crash (which my sync rollback code will sort out), if the ScrText was null, so I don't think it is an issue currently.

That said....I've just added a null check in the finally, as no one likes exceptions in finally blocks.

I haven't added null checks when adding the ScrText objects to the collection as these should not be null, and if they are something is really wrong, so a crash (which the sync-rollback will sort out) is appropriate, and this is what would currently happen in this scenario.

Let me know if you are happy with this, or if I should be a bit heavier with my refactor and include more null checks.

marksvc (approved):
Thanks for your work on this. I approve. :ship: 

I compiled the code and smoke tested connecting a project, looking at text docs, looking at Settings, and Synchronizing a project.

> Reviewed 12 of 14 files at r3, 2 of 2 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-:-Mdc6OoWCxvMxF4BfYoU:b-wa9tk6)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 489 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-MdOwwZx2oMh3hb67Vj7:-Mdbw1SxACRYUhb4ZunD:b-64x7gj) ([raw file](https://github.com/sillsdev/web-xforge/blob/6d2f1385138df1dc7ad4a8d9341b9ef35d2e4689/src/SIL.XForge.Scripture/Services/ParatextService.cs#L489)):*

I sought to minimize the number of changes to the existing logic as far as possible - the line right after - `UsxFragmenter.FindFragments(scrText.ScrStylesheet(bookNum)` - will cause a crash (which my sync rollback code will sort out), if the ScrText was null, so I don't think it is an issue currently.

That said....I've just added a null check in the finally, as no one likes exceptions in finally blocks.

I haven't added null checks when adding the ScrText objects to the collection as these should not be null, and if they are something is really wrong, so a crash (which the sync-rollback will sort out) is appropriate, and this is what would currently happen in this scenario.

Let me know if you are happy with this, or if I should be a bit heavier with my refactor and include more null checks.
</blockquote></details>

Sorry; my concern isn't about scrText being null. Rather, in the old code, we conditionally add scrText to scrTexts, below with

```cs
                Dictionary<string, ScrText> scrTexts = new Dictionary<string, ScrText>();
                foreach (string userId in chapterAuthors.Values.Distinct())
                {
                    if (userId == userSecret.Id)
                    {
                        scrTexts.Add(userId, scrText);
                    }
                    else
                    {
```

But now we _unconditionally_ add scrText to scrTexts in the line above:

```cs
scrTexts.Add(userSecret.Id, scrText);
```

It looks like this would only have a change of how the execution flowed if...
- `chapterAuthors` is not null.
- userSecret.Id is not in chapterAuthors.Values.

And so the scrTexts.Count would be 2 now instead of 1 in that case. And it looks like the only impact is that now instead of writing out the whole book text in that case with 

```cs
                    if (scrTexts.Count == 1)
                    {
                        scrTexts.Values.First().PutText(bookNum, 0, false, usfm, null);
```

we would instead write out each chapter individually, as the single author, with

```cs
                        // Put the individual chapters
                        foreach ((int chapterNum, string authorUserId) in chapterAuthors)
                        {
                            if ((chapterNum - 1) < chapters.Count)
                            {
                                // The ScrText permissions will be the same as the last sync's permissions, so no need to check
                                scrTexts[authorUserId].PutText(bookNum, chapterNum, false, chapters[chapterNum - 1], null);
```

So I guess that's not actually a problem. 

Ok.
___
*[test/SIL.XForge.Scripture.Tests/Services/ComparableProjectPermissionManager.cs, line 27 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-Mdc3Rn_9j1ZDNwyArYw:-Mdc3Rn_9j1ZDNwyArYx:b-2dwrwh) ([raw file](https://github.com/sillsdev/web-xforge/blob/0caa1c8dcd2dff7122b4f8a822bf8d71369520a4/test/SIL.XForge.Scripture.Tests/Services/ComparableProjectPermissionManager.cs#L27)):*
> ```Smalltalk
>         /// This is based on the code for <see cref="PermissionManager.Clone" />.
>         /// </remarks>
>         public string XmlData => Memento.ToXmlString(Data, false, true);
> ```

Nice. Would it be good for this to be `private` , since it's a helper to facilitate Equals(), rather than something we expect clients to call?

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1073)* status: 14 of 15 files reviewed, 2 unresolved discussions (waiting on @marksvc and @pmachapman)

___
*[test/SIL.XForge.Scripture.Tests/Services/ComparableProjectPermissionManager.cs, line 27 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-Mdc3Rn_9j1ZDNwyArYw:-Mdd5Dho586GrYfIYP-I:b-fc35fl) ([raw file](https://github.com/sillsdev/web-xforge/blob/0caa1c8dcd2dff7122b4f8a822bf8d71369520a4/test/SIL.XForge.Scripture.Tests/Services/ComparableProjectPermissionManager.cs#L27)):*

Nice. Would it be good for this to be `private` , since it's a helper to facilitate Equals(), rather than something we expect clients to call?
</blockquote></details>

Have made it private - apologies for that slipping through - it was a hangover from a previous attempted solution where I used a separate IComparer.

marksvc (approved):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1073)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[test/SIL.XForge.Scripture.Tests/Services/ComparableProjectPermissionManager.cs, line 27 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-Mdc3Rn_9j1ZDNwyArYw:-MddImGB5dxAIe8H9FUq:b-s1agie) ([raw file](https://github.com/sillsdev/web-xforge/blob/0caa1c8dcd2dff7122b4f8a822bf8d71369520a4/test/SIL.XForge.Scripture.Tests/Services/ComparableProjectPermissionManager.cs#L27)):*

Have made it private - apologies for that slipping through - it was a hangover from a previous attempted solution where I used a separate IComparer.
</blockquote></details>

No worries :)

irahopkinson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1073)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[src/SIL.XForge/Realtime/RealtimeJsonService.cs, line 40 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1073#-McjU9Xo2blT7fjgARwf:-Mdz5yIFDAOfMziT7UBS:bomt8cv) ([raw file](https://github.com/sillsdev/web-xforge/blob/006e4b31061979f8ccf386a4eb747cf6c759915d/src/SIL.XForge/Realtime/RealtimeJsonService.cs#L40)):*

Yeah okay. @irahopkinson , is there a reason why (git log says) you  used JToken to read+write? I'm assuming not because it was doing something that .Deserialize and .Serialize weren't handling, and that this way is fine.
</blockquote></details>

No reason that I recall.

PR #1074, opened 2021-06-21
SF-1033 localize PT login

irahopkinson:
- pass language through Auth0 so PT login is localized
- stop using Auth0 language picker
- also fixes using the locale from an invite link

This is a draft waiting on a dependent change. Please first review https://github.com/sillsdev/auth0-configs/pull/100. 

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1074)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/package-lock.json
+++ b/src/SIL.XForge.Scripture/ClientApp/package-lock.json
@@ -30,7 +30,7 @@
         "angular-file": "^3.4.0",
         "angular-split": "^4.0.0",
         "arraydiff": "^0.1.3",
-        "auth0-js": "^9.14.3",
+        "auth0-js": "^9.16.2",
         "bootstrap": "^4.6.0",
         "bowser": "^2.11.0",
         "bson-objectid": "^1.3.1",
@@ -11593,12 +11593,13 @@
       }
     },
     "node_modules/auth0-js": {
-      "version": "9.14.3",
-      "resolved": "https://registry.npmjs.org/auth0-js/-/auth0-js-9.14.3.tgz",
-      "integrity": "sha512-UO/fGv9641PUpYjz2nkPaUHzzrhNaJKupJOqt8blj1pD6wBgpZtxUSXBox6Y8md3eTBzpxeWxV+6RKzzERvr1g==",
+      "version": "9.16.2",
+      "resolved": "https://registry.npmjs.org/auth0-js/-/auth0-js-9.16.2.tgz",
+      "integrity": "sha512-cF1nRjmMDezmhJ+ZwwYp23F0gPqU0zNmF/VvTpcwvCrEMl9lAvkCd4iburN1I7G8SYaaIYEfcGedCphpDZw6OQ==",
+      "license": "MIT",
       "dependencies": {
         "base64-js": "^1.3.0",
-        "idtoken-verifier": "^2.0.3",
+        "idtoken-verifier": "^2.1.2",
         "js-cookie": "^2.2.0",
         "qs": "^6.7.0",
         "superagent": "^5.3.1",
@@ -16235,12 +16236,12 @@
       }
     },
     "node_modules/idtoken-verifier": {
-      "version": "2.1.0",
-      "resolved": "https://registry.npmjs.org/idtoken-verifier/-/idtoken-verifier-2.1.0.tgz",
-      "integrity": "sha512-X0423UM4Rc5bFb39Ai0YHr35rcexlu4oakKdYzSGZxtoPy84P86hhAbzlpgbgomcLOFRgzgKRvhY7YjO5g8OPA==",
+      "version": "2.1.2",
+      "resolved": "https://registry.npmjs.org/idtoken-verifier/-/idtoken-verifier-2.1.2.tgz",
+      "integrity": "sha512-YMHiP9zAMjB+pWreV4EHnIj3XCQ168+InWirVRFeRtlsMQIK61S+LLnyLGI8EL0wtlk/v7ya69Gjfio3P9/7Gw==",
       "dependencies": {
         "base64-js": "^1.3.0",
-        "crypto-js": "^3.2.1",
+        "crypto-js": "3.3.0",
         "es6-promise": "^4.2.8",
         "jsbn": "^1.1.0",
         "unfetch": "^4.1.0",
@@ -34112,12 +34113,12 @@
       "dev": true
     },
     "auth0-js": {
-      "version": "9.14.3",
-      "resolved": "https://registry.npmjs.org/auth0-js/-/auth0-js-9.14.3.tgz",
-      "integrity": "sha512-UO/fGv9641PUpYjz2nkPaUHzzrhNaJKupJOqt8blj1pD6wBgpZtxUSXBox6Y8md3eTBzpxeWxV+6RKzzERvr1g==",
+      "version": "9.16.2",
+      "resolved": "https://registry.npmjs.org/auth0-js/-/auth0-js-9.16.2.tgz",
+      "integrity": "sha512-cF1nRjmMDezmhJ+ZwwYp23F0gPqU0zNmF/VvTpcwvCrEMl9lAvkCd4iburN1I7G8SYaaIYEfcGedCphpDZw6OQ==",
       "requires": {
         "base64-js": "^1.3.0",
-        "idtoken-verifier": "^2.0.3",
+        "idtoken-verifier": "^2.1.2",
         "js-cookie": "^2.2.0",
         "qs": "^6.7.0",
         "superagent": "^5.3.1",
@@ -38023,12 +38024,12 @@
       }
     },
     "idtoken-verifier": {
-      "version": "2.1.0",
-      "resolved": "https://registry.npmjs.org/idtoken-verifier/-/idtoken-verifier-2.1.0.tgz",
-      "integrity": "sha512-X0423UM4Rc5bFb39Ai0YHr35rcexlu4oakKdYzSGZxtoPy84P86hhAbzlpgbgomcLOFRgzgKRvhY7YjO5g8OPA==",
+      "version": "2.1.2",
+      "resolved": "https://registry.npmjs.org/idtoken-verifier/-/idtoken-verifier-2.1.2.tgz",
+      "integrity": "sha512-YMHiP9zAMjB+pWreV4EHnIj3XCQ168+InWirVRFeRtlsMQIK61S+LLnyLGI8EL0wtlk/v7ya69Gjfio3P9/7Gw==",
       "requires": {
         "base64-js": "^1.3.0",
-        "crypto-js": "^3.2.1",
+        "crypto-js": "3.3.0",
         "es6-promise": "^4.2.8",
         "jsbn": "^1.1.0",
         "unfetch": "^4.1.0",
--- a/src/SIL.XForge.Scripture/ClientApp/package.json
+++ b/src/SIL.XForge.Scripture/ClientApp/package.json
@@ -44,7 +44,7 @@
     "angular-file": "^3.4.0",
     "angular-split": "^4.0.0",
     "arraydiff": "^0.1.3",
-    "auth0-js": "^9.14.3",
+    "auth0-js": "^9.16.2",
     "bootstrap": "^4.6.0",
     "bowser": "^2.11.0",
     "bson-objectid": "^1.3.1",
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.spec.ts
@@ -1,0 +1,112 @@
+import { ActivatedRouteSnapshot, RouterStateSnapshot } from '@angular/router';
+import { anything, instance, mock, resetCalls, verify, when } from 'ts-mockito';
+import { AuthGuard } from './auth.guard';
+import { AuthService } from './auth.service';
+import { LocationService } from './location.service';
+
+const mockedAuthService = mock(AuthService);
+const mockedLocationService = mock(LocationService);
+const mockedActivatedRouteSnapshot = mock(ActivatedRouteSnapshot);
+const mockedRouterStateSnapshot = mock(RouterStateSnapshot);
+
+describe('AuthGuard', () => {
+  beforeEach(() => {
+    resetCalls(mockedAuthService);
+  });
+
+  it('should do nothing if already logged in', (done: DoneFn) => {
+    // url test pattern: https://scriptureforge.org/projects/<projectId>
+    const authGuard = new AuthGuard(instance(mockedAuthService), instance(mockedLocationService));
+    expect(authGuard).toBeDefined();
+    when(mockedAuthService.isLoggedIn).thenResolve(true);
+    when(mockedActivatedRouteSnapshot.queryParams).thenReturn({});
+
+    const canActivate$ = authGuard.canActivate(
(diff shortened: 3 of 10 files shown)


marksvc (changes requested):
Someone who is familiar with the corresponding auth0-configs change should also give a review. 
Thanks for fixing these things.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1074#-:-MdOsYja5jHHUA7TsV_S:b2a1vbd)* status: all files reviewed, 3 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.ts, line 19 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1074#-MdOTNygCfMOziffl01s:-MdOTNygCfMOziffl01t:b-nvz3bv) ([raw file](https://github.com/sillsdev/web-xforge/blob/f881f68650188d359a714230082eb531ff384e93/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.ts#L19)):*
> ```XML
>         if (!isLoggedIn) {
>           const signUp = next.queryParams['sharing'] === 'true' || next.queryParams['sign-up'] === 'true';
>           const locale = next.queryParams['locale'];
> ```

I'm having trouble finding the tests for AuthGuard as well. 

Let's add to the AuthGuard tests, a specification for what is passed to authService.logIn() if there is no locale query parameter / absent, if it's empty (like in https://example.com/?abc=def&locale=&ghi=jkl, or if it's present, like `mn-OP`.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 168 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1074#-MdOPOxH-cA_PQaQzhL3:-MdOPOxH-cA_PQaQzhL4:b-tgsvsq) ([raw file](https://github.com/sillsdev/web-xforge/blob/f881f68650188d359a714230082eb531ff384e93/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L168)):*
> ```
>     const language = getAspCultureCookieLanguage(this.cookieService.get(ASP_CULTURE_COOKIE_NAME));
>     const ui_locales = language;
> ```

`ui_locales = language` is a bit confusing. It would at least be lessso if we add types to these. I _think_ these are both just strings, but a plural `ui_locales` sounds like it could be an array or something.

(`const ui_locales: string = ...`)
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 173 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1074#-MdOQr2I7esrHB3pPRx7:-MdOQr2I7esrHB3pPRx8:bci8eiz) ([raw file](https://github.com/sillsdev/web-xforge/blob/f881f68650188d359a714230082eb531ff384e93/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L173)):*
> ```XML
>       authOptions.mode = 'signUp';
>       authOptions.login_hint = locale ?? ui_locales;
>     }
> ```

I'm having trouble finding the unit tests for this class. Or, maybe we don't unit tests for our authorization service?

Some tests can be introduced with this that specify
- when `signUp` is false, what authOptions.mode is, and that authOptions.login_hint is ui_locales.
- when `signUp` is true, that authOptions.mode is 'signUp'
- when `signUp` is true and locale is null, authOptions.login_hint is ui_locales.
- when `signUp` is true and locale is 'ab-CD', authOptions.login_hint is 'ab-CD'.

irahopkinson (commented):
For all that effort it didn't increase our coverage much but we are in a much better place to add tests for the auth service.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1074#-:-Mdb-JGV6MrDsOUf39ZA:b-89euyn)* status: 3 of 10 files reviewed, 3 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.ts, line 19 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1074#-MdOTNygCfMOziffl01s:-MdayHrE8vBiYv0wdQc_:bfp3aer) ([raw file](https://github.com/sillsdev/web-xforge/blob/f881f68650188d359a714230082eb531ff384e93/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.ts#L19)):*

I'm having trouble finding the tests for AuthGuard as well. 

Let's add to the AuthGuard tests, a specification for what is passed to authService.logIn() if there is no locale query parameter / absent, if it's empty (like in https://example.com/?abc=def&locale=&ghi=jkl, or if it's present, like `mn-OP`.
</blockquote></details>

I've added tests for this and I think its 100% covered!
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 168 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1074#-MdOPOxH-cA_PQaQzhL3:-MdayQ6_3Or7m1zwlNfT:b-q9vuxg) ([raw file](https://github.com/sillsdev/web-xforge/blob/f881f68650188d359a714230082eb531ff384e93/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L168)):*

> ```
>     const language = getAspCultureCookieLanguage(this.cookieService.get(ASP_CULTURE_COOKIE_NAME));
>     const ui_locales = language;
> ```

`ui_locales = language` is a bit confusing. It would at least be lessso if we add types to these. I _think_ these are both just strings, but a plural `ui_locales` sounds like it could be an array or something.

(`const ui_locales: string = ...`)

</blockquote></details>

I've added the types and I agree that makes things clearer. FYI, `ui_locales` is a common url parameter name for this usage, it may even be a standard somewhere. PT Registry has defined it for their parameter so I'm using the same name here. Its plural because its value is a space separated list of language tags, i.e. `'fr es en'` trying the first and so on until it has one it can support. That's not obvious in our code because we only want to specify one.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 173 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1074#-MdOQr2I7esrHB3pPRx7:-Mdaza3D-R6f7TzWYUtx:bvgl3bx) ([raw file](https://github.com/sillsdev/web-xforge/blob/f881f68650188d359a714230082eb531ff384e93/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L173)):*

I'm having trouble finding the unit tests for this class. Or, maybe we don't unit tests for our authorization service?

Some tests can be introduced with this that specify
- when `signUp` is false, what authOptions.mode is, and that authOptions.login_hint is ui_locales.
- when `signUp` is true, that authOptions.mode is 'signUp'
- when `signUp` is true and locale is null, authOptions.login_hint is ui_locales.
- when `signUp` is true and locale is 'ab-CD', authOptions.login_hint is 'ab-CD'.


</blockquote></details>

This probably wasn't being tested because the service was directly creating an instance of the Auth0 WebAuth class. I've now wrapped the init of this in its own service so we can test this one. I have covered all your suggested scenarios and that looks to be exhaustive.,

marksvc (approved):
This is great Ira, thanks for adding all these!
I know the pain of writing a new spec file for an existing class, so thank you for creating these. It is a good area of the application to have tests for, and a good situation to have it be easy for the next person to write a test.

I did some testing and it all worked well:
- Logging in from the front page with no trouble, when I was really already logged in.
- Logging out. On the front page leaving the UI language as English and clicking Log in. The Log in and Sign up tabs are in English.
- On the front page, setting the UI language as Spanish and clicking Log in. The tabs are now in Spanish. On the log in tab, clicking the Log in with Paratext button takes me to registry-dev.paratext.org, where the UI is also in Spanish. And then past that page, when I get to SF, the SF frontend UI is in Spanish. Nice.
- In the SF frontend, setting the UI language to Russian and logging out brings me to the front page, where the UI is now shown in Russian, and the UI language setting also has Russian selected. Clicking the log in button takes me to the auth0 login page where the UI is in russian. Good.
- On the front page, if I click Sign up, I am taken to the auth0 login page, but the Sign up tab is selected, and the UI is in the language that was specified in the home page language chooser. 

I approve. :ship: 

> Reviewed 4 of 7 files at r2, 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1074#-:-Mdck25T0c5h80QdZEIO:b-xj1lbj)* status: all files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.spec.ts, line 30 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1074#-MdcNPuJ4mSAxK5K01G6:-MdcNPuK95Si9F0g2kHh:bvwzb2l) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4f5f075585caa453074331b0a45c8660471a128/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.spec.ts#L30)):*
> ```XML
> 
>     canActivate$.subscribe(canActivate => {
>       expect(canActivate).toBe(true);
> ```

I was concerned that we might never get into this lambda and do the expect and verify. But debugging the unit test shows me that it does. Why do we get into this instead of just subscribe and then exit the test before running this lambda? Is it because `canActivate$` has a value at the ready from 
```cs
    return from(this.authService.isLoggedIn);
```
and so when we 
```cs
 canActivate$.subscribe(
```
we have something immediately that we can process, and so we do end up processing it before exiting the test?

Might it be that that is just timing that usually works out, but might not always, and so we do have the potential to exit the test before running the lambda that is passed to `.subscribe()`, or are we certainly going to run the lambda before the test exits? If it is certain, how do we know that?
When I debug the test, if I break on line 33 where the `it()` closes:
```cs
});
```
and on line 30 inside the lambda:
```cs
      expect(canActivate).toBe(true);
```
, I observe that we hit the breakpoint on line 33 _before_ getting to line 30. Almost like when the `canActivate$` goes out of scope, it's caused to process any incoming data in its subscription. But significantly, 
```cs
    canActivate$.subscribe(canActivate => {
```
doesn't just process its incoming data with the lambda passed to it right away, before getting to the end of the containing block.

Thanks for any explanations you can give here.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 168 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1074#-MdOPOxH-cA_PQaQzhL3:-MdccLBQ5wB15ZQW58Tc:b-m2neya) ([raw file](https://github.com/sillsdev/web-xforge/blob/f881f68650188d359a714230082eb531ff384e93/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L168)):*

I've added the types and I agree that makes things clearer. FYI, `ui_locales` is a common url parameter name for this usage, it may even be a standard somewhere. PT Registry has defined it for their parameter so I'm using the same name here. Its plural because its value is a space separated list of language tags, i.e. `'fr es en'` trying the first and so on until it has one it can support. That's not obvious in our code because we only want to specify one.
</blockquote></details>

Ok. Thanks. 
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 112 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1074#-MdcjNxI2DL1Tu68vVF-:-MdcjNxJ7Zg9eIdlvBU-:b-o76wzx) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4f5f075585caa453074331b0a45c8660471a128/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L112)):*
>     expect(locale).not.toEqual(env.language, 'setup');

:thumbsup:

irahopkinson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1074)* status: 9 of 10 files reviewed, all discussions resolved (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.spec.ts, line 30 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1074#-MdcNPuJ4mSAxK5K01G6:-MdgBD2b6cANv5vXLrIs:b-lab46s) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4f5f075585caa453074331b0a45c8660471a128/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.spec.ts#L30)):*

I was concerned that we might never get into this lambda and do the expect and verify. But debugging the unit test shows me that it does. Why do we get into this instead of just subscribe and then exit the test before running this lambda? Is it because `canActivate$` has a value at the ready from 
```cs
    return from(this.authService.isLoggedIn);
```
and so when we 
```cs
 canActivate$.subscribe(
```
we have something immediately that we can process, and so we do end up processing it before exiting the test?

Might it be that that is just timing that usually works out, but might not always, and so we do have the potential to exit the test before running the lambda that is passed to `.subscribe()`, or are we certainly going to run the lambda before the test exits? If it is certain, how do we know that?
When I debug the test, if I break on line 33 where the `it()` closes:
```cs
});
```
and on line 30 inside the lambda:
```cs
      expect(canActivate).toBe(true);
```
, I observe that we hit the breakpoint on line 33 _before_ getting to line 30. Almost like when the `canActivate$` goes out of scope, it's caused to process any incoming data in its subscription. But significantly, 
```cs
    canActivate$.subscribe(canActivate => {
```
doesn't just process its incoming data with the lambda passed to it right away, before getting to the end of the containing block.

Thanks for any explanations you can give here.
</blockquote></details>

Yes, I had wondered about this so did quite a bit of research on this. I was tempted to use marbles to test this so as to set up the right pattern for other tests with observables but I decided that was overkill for this. First thing to note is that we are testing a synchronous observable not an asynchronous one. Second, the call to the `subscribe` arrow function is potentially asynchronous so I would expect to hit the breakpoint at line 33 first. However since the observable is synchronous it should immediately then call the arrow function and hit your breakpoint in line 30. However, your question has prompted me to make a small change to add a call to `done`. This guards against the issue you are wondering about if the observable ever become asynchronous as the test now won't finish until `done` is called.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1074)* status: 9 of 10 files reviewed, all discussions resolved

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.spec.ts, line 30 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1074#-MdcNPuJ4mSAxK5K01G6:-MdrVbpa8zMzGJlz7mlN:bp4b2p5) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4f5f075585caa453074331b0a45c8660471a128/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.spec.ts#L30)):*

Yes, I had wondered about this so did quite a bit of research on this. I was tempted to use marbles to test this so as to set up the right pattern for other tests with observables but I decided that was overkill for this. First thing to note is that we are testing a synchronous observable not an asynchronous one. Second, the call to the `subscribe` arrow function is potentially asynchronous so I would expect to hit the breakpoint at line 33 first. However since the observable is synchronous it should immediately then call the arrow function and hit your breakpoint in line 30. However, your question has prompted me to make a small change to add a call to `done`. This guards against the issue you are wondering about if the observable ever become asynchronous as the test now won't finish until `done` is called.
</blockquote></details>

Hmmm. Interesting.  Thank you.

PR #1077, opened 2021-06-23
SF-901 Show USFM labels in text editor

Nateowami:
This is mostly working. What remains:
- [x] Refactor the code a bit. USX_STYLE_DESCRIPTIONS should probably be in a different file, but I haven't figured out where.
- [x] There's an issue where when in a list item that isn't the first, no description is shown. I haven't figured out what to do about this.

Right now looking for a preliminary review and feedback.

LTR | RTL
------|------
 | 

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077)


irahopkinson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-:-Mcw_9B_AYQJzpxULcjS:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

irahopkinson (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts, line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-Mcw_Y8W2k7dSf1FQ9Un:-Mcw_Y8W2k7dSf1FQ9Uo:b-4bhfhu) ([raw file](https://github.com/sillsdev/web-xforge/blob/5a1f6a313a5215b41ef803f0c44449597d81192c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L8)):*

> ```
> // Generated by running
> // curl -s https://raw.githubusercontent.com/ubsicap/usfm/master/sty/usfm.sty | grep '^\\Name ' | cut -c7-
> // | grep -v OBSOLETE | sed 's/^DEPRECATED //g' | sed 's/\.\.\.[[:alnum:]]\{1,\}\*//g'
> const USX_STYLE_DESCRIPTIONS = `id - File - Identification
> ```
</details>

Just put this in its own file right in this folder for now.

irahopkinson (approved):
BTW this is good enough to ship as is. I was tempted to merge it anyway.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-:-McwqITSAQRLhvo756QC:b-kep7w6)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

marksvc (changes requested):
I wonder if these are already localized in Paratext. If not, it may be good to use them via our normal localization process.

I tried out the change locally. It looks good. 

Sorry; I had more issues than Ira :)

> Reviewed 1 of 2 files at r1, 2 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-:-MczXBhREimL5WUolnCy:bo333f8)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts, line 12 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-Mcz_8qT87c3puzLLFsC:-Mcz_8qT87c3puzLLFsD:b-obthnl) ([raw file](https://github.com/sillsdev/web-xforge/blob/74d46dc90eafec7e9ce239b619ff008d93a55002/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L12)):*
> ```XML
> }
> 
> function getUsxDescription(value: string) {
> ```

I think a better return for a miss would be to return something like null, rather than ''. But it's not super important. 
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts, line 436 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-Mcz_Mrw6jaAnIdaFk_S:-Mcz_Mrw6jaAnIdaFk_T:b-6idcyi) ([raw file](https://github.com/sillsdev/web-xforge/blob/74d46dc90eafec7e9ce239b619ff008d93a55002/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L436)):*
> ```XML
>       node.setAttribute(customAttributeName('segment'), value);
>       const styleDescription = getUsxDescription(value);
>       if (styleDescription) {
> ```

I don't think we should interpret `styleDescription` for truth value this way. We also don't declare the type of `styleDescription` on the line before. Let's be explicit about what a true or false condition is by doing something like `if (styleDescription === '')`. 
Also a dozen lines below.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts, line 438 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-Mcz_n_m6hE2wQIgJIq_:-Mcz_n_m6hE2wQIgJIqa:b-hfl47h) ([raw file](https://github.com/sillsdev/web-xforge/blob/74d46dc90eafec7e9ce239b619ff008d93a55002/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L438)):*
> ```XML
>       if (styleDescription) {
>         node.setAttribute(customAttributeName('style-description'), styleDescription);
>       }
> ```

I don't see any unit tests for create, format, or getUsxDescription.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts, line 436 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-Mcz_Mrw6jaAnIdaFk_S:-MczyY0OAUbrJW07XF3V:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/74d46dc90eafec7e9ce239b619ff008d93a55002/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L436)):*

I don't think we should interpret `styleDescription` for truth value this way. We also don't declare the type of `styleDescription` on the line before. Let's be explicit about what a true or false condition is by doing something like `if (styleDescription === '')`. 
Also a dozen lines below.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts, line 438 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-Mcz_n_m6hE2wQIgJIq_:-MczywDN5Ow3E5QMUyLB:bsxf0h) ([raw file](https://github.com/sillsdev/web-xforge/blob/74d46dc90eafec7e9ce239b619ff008d93a55002/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L438)):*

I don't see any unit tests for create, format, or getUsxDescription.
</blockquote></details>

I'm not sure how to test the Quill blots, and there weren't any pre-existing tests, so I left them as-is. As for `getUsxDescription`, it didn't seem significant enough to write a test.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

Nateowami (commented):
@marksvc Did you intend to approve, or add more comments? I'm not sure where we're at.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-:-Md3EDAV15iWP_76Jd34:bgc0pgr)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami (commented):
Actually I think I just figured out a way to test this.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-:-Md3GxJZ9bLjp4Eymw-a:b9s4wx9)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

marksvc (changes requested):
So, let's check whether it inserts the style-description into the DB as a result of getting it into the DOM. 
Otherwise, I approve, minus one more blocking comment. 

> Reviewed 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-:-Md4D08TCAF7jZABWJVd:b-usvzou)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts, line 456 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-Md4Cs3LCj1xQR5Xxy6u:-Md4Cs3LCj1xQR5Xxy6v:b-9xk98c) ([raw file](https://github.com/sillsdev/web-xforge/blob/f539b9a300ec1d02bfc7fea8e7dafc17647c1738/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L456)):*
> ```XML
>         if (styleDescription != null) {
>           this.domNode.setAttribute(customAttributeName('style-description'), styleDescription);
>         }
> ```

So let's undo this hunk, since it didn't seem needed.

irahopkinson:
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts, line 456 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-Md4Cs3LCj1xQR5Xxy6u:-MdIZjt456ZnCVq7y7A4:b-c9g8v8) ([raw file](https://github.com/sillsdev/web-xforge/blob/f539b9a300ec1d02bfc7fea8e7dafc17647c1738/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L456)):*

So let's undo this hunk, since it didn't seem needed.
</blockquote></details>

I wouldn’t undo this. It would be a shame to lose the description when we format something.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077)* status: 3 of 7 files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts, line 456 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-Md4Cs3LCj1xQR5Xxy6u:-MdJ17jL8ZFiPa1voTjD:b-oqp4i0) ([raw file](https://github.com/sillsdev/web-xforge/blob/f539b9a300ec1d02bfc7fea8e7dafc17647c1738/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L456)):*

I wouldn’t undo this. It would be a shame to lose the description when we format something. 
</blockquote></details>

When Mark and I looked into this we came to the conclusion that this doesn't work quite how I thought it did and the format method was not something that we needed to implement this in. Also, I've completely replaced the method by which the attribute is added, so this is no longer relevant.

marksvc:
:eyes:

marksvc (changes requested):
Thanks for your work on this. If any of my requests sound painful, I am willing to pain with you :)

> Reviewed 1 of 2 files at r5, 3 of 3 files at r6.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-:-MdJkmYH0IiUlWIo0TiL:b-1oifr2)* status: all files reviewed, 5 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 211 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdJdM3EE2VQbiwv5wWU:-MdJdM3EE2VQbiwv5wWV:b-l9gjtt) ([raw file](https://github.com/sillsdev/web-xforge/blob/605c9dcd0b222b059804629ff94e1650ebfa35b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L211)):*
> ```XML
>     }
>     editor.updateContents(clearDelta, 'silent');
>     this.addUsfmDescription(Array.from(refs)[0]);
> ```

We might never get here with a refs with length 0 and this not be a problem. But if we ever do, this will crash. We could instead do 
```ts
if (refs.length > 0) {
  this.addUsfmDescription(Array.from(refs)[0]);
}
```

But it depends on whether we think we should gracefully handle the situation, or if we think that if we ever get here and refs.length===0, we should throw up our hands and crash. 

I'm guessing that gracefully handle is better but I have a lot to learn about this area of the code. 
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 221 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdJfsRL27RZ0V8w2ULK:-MdJfsRL27RZ0V8w2ULL:bf0fl1c) ([raw file](https://github.com/sillsdev/web-xforge/blob/605c9dcd0b222b059804629ff94e1650ebfa35b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L221)):*
> ```
> .querySelector('usx-segment')
> ```

What is this fetching here? Is it the element that was also fetched from the first line in the method? I think at this point, a method comment would be helpful. Perhaps along the lines of something like

> Add a `data-style-description` attribute, with a description of the corresponding USFM tag, to the usx-segment element containing segmentRef in the DOM.

I don't understand why we do an `element.querySelectory('usx-segment')`. Were we walking up the ancestry tree to find an ancestor element that contains both the segmentRef element _and_ the desired usx-segment element? Or, what?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 105 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdJc52c8GlrDNUB038i:-MdJc52dByoMWCYvqmhS:b-molurh) ([raw file](https://github.com/sillsdev/web-xforge/blob/605c9dcd0b222b059804629ff94e1650ebfa35b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L105)):*
> ```XML
>     tick();
>     env.fixture.detectChanges();
>     env.component.highlight(['s_1']);
> ```

It appears that this can change the DOM, which I assume lines manipulating `this.highlightMarker.style.visibility` are doing. So after the call to highlight, we should run `env.fixture.detectChanges();`, right?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 109 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdJjO9M0UQclasHeIgj:-MdJjO9M0UQclasHeIgk:b9abaps) ([raw file](https://github.com/sillsdev/web-xforge/blob/605c9dcd0b222b059804629ff94e1650ebfa35b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L109)):*
> ```XML
> 
>     const titleSegment = env.component.editor!.container.querySelector('usx-para[data-style="s"] usx-segment');
>     expect(titleSegment!.getAttribute('data-style-description')).toEqual('s - Heading - Section Level 1');
> ```

Before we call highlight(), we should also there assert that there is no such attribute, or whatever would mean that we _aren't_ showing the popup text when the segment hasn't been clicked. 
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 111 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdJhLQeDfMWmBXwx2Nm:-MdJhLQeDfMWmBXwx2Nn:bxl4wh1) ([raw file](https://github.com/sillsdev/web-xforge/blob/605c9dcd0b222b059804629ff94e1650ebfa35b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L111)):*
> ```XML
>     expect(titleSegment!.getAttribute('data-style-description')).toEqual('s - Heading - Section Level 1');
>   }));
> });
> ```

What about specifying the behaviour when the data-style is a string that isn't in the USFM_STYLE_DESCRIPTIONS dictionary?
What about specifying the behaviour when the element clicked has no associated USFM tag, and there is no parent with an associated USFM tag?

Nateowami (commented):
I've re-written this to use Quill to manipulate the document. It seems to be working correctly. It's ready for re-review.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-:-MdSoMx22KhYDYITARxw:b-wdnyuu)* status: 4 of 8 files reviewed, 5 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 211 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdJdM3EE2VQbiwv5wWU:-MdKnTfs-R00ib4efONy:b-egbbtx) ([raw file](https://github.com/sillsdev/web-xforge/blob/605c9dcd0b222b059804629ff94e1650ebfa35b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L211)):*

We might never get here with a refs with length 0 and this not be a problem. But if we ever do, this will crash. We could instead do 
```ts
if (refs.length > 0) {
  this.addUsfmDescription(Array.from(refs)[0]);
}
```

But it depends on whether we think we should gracefully handle the situation, or if we think that if we ever get here and refs.length===0, we should throw up our hands and crash. 

I'm guessing that gracefully handle is better but I have a lot to learn about this area of the code. 
</blockquote></details>

In reality it wouldn't have errored, but I think that's no longer relevant now that this section has been overhauled.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 221 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdJfsRL27RZ0V8w2ULK:-MdP4CuP1lkZgvHlr4_o:b-d8qabs) ([raw file](https://github.com/sillsdev/web-xforge/blob/605c9dcd0b222b059804629ff94e1650ebfa35b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L221)):*

> ```
> .querySelector('usx-segment')
> ```

What is this fetching here? Is it the element that was also fetched from the first line in the method? I think at this point, a method comment would be helpful. Perhaps along the lines of something like

> Add a `data-style-description` attribute, with a description of the corresponding USFM tag, to the usx-segment element containing segmentRef in the DOM.

I don't understand why we do an `element.querySelectory('usx-segment')`. Were we walking up the ancestry tree to find an ancestor element that contains both the segmentRef element _and_ the desired usx-segment element? Or, what?


</blockquote></details>

I've completely overhauled how this works to now manipulate the document by using Quill. I've also added comments.

However, the strategy remains fundamentally unchanged.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 105 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdJc52c8GlrDNUB038i:-MdP1Zln1-1LDFO7_Pk_:bn73o7y) ([raw file](https://github.com/sillsdev/web-xforge/blob/605c9dcd0b222b059804629ff94e1650ebfa35b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L105)):*

It appears that this can change the DOM, which I assume lines manipulating `this.highlightMarker.style.visibility` are doing. So after the call to highlight, we should run `env.fixture.detectChanges();`, right?
</blockquote></details>

I think `detectChanges()` is telling Angular to check for changes to the properties on components, and consider whether any of them may necessitate updates to the DOM.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 109 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdJjO9M0UQclasHeIgj:-MdSU9Ll1ubH81smJ2gv:bdhj6rq) ([raw file](https://github.com/sillsdev/web-xforge/blob/605c9dcd0b222b059804629ff94e1650ebfa35b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L109)):*

Before we call highlight(), we should also there assert that there is no such attribute, or whatever would mean that we _aren't_ showing the popup text when the segment hasn't been clicked. 
</blockquote></details>

Done. Also I figured out how to test that the CSS rules are creating a pseudo element.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 111 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdJhLQeDfMWmBXwx2Nm:-MdSjExV6W6PwgWcJs1m:bh3h52v) ([raw file](https://github.com/sillsdev/web-xforge/blob/605c9dcd0b222b059804629ff94e1650ebfa35b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L111)):*

What about specifying the behaviour when the data-style is a string that isn't in the USFM_STYLE_DESCRIPTIONS dictionary?
What about specifying the behaviour when the element clicked has no associated USFM tag, and there is no parent with an associated USFM tag?
</blockquote></details>

I've added a test for when we don't want a description shown. I'm not sure how to test the situation where there's a style with no associated USFM description, and it's also not something that's likely to go wrong, as it's a trivially simple edge case that we have a check for.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077)* status: 3 of 8 files reviewed, 6 unresolved discussions (waiting on @marksvc and @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 251 at r10](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdVLgu-3sLXupgSashX:-MdVLgu-3sLXupgSashY:b2swwxo) ([raw file](https://github.com/sillsdev/web-xforge/blob/9739582ce03b030f7e197b44ea34133d1035e82e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L251)):*
> ```XML
> 
>     for (const [i, op] of delta.ops.entries()) {
>       const len = typeof op.insert === 'string' ? op.insert.length : 1;
> ```

Does this work with blank inserts?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077)* status: 3 of 8 files reviewed, 6 unresolved discussions (waiting on @marksvc and @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 251 at r10](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdVLgu-3sLXupgSashX:-MdXVqFPBmW5vuo0cEm5:b-gkgw2q) ([raw file](https://github.com/sillsdev/web-xforge/blob/9739582ce03b030f7e197b44ea34133d1035e82e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L251)):*

Does this work with blank inserts?
</blockquote></details>

I believe so. Do you mean an insert like:
``` js
{
  insert: {
    blank: true
  },
  attributes: {
    segment: "verse_21_1"
  }
}
```
Since it's not inserting a string, if I understand correctly it's inserting an embed, which always has length 1.

I copied the length calculation verbatim from `highlight(segmentRefs: string[])`, the preceding method in this file, which has similar logic.

marksvc (approved):
Good job with this, Nathaniel. 

I did some testing in the SF frontend. Things are working well.
- If I click a segment in the target, a popup tells me the style description, but not if it's just a paragraph. 
- Clicking segments in the readonly source does not show a popup style description.
- Clicking segments in the community checking area does not show a popup style description.
- Clicking segments in the New Question dialog does not show a popup style description.

I haven't tested the behaviour when clicking a segment in a multi-segment paragraph with non-p styling, to observe the style description appearing above the first segment (because I haven't yet constructed a test with that structure). I assume you have been able to show that that works in your testing?

I approve. :sailboat: 

> Reviewed 3 of 3 files at r8, 2 of 2 files at r11.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-:-MddG1QgCe9PJfU9wzpO:b8beifk)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 248 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-Mdd6v7J8bnmp75f7LQC:-Mdd6v7J8bnmp75f7LQD:b-ujedsm) ([raw file](https://github.com/sillsdev/web-xforge/blob/19f399b4734aaa2ba63df53dc55842ce82c979ff/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L248)):*
> ```XML
>     // Math.max will return -Infinity when there is no preceding paragraph
>     const precedingParaIndex = Math.max(...styleOpIndexes.filter(i => i < highlightedSegmentIndex));
>     const segmentIndex = delta.ops.findIndex((op, i) => i > precedingParaIndex && op.attributes?.segment != null);
> ```

Maybe a good variable name here would be`indexOfFirstSegmentInParagraph` ? Hmm; that's long, but `segmentIndex` doesn't communicate the significance of what the segment is.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 427 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-Mdcwl4N9G_nRwCPLx0n:-Mdcwl4N9G_nRwCPLx0o:bbspqau) ([raw file](https://github.com/sillsdev/web-xforge/blob/19f399b4734aaa2ba63df53dc55842ce82c979ff/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L427)):*
> ```XML
>     if (segmentRefs != null) {
>       // this changes the underlying HTML, which can mess up some Quill events, so defer this call
>       Promise.resolve(segmentRefs).then(refs => {
> ```

Why the change to passing segmentRefs to .resolve() ?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 109 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdJjO9M0UQclasHeIgj:-Mdcxk1_ANlsKGOvnmoc:b-ul7oq9) ([raw file](https://github.com/sillsdev/web-xforge/blob/605c9dcd0b222b059804629ff94e1650ebfa35b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L109)):*

Done. Also I figured out how to test that the CSS rules are creating a pseudo element.
</blockquote></details>

Good job.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 106 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdcxPgZ2VkEug08aahB:-MdcxPgZ2VkEug08aahC:b1ednfi) ([raw file](https://github.com/sillsdev/web-xforge/blob/19f399b4734aaa2ba63df53dc55842ce82c979ff/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L106)):*
> ```XML
>     env.fixture.detectChanges();
> 
>     // document starts with no description on title
> ```

This is helpful. I appreciate the commented walk-thru the behaviour that the test is causing and specifying.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/usfm-style-descriptions.ts, line 4 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdcwBre4Rs-96p1NmZs:-MdcwBrf67dG1QHt0hMj:b-ans02q) ([raw file](https://github.com/sillsdev/web-xforge/blob/19f399b4734aaa2ba63df53dc55842ce82c979ff/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/usfm-style-descriptions.ts#L4)):*
> ```
> : { [style: string]: string } 
> ```

:thumbsup: specifying the type

Nateowami (commented):
Yes, it's been tested with multi-segment paragraphs that have a style.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-:-MddInyNEvzrURlXakEE:b8mlp8p)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @marksvc and @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 248 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-Mdd6v7J8bnmp75f7LQC:-MdddxJP5PkfJlZVjUck:b3h63vf) ([raw file](https://github.com/sillsdev/web-xforge/blob/19f399b4734aaa2ba63df53dc55842ce82c979ff/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L248)):*

Maybe a good variable name here would be`indexOfFirstSegmentInParagraph` ? Hmm; that's long, but `segmentIndex` doesn't communicate the significance of what the segment is.
</blockquote></details>

Good point. I was trying too hard to keep it all on one line.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 427 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-Mdcwl4N9G_nRwCPLx0n:-MddPu1u6i8j4Lqn5STe:b4hshik) ([raw file](https://github.com/sillsdev/web-xforge/blob/19f399b4734aaa2ba63df53dc55842ce82c979ff/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L427)):*

Why the change to passing segmentRefs to .resolve() ?
</blockquote></details>

Type safety. Previously we had to assert that the value of `segmentRefs` would still be an array at some point in the future (after `Promise.resolve()` completes), which is basically assuring the type system "don't worry; we promise we won't redefine this variable in the time between the null check above, and whatever point this lambda gets called." Which happens to be true, and we can verify by reading the rest of the method. But if we just pass in the value to the promise, then it no longer matters whether the variable changes value later, and the type system can verify type safety on its own.

Theoretically, if at some point in the future we redefined `segmentRefs` later in the method, the type system wouldn't know anything was wrong, but our code would crash because the type assertion would be wrong.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 106 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-MdcxPgZ2VkEug08aahB:-MddSsTFBSXJC1IGzEF9:b-2jt6xm) ([raw file](https://github.com/sillsdev/web-xforge/blob/19f399b4734aaa2ba63df53dc55842ce82c979ff/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L106)):*

This is helpful. I appreciate the commented walk-thru the behaviour that the test is causing and specifying.
</blockquote></details>

This is something I should do more. I often have trouble understanding tests and my eyes just glaze over.

irahopkinson (approved):
> Reviewed 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc and @nigel-wells)

marksvc (approved):
> Reviewed 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1077)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 427 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1077#-Mdcwl4N9G_nRwCPLx0n:-MdrU3iQ87Vagf0Lryq7:bbr08f7) ([raw file](https://github.com/sillsdev/web-xforge/blob/19f399b4734aaa2ba63df53dc55842ce82c979ff/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L427)):*

Type safety. Previously we had to assert that the value of `segmentRefs` would still be an array at some point in the future (after `Promise.resolve()` completes), which is basically assuring the type system "don't worry; we promise we won't redefine this variable in the time between the null check above, and whatever point this lambda gets called." Which happens to be true, and we can verify by reading the rest of the method. But if we just pass in the value to the promise, then it no longer matters whether the variable changes value later, and the type system can verify type safety on its own.

Theoretically, if at some point in the future we redefined `segmentRefs` later in the method, the type system wouldn't know anything was wrong, but our code would crash because the type assertion would be wrong.
</blockquote></details>

Interesting; thank you.

PR #1088, opened 2021-07-08
SF-1278 - Text lost, verse numbers duplicated, from multiple devices coming online

nigel-wells:
- Change usx-blank into an inline-block with a fixed width
- Move position of cursor to appear after usx-blank in order to keep it inside the usx-segment element
- Adjusted quill left/right handlers to work with new cursor position with usx-blank

**Please note that this fix is just related to data loss from going offline to online. If a bug other than that is found then check Jira as there may be a separate ticket for it.**

## Acceptance Tests
1. Load a project and navigate to a book in the translate app which has lots of blank segments
2. Load the same page in another browser or profile
3. Go offline (shutdown dotnet)
4. Insert text in to any of the segments across both browsers
5. Close down the tab in one of the browsers
6. Go online
7. OBSERVE: The browser that was left open should go online as normal
8. Load the tab again in the other browser
9. OBSERVE: The data in the other browser is synced and now text from both browsers are showing correctly in the correct segments with no missing data

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1088)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
@@ -318,8 +318,10 @@ quill-editor.rtl {
 
 usx-blank {
   width: 35px;
-  font-size: 0;
   display: inline-block;
+  span {
+    font-size: 0;
+  }
 }
 
 usx-char {
--- a/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
@@ -316,6 +316,11 @@ quill-editor.rtl {
   }
 }
 
+usx-blank {
+  width: 35px;
+  display: inline-block;
+}
+
 usx-char {
   &[data-style='it'] {
     font-style: italic;
--- a/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
@@ -318,6 +318,7 @@ quill-editor.rtl {
 
 usx-blank {
   width: 35px;
+  font-size: 0;
   display: inline-block;
 }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
@@ -125,9 +125,7 @@ export function registerScripture(): string[] {
     static tagName = 'usx-blank';
 
     static create(value: boolean): Node {
-      const node = super.create(value) as HTMLElement;
-      node.innerText = NBSP.repeat(8);
-      return node;
+      return super.create(value) as HTMLElement;
     }
 
     static value(_node: HTMLElement): boolean {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
@@ -125,7 +125,9 @@ export function registerScripture(): string[] {
     static tagName = 'usx-blank';
 
     static create(value: boolean): Node {
-      return super.create(value) as HTMLElement;
+      const node = super.create(value) as HTMLElement;
+      node.innerText = NBSP.repeat(8);
+      return node;
     }
 
     static value(_node: HTMLElement): boolean {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -476,10 +476,7 @@ export class TextViewModel {
       fixOffset++;
     } else if (segment.containsBlank && segment.length > 1) {
       // delete blank
-      const delta = new Delta()
-        .retain(segment.index + fixOffset)
-        .retain(segment.length - 1, { segment: segment.ref, 'para-contents': true, 'direction-segment': 'auto' })
-        .delete(1);
+      const delta = new Delta().retain(segment.index + fixOffset).delete(1);
       fixDelta = fixDelta.compose(delta);
       fixOffset--;
       const sel = editor.getSelection();
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -107,7 +107,14 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
         'move next, tab': {
           key: 'tab',
           shiftKey: false,
-          handler: () => this.moveNextSegment()
+          handler: () => {
+            if (this.isRtl && this.isSelectionAtSegmentEnd) {
+              this.moveNextSegment(false);
+              return false;
+            }
+            this.moveNextSegment();
+            return false;
+          }
         },
         'move prev, tab': {
           key: 'tab',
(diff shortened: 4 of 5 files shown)


RaymondLuong3 (changes requested):
It looks like the usx-blank element is now inline, but somehow it puts a 35px spacing above the actual segment that it belongs to.



> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1088#-:-MeBBEnYCpfaFajWtqyq:b9k8ali)* status: all files reviewed, 2 unresolved discussions (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 479 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1088#-MeBC0gA5IpVFE4i9JbZ:-MeBC0gBBJlqSGyarU7O:btrop57) ([raw file](https://github.com/sillsdev/web-xforge/blob/e24ed450c0e8cd820a1ab3e08293476b6df67e95/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L479)):*

> ```
>       const delta = new Delta()
>         .retain(segment.index + fixOffset)
>         .retain(segment.length - 1, { segment: segment.ref, 'para-contents': true, 'direction-segment': 'auto' })
>         .delete(1);
> ```
</details>

What changed so that there is no need for the second retain instruction?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 474 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1088#-MeBEb7Z2Y-layLY_t5M:-MeBEb7_8xANdo9vmwme:bkmd94r) ([raw file](https://github.com/sillsdev/web-xforge/blob/4302a5adfb0ce29d80be475b0d9cf8806b52456b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L474)):*
> ```
>   private getSelectionAtSegmentPosition(end: boolean) {
> ```
Add the return type. Also it's not clear by the name what this method is doing. Maybe a better name isSelectionAtSegmentPosition().

nigel-wells (commented):
Ah well spotted - I thought I was working in paragraphs! Have resolved by also adjusting the font size of the blanks

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1088#-:-MewaAv-8Nj4iqLFX3iY:bbawr8p)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 479 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1088#-MeBC0gA5IpVFE4i9JbZ:-MewaKO08szzGRPkGtdF:bxabsxc) ([raw file](https://github.com/sillsdev/web-xforge/blob/e24ed450c0e8cd820a1ab3e08293476b6df67e95/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L479)):*

> ```
>       const delta = new Delta()
>         .retain(segment.index + fixOffset)
>         .retain(segment.length - 1, { segment: segment.ref, 'para-contents': true, 'direction-segment': 'auto' })
>         .delete(1);
> ```
What changed so that there is no need for the second retain instruction?
</blockquote></details>

The 2nd retain was there in an attempt to fix up the fact the inserted character was outside of an actual segment. Now that the character is inserted inside the segment there is no longer a need to add the information in.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 474 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1088#-MeBEb7Z2Y-layLY_t5M:-Mewawp-8qwN_xMTtqWu:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/4302a5adfb0ce29d80be475b0d9cf8806b52456b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L474)):*

> ```
>   private getSelectionAtSegmentPosition(end: boolean) {
> ```
Add the return type. Also it's not clear by the name what this method is doing. Maybe a better name isSelectionAtSegmentPosition().
</blockquote></details>

Done.

nigel-wells (commented):
Hmm, I've just done an extra test after noticing we're running Angular 11 now and I see the font size has done an undesirable thing with the cursor. I'll look in to that now before you review.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1088#-:-MeweqpE9SSMGRuS5kxn:b-c17213)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

RaymondLuong3 (approved):
I have one non-blocking comment. Otherwise :lgtm: 

> Reviewed 2 of 3 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1088#-:-Mf-cnsZ61IFyKJ9Seo3:bwxwju4)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/_usx.scss, line 322 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1088#-Mf-cQqpBsgYTjjWXIp0:-Mf-cQqpBsgYTjjWXIp1:bjx8vt) ([raw file](https://github.com/sillsdev/web-xforge/blob/25026fd5e18f2d84f646458f292e347ef1d28a42/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss#L322)):*
> ```
>   span {
>     font-size: 0;
>   }
> ```
So this fixes the font size issue you saw? Would it be worth writing a test for this so that we can see if this causes us problems in future upgrades?

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1088)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/_usx.scss, line 322 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1088#-Mf-cQqpBsgYTjjWXIp0:-Mf-y5CL37Z9mpd6hNBY:bm3xe56) ([raw file](https://github.com/sillsdev/web-xforge/blob/25026fd5e18f2d84f646458f292e347ef1d28a42/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss#L322)):*

> ```
>   span {
>     font-size: 0;
>   }
> ```
So this fixes the font size issue you saw? Would it be worth writing a test for this so that we can see if this causes us problems in future upgrades?
</blockquote></details>

It appeared that the child `<span>` element was inheriting a font size causing the extra height you saw. This just forces the size (and therefore height) to something that won't create unwanted height. What exactly did you want to test given this is CSS - just to see if the CSS rule was applied?

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1088)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/_usx.scss, line 322 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1088#-Mf-cQqpBsgYTjjWXIp0:-Mf4goWw0tTOSHG6OwcU:byx4ym1) ([raw file](https://github.com/sillsdev/web-xforge/blob/25026fd5e18f2d84f646458f292e347ef1d28a42/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss#L322)):*

It appeared that the child `<span>` element was inheriting a font size causing the extra height you saw. This just forces the size (and therefore height) to something that won't create unwanted height. What exactly did you want to test given this is CSS - just to see if the CSS rule was applied?
</blockquote></details>

Ok, so the issue was that the font size wasn't specified and came from a parent element. I imagine it is unlikely to regress, so I'm fine with this change as-is.

PR #1093, opened 2021-07-10
Update jwks-rsa

Nateowami:
- Update to latest `jwks-rsa`
- Install jsonwebtoken, because we're using it and didn't have it in our `package.json`. Presumably it used to be required by `jwks-rsa`.

Background:
The `feature/ng11` branch wasn't working on my machine, because a newer version of `jwks-rsa` was supposed to respect the system proxy settings by reading environment variables. Unfortunately, this was implemented poorly and assumed if a proxy was specified that it was an http proxy or an https proxy, rather than checking the actual protocol that was specified. It then attempts to proceed in a way that does not cause any error to be thrown, but also failed to actually fetch the keys.

In version `jwks-rsa` version 2, they've done away with specifying a proxy, and instead if you want to do anything special with how the request is made, you can provide it an object that implements `http.Agent`/`https.Agent` (`http` and `https` are part of Node's native library). This is a much better solution for all the reasons that dependency injection is good, and opens up options that wouldn't have been possible with the previous version, such as mocking the key request in your tests, or reading the keys from a file.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1093)


irahopkinson (approved):
I can verify that `jsonwebtoken` is a dependency of `jwks-rsa` v1.6.0 (what is currently on master).
This is not currently covered by tests so I have manually tested this (but of course I wasn't having the problem).
:lgtm: thanks for fixing this. I did have the thought of looking at upgrading this when I did NG12 but I might not have remembered by then.

> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1093#-:-MeDaGyI0O6y0CwTeAoE:b1wdkgk)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1095, opened 2021-07-10
Add some auth service tests

irahopkinson:
- also changed the change password failure message to a dialog

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1095)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
@@ -475,7 +475,7 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
           this.noticeService.show(translate('app.password_reset_email_sent'));
         })
         .catch(() => {
-          this.noticeService.show(translate('app.cannot_change_password'));
+          this.noticeService.showMessageDialog(() => translate('app.cannot_change_password'));
         });
     }
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts
@@ -1,14 +1,25 @@
-import { fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { discardPeriodicTasks, fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { Router } from '@angular/router';
 import { RouterTestingModule } from '@angular/router/testing';
-import { WebAuth } from 'auth0-js';
+import { Auth0DecodedHash, Auth0Error, WebAuth } from 'auth0-js';
 import { CookieService } from 'ngx-cookie-service';
-import { of } from 'rxjs';
+import { Subject } from 'rxjs';
 import { anyString, anything, capture, instance, mock, resetCalls, verify, when } from 'ts-mockito';
-import { AuthService, EXPIRES_AT_SETTING } from './auth.service';
+import { SystemRole } from 'realtime-server/lib/esm/common/models/system-role';
+import {
+  ACCESS_TOKEN_SETTING,
+  AuthService,
+  AuthState,
+  EXPIRES_AT_SETTING,
+  ID_TOKEN_SETTING,
+  ROLE_SETTING,
+  USER_ID_SETTING,
+  XF_ROLE_CLAIM,
+  XF_USER_ID_CLAIM
+} from './auth.service';
 import { Auth0Service } from './auth0.service';
 import { BugsnagService } from './bugsnag.service';
-import { CommandService } from './command.service';
+import { CommandError, CommandErrorCode, CommandService } from './command.service';
 import { ErrorReportingService } from './error-reporting.service';
 import { LocalSettingsService } from './local-settings.service';
 import { LocationService } from './location.service';
@@ -53,7 +64,7 @@ describe('AuthService', () => {
     ]
   }));
 
-  it('should change password', () => {
+  it('should change password', fakeAsync(() => {
     const env = new TestEnvironment();
     const email = 'test@example.com';
 
@@ -66,14 +77,11 @@ describe('AuthService', () => {
       expect(changePasswordOptions.connection).not.toBeNull();
       expect(changePasswordOptions.email).toEqual(email);
     }
-  });
+  }));
 
   it('should not check online authentication if not logged in', fakeAsync(() => {
     const env = new TestEnvironment({ isOnline: true });
-    let isLoggedIn: boolean = true;
-    env.service.isLoggedIn.then(result => (isLoggedIn = result));
-    tick();
-    expect(isLoggedIn).toBe(false, 'setup');
+    expect(env.isLoggedIn).withContext('setup').toBe(false);
     resetCalls(mockedPwaService);
 
     env.service.checkOnlineAuth();
@@ -84,10 +92,7 @@ describe('AuthService', () => {
 
   it('should check online authentication if logged in', fakeAsync(() => {
     const env = new TestEnvironment();
-    let isLoggedIn: boolean = false;
-    env.service.isLoggedIn.then(result => (isLoggedIn = result));
-    tick();
-    expect(isLoggedIn).toBe(true, 'setup');
+    expect(env.isLoggedIn).withContext('setup').toBe(true);
     resetCalls(mockedPwaService);
 
     env.service.checkOnlineAuth();
@@ -95,6 +100,48 @@ describe('AuthService', () => {
     tick();
     verify(mockedPwaService.checkOnline()).once();
     verify(mockedWebAuth.authorize(anything())).never();
+  }));
+
+  it('check session is valid after returning online and login if session has expired', fakeAsync(() => {
+    const env = new TestEnvironment({
+      isOnline: true,
+      isLoggedIn: true
+    });
+    expect(env.isAuthenticated).toBe(true);
+    resetCalls(mockedPwaService);
+
+    env.setLoginRequiredResponse();
+    env.service.checkOnlineAuth();
+    tick();
+
+    verify(mockedPwaService.checkOnline()).once();
+    verify(mockedWebAuth.authorize(anything())).once();
+  }));
+
+  it('should log out and clear data', fakeAsync(() => {
+    const env = new TestEnvironment({
+      isOnline: true,
+      isLoggedIn: true
+    });
+    expect(env.isAuthenticated).toBe(true);
+    expect(env.service.currentUserId).toBe(TestEnvironment.userId);
+    expect(env.service.idToken).toBe(env.auth0Response.result!.idToken);
+    expect(env.service.currentUserRole).toBe(SystemRole.SystemAdmin);
+    expect(env.service.accessToken).toBe(env.auth0Response.result!.accessToken);
+    expect(env.service.expiresAt).toBeGreaterThan(env.auth0Response.result!.expiresIn!);
+
+    env.service.logOut();
+    tick();
+    expect(env.service.idToken).toBeUndefined();
+    expect(env.service.currentUserRole).toBeUndefined();
+    expect(env.service.accessToken).toBeUndefined();
+    expect(env.service.expiresAt).toBeUndefined();
+    verify(mockedWebAuth.logout(anything())).once();
+    const [logoutOptions] = capture(mockedWebAuth.logout).last();
+    expect(logoutOptions).toBeDefined();
+    if (logoutOptions != null) {
+      expect(logoutOptions.returnTo).toBeDefined();
+    }
   }));
 
   it('should expire token', fakeAsync(() => {
@@ -108,27 +155,40 @@ describe('AuthService', () => {
 
   it('should authenticate if expired', fakeAsync(() => {
     const env = new TestEnvironment();
-    let isAuthenticated: boolean = false;
-
-    env.service.isAuthenticated().then((result: boolean) => (isAuthenticated = result));
-
-    tick();
-    expect(isAuthenticated).toBe(true);
+    expect(env.isAuthenticated).toBe(true);
     verify(mockedWebAuth.checkSession(anything(), anything())).once();
   }));
 
-  it('should authenticate if not expired', fakeAsync(() => {
-    const env = new TestEnvironment({ expiresIn: 10 });
-    let isAuthenticated: boolean = false;
+  it('should renew tokens if expired and authenticating', fakeAsync(() => {
+    const env = new TestEnvironment({ isOnline: true, isLoggedIn: true });
+    expect(env.isAuthenticated).toBe(true);
 
-    env.service.isAuthenticated().then((result: boolean) => (isAuthenticated = result));
+    env.service.expireToken();
+    expect(env.service.expiresAt).toBe(0);
 
-    tick();
-    expect(isAuthenticated).toBe(true);
-    verify(mockedWebAuth.checkSession(anything(), anything())).never();
+    expect(env.isAuthenticated).toBe(true);
+    expect(env.service.expiresAt).toBeGreaterThan(0);
+    verify(mockedWebAuth.checkSession(anything(), anything())).twice();
+    env.discardTokenExpiryTimer();
   }));
 
-  it('should login', () => {
+  it('should renew tokens if expired and idle', fakeAsync(() => {
+    const env = new TestEnvironment({ isOnline: true, isLoggedIn: true });
+    expect(env.isAuthenticated).toBe(true);
+
+    env.service.expireToken();
+    expect(env.service.expiresAt).toBe(0);
+    // expireToken() doesn't actually change the timer subscribed to
+    //   normally isAuthenticated would be triggered next which would call renewTokens()
+    // For this test we'll make sure the timer expires so renewTokens() is called
+    //   to simulate someone using the app and the auth0 token expires
+    env.clearTokenExpiryTimer();
+    expect(env.service.expiresAt).toBeGreaterThan(0);
+    verify(mockedWebAuth.checkSession(anything(), anything())).twice();
+    env.discardTokenExpiryTimer();
+  }));
+
+  it('should attempt login via auth0', fakeAsync(() => {
     const env = new TestEnvironment();
     const returnUrl = 'test-returnUrl';
 
@@ -143,9 +203,9 @@ describe('AuthService', () => {
       expect(authOptions.login_hint).toEqual(env.language);
       expect(authOptions.mode).toBeUndefined();
     }
-  });
+  }));
 
-  it('should login without signup', () => {
+  it('should login without signup', fakeAsync(() => {
     const env = new TestEnvironment();
     const returnUrl = 'test-returnUrl';
     const signUp = false;
@@ -160,9 +220,9 @@ describe('AuthService', () => {
       expect(authOptions.login_hint).toEqual(env.language);
       expect(authOptions.mode).toBeUndefined();
     }
-  });
+  }));
 
-  it('should login with signup', () => {
+  it('should login with signup', fakeAsync(() => {
     const env = new TestEnvironment();
     const returnUrl = 'test-returnUrl';
     const signUp = true;
@@ -177,14 +237,14 @@ describe('AuthService', () => {
       expect(authOptions.login_hint).toEqual(env.language);
       expect(authOptions.mode).toEqual('signUp');
     }
-  });
+  }));
 
-  it('should login with signup and locale', () => {
+  it('should login with signup and locale', fakeAsync(() => {
     const env = new TestEnvironment();
     const returnUrl = 'test-returnUrl';
     const signUp = true;
     const locale = 'es';
-    expect(locale).not.toEqual(env.language, 'setup');
+    expect(locale).withContext('setup').not.toEqual(env.language);
 
     env.service.logIn(returnUrl, signUp, locale);
 
@@ -196,9 +256,9 @@ describe('AuthService', () => {
       expect(authOptions.login_hint).toEqual(locale);
       expect(authOptions.mode).toEqual('signUp');
     }
-  });
+  }));
 
-  it('should link with Paratext', () => {
+  it('should link with Paratext', fakeAsync(() => {
     const env = new TestEnvironment();
     const returnUrl = 'test-returnUrl';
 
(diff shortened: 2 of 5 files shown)


marksvc:
:eyes:

marksvc (changes requested):
Thanks for your work on this. I don't want to add friction to adding tests into the system! But I do have a couple items that stand out for action or might be a mistake, and also a number of items that I would like to learn more about. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-:-MeRJUPi-mh3GeC6SeYo:by67uqj)* status: all files reviewed, 5 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts, line 474 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeRHJWxB09Yrcfzh_7p:-MeRHJWxB09Yrcfzh_7q:b-pya4fk) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L474)):*
> ```XML
>         })
>         .catch(() => {
>           this.noticeService.showMessageDialog(() => translate('app.cannot_change_password'));
> ```

Yeah, an error dialog sounds better than an error snackbar.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeQjT0ZBdR5rDhs263c:-MeQjT0_063C-SgD8nZe:b-kjs006) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L68)):*
<details><summary><i alt="summary">Quoted 7 lines of code…</i></summary>

> ```
>     verify(mockedWebAuth.changePassword(anything(), anything())).once();
>     const [changePasswordOptions] = capture(mockedWebAuth.changePassword).last();
>     expect(changePasswordOptions).toBeDefined();
>     if (changePasswordOptions != null) {
>       expect(changePasswordOptions.connection).not.toBeNull();
>       expect(changePasswordOptions.email).toEqual(email);
>     }
> ```
</details>


It seems like this should be easier. It is possible to assert the arguments with something like

```ts
    verify(
      mockedWebAuth.changePassword(
        objectContaining({ connection: 'Username-Password-Authentication', email }),
        anyFunction()
      )
    ).once();
```

or the simple 
```ts
    verify(mockedWebAuth.changePassword(objectContaining({ email }), anyFunction())).once();
```

or in this case also 

```ts
    verify(
      mockedWebAuth.changePassword(deepEqual({ connection: 'Username-Password-Authentication', email }), anyFunction())
    ).once();
```

But then if there is a problem, you just get a not as helpful error like
> Error: Expected "changePassword(deepEqual([object Object]), anyFunction())" to be called 1 time(s). But has been called 0 time(s).

Rather than the more explanatory 
>         Error: Expected 'Username-Password-Authentication' to equal 'Something-Else'.

I've a couple times wanted to suggest the concise assert format, but... I have to give a thumbs up to this long-winded format because it makes the error messages so much better.

But I at least wanted to point out the objectContaining() option; and maybe we can use that some day if its error message gets spruced up. ts-mockito.d.ts also has some other useful items of note, like notNull().
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 175 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeQy2566k2nX-8N5eh1:-MeQy2566k2nX-8N5eh2:bra56pw) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L175)):*
> ```
> only
> ```

I know this test is a companion with the one below it, but technically this test doesn't specify that it updates the UI language _only_ if the user is logged in. But it does specify that the language is updated _if_ the user is logged in. 
Not a huge deal.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 180 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeQqBW83UWRf2_LDBwC:-MeQqBW97pSMQVO4aj-J:bacxoz6) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L180)):*
> ```XML
>     expect(interfaceLanguage).not.toEqual(env.language, 'setup');
>     let isLoggedIn: boolean = false;
>     env.service.isLoggedIn.then(result => (isLoggedIn = result));
> ```

:(  Seems like there ought to be something like `isLoggedIn = fakeAwait(env.service.isLoggedIn);`
since we can't here do `isLoggedIn = await env.service.isLoggedIn` since we are in fakeAsync not async.

Well, for my own learning, I just tried:

```ts
  function fakeAwait(promise: Promise<any>) {
    let theResult: any = null;
    promise.then((result: any) => (theResult = result));
    flushMicrotasks();
    return theResult;
  }

  it('should update interface language only if logged in', fakeAsync(() => {
    const env = new TestEnvironment();
    const interfaceLanguage = 'es';
    expect(interfaceLanguage).not.toEqual(env.language, 'setup');
    let isLoggedIn: boolean = fakeAwait(env.service.isLoggedIn);
    expect(isLoggedIn).toBe(true, 'setup');
   ...
```

and that's passing. So, interesting to try. 

Hmm, later I found that it can have subtly different behaviour when there are erroring conditions, but I did not look into it further.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 187 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeQupXtDN2-TUFdE4UW:-MeQupXtDN2-TUFdE4UX:b-m5f87e) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L187)):*
> ```XML
> 
>     flushMicrotasks();
>     verify(mockedCommandService.onlineInvoke(anyString(), anyString(), anything())).once();
> ```

The production code doesn't appear to differentiate between a changed or unchanged language specification.  I'm guessing maybe you wrote this part of the test and then realized the production code didn't differentiate between changed or unchanged language specification? Or maybe what you have is a first step toward specifying what should happen if `interfaceLanguage === env.language` in a future test (which it looks like is that the exact same behaviour should happen, according to the current production code) ?
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 188 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeQvLPQE-s5HwzRggmV:-MeQvLPQE-s5HwzRggmW:btylwk4) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L188)):*
> ```XML
>     flushMicrotasks();
>     verify(mockedCommandService.onlineInvoke(anyString(), anyString(), anything())).once();
>   }));
> ```

We may as well assert that the language passed to onlineInvoke() is the right one:

```ts
    capture(mockedCommandService.onlineInvoke).last()[2].language === interfaceLanguage;
```

Except that my  my `capture(mockedCommandService.onlineInvoke)` doesn't work, for reasons I haven't solved yet. 

So, 
```ts
    verify(mockedCommandService.onlineInvoke(anyString(), 'updateInterfaceLanguage', { language: interfaceLanguage }));
```
unless you understand what is going on with capture. I'm guessing it's because it returns a `Promise<T|undefined>` and I haven't specified enough to make that happy.

In context it's not too fair to mark this comment as a 'Blocker' but I want to surface it as a good action item :)
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 191 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeRGaO4CVhEMmRcIbCa:-MeRGaO58Lhmi-jQBoNh:bo8mve5) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L191)):*
> ```XML
> 
>   it('should not update interface language if logged out', fakeAsync(() => {
>     const env = new TestEnvironment({ isOnline: false });
> ```

What's the reason to add in here that we are not online? It seems like that may somewhat cloud the matter of whether we are in fact not updating the interface language because we are not logged in.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 223 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeR9zayDyeGcMF8qBhe:-MeR9zaz6939cCdU69Lj:bj2a08z) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L223)):*
> ```
> (!isOnline)
> ```

Why `!`? From pwa.service checkOnline() it looks like it should just be `isOnline`. (But clearly changing it causes the tests to fail.)
Is it something to do with a distinction between being logged in while online and being logged in while offline?

On the face of it, it appears backward.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 228 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeRDL9F8nkYSjwpCiJc:-MeRDL9F8nkYSjwpCiJd:b-8pfdwu) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L228)):*
> ```XML
>     } else {
>       when(mockedWebAuth.parseHash(anything())).thenCall(callback => callback({}));
>       when(mockedWebAuth.checkSession(anything(), anything())).thenCall(callback => callback({}));
> ```

I did not carefully study how we use WebAuth long enough to know if this is right.

Although... the signature of parseHash is 

```ts
    parseHash(callback: Auth0Callback<Auth0DecodedHash | null, Auth0ParseHashError>): void;
```

(and there is another one but I don't think it is relevant).
How then can we `thenResolve()` it if it's returning _void_, not a Promise? Is it something like that a Promise is essentially the same as the lambda that is being passed to parseHash, and so it just works out? But if that's so then why do you distinguish between 
```ts
when(mockedWebAuth.parseHash(anything())).thenResolve();
```
and
```ts
when(mockedWebAuth.parseHash(anything())).thenCall(callback => callback({}));
```
What's going on here? Maybe this would be a good item to talk about in our meeting.

The signature of checkSession() is 
```ts
checkSession(options: CheckSessionOptions, cb: Auth0Callback<any>): void;
```
How then can we write 
```ts
.thenCall(callback => callback({}))
```
because that means we use the _first_ argument as a function. But the first argument is the `options` not the `cb`. What's going on here?

marksvc (commented):
:heavy_check_mark: :  I smoke tested by doing a clean build + run and showing a text doc. Then, If I'm logged in with a username and password, and I click Change Password, then I get a snackbar message saying an email has been sent, and I can use that email to change my password at auth0. 
I see that the noticeService.showMessageDialog() happens if authService.changePassword fails. So I disable Internet access higher up in my network path (so the SF frontend wouldn't know fast enough that I'm offline and so to disable the Change Password menu option) and then clicked Change Password. It produced a dialog saying "Can't change ..." with Dismiss button. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-:-MeRL_D_03z08hQJSd7F:b-ucx7tq)* status: all files reviewed, 5 unresolved discussions (waiting on @irahopkinson)

irahopkinson (commented):
@nigel-wells a few minor things to fix and a question. I've responded to Mark's comments (his earlier review while a draft) but left 2 that you should respond to. I haven't thought through super carefully about everything that should be covered but I'm happy that we have got at least this far for now.

> Reviewed 2 of 4 files at r2, 3 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-:-MnycEdHDb2CmA-Fnfgp:by3ntgx)* status: all files reviewed, 8 unresolved discussions (waiting on @irahopkinson and @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 174 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MnyF_tj2IIUkp-orWv6:-MnyF_tj2IIUkp-orWv7:b6ir8cn) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc17d01d775dbac4ca9c4b85ffc4961609e658ba/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L174)):*
> ```
> this.unscheduleRenewal();
> ```
@nigel-wells where the 2 additions of this (here and in logOut) found to be needed as a result of testing? What effect will these fixes have on users?
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeQjT0ZBdR5rDhs263c:-MnyEG0LD6Z5fN79qHXL:b-cnzbz2) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L68)):*

> ```
>     verify(mockedWebAuth.changePassword(anything(), anything())).once();
>     const [changePasswordOptions] = capture(mockedWebAuth.changePassword).last();
>     expect(changePasswordOptions).toBeDefined();
>     if (changePasswordOptions != null) {
>       expect(changePasswordOptions.connection).not.toBeNull();
>       expect(changePasswordOptions.email).toEqual(email);
>     }
> ```

It seems like this should be easier. It is possible to assert the arguments with something like

```ts
    verify(
      mockedWebAuth.changePassword(
        objectContaining({ connection: 'Username-Password-Authentication', email }),
        anyFunction()
      )
    ).once();
```

or the simple 
```ts
    verify(mockedWebAuth.changePassword(objectContaining({ email }), anyFunction())).once();
```

or in this case also 

```ts
    verify(
      mockedWebAuth.changePassword(deepEqual({ connection: 'Username-Password-Authentication', email }), anyFunction())
    ).once();
```

But then if there is a problem, you just get a not as helpful error like
> Error: Expected "changePassword(deepEqual([object Object]), anyFunction())" to be called 1 time(s). But has been called 0 time(s).

Rather than the more explanatory 
>         Error: Expected 'Username-Password-Authentication' to equal 'Something-Else'.

I've a couple times wanted to suggest the concise assert format, but... I have to give a thumbs up to this long-winded format because it makes the error messages so much better.

But I at least wanted to point out the objectContaining() option; and maybe we can use that some day if its error message gets spruced up. ts-mockito.d.ts also has some other useful items of note, like notNull().

</blockquote></details>

Yep, I hear you, but at least what we have is easy to read and understand as well as give useful errors.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 180 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeQqBW83UWRf2_LDBwC:-Mny8zoh2RcQvL7P2g33:b-lf9zn3) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L180)):*

:(  Seems like there ought to be something like `isLoggedIn = fakeAwait(env.service.isLoggedIn);`
since we can't here do `isLoggedIn = await env.service.isLoggedIn` since we are in fakeAsync not async.

Well, for my own learning, I just tried:

```ts
  function fakeAwait(promise: Promise<any>) {
    let theResult: any = null;
    promise.then((result: any) => (theResult = result));
    flushMicrotasks();
    return theResult;
  }

  it('should update interface language only if logged in', fakeAsync(() => {
    const env = new TestEnvironment();
    const interfaceLanguage = 'es';
    expect(interfaceLanguage).not.toEqual(env.language, 'setup');
    let isLoggedIn: boolean = fakeAwait(env.service.isLoggedIn);
    expect(isLoggedIn).toBe(true, 'setup');
   ...
```

and that's passing. So, interesting to try. 

Hmm, later I found that it can have subtly different behaviour when there are erroring conditions, but I did not look into it further.
</blockquote></details>

Yes, it took me a long time to figure out how I could test this in the first place, but settled on using the promise as is.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 187 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeQupXtDN2-TUFdE4UW:-MnyBxnC2gDGNkOBQnn9:b-ddj5kc) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L187)):*

The production code doesn't appear to differentiate between a changed or unchanged language specification.  I'm guessing maybe you wrote this part of the test and then realized the production code didn't differentiate between changed or unchanged language specification? Or maybe what you have is a first step toward specifying what should happen if `interfaceLanguage === env.language` in a future test (which it looks like is that the exact same behaviour should happen, according to the current production code) ?
</blockquote></details>

No, I wasn't thinking about any of that. The 2 companion tests are just checking the current functionality that it only tries attempts the update if we are online.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 191 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeRGaO4CVhEMmRcIbCa:-MnyCdSvAn9_Cp4PwtR0:b-rtray7) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L191)):*

What's the reason to add in here that we are not online? It seems like that may somewhat cloud the matter of whether we are in fact not updating the interface language because we are not logged in.
</blockquote></details>

This comment was posted while the PR was draft. No longer relevant.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 223 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeR9zayDyeGcMF8qBhe:-MniKRp82YEpkf5242mn:b-rw1tkf) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L223)):*

> ```
> (!isOnline)
> ```

Why `!`? From pwa.service checkOnline() it looks like it should just be `isOnline`. (But clearly changing it causes the tests to fail.)
Is it something to do with a distinction between being logged in while online and being logged in while offline?

On the face of it, it appears backward.
</blockquote></details>

It was, fixed now.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 228 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeRDL9F8nkYSjwpCiJc:-MnyD5Ta3HW9hca-SsPj:b-rtray7) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L228)):*

I did not carefully study how we use WebAuth long enough to know if this is right.

Although... the signature of parseHash is 

```ts
    parseHash(callback: Auth0Callback<Auth0DecodedHash | null, Auth0ParseHashError>): void;
```

(and there is another one but I don't think it is relevant).
How then can we `thenResolve()` it if it's returning _void_, not a Promise? Is it something like that a Promise is essentially the same as the lambda that is being passed to parseHash, and so it just works out? But if that's so then why do you distinguish between 
```ts
when(mockedWebAuth.parseHash(anything())).thenResolve();
```
and
```ts
when(mockedWebAuth.parseHash(anything())).thenCall(callback => callback({}));
```
What's going on here? Maybe this would be a good item to talk about in our meeting.

The signature of checkSession() is 
```ts
checkSession(options: CheckSessionOptions, cb: Auth0Callback<any>): void;
```
How then can we write 
```ts
.thenCall(callback => callback({}))
```
because that means we use the _first_ argument as a function. But the first argument is the `options` not the `cb`. What's going on here?


</blockquote></details>

This comment was posted while the PR was draft. No longer relevant.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 84 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MnySrXq2JImW6BPhzdz:-MnySrXq2JImW6BPhze-:bissgid) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc17d01d775dbac4ca9c4b85ffc4961609e658ba/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L84)):*
> ```
> expect(env.isLoggedIn).toBe(false, 'setup');
> ```
nit: using **expectationFailOutput**, e.g. the second argument `expect(env.isLoggedIn).toBe(false, 'setup');` is deprecated. Use this syntax instead `expect(env.isLoggedIn).withContext('setup').toBe(false, 'setup');`. Here and below.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 452 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MnyPfd09ZpQUM-HV1hO:-MnyPfd09ZpQUM-HV1hP:bua2m5t) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc17d01d775dbac4ca9c4b85ffc4961609e658ba/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L452)):*
> ```
>   private localSettings: Map<string, string | number> = new Map<string, string | number>();
>   private _localeSettingsRemoveChanges: Subject<StorageEvent> = new Subject<StorageEvent>();
> ```
It's redundant to declare the type when its inferred from the value. Remove the type declaration to improve readability.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 458 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MnyMV0L0fhYi6YuLPGb:-MnyMV0L0fhYi6YuLPGc:b-gt67lc) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc17d01d775dbac4ca9c4b85ffc4961609e658ba/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L458)):*
> ```
> seperated
> ```
nit: sp
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 504 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MnyR6UjAUsqr4FxbGX2:-MnyR6UjAUsqr4FxbGX3:bzgc5uz) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc17d01d775dbac4ca9c4b85ffc4961609e658ba/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L504)):*
> ```
> url, method
> ```
nit: add an '_' to the beginning of unused parameters to declare them as such.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/memory-realtime-remote-store.ts, line 13 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MnyGb8V7L_1BKme8rDQ:-MnyGb8V7L_1BKme8rDR:b78vx7m) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc17d01d775dbac4ca9c4b85ffc4961609e658ba/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/memory-realtime-remote-store.ts#L13)):*
> ```
> private getAccessToken?: () => string | undefined;
> ```
Its value is never read. Tell me why this (and the init method) was added? It could be that you just need the `init` method to exist in which case we should consider adding it to the abstract class `RealtimeRemoteStore` so it can't be forgotten again.

marksvc (changes requested):
I did not examine this version of the PR, but I did look at discussion related to the earlier version of the PR to process those comments. And I'll otherwise leave the review to Ira, but can examine more if requested. I posted a solution for checking the arguments to `mockedCommandService.onlineInvoke`. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-:-Mo0l7j73zXtStqxcRb6:by5d5ir)* status: all files reviewed, 7 unresolved discussions (waiting on @irahopkinson and @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 188 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeQvLPQE-s5HwzRggmV:-Mo0j5Q6CLQq8Jy4x-7u:b-4e1v7u) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L188)):*

We may as well assert that the language passed to onlineInvoke() is the right one:

```ts
    capture(mockedCommandService.onlineInvoke).last()[2].language === interfaceLanguage;
```

Except that my  my `capture(mockedCommandService.onlineInvoke)` doesn't work, for reasons I haven't solved yet. 

So, 
```ts
    verify(mockedCommandService.onlineInvoke(anyString(), 'updateInterfaceLanguage', { language: interfaceLanguage }));
```
unless you understand what is going on with capture. I'm guessing it's because it returns a `Promise<T|undefined>` and I haven't specified enough to make that happy.

In context it's not too fair to mark this comment as a 'Blocker' but I want to surface it as a good action item :)
</blockquote></details>

I figured out how to make it work. This tests that onlineInvoke() is called, and that it's really called with the `updateInterfaceLanguage` method, not some unrelated method (like `double-check-that-i-am-logged-in`). 

```ts
    const [, method, params] = capture<string, string, any>(mockedCommandService.onlineInvoke).last();
    expect(method).toEqual('updateInterfaceLanguage');
    expect(params).toEqual({ language: interfaceLanguage });
```

nigel-wells (commented):
Thanks guys - I've made some tweaks based on your feedback Ira.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-:-Mo1O2y6Ae700USHTDNW:bjw69q2)* status: 2 of 6 files reviewed, 5 unresolved discussions (waiting on @irahopkinson, @marksvc, and @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 174 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MnyF_tj2IIUkp-orWv6:-Mo1Jh9J6miS6bP7yDf2:brcoxn1) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc17d01d775dbac4ca9c4b85ffc4961609e658ba/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L174)):*

> ```
> this.unscheduleRenewal();
> ```
@nigel-wells where the 2 additions of this (here and in logOut) found to be needed as a result of testing? What effect will these fixes have on users?
</blockquote></details>

They probably have zero affect on the end user. I noticed them during testing as, after the test, there were still timers left running so unsubscribing resolves those timers. In reality the browser will redirect so it won't make a difference. I can just as easily dispose the timer in the test rather than adding using `unscheduleRenewal()` if you felt that was a better approach.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 84 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MnySrXq2JImW6BPhzdz:-Mo1KmLH3eYjYrxNkHUE:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc17d01d775dbac4ca9c4b85ffc4961609e658ba/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L84)):*

> ```
> expect(env.isLoggedIn).toBe(false, 'setup');
> ```
nit: using **expectationFailOutput**, e.g. the second argument `expect(env.isLoggedIn).toBe(false, 'setup');` is deprecated. Use this syntax instead `expect(env.isLoggedIn).withContext('setup').toBe(false, 'setup');`. Here and below.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 452 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MnyPfd09ZpQUM-HV1hO:-Mo1Lie9CtziaOIuREKO:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc17d01d775dbac4ca9c4b85ffc4961609e658ba/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L452)):*

> ```
>   private localSettings: Map<string, string | number> = new Map<string, string | number>();
>   private _localeSettingsRemoveChanges: Subject<StorageEvent> = new Subject<StorageEvent>();
> ```
It's redundant to declare the type when its inferred from the value. Remove the type declaration to improve readability.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 458 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MnyMV0L0fhYi6YuLPGb:-Mo1LN_DBHsYT0I831W1:bc6uln9) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc17d01d775dbac4ca9c4b85ffc4961609e658ba/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L458)):*

> ```
> seperated
> ```
nit: sp
</blockquote></details>

Always spell that word wrong!
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 504 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MnyR6UjAUsqr4FxbGX2:-Mo1LmZ25jq0qVi4UPGy:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc17d01d775dbac4ca9c4b85ffc4961609e658ba/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L504)):*

> ```
> url, method
> ```
nit: add an '_' to the beginning of unused parameters to declare them as such.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/memory-realtime-remote-store.ts, line 13 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MnyGb8V7L_1BKme8rDQ:-Mo1MBleCTZ6ULgz1Y02:b-49lodd) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc17d01d775dbac4ca9c4b85ffc4961609e658ba/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/memory-realtime-remote-store.ts#L13)):*

> ```
> private getAccessToken?: () => string | undefined;
> ```
Its value is never read. Tell me why this (and the init method) was added? It could be that you just need the `init` method to exist in which case we should consider adding it to the abstract class `RealtimeRemoteStore` so it can't be forgotten again.
</blockquote></details>

I've added to the abstract class. Yes I was aware the access token isn't used in my instance but thought it might be for a test later on down the track - I just followed what the original purpose of the remote store did with the token.

irahopkinson (commented):
@nigel are you able to see the 2 remaining posts from Mark that still need you to respond?

> Reviewed 4 of 4 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-:-Mo1fsKBEkamXHYc4fhP:b-m6o525)* status: all files reviewed, 2 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 174 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MnyF_tj2IIUkp-orWv6:-Mo1eGHj0hHiX-DbuWba:b-rzenrw) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc17d01d775dbac4ca9c4b85ffc4961609e658ba/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L174)):*

They probably have zero affect on the end user. I noticed them during testing as, after the test, there were still timers left running so unsubscribing resolves those timers. In reality the browser will redirect so it won't make a difference. I can just as easily dispose the timer in the test rather than adding using `unscheduleRenewal()` if you felt that was a better approach.
</blockquote></details>

Leave as is since the tests exposed the need for cleanup, that is good.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 452 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MnyPfd09ZpQUM-HV1hO:-Mo1euGRCBsPUfAVlhGZ:b-7frh0g) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc17d01d775dbac4ca9c4b85ffc4961609e658ba/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L452)):*

Done.
</blockquote></details>

I highlighted 2 lines. You fixed the second one but still need to do the first.

nigel-wells (commented):
No I don't see anything outstanding or unresolved from my end - I guess it's because you own the PR so it is waiting for you to respond? 

Oh wait - it has just updated! I'll take a look

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-:-Mo2R7In2fKEHteiqpAP:b-ufammr)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 452 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MnyPfd09ZpQUM-HV1hO:-Mo2QdWDEBiSwtsfv0Pf:b-mveesh) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc17d01d775dbac4ca9c4b85ffc4961609e658ba/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L452)):*

I highlighted 2 lines. You fixed the second one but still need to do the first.
</blockquote></details>

Ah so you did - fixed.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1095)* status: 5 of 6 files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 188 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeQvLPQE-s5HwzRggmV:-Mo2V7Ea-RVY-S8xt7aA:bn8ig2y) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L188)):*
> const [, method, params] = capture<string, string, any>(mockedCommandService.onlineInvoke).last();
>     expect(method).toEqual('updateInterfaceLanguage');
>     expect(params).toEqual({ language: interfaceLanguage });
Thanks @marksvc I've updated to what you've suggested. This is a good approach when you do want to test a specific instance i.e. last and not just if it happened to be called by somewhere in the code.

irahopkinson (commented):
> Reviewed 1 of 1 files at r7.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1095)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 175 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeQy2566k2nX-8N5eh1:-Mo2hE_D2lhLvbULGiEo:bb6un1j) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L175)):*

> ```
> only
> ```

I know this test is a companion with the one below it, but technically this test doesn't specify that it updates the UI language _only_ if the user is logged in. But it does specify that the language is updated _if_ the user is logged in. 
Not a huge deal.
</blockquote></details>

Done
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 188 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1095#-MeQvLPQE-s5HwzRggmV:-Mo2hK_U-mh4OrM5Hc0x:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/60659bf5ea2db9375fd5298a96032542d044efe7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L188)):*

> const [, method, params] = capture<string, string, any>(mockedCommandService.onlineInvoke).last();
>     expect(method).toEqual('updateInterfaceLanguage');
>     expect(params).toEqual({ language: interfaceLanguage });
Thanks @marksvc I've updated to what you've suggested. This is a good approach when you do want to test a specific instance i.e. last and not just if it happened to be called by somewhere in the code.
</blockquote></details>

Done.

PR #1097, opened 2021-07-13
SF-1344 Remove formatting and newlines when drag and drop

marksvc:
- Turn off browser handling of drag and drop.
- Insert and remove plain text using deltas.
- In the made-up string, `x-dragstart-in-quill`, the `x-` is not
following any documentation for how to describe entries in
DataTransfer.types.
- text.component.html: Remove the disabling of drag-and-drop handling.
- text.component.ts: Add dragAndDrop to list of modules, and
specifying the TextComponent itself to pass to the DragAndDrop
constructor options.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1097)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
@@ -1,0 +1,101 @@
+import Quill, { RangeStatic } from 'quill';
+import { Delta } from 'rich-text';
+import { TextComponent } from './text.component';
+
+export interface DragAndDropOptions {
+  textComponent: TextComponent;
+}
+
+/** Drag-and-drop module for Quill, for dropping only unformatted text into a usx-segment. Unit tests on the behaviour
+ *  specification can be found in text.component.spec.ts.*/
+export class DragAndDrop {
+  /** Mostly arbitrary value to indicate that our quill is the source of the drag data. */
+  static readonly quillIsSourceToken: string = 'x-dragstart-in-quill';
+
+  constructor(quill: Quill, options: DragAndDropOptions) {
+    quill.container.addEventListener('dragstart', (event: Event) => {
+      const dragEvent = event as DragEvent;
+      // Write a custom note on the event. This will only end up being done if the drag started from in quill. In this
+      // way, we can differentiate between drag-and-drops that have their source as quill or from elsewhere, such as
+      // elsewhere in the same web page window or from another window or application.
+      dragEvent.dataTransfer?.setData(DragAndDrop.quillIsSourceToken, '');
+    });
+
+    quill.container.addEventListener('drop', (event: Event) => {
+      const dragEvent = event as DragEvent;
+      // Stop the browser from doing any drag-and-drop behaviour itself, such as inserting text with formatting.
+      dragEvent.preventDefault();
+
+      if ((dragEvent.target as Element)?.localName !== 'usx-segment') {
+        // We need to be able to know where to insert the dropped text, such as from the drop target being a usx-segment
+        // element.
+        return;
+      }
+      if (dragEvent.dataTransfer == null) {
+        console.warn(`Warning: DragEvent unexpectedly has null dataTransfer property.`);
+        return;
+      }
+
+      // Determine where we should be placing the dropped text, using the location of the destination segment, and the
+      // index within that segment.
+
+      const destinationSegmentRef: string = (dragEvent.target as Element).attributes['data-segment'].value;
+      const destinationSegmentRange: RangeStatic | undefined =
+        options.textComponent.getSegmentRange(destinationSegmentRef);
+      if (destinationSegmentRange == null) {
+        console.warn('Warning: drag-and-drop destination segment had an undefined range.');
+        return;
+      }
+
+      // Determine character index of drop location in destination segment, using a browser-specific method.
+      let startPositionInSegment: number = 0;
+      if (document.caretRangeFromPoint !== undefined) {
+        // Chromium/Chrome, Edge, and Safari browsers
+        const range: Range = document.caretRangeFromPoint(dragEvent.clientX, dragEvent.clientY);
+        startPositionInSegment = range.startOffset;
+      } else if (document.caretPositionFromPoint !== undefined) {
+        // Firefox browser
+        const range: CaretPosition | null = document.caretPositionFromPoint(dragEvent.clientX, dragEvent.clientY);
+        if (range == null) {
+          console.warn('Warning: drag-and-drop inferred a null caret position for insertion.');
+          return;
+        }
+        startPositionInSegment = range.offset;
+      } else {
+        console.warn(`Warning: Could not determine insertion position for drag-and-drop.`);
+        return;
+      }
+
+      const startingPositionInDocument: number = destinationSegmentRange.index + startPositionInSegment;
+
+      const insertionDelta = new Delta();
+      insertionDelta.retain(startingPositionInDocument);
+      let newText: string = dragEvent.dataTransfer.getData('text/plain');
+      // Omit newline characters
+      newText = newText.replace(/(?:\r?\n)+/g, ' ');
+      insertionDelta.insert(newText);
+      quill.updateContents(insertionDelta, 'user');
+
+      // Identify the selection range, if any, now that we updated the document with an insert. This will be a
+      // selection that was already present before the drop, whether it is the text that was dragged, or a selection
+      // not related to the drag.
+      const originalSelection: RangeStatic | null = quill.getSelection();
+
+      // Select the inserted text.
+      quill.setSelection(startingPositionInDocument, newText.length);
+
+      if (originalSelection != null) {
+        // There was a selection before the drop occurred.
+        const removalDelta = new Delta();
+        removalDelta.retain(originalSelection.index);
+        if (dragEvent.dataTransfer.types.includes(DragAndDrop.quillIsSourceToken)) {
+          // If the drag was started from within quill, then treat the selection as the source data of the drag, and
+          // delete the selection.
+          removalDelta.delete(originalSelection.length);
+          quill.updateContents(removalDelta, 'user');
+        }
+        // Or if the drag was not started from within quill, then don't delete the selection.
+      }
+    });
+  }
+}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
@@ -1,6 +1,7 @@
 import cloneDeep from 'lodash-es/cloneDeep';
 import Parchment from 'parchment';
 import Quill, { Clipboard, DeltaOperation, DeltaStatic, History, HistoryStackType } from 'quill';
+import { DragAndDrop } from './drag-and-drop';
 
 const Delta: new () => DeltaStatic = Quill.import('delta');
 
@@ -681,6 +682,7 @@ export function registerScripture(): string[] {
   Quill.register('blots/text', NotNormalizedText, true);
   Quill.register('modules/clipboard', DisableHtmlClipboard, true);
   Quill.register('modules/history', FixSelectionHistory, true);
+  Quill.register('modules/dragAndDrop', DragAndDrop);
 
   return formatNames;
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.html
@@ -1,7 +1,5 @@
 <quill-editor
   #quillEditor
-  ondragstart="return false;"
-  ondrop="return false;"
   spellcheck="false"
   [placeholder]="placeholder"
   [readOnly]="readOnlyEnabled"
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts
@@ -3,12 +3,12 @@ import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { Component, ViewChild } from '@angular/core';
 import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { TranslocoService } from '@ngneat/transloco';
-import Quill from 'quill';
+import Quill, { RangeStatic } from 'quill';
 import { SFProject } from 'realtime-server/lib/esm/scriptureforge/models/sf-project';
 import { TextData } from 'realtime-server/lib/esm/scriptureforge/models/text-data';
 import * as RichText from 'rich-text';
 import { BehaviorSubject } from 'rxjs';
-import { anything, mock, when } from 'ts-mockito';
+import { anything, instance, mock, when } from 'ts-mockito';
 import { AvatarTestingModule } from 'xforge-common/avatar/avatar-testing.module';
 import { BugsnagService } from 'xforge-common/bugsnag.service';
 import { PwaService } from 'xforge-common/pwa.service';
@@ -23,6 +23,7 @@ import { Delta, TextDoc, TextDocId } from '../../core/models/text-doc';
 import { SFProjectService } from '../../core/sf-project.service';
 import { SharedModule } from '../../shared/shared.module';
 import { getSFProject, getTextDoc } from '../test-utils';
+import { DragAndDrop } from './drag-and-drop';
 import { TextComponent } from './text.component';
 
 const mockedBugsnagService = mock(BugsnagService);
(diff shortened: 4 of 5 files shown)


irahopkinson (changes requested):
On the wider perspective, did you investigate not doing the `preventDefault` and just modifying the event to only contain plain text on the way through? This approach might remove the need to special things for each browser. Either way we should talk about the implications of this change.

Also how are you interpreting the codecov report?

Good job with your tests, they are easy to follow and therefore should be easy to modify if needed.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-:-Mf7TdcZDlxRkxF96vNo:b-41k8c1)* status: all files reviewed, 4 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 16 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-Mf7Nzn7CILfdigXpZ7Q:-Mf7Nzn7CILfdigXpZ7R:bn7knf4) ([raw file](https://github.com/sillsdev/web-xforge/blob/86352b9317890801ab5f555dc58552645fb7799a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L16)):*
> ```
> event: Event
> ```
Is there some reason why this is not declared as `DragEvent` rather than using the cast  ('as') on the next line? Here and below.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 35 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-Mf7QBE2B95Kr5eL1hLF:-Mf7QBE2B95Kr5eL1hLG:b-4br94t) ([raw file](https://github.com/sillsdev/web-xforge/blob/86352b9317890801ab5f555dc58552645fb7799a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L35)):*
> ```
> console.log
> ```
How about using `console.warn` here? Here and below.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 43 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-Mf7ROoY6EzGtTq0A9JE:-Mf7ROoY6EzGtTq0A9JF:b-ey9ckd) ([raw file](https://github.com/sillsdev/web-xforge/blob/86352b9317890801ab5f555dc58552645fb7799a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L43)):*
> ```
> destanation
> ```
spelling
Does your IDE give you hints when you have a spelling error? In VSCode I have the Cod Spell Checker extension that uses the dictionary in our repo and lets me know of spell errors in variable names.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 64 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-Mf7ST_z8Kf_q1DJ4FRG:-Mf7ST_z8Kf_q1DJ4FRH:bpi2k4p) ([raw file](https://github.com/sillsdev/web-xforge/blob/86352b9317890801ab5f555dc58552645fb7799a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L64)):*
> ```
>  } else {
> ```
Will either of the above methods work in Safari?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 154 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-Mf7VSil6X2MT33PCV3k:-Mf7VSil6X2MT33PCV3l:b-makj99) ([raw file](https://github.com/sillsdev/web-xforge/blob/86352b9317890801ab5f555dc58552645fb7799a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L154)):*
> ```
> MockDragEvent
> ```
Good job figuring out how to test using this.

marksvc (commented):
> On the wider perspective, did you investigate not doing the preventDefault and just modifying the event to only contain plain text on the way through?

Yes! And this would be preferable. When you drag text in quill, you have the data in both text/plain and text/html formats. And so you can just clear out the text/html format data, only the text/plain format remains, and so only plain text is dragged. The problem is that this is only possible when we are dragging within the same quill. If we drag text in from another window, we can't just clear away the text/html data. Since we need to do the more tricky work anyway, I just have both quill-to-quill and otherwindow-to-quill handled by the same route. For more info on this limitation, see a 2021-07-05 comment in JIRA that discusses this. 

> Also how are you interpreting the codecov report?

The codecov report summary doesn't look glowing, does it. 

The introduced misses are basically from two sources:
1. The Firefox if-branch of getting the character position from a point.
2. Instances of only continuing if something is defined (`foo`), so I can later look at its fields (`foo.bar`) and know that it isn't going to be null. For example, 
```ts
if (destanationSegmentRange == null) { return; }
```
means I can later do
```ts
baz = destanationSegmentRange.index
```
without there defending against destanationSegmentRange being null, such as with `destanationSegmentRange?.index` etc.

I added a test today for the Firefox if-branch to be run thru.

> Good job with your tests, they are easy to follow and therefore should be easy to modify if needed.

Thank you. Great, I'm glad they were easy to understand.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-:-MfFJcXF0n-dn14Qxbxq:brgzpbh)* status: all files reviewed, 3 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 16 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-Mf7Nzn7CILfdigXpZ7Q:-MfF8Wns9un7qbwR8KLm:bduvydx) ([raw file](https://github.com/sillsdev/web-xforge/blob/86352b9317890801ab5f555dc58552645fb7799a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L16)):*
Yes. That would have been nice to do. But if I do that, then I get error:
> Argument of type '(event: DragEvent) => void' is not assignable to parameter of type 'EventListenerOrEventListenerObject'.
>       Type '(event: DragEvent) => void' is not assignable to type 'EventListener'.

So I just let start out as an Event.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 35 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-Mf7QBE2B95Kr5eL1hLF:-MfFABUa8dIg1chFjQfA:b-ws23pk) ([raw file](https://github.com/sillsdev/web-xforge/blob/86352b9317890801ab5f555dc58552645fb7799a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L35)):*

> ```
> console.log
> ```
How about using `console.warn` here? Here and below.
</blockquote></details>

Good. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 43 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-Mf7ROoY6EzGtTq0A9JE:-MfFAOqK8smfenFPakmv:b-uytif0) ([raw file](https://github.com/sillsdev/web-xforge/blob/86352b9317890801ab5f555dc58552645fb7799a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L43)):*

> ```
> destanation
> ```
spelling
Does your IDE give you hints when you have a spelling error? In VSCode I have the Cod Spell Checker extension that uses the dictionary in our repo and lets me know of spell errors in variable names.
</blockquote></details>

Yes, I have that extension but missed that squiggle. Thank you.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 64 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-Mf7ST_z8Kf_q1DJ4FRG:-MfFBARO-TC4iyMuSlMH:b59s248) ([raw file](https://github.com/sillsdev/web-xforge/blob/86352b9317890801ab5f555dc58552645fb7799a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L64)):*

> ```
>  } else {
> ```
Will either of the above methods work in Safari?
</blockquote></details>

Yes, according to https://developer.mozilla.org/en-US/docs/Web/API/Document/caretRangeFromPoint , where Edge and Safari both have caretRangeFromPoint. But we will do well to test it. I tested it in Edge today in Linux, and have so far also tested in Chromium and Firefox. All on a desktop computer.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 154 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-Mf7VSil6X2MT33PCV3k:-MfFBlCa6laInU8u78uB:b-vuj8ut) ([raw file](https://github.com/sillsdev/web-xforge/blob/86352b9317890801ab5f555dc58552645fb7799a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L154)):*

> ```
> MockDragEvent
> ```
Good job figuring out how to test using this.
</blockquote></details>

Thank you :)

marksvc (commented):
I'm a bit confused as to why codecov still thinks the Firefox if-branch hasn't been run thru. I modified my Firefox test locally to say
```ts
     // Override the Firefox ..
     document.caretPositionFromPoint = (x: number, y: number) => ({ offset: desiredIndexInSegment } as CaretPosition);
      expect(document.caretPositionFromPoint).toBeDefined();
      // Remove the Chromium point-to-index method so the Firefox one will be used (in our Chromium test runner).
      expect(document.caretRangeFromPoint).toBeDefined();
      (document as any).caretRangeFromPoint = undefined;
      expect(document.caretRangeFromPoint).toBeUndefined();
```
and it still passes.
I put a console.log in the Firefox if-branch, and it prints.
Maybe codecov hasn't actually updated its report yet, or maybe my approach messing with `document` was too much for it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-:-MfFL0w5-Hwd0zNQnPqh:b-24y7uj)* status: 3 of 5 files reviewed, 3 unresolved discussions (waiting on @irahopkinson)

irahopkinson (commented):
Ok, that makes sense but can you tell me why we can't just clear away the text/html data when we drag text in from another window?

For codecov, the current report is based on commit 86352b9 which is 9 days old. In fact in GitHub it says:

The first question is did the TC upload new coverage data and if it did why doesn't codecov deal with it correctly? Pushing a commit (that isn't squashed to the existing commit) will likely fix this (even if its just a trivial change).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-:-MfHBImz93L1kEWo3-VH:blpx77r)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 16 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-Mf7Nzn7CILfdigXpZ7Q:-MfHEpBkCLJNxRe1zhna:bumi1ev) ([raw file](https://github.com/sillsdev/web-xforge/blob/86352b9317890801ab5f555dc58552645fb7799a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L16)):*

Yes. That would have been nice to do. But if I do that, then I get error:
> Argument of type '(event: DragEvent) => void' is not assignable to parameter of type 'EventListenerOrEventListenerObject'.
>       Type '(event: DragEvent) => void' is not assignable to type 'EventListener'.

So I just let start out as an Event.
</blockquote></details>

Right, I guess we have to cast then, what a shame.

irahopkinson (commented):
Actually now we have a new commit on master, a rebase might be all that is needed to update codecov.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-:-MfIzwDY315w0_O6OZA0:b-cu259e)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

marksvc:
> Ok, that makes sense but can you tell me why we can't just clear away the text/html data when we drag text in from another window?

Right. The reason is that we can only manipulate the data in the dragstart event. Every other event is too late (even the drop event). Note that drag and drop is implemented with some security precautions, where if you drag from A to C, and window B happens to be on the way and so you mouse over window B, window B isn't allowed to read the data being dragged. Only A and C are allowed to read the data. Why C can't _modify_ the data when it's dropped, I don't know, but it may be tied into the precautions. 

Consider https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/clearData 's statement :
> This method [clearData] can only be used in the handler for the dragstart event, because that's the only time the drag operation's data store is writeable.

Interestingly, an example shows usage of dataTransfer.clearData() within the drop handler at https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/setData .

Today I tested again by modifying drag-and-drop.ts's drop handler to clear the text/html data and then to clear all data. In the unit tests, it did do just as I asked it. But when running SF, it does not actually clear the data. I also tried commenting out preventDefault(), thus allowing the browser to do its own drag-and-drop behaviour. Despite clearing the data, the undesired formatting was coming back in from drag-and-drop. I tested by dragging within quill, and dragging content from elsewhere on the same web page into quill.

My test involved a change to the drop event in drag-and-drop.ts so it was changed to:
```ts
    quill.container.addEventListener('drop', (event: Event) => {
      const dragEvent = event as DragEvent;
      let dataPlain = dragEvent.dataTransfer?.getData('text/plain');
      let dataHtml = dragEvent.dataTransfer?.getData('text/html');
      console.log(`Before any clear, dataPlain: ${dataPlain}, dataHtml: ${dataHtml}`);
      dragEvent.dataTransfer?.clearData('text/html');
      dataPlain = dragEvent.dataTransfer?.getData('text/plain');
      dataHtml = dragEvent.dataTransfer?.getData('text/html');
      console.log(`After clearing html, dataPlain: ${dataPlain}, dataHtml: ${dataHtml}`);
      dragEvent.dataTransfer?.clearData();
      dataPlain = dragEvent.dataTransfer?.getData('text/plain');
      dataHtml = dragEvent.dataTransfer?.getData('text/html');
      console.log(`After clearing all, dataPlain: ${dataPlain}, dataHtml: ${dataHtml}`);
      // Stop the browser from doing any drag-and-drop behaviour itself, such as inserting text with formatting.
      // dragEvent.preventDefault();
      return;
```

(BTW it occurs to me that maybe `dragEvent.dataTransfer` was null/undefined, but when running SF, and breaking, I can observe that `dataTransfer` is defined.)

We can see that output during a unit test looks great for this:
```
LOG: 'Before any clear, dataPlain: chapter 1, dataHtml: <span background="white">chapter 1</span>'
LOG: 'After clearing html, dataPlain: chapter 1, dataHtml: '
LOG: 'After clearing all, dataPlain: , dataHtml: '
```

But when running SF, things don't go so well:

```
Before any clear, dataPlain: ruth, dataHtml: <span style="color: rgb(0, 0, 0); font-family: Roboto, sans-serif; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: 0.5px; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: pre-wrap; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;">ruth</span>
After clearing html, dataPlain: ruth, dataHtml: <span style="color: rgb(0, 0, 0); font-family: Roboto, sans-serif; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: 0.5px; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: pre-wrap; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;">ruth</span>
After clearing all, dataPlain: ruth, dataHtml: <span style="color: rgb(0, 0, 0); font-family: Roboto, sans-serif; font-size: 16px; font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: 0.5px; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: pre-wrap; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; background-color: rgb(255, 255, 255); text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; display: inline !important; float: none;">ruth</span>
```

And in my text I see there is formatting around the dropped text, "ruth".

I'm not sure why I was able to edit the dragEvent.dataTransfer data in the unit tests. I tried changing from using MockDragEvent to just DragEvent. I suspect it has to do with some aspect of how I manually created my DataTransfer object, or my DragEvent object, rather than those being generated by actual user behaviour in a web browser, or less likely, perhaps some aspect of our test runner. This doesn't help us have confidence that the tests are covering the behaviour we will experience in production, does it. But so far so good in other respects.

marksvc (commented):
I fiddled with codecov and plan to paste an analysis into chat.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-:-MfJfeq48LoBzgthgP1C:bs15z78)* status: 4 of 5 files reviewed, all discussions resolved (waiting on @irahopkinson)

irahopkinson (commented):
I wonder if it is possible to copy the plain text from the event, prevent its default then create a new event with only the plain text and fire that event. Would that work and then would each browser do the right thing?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1097#-:-MfLTzuf7q16v-v8uyHy:bclqybo)* status: 4 of 5 files reviewed, all discussions resolved (waiting on @irahopkinson)

irahopkinson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1097)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

marksvc:
> I wonder if it is possible to copy the plain text from the event, prevent its default then create a new event with only the plain text and fire that event. Would that work and then would each browser do the right thing?

I spent some time looking into this. There is still the potential for this to work, perhaps if we populate the new DragEvent with enough of the right information. A bunch of info on what was attempted is posted to the jira issue.

marksvc:
@irahopkinson Reviewable doesn't tell me what your requested change is; just that you have requested changes :-).   (Although it also says:
> Reviewable status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

so one would think that means the review is done and it's waiting for me to submit it or something). 
I'm assuming that any remaining todo on the review relates to the thoughts regarding DragEvents and plain/text. So I can continue investigating how we might successfully create a DragEvent.  A bunch of details from my investigation so far are in chat.

PR #1098, opened 2021-07-18
SF-902: Back Translations

pmachapman:
This adds support for back translations by providing an alternate implementation of Registry functions for projects that are not found in the registry. The non-Registry implementations provide the same information as the Registry, but are slower, so are only used for projects where they are required.

Another feature added in this PR is that the Project FullName is updated if it is changed. These are changed only on the Registry, but in the case of Back Translations which are not in the Registry, this is required, as the FullName is loaded from the ScrText object.

Note that only the short name is shown when you connect to a project - this parallels Paratext - and so the short name will be updated with the full name on first sync.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1098)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/IParatextNotesMapper.cs
+++ b/src/SIL.XForge.Scripture/Services/IParatextNotesMapper.cs
@@ -13,7 +13,7 @@ namespace SIL.XForge.Scripture.Services
         List<SyncUser> NewSyncUsers { get; }
 
         Task InitAsync(UserSecret currentUserSecret, SFProjectSecret projectSecret, List<User> ptUsers,
-            string paratextProjectId, CancellationToken token);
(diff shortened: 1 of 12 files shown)


ddaspit (approved):
:lgtm:

> Reviewed 12 of 12 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1098#-:-MfaHwv37haGZVW4IDt0:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

irahopkinson (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1098)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #1104, opened 2021-07-22
Add query for projects that are in syncing state

Nateowami:
This is a trivially simple query, but I think it's worth having in the repo so we can easily check for projects that are in the syncing state without writing the query from scratch every time.

Also, since a query like this is typically (hopefully) going to return no results, it's hard to be confident a query was written correctly, since most incorrect queries don't find any results either (in a case like this I would typically modify the query to something like `{'sync.queuedCount': {$ne: 1}}` and check that it's returning a bunch of projects). Having a known-working version in the repo is therefore simpler than writing it from scratch every time.

In the past I've seen up to six or so projects that were all stuck in a syncing state on live.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1104)


RaymondLuong3 (approved):
I didn't even know I had a mongo folder. That's useful.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1104#-:-MfZdGBA4JtFnGWqDuQy:b-w8uauf)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1125, opened 2021-08-17
SF-1364b Fix error calling isAlreadyInvited when checking has sharing disabled

Nateowami:
The condition where the error would occur is:
- Checking sharing is disabled
- Translate sharing is enabled
- User opens share dialog and types email to invite

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1125)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/SFProjectService.cs
+++ b/src/SIL.XForge.Scripture/Services/SFProjectService.cs
@@ -484,7 +484,9 @@ namespace SIL.XForge.Scripture.Services
         public async Task<bool> IsAlreadyInvitedAsync(string curUserId, string projectId, string email)
         {
             SFProject project = await GetProjectAsync(projectId);
-            if (!IsProjectAdmin(project, curUserId) && !project.CheckingConfig.ShareEnabled)
+            bool sharingEnabled = project.TranslateConfig.ShareEnabled ||
+                (project.CheckingConfig.CheckingEnabled && project.CheckingConfig.ShareEnabled);
+            if (!IsProjectAdmin(project, curUserId) && !(IsOnProject(project, curUserId) && sharingEnabled))
                 throw new ForbiddenException();
 
             if (email == null)
--- a/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs
+++ b/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs
@@ -31,6 +31,7 @@ namespace SIL.XForge.Scripture.Services
         private const string Project03 = "project03";
         private const string Project04 = "project04";
         private const string Project05 = "project05";
+        private const string Project06 = "project06";
         private const string Resource01 = "resource_project";
         private const string DisabledSource = "disabled_source";
         private const string Resource01PTId = "resid_is_16_char";
@@ -816,7 +817,7 @@ namespace SIL.XForge.Scripture.Services
 
             Assert.That(await env.Service.IsAlreadyInvitedAsync(User01, Project03, "junk"), Is.False);
 
-            Assert.That(await env.Service.IsAlreadyInvitedAsync(User03, Project02, "nothing"), Is.False);
+            Assert.That(await env.Service.IsAlreadyInvitedAsync(User02, Project02, "nothing"), Is.False);
         }
 
         [Test]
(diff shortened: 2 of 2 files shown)


marksvc (commented):
@nateowami Thanks for fixing this. This is a good place to write a test that shows the problem, and then fix it with the code change. The SUT here, with some simple `if` checks and a `bool` return will lend itself well to unit tests specifying permutations of project share settings, and the expected result of the method.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1125#-:-MhL3uPT7IlkyYF9lt-B:b-6ww3g7)* status: 0 of 1 files reviewed, all discussions resolved

Nateowami (commented):
Thanks. Done.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1125#-:-MhPe1h-E_rq48X13-kX:b-i21bbo)* status: 0 of 2 files reviewed, all discussions resolved

marksvc (approved):
I tested the behaviour by Sending and Resending invitations to an email address, with role 'View Translation' or Community checker, with different combinations of translation sharing enabled, checking enabled, checking sharing enabled. It seems to work fine.
I used a non-admin user for doing the inviting.

I have a comment for how to improve one of the tests, but basically the PR is good to go.

We mentioned only running the IsAlreadyInvited if the requesting user is a member of the project in question, which will be good to include here or we can do it as followup.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1125#-:-MhVKpWEEo4o35opeYXk:b3zt113)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs, line 863 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1125#-MhVGod59BRYQXU0zk-h:-MhVGod59BRYQXU0zk-i:b-okp107) ([raw file](https://github.com/sillsdev/web-xforge/blob/cfbffd98745f8017e742181e7cee5b67456f6295/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L863)):*
> ```Smalltalk
>         {
>             Assert.That(env.GetProject(Project04).CheckingConfig.ShareEnabled, Is.False);
> ```

Because the test data/configs may get changed in various ways as more tests are added, it would be a good idea to also assert here that TranslateConfig.ShareEnabled==true, so that we know that the situation/config we are intending to test, is in fact so.

marksvc (changes requested):
> Reviewed 1 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1125)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 489 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1125#-Mhnwgpl4SupPRIA6On7:-Mhnwgpl4SupPRIA6On8:b5peq3j) ([raw file](https://github.com/sillsdev/web-xforge/blob/7b58f41b68667d5118ab7321236662032b3eb7e6/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L489)):*
> ```
> IsOnProject(project, curUserId) 
> ```

And what about a test that asserts that IsAlreadyInvitedAsync() throws if curUserId is not on the project, when curUserId is not an admin, even if sharing is enabled?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1125)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/Services/SFProjectService.cs, line 489 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1125#-Mhnwgpl4SupPRIA6On7:-Mhpa2CY5Y8399jt2_Hz:b-9w00t0) ([raw file](https://github.com/sillsdev/web-xforge/blob/7b58f41b68667d5118ab7321236662032b3eb7e6/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L489)):*
> curUserId is not on the project, when curUserId is not an admin

I've created a test, but I'm not sure I fully understand your sentence. If the user is not on the project, then the user has no role, and specifying what their role is or isn't doesn't make sense.

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1125)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1127, opened 2021-08-24
SF-1344c Handle drag-and-drop to empty verse

marksvc:
- When dragging to an empty verse, a usx-blank element will be the
target. Use the parent element instead, which should be a usx-segment
element.
- When inserting into an empty verse, position the new text just after
the blank.

text-view-model.ts:
- In update(), call editor.updateContents() right away. Otherwise, the
`updateDelta` can contain outdated information by the time it is sent
to updateContents(), such as if multiple changes are happening to the
quill editor, such as in the case of inserting dropped text and
removing source text. Alternatively, the call to updateContents()
could continue to be deferred if we continue to call updateSegments()
earlier (to set segment boundaries) and then call updateSegments()
again, setting updateDelta, in the deferred processing so that the
updateDelta sent to updateContents() is up to date. But in testing so
far there doesn't seem to be a need to continue deferring the call to
updateContents().
- updateSegments(): Invert the containing `if` to reduce nesting.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1127)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
@@ -1,5 +1,4 @@
 import Quill, { RangeStatic } from 'quill';
-import { Delta } from 'rich-text';
 import { TextComponent } from './text.component';
 
 export interface DragAndDropOptions {
@@ -26,9 +25,28 @@ export class DragAndDrop {
       // Stop the browser from doing any drag-and-drop behaviour itself, such as inserting text with formatting.
       dragEvent.preventDefault();
 
-      if ((dragEvent.target as Element)?.localName !== 'usx-segment') {
+      let targetElement: Element | null = dragEvent.target as Element | null;
+      if (targetElement == null) {
+        console.warn(`Warning: DragEvent unexpectedly has null target.`);
+        return;
+      }
+      let targetElementName: string = targetElement.localName;
+
+      let droppingInBlankSegment: boolean = false;
+      if (targetElementName === 'usx-blank') {
+        droppingInBlankSegment = true;
+        targetElement = targetElement.parentElement as Element | null;
+        if (targetElement == null) {
+          console.warn(`Warning: DragEvent usx-blank target parent is unexpectedly null. Target:`, targetElement);
+          return;
+        }
+        targetElementName = targetElement.localName;
+      }
+
+      if (targetElementName !== 'usx-segment') {
         // We need to be able to know where to insert the dropped text, such as from the drop target being a usx-segment
-        // element.
+        // element. Give up.
+        console.warn('Warning: DragEvent to invalid target:', targetElement);
         return;
       }
       if (dragEvent.dataTransfer == null) {
@@ -39,17 +57,32 @@ export class DragAndDrop {
       // Determine where we should be placing the dropped text, using the location of the destination segment, and the
       // index within that segment.
 
-      const destinationSegmentRef: string = (dragEvent.target as Element).attributes['data-segment'].value;
+      const dataSegmentAttribute: any = targetElement.attributes['data-segment'];
+      if (dataSegmentAttribute == null) {
+        console.warn('Warning: drag-and-drop to target with no segment ref attribute:', targetElement);
+        return;
+      }
+      const destinationSegmentRef: string = dataSegmentAttribute.value;
       const destinationSegmentRange: RangeStatic | undefined =
         options.textComponent.getSegmentRange(destinationSegmentRef);
       if (destinationSegmentRange == null) {
-        console.warn('Warning: drag-and-drop destination segment had an undefined range.');
+        console.warn(
+          'Warning: No defined range for drag-and-drop destination segment. Invalid segment specification? Destination segment ref:',
+          destinationSegmentRef
+        );
+
         return;
       }
 
       // Determine character index of drop location in destination segment, using a browser-specific method.
       let startPositionInSegment: number = 0;
-      if (document.caretRangeFromPoint !== undefined) {
+      if (droppingInBlankSegment) {
+        // If we are dropping into an empty segment, use the position at the end of the segment rather than using a
+        // browser-determined index into the segment.
+        // Use the position at the end of the segment, rather than at the beginning, so that it is after the blank and
+        // any other embeds.
+        startPositionInSegment = destinationSegmentRange.length;
+      } else if (document.caretRangeFromPoint !== undefined) {
         // Chromium/Chrome, Edge, and Safari browsers
         const range: Range = document.caretRangeFromPoint(dragEvent.clientX, dragEvent.clientY);
         startPositionInSegment = range.startOffset;
@@ -57,24 +90,20 @@ export class DragAndDrop {
         // Firefox browser
         const range: CaretPosition | null = document.caretPositionFromPoint(dragEvent.clientX, dragEvent.clientY);
         if (range == null) {
-          console.warn('Warning: drag-and-drop inferred a null caret position for insertion.');
+          console.warn('Warning: drag-and-drop inferred a null caret position for insertion. Target:', targetElement);
           return;
         }
         startPositionInSegment = range.offset;
       } else {
-        console.warn(`Warning: Could not determine insertion position for drag-and-drop.`);
+        console.warn(`Warning: Could not determine insertion position for drag-and-drop. Target:`, targetElement);
         return;
       }
 
-      const startingPositionInDocument: number = destinationSegmentRange.index + startPositionInSegment;
-
-      const insertionDelta = new Delta();
-      insertionDelta.retain(startingPositionInDocument);
+      const insertionPositionInDocument: number = destinationSegmentRange.index + startPositionInSegment;
       let newText: string = dragEvent.dataTransfer.getData('text/plain');
-      // Omit newline characters
+      // Replace newlines with a space.
       newText = newText.replace(/(?:\r?\n)+/g, ' ');
-      insertionDelta.insert(newText);
-      quill.updateContents(insertionDelta, 'user');
+      quill.insertText(insertionPositionInDocument, newText, 'user');
 
       // Identify the selection range, if any, now that we updated the document with an insert. This will be a
       // selection that was already present before the drop, whether it is the text that was dragged, or a selection
@@ -82,17 +111,14 @@ export class DragAndDrop {
       const originalSelection: RangeStatic | null = quill.getSelection();
 
       // Select the inserted text.
-      quill.setSelection(startingPositionInDocument, newText.length);
+      quill.setSelection(insertionPositionInDocument, newText.length);
 
       if (originalSelection != null) {
         // There was a selection before the drop occurred.
-        const removalDelta = new Delta();
-        removalDelta.retain(originalSelection.index);
         if (dragEvent.dataTransfer.types.includes(DragAndDrop.quillIsSourceToken) && !dragEvent.ctrlKey) {
           // If the drag was started from within quill, then treat the selection as the source data of the drag, and
           // delete the selection. Unless the user was holding the ctrl key to copy text instead of move it.
-          removalDelta.delete(originalSelection.length);
-          quill.updateContents(removalDelta, 'user');
+          quill.deleteText(originalSelection.index, originalSelection.length, 'user');
         }
         // Or if the drag was not started from within quill, or the user was holding the ctrl key, then don't delete
         // the selection.
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -149,7 +149,8 @@ export class TextViewModel {
   }
 
   /**
-   * Updates the view model and Quill contents when text is changed.
+   * Updates the view model (textDoc), segment ranges, and slightly the Quill contents, such as in response to text
+   * changing in the quill editor.
    *
    * @param {DeltaStatic} delta The view model delta.
    * @param {Sources} source The source of the change.
@@ -160,6 +161,7 @@ export class TextViewModel {
       return;
     }
 
+    // The incoming change already happened in the quill editor. Now apply the change to the view model.
     if (source === 'user' && editor.isEnabled()) {
       const modelDelta = this.viewToData(delta);
       if (modelDelta.ops != null && modelDelta.ops.length > 0) {
@@ -167,11 +169,17 @@ export class TextViewModel {
       }
     }
 
-    const updateDelta = this.updateSegments(editor);
-    if (updateDelta.ops != null && updateDelta.ops.length > 0) {
-      // defer the update, since it might cause the segment ranges to be out-of-sync with the view model
-      Promise.resolve().then(() => editor.updateContents(updateDelta, source));
-    }
+    // Re-compute segment boundaries so the insertion point stays in the right place.
+    this.updateSegments(editor);
+
+    // Defer the update, since it might cause the segment ranges to be out-of-sync with the view model
+    Promise.resolve().then(() => {
+      const updateDelta = this.updateSegments(editor);
+      if (updateDelta.ops != null && updateDelta.ops.length > 0) {
+        // Clean up blanks in quill editor. This may result in re-entering the update() method.
+        editor.updateContents(updateDelta, source);
+      }
+    });
   }
 
   highlight(segmentRefs: string[]): void {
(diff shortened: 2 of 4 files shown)


marksvc:
@ddaspit  I have a question about a comment in text-view-model.ts. In text-view-model.ts there is a method called update() which receives a delta, applies it to the textDoc, computes segment boundaries and blank fixups, and then schedules an editor change to apply the blank fixups (such as removing a blank if a verse is no longer empty). Deferring this editor change can result in incorrect behaviour if something else updates the editor after blank-fixup ops are created (from `const updateDelta = this.updateSegments(editor);`) and before they are applied to the quill editor (in `Promise.resolve().then(() => editor.updateContents(updateDelta, source));`). (For example, if drag-and-drop.ts does more than one call to `quill.updateContents()`.) 

There are various ways to go about remedying this. One is to continue to call updateSegments() before the Promise.resolve() line, to continue to update segment boundaries right away so that the insertion point stays in the right place. And then to call updateSegments() again inside the Promise.resolve().then(), so that the `updateDelta` has up-to-date information when it's passed to `editor.updateContents()`. 

I chose to go about fixing the situation by not deferring the call to `editor.updateContents()`.  So far in my testing, this has been fine, but there is a message there explaining that 
> // defer the update, since it might cause the segment ranges to be out-of-sync with the view model

Can you tell me how I might go about finding out if this will be a problem? I'm also a bit confused as to how this out-of-sync situation would happen. Or maybe there was the potential for a problem but the codebase has changed since then. So far it's been fine in my limited testing, but maybe a particular situation needs to trigger it getting out of sync. Thanks for any tips or explanation you can give.

marksvc:
Well these failing tests in editor.component.spec.ts may be showing why we need to defer calling updateContents. I will need to look at this further. 
I will bring this down to a draft.

marksvc:
Un-drafting - Ready for review.

RaymondLuong3 (changes requested):
Thanks for working through this drag-and-drop mystery. I have a few suggestions, but for the most part, it looks good.

> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-:-MitL05L0MHgFSH9Lfk7:b4f2dzo)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 73 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitC_jB283mNL2sGDMo:-MitC_jB283mNL2sGDMp:b67v6rb) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L73)):*
> ```
>         startPositionInSegment = 1;
> ```
Could you adjust this to drop at the end of the segment instead? For instance, get the index of the segment, and drop it at the start index + segment length. Once Paratext notes get added, position in segment will not necessarily be 1.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 365 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitEkvZ1ndD0iUH0XrB:-MitEkvZ1ndD0iUH0XrC:bopxa0r) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L365)):*
> ```
>     if (delta.ops == null) {
>       return convertDelta.compose(fixDelta).chop();
>     }
> ```
Could you simply just return convertDelta? Since an empty delta combined with an empty delta would still be an empty delta?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 474 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitF8q-8JLW8BR0PYOs:-MitF8q-8JLW8BR0PYOt:b4rilk4) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L474)):*
> ```
>   /** Adds to fixDelta some cleaning up, such as whether there is a blank. */
> ```
Maybe a better comment would be computes and returns a delta that adds or removed the blank spacing on blank segments, or something similar.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 672 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitK8lD18TKWMYWGr6d:-MitK8lECE6bJcUDlFmG:bhcpcmo) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L672)):*
> ```
>    // End drag-and-drop section of tests.
> ```
It should be easy to also check that the expected text is in verse 2, and the leftover text is in verse 1.

nigel-wells (changes requested):
> Reviewed 3 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1127)* status: 3 of 4 files reviewed, 7 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 173 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitDofE-eoI-S2TB_BW:-MitDofE-eoI-S2TB_BX:bazv1vv) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L173)):*
> ```XML
> 
>     // Re-compute segment boundaries so the insertion point stays in the right place.
>     this.updateSegments(editor);
> ```

Does doing this then affect the deferred one in terms of no updated ops would be returned any more to update quill (and therefor  the text doc)?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 372 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitEQ3q0Y0gKSPU3091:-MitEQ3q0Y0gKSPU3092:bxld6t) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L372)):*
> ```XML
>     let curIndex = 0;
>     let curSegment: SegmentInfo | undefined;
>     for (const op of delta.ops) {
> ```

I assume the only actual change below was the removal of the if null check which you moved above?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 586 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitFbSV83JfGb5tL_QL:-MitFbSV83JfGb5tL_QM:b3m7bzx) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L586)):*
> ```XML
>       // Set up elements with a structure that mimics production.
>       // <usx-para-contents>
>       const usxParaContents: HTMLElement = document.createElement('usx-para-contents');
> ```

Was the reason for setting the elements outside of quill because it isn't possible to select text and then emulate the drag?

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1127)* status: 1 of 4 files reviewed, 5 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 73 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitC_jB283mNL2sGDMo:-MjG4CBR0jHJmfsgW_VZ:b-82fzpm) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L73)):*

> ```
>         startPositionInSegment = 1;
> ```
Could you adjust this to drop at the end of the segment instead? For instance, get the index of the segment, and drop it at the start index + segment length. Once Paratext notes get added, position in segment will not necessarily be 1.
</blockquote></details>

Thank you. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 173 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitDofE-eoI-S2TB_BW:-MjG8R0m8uUh7zzWzzqC:b-yuqj0r) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L173)):*

Does doing this then affect the deferred one in terms of no updated ops would be returned any more to update quill (and therefor  the text doc)?
</blockquote></details>

No. But there are side effects to calling updateSegments(). I've been studying this and doing some testing, and although it seems like something might go wrong, in practice it seems to be okay. I also don't entirely understand some of the workings. I am recording lots more analysis in a jira comment of what I tested and what differences I found.

With respect to specifically the output of the initial call and deferred call to updateSegments(), we do lose some information but it's not information we want to keep. Having different return values is good because the drag-and-drop.ts code is causing more than one update, and by the time the deferred promise executes, I don't want it to do what was calculated from the first call to updateSegments(), but what was _just_ calculated by the deferred call to updateSegments().

If the code in text-view-model.ts update() should stay as is, I may be able to still get what I want by having drag-and-drop.ts combine its two edits into one edit. But in my opinion, the operation of text-view-model.ts update() (prior to this PR) was incorrect. I may also be using the library wrong, if I'm supposed to pause between each edit. 
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 365 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitEkvZ1ndD0iUH0XrB:-MjWOCLa5PuG-EUvJ8XU:b6rzo60) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L365)):*

> ```
>     if (delta.ops == null) {
>       return convertDelta.compose(fixDelta).chop();
>     }
> ```
Could you simply just return convertDelta? Since an empty delta combined with an empty delta would still be an empty delta?
</blockquote></details>

Yes. I looked up compose and chop today to be sure of what they do. Thank you. 
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 372 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitEQ3q0Y0gKSPU3091:-MjWKhSn77NiX42nRhxf:b-sxx2wv) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L372)):*

I assume the only actual change below was the removal of the if null check which you moved above?
</blockquote></details>

Yes, that's my intention. 
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 474 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitF8q-8JLW8BR0PYOs:-MjWKBcH0sLCCrvX6T6f:bkacvgc) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L474)):*

> ```
>   /** Adds to fixDelta some cleaning up, such as whether there is a blank. */
> ```
Maybe a better comment would be computes and returns a delta that adds or removed the blank spacing on blank segments, or something similar.
</blockquote></details>

Thank you. I don't want to define it too narrowly since it also can add `{initial:true}` and adjust segment refs. How about this change?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 586 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitFbSV83JfGb5tL_QL:-MjWClCP3hwsE5buJ_7f:b-75qq0a) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L586)):*

Was the reason for setting the elements outside of quill because it isn't possible to select text and then emulate the drag?
</blockquote></details>

Thank you. It is selecting text and simulating the drag.  I wanted to have a DOM structure that the implementation code could work with to get the right result and it didn't occur to me to _use the DOM that quill had constructed_. Now it is.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 672 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitK8lD18TKWMYWGr6d:-MjW-3C43dim_rTc3xLr:b-yzp6vc) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L672)):*

> ```
>    // End drag-and-drop section of tests.
> ```
It should be easy to also check that the expected text is in verse 2, and the leftover text is in verse 1.
</blockquote></details>

Done. Ah ha and it exposed a problem in the test :)

irahopkinson (commented):
Can you do anything about the **Misses** and the **Partials** as reported by Codecov so that we are not losing coverage? You should be able to drill down on the Codecov report to find them. To point out the perhaps obvious. You have added 12 lines of code and covered 6 of them with tests (Hits), 5 lines don't have coverage (Misses) and one line is partially covered (usually this is a conditional/`if` statement).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-:-MjgIhC5AmQhxTElGFbP:bb29ypk)* status: 1 of 4 files reviewed, 5 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

RaymondLuong3 (commented):
It looks good once Ira's comment is addressed.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-:-MjjEg244bcDIu5g0fSp:b-8jo6n8)* status: all files reviewed, 2 unresolved discussions (waiting on @nigel-wells)

nigel-wells (approved):
Thanks @marksvc , I've commented on my points and marked them as satisfied :)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-:-MjkSu4WBlYpAA42FvJI:bz2kn72)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 173 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitDofE-eoI-S2TB_BW:-MjkSIFs2vwlxhe3zSez:b-beexkj) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L173)):*

No. But there are side effects to calling updateSegments(). I've been studying this and doing some testing, and although it seems like something might go wrong, in practice it seems to be okay. I also don't entirely understand some of the workings. I am recording lots more analysis in a jira comment of what I tested and what differences I found.

With respect to specifically the output of the initial call and deferred call to updateSegments(), we do lose some information but it's not information we want to keep. Having different return values is good because the drag-and-drop.ts code is causing more than one update, and by the time the deferred promise executes, I don't want it to do what was calculated from the first call to updateSegments(), but what was _just_ calculated by the deferred call to updateSegments().

If the code in text-view-model.ts update() should stay as is, I may be able to still get what I want by having drag-and-drop.ts combine its two edits into one edit. But in my opinion, the operation of text-view-model.ts update() (prior to this PR) was incorrect. I may also be using the library wrong, if I'm supposed to pause between each edit. 


</blockquote></details>

Yes I understand now the approach. My initial thought was we're discarding ops from  updateSegments() but because you don't tell Quill it then still returns the equivalent ops when the deferred call is made. I'm ok with two separate calls being made for an insertion and then delete - looking at the history it will look like a cut and paste than a single drag n drop op but I don't think that matters.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 586 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitFbSV83JfGb5tL_QL:-MjkM18f8DzCvRRqeqpH:b-nkkj6z) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L586)):*

Thank you. It is selecting text and simulating the drag.  I wanted to have a DOM structure that the implementation code could work with to get the right result and it didn't occur to me to _use the DOM that quill had constructed_. Now it is.
</blockquote></details>

Thanks. My concern was writing a heap of code outside of component assuming that the component would always produce the same code over time. Now that you're using the DOM from Quill we'll better know if this test still passes after a potential update to Quill in the future.

marksvc (commented):
I added to the console.warn lines some additional information.

I was creating elements on the fly as part of the testing (`document.createElement'usx-segment');) and have changed that to grab the element out of the DOM.

I removed a useless setup check a couple times:
```ts
expect(env.component.editor!.getText()).toEqual(originalAllText, 'setup');
```

@irahopkinson : My understanding is that 4 of the introduced misses are early null returns, like
```ts
someVariable = fetchIt();
if (someVariable == null) {
  console.warn(`Warning: This was null for some reason.`);
  return;
}
anotherVariable = someVariable.neatProperty;
```
where I can check the fields of `someVariable` with a guarantee that it's not null. I'm not certain if you were wanting me to include these in the test suite, but I have added to the tests so it now runs into these unexpected situations, and asserts that there was no insertion or deletion, and that the selection is unchanged.

The 5th introduced miss is from text-view-model.ts where I refactored from 
```ts
if (thing != null) {
  // here a bunch of code with many nested if and for.
}
return empty;
```
to reduce at least one level of nesting:
```ts
if (thing == null) {
  return empty;
}
// here a bunch of code with many nested if and for
```

The original `if (thing == null)` was originally a yellow partial. My refactor just exposes what wasn't tested before :) 
If I get a chance today I'll plan to see if I can get a test to bump into this if it isn't too difficult.

And the introduced Partial will probably go away when I push my update.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-:-Mk49FCt6tgbwLRyBJ1S:bs3wghr)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 586 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-MitFbSV83JfGb5tL_QL:-Mk4GOAX4mC9ZApd5uHq:b4c9lyr) ([raw file](https://github.com/sillsdev/web-xforge/blob/3040b965eb77b9699c13adaab30175254ab2416c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L586)):*

Thanks. My concern was writing a heap of code outside of component assuming that the component would always produce the same code over time. Now that you're using the DOM from Quill we'll better know if this test still passes after a potential update to Quill in the future.
</blockquote></details>

:thumbsup:

marksvc:
@RaymondLuong3 The PR thinks you still want a change. Can you point out which that is, or clear it if it's taken care of? Thanks!

RaymondLuong3 (approved):
Ok, it looks good to go. It would have been enough for me just to explain that the misses were the null checks, but thanks for the additional tests.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1127#-:-Mk4PJhAC5Tm-KKSInYa:b-ffrwc1)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1134, opened 2021-09-01
Fix incorrect message saying no import questions are available

Nateowami:
If the user clicks the import button before the list of books has been loaded, then the dialog will incorrectly state that no questions are available for import. This is because we only list questions that are for books that are available in the project.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1134)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts
@@ -30,6 +30,7 @@ import { BehaviorSubject, of } from 'rxjs';
 import { anything, instance, mock, resetCalls, verify, when } from 'ts-mockito';
 import { AuthService } from 'xforge-common/auth.service';
 import { BugsnagService } from 'xforge-common/bugsnag.service';
+import { RealtimeQuery } from 'xforge-common/models/realtime-query';
 import { NoticeService } from 'xforge-common/notice.service';
 import { PwaService } from 'xforge-common/pwa.service';
 import { TestRealtimeModule } from 'xforge-common/test-realtime.module';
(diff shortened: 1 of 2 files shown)


irahopkinson (changes requested):
Thanks for this fix; it looks good.

IIRC I likely broke this with the change to line 46 (which you just correctly reverted) when "fixing" files after I made our compiler stricter. I was happy to make the change because it didn't make any tests fail. To improve our quality and stability we really do want to cover features like this with tests. This kind of fix is a dead ringer for TDD, i.e. write the test first to prove that its currently failing, then add your fix and the tests should pass.

Since I broke it I'm happy to write the tests if you'd like but I won't be able to get them until at least Friday.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1134#-:-MiUVp5o4-Uo47hKfI-y:bdm8bkf)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1134)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts, line 279 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1134#-Mii4Tqd84SCkRCSOpl1:-Mii4Tqd84SCkRCSOpl2:b-rdi4tk) ([raw file](https://github.com/sillsdev/web-xforge/blob/2558983d26b9ed09d9b884ce32ad6db29562c44f/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts#L279)):*
> ```XML
>     }));
> 
>     it('should not show import questions button until list of texts have loaded', fakeAsync(() => {
> ```

Turns out I was overly optimistic about being able to test this without actually reproducing the issue in the tests (my test passed on master as well). However, this test actually gets to the heart of the issue and fails on master.

irahopkinson (approved):
> Reviewed 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1134)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts, line 279 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1134#-Mii4Tqd84SCkRCSOpl1:-MiiK_YXCDR36Y-KlZu3:b-v7gob2) ([raw file](https://github.com/sillsdev/web-xforge/blob/2558983d26b9ed09d9b884ce32ad6db29562c44f/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts#L279)):*

Turns out I was overly optimistic about being able to test this without actually reproducing the issue in the tests (my test passed on master as well). However, this test actually gets to the heart of the issue and fails on master.
</blockquote></details>

Well done!

PR #1144, opened 2021-09-13
Allow canceling question import

Nateowami:
Allowing for canceling seems like a good idea overall, but it seemed extra important since the import is super slow and gets slower as more are imported. Since that's a harder problem to solve, I've started by making it easy to cancel (which only stops importing; it doesn't reverse those that are already imported).

Here's how the dialog now looks during import:

And here's how it looks once the cancel button is clicked:

This is shown until the question currently being imported is done being imported, and then the dialog disappears.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1144)


irahopkinson (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1144)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1150, opened 2021-09-25
Add script to delete doc and restore to past version

Nateowami:
This script was tested locally and then used to fix the data corruption on live.

The utils file provides supporting code for multiple scripts. Some of the changes to the utils file aren't really relevant to this particular commit, but it seemed pointless to try to revert those specific portions before committing. There will probably be future PRs where I add scripts that rely on some of these changes.

### DIFF
--- a/scripts/db_tools/revert-doc.js
+++ b/scripts/db_tools/revert-doc.js
@@ -1,0 +1,46 @@
+#!/usr/bin/env node
+
+// Delete a doc (though it still exists in the DB), then recreate it in a state that it was in at a
+// previous version
+
+const utils = require('./utils.js');
+const RichText = utils.requireFromRealTimeServer('rich-text');
+const ShareDB = utils.requireFromRealTimeServer('sharedb/lib/client');
+const WebSocket = utils.requireFromRealTimeServer('ws');
+
+// Edit these settings to specify which doc to revert
+const docId = '';
+const connectionConfig = utils.devConfig;
+const resetToVersion = 0;
+const collection = 'texts';
+const dryRun = true;
+
+ShareDB.types.register(RichText.type);
+
+async function run() {
+  const ws = new WebSocket(connectionConfig.wsConnectionString);
+  const conn = new ShareDB.Connection(ws);
+  try {
+    const doc = conn.get(collection, docId);
+    await utils.fetchDoc(doc);
+
+    const revertToSnapshot = await utils.fetchSnapshotByVersion(conn, collection, docId, resetToVersion);
+
+    if (!dryRun) {
+      await utils.deleteDoc(doc);
+      console.log('deleted doc');
+    }
+
+    if (dryRun) {
+      console.log('would reset doc to:');
+      console.log(revertToSnapshot.data.ops);
+    } else {
+      await utils.createDoc(doc, revertToSnapshot.data.ops, RichText.type.name);
+      console.log('recreated doc');
+    }
+  } finally {
+    conn.close();
+  }
+}
+
+run();
--- a/scripts/db_tools/utils.js
+++ b/scripts/db_tools/utils.js
@@ -2,6 +2,54 @@ function requireFromRealTimeServer(package) {
   return require('../../src/RealtimeServer/node_modules/' + package);
 }
 
+function fetchDoc(doc) {
+  return new Promise((resolve, reject) => {
+    doc.fetch(err => {
+      if (err != null) {
+        reject(err);
+      } else {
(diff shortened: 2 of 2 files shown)


irahopkinson (changes requested):
Thanks for this. ToDo: a request and a question.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1150#-:-MkQUa1I0wZW7K_Dmnve:b-ehpn2t)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[scripts/db_tools/utils.js, line 5 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1150#-MkQQt3n2ttkon0zoYhU:-MkQQt3n2ttkon0zoYhV:b-q82jym) ([raw file](https://github.com/sillsdev/web-xforge/blob/be28f9eed7f8cf9b143891c8bdd3618680bb0312/scripts/db_tools/utils.js#L5)):*
> ```
> fetchDoc
> ```
Given this is JS rather than TS, can you add JSDoc comments to each function here so we can see the types? Use `unknown` instead of `any` if that is what they are.
___
*[scripts/db_tools/utils.js, line 99 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1150#-MkQRdqN7KyBgty4-mKd:-MkQRdqN7KyBgty4-mKe:b-l1gly8) ([raw file](https://github.com/sillsdev/web-xforge/blob/be28f9eed7f8cf9b143891c8bdd3618680bb0312/scripts/db_tools/utils.js#L99)):*
> ```
> module.exports = {
> ```
This is the old style export. Use the new ES Module compatible export on each function. Err... but perhaps that doesn't work with MongoDB Playground?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1150)* status: all files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @Nateowami)

___
*[scripts/db_tools/utils.js, line 5 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1150#-MkQQt3n2ttkon0zoYhU:-MkeDQ6V50mia7ODxxXD:b-cfdele) ([raw file](https://github.com/sillsdev/web-xforge/blob/be28f9eed7f8cf9b143891c8bdd3618680bb0312/scripts/db_tools/utils.js#L5)):*
Will do.

> Use unknown instead of any if that is what they are.

I don't think `unknown` makes any logical sense in this context. It doesn't mean that developer does not know. It means that the code cannot make any assumptions about the value. Unless I'm misunderstanding what you're referring to?
___
*[scripts/db_tools/utils.js, line 99 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1150#-MkQRdqN7KyBgty4-mKd:-MkeBfvz08Ww8J1UaN_N:b-srgnzk) ([raw file](https://github.com/sillsdev/web-xforge/blob/be28f9eed7f8cf9b143891c8bdd3618680bb0312/scripts/db_tools/utils.js#L99)):*

> ```
> module.exports = {
> ```
This is the old style export. Use the new ES Module compatible export on each function. Err... but perhaps that doesn't work with MongoDB Playground?
</blockquote></details>

This isn't the MongoDB playground. It's intended to be run by Node (well, the scripts that import from here are run by Node).

Node doesn't support running a script as a module out of the box. Supposedly if you name the file with `.mjs` and run node with something like `node --experimental-modules filename.mjs` it should work, with some versions of Node. But I couldn't get it to work.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1150)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @irahopkinson)

___
*[scripts/db_tools/utils.js, line 5 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1150#-MkQQt3n2ttkon0zoYhU:-MkeV4CH4Kmf1OSGy2BC:b-2tgsrc) ([raw file](https://github.com/sillsdev/web-xforge/blob/be28f9eed7f8cf9b143891c8bdd3618680bb0312/scripts/db_tools/utils.js#L5)):*

Will do.

> Use unknown instead of any if that is what they are.

I don't think `unknown` makes any logical sense in this context. It doesn't mean that developer does not know. It means that the code cannot make any assumptions about the value. Unless I'm misunderstanding what you're referring to?
</blockquote></details>

Done. I don't think the comments above regarding `unknown` actually ended up being relevant in this case.

PR #1160, opened 2021-10-05
SF-1397 Notes reappear if the user deletes the icon

RaymondLuong3:
* update the note text anchor if deletion involves the note
* re-embed a note after it is deleted in the editor
* account for the note in the delete op in the text view model
* use the segment to cache the embeds info prior to text updates
* skip over blank segments if text anchor is in subsequent segment
* do not update segments if switching texts
* do not update text anchor if deletion is caused by editor logic


nigel-wells (commented):
@raymondluong3 I've had a brief look over it. I do like that we're considering any embed rather than just notes to future proof things. My initial thoughts, before I started looking at the PR, was it could be tackled by examining what is included in the delete op and then changing the op to a mixture of delete, retain, delete rather than delete it and then insert back in all the embeds that were originally there along with subscribed events. For example, if the segment was:  **abcde:black_flag:fghi**. If a delete was run on that whole segment then initial delete op would be 10 including the flag embed. We could then change that to be delete 5 (abcde), retain 1 (flag embed), delete 4 (fghi) and thus leave the embed alone in Quill. It may not be as straightforward as that but it would simplify things if it was possible. Thoughts?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1160#-:-MlNXn1--0tWcjyVs5Y-:b-egedrf)* status: 0 of 7 files reviewed, all discussions resolved

RaymondLuong3 (commented):
Hey Nigel, I did some experimenting, and it appears that it is more complex. The user could have a selection of text highlighted, and if they type a character, we would know initial selection of text, but there is no way of knowing what gets inserted (to my limited knowledge). We would need to know the insert text so that we can apply to op to the quill contents along with the deletion ops.
Any thoughts on this?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1160#-:-MlWJeI499PfeOfPJbEt:b-vnadf7)* status: 0 of 7 files reviewed, all discussions resolved

PR #1162, opened 2021-10-06
Clarify areas around PT anchoring

marksvc:
- When I examined code about PT anchoring, it took me a bit to catch
on to some concepts and terminology. In this commit I have some
changes that I think will make some of these concepts more quickly
evident, such as by comments and renamed methods and variables.

editor.component.ts:
- I recognize that in getUpdatedTextAnchor(), the `embedIndex`
variable comes from `const [threadId, embedIndex] of
oldSegmentEmbeds.entries()` in updateSegmentNoteThreadAnchors(), and
therefore we won't expect `oldVerseEmbedPositions.values()` not to
contain it. But I propose that we have a warning for the exceptional
unexpected case of embedIndex not being in the set, at least for now,
as we get this positioning working.

ParatextServiceTests.cs:
- I think "related verse" can be a misleading string for the text of a
segment that is related to the base segment of a verse, and adjusted
the string.

ParatextSyncRunnerTests.cs:
- Accepting parameter "bookId" in `ContainsNote()` gives the wrong
idea that the method is limiting the lookup to only notes of that
book. Be evident that the book is not in the consideration.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1162)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -57,6 +57,10 @@ export interface TextUpdatedEvent {
   oldSegmentEmbeds?: Map<string, number>;
 }
 
+/**
+ * Info to annotate and draw attention to a verse using a distinguishing mark, such as a community checking question or
+ * a note.
+ */
 export interface FeaturedVerseRefInfo {
   verseRef: VerseRef;
   id: string;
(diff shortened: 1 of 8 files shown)


nigel-wells (approved):
I only made non-blockable comments - looks like it was ready to merge. I've gone ahead and closed mine off anyway. :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1162#-:-Mlv3iTd6KLeFDW56S6P:b8bwarf)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1168, opened 2021-10-14
SF-1404 Use GET requests to check permissions from DBL

Nateowami:
- HEAD requests we previously used to check permissions for a resource are no longer working properly on the DBL QA server, and soon will not work on the live DBL server.
- The `/resources` endpoint on the DBL server is adding an `id` filter query so it is possible to fetch the metadata of just one resource.

So instead of a HEAD request to `server_address/api/resource_entries/the_resource_id`, a GET request is sent to `server_address/api/resource_entries?id=the_resource_id`. At present the live server does not recognize the `id` query and just ignores it, resulting in a large payload. This is inefficient. However, it *will* work because we look at the data that is returned and see if it includes the resource id we were looking for. Once this is deployed the DBL server can be updated and the `resource_entries` endpoint with an `id` query parameter will start to only return the data for that one resource.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1168)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -30,6 +30,7 @@
     "caniuse",
     "cdnjs",
     "dropdown",
+    "ethnologue",
     "flexbox",
     "gtag",
     "helphero",
--- a/src/SIL.XForge.Scripture/Services/SFInstallableDblResource.cs
+++ b/src/SIL.XForge.Scripture/Services/SFInstallableDblResource.cs
@@ -220,12 +220,19 @@ namespace SIL.XForge.Scripture.Services
                 throw new ArgumentNullException(nameof(restClientFactory));
             }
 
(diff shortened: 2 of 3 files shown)


irahopkinson (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1168)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

RaymondLuong3 (approved):
Nice. I haven't tested it, but all is good with me.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1168#-:-MmL7XO-D51brKsfBpnG:b-868qjn)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1172, opened 2021-10-25
SF-1413 Notes respond to realtime remote edits

RaymondLuong3:
* detect when an update is local
* apply remote updates to current editor content
* update adjustDeltaForEmbeddedElements to handle converting to text doc and from text doc
* move method for getting the editor position to text view model

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1172)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -136,7 +136,7 @@ export class TextViewModel {
     editor.history.clear();
     if (subscribeToUpdates) {
       this.remoteChangesSub = this.textDoc.remoteChanges$.subscribe(ops => {
-        const deltaWithEmbeds: DeltaStatic = this.adjustDeltaForEmbeddedElements(ops as DeltaStatic, false);
+        const deltaWithEmbeds: DeltaStatic = this.addEmbeddedElementsToDelta(ops as DeltaStatic);
         editor.updateContents(deltaWithEmbeds);
       });
     }
@@ -407,7 +407,7 @@ export class TextViewModel {
         (modelDelta as any).push(modelOp);
       }
       // Remove Paratext notes from model delta
-      modelDelta = this.adjustDeltaForEmbeddedElements(modelDelta, true);
+      modelDelta = this.removedEmbeddedElementsFromDelta(modelDelta);
     }
     return modelDelta.chop();
   }
@@ -593,33 +593,28 @@ export class TextViewModel {
     return this.editor;
   }
 
-  /** Adjust the delta that is applied to the text doc by stripping off the embedded elements displayed in quill. */
-  private adjustDeltaForEmbeddedElements(modelDelta: DeltaStatic, removeEmbeds: boolean): DeltaStatic {
+  /**
+   * Strip off the embedded elements displayed in quill from the delta. This can be used to convert a delta from
+   * user edits to apply to a text doc.
+   */
+  private removedEmbeddedElementsFromDelta(modelDelta: DeltaStatic): DeltaStatic {
     if (modelDelta.ops == null || modelDelta.ops.length < 1) {
       return new Delta();
     }
     const adjustedDelta = new Delta();
     let curIndex: number = 0;
-    let embedsUpToIndex: number = 0;
     for (const op of modelDelta.ops) {
       let cloneOp: DeltaOperation | undefined = cloneDeep(op);
-      const editorStartPos: number = removeEmbeds ? curIndex : curIndex + embedsUpToIndex;
       if (cloneOp.retain != null) {
-        // If adjusting the delta by removing embeds, editorStartPos is the current index, otherwise
-        // editorStartPos must be the current index plus the number of embeds previous
-        const embedsInRange: number = this.getEmbedsInRange(editorStartPos, cloneOp.retain, removeEmbeds);
-        embedsUpToIndex += embedsInRange;
+        const embedsInRange: number = this.getEmbedsInEditorRange(curIndex, cloneOp.retain);
         curIndex += cloneOp.retain;
-        // add or remove from the retain op the number of embedded elements contained in its content
-        const adjustment: number = removeEmbeds ? embedsInRange * -1 : embedsInRange;
-        cloneOp.retain += adjustment;
+        // remove from the retain op the number of embedded elements contained in its content
+        cloneOp.retain -= embedsInRange;
       } else if (cloneOp.delete != null) {
-        const embedsInRange: number = this.getEmbedsInRange(editorStartPos, cloneOp.delete, removeEmbeds);
+        const embedsInRange: number = this.getEmbedsInEditorRange(curIndex, cloneOp.delete);
         curIndex += cloneOp.delete;
-        embedsUpToIndex += embedsUpToIndex;
-        // add or remove from the delete op the number of embedded elements contained in its content
-        const adjustment: number = removeEmbeds ? embedsInRange * -1 : embedsInRange;
-        cloneOp.delete += adjustment;
+        // remove from the delete op the number of embedded elements contained in its content
+        cloneOp.delete -= embedsInRange;
         if (cloneOp.delete < 1) {
           cloneOp = undefined;
         }
@@ -633,25 +628,67 @@ export class TextViewModel {
     return adjustedDelta;
   }
 
-  private getEmbedsInRange(startIndex: number, length: number, fromEditorContent: boolean): number {
-    const indices: number[] = Array.from(this._embeddedElements.values());
-    // The range of content that this retain applies to
-    const opEndIndex: number = startIndex + length;
-    if (fromEditorContent) {
-      // subtract embeds from range
-      let embeddedElementsCount: number = 0;
-      for (const embedIndex of indices) {
-        if (embedIndex < startIndex) {
-          continue;
-        } else if (embedIndex >= startIndex && embedIndex < opEndIndex) {
-          embeddedElementsCount++;
-        } else {
-          break;
+  /**
+   * Add in the embedded elements displayed in quill to the delta. This can be used to convert a delta from a remote
+   * edit to apply to the current editor content.
+   */
+  private addEmbeddedElementsToDelta(modelDelta: DeltaStatic): DeltaStatic {
+    if (modelDelta.ops == null || modelDelta.ops.length < 1) {
+      return new Delta();
+    }
+    const adjustedDelta = new Delta();
+    let curIndex: number = 0;
+    let embedsUpToIndex: number = 0;
+    for (const op of modelDelta.ops) {
+      let cloneOp: DeltaOperation | undefined = cloneDeep(op);
+      const editorStartPos: number = curIndex + embedsUpToIndex;
+      if (cloneOp.retain != null) {
+        // editorStartPos must be the current index plus the number of embeds previous
+        const embedsInRange: number = this.calculateEmbedsInTextRange(editorStartPos, cloneOp.retain);
+        embedsUpToIndex += embedsInRange;
+        curIndex += cloneOp.retain;
+        // add to the retain op the number of embedded elements contained in its content
+        cloneOp.retain += embedsInRange;
+      } else if (cloneOp.delete != null) {
+        const embedsInRange: number = this.calculateEmbedsInTextRange(editorStartPos, cloneOp.delete);
+        curIndex += cloneOp.delete;
+        embedsUpToIndex += embedsInRange;
+        // add to the delete op the number of embedded elements contained in its content
+        cloneOp.delete += embedsInRange;
+        if (cloneOp.delete < 1) {
+          cloneOp = undefined;
         }
       }
-      return embeddedElementsCount;
+
+      if (cloneOp != null) {
+        (adjustedDelta as any).push(cloneOp);
+      }
     }
-    // get the length that would include embeds
-    return this.getEditorPositionPlusTextPosition(startIndex, length) - (startIndex + length);
+
+    return adjustedDelta;
+  }
+
+  /** Gets the number of embeds in a given range displayed in the quill editor. */
+  private getEmbedsInEditorRange(startIndex: number, length: number): number {
+    const indices: number[] = Array.from(this._embeddedElements.values());
+    const opEndIndex: number = startIndex + length;
+    let embeddedElementsCount: number = 0;
+    for (const embedIndex of indices) {
+      if (embedIndex < startIndex) {
+        continue;
+      } else if (embedIndex >= startIndex && embedIndex < opEndIndex) {
+        embeddedElementsCount++;
+      } else {
+        break;
+      }
+    }
+    return embeddedElementsCount;
+  }
+
+  /** Get the number of embeds displayed in the quill editor for a given range of text data. */
+  private calculateEmbedsInTextRange(editorStartIndex: number, length: number): number {
+    const positionWithoutEmbeds = editorStartIndex + length;
+    const positionWithEmbeds = this.getEditorPositionPlusTextPosition(editorStartIndex, length);
+    return positionWithEmbeds - positionWithoutEmbeds;
   }
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -688,10 +688,11 @@ export class TextViewModel {
   }
 
   /** Get the number of embeds displayed in the quill editor for a given range of text data. */
-  private calculateEmbedsInTextRange(editorStartIndex: number, length: number): number {
-    const positionWithoutEmbeds = editorStartIndex + length;
-    const positionWithEmbeds = this.getEditorPositionPlusTextPosition(editorStartIndex, length);
-    return positionWithEmbeds - positionWithoutEmbeds;
+  private calculateEmbedsInTextRange(editorStartPos: number, length: number): number {
+    const editorEndPos: number = this.getEditorPositionPlusTextPosition(editorStartPos, length);
+    const lengthWithEmbeds: number = editorEndPos - editorStartPos;
+    const embedCount: number = lengthWithEmbeds - length;
+    return embedCount;
   }
 
   private getAttributesAtPosition(editorPosition: number) {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -135,7 +135,10 @@ export class TextViewModel {
     editor.setContents(this.textDoc.data as DeltaStatic);
     editor.history.clear();
     if (subscribeToUpdates) {
-      this.remoteChangesSub = this.textDoc.remoteChanges$.subscribe(ops => editor.updateContents(ops as DeltaStatic));
+      this.remoteChangesSub = this.textDoc.remoteChanges$.subscribe(ops => {
+        const deltaWithEmbeds: DeltaStatic = this.adjustDeltaForEmbeddedElements(ops as DeltaStatic, false);
+        editor.updateContents(deltaWithEmbeds);
+      });
     }
     this.onCreateSub = this.textDoc.create$.subscribe(() => {
       if (textDoc.data != null) {
@@ -366,6 +369,20 @@ export class TextViewModel {
     return undefined;
   }
 
+  /** Returns the editor position that corresponds to a text position past an editor position. */
+  getEditorPositionPlusTextPosition(startingEditorPos: number, textPosPast: number): number {
+    let textCharactersFound = 0;
+    let resultingEditorPos = startingEditorPos;
+    const embedEditorPositions = Array.from(this.embeddedElements.values());
+    while (textCharactersFound < textPosPast) {
+      if (!embedEditorPositions.includes(resultingEditorPos)) {
+        textCharactersFound++;
+      }
+      resultingEditorPos++;
+    }
+    return resultingEditorPos;
+  }
+
   private viewToData(delta: DeltaStatic): DeltaStatic {
     let modelDelta = new Delta();
     if (delta.ops != null) {
@@ -390,7 +407,7 @@ export class TextViewModel {
         (modelDelta as any).push(modelOp);
       }
       // Remove Paratext notes from model delta
-      modelDelta = this.adjustDeltaForEmbeddedElements(modelDelta);
+      modelDelta = this.adjustDeltaForEmbeddedElements(modelDelta, true);
     }
     return modelDelta.chop();
   }
@@ -577,49 +594,64 @@ export class TextViewModel {
   }
 
   /** Adjust the delta that is applied to the text doc by stripping off the embedded elements displayed in quill. */
-  private adjustDeltaForEmbeddedElements(modelDelta: DeltaStatic): DeltaStatic {
+  private adjustDeltaForEmbeddedElements(modelDelta: DeltaStatic, removeEmbeds: boolean): DeltaStatic {
     if (modelDelta.ops == null || modelDelta.ops.length < 1) {
       return new Delta();
     }
     const adjustedDelta = new Delta();
     let curIndex: number = 0;
+    let embedsUpToIndex: number = 0;
     for (const op of modelDelta.ops) {
       let cloneOp: DeltaOperation | undefined = cloneDeep(op);
+      const editorStartPos: number = removeEmbeds ? curIndex : curIndex + embedsUpToIndex;
       if (cloneOp.retain != null) {
-        // The range of content that this retain applies to
-        const opEndIndex: number = curIndex + cloneOp.retain;
-        const embeddedElementsCount = this.getEmbeddedElementsWithin(curIndex, opEndIndex);
+        // If adjusting the delta by removing embeds, editorStartPos is the current index, otherwise
+        // editorStartPos must be the current index plus the number of embeds previous
+        const embedsInRange: number = this.getEmbedsInRange(editorStartPos, cloneOp.retain, removeEmbeds);
+        embedsUpToIndex += embedsInRange;
         curIndex += cloneOp.retain;
-        // remove from the retain op the number of embedded elements contained in its content
-        cloneOp.retain -= embeddedElementsCount;
+        // add or remove from the retain op the number of embedded elements contained in its content
+        const adjustment: number = removeEmbeds ? embedsInRange * -1 : embedsInRange;
+        cloneOp.retain += adjustment;
       } else if (cloneOp.delete != null) {
-        const opEndIndex: number = curIndex + cloneOp.delete;
-        const embeddedElementsCount = this.getEmbeddedElementsWithin(curIndex, opEndIndex);
+        const embedsInRange: number = this.getEmbedsInRange(editorStartPos, cloneOp.delete, removeEmbeds);
         curIndex += cloneOp.delete;
-        // remove from the delete op the number of embedded elements contained in its content
-        cloneOp.delete -= embeddedElementsCount;
+        embedsUpToIndex += embedsUpToIndex;
+        // add or remove from the delete op the number of embedded elements contained in its content
+        const adjustment: number = removeEmbeds ? embedsInRange * -1 : embedsInRange;
+        cloneOp.delete += adjustment;
         if (cloneOp.delete < 1) {
           cloneOp = undefined;
         }
       }
+
       if (cloneOp != null) {
         (adjustedDelta as any).push(cloneOp);
       }
     }
+
     return adjustedDelta;
   }
 
-  private getEmbeddedElementsWithin(startIndex: number, endIndex: number): number {
-    let embeddedElementsCount = 0;
-    for (const embedIndex of this.embeddedElements.values()) {
-      if (embedIndex < startIndex) {
-        continue;
-      } else if (embedIndex >= startIndex && embedIndex < endIndex) {
-        embeddedElementsCount++;
-      } else {
-        break;
+  private getEmbedsInRange(startIndex: number, length: number, fromEditorContent: boolean): number {
+    const indices: number[] = Array.from(this._embeddedElements.values());
+    // The range of content that this retain applies to
+    const opEndIndex: number = startIndex + length;
+    if (fromEditorContent) {
+      // subtract embeds from range
+      let embeddedElementsCount: number = 0;
+      for (const embedIndex of indices) {
+        if (embedIndex < startIndex) {
+          continue;
+        } else if (embedIndex >= startIndex && embedIndex < opEndIndex) {
+          embeddedElementsCount++;
+        } else {
+          break;
+        }
       }
+      return embeddedElementsCount;
     }
-    return embeddedElementsCount;
+    // get the length that would include embeds
+    return this.getEditorPositionPlusTextPosition(startIndex, length) - (startIndex + length);
   }
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -407,7 +407,7 @@ export class TextViewModel {
         (modelDelta as any).push(modelOp);
       }
       // Remove Paratext notes from model delta
-      modelDelta = this.removedEmbeddedElementsFromDelta(modelDelta);
+      modelDelta = this.removeEmbeddedElementsFromDelta(modelDelta);
     }
     return modelDelta.chop();
   }
@@ -597,7 +597,7 @@ export class TextViewModel {
    * Strip off the embedded elements displayed in quill from the delta. This can be used to convert a delta from
    * user edits to apply to a text doc.
    */
-  private removedEmbeddedElementsFromDelta(modelDelta: DeltaStatic): DeltaStatic {
+  private removeEmbeddedElementsFromDelta(modelDelta: DeltaStatic): DeltaStatic {
     if (modelDelta.ops == null || modelDelta.ops.length < 1) {
       return new Delta();
     }
@@ -658,6 +658,8 @@ export class TextViewModel {
         if (cloneOp.delete < 1) {
           cloneOp = undefined;
         }
+      } else if (cloneOp.insert != null) {
+        cloneOp.attributes = this.getAttributesAtPosition(editorStartPos);
       }
 
       if (cloneOp != null) {
@@ -690,5 +692,10 @@ export class TextViewModel {
     const positionWithoutEmbeds = editorStartIndex + length;
     const positionWithEmbeds = this.getEditorPositionPlusTextPosition(editorStartIndex, length);
     return positionWithEmbeds - positionWithoutEmbeds;
+  }
+
+  private getAttributesAtPosition(editorPosition: number) {
+    const editor: Quill = this.checkEditor();
+    return editor.getFormat(editorPosition);
   }
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -672,7 +672,7 @@ export class TextViewModel {
 
   /** Gets the number of embeds in a given range displayed in the quill editor. */
   private getEmbedsInEditorRange(startIndex: number, length: number): number {
-    const indices: number[] = Array.from(this._embeddedElements.values());
+    const indices: IterableIterator<number> = this._embeddedElements.values();
     const opEndIndex: number = startIndex + length;
     let embeddedElementsCount: number = 0;
     for (const embedIndex of indices) {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -58,6 +58,7 @@ export interface TextUpdatedEvent {
   prevSegment?: Segment;
   segment?: Segment;
   oldVerseEmbeds?: Map<string, number>;
+  isLocalUpdate?: boolean;
 }
 
 /**
@@ -536,19 +537,19 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
       return;
     }
 
-    const embedInsertPos: number = this.getEditorPositionPlusTextPosition(
+    const embedInsertPos: number = this.viewModel.getEditorPositionPlusTextPosition(
       editorPosOfSegmentToModify.index,
       startTextPosInVerse
     );
 
     this.editor.insertEmbed(embedInsertPos, formatName, format, 'api');
     const anchorEndPosition: number = startTextPosInVerse + textAnchor.length - 1;
-    const endIndex: number = this.getEditorPositionPlusTextPosition(
+    const endPosition: number = this.viewModel.getEditorPositionPlusTextPosition(
       editorPosOfSegmentToModify.index,
       anchorEndPosition
     );
     // add one to include the embed to underline
-    const formatLength: number = endIndex - embedInsertPos + 1;
+    const formatLength: number = endPosition - embedInsertPos + 1;
     this.editor.formatText(embedInsertPos, formatLength, 'text-anchor', 'true', 'api');
     this.updateSegment();
     return embedSegmentRef;
@@ -560,7 +561,8 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     this.updatePlaceholderText();
     // skip updating when only formatting changes occurred
     if (delta.ops != null && delta.ops.some(op => op.insert != null || op.delete != null)) {
-      this.update(delta);
+      const isUserEdit: boolean = source === 'user';
+      this.update(delta, isUserEdit);
     }
   }
 
@@ -725,7 +727,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     }
   }
 
-  private update(delta?: DeltaStatic): void {
+  private update(delta?: DeltaStatic, isUserEdit?: boolean): void {
     let segmentRef: string | undefined;
     let checksum: number | undefined;
     let focus: boolean | undefined;
@@ -785,7 +787,13 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     }
 
     Promise.resolve().then(() => this.adjustSelection());
-    this.updated.emit({ delta, prevSegment, segment: this._segment, oldVerseEmbeds: oldVerseEmbedsToUpdate });
+    this.updated.emit({
+      delta,
+      prevSegment,
+      segment: this._segment,
+      oldVerseEmbeds: oldVerseEmbedsToUpdate,
+      isLocalUpdate: isUserEdit
+    });
   }
 
   private tryChangeSegment(
@@ -915,22 +923,6 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
   private getEmbedCountInRange(editorStartPos: number, length: number): number {
     const embedPositions: number[] = Array.from(this.embeddedElements.values());
     return embedPositions.filter((pos: number) => pos >= editorStartPos && pos < editorStartPos + length).length;
-  }
-
-  /**
-   * Returns the editor position that corresponds to a text position past an editor position.
-   */
-  private getEditorPositionPlusTextPosition(startingEditorPos: number, textPosPast: number): number {
-    let textCharactersFound = 0;
-    let resultingEditorPos = startingEditorPos;
-    const embedEditorPositions = Array.from(this.embeddedElements.values());
-    while (textCharactersFound < textPosPast) {
-      if (!embedEditorPositions.includes(resultingEditorPos)) {
-        textCharactersFound++;
-      }
-      resultingEditorPos++;
-    }
-    return resultingEditorPos;
   }
 
   private setHighlightMarkerPosition(): void {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
@@ -58,7 +58,15 @@
           <app-text
             #target
             [isReadOnly]="!canEdit"
-            (updated)="onTargetUpdated($event.segment, $event.delta, $event.prevSegment, $event.oldVerseEmbeds)"
+            (updated)="
+              onTargetUpdated(
+                $event.segment,
+                $event.delta,
+                $event.prevSegment,
+                $event.oldVerseEmbeds,
+                $event.isLocalUpdate
+              )
+            "
             (loaded)="onTextLoaded('target')"
             [highlightSegment]="target.hasFocus"
             [markInvalid]="true"
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -17,7 +17,7 @@ import {
 } from '@sillsdev/machine';
 import cloneDeep from 'lodash-es/cloneDeep';
 import { CookieService } from 'ngx-cookie-service';
-import Quill, { DeltaStatic, Sources } from 'quill';
+import Quill, { DeltaOperation, DeltaStatic, Sources } from 'quill';
 import { SystemRole } from 'realtime-server/lib/esm/common/models/system-role';
 import { User } from 'realtime-server/lib/esm/common/models/user';
 import { obj } from 'realtime-server/lib/esm/common/utils/obj-path';
@@ -1150,7 +1150,8 @@ describe('EditorComponent', () => {
 
       let contents = env.targetEditor.getContents();
       let noteThreadEmbedCount = env.countNoteThreadEmbeds(contents.ops!);
-      env.component.target!.removeEmbeddedElements();
+      expect(noteThreadEmbedCount).toEqual(5);
+      env.component.removeEmbeddedElements();
       env.wait();
 
       contents = env.targetEditor.getContents();
@@ -1526,6 +1527,54 @@ describe('EditorComponent', () => {
       expect(noteThreadDoc.data!.position).toEqual({ start: text.length, length: 9 });
       verse4p1Index = env.component.target!.getSegmentRange('verse_1_4/p_1')!.index;
       expect(env.getNoteThreadEditorPosition('thread05')).toEqual(verse4p1Index);
+      env.dispose();
+    }));
+
+    it('remote edits correctly applied to editor', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig();
+      env.wait();
+      const noteThreadDoc: NoteThreadDoc = env.getNoteThreadDoc('project01', 'thread01');
+      expect(noteThreadDoc.data!.position).toEqual({ start: 8, length: 9 });
+
+      // target: chap|ter 1, verse 1.
+      let index: number = 15;
+      const textDoc: TextDoc = env.getTextDoc(new TextDocId('project01', 40, 1));
+      const insertDelta: DeltaStatic = new Delta();
+      (insertDelta as any).push({ retain: index } as DeltaOperation);
+      (insertDelta as any).push({ insert: 'abc' } as DeltaOperation);
+      textDoc.submit(insertDelta);
+      env.wait();
+      expect(env.component.target!.getSegmentText('verse_1_1')).toEqual('target: chapabcter 1, verse 1.');
+
+      index = env.getNoteThreadEditorPosition('thread02');
+      // targe|->t: chap<-|ter 1, verse 3.
+      const verse3InsertionIndex = index + 3;
+      const insertDeleteDelta: DeltaStatic = new Delta();
+      (insertDeleteDelta as any).push({ retain: verse3InsertionIndex } as DeltaOperation);
+      (insertDeleteDelta as any).push({ insert: 'defgh' } as DeltaOperation);
+      (insertDeleteDelta as any).push({ delete: 7 } as DeltaOperation);
+      textDoc.submit(insertDeleteDelta);
+      env.wait();
+      expect(env.component.target!.getSegmentText('verse_1_3')).toEqual('targdefghpter 1, verse 3.');
+
+      index = env.getNoteThreadEditorPosition('thread03');
+      // targdefghpter |->1, v<-|erse 3.
+      const verse3DeletionIndex: number = index - 6;
+      const deleteDelta: DeltaStatic = new Delta();
+      (deleteDelta as any).push({ retain: verse3DeletionIndex } as DeltaOperation);
+      (deleteDelta as any).push({ delete: 4 } as DeltaOperation);
+      textDoc.submit(deleteDelta);
+      env.wait();
+      expect(env.component.target!.getSegmentText('verse_1_3')).toEqual('targdefghpter erse 3.');
+      const notePositionInVerse: number = 19;
+      expect(env.getNoteThreadDoc('project01', 'thread03').data!.position.start).toEqual(notePositionInVerse);
+      expect(env.getNoteThreadDoc('project01', 'thread04').data!.position.start).toEqual(notePositionInVerse);
+      const verse3Index: number = env.component.target!.getSegmentRange('verse_1_3')!.index;
+      // 2 note icons in verse before this
+      expect(env.getNoteThreadEditorPosition('thread03')).toEqual(verse3Index + notePositionInVerse + 2);
+      // 1 note icon in verse before this
+      expect(env.getNoteThreadEditorPosition('thread04')).toEqual(verse3Index + notePositionInVerse + 1);
       env.dispose();
     }));
   });
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -1537,32 +1537,34 @@ describe('EditorComponent', () => {
       const noteThreadDoc: NoteThreadDoc = env.getNoteThreadDoc('project01', 'thread01');
       expect(noteThreadDoc.data!.position).toEqual({ start: 8, length: 9 });
 
-      // target: chap|ter 1, verse 1.
-      let index: number = 15;
+      // insert text
+      let remoteEditPosition: number = 15;
+      // target: $chap|ter 1, verse 1.
       const textDoc: TextDoc = env.getTextDoc(new TextDocId('project01', 40, 1));
       const insertDelta: DeltaStatic = new Delta();
-      (insertDelta as any).push({ retain: index } as DeltaOperation);
+      (insertDelta as any).push({ retain: remoteEditPosition } as DeltaOperation);
       (insertDelta as any).push({ insert: 'abc' } as DeltaOperation);
       textDoc.submit(insertDelta);
       env.wait();
       expect(env.component.target!.getSegmentText('verse_1_1')).toEqual('target: chapabcter 1, verse 1.');
 
-      index = env.getNoteThreadEditorPosition('thread02');
-      // targe|->t: chap<-|ter 1, verse 3.
-      const verse3InsertionIndex = index + 3;
+      // select and insert text e.g. pasting in a selection
+      remoteEditPosition = 40;
+      // $targ|->et: chap<-|ter 1, $$verse 3.
       const insertDeleteDelta: DeltaStatic = new Delta();
-      (insertDeleteDelta as any).push({ retain: verse3InsertionIndex } as DeltaOperation);
+      (insertDeleteDelta as any).push({ retain: remoteEditPosition } as DeltaOperation);
       (insertDeleteDelta as any).push({ insert: 'defgh' } as DeltaOperation);
       (insertDeleteDelta as any).push({ delete: 7 } as DeltaOperation);
       textDoc.submit(insertDeleteDelta);
       env.wait();
       expect(env.component.target!.getSegmentText('verse_1_3')).toEqual('targdefghpter 1, verse 3.');
 
-      index = env.getNoteThreadEditorPosition('thread03');
-      // targdefghpter |->1, v<-|erse 3.
-      const verse3DeletionIndex: number = index - 6;
+      // select and delete text including deleting notes
+      remoteEditPosition = 50;
+      // $targdefghpter |->1, $$v<-|erse 3.
       const deleteDelta: DeltaStatic = new Delta();
-      (deleteDelta as any).push({ retain: verse3DeletionIndex } as DeltaOperation);
+      (deleteDelta as any).push({ retain: remoteEditPosition } as DeltaOperation);
+      // the remote edit deletes 4, but locally it is expanded to 6 to include the 2 note embeds
       (deleteDelta as any).push({ delete: 4 } as DeltaOperation);
       textDoc.submit(deleteDelta);
       env.wait();
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -1537,62 +1537,128 @@ describe('EditorComponent', () => {
       const noteThreadDoc: NoteThreadDoc = env.getNoteThreadDoc('project01', 'thread01');
       expect(noteThreadDoc.data!.position).toEqual({ start: 8, length: 9 });
 
-      // insert text
+      // The remote user inserts text after the thread01 note
       let notePosition: number = env.getNoteThreadEditorPosition('thread01');
-      let positionRelativeToNote: number = 5;
+      let remoteEditPositionAfterNote: number = 5;
       let noteCountBeforePosition: number = 1;
-      let remoteEditPosition: number = env.getRemoteEditPosition(
+      // Text position in the text doc at which the remote user edits
+      let remoteEditTextPos: number = env.getRemoteEditPosition(
         notePosition,
-        positionRelativeToNote,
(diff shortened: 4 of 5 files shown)


irahopkinson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1172)* status: all files reviewed, 6 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 574 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MmutC3LDkPqPKxtQZIW:-MnOgER83dmyM_ZCG7Lj:b3f9f0s) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L574)):*

I wrestled to decide whether to make two separate methods, or just one. I see that with two separate method, it would be clearer what is happening. removeEmbeddedElementsFromDelta will remove any embedded elements from the delta so that the delta can be applied to the text doc (since text docs are unaware of embeded notes). AddEmbeddedElementsToDelta will add the embeds to the retain and delete ops in a delta so the delta can be applied to the current view of the content in the editor. Since the editor may contain note embeds, we need to account for that.
</blockquote></details>

I think splitting helps with the readability, however I don't understand enough of what is going on conceptually. Either you could walk me through it or another developer who understands the concepts should review it.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 652 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnOdB_y5cabFhHkbx9v:-MnOdB_y5cabFhHkbx9w:b-pjoxnp) ([raw file](https://github.com/sillsdev/web-xforge/blob/ee17dae3dc6df82fb625f5479be4ccb4f518c5ca/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L652)):*
> ```
> const indices: number[] = Array.from(this._embeddedElements.values());
> ```
I don't think you need to use `Array.from` since the `values90` method already gives you an iterable. Is this a hangover from the previous refactor?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 617 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mmufq_42DgkJeklv3HA:-MnOEjlU-zo-1AdIdzWB:bdxzsq8) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L617)):*

I took a look at how involved it would be to make a subclass of Editor component. It soon became obvious it would not be trivially easy. I might be tempted to do it, but it seems that exposing this method as public won't do us any harm. What do you think?
</blockquote></details>

Yeah, I think you are right.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1529 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mmukjd1DijuZ3iXHIwG:-MnOKySZ1tOGe8sRWkUh:b-ym8my5) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1529)):*

Hey, it turns out if I think of it differently, and use a note icon as a reference point for where a selection begins, I can explain why a remote edit is at the index that it is in a clear way. I was thinking about it in a totally upside-down way.
</blockquote></details>

Great, that's much clearer (and maintainable).
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1532 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnOVchA8ah96UGqXONo:-MnOVchA8ah96UGqXONp:bsdpeh3) ([raw file](https://github.com/sillsdev/web-xforge/blob/ee17dae3dc6df82fb625f5479be4ccb4f518c5ca/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1532)):*
> ```
> let remoteEditPosition: number = notePosition + positionRelativeToNote - localNoteCount;
> ```
You could make this DRY-er since it's used 3 times and the formulae shouldn't change. Perhaps use a getter or a function call directly?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1543 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnORQQpE6-ETktc3yDn:-MnORQQpE6-ETktc3yDo:b-7k717g) ([raw file](https://github.com/sillsdev/web-xforge/blob/ee17dae3dc6df82fb625f5479be4ccb4f518c5ca/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1543)):*
> ```
> // (1)target: + (2)$ + (3)chapabcter 1 + (4), verse 1.
> ```
I don't immediately understand what the parenthesis, numbers and plus mean? From the comment above I have inferred that $ is a note and | is the insert position. After a bit, it looks like you are indicating the 4 ops. Perhaps a commenting key would help?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1549 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnOUwgt4TORNPpIXSqS:-MnOUwgt4TORNPpIXSqT:b-dekf2x) ([raw file](https://github.com/sillsdev/web-xforge/blob/ee17dae3dc6df82fb625f5479be4ccb4f518c5ca/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1549)):*
> ```
> localNoteCount = 2;
> ```
The comment below has 3 local notes but there is only one before the selection. Why is this 2?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1551 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnOVL_o-mT_MFxUyP7X:-MnOVL_o-mT_MFxUyP7Y:bk4qkvk) ([raw file](https://github.com/sillsdev/web-xforge/blob/ee17dae3dc6df82fb625f5479be4ccb4f518c5ca/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1551)):*
> ```
> // $targ|->et: chap<-|ter 1, $$verse 3.
> ```
Looks like the end selection is too far to the right by one character 'p'.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1172)* status: all files reviewed, 6 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 574 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MmutC3LDkPqPKxtQZIW:-MnSFy5QDph9zwJxQGZ0:b-dl9ni2) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L574)):*

I think splitting helps with the readability, however I don't understand enough of what is going on conceptually. Either you could walk me through it or another developer who understands the concepts should review it.
</blockquote></details>

I'll see if I can poke Mark into reviewing this.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 652 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnOdB_y5cabFhHkbx9v:-MnSBPXv2u0wx7s_Djeb:b-cxwlbi) ([raw file](https://github.com/sillsdev/web-xforge/blob/ee17dae3dc6df82fb625f5479be4ccb4f518c5ca/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L652)):*

> ```
> const indices: number[] = Array.from(this._embeddedElements.values());
> ```
I don't think you need to use `Array.from` since the `values90` method already gives you an iterable. Is this a hangover from the previous refactor?
</blockquote></details>

Good thinking. I've flattened this out.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1532 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnOVchA8ah96UGqXONo:-MnSFdJk2GsBEyGy43mg:b6pm8eo) ([raw file](https://github.com/sillsdev/web-xforge/blob/ee17dae3dc6df82fb625f5479be4ccb4f518c5ca/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1532)):*

> ```
> let remoteEditPosition: number = notePosition + positionRelativeToNote - localNoteCount;
> ```
You could make this DRY-er since it's used 3 times and the formulae shouldn't change. Perhaps use a getter or a function call directly?
</blockquote></details>

I've created a method that can be reused.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1543 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnORQQpE6-ETktc3yDn:-MnSDr409-Samubc-NGo:b2o1jax) ([raw file](https://github.com/sillsdev/web-xforge/blob/ee17dae3dc6df82fb625f5479be4ccb4f518c5ca/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1543)):*

> ```
> // (1)target: + (2)$ + (3)chapabcter 1 + (4), verse 1.
> ```
I don't immediately understand what the parenthesis, numbers and plus mean? From the comment above I have inferred that $ is a note and | is the insert position. After a bit, it looks like you are indicating the 4 ops. Perhaps a commenting key would help?
</blockquote></details>

I made it a little more obvious by using square brackets and 0-based numbering. Any clearer?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1549 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnOUwgt4TORNPpIXSqS:-MnSEY0z1KV787h33sRs:b-w4mcst) ([raw file](https://github.com/sillsdev/web-xforge/blob/ee17dae3dc6df82fb625f5479be4ccb4f518c5ca/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1549)):*

> ```
> localNoteCount = 2;
> ```
The comment below has 3 local notes but there is only one before the selection. Why is this 2?
</blockquote></details>

This is the number of notes that come before a desired position. In this case, there is a note on verse 1, and a note at the beginning of verse 3. Those 2 notes come before our desired selection. To get the position in the text doc corresponding to that same position in the editor, the 2 notes would need to be subtracted. I've updated the name to be more descriptive.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1551 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnOVL_o-mT_MFxUyP7X:-MnSFYPD62MmcVoFHIpL:b-34yyl6) ([raw file](https://github.com/sillsdev/web-xforge/blob/ee17dae3dc6df82fb625f5479be4ccb4f518c5ca/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1551)):*

> ```
> // $targ|->et: chap<-|ter 1, $$verse 3.
> ```
Looks like the end selection is too far to the right by one character 'p'.
</blockquote></details>

Good eye, you are right.

irahopkinson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1172)* status: all files reviewed, 3 unresolved discussions (waiting on @irahopkinson and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1543 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnORQQpE6-ETktc3yDn:-MnUUH7d4MG9yGVvQecC:br4fuc5) ([raw file](https://github.com/sillsdev/web-xforge/blob/ee17dae3dc6df82fb625f5479be4ccb4f518c5ca/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1543)):*

I made it a little more obvious by using square brackets and 0-based numbering. Any clearer?
</blockquote></details>

Better
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1549 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnOUwgt4TORNPpIXSqS:-MnUUwIh3UUu1hG4Zszc:b-4eafnl) ([raw file](https://github.com/sillsdev/web-xforge/blob/ee17dae3dc6df82fb625f5479be4ccb4f518c5ca/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1549)):*

This is the number of notes that come before a desired position. In this case, there is a note on verse 1, and a note at the beginning of verse 3. Those 2 notes come before our desired selection. To get the position in the text doc corresponding to that same position in the editor, the 2 notes would need to be subtracted. I've updated the name to be more descriptive.
</blockquote></details>

Ok, but there are 2 notes before 'verse 3', not just one like you have described. It's still not obvious why only one of those last 2 notes counts? And it appears from your comment that the selection range is way before those last 2 notes anyway? In the test below the range covers the last 2 notes but there is only the first 2 notes counted still?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1548 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnUULiT4cw7kPcjDu2p:-MnUULiT4cw7kPcjDu2q:bq3jzy0) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1548)):*
> ```
> target:,
> ```
Missing a space, i.e. should be 'target: ,'

marksvc (changes requested):
Thanks for your work on this. 

> Reviewed 3 of 5 files at r1, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-:-MnagJee117WpoZRWAv_:btksj3a)* status: all files reviewed, 16 unresolved discussions (waiting on @irahopkinson and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 586 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXg-XZ9CrDxqZS-MSP:-MnXg-XZ9CrDxqZS-MSQ:bko06lh) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L586)):*
> ```XML
>       let cloneOp: DeltaOperation | undefined = cloneDeep(op);
>       if (cloneOp.retain != null) {
>         const embedsInRange: number = this.getEmbedsInEditorRange(curIndex, cloneOp.retain);
> ```

Yay. I like that the method name makes it clear that it's using editor positioning, and that it's now using (index,length).
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 597 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXgObS4F1t4sJytDQn:-MnXgObS4F1t4sJytDQo:bgnc9ip) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L597)):*
> ```XML
>         if (cloneOp.delete < 1) {
>           cloneOp = undefined;
>         }
> ```

Hmm, would there not be an insert op that would include the embeds, that might also need accounted for here?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 620 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXYLp28HZJNEuR99iD:-MnXYLp3C54FrMymOLxG:byk6oz3) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L620)):*
> ```XML
>     let embedsUpToIndex: number = 0;
>     for (const op of modelDelta.ops) {
>       let cloneOp: DeltaOperation | undefined = cloneDeep(op);
> ```

Why are you cloning and making a new DeltaStatic instead of re-using the existing DeltaStatic object and its ops?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 635 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXbwT93OBZQb7e3Jvu:-MnXbwT93OBZQb7e3Jvv:byewp0e) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L635)):*
> ```XML
>         // add to the delete op the number of embedded elements contained in its content
>         cloneOp.delete += embedsInRange;
>         if (cloneOp.delete < 1) {
> ```

Wow does that happen that there is a delete but it is for <=0 characters?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 669 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXhzibBsHx1WSPU5vi:-MnXhzibBsHx1WSPU5vj:b-f7r8wz) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L669)):*
> ```
>  positionWithoutEmbeds 
> ```

I think this variable name is misleading, because `editorStartIndex` will include embeds in its position information (if I'm correct). 
What do you think about re-working the method like the following?

```ts
// End position that includes embeds
const editorEndPos:number = this.getEditorPositionPlusTextPosition(editorStartIndex, length);
const lengthWithEmbeds:number = editorEndPos - editorStartIndex    
const countOfEmbeds:number = lengthWithEmbeds - length
return countOfEmbeds
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 562 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnadoIg1PIGJqW0wZAE:-MnadoIg1PIGJqW0wZAF:bgd9ar7) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L562)):*
> ```
>  endPosition
> ```

I do find that it's easier to think of the "position" in the editor rather than the "index" in the editor.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 580 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnafoBS1XfTwY71ZBKD:-MnafoBS1XfTwY71ZBKE:b-76cgga) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L580)):*
> ```XML
>     if (delta.ops != null && delta.ops.some(op => op.insert != null || op.delete != null)) {
>       const isUserEdit: boolean = source === 'user';
>       this.update(delta, isUserEdit);
> ```

Hmm. It might be worth showing in the spec that when `onContentChanged()` is called, we `updated.emit({islocalUpdate:true})`.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 115 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnackRO-B9BJFyl1zZF:-MnackRO-B9BJFyl1zZG:b-usq6uw) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L115)):*
> ```
> private shouldNoteThreadsRespondToEdits: boolean = false;
> ```

Ahh, I like this renaming.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 491 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnaXf9zDL-7Vjd-biie:-MnaXf9zDL-7Vjd-biif:b-hx7vna) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L491)):*
> ```
>  && isLocalUpdate)
> ```

I had to think and remember why the reason for this is: It's that the remote user, who authored the edit, their computer will have recomputed note positions and sent them separately (via probably a NoteThreadDoc). And so we don't process the anchor updating again on every client. Hmm. I'm undecided as to whether there should be a comment about that. 
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 491 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnaYWe37_GOu6Z73aOj:-MnaYWe37_GOu6Z73aOk:b-mpdp1u) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L491)):*
> ```
>  && isLocalUpdate
> ```

This would be good to show in the spec, that:
- a remote edit coming in for the current segment does not result in locally updating thread anchors, but does re-create deleted note thread embeds  
  - when the 'current segment' === the previous segment.
  - when the 'current segment' !== the previous segment.
- A remote edit coming in for a _different_ segment than the "current" one does not result in localy updating thread anchors, but does re-create deleted note thread embeds 
  - when the 'current segment' === the previous segment.
  - when the 'current segment' !== the previous segment.
- A local edit coming in for the current segment does result, and does result.
  - when the 'current segment' === the previous segment.
  - when the 'current segment' !== the previous segment.

And.. maybe after we think about those permutations a bit we'll end up re-writing parts of this method :/ ?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 617 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mmufq_42DgkJeklv3HA:-MnXlG2A4GjcK45iCIyM:b-9phlw) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L617)):*

Yeah, I think you are right.
</blockquote></details>

Another option is to keep private here: 
```ts
private removeEmbeddedElements()
```
and in the test, call the method after casting the object: 
```ts
(env.component as any).removeEmbeddedElements()
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 644 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mna_Zfb8e2SuEZfnju_:-Mna_Zfb8e2SuEZfnjua:b-9ivxuu) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L644)):*
> ```
>  this.embedNoteThread(featured);
> ```

It looks like we might get doubled up thread embeds if toggleNoteThreadVerses(true) is called twice in a row :). That might never happen, but something to consider if it ever comes up.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 1161 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mnac-SK4LhkJ0ZSu8mz:-Mnac-SK4LhkJ0ZSu8n-:b-f44vm6) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1161)):*
> ```
> !.data!
> ```

I'm nervous to see those in our production code. I tried just removing the `!` symbols and my code still compiled without complaint and the editor tests passed, so maybe they were helpful in an earlier draft of the code but aren't needed now?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1530 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXmA_6ArkC0Icj65oQ:-MnXmA_6ArkC0Icj65oR:b-95mgli) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1530)):*
> ```XML
>       // insert text
>       let notePosition: number = env.getNoteThreadEditorPosition('thread01');
>       let positionRelativeToNote: number = 5;
> ```

~~What does this mean?~~ Okay, after reading the rest of the test, I think this means "where the remote user is going to edit, _relative to `notePosition`_". Is that so?

This test, simulating a multi-user editing environment, I think will benefit from some more clarity. What do you think about the  ideas in the diff below to add clarity? These change some variable names, and some description of what the remote user is doing, and note where the 3 SUTs occur. 
(I only  included here the initial variable renames in the diff, not renames of all their usages, for conciseness.)

```diff
       expect(noteThreadDoc.data!.position).toEqual({ start: 8, length: 9 });
 
-      // insert text
+      // The remote user inserts text after the thread01 embed.
       let notePosition: number = env.getNoteThreadEditorPosition('thread01');
-      let positionRelativeToNote: number = 5;
+      let remoteEditPosAfterNote: number = 5;
-      let noteCountBeforePosition: number = 1;
+      let noteCountBeforeRemoteEditPos: number = 1;
-      let remoteEditPosition: number = env.getRemoteEditPosition(
+      // Text position in the text doc at which the remote user edits. This only regards text, not note embeds.
+      let remoteEditTextPos: number = env.getRemoteEditPosition(
         notePosition,
        positionRelativeToNote,
        noteCountBeforePosition
       );
       // $ represents a note thread embed
       // target: $chap|ter 1, verse 1.
       const textDoc: TextDoc = env.getTextDoc(new TextDocId('project01', 40, 1));
       const insertDelta: DeltaStatic = new Delta();
      (insertDelta as any).push({ retain: remoteEditPosition } as DeltaOperation);
       (insertDelta as any).push({ insert: 'abc' } as DeltaOperation);
+      // Simulate remote changes coming in.
       textDoc.submit(insertDelta);
+
+      // SUT 1
       env.wait();
+
+      // The local editor was updated to apply the remote edit in the right place, considering the note embed.
-      expect(env.component.target!.getSegmentText('verse_1_1')).toEqual('target: chapabcter 1, verse 1.');
+      expect(env.component.target!.getSegmentText('verse_1_1')).toEqual('target: chap' + 'abc' + 'ter 1, verse 1.');
       const verse1Range = env.component.target!.getSegmentRange('verse_1_1')!;
       const verse1Contents = env.targetEditor.getContents(verse1Range.index, verse1Range.length);
       // ops are [0]target:, [1]$, [2]chapabcter 1, [3], verse 1.
-      expect(verse1Contents.ops!.length).toEqual(4);
+      expect(verse1Contents.ops!.length).withContext('should have the expected op structure').toEqual(4);
-      expect(verse1Contents.ops![2].attributes!['text-anchor']).toBe(true);
+      expect(verse1Contents.ops![2].attributes!['text-anchor'])
+        .withContext('should have maintained the formatting')
+        .toBe(true);
 
-      // select and insert text e.g. pasting in a selection
+      // The remote user selects some text and pastes in a replacement. (-----if that really results in ops with an insert before a delete, not the other way around?----)
       notePosition = env.getNoteThreadEditorPosition('thread02');
      noteCountBeforePosition = 2;
      remoteEditPosition = env.getRemoteEditPosition(notePosition, positionRelativeToNote, noteCountBeforePosition);
       // $targ|->et: cha<-|pter 1, $$verse 3.
       const insertDeleteDelta: DeltaStatic = new Delta();
      (insertDeleteDelta as any).push({ retain: remoteEditPosition } as DeltaOperation);
       (insertDeleteDelta as any).push({ insert: 'defgh' } as DeltaOperation);
       (insertDeleteDelta as any).push({ delete: 7 } as DeltaOperation);
       textDoc.submit(insertDeleteDelta);
+
+      // SUT 2
       env.wait();
-      expect(env.component.target!.getSegmentText('verse_1_3')).toEqual('targdefghpter 1, verse 3.');
 
+      expect(env.component.target!.getSegmentText('verse_1_3')).toEqual('targ' + 'defgh' + 'pter 1, verse 3.');
+
-      // select and delete text including deleting notes
+      // The remote user selects and deletes some text. The text contains a couple note thread embeds.
      positionRelativeToNote = 15;
      remoteEditPosition = env.getRemoteEditPosition(notePosition, positionRelativeToNote, noteCountBeforePosition);
       // $targdefghpter |->1, $$v<-|erse 3.
       const deleteDelta: DeltaStatic = new Delta();
      (deleteDelta as any).push({ retain: remoteEditPosition } as DeltaOperation);
       // the remote edit deletes 4, but locally it is expanded to 6 to include the 2 note embeds
       (deleteDelta as any).push({ delete: 4 } as DeltaOperation);
       textDoc.submit(deleteDelta);
+
+      // SUT 3
       env.wait();
+
-      expect(env.component.target!.getSegmentText('verse_1_3')).toEqual('targdefghpter erse 3.');
+      expect(env.component.target!.getSegmentText('verse_1_3')).toEqual('targdefghpter ' + 'erse 3.');
       const notePositionInVerse: number = 19;
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1556 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXnRDx-A5foI3ug_JY:-MnXnRDx-A5foI3ug_JZ:b-7jyyy8) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1556)):*
> ```XML
>       noteCountBeforePosition = 2;
>       remoteEditPosition = env.getRemoteEditPosition(notePosition, positionRelativeToNote, noteCountBeforePosition);
>       // $targ|->et: cha<-|pter 1, $$verse 3.
> ```

I didn't understand this until I read the ops below it :)
It might be more clear to instead have the comment say something like the following. What do you think?
```ts
// Base text: $target: chapter 1, $$verse 3
//                 ^ where remote user applies edits
//                 ||||||| base text that gets deleted by remote user
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1560 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXp2id5NvH78gSYA73:-MnXp2id5NvH78gSYA74:bqzeb9x) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1560)):*
> ```
>  7
> ```

Clearer than `7` would be 
```ts
`get: cha`.length
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1568 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXqBTy7pbBK4wJu7hZ:-MnXqBTy7pbBK4wJu7h_:brtv4xn) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1568)):*
> ```XML
>       positionRelativeToNote = 15;
>       remoteEditPosition = env.getRemoteEditPosition(notePosition, positionRelativeToNote, noteCountBeforePosition);
>       // $targdefghpter |->1, $$v<-|erse 3.
> ```

What do you think about using something like this instead?

```ts
// Base text: $targdefghpter 1, $$verse 3.
//                           |||||| editor area that gets deleted (in remote user editor, and in local user editor)
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1576 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXqkwSAOm9vsgZnLie:-MnXqkwSAOm9vsgZnLif:b34tfi1) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1576)):*
> ```
>  19
> ```

Can we replace the 19 with something more clear where that is? Like maybe the following if it's how we get 19:
`targdefghpter erse `.length

Also, what happened here? Below we assert that thread03 and thread04 are at a specific position, both text-wise and editor-wise. Did the note embeds move to a new place in response to their embeds being deleted? To almost the end of the verse? Were the notes at this position before, or is this a new position? Oh I see, it was at start position 8 at the beginning of the test, so this is a new position.
Let's say what happened with a comment. eg "Thread positions were updated, after their embed icons were deleted, with positions based on the local client's best guess of where they are, which is near the end of the verse." (If that's what's happening?)

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1172)* status: all files reviewed, 16 unresolved discussions (waiting on @irahopkinson and @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 597 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXgObS4F1t4sJytDQn:-MnfcQCa9oa06kHagmfv:b-nn6nqx) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L597)):*

Hmm, would there not be an insert op that would include the embeds, that might also need accounted for here?
</blockquote></details>

There could be embeds if a user decides to undo the deletion of an embed. That will come in another task I am working on.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 620 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXYLp28HZJNEuR99iD:-MnfdoiL1ntYvrn4KW-v:b8gagpr) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L620)):*

Why are you cloning and making a new DeltaStatic instead of re-using the existing DeltaStatic object and its ops?
</blockquote></details>

I don't remember why I made that choice to clone and an entirely new DeltaStatic. I wouldn't be opposed to modifying the original DeltaStatic rather than creating a new one. I believe meant to follow the example of viewToData() which modifies a copy of the delta.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 635 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXbwT93OBZQb7e3Jvu:-MnfeWGPBS5x_kJLXgsm:b-wf359i) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L635)):*

Wow does that happen that there is a delete but it is for <=0 characters?
</blockquote></details>

The delete could be 0, in the case that a user deletes an embedded note. If it is 0, I've seen something strange happen, where the operation deletes the range of text in the previous retain op. Very strange indeed.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 669 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXhzibBsHx1WSPU5vi:-Mnfiz_kAytz8NBvO4h2:b-9fmn3c) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L669)):*

> ```
>  positionWithoutEmbeds 
> ```

I think this variable name is misleading, because `editorStartIndex` will include embeds in its position information (if I'm correct). 
What do you think about re-working the method like the following?

```ts
// End position that includes embeds
const editorEndPos:number = this.getEditorPositionPlusTextPosition(editorStartIndex, length);
const lengthWithEmbeds:number = editorEndPos - editorStartIndex    
const countOfEmbeds:number = lengthWithEmbeds - length
return countOfEmbeds
```
</blockquote></details>

Good suggestion. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 562 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnadoIg1PIGJqW0wZAE:-Mnfj4OtByMlG7VT15Lw:b9m0o6r) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L562)):*

> ```
>  endPosition
> ```

I do find that it's easier to think of the "position" in the editor rather than the "index" in the editor.
</blockquote></details>

Yes, I used to think in terms of index, but now I am finding that consistently using position has helps me to think in that term now.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 580 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnafoBS1XfTwY71ZBKD:-MnfmQEw8Bc6DVDTxv74:b-bbs27z) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L580)):*

Hmm. It might be worth showing in the spec that when `onContentChanged()` is called, we `updated.emit({islocalUpdate:true})`.
</blockquote></details>

I think this is reasonably covered. If I reverse the check for isLocalUpdate in `editor.component.ts` and check that it is not a local update, I get 10 test failure. That should be enough to show that it is emitting as expected.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 491 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnaXf9zDL-7Vjd-biie:-Mnfn2mJ3NJ3oRKdsYks:bs49sb9) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L491)):*

> ```
>  && isLocalUpdate)
> ```

I had to think and remember why the reason for this is: It's that the remote user, who authored the edit, their computer will have recomputed note positions and sent them separately (via probably a NoteThreadDoc). And so we don't process the anchor updating again on every client. Hmm. I'm undecided as to whether there should be a comment about that. 
</blockquote></details>

That's correct, only for the user who makes the edits will the text anchor be updated. I'll add a comment so that we don't forget the reasoning.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 491 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnaYWe37_GOu6Z73aOj:-MngqHNT5ea7y-K5h8kw:b-jic3k9) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L491)):*

> ```
>  && isLocalUpdate
> ```

This would be good to show in the spec, that:
- a remote edit coming in for the current segment does not result in locally updating thread anchors, but does re-create deleted note thread embeds  
  - when the 'current segment' === the previous segment.
  - when the 'current segment' !== the previous segment.
- A remote edit coming in for a _different_ segment than the "current" one does not result in localy updating thread anchors, but does re-create deleted note thread embeds 
  - when the 'current segment' === the previous segment.
  - when the 'current segment' !== the previous segment.
- A local edit coming in for the current segment does result, and does result.
  - when the 'current segment' === the previous segment.
  - when the 'current segment' !== the previous segment.

And.. maybe after we think about those permutations a bit we'll end up re-writing parts of this method :/ ?

</blockquote></details>

Ok, I did some testing and I could not come up with a scenario where a remote edit occurs, and current segment !== previous segment. In this case, we can stick to testing the scenario where the current segment === previous segment. I've added a test for the remote edit scenarios. The local edit scenario is already covered by tests.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 644 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mna_Zfb8e2SuEZfnju_:-MnfwPbN64KkqlGfqFqg:bcctk6u) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L644)):*

> ```
>  this.embedNoteThread(featured);
> ```

It looks like we might get doubled up thread embeds if toggleNoteThreadVerses(true) is called twice in a row :). That might never happen, but something to consider if it ever comes up.
</blockquote></details>

It ok to call this multiple times, the element will only get embedded if it doesn't already exist in the editor. `text.component.embedElementInline()` does the check of this.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 1161 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mnac-SK4LhkJ0ZSu8mz:-Mnfx-MD1hLwTFuTwEha:b-j626eo) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1161)):*

> ```
> !.data!
> ```

I'm nervous to see those in our production code. I tried just removing the `!` symbols and my code still compiled without complaint and the editor tests passed, so maybe they were helpful in an earlier draft of the code but aren't needed now?
</blockquote></details>

You're right. Not sure how I left those in there.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1549 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnOUwgt4TORNPpIXSqS:-Mng2xEj9X1rEkunI3vD:bu76lor) ([raw file](https://github.com/sillsdev/web-xforge/blob/ee17dae3dc6df82fb625f5479be4ccb4f518c5ca/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1549)):*

Ok, but there are 2 notes before 'verse 3', not just one like you have described. It's still not obvious why only one of those last 2 notes counts? And it appears from your comment that the selection range is way before those last 2 notes anyway? In the test below the range covers the last 2 notes but there is only the first 2 notes counted still?
</blockquote></details>

You may be confusing the notes in the comment with the notes I'm referring to. There are 3 notes in verse 3, but only the first note counts as being BEFORE the selection. As you pointed out, those two notes on `verse 3` do not count to be before the selection. So that is one note at the beginning of verse 3, and one note on VERSE 1. Two notes in total that come BEFORE the selection.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1530 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXmA_6ArkC0Icj65oQ:-Mng9KOt8DOUbTlbfvra:bdyg1xk) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1530)):*

~~What does this mean?~~ Okay, after reading the rest of the test, I think this means "where the remote user is going to edit, _relative to `notePosition`_". Is that so?

This test, simulating a multi-user editing environment, I think will benefit from some more clarity. What do you think about the  ideas in the diff below to add clarity? These change some variable names, and some description of what the remote user is doing, and note where the 3 SUTs occur. 
(I only  included here the initial variable renames in the diff, not renames of all their usages, for conciseness.)

```diff
       expect(noteThreadDoc.data!.position).toEqual({ start: 8, length: 9 });
 
-      // insert text
+      // The remote user inserts text after the thread01 embed.
       let notePosition: number = env.getNoteThreadEditorPosition('thread01');
-      let positionRelativeToNote: number = 5;
+      let remoteEditPosAfterNote: number = 5;
-      let noteCountBeforePosition: number = 1;
+      let noteCountBeforeRemoteEditPos: number = 1;
-      let remoteEditPosition: number = env.getRemoteEditPosition(
+      // Text position in the text doc at which the remote user edits. This only regards text, not note embeds.
+      let remoteEditTextPos: number = env.getRemoteEditPosition(
         notePosition,
        positionRelativeToNote,
        noteCountBeforePosition
       );
       // $ represents a note thread embed
       // target: $chap|ter 1, verse 1.
       const textDoc: TextDoc = env.getTextDoc(new TextDocId('project01', 40, 1));
       const insertDelta: DeltaStatic = new Delta();
      (insertDelta as any).push({ retain: remoteEditPosition } as DeltaOperation);
       (insertDelta as any).push({ insert: 'abc' } as DeltaOperation);
+      // Simulate remote changes coming in.
       textDoc.submit(insertDelta);
+
+      // SUT 1
       env.wait();
+
+      // The local editor was updated to apply the remote edit in the right place, considering the note embed.
-      expect(env.component.target!.getSegmentText('verse_1_1')).toEqual('target: chapabcter 1, verse 1.');
+      expect(env.component.target!.getSegmentText('verse_1_1')).toEqual('target: chap' + 'abc' + 'ter 1, verse 1.');
       const verse1Range = env.component.target!.getSegmentRange('verse_1_1')!;
       const verse1Contents = env.targetEditor.getContents(verse1Range.index, verse1Range.length);
       // ops are [0]target:, [1]$, [2]chapabcter 1, [3], verse 1.
-      expect(verse1Contents.ops!.length).toEqual(4);
+      expect(verse1Contents.ops!.length).withContext('should have the expected op structure').toEqual(4);
-      expect(verse1Contents.ops![2].attributes!['text-anchor']).toBe(true);
+      expect(verse1Contents.ops![2].attributes!['text-anchor'])
+        .withContext('should have maintained the formatting')
+        .toBe(true);
 
-      // select and insert text e.g. pasting in a selection
+      // The remote user selects some text and pastes in a replacement. (-----if that really results in ops with an insert before a delete, not the other way around?----)
       notePosition = env.getNoteThreadEditorPosition('thread02');
      noteCountBeforePosition = 2;
      remoteEditPosition = env.getRemoteEditPosition(notePosition, positionRelativeToNote, noteCountBeforePosition);
       // $targ|->et: cha<-|pter 1, $$verse 3.
       const insertDeleteDelta: DeltaStatic = new Delta();
      (insertDeleteDelta as any).push({ retain: remoteEditPosition } as DeltaOperation);
       (insertDeleteDelta as any).push({ insert: 'defgh' } as DeltaOperation);
       (insertDeleteDelta as any).push({ delete: 7 } as DeltaOperation);
       textDoc.submit(insertDeleteDelta);
+
+      // SUT 2
       env.wait();
-      expect(env.component.target!.getSegmentText('verse_1_3')).toEqual('targdefghpter 1, verse 3.');
 
+      expect(env.component.target!.getSegmentText('verse_1_3')).toEqual('targ' + 'defgh' + 'pter 1, verse 3.');
+
-      // select and delete text including deleting notes
+      // The remote user selects and deletes some text. The text contains a couple note thread embeds.
      positionRelativeToNote = 15;
      remoteEditPosition = env.getRemoteEditPosition(notePosition, positionRelativeToNote, noteCountBeforePosition);
       // $targdefghpter |->1, $$v<-|erse 3.
       const deleteDelta: DeltaStatic = new Delta();
      (deleteDelta as any).push({ retain: remoteEditPosition } as DeltaOperation);
       // the remote edit deletes 4, but locally it is expanded to 6 to include the 2 note embeds
       (deleteDelta as any).push({ delete: 4 } as DeltaOperation);
       textDoc.submit(deleteDelta);
+
+      // SUT 3
       env.wait();
+
-      expect(env.component.target!.getSegmentText('verse_1_3')).toEqual('targdefghpter erse 3.');
+      expect(env.component.target!.getSegmentText('verse_1_3')).toEqual('targdefghpter ' + 'erse 3.');
       const notePositionInVerse: number = 19;
```

</blockquote></details>

Thanks. I've taken these suggestions.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1548 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnUULiT4cw7kPcjDu2p:-Mng2gTK7_641x8IppeM:b6zrrk4) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1548)):*

> ```
> target:,
> ```
Missing a space, i.e. should be 'target: ,'
</blockquote></details>

Good catch!
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1556 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXnRDx-A5foI3ug_JY:-Mng4Mui-XEcr8OQ9J91:bhy5gzh) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1556)):*

I didn't understand this until I read the ops below it :)
It might be more clear to instead have the comment say something like the following. What do you think?
```ts
// Base text: $target: chapter 1, $$verse 3
//                 ^ where remote user applies edits
//                 ||||||| base text that gets deleted by remote user
```

</blockquote></details>

I've updated the comment to be more clear
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1560 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXp2id5NvH78gSYA73:-Mng9QJ1Cpj3UqHpQi_x:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1560)):*

> ```
>  7
> ```

Clearer than `7` would be 
```ts
`get: cha`.length
```
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1568 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXqBTy7pbBK4wJu7hZ:-Mng4HDv2ixxny63bZ4m:bgw57cx) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1568)):*

What do you think about using something like this instead?

```ts
// Base text: $targdefghpter 1, $$verse 3.
//                           |||||| editor area that gets deleted (in remote user editor, and in local user editor)
```
</blockquote></details>

Ok, I've updated the comment.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1576 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnXqkwSAOm9vsgZnLie:-Mng8hi57UKbDun8kmMa:b-4os9nq) ([raw file](https://github.com/sillsdev/web-xforge/blob/00577bcd002cf35e9f79e2a5c99ddf3cc8b628b7/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1576)):*

> ```
>  19
> ```

Can we replace the 19 with something more clear where that is? Like maybe the following if it's how we get 19:
`targdefghpter erse `.length

Also, what happened here? Below we assert that thread03 and thread04 are at a specific position, both text-wise and editor-wise. Did the note embeds move to a new place in response to their embeds being deleted? To almost the end of the verse? Were the notes at this position before, or is this a new position? Oh I see, it was at start position 8 at the beginning of the test, so this is a new position.
Let's say what happened with a comment. eg "Thread positions were updated, after their embed icons were deleted, with positions based on the local client's best guess of where they are, which is near the end of the verse." (If that's what's happening?)

</blockquote></details>

Ok, I've updated this section. The 19 comes from the original text position of note threads 3 and 4. In this simulation, we expect that the notes get recreated and placed at the position recorded in the note thread doc, which we have not changed. I've given the 19 the name `originalNotePosInVerse`

irahopkinson (commented):
This looks good. I haven't marked any files as read so @marksvc can finish his review.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-:-Mnxxwn04PMyBMQVZn-d:b-umit17)* status: 1 of 5 files reviewed, 14 unresolved discussions (waiting on @irahopkinson, @marksvc, and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 617 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mmufq_42DgkJeklv3HA:-MnxrcMACsS98_dYDB-7:b-28wm3p) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L617)):*

Another option is to keep private here: 
```ts
private removeEmbeddedElements()
```
and in the test, call the method after casting the object: 
```ts
(env.component as any).removeEmbeddedElements()
```

</blockquote></details>

@marksvc your suggestion works but because it breaks type checking we should only use it where necessary. Generally if a method doesn't need to be public we keep it private and only make it public if it needs to be used externally. But there are really good reasons to keep some things private even though it makes testing hard - that would be a good reason to use your suggestion. In this case though, although there is no reason other than testing to make it public, there is also no reasons to want to keep it private.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1549 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnOUwgt4TORNPpIXSqS:-Mnxwe-U2-DRjaAjRk0Y:b580kfp) ([raw file](https://github.com/sillsdev/web-xforge/blob/ee17dae3dc6df82fb625f5479be4ccb4f518c5ca/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1549)):*

You may be confusing the notes in the comment with the notes I'm referring to. There are 3 notes in verse 3, but only the first note counts as being BEFORE the selection. As you pointed out, those two notes on `verse 3` do not count to be before the selection. So that is one note at the beginning of verse 3, and one note on VERSE 1. Two notes in total that come BEFORE the selection.
</blockquote></details>

Got it, thanks, the note count is per chapter, not per verse. How about a comment that says the count of 2 is 1 from verse 1 and 1 from the note in this verse (3) that is before the selection point?

marksvc (changes requested):
I have some naming + clarity comments, and then good to go. 

> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-:-Mo0YrLxAH4cepniLvgi:b923czp)* status: all files reviewed, 4 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 617 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mmufq_42DgkJeklv3HA:-Mo0XA815Abqsm23VR4z:b7x9iti) ([raw file](https://github.com/sillsdev/web-xforge/blob/46215c7f760e8f117dab880aa711754afc5885ad/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L617)):*

@marksvc your suggestion works but because it breaks type checking we should only use it where necessary. Generally if a method doesn't need to be public we keep it private and only make it public if it needs to be used externally. But there are really good reasons to keep some things private even though it makes testing hard - that would be a good reason to use your suggestion. In this case though, although there is no reason other than testing to make it public, there is also no reasons to want to keep it private.
</blockquote></details>

Okay. Thanks. 
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1591 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mo0Nu6AD7pFxlzlaLmy:-Mo0Nu6AD7pFxlzlaLmz:bt7f6vd) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f8ffea1d70335a4f79f3eea7c88e68979680f19/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1591)):*
> ```
>  6 characters deleted locally
> ```

If we want to say 6 somethings get deleted here, it may be inconsistent or not helpful to say "characters". It is 4 characters that are deleted (well, 4 .. things that have text codepoints), and two embeds of length 1. "Characters" may be nice to reserve for referring to things that would also show up in the text doc. 
What do you think of changing this comment to something like one of the following? 
- editor range deleted
- editor range of 6 deleted
- 4 characters, and the 2 embeds between them, deleted
- 6 editor positions deleted
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1627 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mo0RSGB65vGlDw4FB4v:-Mo0RSGCEEvmZSGRv-Dv:b-b7hu5g) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f8ffea1d70335a4f79f3eea7c88e68979680f19/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1627)):*
> ```
> AfterNote
> ```

Oh, naming. So what we mean here is the position _relative to the beginning of the note embed_, right? Because the editor position _after_ the note embed is 1 (I think). `v` would be 0 after it, and `e` would be 1 after it. And I'm the one who gave the variable name `remoteEditPositionAfterNote`; what can you do? :-)
I must not have counted last time to see the discrepancy. 
Your earlier variable name,  positionRelativeToNote, captured some good aspects of this. 
I'm trying to think of a better variable name to use, but I think we are being limited by the helper function. 
I don't want to be too pedantic :), but what do you think about changing `getRemoteEditPosition()` to instead `    return notePosition + 1 + positionAfter - noteCount;`, and then make the `remoteEditPositionAfterNote` variable keep the same variable name, but decrease all their values by 1? (For example here, it would say `let remoteEditPositionAfterNote: number = 1;`.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1628 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mo0Rp7PB8nDDAo4Th3s:-Mo0Rp7PB8nDDAo4Th3t:b-v0pvay) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f8ffea1d70335a4f79f3eea7c88e68979680f19/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1628)):*
> ```
> noteCountBeforePosition: number = 4;
> ```

That's kind of confusing. Maybe a comment should say `// (Including a note before this verse)` if that's the reason for 4 instead of 3.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1660 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mo0SWYF5b8bo7iraas5:-Mo0SWYF5b8bo7iraas6:bn1m3qc) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f8ffea1d70335a4f79f3eea7c88e68979680f19/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1660)):*
> ```
> ('target: c' + 'def' + 'hapter 1, v
> ```

This, and the `|` character in 
>    // target: $c|hapter 1, verse 1.

provide helpful clarity.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1172)* status: all files reviewed, 4 unresolved discussions (waiting on @irahopkinson and @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1549 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-MnOUwgt4TORNPpIXSqS:-MoKwE_5D828BIam0USC:b-rfkwhz) ([raw file](https://github.com/sillsdev/web-xforge/blob/ee17dae3dc6df82fb625f5479be4ccb4f518c5ca/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1549)):*

Got it, thanks, the note count is per chapter, not per verse. How about a comment that says the count of 2 is 1 from verse 1 and 1 from the note in this verse (3) that is before the selection point?
</blockquote></details>

I can add a comment.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1591 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mo0Nu6AD7pFxlzlaLmy:-MoKwQFMAXeRpa48JCJF:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f8ffea1d70335a4f79f3eea7c88e68979680f19/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1591)):*

> ```
>  6 characters deleted locally
> ```

If we want to say 6 somethings get deleted here, it may be inconsistent or not helpful to say "characters". It is 4 characters that are deleted (well, 4 .. things that have text codepoints), and two embeds of length 1. "Characters" may be nice to reserve for referring to things that would also show up in the text doc. 
What do you think of changing this comment to something like one of the following? 
- editor range deleted
- editor range of 6 deleted
- 4 characters, and the 2 embeds between them, deleted
- 6 editor positions deleted


</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1627 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mo0RSGB65vGlDw4FB4v:-MoKxf5n85TE4WA8NaAG:b-soahpz) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f8ffea1d70335a4f79f3eea7c88e68979680f19/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1627)):*

> ```
> AfterNote
> ```

Oh, naming. So what we mean here is the position _relative to the beginning of the note embed_, right? Because the editor position _after_ the note embed is 1 (I think). `v` would be 0 after it, and `e` would be 1 after it. And I'm the one who gave the variable name `remoteEditPositionAfterNote`; what can you do? :-)
I must not have counted last time to see the discrepancy. 
Your earlier variable name,  positionRelativeToNote, captured some good aspects of this. 
I'm trying to think of a better variable name to use, but I think we are being limited by the helper function. 
I don't want to be too pedantic :), but what do you think about changing `getRemoteEditPosition()` to instead `    return notePosition + 1 + positionAfter - noteCount;`, and then make the `remoteEditPositionAfterNote` variable keep the same variable name, but decrease all their values by 1? (For example here, it would say `let remoteEditPositionAfterNote: number = 1;`.


</blockquote></details>

Ok, I'll add the + 1 to the position so that positions are 0-indexed positions again.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1628 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-Mo0Rp7PB8nDDAo4Th3s:-MoKyByK2xGRO7UXMbIf:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f8ffea1d70335a4f79f3eea7c88e68979680f19/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1628)):*

> ```
> noteCountBeforePosition: number = 4;
> ```

That's kind of confusing. Maybe a comment should say `// (Including a note before this verse)` if that's the reason for 4 instead of 3.
</blockquote></details>

Done.

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1172)* status: all files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

irahopkinson (approved):
:lgtm:

> Reviewed 3 of 4 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1172#-:-MoMjEX850dWX3XbeyJm:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1182, opened 2021-11-09
SF-1410 Bulk import of questions from CSV file

Nateowami:
- Added Papaparse for CSV parsing (6.38 KB gzipped, according to webpack-bundle-analyzer)
- Always show import questions dialog, since questions can now be imported even if Transcelerator questions are unavailable (this also makes it more discoverable)

This is a good branch to check out to review its functionality.

Here's how the import button looks now:

Here's what the dialog now looks like:

Here's how the filtering and selecting now looks. This uses a combination of Angular Material and Angular MDC components, unfortunately (as of now).

This view is shown when importing from a CSV file and some of the lines cannot be parsed:

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182)


RaymondLuong3 (changes requested):
Just beginning to go through this review now. I will get back to it on monday, but I've added my initial comments.

> Reviewed 11 of 29 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-:-MoLTDYa2WA2Se7_0h03:b-tyqdx9)* status: 11 of 29 files reviewed, 12 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html, line 57 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoL1AWE63H4GaqxnNMV:-MoL1AWE63H4GaqxnNMW:brl83en) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L57)):*
> ```
> "urls.communityAnnouncementPages"
> ```
Glad to see these put in it's own file.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.ts, line 383 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoL79ix1j_Jgz7wAqHP:-MoL79ix1j_Jgz7wAqHQ:b3xe3r5) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.ts#L383)):*
> ```
>  this.matDialog.open(ImportQuestionsDialogComponent, { data });
> ```
Is there a reason this could not user mdc Dialog? Oh, I remember now. We are moving things back to angular material. Ok, I guess for the time being we have both in our app.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 34 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoL9lCeD5o8JyteC8Lb:-MoL9lCeD5o8JyteC8Lc:bnz7zdz) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L34)):*
> ```
> target="_blank">Learn more</a>
> ```
This should be internationalized.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 150 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLNMXC76O4P_2c30f7:-MoLNMXC76O4P_2c30f8:b-hcyrcc) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L150)):*
> ```
>       <div *ngIf="status === 'no_questions'">{{ t("no_questions_available") }}</div>
> ```
Angular has an *ngSwitch directive that could have been used. I'm not too familiar with it, so I am happy to keep this way of checking using *ngIf.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 155 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLNDXA0jnHDlwgG03C:-MoLNDXA0jnHDlwgG03D:b-jcy60k) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L155)):*
> ```
>         <mdc-linear-progress [determinate]="true" [progress]="importedCount / toImportCount"></mdc-linear-progress>
> ```
angular material has linear progress bar now. It should be straight forward to use.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 176 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLP-yAD0uEYg6WIXX6:-MoLP-yAD0uEYg6WIXX7:b-8ay1yk) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L176)):*
> ```
> <ng-container *ngIf="importCanceled"><br /><br />{{ t("canceling_import") }}</ng-container>
> ```
Is there a reason for using a container instead of a div? And I think it would be cleaner to use styling in css rather than <br /> tags.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss, line 81 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLR2928AHdbaSfM1SQ:-MoLR2928AHdbaSfM1SR:b-3u9h0e) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss#L81)):*
<details><summary><i alt="summary">Quoted 11 lines of code…</i></summary>

> ```
> .mat-column-rowNumber {
>   border-right: 1px solid rgba(0, 0, 0, 0.12);
>   padding-right: 1em;
>   text-align: right;
>   font-weight: 500;
> }
> 
> .mat-column-reference {
>   padding-left: 1em;
>   padding-right: 1em;
> }
> ```
</details>

I wonder if this styling may leak into other parts of the dialog. Do we have have the one data table. I see that we do define matColumnDef="rowNumber" and "matColumnDef="reference" so that may be enough.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 100 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLS8Is4O2YiYrsVS_x:-MoLS8Is4O2YiYrsVS_y:bw8l6hp) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L100)):*
> ```
>     readonly mdcDialog: MdcDialog,
>     readonly matDialog: MatDialog,
> ```
These look like they should be private
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLSdlh6cngTqvv-K9x:-MoLSdlh6cngTqvv-K9y:b1m83nk) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L165)):*
> ```
> return this.invalidRows!.
> ```
Is that a not null assertion? How do we know here it won't be null?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 172 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoL9IGO6XxibgQ0zn29:-MoL9IGO6XxibgQ0zn2A:brtzilg) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L172)):*
> ```
>  get translocoInstructions(): string[] {
> ```
This should be transcelerator instructions.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 368 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLT5s_1vFQ7TPJi6KH:-MoLT5s_1vFQ7TPJi6KI:b-luw636) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L368)):*
> ```
>     invalidRows = invalidRows.filter(row => row.length > 1);
> ```
What happens if we do not filter for rows with length > 1?
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 100 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLNvqH6RrATc1_zzH0:-MoLNvqH6RrATc1_zzH1:b-8tgm50) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L100)):*
> ```
> "Note: Canceling import will stop importing questions but will not remove questions that are already imported."
> ```
I think this can improve in wording. "Cancelling will not delete questions that are already imported."

irahopkinson (changes requested):
This is looking really good, you have thought through it well, thanks for doing this. I have a few things that need attention.

> Reviewed 29 of 29 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-:-MoXN_6TA7-V0yJlVYT0:b-9r2g43)* status: all files reviewed, 27 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 155 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLNDXA0jnHDlwgG03C:-MoXH9PJCE3rvNFvmd94:bka0805) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L155)):*

> ```
>         <mdc-linear-progress [determinate]="true" [progress]="importedCount / toImportCount"></mdc-linear-progress>
> ```
angular material has linear progress bar now. It should be straight forward to use.
</blockquote></details>

Yep, note that `value` is percent rather than a fraction when you change.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 183 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXS6Zy1IW9B7oSVkEO:-MoXS6Zy1IW9B7oSVkEP:b-sz1w56) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L183)):*
> ```
> 1000
> ```
This number and `25` below are magical to me. How about some help understanding them?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 184 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXRrHLAA0YEZohbz0z:-MoXRrHLAA0YEZohbz1-:b-qupxzm) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L184)):*
> ```
> this.invalidRows!
> ```
Perhaps this could be initialized to an empty array so that we don't have to assert it as defined?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 222 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXSfsF3uAPcpUMF7d2:-MoXSfsF3uAPcpUMF7d3:bijxdhx) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L222)):*
> ```
> showDuplicateImportNote
> ```
This should be moved up with the other getters.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 343 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXU2Rs9bLKUj6EQgcm:-MoXU2Rs9bLKUj6EQgcn:b-4mhox7) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L343)):*
> ```
> fileSelected
> ```
If I put a range in the CSV file, e.g. MAT 1:2-3, for the reference it doesn't seem to set the _EndReference_ for the question. Can we support that?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 57 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXZMsg2Vg4pT4CnmGt:-MoXZMsg2Vg4pT4CnmGu:b-tb2b88) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L57)):*
> ```
> env.getButtonByText('Cancel')
> ```
It would still be better to use `env.cancelButton`, that way if the button text changes we only have to change it in one place. Same below for `ShowAllButton`, `Continue import`, and `Close`.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 202 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoX_7YP4SKwhD1fekET:-MoX_7YP4SKwhD1fekEU:b47wgk4) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L202)):*
> ```
> allows updates questions
> ```
nit: too many final `s`es. Perhaps 'allow updates to questions'? (There is also a double space before 'that'.)
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 327 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXbwRC4xq6anKYb-NO:-MoXbwRC4xq6anKYb-NP:b-fcbf4o) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L327)):*
> ```
> env.overlayContainerElement.querySelectorAll('table[mat-table] tbody tr'
> ```
Please put this in a getter to make it easier to read and maintain.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 332 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXcr1b8-p4D4h97zch:-MoXcr1b8-p4D4h97zci:b1fhr5x) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L332)):*
> ```
>     // Clean up the overlay container. Necessary because we can't do tick() so it's hard to clean up automatically.
>     // Not sure whether this is a good idea, but it stops the error message saying it wasn't cleaned up.
> ```
I'm guessing you tried doing `env.fixture.detectChanges();` after the last click? If so (or it doesn't help) then this comment is good for now.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 339 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXdmls4Io2W41_4_F3:-MoXdmls4Io2W41_4_F4:b45k8l) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L339)):*
> ```
> expect(env.bodyText).toBe('There are no questions for the books in this project.');
> ```
It would be good to demonstrate this doesn't show the rest of the time to make this a valid test.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 496 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXdUbB852imW8wyYQK:-MoXdUbB852imW8wyYQL:b9m0mfy) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L496)):*
> ```
> importFromTranscelerator
> ```
Add 'Button' to the end of the name to keep it consistent with the other names.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 539 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXf4RZ5g9Hb9cbKljX:-MoXf4RZ5g9Hb9cbKljY:bhbe4a6) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L539)):*
> ```
> as HTMLElement
> ```
Should be `HTMLButtonElement`?
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 104 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXKM-p1nfQsk7ov8fo:-MoXKM-p1nfQsk7ov8fp:bv2147w) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L104)):*
> ```
> Send and receive the project in Paratext.
> ```
Why do the CSV instructions include this step of S/R with PT? It seems unnecessary to require this.
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 129 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXMTHq9ZCnnTur-N_Q:-MoXMTHq9ZCnnTur-N_R:b-kzpgwo) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L129)):*
> ```
> \"Import from Transcelerator\"
> ```
According to our style guide, ideally the thing to click on would be bold rather than in speech marks (same for CSV instructions). I'm guessing that may be tricky given your current implementation. Is there some advantage to having the instructions all in one string? It would be fine not to use code to deploy them but to have a list item for each instruction. That could simplify things and allow you to format as needed.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/external-url.service.ts, line 7 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoX3DMC2OKjFO2zfIjU:-MoX3DMC2OKjFO2zfIjV:b-6qe81r) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/external-url.service.ts#L7)):*
> ```
> ExternalUrlService
> ```
Q: what does this do for us that `environment` doesn't?
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/external-url.service.ts, line 16 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoX2gTu6mKpAJxrLgT9:-MoX2gTu6mKpAJxrLgTA:bby0urd) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/external-url.service.ts#L16)):*
> ```
> en
> ```
This needs to be locale specific if the help exists. We do this already in the **Help** menu.

RaymondLuong3 (changes requested):
It appears that if I click "import from csv file", the file selection dialog appears. But when I select a csv file, nothing happens. I went to the question dialog and tried to upload an audio file, and it worked without issue. It looks like they both use the 'ngfSelect' directive correctly. I'm investigating to see what is going on, but so far no luck.

> Reviewed 18 of 29 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-:-Mo_2niPA-8ajJhGiR_2:bb4ph8c)* status: all files reviewed, 30 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 217 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoZZoJBDtdZj1sQHtpi:-MoZZoJBDtdZj1sQHtpj:b4mw945) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L217)):*
> ```
>     expect(env.editedTransceleratorQuestionIds).toEqual(['1', '4']);
> ```
It would be more intuitive if we could move the logic that results in this behaviour of question 1 and 4 being edited into this test. For instance, call a method called `simulateQuestionWithIdEdited('1')`. This way I wouldn't have to scroll down to the test environment class to confirm why this result is expected.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 321 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoZc32j2NN3uuzDvz6u:-MoZc32j2NN3uuzDvz6v:b-t6qqbc) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L321)):*
> ```
> const matQuestions = Array.from(Array(100), (_, i) => `MAT 1:${i + 1},Question for Matthew 1:${i + 1}`).join('\n');
> ```
If you want to show that only the 1 question from Genesis is on display, it would be worth checking that the one question has the expected verse ref.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 528 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoZgn0BDpJVFEiK2NcQ:-MoZgn0BDpJVFEiK2NcR:b25ffk4) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L528)):*
> ```
>   get importSelectedQuestionsButton(): HTMLElement {
>     return this.getButtonWhere(buttonText => /import \d+ selected questions/i.test(buttonText));
>   }
> ```
This seems complex to find the import button. I normally just add an id to the button element `<button id="importButton">` and then use the id to query for the button.

RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182)* status: all files reviewed, 31 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 56 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-Mo_a4wx4WHcfwkHFHTe:-Mo_a4wx4WHcfwkHFHTf:b-twyk4s) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L56)):*
> ```
> ngfSelect accept="text/csv" 
> ```
As we discovered, this needs to be updated to `accept=".csv"` instead of using the MIME type of text/csv for files created in Excel.

Nateowami (commented):
Thanks for your reviews. I've made a number of changes, but not addressed everything yet.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-:-MojRHS310ndV_gO1r1h:b-7czgcr)* status: all files reviewed, 26 unresolved discussions (waiting on @irahopkinson, @Nateowami, and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html, line 57 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoL1AWE63H4GaqxnNMV:-MoZXYGg3UOyaVCYxtZy:b-20ghfi) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L57)):*

> ```
> "urls.communityAnnouncementPages"
> ```
Glad to see these put in it's own file.
</blockquote></details>

I'm changing "Pages" to "Page" though. I think it was a mistake.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.ts, line 383 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoL79ix1j_Jgz7wAqHP:-Mog5q3EBEaV0X_Aqij-:b-6df11e) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.ts#L383)):*

> ```
>  this.matDialog.open(ImportQuestionsDialogComponent, { data });
> ```
Is there a reason this could not user mdc Dialog? Oh, I remember now. We are moving things back to angular material. Ok, I guess for the time being we have both in our app.
</blockquote></details>

Right. It's the things that are still MDC that are out of place and just haven't been changed yet.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 34 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoL9lCeD5o8JyteC8Lb:-MoZXQ7QDn2TrO1Mvoha:boj1qgt) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L34)):*

> ```
> target="_blank">Learn more</a>
> ```
This should be internationalized.
</blockquote></details>

Good catch. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 56 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-Mo_a4wx4WHcfwkHFHTe:-Mo_bJi43F5S9cxqIF5F:b-yhido8) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L56)):*

> ```
> ngfSelect accept="text/csv" 
> ```
As we discovered, this needs to be updated to `accept=".csv"` instead of using the MIME type of text/csv for files created in Excel.
</blockquote></details>

Raymond and I looked at this together and came to the conclusion that we should use `accept=".csv"`. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 155 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLNDXA0jnHDlwgG03C:-MoZVMIq7aLZyHfLz-ec:b-i21bbo) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L155)):*

Yep, note that `value` is percent rather than a fraction when you change.
</blockquote></details>

Thanks. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 176 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLP-yAD0uEYg6WIXX6:-MoZZOAz2ThgBrAyFmWJ:b-etaqep) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L176)):*

> ```
> <ng-container *ngIf="importCanceled"><br /><br />{{ t("canceling_import") }}</ng-container>
> ```
Is there a reason for using a container instead of a div? And I think it would be cleaner to use styling in css rather than <br /> tags.
</blockquote></details>

A `ng-container` doesn't add an element the DOM; that's why I used it rather than a `div`. But your point is well taken. I've changed it.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss, line 81 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLR2928AHdbaSfM1SQ:-MoaWA6G2LZxPS3hta62:b-6j20kt) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss#L81)):*

> ```
> .mat-column-rowNumber {
>   border-right: 1px solid rgba(0, 0, 0, 0.12);
>   padding-right: 1em;
>   text-align: right;
>   font-weight: 500;
> }
> 
> .mat-column-reference {
>   padding-left: 1em;
>   padding-right: 1em;
> }
> ```
I wonder if this styling may leak into other parts of the dialog. Do we have have the one data table. I see that we do define matColumnDef="rowNumber" and "matColumnDef="reference" so that may be enough.
</blockquote></details>

I don't think it will leak, as it's a fairly specific class that will probably only be applied in one place.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 100 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLS8Is4O2YiYrsVS_x:-MoaZla64AmL2cSrzxKX:bcuz34v) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L100)):*

> ```
>     readonly mdcDialog: MdcDialog,
>     readonly matDialog: MatDialog,
> ```
These look like they should be private
</blockquote></details>

They're public so they can be spied on in the tests. I looked into mocking and injecting these and haven't been able to figure out how to do it properly.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLSdlh6cngTqvv-K9x:-Mo_k5jCBsDdgm12IHvC:bneusku) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L165)):*

> ```
> return this.invalidRows!.
> ```
Is that a not null assertion? How do we know here it won't be null?
</blockquote></details>

We discussed this. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 172 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoL9IGO6XxibgQ0zn29:-Mo_k3snDaODjwX1gTTW:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L172)):*

> ```
>  get translocoInstructions(): string[] {
> ```
This should be transcelerator instructions.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 183 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXS6Zy1IW9B7oSVkEO:-MoaaDcKCqHLvLhtpMXt:b-xk2h0o) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L183)):*

> ```
> 1000
> ```
This number and `25` below are magical to me. How about some help understanding them?
</blockquote></details>

As the user scrolls, more list items are added (otherwise having thousands of items causes the dialog to stall).  When the user scrolls, we check whether we're within 1000 pixels of reaching the bottom of the scroll area. If so, 25 more items are added to the view.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 184 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXRrHLAA0YEZohbz0z:-MoaaAG326VXd8JYiXKs:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L184)):*

> ```
> this.invalidRows!
> ```
Perhaps this could be initialized to an empty array so that we don't have to assert it as defined?
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 222 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXSfsF3uAPcpUMF7d2:-Mo_kNh7-x9GafpFLRIm:b-i21bbo) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L222)):*

> ```
> showDuplicateImportNote
> ```
This should be moved up with the other getters.
</blockquote></details>

Thanks. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 368 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLT5s_1vFQ7TPJi6KH:-Moaby9k-RBGchlxv7ul:b5cikv7) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L368)):*

> ```
>     invalidRows = invalidRows.filter(row => row.length > 1);
> ```
What happens if we do not filter for rows with length > 1?
</blockquote></details>

We discussed this and made significant changes. The idea is to ignore lines that are blank, and inform the user about lines that are not blank and are not valid.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 57 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXZMsg2Vg4pT4CnmGt:-MoiXEf73M_v5wsclLz9:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L57)):*

> ```
> env.getButtonByText('Cancel')
> ```
It would still be better to use `env.cancelButton`, that way if the button text changes we only have to change it in one place. Same below for `ShowAllButton`, `Continue import`, and `Close`.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 202 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoX_7YP4SKwhD1fekET:-MoacEaB-wkC0ukRpo5d:b-i21bbo) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L202)):*

> ```
> allows updates questions
> ```
nit: too many final `s`es. Perhaps 'allow updates to questions'? (There is also a double space before 'that'.)
</blockquote></details>

Thanks. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 332 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXcr1b8-p4D4h97zch:-Mois2Yk6FSImHmh3UDv:b-c2ukx7) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L332)):*

> ```
>     // Clean up the overlay container. Necessary because we can't do tick() so it's hard to clean up automatically.
>     // Not sure whether this is a good idea, but it stops the error message saying it wasn't cleaned up.
> ```
I'm guessing you tried doing `env.fixture.detectChanges();` after the last click? If so (or it doesn't help) then this comment is good for now.
</blockquote></details>

Yes, I tried that, and much more. The issue is that this test is not in a `fakeAsync` because (I think):
- Papaparse tries to use a Web Worker (looking at the source code it has a condition checking whether it can use a web worker, and I don't know how that would work out in a test)
- It has to read a file, so presumably it is using a FileReader, which doesn't work with fakeAsync ([fakeAsync cannot work with FileReader event](https://github.com/angular/angular/issues/31702)).
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 496 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXdUbB852imW8wyYQK:-MoadVZ76F0YtO2p7mIK:b-i21bbo) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L496)):*

> ```
> importFromTranscelerator
> ```
Add 'Button' to the end of the name to keep it consistent with the other names.
</blockquote></details>

Thanks. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 528 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoZgn0BDpJVFEiK2NcQ:-Mog5h3k-YShISt2WMmx:blb8omt) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L528)):*

> ```
>   get importSelectedQuestionsButton(): HTMLElement {
>     return this.getButtonWhere(buttonText => /import \d+ selected questions/i.test(buttonText));
>   }
> ```
This seems complex to find the import button. I normally just add an id to the button element `<button id="importButton">` and then use the id to query for the button.
</blockquote></details>

Good point. I've changed this.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 539 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXf4RZ5g9Hb9cbKljX:-MofonYWBeJ9yYN8YQP3:b-3rl8mu) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L539)):*

> ```
> as HTMLElement
> ```
Should be `HTMLButtonElement`?
</blockquote></details>

It could be, but it doesn't particularly need to be, since we aren't using any button-specific properties or methods. I changed it though, because I do think it's better to be specific.
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 100 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLNvqH6RrATc1_zzH0:-MoaliV-9DYzvT1GvPfF:b-i21bbo) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L100)):*

> ```
> "Note: Canceling import will stop importing questions but will not remove questions that are already imported."
> ```
I think this can improve in wording. "Cancelling will not delete questions that are already imported."
</blockquote></details>

Thanks. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 104 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXKM-p1nfQsk7ov8fo:-MoameIy7A8DEgZhGnbO:be53yqf) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L104)):*

> ```
> Send and receive the project in Paratext.
> ```
Why do the CSV instructions include this step of S/R with PT? It seems unnecessary to require this.
</blockquote></details>

Not sure how I missed this. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 129 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXMTHq9ZCnnTur-N_Q:-MoefTL--8X63OShriKt:b-hixfhj) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L129)):*

> ```
> \"Import from Transcelerator\"
> ```
According to our style guide, ideally the thing to click on would be bold rather than in speech marks (same for CSV instructions). I'm guessing that may be tricky given your current implementation. Is there some advantage to having the instructions all in one string? It would be fine not to use code to deploy them but to have a list item for each instruction. That could simplify things and allow you to format as needed.
</blockquote></details>

The advantage is that the translators can see that the list items are together and translate them in context of each other. But good point about style. Here's how it looks now:


I also changed how the internationalization happens so that the text in the last list item is whatever is shown on the button below (otherwise there's a possibility that they get translated differently and don't actually match).
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/external-url.service.ts, line 7 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoX3DMC2OKjFO2zfIjU:-MoigsEs2ODR44wSOOIV:b-3mfajp) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/external-url.service.ts#L7)):*

> ```
> ExternalUrlService
> ```
Q: what does this do for us that `environment` doesn't?
</blockquote></details>

The `environment` variable seems like something that is supposed to give us values that change from one environment to another (or that have the potential to differ between environments at some point in the future).

The `ExternalUrlService` is specifically for bringing all URLs for external sites into one place so we don't have magic strings throughout our views. This also makes it easier in the future if we change the name of a help page, or the format of the URLs for the help pages, or moved the community announcement page, or anything else of the sort.

Also it now makes sure the correct locale is in the URL for the helps pages (which it didn't when you originally wrote this comment).
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/external-url.service.ts, line 16 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoX2gTu6mKpAJxrLgT9:-Moas6KC2ClJBZhatgpd:bl8lja6) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/external-url.service.ts#L16)):*

> ```
> en
> ```
This needs to be locale specific if the help exists. We do this already in the **Help** menu.
</blockquote></details>

Good point. Fixed.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182)* status: 19 of 29 files reviewed, 26 unresolved discussions (waiting on @irahopkinson, @Nateowami, and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 195 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MojSbr1-J_xFkvPwyIj:-MojSbr1-J_xFkvPwyIk:b-aif971) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L195)):*
> ```HTML
>       {{ t("continue_import") }}
>     </button>
>     <button *ngIf="status === 'filter'" mat-flat-button color="primary" cdkFocusInitial (click)="importQuestions()">
> ```

I removed `cdkFocusInitial` because this button doesn't need to be automatically focused (it doesn't even exist yet when the dialog is opened, so it doesn't even make sense anymore).
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 321 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoZc32j2NN3uuzDvz6u:-MolNvlSEoSecUBUk1er:b-qu6n94) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L321)):*

> ```
> const matQuestions = Array.from(Array(100), (_, i) => `MAT 1:${i + 1},Question for Matthew 1:${i + 1}`).join('\n');
> ```
If you want to show that only the 1 question from Genesis is on display, it would be worth checking that the one question has the expected verse ref.
</blockquote></details>

Good point. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 327 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXbwRC4xq6anKYb-NO:-Mojqnj18_pKtaYHAZeE:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L327)):*

> ```
> env.overlayContainerElement.querySelectorAll('table[mat-table] tbody tr'
> ```
Please put this in a getter to make it easier to read and maintain.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 339 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXdmls4Io2W41_4_F3:-Mojozxt0J4npUV2uMU2:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L339)):*

> ```
> expect(env.bodyText).toBe('There are no questions for the books in this project.');
> ```
It would be good to demonstrate this doesn't show the rest of the time to make this a valid test.
</blockquote></details>

Done.

RaymondLuong3 (changes requested):
> Reviewed 8 of 10 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182)* status: all files reviewed, 16 unresolved discussions (waiting on @irahopkinson and @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 176 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLP-yAD0uEYg6WIXX6:-MonqJN57EpEf0Cvj6YQ:bruue6k) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L176)):*

A `ng-container` doesn't add an element the DOM; that's why I used it rather than a `div`. But your point is well taken. I've changed it.
</blockquote></details>

That's neat. I wasn't aware that `ng-container` does not add an element. I'll have to experiment with that next time I work in html
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 100 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLS8Is4O2YiYrsVS_x:-MonqgPaCMdN0sJhB0jy:b146efv) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L100)):*

They're public so they can be spied on in the tests. I looked into mocking and injecting these and haven't been able to figure out how to do it properly.
</blockquote></details>

Ok, that is reasonable. We will try to limit specifically exposing properties to the tests where we can. But it is beneficial in this case so we can spy on the dialogs.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 362 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-Monsdsu7t1jLbw38skc:-Monsdsu7t1jLbw38skd:b-238k78) ([raw file](https://github.com/sillsdev/web-xforge/blob/0fdc9dfdbda86f02ed201a64f3ae3ba018cb1105/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L362)):*

> ```
>       const rowNumber: string = index + 1 + '';
>       const reference = row[0];
>       const questionText = row[1];
>       if (row.length < 2 || reference === '' || questionText === '') {
>         invalidRows.push([rowNumber].concat(row));
>         continue;
>       }
>       try {
>         const verse = VerseRef.parse(reference);
> ```
</details>

These types are not explicit, so lets add the type for these vars.
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 104 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MonuEkY040fzhlhABU8:-MonuEkY040fzhlhABU9:b-kcp1d1) ([raw file](https://github.com/sillsdev/web-xforge/blob/0fdc9dfdbda86f02ed201a64f3ae3ba018cb1105/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L104)):*
> ```
> \n
> ```
I believe that we can use {{ newLine }} instead of \n.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182)* status: 26 of 29 files reviewed, 15 unresolved discussions (waiting on @irahopkinson, @Nateowami, and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 343 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXU2Rs9bLKUj6EQgcm:-Mp8kQdJ-UrHWDzLlSKK:b-c0v34e) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L343)):*

> ```
> fileSelected
> ```
If I put a range in the CSV file, e.g. MAT 1:2-3, for the reference it doesn't seem to set the _EndReference_ for the question. Can we support that?
</blockquote></details>

Thanks. Somehow I missed this. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 362 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-Monsdsu7t1jLbw38skc:-MouL5qe223nZsLxk7yG:bo3vl6y) ([raw file](https://github.com/sillsdev/web-xforge/blob/0fdc9dfdbda86f02ed201a64f3ae3ba018cb1105/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L362)):*

> ```
>       const rowNumber: string = index + 1 + '';
>       const reference = row[0];
>       const questionText = row[1];
>       if (row.length < 2 || reference === '' || questionText === '') {
>         invalidRows.push([rowNumber].concat(row));
>         continue;
>       }
>       try {
>         const verse = VerseRef.parse(reference);
> ```
These types are not explicit, so lets add the type for these vars.
</blockquote></details>

Thanks, done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 104 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MonuEkY040fzhlhABU8:-Mou74XZ-IytTntbxtjn:b-xv175o) ([raw file](https://github.com/sillsdev/web-xforge/blob/0fdc9dfdbda86f02ed201a64f3ae3ba018cb1105/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L104)):*

> ```
> \n
> ```
I believe that we can use {{ newLine }} instead of \n.
</blockquote></details>

Interesting. I was not aware of that. I wonder though whether that might be more difficult for translators to understand and correctly translate this:
```
Create a spreadsheet with a column for the reference and a column for the question, as shown below:{{ newLine }}Export or save the spreadsheet as a CSV file.{{ newLine }}Click {{ boldStart }}{{ importFromCsvFile }}{{ boldEnd }} below and select the file.
```
Than this:
```
Create a spreadsheet with a column for the reference and a column for the question, as shown below:
Export or save the spreadsheet as a CSV file.
Click {{ boldStart }}{{ importFromCsvFile }}{{ boldEnd }} below and select the file.
```
The `\n` will never be seen by the translators, and makes it clearer that this is a list of instructions (they don't necessarily know that there are bullet points).

I've also noticed that we have some strings with `{{ newLine }}String goes here`. That seems quite odd; if we need anything at the beginning or end of a string we don't actually need it to be internationalized.

irahopkinson (changes requested):
> Reviewed 7 of 10 files at r2, 1 of 2 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182)* status: all files reviewed, 4 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 183 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXS6Zy1IW9B7oSVkEO:-MpB0ZoK4t1Ay5CpBCxY:b-nsupbr) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L183)):*

As the user scrolls, more list items are added (otherwise having thousands of items causes the dialog to stall).  When the user scrolls, we check whether we're within 1000 pixels of reaching the bottom of the scroll area. If so, 25 more items are added to the view.
</blockquote></details>

How about assigning both to a meaningfully named `const` or add a comment then?
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 129 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXMTHq9ZCnnTur-N_Q:-MpB3vNa03OKhCLfDvkU:bskb4tl) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L129)):*

The advantage is that the translators can see that the list items are together and translate them in context of each other. But good point about style. Here's how it looks now:


I also changed how the internationalization happens so that the text in the last list item is whatever is shown on the button below (otherwise there's a possibility that they get translated differently and don't actually match).
</blockquote></details>

Nice improvement getting the button names consistent.

RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 104 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MonuEkY040fzhlhABU8:-MpChBAp11yffUSc59U9:b-uzxxsc) ([raw file](https://github.com/sillsdev/web-xforge/blob/0fdc9dfdbda86f02ed201a64f3ae3ba018cb1105/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L104)):*

Interesting. I was not aware of that. I wonder though whether that might be more difficult for translators to understand and correctly translate this:
```
Create a spreadsheet with a column for the reference and a column for the question, as shown below:{{ newLine }}Export or save the spreadsheet as a CSV file.{{ newLine }}Click {{ boldStart }}{{ importFromCsvFile }}{{ boldEnd }} below and select the file.
```
Than this:
```
Create a spreadsheet with a column for the reference and a column for the question, as shown below:
Export or save the spreadsheet as a CSV file.
Click {{ boldStart }}{{ importFromCsvFile }}{{ boldEnd }} below and select the file.
```
The `\n` will never be seen by the translators, and makes it clearer that this is a list of instructions (they don't necessarily know that there are bullet points).

I've also noticed that we have some strings with `{{ newLine }}String goes here`. That seems quite odd; if we need anything at the beginning or end of a string we don't actually need it to be internationalized.
</blockquote></details>

Our format markers in the i18n files should be consistent across our strings. That would be the most helpful thing for the translators. But you do bring up a good point that it looks clearer for the translator.
However, I am inclined to say that the `{{ newLine }}` approach has the advantage that a translator will be more likely to get the translation formatting the way we expect. It won't be up to the translator to know that you should put a \n at the correct positions in the translation. They will see the english text as one long string, and will translate it as one long string.

irahopkinson (approved):
:lgtm: awaiting @raymondluong3 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-:-MpEZMweA4jDvDCouKK0:b3kfg85)* status: 27 of 31 files reviewed, 2 unresolved discussions (waiting on @irahopkinson, @Nateowami, and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 104 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MonuEkY040fzhlhABU8:-MpEXqm1CZAPr_LKVhm_:b-goj139) ([raw file](https://github.com/sillsdev/web-xforge/blob/0fdc9dfdbda86f02ed201a64f3ae3ba018cb1105/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L104)):*

Our format markers in the i18n files should be consistent across our strings. That would be the most helpful thing for the translators. But you do bring up a good point that it looks clearer for the translator.
However, I am inclined to say that the `{{ newLine }}` approach has the advantage that a translator will be more likely to get the translation formatting the way we expect. It won't be up to the translator to know that you should put a \n at the correct positions in the translation. They will see the english text as one long string, and will translate it as one long string.
</blockquote></details>

I can see this both ways. I believe the carriage returns will show in crowdin as such. Also `{{ newLine }}` is still useful in other contexts. Given @nateowami 's experience with crowdin and responding to users there, how about we go with this current approach for lists like this unless we see issues with CRs being removed from the crowdin results.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182)* status: 27 of 31 files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 100 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLS8Is4O2YiYrsVS_x:-MpD615zD8IbxmScGH9Z:bbrhwnv) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L100)):*

Ok, that is reasonable. We will try to limit specifically exposing properties to the tests where we can. But it is beneficial in this case so we can spy on the dialogs.
</blockquote></details>

I'm quite confident there should be a way around this, but when I've tried I haven't been successful.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 183 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoXS6Zy1IW9B7oSVkEO:-MpD69029gi853dmt_jV:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L183)):*

How about assigning both to a meaningfully named `const` or add a comment then?
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts, line 217 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoZZoJBDtdZj1sQHtpi:-MpD6ACH8phwwl02jcQo:b9865gt) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts#L217)):*

> ```
>     expect(env.editedTransceleratorQuestionIds).toEqual(['1', '4']);
> ```
It would be more intuitive if we could move the logic that results in this behaviour of question 1 and 4 being edited into this test. For instance, call a method called `simulateQuestionWithIdEdited('1')`. This way I wouldn't have to scroll down to the test environment class to confirm why this result is expected.
</blockquote></details>

Thanks. Done. Although still not very pretty, because it needs to be done at the point of creating the TestEnvironment, which already has an unwieldy constructor (too many arguments available).

irahopkinson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182)* status: 27 of 31 files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 100 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLS8Is4O2YiYrsVS_x:-MpFV49S1nw2GiZsvoO9:b-grceep) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L100)):*

I'm quite confident there should be a way around this, but when I've tried I haven't been successful.
</blockquote></details>

For an example, see how we do it in `auth.service.spec.ts` L465 and L468. Although that is to call a private method - I imagine it should also work to read a private property.

RaymondLuong3 (approved):
Thanks for your work! I approve.

> Reviewed 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-:-MpHh0MZEwIexNVZ0YPm:b-4ysvro)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182)* status: 29 of 31 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts, line 100 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-MoLS8Is4O2YiYrsVS_x:-MpHt53E4RMUB-kPzXUI:b-y15f6o) ([raw file](https://github.com/sillsdev/web-xforge/blob/f288f691b4c335c63fe9703d80de8ebe53897b6d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L100)):*

For an example, see how we do it in `auth.service.spec.ts` L465 and L468. Although that is to call a private method - I imagine it should also work to read a private property.
</blockquote></details>

I figured it out. The complexity is that this is a dialog that opens other dialogs. The tests need to actually open the main dialog, but mock the sub dialogs.

Right now it's pretty simple because we have a MatDialog that opens MdcDialog. So the solution is to mock MdcDialog and not mock MatDialog. However, in the past it was an MdcDialog opening MdcDialog, and in the future it will be a MatDialog opening MatDialog. I think the solution is something like:

``` ts
@Injectable()
class RealMatDialog extends MatDialog {}

configureTestingModule(() => ({
  providers: [{ provide: RealMatDialog, useClass: MatDialog }]
}));

// In TestEnvironment:
TestBed.inject(RealMatDialog).open(ImportQuestionsDialogComponent, config);
```

Nateowami (commented):
@raymondluong3 Thanks for your review. It's now ready for re-review with minor changes.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-:-MpIAwWU-VuoFnjw_u7M:b-28x5p3)* status: 29 of 31 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

irahopkinson (approved):
:lgtm:

> Reviewed 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1182#-:-MpK0k9aCXN0iPb7bPXk:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1183, opened 2021-11-10
Fix login redirect with 3rd party cookies

nigel-wells:
Based on some recent feedback to the login process around redirect issues with 3rd party cookies, as well as some of the other login redirect related issues, I've reworked some of the login process.

For context, when the app first loads if settings are available in local storage then the user will be logged in "offline". Once the app is fully loaded a background check is made (as long as they're online) with auth0 to get an updated token. When 3rd party cookies are blocked it means any check with auth0 will return "login required" and a redirect to auth0 takes place. 

After doing the recent auth service tests I noticed that the above was the intention but the reality is slightly different. When auth0 returns the user to SF it stores the result of the parsed hash in a promise. From then on any checks with auth0 simply check that result rather than actually checking with auth0 again. If the page was to refresh then the parsed hash would be lost and those with 3rd party cookies blocked would be redirected to auth0 again. Anyone using a browser without 3rd party cookies blocked would be fine as it would do the intended check with auth0 in the background.

Because of the above I felt safe to change our "check with auth0" code once the app loads to simply ensure the token expire scheduler is running. Auth0 provides us an expiry date/time for when the token expires so we really don't need to check with auth0 again until that expires. The proposed changes does just that. This appears to resolve both the redirect issue as well as the refresh issue.

The end goal is still to rework the auth0 flow to remove reliance on 3rd party cookies so this PR is to help cover us until we get there.

Interested in others feedback.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1183)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth-http-interceptor.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth-http-interceptor.ts
@@ -25,7 +25,9 @@ export class AuthHttpInterceptor implements HttpInterceptor {
       this.authService = this.injector.get<AuthService>(AuthService);
     }
     // Make sure the user is authenticated with a valid access token
-    await this.authService.isAuthenticated();
+    if (!(await this.authService.isAuthenticated())) {
+      return await throwError('User not authenticated - login required').toPromise();
+    }
     // Add access token to the request header
     const authReq = req.clone({
       headers: req.headers.set('Authorization', 'Bearer ' + this.authService.accessToken)
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts
@@ -91,30 +91,38 @@ describe('AuthService', () => {
   }));
 
   it('should check online authentication if logged in', fakeAsync(() => {
-    const env = new TestEnvironment();
+    const env = new TestEnvironment({ isOnline: true, isLoggedIn: true });
+
     expect(env.isLoggedIn).withContext('setup').toBe(true);
-    resetCalls(mockedPwaService);
-
-    env.service.checkOnlineAuth();
-
-    tick();
     verify(mockedPwaService.checkOnline()).once();
+    verify(mockedWebAuth.checkSession(anything(), anything())).once();
     verify(mockedWebAuth.authorize(anything())).never();
+    env.discardTokenExpiryTimer();
   }));
 
   it('check session is valid after returning online and login if session has expired', fakeAsync(() => {
     const env = new TestEnvironment({
-      isOnline: true,
+      isOnline: false,
       isLoggedIn: true
     });
     expect(env.isAuthenticated).toBe(true);
-    resetCalls(mockedPwaService);
+    verify(mockedWebAuth.authorize(anything())).never();
 
     env.setLoginRequiredResponse();
-    env.service.checkOnlineAuth();
-    tick();
+    env.setOnline();
+    // Should still be true as the expiry set is still in the future
+    expect(env.isAuthenticated).toBe(true);
+    verify(mockedWebAuth.authorize(anything())).never();
 
-    verify(mockedPwaService.checkOnline()).once();
+    env.setOnline(false);
+    env.service.expireToken();
+    // Should still be authenticated as we were logged in and are now offline so can't do a renewal
+    expect(env.isAuthenticated).toBe(true);
+    verify(mockedWebAuth.authorize(anything())).never();
+
+    env.setOnline();
+    // Should now renew tokens as expired timer is reached
+    expect(env.isAuthenticated).toBe(false);
     verify(mockedWebAuth.authorize(anything())).once();
   }));
 
@@ -154,9 +162,14 @@ describe('AuthService', () => {
   }));
 
   it('should authenticate if expired', fakeAsync(() => {
-    const env = new TestEnvironment();
+    const env = new TestEnvironment({ isOnline: true, isLoggedIn: true });
     expect(env.isAuthenticated).toBe(true);
     verify(mockedWebAuth.checkSession(anything(), anything())).once();
+    resetCalls(mockedWebAuth);
+
+    env.clearTokenExpiryTimer();
+    verify(mockedWebAuth.checkSession(anything(), anything())).once();
+    env.discardTokenExpiryTimer();
   }));
 
   it('should renew tokens if expired and authenticating', fakeAsync(() => {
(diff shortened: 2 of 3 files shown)


irahopkinson (changes requested):
Looks good, thanks for following through on this. Just 2 small clean-ups and a question.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1183#-:-MoWr4ogCjDgAe6Zpc9w:b-1pq2ek)* status: all files reviewed, 3 unresolved discussions (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 326 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1183#-MoWZ84f3_uVQkz8DZz8:-MoWZ84f3_uVQkz8DZz9:bjtym4b) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc79acb26e5bc40968ffadaa7b3226b7f2cf1748/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L326)):*
> ```
> return true;
> ```
Remove this. I think I missed this clean-up when I simplified the call to this method. Is this still the right name for this method?
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 429 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1183#-MoWlGUg2zU211VGzAVp:-MoWlGUg2zU211VGzAVq:bmzjelo) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc79acb26e5bc40968ffadaa7b3226b7f2cf1748/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L429)):*
> ```XML
> 
>     // Expiry 30 seconds sooner than the actual expiry date to avoid any inflight expiry issues
>     const expiresAt = (expiresIn - 30) * 1000 + Date.now();
> ```

What happens if `expiresIn` is 30 or less?
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 105 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1183#-MoWmjys0jD6FqeGUF-q:-MoWmjys0jD6FqeGUF-r:bgxu7j4) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc79acb26e5bc40968ffadaa7b3226b7f2cf1748/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L105)):*

> ```
>     const env = new TestEnvironment({
>       isOnline: false,
>       isLoggedIn: true
>     });
> ```
</details>

nit style: this could all be on one line.

nigel-wells (commented):
Done. One scenario not covered as I couldn't consistantly replicate it (perhaps because the changes coincidently fixed it?) is the occasional `checkUserNeedsMigrating` dialog. I added some code to attempt to deal with that but I'm not sure. I did try testing the scenario but couldn't trigger it. The times I have seen it is when using the app after a long period of time and my token has expired. The dialog briefly appears and then you're directed to auth0.

We probably could add another tidy up task to remove beta at some stage but that is potentially just hiding the bug for another day :)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1183#-:-MoXNo8W4jde_A3CGvu6:b9md8n3)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 326 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1183#-MoWZ84f3_uVQkz8DZz8:-MoXMQknAhIW7pEI7WY-:b-e8z8b0) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc79acb26e5bc40968ffadaa7b3226b7f2cf1748/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L326)):*

> ```
> return true;
> ```
Remove this. I think I missed this clean-up when I simplified the call to this method. Is this still the right name for this method?
</blockquote></details>

Done. It isn't even needed any more to be honest now that it has been cut right back. I've moved the one line of code in there to `tryLogIn()` which is where it was called from.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 429 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1183#-MoWlGUg2zU211VGzAVp:-MoXLTMW0-Rrhyv2NTB9:b-48x1jb) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc79acb26e5bc40968ffadaa7b3226b7f2cf1748/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L429)):*

What happens if `expiresIn` is 30 or less?
</blockquote></details>

That scenario is taken care of in `scheduleRenewal()` which ensures the timer is at least 1. This covers cases where the expiry date is well and truly expired i.e. it expired overnight and then you try to use the app in the morning.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 105 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1183#-MoWmjys0jD6FqeGUF-q:-MoXMDHB4JEIsSK6wK8H:bxpirej) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc79acb26e5bc40968ffadaa7b3226b7f2cf1748/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L105)):*

> ```
>     const env = new TestEnvironment({
>       isOnline: false,
>       isLoggedIn: true
>     });
> ```
nit style: this could all be on one line.
</blockquote></details>

Guess I assumed prettier would sort all those out. Have fixed.

irahopkinson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1183#-:-MoXXcVXADzNgaQ5xvOy:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts, line 429 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1183#-MoWlGUg2zU211VGzAVp:-MoXXDE3399fynFSXdp1:b-tocur2) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc79acb26e5bc40968ffadaa7b3226b7f2cf1748/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L429)):*

That scenario is taken care of in `scheduleRenewal()` which ensures the timer is at least 1. This covers cases where the expiry date is well and truly expired i.e. it expired overnight and then you try to use the app in the morning.
</blockquote></details>

Right, I can see that now.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts, line 105 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1183#-MoWmjys0jD6FqeGUF-q:-MoXXJBF3Lahh1kenJab:b-p1gpqd) ([raw file](https://github.com/sillsdev/web-xforge/blob/bc79acb26e5bc40968ffadaa7b3226b7f2cf1748/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L105)):*

Guess I assumed prettier would sort all those out. Have fixed.
</blockquote></details>

Prettier won't tidy up if you intentionally add style like CR, which is actually nice when you want it.

irahopkinson:
We just need to get rid of the `checkUserNeedsMigrating` now.

PR #1191, opened 2021-11-17
Gracefully abort handling of HttpRequest intercepts when not authenticated

nigel-wells:
This addresses a problem that would occur when auth0 issued a login required error either by the session expiring or from 3rd party cookies being blocked. If this instance occurred as well as the access token expiry (captured on first login and stored in local storage) also expiring AND the app first loaded OR the app loaded while offline and then came online an error dialog would occur due to the `checkUserNeedsMigrating` check that is run when the app first loads. 

This issue was minimized in a recent update that ensured if the access token had expired, and the user was online, it would check immediately and the user would be redirected to auth0 before the main app would ever need to run `checkUserNeedsMigrating`.

This is the most common scenario but the same dialog could have occurred when other checks with the backend are taking place and a 401 response comes back.

The dialog would only ever appear for a couple of seconds because when auth0 responds with a login required error the app immediately starts redirecting but while waiting for the redirect to occur SF continues with the reset of its execution and displays the error.

This fix in the http interceptor simple keeps the request in limbo using `EMPTY` which means the observable never confirms to any subscriptions that it completes resulting in the app sitting and waiting patiently for the redirect to occur.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1191)

Currently on `master` we have the auth0 tenant to force a login after 2 minutes or 1 minute of no activity with auth0. There is another setting that only expires our access token after 5 minutes which makes it a little tricky to test. The best way to replicate is to:
1. Build PWA
2. Login
3. Go offline
4. Wait 1 minute
5. Edit local storage and change `expires_at` to something low - just deleting one of the numbers is enough
6. Refresh the page making the PWA reload the app in offline mode
7. Go online

Current `master` will show the dialog whereas this branch will take you to auth0 without showing an error dialog.

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth-http-interceptor.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth-http-interceptor.spec.ts
@@ -1,0 +1,154 @@
+import { HTTP_INTERCEPTORS, HttpClient, HttpErrorResponse, HttpRequest } from '@angular/common/http';
+import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing';
+import { fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { mock, verify, when } from 'ts-mockito';
+import { AuthService } from 'xforge-common/auth.service';
+import { AUTH_APIS, AuthHttpInterceptor } from 'xforge-common/auth-http-interceptor';
+import { CommandErrorCode, JsonRpcResponse } from 'xforge-common/command.service';
+import { configureTestingModule } from 'xforge-common/test-utils';
+import { COMMAND_API_NAMESPACE, PROJECTS_URL, USERS_URL } from 'xforge-common/url-constants';
+
+const mockedAuthService = mock(AuthService);
+
+describe('AuthHttpInterceptor', () => {
+  configureTestingModule(() => ({
+    imports: [HttpClientTestingModule],
+    providers: [
+      { provide: HTTP_INTERCEPTORS, useClass: AuthHttpInterceptor, multi: true },
+      { provide: AuthService, useMock: mockedAuthService }
+    ]
+  }));
+
+  it('Sets authorization bearer only for valid auth apis', fakeAsync(() => {
+    const env = new TestEnvironment({ isAuthenticated: true });
+    const tests: { url: string; authorizationRequired: boolean }[] = [USERS_URL, PROJECTS_URL]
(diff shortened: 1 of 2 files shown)


irahopkinson (commented):
I see we don't have any tests for this. This fixes the problem at hand however what would happen if something goes wrong with the redirect? I guess the user would hang forever. Would it be better to await a timeout in which a redirect should have finished and then still throw the error?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1191#-:-MogHCurEhjZKPttrr6u:b-lgupl9)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

nigel-wells (commented):
No tests exist yet for this file - I could write some up though or perhaps wait until the new auth0 flow? Do we have any errors in bugsnag to indicate the redirect wouldn't work? It could be a connection issue or if someone kept pressing escape to stop the redirect happening. We'd need quite a different error dialog and then also force offline mode otherwise there will be heaps of connection issues occuring everytime the app tries to connect to the backend. There is also no way to know how long the redirect should take as it depends on the connection to auth0 from the client machine. From expierence it has been a few seconds maximum. The same situation applies to trying to login as what happens if that redirect doesn't happen. I think a separate task should be created to address handling auth0 connection issues.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1191#-:-MogMu2S9C-vxXb_ksE6:b-kn7s3f)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

nigel-wells (commented):
I've created some tests for this - let me know what you think.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1191#-:-MoqF4_h0fTaDBcgjCsA:bmfwff2)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @irahopkinson)

irahopkinson (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1191)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

irahopkinson (approved):
:lgtm: Thanks for the tests.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1191#-:-MpKAlGp2Kyej57E-SL7:bnxa263)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

PR #1196, opened 2021-11-24
SF-1417d Order drag-and-drop activity like user

marksvc:
- text.component.ts update(), editor.component.ts onTargetUpdated(),
and updateVerseNoteThreadAnchors() run with the assumption that the
editor's selection is in the current segment, and that the incoming
delta ops regard that segment. This has not been so for
drag-and-drop, where text is inserted in a destination segment while
there may be a selection in a different segment.
- Modify drag-and-drop to (optionally) remove the selection, set the
insertion point in the destination, and then insert text. Let the
activity be interleaved with TextComponent.updated events. Behaving
more like a user in this regard allows TextComponent and
EditorComponent to continue their assumptions. This allows threads to
be re-anchored correctly. Deletes happen in the segment with text
being deleted. In this way, the source segment of a drag-and-drop
has its anchors updated even if another segment is the target.

====

This PR makes use of `setTimeout()` to allow interleaving multiple programmatic edits with multiple TextComponent.updated events, rather than having all the TextComponent.updated events happen at once at the end (and do the wrong things).

This PR also has a line saying `determineInsertionPosition(...) ?? -1`. 

Ira previously commented about the `determineInsertionPosition(...) ?? -1`, but it was before the justification was uploaded, which is for the setTimeouts(). There may be another way to make it neat, but allowing `insertionPositionInDocument` to by type `number` instead of `number|undefined` means we can later skip probably 3 times when we would otherwise need to do `if (insertionPositionIndocument != null)`, in the `setTimeout()`s. 

Hmm, perhaps another approach could be to add another variable into the mix, so we have a temporary
```ts
const temporary: number | undefined = this.determineInsertionPosition(...)
```
and then set it to the other variable
```ts
if (temporary == null) {
  :-/
  return
}
const insertionPositionInDocument = temporary
```
I won't worry about that too much until the `setTimeouts()` are considered tho.

Here is a 20 s animation showing some dragging into and out of verses with note markers. This is perhaps slightly misleading since the problem may not show up unless you navigate away and back to the chapter. But when I later did that, the flags are in the right place. And I also did a sync and the flags are still in the right place. 

dnd-w-flags

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1196)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
@@ -54,40 +54,67 @@ export class DragAndDrop {
         return;
       }
 
-      let insertionPositionInDocument: number | undefined = this.determineInsertionPosition(
+      const insPos: number | undefined = this.determineInsertionPosition(
         targetElement,
         destinationSegmentRange,
         droppingIntoBlankSegment,
         dragEvent.clientX,
         dragEvent.clientY
       );
-      if (insertionPositionInDocument == null) {
+      if (insPos == null) {
         return;
       }
+      let insertionPositionInDocument: number = insPos;
 
-      let newText: string = dragEvent.dataTransfer.getData('text/plain');
-      // Replace newlines with a space.
-      newText = newText.replace(/(?:\r?\n)+/g, ' ');
-      quill.insertText(insertionPositionInDocument, newText, 'user');
-
-      // Identify the selection range, if any, now that we updated the document with an insert. This will be a
+      // Identify the selection range, if any. This will be a
       // selection that was already present before the drop, whether it is the text that was dragged, or a selection
       // not related to the drag.
       const originalSelection: RangeStatic | null = quill.getSelection();
 
-      // Select the inserted text.
-      quill.setSelection(insertionPositionInDocument, newText.length);
+      let newText: string = dragEvent.dataTransfer.getData('text/plain');
+      // Replace newlines with a space.
+      newText = newText.replace(/(?:\r?\n)+/g, ' ');
 
+      let quillIsSource: boolean = dragEvent.dataTransfer.types.includes(DragAndDrop.quillIsSourceToken);
+      let userIsHoldingCtrlKey: boolean = dragEvent.ctrlKey;
       if (originalSelection != null) {
         // There was a selection before the drop occurred.
-        if (dragEvent.dataTransfer.types.includes(DragAndDrop.quillIsSourceToken) && !dragEvent.ctrlKey) {
+        if (quillIsSource && !userIsHoldingCtrlKey) {
           // If the drag was started from within quill, then treat the selection as the source data of the drag, and
           // delete the selection. Unless the user was holding the ctrl key to copy text instead of move it.
           quill.deleteText(originalSelection.index, originalSelection.length, 'user');
+          // Adjust insertion position accordingly if preceding text was just deleted
+          if (insertionPositionInDocument > originalSelection.index) {
+            insertionPositionInDocument -= originalSelection.length;
+          }
         }
         // Or if the drag was not started from within quill, or the user was holding the ctrl key, then don't delete
         // the selection.
       }
+
+      // Behave like a user, to help editor event processing, by cutting the source text first (if applicable),
+      // putting the insertion point in at the target, and then inserting. Give an opportunity between each of these
+      // for EditorComponent.onTargetUpdated() to run, so that thread anchors are updated even in a source verse that
+      // is dragged out of. Use of setTimeout() allows TextComponent.updated to emit and be received by subscribers
+      // before we take further action here. Therefore, TextComponent.updated events are interleaved with our calls to
+      // quill.
+
+      setTimeout(() => {
+        quill.setSelection(insertionPositionInDocument, 0);
+        setTimeout(() => {
+          quill.insertText(insertionPositionInDocument, newText, 'user');
+          setTimeout(() => {
+            // If we inserted into a blank segment, and let SF respond by removing the blank in between our quill
+            // changes, then the position of insertion needs to move back by 1.
+            const lengthOfBlank: number = 1;
+            if (droppingIntoBlankSegment) {
+              insertionPositionInDocument -= lengthOfBlank;
+            }
+            // Select the inserted text.
+            quill.setSelection(insertionPositionInDocument, newText.length);
+          }, 1);
+        }, 1);
+      }, 1);
     });
   }
 
@@ -117,7 +144,7 @@ export class DragAndDrop {
       // We need to be able to know where to insert the dropped text, such as from the drop target being a usx-segment
       // element. Give up.
       console.warn('Warning: DragEvent to invalid target:', targetElement);
-      return [null, droppingIntoBlankSegment];
+      return [targetElement, droppingIntoBlankSegment];
     }
     return [targetElement, droppingIntoBlankSegment];
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts
@@ -1,7 +1,7 @@
 import { CommonModule } from '@angular/common';
 import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { Component, ViewChild } from '@angular/core';
-import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { ComponentFixture, fakeAsync, flush, TestBed, tick } from '@angular/core/testing';
 import { TranslocoService } from '@ngneat/transloco';
 import Quill, { RangeStatic } from 'quill';
 import { SFProject } from 'realtime-server/lib/esm/scriptureforge/models/sf-project';
@@ -9,7 +9,7 @@ import { TextAnchor } from 'realtime-server/lib/esm/scriptureforge/models/text-a
 import { TextData } from 'realtime-server/lib/esm/scriptureforge/models/text-data';
 import { VerseRef } from 'realtime-server/lib/esm/scriptureforge/scripture-utils/verse-ref';
 import * as RichText from 'rich-text';
-import { BehaviorSubject } from 'rxjs';
+import { BehaviorSubject, Subscription } from 'rxjs';
 import { anything, mock, when } from 'ts-mockito';
 import { AvatarTestingModule } from 'xforge-common/avatar/avatar-testing.module';
 import { BugsnagService } from 'xforge-common/bugsnag.service';
@@ -200,7 +200,7 @@ describe('TextComponent', () => {
 
       // SUT
       const cancelled = !env.component.editor?.container.dispatchEvent(dragEvent);
-      tick();
+      flush();
 
       expect(env.component.getSegmentText(targetSegmentRef)).toEqual(expectedFinalText);
       expect(env.component.editor?.getText()).toContain(expectedFinalText);
@@ -262,7 +262,7 @@ describe('TextComponent', () => {
 
       // SUT
       env.component.editor?.container.dispatchEvent(dragEvent);
-      tick();
+      flush();
 
       expect(env.component.getSegmentText(targetSegmentRef)).toEqual(expectedFinalText);
       expect(env.component.editor?.getText()).toContain(expectedFinalText);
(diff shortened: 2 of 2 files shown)


irahopkinson (changes requested):
Looking good. A few things to respond to.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-:-MpQ58eAABWO6kYkwhgV:bokpobq)* status: all files reviewed, 5 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 105 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-MpQ4koO1JAxKimtyZgO:-MpQ4koO1JAxKimtyZgP:b-7v7ksc) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba700a924c75f2f9ec3d2a21ac7ce1259eb92acd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L105)):*
> ```
> setTimeout(
> ```
Obviously not excited about indented `setTimeout`s but its only 3 deep so far and I don't see another way that would be simpler.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-MpQ1U7H2E7V5t45ziYu:-MpQ1U7H2E7V5t45ziYv:bcbsan6) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba700a924c75f2f9ec3d2a21ac7ce1259eb92acd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L13)):*
> ```
> import { Subscription } from 'rxjs';
> import { BehaviorSubject } from 'rxjs';
> ```
Combine these. In VS Code, if you use _ctrl + ._ on the item that hasn't yet been imported, it will do the right thing for you.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 204 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-MpQ1z7vAXhERMW0s8-1:-MpQ1z7vAXhERMW0s8-2:b-nrey5p) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba700a924c75f2f9ec3d2a21ac7ce1259eb92acd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L204)):*
> ```
> flush
> ```
Can you explain the change from `tick` to `flush` here and below? Is it because of the `setTimeout`s?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 788 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-MpQ2RFA4TwOD1dssSgw:-MpQ2RFA4TwOD1dssSgx:bhixyvs) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba700a924c75f2f9ec3d2a21ac7ce1259eb92acd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L788)):*
> ```
> , and updated and quill changes are interleaved
> ```
I'm having trouble understanding the meaning of this addition?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 864 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-MpQ2y0e0SSbVV7pHUVE:-MpQ2y0e0SSbVV7pHUVF:b-6psd7g) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba700a924c75f2f9ec3d2a21ac7ce1259eb92acd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L864)):*
> ```
> Array<{ ... }>
> ```
Use the shorthand array notation instead, i.e. `{ ... }[]`.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 895 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-MpQ4-1O8c9v5SVuoZna:-MpQ4-1O8c9v5SVuoZnb:b-lgwvww) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba700a924c75f2f9ec3d2a21ac7ce1259eb92acd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L895)):*
> ```
> toContain
> ```
Your comment above says 'then ...', but `toContain` won't check the order. It seems like we should care about order. Is it inherent in some way I can't see? If so perhaps a comment to explain why.

irahopkinson (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1196)* status: all files reviewed, 6 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 67 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-MpQ9Ve3CjOZbpM_7Wpl:-MpQ9Ve3CjOZbpM_7Wpm:bw090dl) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba700a924c75f2f9ec3d2a21ac7ce1259eb92acd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L67)):*
> ```
> ) ?? -1;
> ```
I checked the code in quill for `setSelection` and if the index is less than 0 it sets the index to 0 and still sets the selection. Is that what you expected?

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1196)* status: all files reviewed, 6 unresolved discussions (waiting on @irahopkinson)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 67 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-MpQ9Ve3CjOZbpM_7Wpl:-MpsC7En6vOINMVdGkSh:b7j8q9p) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba700a924c75f2f9ec3d2a21ac7ce1259eb92acd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L67)):*

> ```
> ) ?? -1;
> ```
I checked the code in quill for `setSelection` and if the index is less than 0 it sets the index to 0 and still sets the selection. Is that what you expected?
</blockquote></details>

Ah. Well, I was using -1 as a specific problem value to help with types, and wasn't concerned about whether it was something allowable to pass to quill or not. But I re-wrote this section to be a little bit cleaner.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-MpQ1U7H2E7V5t45ziYu:-Mprw8qs7uuvzGBvuU8M:b1x1l2w) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba700a924c75f2f9ec3d2a21ac7ce1259eb92acd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L13)):*

> ```
> import { Subscription } from 'rxjs';
> import { BehaviorSubject } from 'rxjs';
> ```
Combine these. In VS Code, if you use _ctrl + ._ on the item that hasn't yet been imported, it will do the right thing for you.
</blockquote></details>

Oops. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 204 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-MpQ1z7vAXhERMW0s8-1:-Mps0bdQ6BCigvQ_j_5h:b-qlbzj2) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba700a924c75f2f9ec3d2a21ac7ce1259eb92acd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L204)):*

> ```
> flush
> ```
Can you explain the change from `tick` to `flush` here and below? Is it because of the `setTimeout`s?
</blockquote></details>

Yes. It appears that `flush()` is needed now because of the setTimeouts(). I see from the docs for tick and flush that tick drains the microtask queue while flush drains the macrotask queue. I'll put a note about flush in the commit message.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 788 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-MpQ2RFA4TwOD1dssSgw:-Mps2hNv7gE9ipA0zfIb:b-5nivlx) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba700a924c75f2f9ec3d2a21ac7ce1259eb92acd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L788)):*

> ```
> , and updated and quill changes are interleaved
> ```
I'm having trouble understanding the meaning of this addition?
</blockquote></details>

I am adding some additional explanation into the comment at the beginning of the text. I can add more if it's still not clear. Oh, I bet the abbreviated reference to "updated" was part of the confusion. By "updated" I mean the `TextComponent.updated` event. I'll adjust the test name.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 864 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-MpQ2y0e0SSbVV7pHUVE:-Mps3Q4O26iDztHRtLPx:b-dz0mn5) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba700a924c75f2f9ec3d2a21ac7ce1259eb92acd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L864)):*

> ```
> Array<{ ... }>
> ```
Use the shorthand array notation instead, i.e. `{ ... }[]`.
</blockquote></details>

Done. 
Why? `Array<something>` seems so nicely descriptive :)
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 895 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-MpQ4-1O8c9v5SVuoZna:-Mps844cAvBExdAN5ikL:bydvdhu) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba700a924c75f2f9ec3d2a21ac7ce1259eb92acd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L895)):*

> ```
> toContain
> ```
Your comment above says 'then ...', but `toContain` won't check the order. It seems like we should care about order. Is it inherent in some way I can't see? If so perhaps a comment to explain why.
</blockquote></details>

It's not so much checking the order, in the sense that: It's not really checking that this happened after the previous thing checked. 
But what it's checking is that at a specific processing of  the `TextComponent.updated` event, a specific number of setSelection, insertText, and deleteText calls had occurred prior to that point. 

So it's implied that there is a certain order. But what we're checking isn't the order but rather than  a certain number of edit or selection activities happened before a given `TextComponent.updated` event. 

I wrote out a verbose explanation to include with the test in case it's not clear to the next person.

Actually: Writing a unit test to assert the order of some event processing is not something I do frequently, and this is what I came up with. But if you have another suggestion, I would be happy to listen to it.

irahopkinson (approved):
:lgtm: Please rebase and merge (if I don't get to it first).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-:-MpywzAb8MypEWBVQWre:busu16b)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 864 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1196#-MpQ2y0e0SSbVV7pHUVE:-MpyvjIh6TLn_gEaX0Vw:b-df00k) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba700a924c75f2f9ec3d2a21ac7ce1259eb92acd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L864)):*

Done. 
Why? `Array<something>` seems so nicely descriptive :)

</blockquote></details>

Yes, we use the longer notation only when necessary, i.e. no other way to describe it. We use the shorthand notation where we can because it matches other usual notation, e.g. `string[]` or `Verse[]`.

PR #1204, opened 2021-12-08
SF-1424 Reattached notes in Paratext appear in the reattached location

RaymondLuong3:
When notes have been reattached in the chapter, notes in SF need to appear in the updated location. The backend had to be updated to add the Reattached property to Note.cs. Using the Reattached property, the backend can determine what is the new verse that the note is attached to, and update the Position property of the NoteThread.
On the frontend, the majority of the changes involve updating the note dialog to correctly display Notes in the NoteThread to reflect the reattachment of the note thread.

Note reattached

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1204)


marksvc:
Examining.

marksvc (changes requested):
> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1204)* status: all files reviewed, 20 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts, line 38 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqtcs0J-N_p1LMGK7ND:-Mqtcs0M3gNNCcNEXxit:b-k3bx0t) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts#L38)):*
> ```
>   currentVerseRef(noteThread: NoteThread):
> ```

Is there a special reason that this method takes a `NoteThread` rather than operate on `this.data`?
There are only a couple places this method is called, and in both we write something like `foo.currentVerseRef(foo.data)`, which seems odd. 
That's kind of more of a static usage. What do you think about re-writing the method signature and beginning as 
```ts
currentVerseRef(): VerseRef {
  if (this.data==null) {
    ?
  }
  noteThread: NoteThread == this.data
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts, line 41 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqpp_s-66l-TQSqani5:-Mqpp_s-66l-TQSqani6:b88duv4) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts#L41)):*
> ```
> d!)
> ```


That's too bad for the compiler not to understand that reattached would always be not-null here. But I don't like the introduction of `!` :). 

I played with this a bit and came up with the following. What do you think about doing something like this so we don't introduce `!` into the production code? 

It didn't look like `.map()` had a way to _not_ return/set a value. I didn't try throwing from inside the map lambda, which may have been similar to the following code. 

And I noticed that we don't need `map` anyway and we can just query the `.reattached` property on the filtered array (or on the single resulting hit in the code below).

Maybe I should always just think of a `foo!` related crash as a "Programming error" and so short-hand for what I did in the following code. But I think I'm inclined to just not put `!` into the production code.


```ts
  currentVerseRef(noteThread: NoteThread): VerseRef {
    const lastReattach: Note | undefined = this.notesInOrderClone(noteThread.notes)
      .reverse()
      .find((n: Note) => n.reattached != null);
    if (lastReattach == null) {
      return toVerseRef(noteThread.verseRef);
    }
    if (lastReattach.reattached == null) {
      throw new Error('Programming error');
    }
    const verseStr: string = lastReattach.reattached.split(REATTACH_SEPARATOR)[0];
    return VerseRef.parse(verseStr);
  }

  /** Returns notes sorted by creation date from earlier to later. */
  notesInOrderClone(notes: Note[]): Note[] {
    ...
  }
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts, line 51 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqtbYQk518hylGZFjCA:-MqtbYQo2IZY4gmbJ_v0:b3dhiyd) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts#L51)):*
> ```
> this.data == null
> ```

Oops, I think here we want `notes == null`.

Actually. This method is acting like a static method as well, and every call to it looks something like `foo.notesInOrderClone(foo.data.notes)`. What do you think about re-writing it as
```ts
notesInOrderClone(): Note[] {
  if (this.data == null || this.data.notes == null) {
    return []
  }
  notes: Notes[] = this.data.notes
  ...
}
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.spec.ts, line 199 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqtfrxfBXRjORl-HWEe:-Mqtfrxi5iTnX-2Nt-iu:b2s83zj) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.spec.ts#L199)):*
> ```
> currentVerseRef
> ```

It looks like we don't test the other path of this method: When there is no re-attached note. Let's have a test specify what the output is when there are no re-attached notes, i.e. the noteThread.verseRef.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1243 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqu7MLc0aAYTuM_Chg-:-Mqu7MLh4mVZthvp5By3:b-1kuyk9) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1243)):*
> ```XML
>       const note4Position: number = env.getNoteThreadEditorPosition('thread04');
>       const note4Anchor: TextAnchor = env.getNoteThreadDoc('project01', 'thread04')!.data!.position;
>       expect(note4Anchor).toEqual({ start: 19, length: 5 });
> ```

It's not clear from the test if 19,5 is new and different, or the same thing as before the note was re-attached. Maybe that could be made more clear somehow?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 2611 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqu1hG66u2gPeDi711F:-Mqu1hG66u2gPeDi711G:bbcnq2c) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L2611)):*
> ```
> 19, length: 5 };
> ```

Can we replace this with explanatory constants or derived values?
I don't suppose it's arbitrary and just needs to match the numbers in the unit test ?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts, line 140 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MquKYx41y5zRFXjZIGb:-MquKYx54l5TbhogaBso:b-ro8mb8) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L140)):*
> ```XML
>       return '';
>     }
>     if (note.reattached != null) {
> ```

Hmm. It seems that the fact that the note is re-attached would be less important than whether it's a "todo" or "done" or other meaningful icon. What's the thinking behind prioritizing the reattached icon? 
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts, line 155 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MquKw7K7QwCcHhKI0ZT:-MquKw7K7QwCcHhKI0ZU:bp8giun) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L155)):*
> ```XML
>   noteTitle(note: Note) {
>     if (note.reattached != null) {
>       return translate('note_dialog.note_reattached');
> ```

Hmm, same thing here as with the icon. Isn't it more important if the note has status ToDo or Resolved, than whether it was moved to a different verse?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts, line 172 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MquEo6i3SSuqHJYnouV:-MquEo6kC2Enlky8OFA-:b-nchpe2) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L172)):*
> ```
> reattachedParts[3]
> ```

It's hard to know if [3] here and [1] there etc makes any sense :)
Can we assign these to meaningfully named variables first, like
```ts
const contextBefore = reattachedParts[n]
const something = reattachedParts[m]
```
before using them to build the output string?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts, line 177 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MquJzqr6__vKjfimm9T:-MquJzqu-HEpovYRA6-R:b-qqds2t) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L177)):*
> ```XML
>     const reattached: string = translate('note_dialog.reattached');
> 
>     return `${verseRef} ${reattached}</br>${reattachedText}`;
> ```

Reattached notes can also contain `<Contents>` with non-empty strings. I don't think we are accounting for that here.
___
*[src/SIL.XForge.Scripture/Models/Note.cs, line 11 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpHvfe21_RVlTNLn6i:-MqpHvfe21_RVlTNLn6j:bmlr008) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/Models/Note.cs#L11)):*
> ```Smalltalk
>         public string Status { get; set; }
>         public string TagIcon { get; set; }
>         public string Reattached { get; set; }
> ```

It's too bad that this doesn't let us also specify whether it's a required property.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1528 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpL2so1w7C9XNGPboQ:-MqpL2spBUZMewX-D77y:bq4dktn) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1528)):*
> ```Smalltalk
>             // Check if fields have been updated in Paratext
>             bool statusChanged =
>                 !string.IsNullOrEmpty(comment.Status.InternalValue) && comment.Status.InternalValue != note.Status;
> ```

Might it be that the SF note.Status is present, and the PT comment.Status.InternalValue is (newly) null or empty, and so the status has changed, but we here set `statusChanged = false`? 
Status could be handled the same way content was in this change, if significant.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1547 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpXBqnAlo-nzhHvqGq:-MqpXBqnAlo-nzhHvqGr:b1qa7lt) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1547)):*
> ```Smalltalk
>             {
>                 comment.AddTextToContent("", false);
>                 comment.Contents.InnerXml = note.Content;
> ```

I was able to make a Comment with a Reattached element and no Contents element. 
Hmm but I haven't yet been able to take a Comment with a non-empty Contents element and make it lose its Contents element, to say that this line above may crash (if `comment.Contents` is null). 

However, once we let users edit notes in SF, it could be that there is a Paratext Comment with absent `<Contents>` element, which creates a SF Note, with equivalent .Content property. And then in SF the Note is edited by adding content (or not allowed possible maybe). Then in this method we do
```
if (! null or empty (note.Content)):  // true
  ..
  comment.Contents.InnerXml = note.Content // Crash
```
where we crash because `comment.Contents` is null. I looked at Paratext's CommentList.cs to see if Comment.Contents is always non-null but it looks like it can be null. 

What do you think about modifying this so that ahead of time we are prepared to write content to comment.Contents.InnerXml in situations where comment.Contents starts off null?
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1596 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpMM8OBf-OKpMNYatj:-MqpMM8OBf-OKpMNYatk:b2vtcau) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1596)):*
> ```
>  GetCommentTextAnchor
> ```

Would it be a good idea to rename this now, to something like GetThreadTextAnchor ? 
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1597 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpTOfi3PS2XoVh44ti:-MqpTOfkEhvOurqs8e-T:bfatomx) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1597)):*
> ```Smalltalk
> 
>         private TextAnchor GetCommentTextAnchor(CommentThread thread, Dictionary<int, ChapterDelta> chapterDeltas)
>         {
> ```

FWIW

So I was playing around in PT9.0 and found that I could create a note, click re-attach and re-attach it. Then enter some comment text and click OK. The result is that in the note XML file I have a thread with a single comment, that has both a `<Reattached>` element and a non-empty `<Contents>` element. 
I was able to add a second <Comment> to that thread with another Reattached and no Contents element.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1619 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpW_Wi2zJ9LeXfL6hb:-MqpW_Wk-MRBKaeZ8Y7z:b6mdr34) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1619)):*
> ```
>  ref startPos,
> ```

It looks good to pass comment.StartPosition here, but I note with interest that before we were just passing 0. 
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 847 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpdCkK18WaV031JAz_:-MqpdCkK18WaV031JAza:bnvswqn) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L847)):*
> ```Smalltalk
>                 new ThreadComponents { threadNum = 3, noteCount = 1, reattachedVerseStr = verseStr }
>             });
>             env.AddParatextComments(new[]
> ```

How about a thread4 where SF knows about 1 comment, and PT is introducing a new comment where the new comment also has a Reattached element? (The test at hand appears to add Reattached to an existing comment).
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 848 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpeAXV0sYz7j2UQfmt:-MqpeAXV0sYz7j2UQfmu:bg3cufy) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L848)):*
> ```Smalltalk
>             });
>             env.AddParatextComments(new[]
>             {
> ```

I am able to make a thread with a comment, then add a reattach with comment, then remove the reattach+comment comment. And so then the thread goes  back to its original position. Can we add to our tests this situation?
- SF knows about a thread that has a reference, but is re-attached to a different reference. 
- SF receives a change from PT where the Comment with Reattached is no longer present.
- SF now places the thread at its original reference.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 2041 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqphp6o4PKOUQmVdlNR:-Mqphp6o4PKOUQmVdlNS:b3gmibq) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2041)):*
> ```Smalltalk
>                             Thread = threadId,
>                             VerseRefStr = verseStr,
>                             SelectedText = selectedText,
> ```

And shouldn't this instead be rnt.selectedText ? I guess the test should be checking that the reattached comment item has rnt.SelectedText, rnt.before, and rnt.after.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 2044 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpgxWX6NYjR0cZS8jd:-MqpgxWX6NYjR0cZS8je:b-k4bz8e) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2044)):*
> ```
>                             ContextBefore = comp.appliesToVerse ? "" : before,
>                             ContextAfter = comp.appliesToVerse ? "" : after,
>                             StartPosition = comp.appliesToVerse ? 0 : before.Length,
> ```

Shouldn't this instead be using rnt.before, rnt.after, rnt.before.Length ?
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 2104 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqpi-B36xItv-wJ_W6x:-Mqpi-B36xItv-wJ_W6y:b340p6p) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2104)):*
> ```
>                 string before = "Context before ";
>                 string after = " context after.";
> ```

How about changing these so they are different for the re-attached verse than for the original verse, so we can check that we used the new values? eg 
```
string before = "Reattached context before "
string after = " reattached context after."
```
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 2131 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpiwEE9uAD9t_JV5nj:-MqpiwEE9uAD9t_JV5nk:bvif82b) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2131)):*
> ```
>                     if (i == 7)
>                         noteSelectedText = "reattached text";
> ```

Hmm, okay. What do you think about adding a comment before this like
// Make one of the verses have different text to optionally use to re-attach to.

RaymondLuong3 (commented):
I've made some changes, but there are a few more I need to make before this is ready for re-review

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-:-MqztE9C208xBoWZdGr_:brvg850)* status: all files reviewed, 20 unresolved discussions (waiting on @marksvc and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts, line 38 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqtcs0J-N_p1LMGK7ND:-MqzDSTK9XKDHDo2NJLf:bl5ccbp) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts#L38)):*

> ```
>   currentVerseRef(noteThread: NoteThread):
> ```

Is there a special reason that this method takes a `NoteThread` rather than operate on `this.data`?
There are only a couple places this method is called, and in both we write something like `foo.currentVerseRef(foo.data)`, which seems odd. 
That's kind of more of a static usage. What do you think about re-writing the method signature and beginning as 
```ts
currentVerseRef(): VerseRef {
  if (this.data==null) {
    ?
  }
  noteThread: NoteThread == this.data
```
</blockquote></details>

Ok, I've changed a few things around to make this work. currentVerseRef now returns VerseRef | undefined.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts, line 41 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqpp_s-66l-TQSqani5:-MqzFS2q10Kn3zzSl5Eo:bi4lwu8) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts#L41)):*

> ```
> d!)
> ```


That's too bad for the compiler not to understand that reattached would always be not-null here. But I don't like the introduction of `!` :). 

I played with this a bit and came up with the following. What do you think about doing something like this so we don't introduce `!` into the production code? 

It didn't look like `.map()` had a way to _not_ return/set a value. I didn't try throwing from inside the map lambda, which may have been similar to the following code. 

And I noticed that we don't need `map` anyway and we can just query the `.reattached` property on the filtered array (or on the single resulting hit in the code below).

Maybe I should always just think of a `foo!` related crash as a "Programming error" and so short-hand for what I did in the following code. But I think I'm inclined to just not put `!` into the production code.


```ts
  currentVerseRef(noteThread: NoteThread): VerseRef {
    const lastReattach: Note | undefined = this.notesInOrderClone(noteThread.notes)
      .reverse()
      .find((n: Note) => n.reattached != null);
    if (lastReattach == null) {
      return toVerseRef(noteThread.verseRef);
    }
    if (lastReattach.reattached == null) {
      throw new Error('Programming error');
    }
    const verseStr: string = lastReattach.reattached.split(REATTACH_SEPARATOR)[0];
    return VerseRef.parse(verseStr);
  }

  /** Returns notes sorted by creation date from earlier to later. */
  notesInOrderClone(notes: Note[]): Note[] {
    ...
  }
```
</blockquote></details>

I've updated the code to follow your suggestion. It is certainly nice when we can avoid `!` assertions in our production code. I did not include the programming error line, I think we can keep going if the lastReattach.reattached is null.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts, line 51 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqtbYQk518hylGZFjCA:-MqzGI_y2BRvJ42J649j:bu02aup) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts#L51)):*

> ```
> this.data == null
> ```

Oops, I think here we want `notes == null`.

Actually. This method is acting like a static method as well, and every call to it looks something like `foo.notesInOrderClone(foo.data.notes)`. What do you think about re-writing it as
```ts
notesInOrderClone(): Note[] {
  if (this.data == null || this.data.notes == null) {
    return []
  }
  notes: Notes[] = this.data.notes
  ...
}
```

</blockquote></details>

Turns out we do not need the null check at all.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.spec.ts, line 199 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqtfrxfBXRjORl-HWEe:-MqzQe0A90RdySv6Olin:brrvei6) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.spec.ts#L199)):*

> ```
> currentVerseRef
> ```

It looks like we don't test the other path of this method: When there is no re-attached note. Let's have a test specify what the output is when there are no re-attached notes, i.e. the noteThread.verseRef.
</blockquote></details>

Good thinking. I added two checks on previous tests. One for a thread with existing notes, one with no existing notes.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1243 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqu7MLc0aAYTuM_Chg-:-MqzTCKDALsKhNbRuII5:b-ddr6nz) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1243)):*

It's not clear from the test if 19,5 is new and different, or the same thing as before the note was re-attached. Maybe that could be made more clear somehow?
</blockquote></details>

The position of a note thread will always be the up-to-date value. I've adjust the test a little, so we use the expected position value to detect if the position in-editor is correct.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 2611 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqu1hG66u2gPeDi711F:-MqzUYPj046qUATx18H-:b-xms6cj) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L2611)):*

> ```
> 19, length: 5 };
> ```

Can we replace this with explanatory constants or derived values?
I don't suppose it's arbitrary and just needs to match the numbers in the unit test ?
</blockquote></details>

Good thinking. The reattachNote method now accepts the position as a param.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts, line 140 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MquKYx41y5zRFXjZIGb:-MqzWXif29wl8Trgr162:b-v2hjnf) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L140)):*

Hmm. It seems that the fact that the note is re-attached would be less important than whether it's a "todo" or "done" or other meaningful icon. What's the thinking behind prioritizing the reattached icon? 
</blockquote></details>

It seems kind of overloaded for a comment to both have content, and represent a reattachment. But this seems to be the case. So I have had to rethink how we handle reattached notes. In this case, we will only show the reattach icon if the note has no status.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts, line 155 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MquKw7K7QwCcHhKI0ZT:-MqzkBcG7FDrKdqe3ked:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L155)):*

Hmm, same thing here as with the icon. Isn't it more important if the note has status ToDo or Resolved, than whether it was moved to a different verse?
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts, line 172 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MquEo6i3SSuqHJYnouV:-MqzkiO77zWIxp6cww_8:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L172)):*

> ```
> reattachedParts[3]
> ```

It's hard to know if [3] here and [1] there etc makes any sense :)
Can we assign these to meaningfully named variables first, like
```ts
const contextBefore = reattachedParts[n]
const something = reattachedParts[m]
```
before using them to build the output string?
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts, line 177 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MquJzqr6__vKjfimm9T:-Mqzpy8t9cz6hGDY9k92:bg8c5yv) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L177)):*

Reattached notes can also contain `<Contents>` with non-empty strings. I don't think we are accounting for that here.
</blockquote></details>

Yes, I see that now. I've updated the note so that both the reattached context and the content appear in the dialog.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1528 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpL2so1w7C9XNGPboQ:-MqzvOt30p_DjlIX9rHS:b1narn4) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1528)):*

Might it be that the SF note.Status is present, and the PT comment.Status.InternalValue is (newly) null or empty, and so the status has changed, but we here set `statusChanged = false`? 
Status could be handled the same way content was in this change, if significant.


</blockquote></details>

Good thinking. I don't know enough about comment status, but to be safe, I've updated this as per your suggestion.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1547 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpXBqnAlo-nzhHvqGq:-MqzwnlmAZ2bGU3UrSTe:b-shxcdh) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1547)):*

I was able to make a Comment with a Reattached element and no Contents element. 
Hmm but I haven't yet been able to take a Comment with a non-empty Contents element and make it lose its Contents element, to say that this line above may crash (if `comment.Contents` is null). 

However, once we let users edit notes in SF, it could be that there is a Paratext Comment with absent `<Contents>` element, which creates a SF Note, with equivalent .Content property. And then in SF the Note is edited by adding content (or not allowed possible maybe). Then in this method we do
```
if (! null or empty (note.Content)):  // true
  ..
  comment.Contents.InnerXml = note.Content // Crash
```
where we crash because `comment.Contents` is null. I looked at Paratext's CommentList.cs to see if Comment.Contents is always non-null but it looks like it can be null. 

What do you think about modifying this so that ahead of time we are prepared to write content to comment.Contents.InnerXml in situations where comment.Contents starts off null?
</blockquote></details>

Initially I used comment.AddTextToContent() to guarantee that comment.Contents is not null because AddTextToContent creates a Contents node. I updated this method to use the comment.GetOrCreateCommentNode() directly to guarantee that the content node exists.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1596 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpMM8OBf-OKpMNYatj:-MqzyH2kE58-8ulwuao8:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1596)):*

> ```
>  GetCommentTextAnchor
> ```

Would it be a good idea to rename this now, to something like GetThreadTextAnchor ? 
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1597 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpTOfi3PS2XoVh44ti:-MqzyKYUBFQsNlHj6UH9:bpvllg3) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1597)):*

FWIW

So I was playing around in PT9.0 and found that I could create a note, click re-attach and re-attach it. Then enter some comment text and click OK. The result is that in the note XML file I have a thread with a single comment, that has both a `<Reattached>` element and a non-empty `<Contents>` element. 
I was able to add a second <Comment> to that thread with another Reattached and no Contents element.
</blockquote></details>

Yes, I see that we need to assume that the Contents and the Reattached properties can coexist.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1619 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpW_Wi2zJ9LeXfL6hb:-MqzyhXi5CdY5gwnWLfv:bqzyqyq) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1619)):*

> ```
>  ref startPos,
> ```

It looks good to pass comment.StartPosition here, but I note with interest that before we were just passing 0. 
</blockquote></details>

This one is interesting. We are using the StringUtils.MatchContexts() in a nuanced way. We use the chapter deltas to give us the verse text, but the deltas do no include verse markers. The startPos that Paratext records includes verse markers. So the startPos we pass here is meaningless. So either 0, or the comment.StartPosition would be fine to use.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 847 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpdCkK18WaV031JAz_:-Mr-0zDl4jQX2yn-jane:bwe4hjl) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L847)):*

How about a thread4 where SF knows about 1 comment, and PT is introducing a new comment where the new comment also has a Reattached element? (The test at hand appears to add Reattached to an existing comment).
</blockquote></details>

I can add that to this test.

RaymondLuong3 (commented):
Ok, the comments have been addressed and is now ready for re-review.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-:-Mr3W1Bl5rifG1jSAuXS:b-byshey)* status: all files reviewed, 20 unresolved discussions (waiting on @marksvc)

___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 848 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpeAXV0sYz7j2UQfmt:-Mr3Ggg4DSDW5-qpafW3:b-8f42pb) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L848)):*

I am able to make a thread with a comment, then add a reattach with comment, then remove the reattach+comment comment. And so then the thread goes  back to its original position. Can we add to our tests this situation?
- SF knows about a thread that has a reference, but is re-attached to a different reference. 
- SF receives a change from PT where the Comment with Reattached is no longer present.
- SF now places the thread at its original reference.


</blockquote></details>

How does this sound. The SF NoteThreadDocs start with a thread5 with the second note being a reattached note. The Paratext commentThread for thread5 contains just one comment (the second comment is assumed to be deleted). Then after the operation, test that thread5's second second noteThread in SF is marked for removal.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 2041 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqphp6o4PKOUQmVdlNR:-Mr3HE0FDYF8VygwFEEt:bjho6q) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2041)):*

And shouldn't this instead be rnt.selectedText ? I guess the test should be checking that the reattached comment item has rnt.SelectedText, rnt.before, and rnt.after.
</blockquote></details>

No, these properties never change on the Comment thread. The only thing that changes is the content and the reattached string.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 2044 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpgxWX6NYjR0cZS8jd:-Mr3HT8J9ac21XqhI7Nk:b2nreh6) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2044)):*

> ```
>                             ContextBefore = comp.appliesToVerse ? "" : before,
>                             ContextAfter = comp.appliesToVerse ? "" : after,
>                             StartPosition = comp.appliesToVerse ? 0 : before.Length,
> ```

Shouldn't this instead be using rnt.before, rnt.after, rnt.before.Length ?
</blockquote></details>

Same as above.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 2104 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqpi-B36xItv-wJ_W6x:-Mr3UHSFD7-tGpc1DGxw:b-nh1smh) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2104)):*

> ```
>                 string before = "Context before ";
>                 string after = " context after.";
> ```

How about changing these so they are different for the re-attached verse than for the original verse, so we can check that we used the new values? eg 
```
string before = "Reattached context before "
string after = " reattached context after."
```
</blockquote></details>

I've created three new constants for reattached selected text, an alternate context before and after. This allows us to confirm that the changed position matches what we expect.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 2131 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpiwEE9uAD9t_JV5nj:-Mr3VzS8BnBKDFL1EDZg:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2131)):*

> ```
>                     if (i == 7)
>                         noteSelectedText = "reattached text";
> ```

Hmm, okay. What do you think about adding a comment before this like
// Make one of the verses have different text to optionally use to re-attach to. 
</blockquote></details>

Done.

marksvc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1204)* status: all files reviewed, 4 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1243 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqu7MLc0aAYTuM_Chg-:-Mr4HL0u7vt-mpWgNCBG:b-qggjd4) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1243)):*

The position of a note thread will always be the up-to-date value. I've adjust the test a little, so we use the expected position value to detect if the position in-editor is correct.
</blockquote></details>

Hmm, okay. I have been puzzled by this test today and think it could use some clarity. Could we assert earlier in the test that the note text position is not equal to {start:19,length:5}, and that thread4 is in verse 3? For example:

```
expect(initial thread 4 text anchoring != `position`) (and I guess quite significantly that .start is different between them)
expect(thread 4 editor position === (getSegmentRange('verse_1_3')).index + (initial thread 4 text anchoring).start)
reattachNote(thread04 to 'MAT 1:4', newTextPosition)
expect(new thread 4 text anchoring === `position`)
expect(thread 4 editor position === (getSegmentRange('verse_1_4')).index + (new thread 4 text anchoring).start)
```

As the test is written now, the significance of {start:19,length:5} is not obvious. Was it {start:19,length:5} before and it is still {start:19,length:5} now? Was it some other value and we changed it to {start:19,length:5}? Is it significant that it be 19,5 or if we wrote 100,200, would that still make the test pass? Presumably 19,5 relates to text of MAT 1:4, but does it relate the same way to the text of MAT 1:3?

I can see that we assert that thread04 is in verse_1_4, but the test doesn't show that it wasn't there before. 

I'm also a bit confused about where the action happens. Is there a line that would be a good place to label as SUT?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts, line 172 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MquEo6i3SSuqHJYnouV:-Mr4NpKl1PAonrDr3RuR:b5ugbmh) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L172)):*

Done.
</blockquote></details>

Much easier to read!
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts, line 175 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mr4MiI-AioOR1-xywZQ:-Mr4MiI1-2N20zK_l_So:b-lz7mfg) ([raw file](https://github.com/sillsdev/web-xforge/blob/aa831ebb90f6dae82e539a9867da8141fdec3d11/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L175)):*
> ```XML
>     const reattached: string = translate('note_dialog.reattached');
>     let noteContent: string = this.parseNote(note.content);
>     noteContent = noteContent === '' ? '' : `<br>${noteContent}`;
> ```

I find this kind of awkward. What do you think about something like this instead? 
```ts
if (noteContent !== '') {
  noteContent = `<br>${noteContent}`
}
```

Also, I'm guessing it might benefit from `<p>` rather than `<br>`, but haven't seen at how it looks yet.



_Code quote:_
```XML
  noteContent = noteContent === '' ? '' : `<br>${noteContent}`;
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts, line 177 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mr4OD-0-mLIduOFAbMK:-Mr4OD-0-mLIduOFAbML:b-tqer1e) ([raw file](https://github.com/sillsdev/web-xforge/blob/aa831ebb90f6dae82e539a9867da8141fdec3d11/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L177)):*
> ```XML
>     noteContent = noteContent === '' ? '' : `<br>${noteContent}`;
> 
>     return `${verseRef} ${reattached}<br>${reattachedText}${noteContent}`;
> ```

```diff
-</br>
+<br>
```

Oops, good catch.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1528 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpL2so1w7C9XNGPboQ:-Mr4S0vI-a0EDbY995SV:bo9cv1s) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1528)):*

Good thinking. I don't know enough about comment status, but to be safe, I've updated this as per your suggestion.
</blockquote></details>

Ha, ya know, as I look at this more I am wondering if we can simplify it a lot. What do you think about changing it to this? (I'm not as familiar with these values, so it may not make sense to do it this way tho.)

```diff
             // Check if fields have been updated in Paratext
-            bool statusExists = !string.IsNullOrEmpty(comment.Status.InternalValue) ||
-                !string.IsNullOrEmpty(note.Status);
-            bool statusChanged = statusExists && comment.Status.InternalValue != note.Status;
-            bool contentExists = !string.IsNullOrEmpty(note.Content) ||
-                !string.IsNullOrEmpty(comment.Contents?.InnerXml);
-            bool contentChanged = contentExists && comment.Contents.InnerXml != note.Content;
+            bool statusChanged = comment.Status.InternalValue != note.Status;
+            bool contentChanged = comment.Contents?.InnerXml != note.Content;
             if (contentChanged || statusChanged)
                 return ChangeType.Updated;
             return ChangeType.None;
```

When I did this, I noticed that the test setup code wasn't setting Status for a Note or a Comment when Reattached is set. But it didn't look like that would be so in production. So I modified the ParatextServiceTests.cs with

```diff
+++ b/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs
@@ -1929,6 +1929,7 @@ namespace SIL.XForge.Scripture.Services
                             ExtUserId = "user02",
                             SyncUserRef = "syncuser01",
                             DateCreated = new DateTime(2019, 1, 20, 8, 0, 0, DateTimeKind.Utc),
+                            Status = NoteStatus.Unspecified.InternalValue,
                             Reattached = ReattachedThreadInfoStr(rti)
                         });
                         noteThread.Position = new TextAnchor
@@ -2069,6 +2070,7 @@ namespace SIL.XForge.Scripture.Services
                             ContextAfter = comp.appliesToVerse ? "" : after,
                             StartPosition = comp.appliesToVerse ? 0 : before.Length,
                             Date = "2019-01-20T08:00:00.0000000+00:00",
+                            Status = NoteStatus.Unspecified,
                             Reattached = ReattachedThreadInfoStr(rti)
                         });
                     }
```

and the tests are passing.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1547 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpXBqnAlo-nzhHvqGq:-Mr8M6Zi3JdQkEMbyG0M:b-k77929) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1547)):*

Initially I used comment.AddTextToContent() to guarantee that comment.Contents is not null because AddTextToContent creates a Contents node. I updated this method to use the comment.GetOrCreateCommentNode() directly to guarantee that the content node exists.
</blockquote></details>

Ahh, okay. That makes sense then. 
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1619 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpW_Wi2zJ9LeXfL6hb:-Mr8MWyJDbvMBjFSgOKf:b-ke6dnq) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1619)):*

This one is interesting. We are using the StringUtils.MatchContexts() in a nuanced way. We use the chapter deltas to give us the verse text, but the deltas do no include verse markers. The startPos that Paratext records includes verse markers. So the startPos we pass here is meaningless. So either 0, or the comment.StartPosition would be fine to use.
</blockquote></details>

Interesting. Thanks for explaining.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 2041 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqphp6o4PKOUQmVdlNR:-Mr8Pnpx1LHyG0GY5B-C:bdlsc70) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2041)):*

No, these properties never change on the Comment thread. The only thing that changes is the content and the reattached string.
</blockquote></details>

Thank you. I see I was mis-reading my Notes .xml file.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 845 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mr8SODN9HIMIKyx9vak:-Mr8SODO7vPxVHeBQAWY:b-2ajdbp) ([raw file](https://github.com/sillsdev/web-xforge/blob/aa831ebb90f6dae82e539a9867da8141fdec3d11/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L845)):*
> ```Smalltalk
>             {
>                 new ThreadComponents { threadNum = 1, noteCount = 1 },
>                 new ThreadComponents { threadNum = 3, noteCount = 1, reattachedVerseStr = verseStr },
> ```

I keep forgetting if noteCount includes the reattached note. It's confusing that it doesn't. But at least AddNoteThreadData and AddParatextComments are both the same in the regard :)
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 901 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mr8VxMx63knFNE7d6aT:-Mr8VxN06KD_m3lRq7Sc:bj2covp) ([raw file](https://github.com/sillsdev/web-xforge/blob/aa831ebb90f6dae82e539a9867da8141fdec3d11/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L901)):*
> ```Smalltalk
>                 Assert.That(change5.NoteIdsRemoved.Count, Is.EqualTo(1));
>                 Assert.That(change5.NoteIdsRemoved[0], Is.EqualTo("reattachedthread5"));
>                 TextAnchor originalAnchor = new TextAnchor
> ```

Right. That's good. I think it would be good to tack in the following change as well. The usefulness of the `Assert.That(change5.Position, Is.EqualTo(originalAnchor));` partly relies on our test setup having helpful values for some of these items (like you have done by introducing TestEnvironment.AlternateBefore/After). 


```diff
+++ b/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs
@@ -898,6 +898,9 @@ namespace SIL.XForge.Scripture.Services
                 NoteThreadChange change5 = changes.Single(c => c.ThreadId == "thread5");
                 Assert.That(change5.NoteIdsRemoved.Count, Is.EqualTo(1));
                 Assert.That(change5.NoteIdsRemoved[0], Is.EqualTo("reattachedthread5"));
+                // The thread is un-reattached back to its original position.
+                Assert.That(change5.ContextBefore, Is.Not.EqualTo(rnt.contextBefore), "test framework");
+                Assert.That(change5.SelectedText, Is.Not.EqualTo(rnt.selectedText), "test framework");
                 TextAnchor originalAnchor = new TextAnchor
                 {
                     Start = change5.ContextBefore.Length,
```

I at first described the new lines as `"setup"`. It's something that will fail if our test system doesn't remain constructed in a way that helps us test this and so I ended up with "test framework". But you might have a better name :).  Or maybe "setup" is better.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 1332 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mr8RqeiCCF6ixYi0aMw:-Mr8RqeiCCF6ixYi0aMx:bypj500) ([raw file](https://github.com/sillsdev/web-xforge/blob/aa831ebb90f6dae82e539a9867da8141fdec3d11/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L1332)):*
> ```Smalltalk
>         }
> 
>         struct ReattachedThreadInfo
> ```

Yes I like this better.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1204)* status: 6 of 14 files reviewed, 4 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1243 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqu7MLc0aAYTuM_Chg-:-Mr9FFkb3OILv0qwI2Ju:bltt5v5) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1243)):*

Hmm, okay. I have been puzzled by this test today and think it could use some clarity. Could we assert earlier in the test that the note text position is not equal to {start:19,length:5}, and that thread4 is in verse 3? For example:

```
expect(initial thread 4 text anchoring != `position`) (and I guess quite significantly that .start is different between them)
expect(thread 4 editor position === (getSegmentRange('verse_1_3')).index + (initial thread 4 text anchoring).start)
reattachNote(thread04 to 'MAT 1:4', newTextPosition)
expect(new thread 4 text anchoring === `position`)
expect(thread 4 editor position === (getSegmentRange('verse_1_4')).index + (new thread 4 text anchoring).start)
```

As the test is written now, the significance of {start:19,length:5} is not obvious. Was it {start:19,length:5} before and it is still {start:19,length:5} now? Was it some other value and we changed it to {start:19,length:5}? Is it significant that it be 19,5 or if we wrote 100,200, would that still make the test pass? Presumably 19,5 relates to text of MAT 1:4, but does it relate the same way to the text of MAT 1:3?

I can see that we assert that thread04 is in verse_1_4, but the test doesn't show that it wasn't there before. 

I'm also a bit confused about where the action happens. Is there a line that would be a good place to label as SUT?


</blockquote></details>

I think the biggest part of the confusion is that reattachNote sounds like it is in realtime. And it is, but this test is not concerned about the realtime behaviour of notes (since editing notes is note yet a SF feature).
So read this as note loading behaviour. The editor is loaded after the wait() is called, and everything before that is setup.
If you read it that way, they you will see that we are testing that a note is loaded at the correct position, and `position` is the up-to-date position in the verse to load the note. So there is no "before" and "after" sense to this test. Only the "on load" sense.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts, line 175 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mr4MiI-AioOR1-xywZQ:-Mr9f4yP8SrY8GS0rjxq:b-uff1ei) ([raw file](https://github.com/sillsdev/web-xforge/blob/aa831ebb90f6dae82e539a9867da8141fdec3d11/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L175)):*

I find this kind of awkward. What do you think about something like this instead? 
```ts
if (noteContent !== '') {
  noteContent = `<br>${noteContent}`
}
```

Also, I'm guessing it might benefit from `<p>` rather than `<br>`, but haven't seen at how it looks yet.


</blockquote></details>

Even better. I've updated the note dialog so that the content will show up like such.

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1528 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MqpL2so1w7C9XNGPboQ:-Mr9LYT_EK1WzqXr8gCc:b-wanonu) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1528)):*

Ha, ya know, as I look at this more I am wondering if we can simplify it a lot. What do you think about changing it to this? (I'm not as familiar with these values, so it may not make sense to do it this way tho.)

```diff
             // Check if fields have been updated in Paratext
-            bool statusExists = !string.IsNullOrEmpty(comment.Status.InternalValue) ||
-                !string.IsNullOrEmpty(note.Status);
-            bool statusChanged = statusExists && comment.Status.InternalValue != note.Status;
-            bool contentExists = !string.IsNullOrEmpty(note.Content) ||
-                !string.IsNullOrEmpty(comment.Contents?.InnerXml);
-            bool contentChanged = contentExists && comment.Contents.InnerXml != note.Content;
+            bool statusChanged = comment.Status.InternalValue != note.Status;
+            bool contentChanged = comment.Contents?.InnerXml != note.Content;
             if (contentChanged || statusChanged)
                 return ChangeType.Updated;
             return ChangeType.None;
```

When I did this, I noticed that the test setup code wasn't setting Status for a Note or a Comment when Reattached is set. But it didn't look like that would be so in production. So I modified the ParatextServiceTests.cs with

```diff
+++ b/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs
@@ -1929,6 +1929,7 @@ namespace SIL.XForge.Scripture.Services
                             ExtUserId = "user02",
                             SyncUserRef = "syncuser01",
                             DateCreated = new DateTime(2019, 1, 20, 8, 0, 0, DateTimeKind.Utc),
+                            Status = NoteStatus.Unspecified.InternalValue,
                             Reattached = ReattachedThreadInfoStr(rti)
                         });
                         noteThread.Position = new TextAnchor
@@ -2069,6 +2070,7 @@ namespace SIL.XForge.Scripture.Services
                             ContextAfter = comp.appliesToVerse ? "" : after,
                             StartPosition = comp.appliesToVerse ? 0 : before.Length,
                             Date = "2019-01-20T08:00:00.0000000+00:00",
+                            Status = NoteStatus.Unspecified,
                             Reattached = ReattachedThreadInfoStr(rti)
                         });
                     }
```

and the tests are passing.
</blockquote></details>

Yes, that looks like an improvement. Note that it is no longer expected that comments with null content will map to the empty string in SF data.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 901 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mr8VxMx63knFNE7d6aT:-Mr9NmmK8BQXsvSfKWus:b2jof0n) ([raw file](https://github.com/sillsdev/web-xforge/blob/aa831ebb90f6dae82e539a9867da8141fdec3d11/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L901)):*

Right. That's good. I think it would be good to tack in the following change as well. The usefulness of the `Assert.That(change5.Position, Is.EqualTo(originalAnchor));` partly relies on our test setup having helpful values for some of these items (like you have done by introducing TestEnvironment.AlternateBefore/After). 


```diff
+++ b/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs
@@ -898,6 +898,9 @@ namespace SIL.XForge.Scripture.Services
                 NoteThreadChange change5 = changes.Single(c => c.ThreadId == "thread5");
                 Assert.That(change5.NoteIdsRemoved.Count, Is.EqualTo(1));
                 Assert.That(change5.NoteIdsRemoved[0], Is.EqualTo("reattachedthread5"));
+                // The thread is un-reattached back to its original position.
+                Assert.That(change5.ContextBefore, Is.Not.EqualTo(rnt.contextBefore), "test framework");
+                Assert.That(change5.SelectedText, Is.Not.EqualTo(rnt.selectedText), "test framework");
                 TextAnchor originalAnchor = new TextAnchor
                 {
                     Start = change5.ContextBefore.Length,
```

I at first described the new lines as `"setup"`. It's something that will fail if our test system doesn't remain constructed in a way that helps us test this and so I ended up with "test framework". But you might have a better name :).  Or maybe "setup" is better.

</blockquote></details>

Actually, the NoteThreadChange.ContextBefore will never be different regardless of whether the note is attached or not. Those properties are needed to create a note thread, but they are meaningless when the note thread doc already exists. Same with ContextBefore.

marksvc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1204)* status: all files reviewed, 3 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1243 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqu7MLc0aAYTuM_Chg-:-MrNzTKE1anPADeHRTVj:b-b1c5c1) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1243)):*

I think the biggest part of the confusion is that reattachNote sounds like it is in realtime. And it is, but this test is not concerned about the realtime behaviour of notes (since editing notes is note yet a SF feature).
So read this as note loading behaviour. The editor is loaded after the wait() is called, and everything before that is setup.
If you read it that way, they you will see that we are testing that a note is loaded at the correct position, and `position` is the up-to-date position in the verse to load the note. So there is no "before" and "after" sense to this test. Only the "on load" sense.
</blockquote></details>

That's helpful, thank you.
What is the meaning of {start: 19, length:5}? Is it the anchoring that was determined when the thread was first created, which would be used to anchor the thread to a particular part of the verse 3 text? Are they values that somehow relate to verse 4? 
Would it be accurate to give `position` a comment above it saying something like: // Original anchoring of thread 4 in verse 3
or something like // Anchoring of tread 4 in verse 4
?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts, line 175 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mr4MiI-AioOR1-xywZQ:-MrNlmem-4kdw1uPsmzE:bb6wk0y) ([raw file](https://github.com/sillsdev/web-xforge/blob/aa831ebb90f6dae82e539a9867da8141fdec3d11/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L175)):*

Even better. I've updated the note dialog so that the content will show up like such.


</blockquote></details>

Good
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts, line 122 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MrNmaBE9T-cQAbR9SCx:-MrNmaBE9T-cQAbR9SCy:bpeijz9) ([raw file](https://github.com/sillsdev/web-xforge/blob/9814097969bd73d8d7dee7c411ad3dfa1cc87525/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts#L122)):*
> ```XML
>   }));
> 
>   it('should show notes for reattachment', fakeAsync(() => {
> ```

How about also here (or in the below test) asserting that the re-attached location verse reference appears where it should?
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 901 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mr8VxMx63knFNE7d6aT:-MrNnjvDEvZZU3xHHPy4:b5otyvi) ([raw file](https://github.com/sillsdev/web-xforge/blob/aa831ebb90f6dae82e539a9867da8141fdec3d11/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L901)):*

Actually, the NoteThreadChange.ContextBefore will never be different regardless of whether the note is attached or not. Those properties are needed to create a note thread, but they are meaningless when the note thread doc already exists. Same with ContextBefore.
</blockquote></details>

Hmm. But why then do we earlier do

```cs
                TextAnchor expectedAnchor = new TextAnchor
                {
                    Start = rnt.contextBefore.Length,
                    Length = rnt.selectedText.Length
                };
                Assert.That(change1.Position, Is.EqualTo(expectedAnchor));
```
is this the same as asserting that change1.Position is equal to originalAnchor?

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1204)* status: all files reviewed, 3 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1243 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mqu7MLc0aAYTuM_Chg-:-MrP0aA2DKBlqFKWs84H:b-55y2mk) ([raw file](https://github.com/sillsdev/web-xforge/blob/f7941886fbac112ea1cfa34380cadb90101a75b0/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1243)):*

That's helpful, thank you.
What is the meaning of {start: 19, length:5}? Is it the anchoring that was determined when the thread was first created, which would be used to anchor the thread to a particular part of the verse 3 text? Are they values that somehow relate to verse 4? 
Would it be accurate to give `position` a comment above it saying something like: // Original anchoring of thread 4 in verse 3
or something like // Anchoring of tread 4 in verse 4
?
</blockquote></details>

The position could have been any position, we just need to give our test a place to anchor the note thread after it is reattached to verse 4. I left a comment saying "active position of thread04 when reattached to verse 4.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts, line 122 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MrNmaBE9T-cQAbR9SCx:-MrP2mta0S8LCXQ2EyI_:bjjp38z) ([raw file](https://github.com/sillsdev/web-xforge/blob/9814097969bd73d8d7dee7c411ad3dfa1cc87525/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts#L122)):*

How about also here (or in the below test) asserting that the re-attached location verse reference appears where it should?
</blockquote></details>

I've added that to both.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 901 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-Mr8VxMx63knFNE7d6aT:-MrP3RTQAZmVMUKOQ6m0:b-lwj00j) ([raw file](https://github.com/sillsdev/web-xforge/blob/aa831ebb90f6dae82e539a9867da8141fdec3d11/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L901)):*

Hmm. But why then do we earlier do

```cs
                TextAnchor expectedAnchor = new TextAnchor
                {
                    Start = rnt.contextBefore.Length,
                    Length = rnt.selectedText.Length
                };
                Assert.That(change1.Position, Is.EqualTo(expectedAnchor));
```
is this the same as asserting that change1.Position is equal to originalAnchor?
</blockquote></details>

I think you are asking why it appears that the note thread anchor looks important and relates the reattached note thread position, but the ContextBefore and SelectedText appear to be meaningless.
In the NoteThreadChange class, the ContextBefore and SelectedText are only necessary if the note thread doc does not yet exist, it needs to be created with these values. These properties do not change on comments in a thread for any reason, so we never have to check what values they hold apart from creation time. However, other properties like Position do have meaning. If they exist, it means that these values are new and need to be stored.

marksvc (approved):
> Reviewed 8 of 8 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1204)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts, line 141 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1204#-MrT269P1Segxyjb2KJJ:-MrT269P1Segxyjb2KJK:b2brqnz) ([raw file](https://github.com/sillsdev/web-xforge/blob/b7d0128149215b74d5634253f6d33e9a00edb717/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L141)):*
> ```XML
>     }
>     return note.reattached != null ? this.threadDoc.iconReattached.url : this.threadDoc.getNoteIcon(note).url;
>   }
> ```

This seems like a strange fall thru. We do
```py
if note status is Todo
  return whatever the note icon is supposed to be
if note status is Done or Resolved,
  return the resolved icon
if the note is reattached,
  return the reattached url
In any other case not already handled above,
  return whatever the note icon is supposed to be
```
Hmm. Okay :)



_Code quote:_
```XML
    switch (note.status) {
      case NoteStatus.Todo:
        return this.threadDoc.getNoteIcon(note).url;
      case NoteStatus.Done:
      case NoteStatus.Resolved:
        return this.threadDoc.getNoteResolvedIcon(note).url;
    }
    return note.reattached != null ? this.threadDoc.iconReattached.url : this.threadDoc.getNoteIcon(note).url;
  }
```

PR #1205, opened 2021-12-09
Don't request thread position updates when not changed

marksvc:
- Needless.

====
I identified this while working on SF-1417, but am proposing it afterward since it wasn't part of the issue.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1205)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -1300,6 +1300,49 @@ describe('EditorComponent', () => {
       env.dispose();
     }));
 
+    it('does not try to update positions with an unchanged value', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig({ selectedBookNum: 40, selectedChapterNum: 1, selectedSegment: 'verse_1_1' });
+      env.wait();
+
+      const priorThreadId = 'thread02';
+      const priorThreadDoc: NoteThreadDoc = env.getNoteThreadDoc('project01', priorThreadId);
+      const laterThreadDoc: NoteThreadDoc = env.getNoteThreadDoc('project01', 'thread04');
+      const origPriorThreadDocAnchorStart: number = priorThreadDoc.data!.position.start;
+      const origPriorThreadDocAnchorLength: number = priorThreadDoc.data!.position.length;
+      const origLaterThreadDocAnchorStart: number = laterThreadDoc.data!.position.start;
+      const origLaterThreadDocAnchorLength: number = laterThreadDoc.data!.position.length;
+      expect(laterThreadDoc.data!.position.start)
+        .withContext('setup: have some space between the anchorings')
+        .toBeGreaterThan(origPriorThreadDocAnchorStart + origPriorThreadDocAnchorLength);
+
+      const insertedText = 'inserted text';
+      const insertedTextLength = insertedText.length;
+      let priorThreadEditorPos = env.getNoteThreadEditorPosition(priorThreadId);
+
+      // Edit between anchorings
+      env.targetEditor.setSelection(priorThreadEditorPos, 0, 'user');
+      env.wait();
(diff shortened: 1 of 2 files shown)


RaymondLuong3 (changes requested):
Just a small change, then it is good to go. Thanks for cleaning up my previous work.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1205#-:-Mq_Q7ft3s_D_w_oRuqC:b5mjlxt)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1333 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1205#-Mq_PeKp9CXOvNk7DKWj:-Mq_PeKp9CXOvNk7DKWk:b-r45f5j) ([raw file](https://github.com/sillsdev/web-xforge/blob/e9cefc9e5d30c39f48fd4bc07aa8c2be7f80ad67/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1333)):*
> ```
>       expect(laterThreadDoc.data!.position)
>         .withContext('pushed over')
>         .toEqual({ start: origLaterThreadDocAnchorStart + insertedTextLength, length: origLaterThreadDocAnchorLength });
> ```
What if we check the noteThreadDoc's position after this update instead of only checking that submitJson0Op is called. Then we know if the expected update is applied.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1205)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1333 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1205#-Mq_PeKp9CXOvNk7DKWj:-MqaV0u890a6rJewzoJq:bw2ruuo) ([raw file](https://github.com/sillsdev/web-xforge/blob/e9cefc9e5d30c39f48fd4bc07aa8c2be7f80ad67/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1333)):*

> ```
>       expect(laterThreadDoc.data!.position)
>         .withContext('pushed over')
>         .toEqual({ start: origLaterThreadDocAnchorStart + insertedTextLength, length: origLaterThreadDocAnchorLength });
> ```
What if we check the noteThreadDoc's position after this update instead of only checking that submitJson0Op is called. Then we know if the expected update is applied.
</blockquote></details>

I'm a bit confused. So in the test we call `typeCharacters()`, then we assert that the two threads in question have the expected resulting positions, then we assert that the expected calls to submitJson0Op did/didn't happen. 
So isn't it so that we _are_ checking the note thread doc's position after the update, and are confirming that the update is applied?

RaymondLuong3 (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1205#-:-MqodSkZ7l2VxyEyXwqL:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1333 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1205#-Mq_PeKp9CXOvNk7DKWj:-MqodHNt2Ya-igHpswcU:b373czz) ([raw file](https://github.com/sillsdev/web-xforge/blob/e9cefc9e5d30c39f48fd4bc07aa8c2be7f80ad67/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1333)):*

I'm a bit confused. So in the test we call `typeCharacters()`, then we assert that the two threads in question have the expected resulting positions, then we assert that the expected calls to submitJson0Op did/didn't happen. 
So isn't it so that we _are_ checking the note thread doc's position after the update, and are confirming that the update is applied?


</blockquote></details>

Oops, I don't know what I was thinking at the time. This all seems like how I imagined it needs to be. Thanks!

PR #1206, opened 2021-12-10
SF-1447 Correctly show remote edits following a note

RaymondLuong3:
Previously when a user inserts text, remote users see the text update in real-time. Unfortunately, if text is inserted just after a note embed, remote users see the text inserted immediately before the note embed. The first commit changes the behaviour so that remote edits will always be inserted after the note icon if the edit occurs right at the icon's position. The second change listens the the textDoc's remote changes observable, and redraws the note icon based on the note thread's position data. It would be ideal to review the 2 commits separately.
The majority of the first change makes adjustments to the spec, where thread03 used to follow thread04. Because of this change, now thread04 follows thread03.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1206)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -71,6 +71,16 @@ function removeAttribute(op: DeltaOperation, name: string): void {
   }
 }
 
+/** Provides information about a range of text that includes the embedded elements */
+export interface EditorRange {
+  startEditorPosition: number;
+  endEditorPosition: number;
+  embedsWithinRange: number;
+  leadingEmbedCount: number;
+  // not counted in embedsWithinRange
+  trailingEmbedCount: number;
+}
+
 class SegmentInfo {
   length: number = 0;
   origRef?: string;
@@ -381,22 +391,40 @@ export class TextViewModel {
     return undefined;
   }
 
-  /** Returns the editor position that corresponds to a text position past an editor position. */
-  getEditorPositionPlusTextPosition(startingEditorPos: number, textPosPast: number): number {
-    let textCharactersFound = 0;
-    let resultingEditorPos = startingEditorPos;
+  /** Returns editor range information that corresponds to a text position past an editor position. */
+  getEditorContentRange(startEditorPosition: number, textPosPast: number): EditorRange {
     const embedEditorPositions = Array.from(this.embeddedElements.values());
+    const leadingEmbedCount = this.countEmbedsAtPosition(startEditorPosition);
+    let resultingEditorPos = startEditorPosition + leadingEmbedCount;
+    let textCharactersFound = 0;
+    let embedsWithinRange = leadingEmbedCount;
     while (textCharactersFound < textPosPast) {
       if (!embedEditorPositions.includes(resultingEditorPos)) {
         textCharactersFound++;
+      } else {
+        embedsWithinRange++;
       }
       resultingEditorPos++;
     }
-    // if the position ends on an embed, include it in the final editor position i.e. make this greedy
-    while (embedEditorPositions.includes(resultingEditorPos)) {
-      resultingEditorPos++;
+    // trailing embeds do not count towards embedsWithinRange
+    const trailingEmbedCount: number = this.countEmbedsAtPosition(resultingEditorPos);
+    return {
+      startEditorPosition,
+      endEditorPosition: resultingEditorPos,
+      embedsWithinRange,
+      leadingEmbedCount,
+      trailingEmbedCount
+    };
+  }
+
+  private countEmbedsAtPosition(startEditorPosition: number): number {
+    const embedEditorPositions = Array.from(this.embeddedElements.values());
+    // add up the leading embeds
+    let leadingEmbedCount = 0;
+    while (embedEditorPositions.includes(startEditorPosition + leadingEmbedCount)) {
+      leadingEmbedCount++;
     }
-    return resultingEditorPos;
+    return leadingEmbedCount;
   }
 
   private viewToData(delta: DeltaStatic): DeltaStatic {
@@ -657,32 +685,45 @@ export class TextViewModel {
     const adjustedDelta = new Delta();
     let curIndex: number = 0;
     let embedsUpToIndex: number = 0;
+    let previousOp: 'retain' | 'insert' | 'delete' | undefined;
+    let editorStartPos: number = 0;
     for (const op of modelDelta.ops) {
-      let cloneOp: DeltaOperation | undefined = cloneDeep(op);
-      const editorStartPos: number = curIndex + embedsUpToIndex;
+      let cloneOp: DeltaOperation = cloneDeep(op);
+      editorStartPos = curIndex + embedsUpToIndex;
       if (cloneOp.retain != null) {
         // editorStartPos must be the current index plus the number of embeds previous
-        const embedsInRange: number = this.calculateEmbedsInTextRange(editorStartPos, cloneOp.retain);
-        embedsUpToIndex += embedsInRange;
+        const editorRange: EditorRange = this.getEditorContentRange(editorStartPos, cloneOp.retain);
+        embedsUpToIndex += editorRange.embedsWithinRange;
         curIndex += cloneOp.retain;
-        // add to the retain op the number of embedded elements contained in its content
-        cloneOp.retain += embedsInRange;
-      } else if (cloneOp.delete != null) {
-        const embedsInRange: number = this.calculateEmbedsInTextRange(editorStartPos, cloneOp.delete);
-        curIndex += cloneOp.delete;
-        embedsUpToIndex += embedsInRange;
-        // add to the delete op the number of embedded elements contained in its content
-        cloneOp.delete += embedsInRange;
-        if (cloneOp.delete < 1) {
-          cloneOp = undefined;
+        // remove any embeds subsequent to the previous insert so they can be redrawn
+        if (editorRange.leadingEmbedCount > 0 && editorStartPos > 0 && previousOp === 'insert') {
+          (adjustedDelta as any).push({ delete: editorRange.leadingEmbedCount } as DeltaOperation);
         }
+        // add to the retain op the number of embedded elements contained in its content
+        cloneOp.retain += editorRange.embedsWithinRange - editorRange.leadingEmbedCount;
+        previousOp = 'retain';
+      } else if (cloneOp.delete != null) {
+        const editorRange: EditorRange = this.getEditorContentRange(editorStartPos, cloneOp.delete);
+        curIndex += cloneOp.delete;
+        embedsUpToIndex += editorRange.embedsWithinRange;
+        // add to the delete op the number of embedded elements contained in its content
+        cloneOp.delete += editorRange.embedsWithinRange;
+        if (cloneOp.delete < 1) {
+          continue;
+        }
+        previousOp = 'delete';
       } else if (cloneOp.insert != null) {
         cloneOp.attributes = this.getAttributesAtPosition(editorStartPos);
+        previousOp = 'insert';
       }
+      (adjustedDelta as any).push(cloneOp);
+    }
 
-      if (cloneOp != null) {
-        (adjustedDelta as any).push(cloneOp);
-      }
+    editorStartPos = curIndex + embedsUpToIndex;
+    // remove any embeds subsequent the previous insert so they can be redrawn
+    const embedsAfterLastEdit = this.countEmbedsAtPosition(editorStartPos);
+    if (embedsAfterLastEdit > 0 && previousOp === 'insert') {
+      (adjustedDelta as any).push({ delete: embedsAfterLastEdit } as DeltaOperation);
     }
 
     return adjustedDelta;
@@ -703,14 +744,6 @@ export class TextViewModel {
       }
     }
     return embeddedElementsCount;
-  }
-
-  /** Get the number of embeds displayed in the quill editor for a given range of text data. */
-  private calculateEmbedsInTextRange(editorStartPos: number, length: number): number {
-    const editorEndPos: number = this.getEditorPositionPlusTextPosition(editorStartPos, length);
-    const lengthWithEmbeds: number = editorEndPos - editorStartPos;
-    const embedCount: number = lengthWithEmbeds - length;
-    return embedCount;
   }
 
   private getAttributesAtPosition(editorPosition: number) {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -392,6 +392,10 @@ export class TextViewModel {
       }
       resultingEditorPos++;
     }
+    // if the position ends on an embed, include it in the final editor position i.e. make this greedy
+    while (embedEditorPositions.includes(resultingEditorPos)) {
+      resultingEditorPos++;
+    }
     return resultingEditorPos;
   }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -137,7 +137,7 @@ export class TextViewModel {
     if (subscribeToUpdates) {
       this.remoteChangesSub = this.textDoc.remoteChanges$.subscribe(ops => {
         const deltaWithEmbeds: DeltaStatic = this.addEmbeddedElementsToDelta(ops as DeltaStatic);
-        editor.updateContents(deltaWithEmbeds);
+        editor.updateContents(deltaWithEmbeds, 'api');
       });
     }
     this.onCreateSub = this.textDoc.create$.subscribe(() => {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -74,10 +74,11 @@ function removeAttribute(op: DeltaOperation, name: string): void {
 /** Provides information about a range of text that includes the embedded elements */
 export interface EditorRange {
   startEditorPosition: number;
-  endEditorPosition: number;
+  editorLength: number;
   embedsWithinRange: number;
+  /** Count of sequential embeds leading off the range */
   leadingEmbedCount: number;
-  // not counted in embedsWithinRange
+  /** Count of sequential embeds immediately following the range */
   trailingEmbedCount: number;
 }
 
@@ -394,7 +395,7 @@ export class TextViewModel {
   /** Returns editor range information that corresponds to a text position past an editor position. */
   getEditorContentRange(startEditorPosition: number, textPosPast: number): EditorRange {
     const embedEditorPositions = Array.from(this.embeddedElements.values());
-    const leadingEmbedCount = this.countEmbedsAtPosition(startEditorPosition);
+    const leadingEmbedCount = this.countSequentialEmbedsStartingAt(startEditorPosition);
     let resultingEditorPos = startEditorPosition + leadingEmbedCount;
     let textCharactersFound = 0;
     let embedsWithinRange = leadingEmbedCount;
@@ -407,17 +408,17 @@ export class TextViewModel {
       resultingEditorPos++;
     }
     // trailing embeds do not count towards embedsWithinRange
-    const trailingEmbedCount: number = this.countEmbedsAtPosition(resultingEditorPos);
+    const trailingEmbedCount: number = this.countSequentialEmbedsStartingAt(resultingEditorPos);
     return {
       startEditorPosition,
-      endEditorPosition: resultingEditorPos,
+      editorLength: resultingEditorPos - startEditorPosition,
       embedsWithinRange,
       leadingEmbedCount,
       trailingEmbedCount
     };
   }
 
-  private countEmbedsAtPosition(startEditorPosition: number): number {
+  private countSequentialEmbedsStartingAt(startEditorPosition: number): number {
     const embedEditorPositions = Array.from(this.embeddedElements.values());
     // add up the leading embeds
     let leadingEmbedCount = 0;
@@ -695,12 +696,14 @@ export class TextViewModel {
         const editorRange: EditorRange = this.getEditorContentRange(editorStartPos, cloneOp.retain);
         embedsUpToIndex += editorRange.embedsWithinRange;
         curIndex += cloneOp.retain;
-        // remove any embeds subsequent to the previous insert so they can be redrawn
-        if (editorRange.leadingEmbedCount > 0 && editorStartPos > 0 && previousOp === 'insert') {
+        let embedsToRetain: number = editorRange.embedsWithinRange;
+        // remove any embeds subsequent to the previous insert so they can be redrawn in the right place
+        if (editorRange.leadingEmbedCount > 0 && previousOp === 'insert') {
           (adjustedDelta as any).push({ delete: editorRange.leadingEmbedCount } as DeltaOperation);
+          embedsToRetain -= editorRange.leadingEmbedCount;
         }
         // add to the retain op the number of embedded elements contained in its content
-        cloneOp.retain += editorRange.embedsWithinRange - editorRange.leadingEmbedCount;
+        cloneOp.retain += embedsToRetain;
         previousOp = 'retain';
       } else if (cloneOp.delete != null) {
         const editorRange: EditorRange = this.getEditorContentRange(editorStartPos, cloneOp.delete);
@@ -721,7 +724,7 @@ export class TextViewModel {
 
     editorStartPos = curIndex + embedsUpToIndex;
     // remove any embeds subsequent the previous insert so they can be redrawn
-    const embedsAfterLastEdit = this.countEmbedsAtPosition(editorStartPos);
+    const embedsAfterLastEdit = this.countSequentialEmbedsStartingAt(editorStartPos);
     if (embedsAfterLastEdit > 0 && previousOp === 'insert') {
       (adjustedDelta as any).push({ delete: embedsAfterLastEdit } as DeltaOperation);
     }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -74,6 +74,12 @@ export interface FeaturedVerseRefInfo {
   highlight?: boolean;
 }
 
+/** Information about an embed and its associated formatting. */
+export interface EmbedFormat {
+  id: string;
+  formatLength: number;
+}
+
 /** View of an editable text document. Used for displaying Scripture. */
 @Component({
   selector: 'app-text',
@@ -612,18 +618,24 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     }
   }
 
-  removeEmbeddedElements(embedIdsWithLength?: [string, number][]): void {
+  /** Remove embedded elements that are not part of the text data. This may be necessary to preserve the cursor
+   * location when switching between texts. This is also helpful if a particular embed needs to be redrawn.
+   * @param embedFormats An array of the ids and format lengths for each embed to remove. The embed is removed
+   * along with the corresponding length for formatting associated with the embed.
+   * Set to undefined to remove all embeds.
+   */
+  removeEmbeddedElements(embedFormats?: EmbedFormat[]): void {
     if (this.editor == null) {
       return;
     }
     let embedIndicesWithLength: [number, number][] = [];
-    if (embedIdsWithLength == null) {
+    if (embedFormats == null) {
       embedIndicesWithLength = Array.from(this.viewModel.embeddedElements.values()).map(e => [e, 0]);
     } else {
-      for (const [embedId, length] of embedIdsWithLength) {
-        const index: number | undefined = this.viewModel.embeddedElements.get(embedId);
+      for (const embed of embedFormats) {
+        const index: number | undefined = this.viewModel.embeddedElements.get(embed.id);
         if (index != null) {
-          embedIndicesWithLength.push([index, length]);
+          embedIndicesWithLength.push([index, embed.formatLength]);
         }
       }
     }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -13,20 +13,20 @@ import { TranslocoService } from '@ngneat/transloco';
 import { clone } from 'lodash-es';
 import isEqual from 'lodash-es/isEqual';
 import merge from 'lodash-es/merge';
-import Quill, { DeltaOperation, DeltaStatic, RangeStatic, Sources, StringMap } from 'quill';
+import Quill, { DeltaOperation, DeltaStatic, RangeStatic, Sources } from 'quill';
 import { TextAnchor } from 'realtime-server/lib/esm/scriptureforge/models/text-anchor';
 import { VerseRef } from 'realtime-server/lib/esm/scriptureforge/scripture-utils/verse-ref';
 import { fromEvent } from 'rxjs';
 import { PwaService } from 'xforge-common/pwa.service';
 import { SubscriptionDisposable } from 'xforge-common/subscription-disposable';
 import { getBrowserEngine } from 'xforge-common/utils';
-import { Delta, TextDoc, TextDocId } from '../../core/models/text-doc';
+import { Delta, TextDocId } from '../../core/models/text-doc';
 import { SFProjectService } from '../../core/sf-project.service';
 import { NoteThreadIcon } from '../../core/models/note-thread-doc';
 import { VERSE_REGEX } from '../utils';
 import { registerScripture } from './quill-scripture';
 import { Segment } from './segment';
-import { TextViewModel } from './text-view-model';
+import { EditorRange, TextViewModel } from './text-view-model';
 
 const EDITORS = new Set<Quill>();
 
@@ -74,12 +74,6 @@ export interface FeaturedVerseRefInfo {
   highlight?: boolean;
 }
 
-/** Information about an embed and its associated formatting. */
-export interface EmbedFormat {
-  id: string;
-  formatLength: number;
-}
-
 /** View of an editable text document. Used for displaying Scripture. */
 @Component({
   selector: 'app-text',
@@ -93,8 +87,8 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
   @Input() subscribeToUpdates = true;
   @Output() updated = new EventEmitter<TextUpdatedEvent>(true);
   @Output() segmentRefChange = new EventEmitter<string>();
+  @Output() loaded = new EventEmitter(true);
   @Output() focused = new EventEmitter<boolean>(true);
-  @Output() loaded = new EventEmitter<TextDoc>(true);
   lang: string = '';
   // only use USX formats and not default Quill formats
   readonly allowedFormats: string[] = USX_FORMATS;
@@ -205,6 +199,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
   private highlightMarkerHeight: number = 0;
   private _placeholder?: string;
   private displayMessage: string = '';
+
   constructor(
     private readonly projectService: SFProjectService,
     private readonly transloco: TranslocoService,
@@ -545,19 +540,19 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
       return;
     }
 
-    const embedInsertPos: number = this.viewModel.getEditorPositionPlusTextPosition(
+    const editorRange: EditorRange = this.viewModel.getEditorContentRange(
       editorPosOfSegmentToModify.index,
       startTextPosInVerse
     );
+    const embedInsertPos: number = editorRange.endEditorPosition + editorRange.trailingEmbedCount;
 
     this.editor.insertEmbed(embedInsertPos, formatName, format, 'api');
-    const anchorEndPosition: number = startTextPosInVerse + textAnchor.length - 1;
-    const endPosition: number = this.viewModel.getEditorPositionPlusTextPosition(
-      editorPosOfSegmentToModify.index,
-      anchorEndPosition
-    );
+    const endPosition: number = this.viewModel.getEditorContentRange(
+      embedInsertPos,
+      textAnchor.length
+    ).endEditorPosition;
     // add one to include the embed to underline
-    const formatLength: number = endPosition - embedInsertPos + 1;
+    const formatLength: number = endPosition - embedInsertPos;
     this.editor.formatText(embedInsertPos, formatLength, 'text-anchor', 'true', 'api');
     this.updateSegment();
     return embedSegmentRef;
@@ -618,48 +613,24 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     }
   }
 
-  /** Remove embedded elements that are not part of the text data. This may be necessary to preserve the cursor
-   * location when switching between texts. This is also helpful if a particular embed needs to be redrawn.
-   * @param embedFormats An array of the ids and format lengths for each embed to remove. The embed is removed
-   * along with the corresponding length for formatting associated with the embed.
-   * Set to undefined to remove all embeds.
+  /**
+   * Remove embedded elements that are not part of the text data. This may be necessary to preserve the cursor
+   * location when switching between texts.
    */
-  removeEmbeddedElements(embedFormats?: EmbedFormat[]): void {
+  removeEmbeddedElements(): void {
     if (this.editor == null) {
       return;
     }
-    let embedIndicesWithLength: [number, number][] = [];
-    if (embedFormats == null) {
-      embedIndicesWithLength = Array.from(this.viewModel.embeddedElements.values()).map(e => [e, 0]);
-    } else {
-      for (const embed of embedFormats) {
-        const index: number | undefined = this.viewModel.embeddedElements.get(embed.id);
-        if (index != null) {
-          embedIndicesWithLength.push([index, embed.formatLength]);
-        }
-      }
-    }
-
-    const deleteDelta = new Delta();
     let previousEmbedIndex = -1;
-    let textAnchorRemovalLength = 0;
-    const ops: StringMap = { 'text-anchor': false };
-    for (const [embedIndex, textAnchorLength] of embedIndicesWithLength) {
+    const deleteDelta = new Delta();
+    for (const embedIndex of this.viewModel.embeddedElements.values()) {
       const lengthBetweenEmbeds: number = embedIndex - (previousEmbedIndex + 1);
       if (lengthBetweenEmbeds > 0) {
         // retain elements other than notes between the previous and current embed
-        if (textAnchorRemovalLength > 0) {
-          deleteDelta.retain(textAnchorRemovalLength, ops);
-        }
-        deleteDelta.retain(lengthBetweenEmbeds - textAnchorRemovalLength);
-        textAnchorRemovalLength = 0;
+        deleteDelta.retain(lengthBetweenEmbeds);
       }
       deleteDelta.delete(1);
       previousEmbedIndex = embedIndex;
-      textAnchorRemovalLength = Math.max(textAnchorRemovalLength, textAnchorLength);
-    }
-    if (textAnchorRemovalLength > 0) {
-      deleteDelta.retain(textAnchorRemovalLength, ops);
     }
     deleteDelta.chop();
     if (deleteDelta.ops != null && deleteDelta.ops.length > 0) {
@@ -705,7 +676,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     this.viewModel.bind(textDoc, this.subscribeToUpdates);
     this.updatePlaceholderText();
 
-    this.loaded.emit(textDoc);
+    this.loaded.emit();
     this.applyEditorStyles();
   }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -484,7 +484,8 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     return segments;
   }
 
-  /** Embeds an element, with the specified format, into the editor, at the editor position that corresponds to
+  /**
+   * Embeds an element, with the specified format, into the editor, at an editor position that corresponds to
    * the beginning of textAnchor.
    */
   embedElementInline(
@@ -544,15 +545,12 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
       editorPosOfSegmentToModify.index,
       startTextPosInVerse
     );
-    const embedInsertPos: number = editorRange.endEditorPosition + editorRange.trailingEmbedCount;
+    const embedInsertPos: number =
+      editorRange.startEditorPosition + editorRange.editorLength + editorRange.trailingEmbedCount;
 
     this.editor.insertEmbed(embedInsertPos, formatName, format, 'api');
-    const endPosition: number = this.viewModel.getEditorContentRange(
-      embedInsertPos,
-      textAnchor.length
-    ).endEditorPosition;
-    // add one to include the embed to underline
-    const formatLength: number = endPosition - embedInsertPos;
+    const textAnchorRange = this.viewModel.getEditorContentRange(embedInsertPos, textAnchor.length);
+    const formatLength: number = textAnchorRange.editorLength;
     this.editor.formatText(embedInsertPos, formatLength, 'text-anchor', 'true', 'api');
     this.updateSegment();
     return embedSegmentRef;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -13,14 +13,14 @@ import { TranslocoService } from '@ngneat/transloco';
 import { clone } from 'lodash-es';
 import isEqual from 'lodash-es/isEqual';
 import merge from 'lodash-es/merge';
-import Quill, { DeltaOperation, DeltaStatic, RangeStatic, Sources } from 'quill';
+import Quill, { DeltaOperation, DeltaStatic, RangeStatic, Sources, StringMap } from 'quill';
 import { TextAnchor } from 'realtime-server/lib/esm/scriptureforge/models/text-anchor';
 import { VerseRef } from 'realtime-server/lib/esm/scriptureforge/scripture-utils/verse-ref';
 import { fromEvent } from 'rxjs';
 import { PwaService } from 'xforge-common/pwa.service';
 import { SubscriptionDisposable } from 'xforge-common/subscription-disposable';
 import { getBrowserEngine } from 'xforge-common/utils';
-import { Delta, TextDocId } from '../../core/models/text-doc';
+import { Delta, TextDoc, TextDocId } from '../../core/models/text-doc';
 import { SFProjectService } from '../../core/sf-project.service';
 import { NoteThreadIcon } from '../../core/models/note-thread-doc';
 import { VERSE_REGEX } from '../utils';
@@ -87,8 +87,8 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
   @Input() subscribeToUpdates = true;
   @Output() updated = new EventEmitter<TextUpdatedEvent>(true);
   @Output() segmentRefChange = new EventEmitter<string>();
-  @Output() loaded = new EventEmitter(true);
   @Output() focused = new EventEmitter<boolean>(true);
+  @Output() loaded = new EventEmitter<TextDoc>(true);
   lang: string = '';
   // only use USX formats and not default Quill formats
   readonly allowedFormats: string[] = USX_FORMATS;
@@ -199,7 +199,6 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
   private highlightMarkerHeight: number = 0;
   private _placeholder?: string;
   private displayMessage: string = '';
-
   constructor(
     private readonly projectService: SFProjectService,
     private readonly transloco: TranslocoService,
@@ -613,19 +612,42 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     }
   }
 
-  removeEmbeddedElements(): void {
+  removeEmbeddedElements(embedIdsWithLength?: [string, number][]): void {
     if (this.editor == null) {
       return;
     }
-    let previousEmbedIndex = -1;
+    let embedIndicesWithLength: [number, number][] = [];
+    if (embedIdsWithLength == null) {
+      embedIndicesWithLength = Array.from(this.viewModel.embeddedElements.values()).map(e => [e, 0]);
+    } else {
+      for (const [embedId, length] of embedIdsWithLength) {
+        const index: number | undefined = this.viewModel.embeddedElements.get(embedId);
+        if (index != null) {
+          embedIndicesWithLength.push([index, length]);
+        }
+      }
+    }
+
     const deleteDelta = new Delta();
-    for (const embedIndex of this.viewModel.embeddedElements.values()) {
-      // retain elements other than notes between the previous and current embed
-      if (embedIndex > previousEmbedIndex + 1) {
-        deleteDelta.retain(embedIndex - (previousEmbedIndex + 1));
+    let previousEmbedIndex = -1;
+    let textAnchorRemovalLength = 0;
+    const ops: StringMap = { 'text-anchor': false };
+    for (const [embedIndex, textAnchorLength] of embedIndicesWithLength) {
+      const lengthBetweenEmbeds: number = embedIndex - (previousEmbedIndex + 1);
+      if (lengthBetweenEmbeds > 0) {
+        // retain elements other than notes between the previous and current embed
+        if (textAnchorRemovalLength > 0) {
+          deleteDelta.retain(textAnchorRemovalLength, ops);
+        }
+        deleteDelta.retain(lengthBetweenEmbeds - textAnchorRemovalLength);
+        textAnchorRemovalLength = 0;
       }
       deleteDelta.delete(1);
       previousEmbedIndex = embedIndex;
+      textAnchorRemovalLength = Math.max(textAnchorRemovalLength, textAnchorLength);
+    }
+    if (textAnchorRemovalLength > 0) {
+      deleteDelta.retain(textAnchorRemovalLength, ops);
     }
     deleteDelta.chop();
     if (deleteDelta.ops != null && deleteDelta.ops.length > 0) {
@@ -671,7 +693,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     this.viewModel.bind(textDoc, this.subscribeToUpdates);
     this.updatePlaceholderText();
 
-    this.loaded.emit();
+    this.loaded.emit(textDoc);
     this.applyEditorStyles();
   }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
@@ -67,7 +67,7 @@
                 $event.isLocalUpdate
               )
             "
-            (loaded)="onTextLoaded('target', $event)"
+            (loaded)="onTextLoaded('target')"
             (focused)="targetFocused = $event"
             [highlightSegment]="targetFocused"
             [markInvalid]="true"
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
@@ -67,7 +67,7 @@
                 $event.isLocalUpdate
               )
             "
-            (loaded)="onTextLoaded('target')"
+            (loaded)="onTextLoaded('target', $event)"
             (focused)="targetFocused = $event"
             [highlightSegment]="targetFocused"
             [markInvalid]="true"
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -1818,8 +1818,48 @@ describe('EditorComponent', () => {
 
       // SUT 2
       env.wait();
-      expect(env.component.target!.getSegmentText('verse_1_1')).toEqual('target: c' + 'def' + 'hapter 1, verse 1.');
+      expect(env.component.target!.getSegmentText('verse_1_1')).toEqual('target: c' + insert + 'hapter 1, verse 1.');
       expect(noteThread1Doc.data!.position).toEqual(originalNoteThread1TextPos);
+
+      // simulate text changes just before a note embed
+      remoteEditPositionAfterNote = -1;
+      noteCountBeforePosition = 0;
+      // target: |$cdefhapter 1, verse 1.
+      remoteEditTextPos = env.getRemoteEditPosition(notePosition, remoteEditPositionAfterNote, noteCountBeforePosition);
+      insert = 'before';
+      deltaOps = [{ retain: remoteEditTextPos }, { insert: insert }];
+      const insertDelta = new Delta(deltaOps);
+      textDoc.submit(insertDelta);
+      const note1Doc: NoteThreadDoc = env.getNoteThreadDoc('project01', 'thread01');
+      const anchor: TextAnchor = { start: 8 + insert.length, length: 12 };
+      note1Doc.submitJson0Op(op => op.set(nt => nt.position, anchor));
+
+      // SUT 3
+      env.wait();
+      expect(env.component.target!.getSegmentText('verse_1_1')).toEqual('target: ' + insert + 'cdefhapter 1, verse 1.');
+      const range: RangeStatic = env.component.target!.getSegmentRange('verse_1_1')!;
+      expect(env.getNoteThreadEditorPosition('thread01')).toEqual(range.index + anchor.start);
+      const contents = env.targetEditor.getContents(range.index, range.length);
+      expect(contents.ops![0].insert).toEqual('target: ' + insert);
+      expect(contents.ops![0].attributes!['text-anchor']).toBeUndefined();
+
+      // simulate text changes just after a note embed
+      notePosition = env.getNoteThreadEditorPosition('thread01');
+      remoteEditPositionAfterNote = 0;
+      noteCountBeforePosition = 1;
+      // target: before$|cdefhapter 1, verse 1.
+      remoteEditTextPos = env.getRemoteEditPosition(notePosition, remoteEditPositionAfterNote, noteCountBeforePosition);
+      insert = 'ghi';
+      deltaOps = [{ retain: remoteEditTextPos }, { insert: insert }];
+      const insertAfterNoteDelta = new Delta(deltaOps);
+      textDoc.submit(insertAfterNoteDelta);
+
+      // SUT 4
+      env.wait();
+      expect(env.getNoteThreadEditorPosition('thread01')).toEqual(notePosition);
+      expect(env.component.target!.getSegmentText('verse_1_1')).toEqual(
+        'target: before' + insert + 'cdefhapter 1, verse 1.'
+      );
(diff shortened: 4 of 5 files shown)


irahopkinson:
Please explain your second sentence "_There is no way to accurately decide if the text inserted next to an embed should occur before or after the embed for remote users._"? That's not what I expect from ShareDB - I expect that we always know how to handle any edit.

RaymondLuong3 (commented):
Yes, thanks for the question. When a remote change is emitted, it is a remote change based on the text doc data that is void of some formatting information including note embeds. So whether a user is making an edit immediately before or after a note embed, it will look exactly the same for remote users, there is no distinction since note embeds are added on top of text doc data. However, now that I think if it, we could re-draw the note embeds based on the note thread doc. I will explore that.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-:-MqokGADAVvZAGoAQiXp:bauvi6)* status: 0 of 2 files reviewed, all discussions resolved

RaymondLuong3 (commented):
After looking at this again, it seems like the best solution is to redraw note embeds if a user inserts text, and remote users receive those edits. If the edits happen to occur at the location of a note thread embed, the note thread gets redrawn so the edits will appear at the right place relative to the note embed.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-:-MtEQcKk7yGUmRekbNYr:bw5wbvi)* status: 0 of 5 files reviewed, all discussions resolved

marksvc:
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 1280 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwF1za8SY_InZkn3Fp:-MuwF1zb2VFx0XqevPec:b-qvveql) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1280)):*
> ```XML
>   }
> 
>   private refreshNoteEmbedPositions(delta: DeltaStatic): void {
> ```

Can we give a bit more explanation, like is this trying to refresh all note embed positions that occur after the location of an op that changes the text? 
I see we only seem to actually _do_ something if there was an insert. But what if there was a delete, is that not important to likewise process?

marksvc:
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 615 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwAggK9w9U9Xh_3_qG:-MuwAggLCZ0RCvdfzJaQ:b-l0s1vw) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L615)):*
> ```XML
>   }
> 
>   removeEmbeddedElements(embedIdsWithLength?: [string, number][]): void {
> ```

On my first reading, I don't understand why we are doing this or or haven't followed closely enough the different steps to understand what we are now keeping or deleting. Can you make that obvious with a comment?

marksvc:
:eyes:

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1206)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 615 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwAggK9w9U9Xh_3_qG:-Mv03lbn1wGiTGyj7oN6:b1bftt7) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L615)):*

On my first reading, I don't understand why we are doing this or or haven't followed closely enough the different steps to understand what we are now keeping or deleting. Can you make that obvious with a comment?
</blockquote></details>

It's true, this method got more complex. I've added a method comment that should explain more what its purpose is. Primarily, this allows us to remove and redraw embeds that need to be redrawn. It is also necessary to strip out embeds so that the cursor location is maintained when switching between texts.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 1280 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwF1za8SY_InZkn3Fp:-Mv09GVs6YK0Suw1TPKd:b-24t52s) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1280)):*

Can we give a bit more explanation, like is this trying to refresh all note embed positions that occur after the location of an op that changes the text? 
I see we only seem to actually _do_ something if there was an insert. But what if there was a delete, is that not important to likewise process?
</blockquote></details>

I've added an explanation. This method exists because of the scenario that an insertion occurred at the same location as a note embed. Since the remote delta has no knowledge of embeds, there is no way to tell if an insertion occurring at the same index as a note should be inserted before or after the note embed. This redraws those specific notes that are affected.

marksvc (changes requested):
> Reviewed 1 of 5 files at r2, 1 of 4 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1206)* status: 2 of 5 files reviewed, 5 unresolved discussions (waiting on @marksvc and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 615 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwAggK9w9U9Xh_3_qG:-MvL9442BnexIP2EwTJd:b-2j53nk) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L615)):*

It's true, this method got more complex. I've added a method comment that should explain more what its purpose is. Primarily, this allows us to remove and redraw embeds that need to be redrawn. It is also necessary to strip out embeds so that the cursor location is maintained when switching between texts.
</blockquote></details>

What do you mean by switching between texts? Do you mean switching between chapters? Source and target translations? Between different projects? Maybe, interpreting locations between two users' computers?

What does it mean for an embed to have formatting? Is it related to text formatting, like bold and italic? Maybe it allows a custom specification of the editor length of a particular embed, if maybe it's not just of length 1? 

Does it relate to text anchors? I see in code below `textAnchorRemovalLength` is being used. Is the format length to specify the length of anchored text?

What wouldn't get removed if format length were set to 0? I see that when `embedFormats == null` the code just assumes a format length of 0 for all embeds, and it looks like that's assumed not to have trouble.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 1303 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwCvAxBFSWLqLGNojJ:-MuwCvAxBFSWLqLGNojK:bj4f7ve) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1303)):*
> ```XML
>         }
>       } else if (op.insert != null) {
>         insertLength = typeof op.insert === 'string' ? op.insert.length : 1;
> ```

Should this maybe have been `+=`? Or maybe we should have a `break;` that follows?

Also, why are we recording the insert length to pass to removeEmbeddedElements() ?



_Code quote:_
```XML
h = 
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 1290 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MvLDPZM0i_2EpCF8IPX:-MvLDPZM0i_2EpCF8IPY:b-qv841y) ([raw file](https://github.com/sillsdev/web-xforge/blob/06e1db24db3e578eba48dc3a12279fcd777c8dd2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1290)):*
> ```XML
>     const embedsToRefresh: string[] = [];
>     let insertLength: number = 0;
>     for (const op of remoteDelta.ops) {
> ```

What kinds of sets of ops are we expecting, that we would loop over them here? Will we have series of ops with a bunch of ops in the set, like: "retain, insert, retain, delete, insert, retain, insert" ? Or will we only ever have simple sets, like "retain, insert", "insert", and "retain, delete" ?

If we will have a set of a bunch of ops, then won't additional hits of `if (op.retain != null)` below not work correctly, since that section of code (appears to) only know about where it is relative to the start of the retain, not where the retain is in the whole document? In other words, if we have "retain 5, insert abc, retain 7, insert def", and we do the `if (op.retain != null)` for the first retain, it correctly looks at embed editor positions and considers the text position 5. But once we get to the "retain 7", the `if (op.retain != null)` section doesn't know that it should be using text position 5+"abc".length+7=15. Instead, it continues incorrectly using text position 7 when considering where embeds are.

If the set of ops will actually be smaller, like "retain, insert", then the `for` loop seems like an adequate but odd  way to process it. I think in that situation we should mention in a comment the kind of data we expect to be processing, like "deltas will come in with ops that contain no more than 1 retain, insert, or delete" or something. To help people reading the code understand what's being worked with.



_Code quote:_
```XML
   for (const op of remoteDelta.ops) {
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 1303 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MvLBJsX5hrupouf9cft:-MvLBJsX5hrupouf9cfu:bk2ws1k) ([raw file](https://github.com/sillsdev/web-xforge/blob/06e1db24db3e578eba48dc3a12279fcd777c8dd2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1303)):*
> ```XML
>           }
>           // an embed is positioned before or at the remote edit position, so increment the insertion position
>           // so we can compare it to an embed's editor position
> ```

IIRC, we have helper methods that do things like compute an editor position based on a text position. It seems like this block of code could be simpler if we implemented it like the following, rather than using a loop and gradually walking toward where the editor position and text position correspond. What do you think?

```
for each op of remoteDelta.ops:
  if the op is a retain:
    const whereWeStoppedRetaining = op.retain
    const editorPos = convertTextPosToEditorPos(whereWeStoppedRetaining)
    if editorPos is in this.target.embeddedElements.values():
      embedsToRefresh.push(embedId for embed at pos editorPos)
  if the op is an insert: 
    ...
```

Hmm, it occurs to me that if there are multiple embeds that are adjacent, they correspond to the same text position. So we would need to be comparing positions, not by converting a text pos to an editor pos, but by converting from editor-to-text positions. Maybe something like this:
```
for each op of remoteDelta.ops:
  if the op is a retain:
    const insertionTextPos: number = op.retain
    const embedsAtInsertionPos = this.target.embeddedElements.filter(embed => {
      const correspondingTextPos = editorPosToTextPos(embed.editorPosition)
      return correspondingTextPos === insertionTextPos
    })
```



_Code quote:_
```XML
          // an embed is positioned before or at the remote edit position, so increment the insertion position
          // so we can compare it to an embed's editor position
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1206)* status: 2 of 5 files reviewed, 5 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 615 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwAggK9w9U9Xh_3_qG:-MvPrarN90hrlqCAv0wZ:bbc7fft) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L615)):*

What do you mean by switching between texts? Do you mean switching between chapters? Source and target translations? Between different projects? Maybe, interpreting locations between two users' computers?

What does it mean for an embed to have formatting? Is it related to text formatting, like bold and italic? Maybe it allows a custom specification of the editor length of a particular embed, if maybe it's not just of length 1? 

Does it relate to text anchors? I see in code below `textAnchorRemovalLength` is being used. Is the format length to specify the length of anchored text?

What wouldn't get removed if format length were set to 0? I see that when `embedFormats == null` the code just assumes a format length of 0 for all embeds, and it looks like that's assumed not to have trouble.

</blockquote></details>

1. Switching between text docs (chapters, books)
2. Embeds can be formatted such as with the text-anchor format, and presumably bold/italics. In this case we are only concerned about the text anchor formatting. This is because when text is inserted, it picks up the formatting of the preceeding text. If it turns out that the text inserted should come before the note icon, then it should have its text-anchor formatting removed.
3. The format length is the length of text to strip the text anchor format off, in this case, it applies to the length of text that was inserted remotely at the location of the note icon.
4. If the format length to remove is 0, do not update any formatting on the text anchor of that embed that was removed.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 1303 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwCvAxBFSWLqLGNojJ:-MvQ2Uiw44-VVdkEc-Hv:b-2u2bxq) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1303)):*

Should this maybe have been `+=`? Or maybe we should have a `break;` that follows?

Also, why are we recording the insert length to pass to removeEmbeddedElements() ?
</blockquote></details>

The insert length helps us to identify what is the length of text inserted by the remote user that we need to strip the text-anchor formatting from (since we don't want that remote inserted text to be underlined if it does not come after the note icon.)
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 1290 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MvLDPZM0i_2EpCF8IPX:-MvQ11-oEZUVGhW5wGm7:b-aipafn) ([raw file](https://github.com/sillsdev/web-xforge/blob/06e1db24db3e578eba48dc3a12279fcd777c8dd2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1290)):*

What kinds of sets of ops are we expecting, that we would loop over them here? Will we have series of ops with a bunch of ops in the set, like: "retain, insert, retain, delete, insert, retain, insert" ? Or will we only ever have simple sets, like "retain, insert", "insert", and "retain, delete" ?

If we will have a set of a bunch of ops, then won't additional hits of `if (op.retain != null)` below not work correctly, since that section of code (appears to) only know about where it is relative to the start of the retain, not where the retain is in the whole document? In other words, if we have "retain 5, insert abc, retain 7, insert def", and we do the `if (op.retain != null)` for the first retain, it correctly looks at embed editor positions and considers the text position 5. But once we get to the "retain 7", the `if (op.retain != null)` section doesn't know that it should be using text position 5+"abc".length+7=15. Instead, it continues incorrectly using text position 7 when considering where embeds are.

If the set of ops will actually be smaller, like "retain, insert", then the `for` loop seems like an adequate but odd  way to process it. I think in that situation we should mention in a comment the kind of data we expect to be processing, like "deltas will come in with ops that contain no more than 1 retain, insert, or delete" or something. To help people reading the code understand what's being worked with.


</blockquote></details>

Yes, that is a good question. In this case, we are only concerned with the first retain and insert ops since remote edits generally only have those two ops. I've added a break statement for the case that we found our insert op. I can imagine that maybe we need to consider the case where a user is offline and makes a series of edits, then goes back online. In that case, we need to consider all ops. But that is worth its own issue. I've left a comment in the jira card.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 1303 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MvLBJsX5hrupouf9cft:-MvQ1xz9A5iJJ4ImUzaA:b-tsej1u) ([raw file](https://github.com/sillsdev/web-xforge/blob/06e1db24db3e578eba48dc3a12279fcd777c8dd2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1303)):*

IIRC, we have helper methods that do things like compute an editor position based on a text position. It seems like this block of code could be simpler if we implemented it like the following, rather than using a loop and gradually walking toward where the editor position and text position correspond. What do you think?

```
for each op of remoteDelta.ops:
  if the op is a retain:
    const whereWeStoppedRetaining = op.retain
    const editorPos = convertTextPosToEditorPos(whereWeStoppedRetaining)
    if editorPos is in this.target.embeddedElements.values():
      embedsToRefresh.push(embedId for embed at pos editorPos)
  if the op is an insert: 
    ...
```

Hmm, it occurs to me that if there are multiple embeds that are adjacent, they correspond to the same text position. So we would need to be comparing positions, not by converting a text pos to an editor pos, but by converting from editor-to-text positions. Maybe something like this:
```
for each op of remoteDelta.ops:
  if the op is a retain:
    const insertionTextPos: number = op.retain
    const embedsAtInsertionPos = this.target.embeddedElements.filter(embed => {
      const correspondingTextPos = editorPosToTextPos(embed.editorPosition)
      return correspondingTextPos === insertionTextPos
    })
```

</blockquote></details>

Good thinking. I've updated my method to determine what the corresponding text position of a given embed is, and compare that to the position of the insert. It turned out to be pretty simple to incorporate your second pseudocode suggestion.

marksvc:
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 615 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwAggK9w9U9Xh_3_qG:-MvQUTYZ24bQxKUFA_18:b8tb68e) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L615)):*

1. Switching between text docs (chapters, books)
2. Embeds can be formatted such as with the text-anchor format, and presumably bold/italics. In this case we are only concerned about the text anchor formatting. This is because when text is inserted, it picks up the formatting of the preceeding text. If it turns out that the text inserted should come before the note icon, then it should have its text-anchor formatting removed.
3. The format length is the length of text to strip the text anchor format off, in this case, it applies to the length of text that was inserted remotely at the location of the note icon.
4. If the format length to remove is 0, do not update any formatting on the text anchor of that embed that was removed.
</blockquote></details>

What do you mean, then, that the cursor location is maintained when switching between text docs. I tried putting the insertion point in a verse and moving to another chapter, but the insertion point isn't there any more. Are you meaning that if I put the insertion point in Ruth 2:3 and click to go to Ruth 3, that the desired behaviour is for my insertion point to be in Ruth 3:3? Or what do you mean?

When we call removeEmbeddedElements() with no arguments (and it looks like the only place that happens is from editor.component.ts), and the note thread embeds are removed,  we would be  left with a bunch of text anchor area underlining formatting, is that so? I see that we only ever do this right before changing the text content.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1206)* status: 2 of 5 files reviewed, 5 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 615 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwAggK9w9U9Xh_3_qG:-MvUWxNxEAibjs41d70z:bw879mx) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L615)):*

What do you mean, then, that the cursor location is maintained when switching between text docs. I tried putting the insertion point in a verse and moving to another chapter, but the insertion point isn't there any more. Are you meaning that if I put the insertion point in Ruth 2:3 and click to go to Ruth 3, that the desired behaviour is for my insertion point to be in Ruth 3:3? Or what do you mean?

When we call removeEmbeddedElements() with no arguments (and it looks like the only place that happens is from editor.component.ts), and the note thread embeds are removed,  we would be  left with a bunch of text anchor area underlining formatting, is that so? I see that we only ever do this right before changing the text content.




</blockquote></details>

If a user has their cursor in a particular position, and then goes to the next chapter and back, the cursor will be in the last position on the original text. This only happens if a user does not make a text selection in the second text.

When calling removeEmbeddedElements, the goal is to remove the element and text anchor formatting if necessary. Removing formatting is only necessary for remote edits so that you do not get underlined text where it does not belong. When removing embeds when switching texts, the text anchor formatting could be removed if desired, but since there is no benefit to that, we simply pass 0 to the formatLength resulting in no formatting removed from the embed's text anchor.

marksvc (changes requested):
In my testing, if I have 

`abXcd`

on one client, where X is a note thread icon, and I type anything after the X, it displays in the same place on another client.
But if I type between `b` and `X`, then although on my editing client I see the new character between `b` and `X`, on the other client it appears between `X` and `c`. On the other client, if I then move away from the chapter and back to the chapter, it then correctly shows the new character between `b` and `X` (to my surprise and interest).

Is this the behaviour that's expected?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-:-MvVJpcI0WvTIvplWZw0:b-mwnli0)* status: 2 of 5 files reviewed, 4 unresolved discussions (waiting on @marksvc and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 615 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwAggK9w9U9Xh_3_qG:-MvQ_cjhE9G_qDxi8XpU:bryqnm8) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L615)):*

If a user has their cursor in a particular position, and then goes to the next chapter and back, the cursor will be in the last position on the original text. This only happens if a user does not make a text selection in the second text.

When calling removeEmbeddedElements, the goal is to remove the element and text anchor formatting if necessary. Removing formatting is only necessary for remote edits so that you do not get underlined text where it does not belong. When removing embeds when switching texts, the text anchor formatting could be removed if desired, but since there is no benefit to that, we simply pass 0 to the formatLength resulting in no formatting removed from the embed's text anchor.
</blockquote></details>

Thank you for your explanations and patience while I learn what's happening here. 
I don't understand yet from removeEmbeddedElements() where it actually de-formats text.

I am uneasy about the change to the removeEmbeddedElements() method. It seems overloaded and fiddly that it now receives a set of embed positions and optional lengths of text to de-format. 

What do you think about having removedEmbeddedElements() only remove the actual embedded elements (not handling formatting), and then in calling code, subsequently do something like
```ts
foreach (editorPos in positionsOfAffectedEmbeds) {
  this.target.editor.formatText(editorPos, newRemoteTextLength, 'text-anchor', 'false', 'api');
}
``` 
to remove the undesired formatting?

I also wonder if we could get away with retaining the prior removeEmbeddedElements() behaviour, and then calling something like what EditorComponent updateVerseNoteThreadAnchors() is doing. We could compute the note icon positions for an affected segment, and apply those new positions and resulting formatting. 
Hmm, I see that EditorComponent is calling updateVerseNoteThreadAnchors() from onTargetUpdated(). But it specifically does _not_ do so if it is a remote edit. I wonder if we could get the functionality we want by helping EditorComponent onTargetUpdated() fix up the note icons and formatting (by allowing it to do it in response to remote edits).
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 1280 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwF1za8SY_InZkn3Fp:-MvVKlUv6ORV9FlCFVk-:b-9crms3) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1280)):*

I've added an explanation. This method exists because of the scenario that an insertion occurred at the same location as a note embed. Since the remote delta has no knowledge of embeds, there is no way to tell if an insertion occurring at the same index as a note should be inserted before or after the note embed. This redraws those specific notes that are affected.
</blockquote></details>

Okay. And what about deletes? If I backspace or delete characters around a note icon, I see funny behaviour on my other client. Is that out of scope for this change?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 1303 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwCvAxBFSWLqLGNojJ:-MvVLQAXEKAGsYt_2VA0:b-kh1tyh) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1303)):*

The insert length helps us to identify what is the length of text inserted by the remote user that we need to strip the text-anchor formatting from (since we don't want that remote inserted text to be underlined if it does not come after the note icon.)
</blockquote></details>

Thank you. Hmm. When a text change happens, and it corresponds to a note icon position, it's ambiguous on the receiving client as to whether it should be before of after the flag. In my testing on my machine, I am generally seeing the change come thru on the other client _after_  the flag, and  with underlining formatting. I am suspecting that I must be observing something that isn't intentional if we are meaning to be _removing_ formatting from the text?

RaymondLuong3 (commented):
Ok, you found something that I had missed. It looks like there must be some race-conditions going on and we cannot trust that the editor embedded elements are up to date. I found that a much simpler solution is to delete the embeds that need to be redrawn when we update the current view of the text with the remote delta instead of doing it after we apply the remote delta.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-:-MvepRz5D0KIt8hhNeRF:b-yewt41)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 615 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwAggK9w9U9Xh_3_qG:-Mveq80pB8I21bq0HXiK:b-pp040x) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L615)):*

Thank you for your explanations and patience while I learn what's happening here. 
I don't understand yet from removeEmbeddedElements() where it actually de-formats text.

I am uneasy about the change to the removeEmbeddedElements() method. It seems overloaded and fiddly that it now receives a set of embed positions and optional lengths of text to de-format. 

What do you think about having removedEmbeddedElements() only remove the actual embedded elements (not handling formatting), and then in calling code, subsequently do something like
```ts
foreach (editorPos in positionsOfAffectedEmbeds) {
  this.target.editor.formatText(editorPos, newRemoteTextLength, 'text-anchor', 'false', 'api');
}
``` 
to remove the undesired formatting?

I also wonder if we could get away with retaining the prior removeEmbeddedElements() behaviour, and then calling something like what EditorComponent updateVerseNoteThreadAnchors() is doing. We could compute the note icon positions for an affected segment, and apply those new positions and resulting formatting. 
Hmm, I see that EditorComponent is calling updateVerseNoteThreadAnchors() from onTargetUpdated(). But it specifically does _not_ do so if it is a remote edit. I wonder if we could get the functionality we want by helping EditorComponent onTargetUpdated() fix up the note icons and formatting (by allowing it to do it in response to remote edits).


</blockquote></details>

You are right, this method started to get cluggy, and I'm glad that we no longer need to overload this method since embeds are now removed as part of applying the remote edit op to our quill editor view.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 1280 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwF1za8SY_InZkn3Fp:-MveqUKw7GNsu1TeALX7:b-cxchxo) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1280)):*

Okay. And what about deletes? If I backspace or delete characters around a note icon, I see funny behaviour on my other client. Is that out of scope for this change?
</blockquote></details>

No longer relevant.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts, line 1303 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MuwCvAxBFSWLqLGNojJ:-MveqbLqDabMuoDFHRpP:b33n096) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfac5feb00fede32088e653e06cd2d9183c6054c/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1303)):*

Thank you. Hmm. When a text change happens, and it corresponds to a note icon position, it's ambiguous on the receiving client as to whether it should be before of after the flag. In my testing on my machine, I am generally seeing the change come thru on the other client _after_  the flag, and  with underlining formatting. I am suspecting that I must be observing something that isn't intentional if we are meaning to be _removing_ formatting from the text?
</blockquote></details>

It appears that some times the embedded elements are not always up to date, this was causing issues that I'm glad we found. So now we do this a different way that is less prone to race conditions.

marksvc (changes requested):
:thumbsup: 

> Reviewed 4 of 5 files at r6.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-:-MwcgqDv4JGIr84bWuBD:b-jd3vnx)* status: 4 of 5 files reviewed, 15 unresolved discussions (waiting on @marksvc and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 75 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwcA5OE3qouZb5MNefN:-MwcA5OE3qouZb5MNefO:b-5mj8pi) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L75)):*
> ```XML
> 
> /** Provides information about a range of text that includes the embedded elements */
> export interface EditorRange {
> ```

[Okay, well, after looking at how EditorRange is used in the code, I see this wouldn't be helpful to provide `startTextPosition` and `textLength`, but I'll mention it anyway in case it's a useful idea for another time.]
BTW, one way this could be defined is with both editor and text positions, like this, if found to be helpful:

```ts
export class EditorRange {
  /** Position within the whole of the editor. */
  public startEditorPosition: number;
  /** Corresponding text position, within the whole of the text doc loaded in the editor. */
  public startTextPosition: number;
  public editorLength: number;
  public textLength: number;
  public get embedCount(): number {
    return editorLength - textLength;
  }
  constructor(startEditorPosition: number, editorLength: number) {
    ...
    this.startTextPosition = editorToTextPosition(startEditorPosition);
    ...
  }
}
```
And it wouldn't need to define `leadingEmbedCount` or `trailingEmbedCount`, but rather clients could be expected to call 
`sequenceOfEmbedsStartingAt(editorRange.startEditorPosition)` and 
`sequenceOfEmbedsStartingAt(editorRange.startEditorPosition + editorRange.editorLength)`
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 78 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYPPDX8rLWdyVfvvFL:-MwYPPDX8rLWdyVfvvFM:b-r1036v) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L78)):*
> ```XML
>   startEditorPosition: number;
>   endEditorPosition: number;
>   embedsWithinRange: number;
> ```

Using 'endEditorPosition' can be unclear if we mean an inclusive or exclusive end, and so too "embedsWithinRange" can therefore be unclear as well. Let's either specify if endEditorPosition is inclusive or exclusive (for example, if we start using position 3 and we end using position 4, does the range include the characters at positions 3 and 4, or just the character at position 3? From looking at how we make an editor range, I think endEditorPosition is _exclusive_, not including the character at position 4), **or use `editorLength` instead**, to give a length instead of a final position. 
In other places where we had a start and a _length_, I've found it to be more clear and also a bit helpful for interacting with other methods.



_Code quote:_
```XML
  endEditorPosition: number;
  embedsWithinRange: number;
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 79 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYUamf9hnG1GWO6Ocz:-MwYUamf9hnG1GWO6Od-:b873izz) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L79)):*
> ```XML
>   endEditorPosition: number;
>   embedsWithinRange: number;
>   leadingEmbedCount: number;
> ```

So I definitely thought "leadingEmbedCount" was the number of embeds _before_ the EditorRange object. But now that I'm reading some of the rest of the diff, I see it means "number of embeds at the beginning of the EditorRange". Maybe.. we should at least add a comment to `leadingEmbedCount` about what is meant :)
let's either 
(1) add a comment, like // number of sequential embeds at the beginning of the range
or (2) rename field, like "startingEmbedCount" or "startingSequentialEmbedCount".



_Code quote:_
```XML
leadingEmbedCount:
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 81 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYX9R-Cfb-qantfXyb:-MwYX9R-Cfb-qantfXyc:bgmugcg) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L81)):*
> ```XML
>   leadingEmbedCount: number;
>   // not counted in embedsWithinRange
>   trailingEmbedCount: number;
> ```

So, I had thought trailingEmbedCount was the count of embeds that come after the EditorRange. But I see from the code that it is the count of the sequential embeds that immediately follow the EditorRange. 
In other words, with editor content: abcXXdefXXghiXXjkl
If an EditorRange went from "bcXXdef", I thought trailingEmbedCount would be 4, since there are 4 embeds that occur after the Editor Range. But I see that the trailingEmbedCount would actually be 2.
Can we either
(1) add a comment, like "count of sequential embeds that immediately follow the range"
or (2) rename the field to something like "trailingSequentialEmbedCount" or "adjacentEmbedsFollowingCount".
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 397 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYXbWT57MLas4c0H1m:-MwYXbWT57MLas4c0H1n:b-ma95zg) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L397)):*
> ```XML
>   getEditorContentRange(startEditorPosition: number, textPosPast: number): EditorRange {
>     const embedEditorPositions = Array.from(this.embeddedElements.values());
>     const leadingEmbedCount = this.countEmbedsAtPosition(startEditorPosition);
> ```

Might there be a special case for when `textPosPast === 0`? What should be the behaviour? Should the resulting EditorRange contain the count of the sequential embeds that begin at `startEditorPosition`? I think "yes" would be okay. 



_Code quote:_
```XML
 leadingEmbedCount =
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 420 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYR5Ei33MqViVYH6Ko:-MwYR5Ej0zziWLJGJURO:b2br497) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L420)):*
> ```XML
>   }
> 
>   private countEmbedsAtPosition(startEditorPosition: number): number {
> ```

I think this method name is .. maybe not super accurate (because we should only have one embed at a given editor position (if I'm correct in that)). Let's do one of 
(1) Rename the method, like "countSequentialEmbedsStartingAt" (which which still doesn't 100% solve it :), or 
(2) Writing a method comment like `return count of sequential embeds starting at editor position`.



_Code quote:_
```XML
countEmbedsAtPosition(
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 423 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYQnxh28aYvlkteh3u:-MwYQnxh28aYvlkteh3v:b-qkp9u8) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L423)):*
> ```XML
>     const embedEditorPositions = Array.from(this.embeddedElements.values());
>     // add up the leading embeds
>     let leadingEmbedCount = 0;
> ```

Let's use a different name than "leading". I was confused and thought we were going to count embeds "leading up to" the starting position.

Maybe just call them embeds? Like:
```ts
// add up the embeds
let embedCount = 0;
...
```

_Hmm_. Actually: I am looking at the dictionary and think we can keep "leading" if you prefer it.



_Code quote:_
```XML
    // add up the leading embeds
    let leadingEmbedCount = 0;
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 698 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYZeKI3LhNaZDcbekc:-MwYZeKI3LhNaZDcbekd:b-2x3ep4) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L698)):*
> ```XML
>         embedsUpToIndex += editorRange.embedsWithinRange;
>         curIndex += cloneOp.retain;
>         // remove any embeds subsequent to the previous insert so they can be redrawn
> ```

If ops come thru like this:
```
ops {
  0: { insert "hello" }
  1: { delete: 2 }
  2: { retain: 3 }
}
```
where we have an insert, but something other than a retain that immediately follows, we won't do this delete step.



_Code quote:_
```XML
 // remove any embeds subsequent to the previous insert so they can be redrawn
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 698 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mwcb1NV-vqVk4fvqfPL:-Mwcb1NV-vqVk4fvqfPM:bmtk23q) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L698)):*
> ```XML
>         embedsUpToIndex += editorRange.embedsWithinRange;
>         curIndex += cloneOp.retain;
>         // remove any embeds subsequent to the previous insert so they can be redrawn
> ```

How about "redrawn in the right place"



_Code quote:_
```XML
redrawn
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 699 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwcbAxKDhiXmbjHTpMr:-MwcbAxKDhiXmbjHTpMs:b-d78bth) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L699)):*
> ```XML
>         curIndex += cloneOp.retain;
>         // remove any embeds subsequent to the previous insert so they can be redrawn
>         if (editorRange.leadingEmbedCount > 0 && editorStartPos > 0 && previousOp === 'insert') {
> ```

What's the purpose of `editorStartPos>0` here?



_Code quote:_
```XML
editorStartPos > 0
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 703 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYaX9sBHYcYwNDmb4n:-MwYaX9sBHYcYwNDmb4o:bgwrd8w) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L703)):*
> ```XML
>         }
>         // add to the retain op the number of embedded elements contained in its content
>         cloneOp.retain += editorRange.embedsWithinRange - editorRange.leadingEmbedCount;
> ```

Don't we only want to subtract editorRange.leadingEmbedCount if a delete op was pushed a few lines above?



_Code quote:_
```XML
 - editorRange.leadingEmbedCount;
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 711 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYb8ar4w4b3T_iio5d:-MwYb8ar4w4b3T_iio5e:b6z0p3v) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L711)):*
> ```XML
>         // add to the delete op the number of embedded elements contained in its content
>         cloneOp.delete += editorRange.embedsWithinRange;
>         if (cloneOp.delete < 1) {
> ```

I'm a bit confused how the  placement of this condition, and its relation to prior lines of code, are intended to relate. 
(Would moving `if cloneOp.delete <1; continue` up higher be as good?) 
If cloneOp.delete===0, then getEditorContentRange() could possibly provide a count of initial sequential embeds, iirc, and so embedsUpToIndex and cloneOp.delete will be increased. Tho then cloneOp.delete won't equal 0 anymore to do the `continue` here. Is that what the thinking is?



_Code quote:_
```XML
        if (cloneOp.delete < 1) {
          continue;
        }
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 714 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYc4xU8npAtsjj_xJQ:-MwYc4xU8npAtsjj_xJR:bcikyy8) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L714)):*
> ```XML
>           continue;
>         }
>         previousOp = 'delete';
> ```

This previousOp setting will need to be done earlier since there is a `continue` above. Oh, unless you are meaning to kind of ignore the delete op if it's a meaningless 0-length delete?



_Code quote:_
```XML
 previousOp = 'delete';
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 487 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mwc4B0f6AHXOkB3PMuD:-Mwc4B0f6AHXOkB3PMuE:b-uk0lbu) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L487)):*
> ```XML
>   }
> 
>   /** Embeds an element, with the specified format, into the editor, at the editor position that corresponds to
> ```

Hmm this should probably be  "at an editor position" ("an" not "the"), since multiple editor positions correspond to the same text position.



_Code quote:_
```XML
the
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 549 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mwc4pOl49j_iz2SZQmd:-Mwc4pOm8wRnFB7Y8Roj:bdt391a) ([raw file](https://github.com/sillsdev/web-xforge/blob/eb8a1c10ae30a0c277a1c6f30faa6157a54c98a9/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L549)):*
> ```XML
> 
>     this.editor.insertEmbed(embedInsertPos, formatName, format, 'api');
>     const anchorEndPosition: number = startTextPosInVerse + textAnchor.length - 1;
> ```

Hmm, why was there a -1 here? I don't understand, to know if maybe we still need it :)



_Code quote:_
```XML
 - 1;
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 554 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mwc59gF0ixanzXNZNXR:-Mwc59gF0ixanzXNZNXS:b5jszwn) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L554)):*
> ```XML
>       textAnchor.length
>     ).endEditorPosition;
>     // add one to include the embed to underline
> ```

And we'll want to remove this line if we aren't doing `+1` below any more.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1206)* status: 4 of 5 files reviewed, 14 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 75 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwcA5OE3qouZb5MNefN:-MwgyKrz361mbzGWAvA3:b4xgvop) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L75)):*

[Okay, well, after looking at how EditorRange is used in the code, I see this wouldn't be helpful to provide `startTextPosition` and `textLength`, but I'll mention it anyway in case it's a useful idea for another time.]
BTW, one way this could be defined is with both editor and text positions, like this, if found to be helpful:

```ts
export class EditorRange {
  /** Position within the whole of the editor. */
  public startEditorPosition: number;
  /** Corresponding text position, within the whole of the text doc loaded in the editor. */
  public startTextPosition: number;
  public editorLength: number;
  public textLength: number;
  public get embedCount(): number {
    return editorLength - textLength;
  }
  constructor(startEditorPosition: number, editorLength: number) {
    ...
    this.startTextPosition = editorToTextPosition(startEditorPosition);
    ...
  }
}
```
And it wouldn't need to define `leadingEmbedCount` or `trailingEmbedCount`, but rather clients could be expected to call 
`sequenceOfEmbedsStartingAt(editorRange.startEditorPosition)` and 
`sequenceOfEmbedsStartingAt(editorRange.startEditorPosition + editorRange.editorLength)`



</blockquote></details>

Yea, we can keep this in mind as we keep adding features that require a more complex EditorRange.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 78 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYPPDX8rLWdyVfvvFL:-Mwgh4Z9DJSeQuyAF1Lv:b9ipjep) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L78)):*

Using 'endEditorPosition' can be unclear if we mean an inclusive or exclusive end, and so too "embedsWithinRange" can therefore be unclear as well. Let's either specify if endEditorPosition is inclusive or exclusive (for example, if we start using position 3 and we end using position 4, does the range include the characters at positions 3 and 4, or just the character at position 3? From looking at how we make an editor range, I think endEditorPosition is _exclusive_, not including the character at position 4), **or use `editorLength` instead**, to give a length instead of a final position. 
In other places where we had a start and a _length_, I've found it to be more clear and also a bit helpful for interacting with other methods.
</blockquote></details>

I like the idea of the editor length, since that is not confusing. I'll update the EditorRange interface. You are correct, the endEditorPosition was exclusive. 
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 79 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYUamf9hnG1GWO6Ocz:-MwghHWqB2BE8dd9GRS_:b-d7eqdd) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L79)):*

So I definitely thought "leadingEmbedCount" was the number of embeds _before_ the EditorRange object. But now that I'm reading some of the rest of the diff, I see it means "number of embeds at the beginning of the EditorRange". Maybe.. we should at least add a comment to `leadingEmbedCount` about what is meant :)
let's either 
(1) add a comment, like // number of sequential embeds at the beginning of the range
or (2) rename field, like "startingEmbedCount" or "startingSequentialEmbedCount".
</blockquote></details>

I've added a comment to the leadingEmbedCount field in the EditorRange interface.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 81 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYX9R-Cfb-qantfXyb:-MwgheuaEIHawAAfLDnM:b-3uhmwk) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L81)):*

So, I had thought trailingEmbedCount was the count of embeds that come after the EditorRange. But I see from the code that it is the count of the sequential embeds that immediately follow the EditorRange. 
In other words, with editor content: abcXXdefXXghiXXjkl
If an EditorRange went from "bcXXdef", I thought trailingEmbedCount would be 4, since there are 4 embeds that occur after the Editor Range. But I see that the trailingEmbedCount would actually be 2.
Can we either
(1) add a comment, like "count of sequential embeds that immediately follow the range"
or (2) rename the field to something like "trailingSequentialEmbedCount" or "adjacentEmbedsFollowingCount".

</blockquote></details>

I've added the comment.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 397 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYXbWT57MLas4c0H1m:-MwgmWQe0rPsVKlcFeBa:b-k7pzou) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L397)):*

Might there be a special case for when `textPosPast === 0`? What should be the behaviour? Should the resulting EditorRange contain the count of the sequential embeds that begin at `startEditorPosition`? I think "yes" would be okay. 
</blockquote></details>

I think yes would be reasonable. If we call getEditorContentRange for a text length of 0, then we return the number of embeds at and after the editor position before the first text character
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 420 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYR5Ei33MqViVYH6Ko:-MwgnNvyEMV2YFhjX_Vd:b-qy6a43) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L420)):*

I think this method name is .. maybe not super accurate (because we should only have one embed at a given editor position (if I'm correct in that)). Let's do one of 
(1) Rename the method, like "countSequentialEmbedsStartingAt" (which which still doesn't 100% solve it :), or 
(2) Writing a method comment like `return count of sequential embeds starting at editor position`.

</blockquote></details>

Good thinking. I renamed the method.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 698 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYZeKI3LhNaZDcbekc:-MwgocaE4Vcn-ziPq3Ve:byio8ui) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L698)):*

If ops come thru like this:
```
ops {
  0: { insert "hello" }
  1: { delete: 2 }
  2: { retain: 3 }
}
```
where we have an insert, but something other than a retain that immediately follows, we won't do this delete step.


</blockquote></details>

That's right.  Delete op automatically deletes the leading embeds. We needed to add this for retain ops, too. And we should not see insert ops in sequence.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 698 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mwcb1NV-vqVk4fvqfPL:-Mwgp3Ye3Hi0OXC64m1N:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L698)):*

How about "redrawn in the right place"
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 699 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwcbAxKDhiXmbjHTpMr:-MwgpmHP9X4YgQ26gA0R:b-dlfoir) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L699)):*

What's the purpose of `editorStartPos>0` here?
</blockquote></details>

Hmm, I wanted to guard against ops that start with in insert. But now I think of it we don't allow editing the chapter or header text formatting in SF, so I'll remove that.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 703 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYaX9sBHYcYwNDmb4n:-MwgtrO14AHPKL0DC6H2:boj1qgt) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L703)):*

Don't we only want to subtract editorRange.leadingEmbedCount if a delete op was pushed a few lines above?
</blockquote></details>

Good catch. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 711 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYb8ar4w4b3T_iio5d:-MwguYASB0FNpnsbYHd8:beppdeo) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L711)):*

I'm a bit confused how the  placement of this condition, and its relation to prior lines of code, are intended to relate. 
(Would moving `if cloneOp.delete <1; continue` up higher be as good?) 
If cloneOp.delete===0, then getEditorContentRange() could possibly provide a count of initial sequential embeds, iirc, and so embedsUpToIndex and cloneOp.delete will be increased. Tho then cloneOp.delete won't equal 0 anymore to do the `continue` here. Is that what the thinking is?

</blockquote></details>

I think that is right. After checking that there are any embeds to delete, if we find them, we add them to the delete op. So then we are in position to check if the delete is 1 or greater, and push that delete op tot he adjustedDelta.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 714 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-MwYc4xU8npAtsjj_xJQ:-Mwgv8_56GNUPXJUQRPp:b-cjbtuv) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L714)):*

This previousOp setting will need to be done earlier since there is a `continue` above. Oh, unless you are meaning to kind of ignore the delete op if it's a meaningless 0-length delete?
</blockquote></details>

That's right, we want to ignore the cloneOp.delete == 0 situation.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 487 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mwc4B0f6AHXOkB3PMuD:-MwgvTp08UIzwCZfyvFF:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L487)):*

Hmm this should probably be  "at an editor position" ("an" not "the"), since multiple editor positions correspond to the same text position.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 549 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mwc4pOl49j_iz2SZQmd:-MwgvvpyDrfcb5njLikY:b-cpn5pu) ([raw file](https://github.com/sillsdev/web-xforge/blob/eb8a1c10ae30a0c277a1c6f30faa6157a54c98a9/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L549)):*

Hmm, why was there a -1 here? I don't understand, to know if maybe we still need it :)
</blockquote></details>

It looks like we use the - 1 to get the inclusive anchorEndPosition. I've updated it so we determine the format length using the EditorRange.editorLength. It is simpler and inherently includes the embed in the calculation.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 554 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mwc59gF0ixanzXNZNXR:-MwgvaG666QVOrElrV9C:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/8227614ae1c727ab6573fa48b00ac73afb4ba0c2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L554)):*

And we'll want to remove this line if we aren't doing `+1` below any more.
</blockquote></details>

Done.

marksvc (changes requested):
> Reviewed 1 of 5 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1206)* status: all files reviewed, 5 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1230 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mx5RDyIBetNJ6JAqffW:-Mx5RDyIBetNJ6JAqffX:b-wnwiwd) ([raw file](https://github.com/sillsdev/web-xforge/blob/b45c3bd10db43a93bc9bc186c8828aab8f300653/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1230)):*
> ```XML
> 
>       const note7Position = env.getNoteThreadEditorPosition('thread07');
>       const adjacentNoteLength = 1;
> ```

Hmm? What's with this?



_Code quote:_
```XML
 const adjacentNoteLength = 1;
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1780 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mx1UOI954WOo5d9nhN5:-Mx1UOI954WOo5d9nhN6:bboflqh) ([raw file](https://github.com/sillsdev/web-xforge/blob/b45c3bd10db43a93bc9bc186c8828aab8f300653/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1780)):*
> ```XML
> 
>       const noteThread1Doc: NoteThreadDoc = env.getNoteThreadDoc('project01', 'thread01');
>       const noteThread3Doc: NoteThreadDoc = env.getNoteThreadDoc('project01', 'thread04');
> ```

Oops. This should probably be `noteThread4Doc`.



_Code quote:_
```XML
 noteThread3Doc
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1782 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mx1UVUG2Li3v0FkmbDK:-Mx1UVUG2Li3v0FkmbDL:btscsa0) ([raw file](https://github.com/sillsdev/web-xforge/blob/b45c3bd10db43a93bc9bc186c8828aab8f300653/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1782)):*
> ```XML
>       const noteThread3Doc: NoteThreadDoc = env.getNoteThreadDoc('project01', 'thread04');
>       const originalNoteThread1TextPos: TextAnchor = noteThread1Doc.data!.position;
>       const originalNoteThread3TextPos: TextAnchor = noteThread3Doc.data!.position;
> ```

(And so also originalNoteThread4TextPos)



_Code quote:_
```XML
originalNoteThread3TextPos
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1809 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mx1WFjM6PJLbRBWrnwe:-Mx1WFjM6PJLbRBWrnwf:b9nwa3q) ([raw file](https://github.com/sillsdev/web-xforge/blob/b45c3bd10db43a93bc9bc186c8828aab8f300653/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1809)):*
> ```XML
>       expect(noteThread3Doc.data!.position).toEqual(originalNoteThread3TextPos);
> 
>       // simulate text changes at a different segment
> ```

I was surprised at what this one is asserting. What's the purpose of doing this "at a different segment" if we just check that the thread 1 is still where we expect it to be? (which seems to be the same as what was tested in SUT 1). I would have thought we would go on to check that thread 4 is in the expected place. 



_Code quote:_
```XML
 // simulate text changes at a different segment
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1835 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mx1XvWc1zxFD5RArgrB:-Mx1XvWc1zxFD5RArgrC:b-91cgw) ([raw file](https://github.com/sillsdev/web-xforge/blob/b45c3bd10db43a93bc9bc186c8828aab8f300653/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1835)):*
> ```XML
>       const note1Doc: NoteThreadDoc = env.getNoteThreadDoc('project01', 'thread01');
>       const anchor: TextAnchor = { start: 8 + insert.length, length: 12 };
>       note1Doc.submitJson0Op(op => op.set(nt => nt.position, anchor));
> ```

This is a position change to the thread doc  which would be coming thru to the client from the server, and so a remote change, right? (Tho not necessarily triggering the same remoteChange$ Subject)



_Code quote:_
```XML
      const note1Doc: NoteThreadDoc = env.getNoteThreadDoc('project01', 'thread01');
      const anchor: TextAnchor = { start: 8 + insert.length, length: 12 };
      note1Doc.submitJson0Op(op => op.set(nt => nt.position, anchor));
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1206)* status: all files reviewed, 5 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1230 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mx5RDyIBetNJ6JAqffW:-Mx5URdXBXEvETvxba1a:bjcmg9s) ([raw file](https://github.com/sillsdev/web-xforge/blob/b45c3bd10db43a93bc9bc186c8828aab8f300653/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1230)):*

Hmm? What's with this?
</blockquote></details>

I had to double check also. note thread 4 is also at this location, these notes are side-by-side. I've updated the comment to make it more obvious.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1780 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mx1UOI954WOo5d9nhN5:-Mx5VlCe5QJleDqnubAd:boj1qgt) ([raw file](https://github.com/sillsdev/web-xforge/blob/b45c3bd10db43a93bc9bc186c8828aab8f300653/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1780)):*

Oops. This should probably be `noteThread4Doc`.
</blockquote></details>

Good catch. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1782 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mx1UVUG2Li3v0FkmbDK:-Mx5VnXpELguTzlFVXSb:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/b45c3bd10db43a93bc9bc186c8828aab8f300653/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1782)):*

(And so also originalNoteThread4TextPos)
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1809 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mx1WFjM6PJLbRBWrnwe:-Mx5WEm600qT1DL0GBOW:b-ubrm5q) ([raw file](https://github.com/sillsdev/web-xforge/blob/b45c3bd10db43a93bc9bc186c8828aab8f300653/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1809)):*

I was surprised at what this one is asserting. What's the purpose of doing this "at a different segment" if we just check that the thread 1 is still where we expect it to be? (which seems to be the same as what was tested in SUT 1). I would have thought we would go on to check that thread 4 is in the expected place. 
</blockquote></details>

Good thinking. I've added a test to check that noteThread4 is still in the original position.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1835 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1206#-Mx1XvWc1zxFD5RArgrB:-Mx5WssG4pq_0Sb321Mq:b-m9bm6r) ([raw file](https://github.com/sillsdev/web-xforge/blob/b45c3bd10db43a93bc9bc186c8828aab8f300653/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1835)):*

This is a position change to the thread doc  which would be coming thru to the client from the server, and so a remote change, right? (Tho not necessarily triggering the same remoteChange$ Subject)
</blockquote></details>

That's right. The text anchor is updated, and the `noteThreadDoc.remoteChanges$` would emit. But we do not listen for that change since it is sufficient to listen for the `textDoc.remoteChange$`

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1206)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1208, opened 2021-12-13
SF-1456 Remove excessive space in import question dialog

Nateowami:
There are several possible messages that can be shown at the bottom of the import questions dialog. Originally they were styled `visibility: hidden` until shown so the layout would not change at all once they were revealed. However, this is not great on mobile where screen space is very limited. For that reason this PR changes from using `visibility: hidden` to using `*ngIf`. This required changes to the tests. The changes to the UI are shown in the screenshots below. It is assumed at least one of the questions has already been imported (otherwise this message wouldn't show at all).
 
Description | Before this change | After this change
------------|--------------------|------------------
Initial dialog screen |  | 
After selecting questions |  | 

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1208)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html
@@ -157,13 +157,13 @@
     </div>
 
     <div *ngIf="status === 'filter'" class="dialog-content-footer">
-      <mat-error [class.hidden]="!(importClicked && selectedCount < 1)">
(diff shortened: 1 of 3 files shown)


irahopkinson (approved):
When  I initially read your PR message I thought it might be better to have the original behavior on larger viewports and this new behavior on only mobile viewports. However having seen the screenshots, I think the simpler solution for all viewports is fine.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1208#-:-MqtW3ynBm61VOXBjKmO:b-gkgnn)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1211, opened 2021-12-17
SF-1430 - Note icon does not change into default flag icon in SF if note icon deleted in PT

nigel-wells:
- Check tag icon change during sync

PR addresses two points:
1. Setting the correct tag icon for a note which covers all the different statuses including resolving and reopening a thread
2. Ensuring the thread icon matches the correct tag icon (last used) in the list of notes

The end result will look similar to this:
image

I also changed the `Note` interface to set `content` as optional which is how the backend is treating it when saving to Mongo

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1211)

### DIFF
--- a/src/RealtimeServer/scriptureforge/models/note.ts
+++ b/src/RealtimeServer/scriptureforge/models/note.ts
@@ -3,7 +3,7 @@ import { NoteStatus } from './note-thread';
 
 export interface Note extends Comment {
   threadId: string;
-  content: string;
+  content?: string;
   extUserId: string;
   deleted: boolean;
   tagIcon?: string;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts
@@ -21,7 +21,7 @@ export class NoteThreadDoc extends ProjectDataDoc<NoteThread> {
   }
 
   get iconResolved(): NoteThreadIcon {
-    let iconTag = this.getResolvedTag(this.getTag());
+    const iconTag = this.getResolvedTag(this.getTag());
     return this.iconProperties(iconTag);
   }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts
@@ -21,11 +21,16 @@ export class NoteThreadDoc extends ProjectDataDoc<NoteThread> {
   }
 
   get iconResolved(): NoteThreadIcon {
-    let iconTag = this.getTag();
-    if (iconTag !== '') {
-      // Resolved tags use 5 in the filename instead of the current number suffix
-      iconTag = iconTag.slice(0, iconTag.length - 1) + '5';
-    }
+    let iconTag = this.getResolvedTag(this.getTag());
+    return this.iconProperties(iconTag);
+  }
+
+  getNoteIcon(note: Note): NoteThreadIcon {
+    return this.iconProperties(note.tagIcon ? note.tagIcon : '');
+  }
+
+  getNoteResolvedIcon(note: Note): NoteThreadIcon {
+    const iconTag = this.getResolvedTag(note.tagIcon ? note.tagIcon : '');
     return this.iconProperties(iconTag);
   }
 
@@ -43,7 +48,18 @@ export class NoteThreadDoc extends ProjectDataDoc<NoteThread> {
     return iconTag;
   }
 
+  private getResolvedTag(iconTag: string = ''): string {
+    if (iconTag !== '') {
+      // Resolved tags use 5 in the filename instead of the current number suffix
+      iconTag = iconTag.slice(0, iconTag.length - 1) + '5';
+    }
+    return iconTag;
+  }
+
   private iconProperties(iconTag: string): NoteThreadIcon {
+    if (iconTag === '') {
+      return { cssVar: '', url: '' };
+    }
     const iconUrl = `/assets/icons/TagIcons/${iconTag}.png`;
     return { cssVar: `--icon-file: url(${iconUrl});`, url: iconUrl };
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
@@ -976,7 +976,7 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
     const notes: Note[] = clone(threadDoc.data!.notes).sort(
       (a, b) => Date.parse(a.dateCreated) - Date.parse(b.dateCreated)
     );
-    let preview: string = this.stripXml(notes[0].content.trim());
+    let preview: string = notes[0].content != null ? this.stripXml(notes[0].content.trim()) : '';
     if (notes.length > 1) {
       preview += '\n' + translate('editor.more_notes', { count: notes.length - 1 });
     }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts
@@ -66,7 +66,7 @@ describe('NoteDialogComponent', () => {
 
   it('should style notes', fakeAsync(() => {
     env = new TestEnvironment();
-    const tests: { text: string; expected: string }[] = [
+    const tests: { text: string | undefined; expected: string }[] = [
       {
         text: 'turn <bold>text bold</bold>',
         expected: 'turn <b>text bold</b>'
@@ -82,6 +82,14 @@ describe('NoteDialogComponent', () => {
       {
         text: 'check <unknown id="anything">unknown</unknown> <italic>text</italic>',
         expected: 'check unknown <i>text</i>'
+      },
+      {
+        text: '',
+        expected: ''
+      },
+      {
+        text: undefined,
+        expected: ''
       }
     ];
     tests.forEach(test => {
@@ -98,24 +106,34 @@ describe('NoteDialogComponent', () => {
     env = new TestEnvironment();
 
     // To do
-    expect(env.notes[0].nativeElement.querySelector('img').getAttribute('src')).toEqual(
-      '/assets/icons/TagIcons/flag02.png'
-    );
-    expect(env.notes[0].nativeElement.querySelector('img').getAttribute('title')).toEqual('To do');
+    expect(env.notes[0].nativeElement.querySelector('img').getAttribute('src'))
+      .withContext('[n0] to do - src')
+      .toEqual('/assets/icons/TagIcons/flag02.png');
+    expect(env.notes[0].nativeElement.querySelector('img').getAttribute('title'))
+      .withContext('[n0] to do - title')
+      .toEqual('To do');
 
     // Resolved
-    expect(env.notes[1].nativeElement.querySelector('img').getAttribute('src')).toEqual(
-      '/assets/icons/TagIcons/flag05.png'
-    );
-    expect(env.notes[1].nativeElement.querySelector('img').getAttribute('title')).toEqual('Resolved');
-    expect(env.notes[3].nativeElement.querySelector('img').getAttribute('src')).toEqual(
-      '/assets/icons/TagIcons/flag05.png'
-    );
-    expect(env.notes[3].nativeElement.querySelector('img').getAttribute('title')).toEqual('Resolved');
+    expect(env.notes[1].nativeElement.querySelector('img').getAttribute('src'))
+      .withContext('[n1] resolved - src')
+      .toEqual('/assets/icons/TagIcons/flag05.png');
+    expect(env.notes[1].nativeElement.querySelector('img').getAttribute('title'))
+      .withContext('[n1] resolved - title')
+      .toEqual('Resolved');
+    expect(env.notes[3].nativeElement.querySelector('img').getAttribute('src'))
+      .withContext('[n3] resolved - src')
+      .toEqual('/assets/icons/TagIcons/flag05.png');
+    expect(env.notes[3].nativeElement.querySelector('img').getAttribute('title'))
+      .withContext('[n3] resolved - title')
+      .toEqual('Resolved');
 
     // Blank/unspecified
-    expect(env.notes[2].nativeElement.querySelector('img').getAttribute('src')).toEqual('');
-    expect(env.notes[2].nativeElement.querySelector('img').getAttribute('title')).toEqual('');
+    expect(env.notes[2].nativeElement.querySelector('img').getAttribute('src'))
+      .withContext('[n2] blank - src')
+      .toEqual('');
+    expect(env.notes[2].nativeElement.querySelector('img').getAttribute('title'))
+      .withContext('[n2] blank - title')
+      .toEqual('');
   }));
 
   it('should gracefully return when data not ready', fakeAsync(() => {
@@ -223,6 +241,7 @@ class TestEnvironment {
         deleted: false,
         ownerRef: 'user01',
         status: NoteStatus.Todo,
+        tagIcon: 'flag02',
         dateCreated: '',
         dateModified: ''
       },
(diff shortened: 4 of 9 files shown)


RaymondLuong3 (changes requested):
This looks good. I wonder if I am doing something different than you are. When I create comments in Paratext, I generally do not see a `status` for the comment. The only time I've seen a status appear is when a conflict note is created. Otherwise the normal status is empty. When do you see status defined in your notes?

> Reviewed 8 of 9 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-:-Mr8m8ZIAHGKjySK17ap:b1pxri)* status: all files reviewed, 6 unresolved discussions (waiting on @nigel-wells)

___
*[src/RealtimeServer/scriptureforge/models/note.ts, line 6 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8KqiV2V-IpOnEijSO:-Mr8KqiV2V-IpOnEijSP:bcze5fh) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/RealtimeServer/scriptureforge/models/note.ts#L6)):*
> ```XML
> export interface Note extends Comment {
>   threadId: string;
>   content?: string;
> ```

I had been thinking about saving this as an empty string to the mongoDB when the comment.Content is null, but I also see the advantage of treating this as nullable. Let's go with this to be consistent with Paratext comments.



_Code quote:_
```XML
  content?: string;
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts, line 24 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8K6Ef0KaGdUhIymVS:-Mr8K6Ef0KaGdUhIymVT:by00se5) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts#L24)):*
> ```XML
> 
>   get iconResolved(): NoteThreadIcon {
>     let iconTag = this.getResolvedTag(this.getTag());
> ```

This should be a const variable. Please add the type here and below.



_Code quote:_
```XML
    let iconTag = this.getResolvedTag(this.getTag());
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts, line 244 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8WCn90iq3bcCY9aAf:-Mr8WCn90iq3bcCY9aAg:bgx74st) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts#L244)):*
> ```XML
>         ownerRef: 'user01',
>         status: NoteStatus.Todo,
>         tagIcon: 'flag02',
> ```

From my experience, most Notes do not have tagIcon defined. Is there a reason to define the tagIcon for every note?



_Code quote:_
```XML
      tagIcon: 'flag02',
```
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 777 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8hryw-0LBl2S4S67w:-Mr8hryw-0LBl2S4S67x:b8shpd1) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/Services/ParatextService.cs#L777)):*
> ```Smalltalk
>                     {
>                         matchedCommentIds.Add(matchedComment.Id);
>                         CommentTag commentIconTag = this.GetCommentTag(existingThread, matchedComment, commentTags);
> ```

Generally speaking we do not need `this` before calling GetCommentTag(), so we can remove it to stick with the convention.



_Code quote:_
```Smalltalk
                        CommentTag commentIconTag = this.GetCommentTag(existingThread, matchedComment, commentTags);
```
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1579 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8cSJ0Ege3EF0bwSQ0:-Mr8cSJ0Ege3EF0bwSQ1:b40g1x2) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1579)):*
> ```Smalltalk
>         {
>             // Use the main to do tag as default
>             CommentTag tagInUse = commentTags.Get(1);
> ```

You can set this as commentTags.Get(CommentTag.toDoTagId);



_Code quote:_
```Smalltalk
            CommentTag tagInUse = commentTags.Get(1);
```
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1592 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8_igrBe3fKO37p8mn:-Mr8_igrBe3fKO37p8mo:b7qloji) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1592)):*
> ```Smalltalk
>                 // When checking the tag for the thread and a conflict is found then use that
>                 if (threadComment.Type == NoteType.Conflict && comment == null)
>                     return CommentTag.ConflictTag;
> ```

You can check if comment is null just once like this.
```
if (comment == null)
{
  if (threadComment.Type == NoteType.Conflict)
    return CommentTag.ConflictTag;
  continue;
}
```



_Code quote:_
```Smalltalk
                if (threadComment.Type == NoteType.Conflict && comment == null)
                    return CommentTag.ConflictTag;

                if (comment == null)
                    continue;
```
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1597 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8oJDX8O2vPSGbcIIo:-Mr8oJDY4DBcGMcOglBW:b-smy3bp) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1597)):*
> ```Smalltalk
>                     continue;
> 
>                 if (threadComment.Id == comment.Id)
> ```

What if you simply returned the first tag in the `threadComment.TagsAdded` list when the comment matches or null if it does not exist? It seems that property is used when the icon has been changed, otherwise it is assumed that the icon is the same as the previous comment. We could trim that logic out of this method.



_Code quote:_
```Smalltalk
                if (threadComment.Id == comment.Id)
```
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1604 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8f431BUipk7BeER25:-Mr8f431BUipk7BeER26:b-ormohr) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1604)):*
> ```Smalltalk
> 
>                     // Ignore any repeat tag icons i.e. only use a tag if it has changed
>                     if (lastTodoTagUsed != null && lastTodoTagUsed.Icon == tagInUse.Icon && (comment.Status == NoteStatus.Todo || tagAddedInUse))
> ```

When the tagAddedInUse is false and the note status is unspecified, we do not return null. I would have thought in this situation, we would want to return null (since in this situation, the tag would be a repeat of the previous comment)



_Code quote:_
```Smalltalk
                    if (lastTodoTagUsed != null && lastTodoTagUsed.Icon == tagInUse.Icon && (comment.Status == NoteStatus.Todo || tagAddedInUse))
```

nigel-wells (commented):
The status will change when you resolve a thread and then reopen it again. The tag can change from also doing either of those or by simply assigning a new tag. Resoling has the status as "deleted" and opening it has the status of "todo". That help?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-:-MrKZ9QS6oKE0UgN3sPy:banror0)* status: 7 of 9 files reviewed, 6 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts, line 24 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8K6Ef0KaGdUhIymVS:-MrK_5d-BDD1_wd-ZZOb:b-8clqav) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts#L24)):*

This should be a const variable. Please add the type here and below.
</blockquote></details>

Done. Not sure why my IDE didn't point that out - it did on my old machine so I'll need to look in to that.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts, line 244 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8WCn90iq3bcCY9aAf:-MrK_M-O-n1QGAvP-Lav:b-n6mvaa) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts#L244)):*

From my experience, most Notes do not have tagIcon defined. Is there a reason to define the tagIcon for every note?
</blockquote></details>

It's on all but one. I wanted to try and closely replicate the data that should be there after a sync had taken place rather than just putting in test data that doesn't reflect what the sync does. In this case I was testing the different statuses with a icon supplied.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 777 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8hryw-0LBl2S4S67w:-MrKvqud6dMk_lJ7jDg2:b7ap0tn) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/Services/ParatextService.cs#L777)):*

Generally speaking we do not need `this` before calling GetCommentTag(), so we can remove it to stick with the convention.
</blockquote></details>

Ah nice! I guess C# had to be different from the others :)
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1579 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8cSJ0Ege3EF0bwSQ0:-MrKw4cOCVwB2qmLNghF:b-j6u9bs) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1579)):*

You can set this as commentTags.Get(CommentTag.toDoTagId);
</blockquote></details>

Perfect
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1592 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8_igrBe3fKO37p8mn:-MrKwU3tEbc7ecE2V5Do:bxf8vn) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1592)):*
> if (comment == null)
> {
>   if (threadComment.Type == NoteType.Conflict)
>     return CommentTag.ConflictTag;
>   continue;
> }
Yep that works too - have updated
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1597 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8oJDX8O2vPSGbcIIo:-MrKwm8jCT0bVVXnq7xk:b-lg8or9) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1597)):*

What if you simply returned the first tag in the `threadComment.TagsAdded` list when the comment matches or null if it does not exist? It seems that property is used when the icon has been changed, otherwise it is assumed that the icon is the same as the previous comment. We could trim that logic out of this method.
</blockquote></details>

I found if a status is resolved and then reopened with the status "todo" it doesn't contain anything in `TagsAdded` hence the need to track what was being used. PT appeared to do all this logic when they displayed the notes. We possibly could have done the same but we were already recording the various icons so I continued with that and added the logic to the sync instead.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1604 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8f431BUipk7BeER25:-MrKxdnjCYB8Ft3rFHLW:bhxz497) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1604)):*

When the tagAddedInUse is false and the note status is unspecified, we do not return null. I would have thought in this situation, we would want to return null (since in this situation, the tag would be a repeat of the previous comment)
</blockquote></details>

That is taken care of further under this if statement - there is another check for status to ensure it isn't `Unspecified` and, if it is `Unspecified` then `null` is returned.

RaymondLuong3 (changes requested):
Ok, that would explain why I rarely see status in Comment assigned. Thanks for explaining. I just have one remaining change request, then it is good to go.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-:-MrNnuDcA2B0HLEA3TFW:b9ylic8)* status: all files reviewed, 3 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 777 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8hryw-0LBl2S4S67w:-MrNn6EV8KuRD3rgFwZ2:b-s3wy7t) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/Services/ParatextService.cs#L777)):*

Ah nice! I guess C# had to be different from the others :)
</blockquote></details>

Below are some more instances of unnecessary `this` references that can be fixed up.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1597 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8oJDX8O2vPSGbcIIo:-MrNnVd0-7i5bnpJM_nR:b-2osbbr) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1597)):*

I found if a status is resolved and then reopened with the status "todo" it doesn't contain anything in `TagsAdded` hence the need to track what was being used. PT appeared to do all this logic when they displayed the notes. We possibly could have done the same but we were already recording the various icons so I continued with that and added the logic to the sync instead.
</blockquote></details>

Ah good job discovering that. It is too bad that when a note is resolved and then reopened, it does not pick up on the relevant tag. Ok, this will have to do then.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1211)* status: 8 of 9 files reviewed, 3 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 777 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-Mr8hryw-0LBl2S4S67w:-MrOHfqN0oY2fiZPKiC0:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab9a88127b197b07ca3a5f8bbd4385ad7129febe/src/SIL.XForge.Scripture/Services/ParatextService.cs#L777)):*

Below are some more instances of unnecessary `this` references that can be fixed up.
</blockquote></details>

Done.

RaymondLuong3 (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1211#-:-MrOsmVt-7i7GI11PmLf:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

PR #1213, opened 2021-12-21
Improve tests by making CSV parsing mockable

Nateowami:
One of the tests for the question import dialog couldn't use `fakeAsync` because Jasmine doesn't redefine FileReader, so reading files can't be controlled using `tick()` and `flush()` (you can `tick()` and `flush()`, but the FileReader doesn't fire its event handlers).

What I've done in this PR is pull the file parsing out into a separate service so it can be more easily mocked in the tests. Now a File object isn't actually needed in the tests, and `tick()` and `flush()` work.

I'm making this change because I'm going to need to write another test, and I couldn't figure out how to get it to work. It should be a lot easier after this.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1213)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts
@@ -17,6 +17,7 @@ import { VerseRef } from 'realtime-server/lib/esm/scriptureforge/scripture-utils
 import { Observable, of, Subject } from 'rxjs';
 import { anything, capture, instance, mock, verify, when } from 'ts-mockito';
 import { AuthService } from 'xforge-common/auth.service';
+import { CsvService } from 'xforge-common/csv-service.service';
 import { RealtimeQuery } from 'xforge-common/models/realtime-query';
 import { PwaService } from 'xforge-common/pwa.service';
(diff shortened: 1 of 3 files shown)


RaymondLuong3 (approved):
Nice work. Just one non-blocking comment.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1213#-:-MrYktbXDHjUlXEbmTIQ:b-619ysz)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/csv-parser.service.ts, line 7 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1213#-MrYkJ9EEbqYCW2f3nC2:-MrYkJ9EEbqYCW2f3nC3:bxmgu8v) ([raw file](https://github.com/sillsdev/web-xforge/blob/12c3ffbf748d2431976437ce705ef671ac5416e2/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/csv-parser.service.ts#L7)):*
> ```XML
>   providedIn: 'root'
> })
> export class CsvParser {
> ```

I would name this CsvService, it would have the benefit of matching the convention of having service in its name, and maybe more methods can be added later pertaining to CSV that may not be parsing specific.



_Code quote:_
```XML
export class CsvParser {
```

PR #1218, opened 2021-12-23
SF-1455 Communicate editor focus change via events

Nateowami:
Avoids ExpressionChangedAfterItHasBeenCheckedError when double clicking a note icon.

After our call I also went through and updated the other locations in the editor component where we access `target.hasFocus` and swapped it out for the value ascertained from the events. I think this is desirable, as it means the entire component has one single variable it relies on for determining whether the editor has focus.

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.html
@@ -19,5 +19,7 @@
   [dir]="$any(textDirection)"
   [lang]="lang"
   [attr.data-browser-engine]="browserEngine"
+  (onBlur)="focused.next(false)"
+  (onFocus)="focused.next(true)"
 >
 </quill-editor>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -87,6 +87,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
   @Output() updated = new EventEmitter<TextUpdatedEvent>(true);
   @Output() segmentRefChange = new EventEmitter<string>();
   @Output() loaded = new EventEmitter(true);
+  @Output() focused = new EventEmitter<boolean>(true);
   lang: string = '';
   // only use USX formats and not default Quill formats
   readonly allowedFormats: string[] = USX_FORMATS;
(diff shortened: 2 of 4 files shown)


RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1218)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc and @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.html, line 22 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1218#-MrdBso89pg7XdIVEVxA:-MrdBso89pg7XdIVEVxB:b2at0aq) ([raw file](https://github.com/sillsdev/web-xforge/blob/30bdb34b3383e4ee851f6c9056905cf7d5b73c7e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.html#L22)):*
> ```HTML
>   [lang]="lang"
>   [attr.data-browser-engine]="browserEngine"
>   (onBlur)="focused.next(false)"
> ```

This eventEmitter should be using emit() to emit the values. I'm not sure why next() seems to work.



_Code quote:_
```HTML
  (onBlur)="focused.next(false)"
  (onFocus)="focused.next(true)"
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1218)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @marksvc and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.html, line 22 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1218#-MrdBso89pg7XdIVEVxA:-Mrdj0zH8S6qvyziwSYn:b-crstf9) ([raw file](https://github.com/sillsdev/web-xforge/blob/30bdb34b3383e4ee851f6c9056905cf7d5b73c7e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.html#L22)):*

This eventEmitter should be using emit() to emit the values. I'm not sure why next() seems to work.
</blockquote></details>

Good point. The reason it works is because it extends Subject. But that's not how EventEmitter is intended to be used, and I don't know what the consequences of using it that way would be. Done.

RaymondLuong3 (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1218#-:-Ms_zgaKEa32xqpoDjI1:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1220, opened 2022-01-05
Optimize importing of questions

Nateowami:
There are three commits here, the first by Damien, and the second by me.

First commit:
Creates a way to speed up import by skipping the local update until after the last question is imported.

Second commit:
Changes Edge required version to 79 because the IndexedDB getAll method is not supported below version 79. Since Edge is now Chromium based, I would imagine Edge 79 is fairly outdated (and likely broken for other reasons?). But we know below version 79 it will be broken, so it seems best to bump the requirement.

Third commit:
Fixes assertions in tests.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1220)


Nateowami:
I'm closing this PR for now because I think we are getting quite close to an ideal import speed without going down this route (which is kind of ugly, and surprisingly isn't as effective as some other changes, despite addressing what previously appeared to be the root problem).

PR #1221, opened 2022-01-06
Optimize checking overview page to speed up question imports

Nateowami:
This is a change to the checking overview page that improves question import speed. Looking at the changes may lead one to believe this is premature optimization, but the data consistently says otherwise.

This chart shows how long it takes to import questions. The vertical axis is milliseconds, and the horizontal axis shows how many questions have been added. So what this chart is showing is that when you import questions, each question takes longer (on average) to import than the previous question. This pull request doesn't fix this entirely, but it makes the slope much more gradual. The total width of the chart is 1533, corresponding to the total number of questions I had to import. The blue line doesn't reach all the way to the right side of the chart because it was progressing too slowly to import all 1533 questions, and there's enough data to draw conclusions.

In case the explanation above isn't clear enough, here's how I produced this chart:
- On master, I ran imported a number of questions and logged how long it took to import each one. Then I plotted those values, and that produced the blue (before) line. This shows that by the time about 1000 questions were imported, it was starting to take almost 500ms to import each question. At this point the import had been running for about four and a half minutes and the page reloaded because Auth0.
- I then ran the same test with this PR and produced the red line. This is much better.
- I then wondered how much the getters I optimized were now contributing to the overall time. To test this I replaced each of them with `return 0` and ran it again. This did speed things up a tad, but not by much. I think I would need to run this test several times to be sure there's a real effect. The point is, it doesn't appear it accounts for the bulk of the rise in the slope, indicating there's probably something else causing most of the slowdown.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1221)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.html
@@ -110,7 +110,7 @@
     fxLayout="row"
     fxLayoutAlign="space-around"
     class="reviewer-panels"
-    *ngIf="allQuestionsCount === '0'"
+    *ngIf="allQuestionsCount === 0"
     id="no-questions-label"
   >
     <p>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts
@@ -334,10 +334,10 @@ describe('CheckingOverviewComponent', () => {
       expect(read).toBe(2);
       expect(answered).toBe(1);
       // 1 of 7 questions of MAT is archived + 1 in LUK
-      expect(env.component.allQuestionsCount).toBe('7');
-      expect(env.component.myAnswerCount).toBe('1');
-      expect(env.component.myCommentCount).toBe('2');
-      expect(env.component.myLikeCount).toBe('3');
+      expect(env.component.allQuestionsCount).toBe(7);
+      expect(env.component.myAnswerCount).toBe(1);
+      expect(env.component.myCommentCount).toBe(2);
+      expect(env.component.myLikeCount).toBe(3);
     }));
 
     it('should calculate the right stats for project admin', fakeAsync(() => {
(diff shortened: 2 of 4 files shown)


RaymondLuong3 (approved):
Nice work. :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1221#-:-Mspj_ubDEMRFDnRolqQ:b-a5e3ql)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.ts, line 71 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1221#-Mspi69h9SE5VrAV-Upg:-Mspi69h9SE5VrAV-Uph:b-51re62) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca95431427838e39bd8a3b3c3bad5fd5fb72a368/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.ts#L71)):*
> ```XML
>   get myAnswerCount(): number {
>     let count: number = 0;
>     const canCreateQuestion = this.canCreateQuestion;
> ```

How does assigning these getters to a const make such a significant improvement? Maybe copying a value of the getter to a const reduces the number of times change detection occurs? But glad that this makes a difference.



_Code quote:_
```XML
    const canCreateQuestion = this.canCreateQuestion;
    const currentUserId = this.userService.currentUserId;
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1221)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.ts, line 71 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1221#-Mspi69h9SE5VrAV-Upg:-Msq5y6uB9kpaCq2G0co:b-bp7hhl) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca95431427838e39bd8a3b3c3bad5fd5fb72a368/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.ts#L71)):*

How does assigning these getters to a const make such a significant improvement? Maybe copying a value of the getter to a const reduces the number of times change detection occurs? But glad that this makes a difference.
</blockquote></details>

I think it's because accessing a variable is cheaper than calling a function that in turn calls other functions.

`canCreateQuestion` does null checks, string concatenation, and looking things up in an array.

`currentUserId` hits local storage.

The profiler also seemed to indicate that these methods where bottlenecks.

They're still not very efficient, but they're no longer the bottleneck.

PR #1223, opened 2022-01-11
SF-1454 Handle dropping in more nodes

marksvc:
- discernTarget() still returns the <usx-segment> that is dropped
into, but also the specific element inside the usx-segment that was
dropped into, such as a <display-text-anchor>. This allows text to be
dropped into a thread anchoring area.
- determineInsertionPosition() processes a tree of nodes to determine
the length from the beginning of a <usx-segment> thru to the position
of the drop as reported by the browser.
- New tests specify behaviour for dropping into a note thread
anchoring area as well as a location that is in the anchoring area of
two note threads.

===

Handling of `<usx-blank>` elements can be somewhat rolled into this, but I want to do that in followup after this initial change is reviewed.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1223)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
@@ -269,10 +269,10 @@ export class DragAndDrop {
     }
 
     const lengthOfAnythingElse = 1;
-    const elementsKnownToBeLengthOne: string[] = ['display-note', 'usx-figure'];
+    const elementsKnownToBeLengthOne: string[] = ['display-note', 'usx-figure', 'usx-note'];
     if (!elementsKnownToBeLengthOne.includes(node.nodeName.toLowerCase())) {
       // Also return 1 for anything else. But note it in the console until it too is incorporated.
-      console.warn(`Warning: Drag-and-drop is assuming length 1 for node ${node.nodeName}`);
+      console.warn(`Warning: Drag-and-drop is assuming length 1 for node ${node.nodeName.toLowerCase()}`);
     }
     return lengthOfAnythingElse;
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
@@ -255,7 +255,7 @@ export class DragAndDrop {
       // Text nodes will not have children to further process.
     }
 
-    const elementsToLookInside: string[] = ['usx-segment', 'display-text-anchor'];
+    const elementsToLookInside: string[] = ['usx-segment', 'display-text-anchor', 'usx-char'];
     if (elementsToLookInside.includes(node.nodeName.toLowerCase())) {
       let lengthOfDescendants: number = 0;
       for (let i = 0; i < node.childNodes.length; i++) {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
@@ -25,8 +25,12 @@ export class DragAndDrop {
       // Stop the browser from doing any drag-and-drop behaviour itself, such as inserting text with formatting.
       dragEvent.preventDefault();
 
-      const [targetElement, droppingIntoBlankSegment]: [Element | null, boolean] = this.discernTarget(dragEvent);
-      if (targetElement == null) {
+      const [targetUsxSegmentElement, targetElement, droppingIntoBlankSegment]: [
+        Element | null,
+        Element | null,
+        boolean
+      ] = this.discernTarget(dragEvent);
+      if (targetUsxSegmentElement == null || targetElement == null) {
         return;
       }
 
@@ -38,9 +42,12 @@ export class DragAndDrop {
       // Determine where we should be placing the dropped text, using the location of the destination segment, and the
       // index within that segment.
 
-      const dataSegmentAttribute: any = targetElement.attributes['data-segment'];
+      const dataSegmentAttribute: any = targetUsxSegmentElement.attributes['data-segment'];
       if (dataSegmentAttribute == null) {
-        console.warn('Warning: drag-and-drop to target with no segment ref attribute:', targetElement);
+        console.warn(
+          'Warning: drag-and-drop to usx segment element with no segment ref attribute:',
+          targetUsxSegmentElement
+        );
         return;
       }
       const destinationSegmentRef: string = dataSegmentAttribute.value;
@@ -55,6 +62,7 @@ export class DragAndDrop {
       }
 
       const insPos: number | undefined = this.determineInsertionPosition(
+        targetUsxSegmentElement,
         targetElement,
         destinationSegmentRange,
         droppingIntoBlankSegment,
@@ -118,39 +126,58 @@ export class DragAndDrop {
     });
   }
 
-  /** Return the usx-segment element that the user is trying to drop into, if possible, and whether the original
-   * target was a usx-blank. */
-  private discernTarget(dragEvent: DragEvent): [Element | null, boolean] {
-    let targetElement: Element | null = dragEvent.target as Element | null;
+  /** Return the usx-segment element that the user is trying to drop into (if possible), the actual target element
+   * being dropped into, and whether the original target was a usx-blank. The actual element being dropped into may be
+   * something other than a usx-segment, such as a display-text-anchor, in which case the usx segment element returned
+   * will be the one containing the display-text-anchor being dropped into. */
+  private discernTarget(dragEvent: DragEvent): [Element | null, Element | null, boolean] {
+    const targetElement: Element | null = dragEvent.target as Element | null;
     let droppingIntoBlankSegment: boolean = false;
+    let targetUsxSegmentElement: Element | null = null;
 
     if (targetElement == null) {
       console.warn(`Warning: DragEvent unexpectedly has null target.`);
-      return [targetElement, droppingIntoBlankSegment];
+      return [targetUsxSegmentElement, targetElement, droppingIntoBlankSegment];
     }
     let targetElementName: string = targetElement.localName;
 
-    if (targetElementName === 'usx-blank') {
-      droppingIntoBlankSegment = true;
-      targetElement = targetElement.parentElement as Element | null;
-      if (targetElement == null) {
-        console.warn(`Warning: DragEvent usx-blank target parent is unexpectedly null. Target:`, targetElement);
-        return [targetElement, droppingIntoBlankSegment];
-      }
-      targetElementName = targetElement.localName;
+    if (targetElementName === 'usx-segment') {
+      targetUsxSegmentElement = targetElement;
+      return [targetUsxSegmentElement, targetElement, droppingIntoBlankSegment];
     }
 
-    if (targetElementName !== 'usx-segment') {
-      // We need to be able to know where to insert the dropped text, such as from the drop target being a usx-segment
-      // element. Give up.
-      console.warn('Warning: DragEvent to invalid target:', targetElement);
-      return [targetElement, droppingIntoBlankSegment];
+    if (targetElementName === 'usx-blank') {
+      droppingIntoBlankSegment = true;
+      targetUsxSegmentElement = targetElement.parentElement as Element | null;
+      if (targetUsxSegmentElement == null) {
+        console.warn(`Warning: DragEvent usx-blank target parent is unexpectedly null. Target:`, targetElement);
+        return [null, targetElement, droppingIntoBlankSegment];
+      }
+      if (targetUsxSegmentElement.localName === 'usx-segment') {
+        return [targetUsxSegmentElement, targetElement, droppingIntoBlankSegment];
+      } else {
+        console.warn(
+          'Warning: DragEvent usx-blank target parent is unexpectedly not a usx-segment. Target:`, targetElement'
+        );
+        return [null, targetElement, droppingIntoBlankSegment];
+      }
     }
-    return [targetElement, droppingIntoBlankSegment];
+
+    let element: Element = targetElement;
+    while (element.localName != 'usx-segment' && element.parentElement != null) {
+      element = element.parentElement;
+    }
+    if (element == null || element.localName != 'usx-segment') {
+      console.warn('Warning: DragEvent never found a needed usx-segment ancestor for drop target:', targetElement);
+      return [null, targetElement, droppingIntoBlankSegment];
+    }
+    targetUsxSegmentElement = element;
+    return [targetUsxSegmentElement, targetElement, droppingIntoBlankSegment];
   }
 
   /** Return the position in the quill editor where we want to insert text from a drop. */
   private determineInsertionPosition(
+    targetUsxSegmentElement: Element,
     targetElement: Element,
     destinationSegmentRange: RangeStatic,
     droppingIntoBlankSegment: boolean,
@@ -192,34 +219,61 @@ export class DragAndDrop {
         console.warn('Warning: Could not get the node that the text was dropped into.');
         return;
       }
-      let node: Node = nodeDroppedInto;
-      const textNodeType = 3;
-      // Add up the length of text and embeds that are previous nodes in the usx-segment
-      while (node.previousSibling != null) {
-        node = node.previousSibling;
-        if (node.nodeType === textNodeType) {
-          if (node.nodeValue != null) {
-            startPositionInSegment += node.nodeValue.length;
-          } else {
-            console.warn(`Warning: Unexpected situation of null text node value`);
-          }
-        } else {
-          if (node.nodeName.toLowerCase() === 'display-text-anchor') {
-            const anchoredTextOfNote: string = node.lastChild?.nodeValue ?? '';
-            startPositionInSegment += anchoredTextOfNote.length;
-            const lengthOfEmbed = 1;
-            startPositionInSegment += lengthOfEmbed;
-          } else {
-            console.warn(
-              `Warning: drag-and-drop is assuming length 1 for unknown element: ${node.nodeName.toLowerCase()}`
-            );
-            startPositionInSegment++;
-          }
-        }
-      }
-      startPositionInSegment += startPositionInTextNode;
+      startPositionInSegment = this.determineEditorLengthOfNodesUpToPosition(
+        targetUsxSegmentElement,
+        nodeDroppedInto,
+        startPositionInTextNode
+      );
     }
     const insertionPositionInDocument: number = destinationSegmentRange.index + startPositionInSegment;
     return insertionPositionInDocument;
+  }
+
+  private determineEditorLengthOfNodesUpToPosition(
+    targetUsxSegmentElement: Element,
+    nodeDroppedInto: Node,
+    positionInTextNode: number
+  ): number {
+    return positionInTextNode + this.determineLengthBeforeTerminationNode(targetUsxSegmentElement, nodeDroppedInto);
+  }
+
+  /**
+   * Determine the editor length for `node` up until a contained `terminationChildNode`. Length is
+   * determined from text nodes and some elements.
+   */
+  private determineLengthBeforeTerminationNode(node: Node, terminationChildNode: Node): number {
+    // Using preorder traversal and stops traversing when the terminationNode is reached.
+    if (node === terminationChildNode) {
+      return 0;
+    }
+    if (node.nodeType === Node.TEXT_NODE) {
+      if (node.nodeValue == null) {
+        // Based on the API docs, text nodes should not have a null value. So this should never happen.
+        return 0;
+      }
+      return node.nodeValue.length;
+      // Text nodes will not have children to further process.
+    }
+
+    const elementsToLookInside: string[] = ['usx-segment', 'display-text-anchor'];
+    if (elementsToLookInside.includes(node.nodeName.toLowerCase())) {
+      let lengthOfDescendants: number = 0;
+      for (let i = 0; i < node.childNodes.length; i++) {
+        let childNode: Node = node.childNodes[i];
+        lengthOfDescendants += this.determineLengthBeforeTerminationNode(childNode, terminationChildNode);
+        if (childNode.contains(terminationChildNode)) {
+          break;
+        }
+      }
+      return lengthOfDescendants;
+    }
+
+    const lengthOfAnythingElse = 1;
+    const elementsKnownToBeLengthOne: string[] = ['display-note', 'usx-figure'];
+    if (!elementsKnownToBeLengthOne.includes(node.nodeName.toLowerCase())) {
+      // Also return 1 for anything else. But note it in the console until it too is incorporated.
+      console.warn(`Warning: Drag-and-drop is assuming length 1 for node ${node.nodeName}`);
+    }
+    return lengthOfAnythingElse;
   }
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
@@ -8,14 +8,14 @@ export interface DragAndDropOptions {
 /** Drag-and-drop module for Quill, for dropping only unformatted text into a usx-segment. Unit tests on the behaviour
  *  specification can be found in text.component.spec.ts.*/
 export class DragAndDrop {
-  /** Mostly arbitrary value to indicate that our quill is the source of the drag data. */
+  /** Mostly arbitrary value to indicate that our quill is the origin of the drag data. */
   static readonly quillIsSourceToken: string = 'x-dragstart-in-quill';
 
   constructor(quill: Quill, options: DragAndDropOptions) {
     quill.container.addEventListener('dragstart', (event: Event) => {
       const dragEvent = event as DragEvent;
       // Write a custom note on the event. This will only end up being done if the drag started from in quill. In this
-      // way, we can differentiate between drag-and-drops that have their source as quill or from elsewhere, such as
+      // way, we can differentiate between drag-and-drops that have their origin as quill or from elsewhere, such as
       // elsewhere in the same web page window or from another window or application.
       dragEvent.dataTransfer?.setData(DragAndDrop.quillIsSourceToken, '');
     });
@@ -88,7 +88,7 @@ export class DragAndDrop {
       if (originalSelection != null) {
         // There was a selection before the drop occurred.
         if (quillIsSource && !userIsHoldingCtrlKey) {
-          // If the drag was started from within quill, then treat the selection as the source data of the drag, and
+          // If the drag was started from within quill, then treat the selection as the origin data of the drag, and
           // delete the selection. Unless the user was holding the ctrl key to copy text instead of move it.
           quill.deleteText(originalSelection.index, originalSelection.length, 'user');
           // Adjust insertion position accordingly if preceding text was just deleted
@@ -100,9 +100,9 @@ export class DragAndDrop {
         // the selection.
       }
 
-      // Behave like a user, to help editor event processing, by cutting the source text first (if applicable),
+      // Behave like a user, to help editor event processing, by cutting the origin text first (if applicable),
       // putting the insertion point in at the target, and then inserting. Give an opportunity between each of these
-      // for EditorComponent.onTargetUpdated() to run, so that thread anchors are updated even in a source verse that
+      // for EditorComponent.onTargetUpdated() to run, so that thread anchors are updated even in an origin verse that
       // is dragged out of. Use of setTimeout() allows TextComponent.updated to emit and be received by subscribers
       // before we take further action here. Therefore, TextComponent.updated events are interleaved with our calls to
       // quill.
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
@@ -181,6 +181,7 @@ export function registerScripture(): string[] {
   }
   formats.push(EmptyEmbed);
 
+  /** Span of characters or elements, that can have formatting. */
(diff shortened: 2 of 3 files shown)


RaymondLuong3 (changes requested):
Just beginning to review now. I'll add comments as they arise.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-:-MtAKS9g2k7EyF_bn-nX:b-jeocj8)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @marksvc and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 983 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtAJ0gq-cyyoS8wQaWV:-MtAJ0gq-cyyoS8wQaWW:bo6k187) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L983)):*
> ```XML
> 
>     it('can drag-and-drop into note thread anchoring area', fakeAsync(() => {
>       // The user drags into the region of the editor that is anchored to by a thread's text anchor. The dragged text
> ```

Doesn't the user drag into 



_Code quote:_
```XML
      // The user drags into the region of the editor that is anchored to by a thread's text anchor. The dragged text
      // should appear where it's dropped.
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 994 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtAJqqm3W3pP_5-auZ5:-MtAJqqm3W3pP_5-auZ6:b-xjcbb4) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L994)):*
> ```XML
>       const textLeadingUpToSelection_1_1 = 'target: ';
>       const initialTextInDoc_1_1 = 'target: chapter 1, verse 1.';
>       //                          move this ---------   ^^^^ thread anchoring
> ```

This isn't correct. The underlined text should just be on 'chapter'. What do the ^^^^ mean? 



_Code quote:_
```XML
      //                          move this ---------   ^^^^ thread anchoring
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 999 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtAI35-5nyMY9KraBMR:-MtAI35-5nyMY9KraBMS:blu5il5) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L999)):*
> ```XML
>       const expectedTextInDoc_1_1 = 'target: , verchapter 1se 1.';
>       const textLeadingUpToTargetLocationAfterEvent_1_1 = 'target: , ver';
>       const textNodeTextBeforeEvent_1_1 = 'erse';
> ```

From the name of the variable, I would have expected this to be ', verse 1.'



_Code quote:_
```XML
const textNodeTextBeforeEvent_1_1 = 'erse';
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1003 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtAILF171rVmS2K3eMb:-MtAILF171rVmS2K3eMc:bgg6rtx) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1003)):*
> ```XML
>       // The drop location is in a text node in a display-text-anchor element. These are the characters in the text
>       // node that lead up to the drop location.
>       const textNodeTextLeadingUpToTargetLocationAfterEvent_1_1 = 'er';
> ```

Same with this, isn't the 'chapter 1' text moved to be between ', ver' and 'se 1.' So I would have expected this to be ', ver'.



_Code quote:_
```XML
      const textNodeTextLeadingUpToTargetLocationAfterEvent_1_1 = 'er';
```

RaymondLuong3 (changes requested):
Good work figuring out how to do this. What about testing whether dragging text from within a note anchor? Are note threads correctly updated? I assume this already has tests in editor.component but just wanted to be sure you had thought of that.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-:-MtBCNCV32vbq3z87hxb:b-d51tql)* status: all files reviewed, 6 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 171 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtBE3nQEKbWYAVs2Wxs:-MtBE3nQEKbWYAVs2Wxt:b-9obsib) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L171)):*
> ```XML
>     }
>     if (element == null || element.localName != 'usx-segment') {
>       console.warn('Warning: DragEvent never found a needed usx-segment ancestor for drop target:', targetElement);
> ```

I wonder if it would be worth using a loop to discover if any parent is a usx-segment. It could be useful if we ever add more elements within a usx-segment.



_Code quote:_
```XML
      console.warn('Warning: DragEvent never found a needed usx-segment ancestor for drop target:', targetElement);
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 249 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtBFMxg6c_4xiPDbZNb:-MtBFMxg6c_4xiPDbZNc:bftjqje) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L249)):*
> ```XML
>       return 0;
>     }
>     if (node.nodeName.toLowerCase() === 'display-note') {
> ```

There can also be nodes such as usx-note or usx-figure that are embed just like the display-note. I believe those should be addressed here.



_Code quote:_
```XML
    if (node.nodeName.toLowerCase() === 'display-note') {
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 255 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtBFruoBiQ5lO22yKdV:-MtBFruoBiQ5lO22yKdW:b-b461de) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L255)):*
> ```XML
>     if (node.nodeType === Node.TEXT_NODE) {
>       if (node.nodeValue == null) {
>         throw new Error('Did not expect a text node to have a null value.');
> ```

Does this need to be an error? I guess it would prevent the user from possibly doing something that results in corrupted data.



_Code quote:_
```XML
        throw new Error('Did not expect a text node to have a null value.');
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 983 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtAJ0gq-cyyoS8wQaWV:-MtB2XK748QUNwZI_UE4:btslkkp) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L983)):*

Doesn't the user drag into 
</blockquote></details>

Nvm, it doesn't look like it is relevant.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 994 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtAJqqm3W3pP_5-auZ5:-MtB2pGx41jyYsMYcGE_:b-7m1p2s) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L994)):*

This isn't correct. The underlined text should just be on 'chapter'. What do the ^^^^ mean? 
</blockquote></details>

I see now that ------- denotes the text selection and the ^^^^ denotes the text anchor. I would have swapped those since a note anchor is represented by dashed lines. Ok good I'm starting to understand this.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 999 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtAI35-5nyMY9KraBMR:-MtBBemyB_zjUNSt_lik:bn7de1y) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L999)):*

From the name of the variable, I would have expected this to be ', verse 1.'
</blockquote></details>

Nvm, I see I had it swapped in my head what --- and ^^^ means
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1003 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtAILF171rVmS2K3eMb:-MtBBrQi7mzwTMkCAepe:b-uo3i1k) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1003)):*

Same with this, isn't the 'chapter 1' text moved to be between ', ver' and 'se 1.' So I would have expected this to be ', ver'.
</blockquote></details>

Nvm, I had it swapped around --- vs ^^^
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1146 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtB6b3E9LGkqOf255Ug:-MtB6b3F7ZahmKg67rAk:bi7seps) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1146)):*
> ```XML
>       // <display-text-anchor> elements for the area of overlapping anchoring. The text node that the drop happens in
>       // will be the thread2 text node.
>       const textNodeTextLeadingUpToTargetLocationAfterEvent_1_1 = 'r';
> ```

In this case, since the text anchor overlaps beginning at this character, it is a distinct text node from the letter 'e' that comes right before. Ok.



_Code quote:_
```XML
      const textNodeTextLeadingUpToTargetLocationAfterEvent_1_1 = 'r';
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1528 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtBC4cA8PmiHCrXIC5G:-MtBC4cA8PmiHCrXIC5H:bb5fmhu) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1528)):*
> ```XML
> 
>   /** Where reference is like 'MAT 1:2'. */
>   embedThreadAt(reference: string, textAnchor: TextAnchor): void {
> ```

This could be private



_Code quote:_
```XML
  embedThreadAt(reference: string, textAnchor: TextAnchor): void {
```

marksvc (commented):
Thank you. I wrote to you in chat about what to do regarding dragging from a note thread anchoring area, and the existing tests for thread anchoring.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-:-MtdoTl_1XkO4gTKF9Dx:b6b7d0v)* status: all files reviewed, 4 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 171 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtBE3nQEKbWYAVs2Wxs:-MtEcibk-ibtVXtpeytW:bjypqnt) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L171)):*

I wonder if it would be worth using a loop to discover if any parent is a usx-segment. It could be useful if we ever add more elements within a usx-segment.
</blockquote></details>

Yes! So we do this a few lines above when we do 
```ts
    while (element.localName != 'usx-segment' && element.parentElement != null) {
      element = element.parentElement;
    }
```
which will walk up the tree until we find the containing usx-segment element, or until we get to the root and never found a usx-segment. 
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 249 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtBFMxg6c_4xiPDbZNb:-MtdnkUB5KdfR98UWKvO:bnr310d) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L249)):*

There can also be nodes such as usx-note or usx-figure that are embed just like the display-note. I believe those should be addressed here.
</blockquote></details>

I'll mention that a couple ways to approach this would be:

- if we recognize an element of fixed length, return its known length and don't descend into its children. 
- otherwise, descend into any other element, looking for children elements and text nodes.

or

- If we recognize an element as having children we want to measure, descend into the element and examine its children.
- otherwise, assume the element is length 1 and don't process it further.

I suspect that in our case, the second one is more likely to do the right thing when we encounter an element we didn't plan for. I am posting a revision of the code which does this.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 255 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtBFruoBiQ5lO22yKdV:-MtEdl7y5NvbtCg_-5YE:b7qe5wz) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L255)):*

Does this need to be an error? I guess it would prevent the user from possibly doing something that results in corrupted data.
</blockquote></details>

This situation should never happen. Well, based on looking at https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeValue
As an alternative, we could detect and return: `if nodeValue is null then return 0`. But it shouldn't happen.
We could write a new class to represent a text node that that specifies that its nodeValue is always a non-null string.

The error message is perhaps not worded well since it's not clear if it means "It is possible that the text node could have a null value, but we didn't expect that it was going to happen in this situation" or "The programmer thought this situation was impossible and did not expect it to happen!". 

Good point tho that it doesn't need to be an error. I ended up changing to return 0 if nodeValue is null here, with a comment about the situation.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 994 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtAJqqm3W3pP_5-auZ5:-MtEigqA539PyV6ywAdq:bmdmld5) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L994)):*

I see now that ------- denotes the text selection and the ^^^^ denotes the text anchor. I would have swapped those since a note anchor is represented by dashed lines. Ok good I'm starting to understand this.
</blockquote></details>

Sorry about the misunderstandings! I don't mean for the characters here to be mimicking our UI, but just pointing to things to show what's happening where, and where some boundaries and pieces are. I changed some dashes to the caret so it's all just pointing rather than underlying. I see why that was confusing. 

I meant for the text next to the carets (or dashes) to describe what was being marked. For example, the 'chapter 1' section of text that is marked has a label "move this", and the 'erse' section of text that is marked has a label "thread anchoring". But I guess it wasn't without some getting used to. 

Ok I also put in horizontal lines from the labels, to the arrows they relate to. 
Well, and in some further work I ended up removing the labels since it was getting too complex.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1146 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtB6b3E9LGkqOf255Ug:-MtEl8ox5NtqtUu1Y7hP:b-4ws3g8) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1146)):*

In this case, since the text anchor overlaps beginning at this character, it is a distinct text node from the letter 'e' that comes right before. Ok.
</blockquote></details>

Yes. There is a text node for `e` and another text node for `rse 1`. Between the text nodes is a `display-note` element showing the flag icon. The drop occurs just after the `r` in the `rse 1` text node. And so after the drag-and-drop event(s), the text in the text node where the drop occurred, has text of `r` leading up to the location of the drop.
This variable is later on used to specify what the web browser will return for the position on the text node at which the drop occurred. Hm I guess it should be "BeforeEvent" not "AfterEvent". I'll change that. 
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1528 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtBC4cA8PmiHCrXIC5G:-MtEpg7B-k1KsAk90Nv2:bfhaoox) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1528)):*

This could be private
</blockquote></details>

Thanks. I just gave that a try but it looks like a few places call `env.embedThreadAt()` from outside the class.

RaymondLuong3 (approved):
Looks like it needs a rebase, then it looks good to me.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-:-MtzEo6A-RDfrJ660gof:b-hk79oh)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 171 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtBE3nQEKbWYAVs2Wxs:-MtzCVJfEQYH9VLz7Vq4:byzvq5z) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L171)):*

Yes! So we do this a few lines above when we do 
```ts
    while (element.localName != 'usx-segment' && element.parentElement != null) {
      element = element.parentElement;
    }
```
which will walk up the tree until we find the containing usx-segment element, or until we get to the root and never found a usx-segment. 
</blockquote></details>

Good, glad this is here for future-proofing.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 249 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtBFMxg6c_4xiPDbZNb:-MtzCiBP4hMdShv-qY9m:b-6xawy6) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L249)):*

I'll mention that a couple ways to approach this would be:

- if we recognize an element of fixed length, return its known length and don't descend into its children. 
- otherwise, descend into any other element, looking for children elements and text nodes.

or

- If we recognize an element as having children we want to measure, descend into the element and examine its children.
- otherwise, assume the element is length 1 and don't process it further.

I suspect that in our case, the second one is more likely to do the right thing when we encounter an element we didn't plan for. I am posting a revision of the code which does this.



</blockquote></details>

Yes, I believe the second is the case we should be prepared for. I haven't looked too closely at this, but from looking at these changes I get the understanding that most non-text elements have a length of 1, except for `usx-char` elements which have a length based on the enclosed text.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 255 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtBFruoBiQ5lO22yKdV:-MtzD9bT-e_zN7gfy_BE:bl34lks) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L255)):*

This situation should never happen. Well, based on looking at https://developer.mozilla.org/en-US/docs/Web/API/Node/nodeValue
As an alternative, we could detect and return: `if nodeValue is null then return 0`. But it shouldn't happen.
We could write a new class to represent a text node that that specifies that its nodeValue is always a non-null string.

The error message is perhaps not worded well since it's not clear if it means "It is possible that the text node could have a null value, but we didn't expect that it was going to happen in this situation" or "The programmer thought this situation was impossible and did not expect it to happen!". 

Good point tho that it doesn't need to be an error. I ended up changing to return 0 if nodeValue is null here, with a comment about the situation.


</blockquote></details>

Ok, good.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 994 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtAJqqm3W3pP_5-auZ5:-MtzDBp_4dx24AY0YxSR:b-96514h) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L994)):*

Sorry about the misunderstandings! I don't mean for the characters here to be mimicking our UI, but just pointing to things to show what's happening where, and where some boundaries and pieces are. I changed some dashes to the caret so it's all just pointing rather than underlying. I see why that was confusing. 

I meant for the text next to the carets (or dashes) to describe what was being marked. For example, the 'chapter 1' section of text that is marked has a label "move this", and the 'erse' section of text that is marked has a label "thread anchoring". But I guess it wasn't without some getting used to. 

Ok I also put in horizontal lines from the labels, to the arrows they relate to. 
Well, and in some further work I ended up removing the labels since it was getting too complex.
</blockquote></details>

Thanks, I think this helps to clarify things.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1528 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtBC4cA8PmiHCrXIC5G:-MtzCI6bBk-tL3MD7JeM:b-o4tfnc) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a19e071f7384ff935485683505195ae5bc02172/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1528)):*

Thanks. I just gave that a try but it looks like a few places call `env.embedThreadAt()` from outside the class.
</blockquote></details>

Good job testing that. Ok.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1597 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-Mtz9ibU1cVkweTFTjpU:-Mtz9ibU1cVkweTFTjpV:b-qqrpun) ([raw file](https://github.com/sillsdev/web-xforge/blob/95fb5fc96fe6cc45bedde4256758f32c5c51a141/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1597)):*
> ```XML
>       const expectedElementCountNugget = initialNuggetElementCount;
> 
>       const sourceSegmentRange: RangeStatic | undefined = env.component.getSegmentRange(sourceSegmentRef);
> ```

Consider renaming this so that someone else won't mistake this to be the source text, as in the mother translation segment range. Maybe dragStartSegmentRange.



_Code quote:_
```XML
 const sourceSegmentRange
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1662 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtzB50q8oa6VWsY0YdL:-MtzB50q8oa6VWsY0YdM:b2f38r8) ([raw file](https://github.com/sillsdev/web-xforge/blob/95fb5fc96fe6cc45bedde4256758f32c5c51a141/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1662)):*
> ```XML
>         textLeadingUpToTargetLocationAfterEvent.length +
>         args.nuggetEditorLength -
>         args.nuggetTextDocText!.length;
> ```

Why is this subtracted from the desiredSelectionStart? I see that when we assign textLeadingUpToTargetLocationAfterEvent, we use the length of thenuggetTextDocText. Is this trying to say that the nuggetTextDocText length is somehow excluded when we call getSelection()?



_Code quote:_
```XML
       args.nuggetTextDocText!.length;
```

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1223)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1597 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-Mtz9ibU1cVkweTFTjpU:-MuCrPBz8BW3scSpDWyd:b20g0fx) ([raw file](https://github.com/sillsdev/web-xforge/blob/95fb5fc96fe6cc45bedde4256758f32c5c51a141/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1597)):*

Consider renaming this so that someone else won't mistake this to be the source text, as in the mother translation segment range. Maybe dragStartSegmentRange.
</blockquote></details>

How about "origin"?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1662 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtzB50q8oa6VWsY0YdL:-MuCnyQB-8_mt4WdR56r:b-aub6w8) ([raw file](https://github.com/sillsdev/web-xforge/blob/95fb5fc96fe6cc45bedde4256758f32c5c51a141/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1662)):*

Why is this subtracted from the desiredSelectionStart? I see that when we assign textLeadingUpToTargetLocationAfterEvent, we use the length of thenuggetTextDocText. Is this trying to say that the nuggetTextDocText length is somehow excluded when we call getSelection()?
</blockquote></details>

No,
The selection start needs to be in terms of editor position. And so it needs to include editor-only items that may be present in `args.nuggetEditorLength` but not in `args.nuggetTextDocText!.length`, like a footnote marker. 

I am revising it to try to clarify what is happening, by introducing variable `editorLengthToTargetLocationAfterEvent`.

Another way to clarify may have been to write:
```ts
const lengthOnlyInEditor: number = args.nuggetEditorLength - args.nuggettextDocText.length
const desiredSelectionstart = targetSegmentRange.index + textLeadingUpToTargetLocationAfterEvent.length + lengthOnlyInEditor
```

RaymondLuong3 (approved):
> Reviewed 1 of 3 files at r3, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1223)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1597 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-Mtz9ibU1cVkweTFTjpU:-MuCyq2u5Wf8kyeZQfMP:b-speyhe) ([raw file](https://github.com/sillsdev/web-xforge/blob/95fb5fc96fe6cc45bedde4256758f32c5c51a141/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1597)):*

How about "origin"?
</blockquote></details>

I like origin!
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1662 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1223#-MtzB50q8oa6VWsY0YdL:-MuCzBZoEj_w-VJAyVNH:b-nbajas) ([raw file](https://github.com/sillsdev/web-xforge/blob/95fb5fc96fe6cc45bedde4256758f32c5c51a141/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1662)):*

No,
The selection start needs to be in terms of editor position. And so it needs to include editor-only items that may be present in `args.nuggetEditorLength` but not in `args.nuggetTextDocText!.length`, like a footnote marker. 

I am revising it to try to clarify what is happening, by introducing variable `editorLengthToTargetLocationAfterEvent`.

Another way to clarify may have been to write:
```ts
const lengthOnlyInEditor: number = args.nuggetEditorLength - args.nuggettextDocText.length
const desiredSelectionstart = targetSegmentRange.index + textLeadingUpToTargetLocationAfterEvent.length + lengthOnlyInEditor
```

</blockquote></details>

Hmm, I think I understand. The textLeadingUpToTargetLocationAfterEvent including text in the nugget, but not non-text length that the nugget includes. I like they updated name much better, because it represents all the length of the text and non-text objects that come before the selection.

PR #1224, opened 2022-01-11
SF-1467 Note can attach to text outside of verse reference

RaymondLuong3:
* GetVerseText gets the texts in the delta that also includes section headings
* determine note anchoring position using all text belonging to a verse
* handle anchoring notes to verse text on the front end

Section heading notes

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1224)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/test-utils.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/test-utils.ts
@@ -47,6 +47,12 @@ export function getCombinedVerseTextDoc(id: TextDocId): TextData {
   delta.insert({ verse: { number: '2-3', style: 'v' } });
   delta.insert(`${id.textType}: chapter ${id.chapterNum}, verse 2-3.`, { segment: `verse_${id.chapterNum}_2-3` });
   delta.insert('\n', { para: { style: 'p' } });
+  delta.insert('Text in section heading', { segment: 's_2' });
+  delta.insert('\n', { para: { style: 's' } });
+  delta.insert({ blank: true }, { segment: 'p_2' });
+  delta.insert({ verse: { number: '4', style: 'v' } });
+  delta.insert(`${id.textType}: chapter ${id.chapterNum}, verse 4.`, { segment: `verse_${id.chapterNum}_4` });
+  delta.insert('\n', { para: { style: 'p' } });
   return delta;
 }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -287,6 +287,11 @@ export class TextViewModel {
     return Array.from(this._segments.keys()).filter(r => r.indexOf(ref + '/') === 0);
   }
 
+  /** Get the segments that fall within a given verse reference. A segment is considered
+   * to be in the reference if its ref is (or the first preceding segment with a ref) of the
+   * format verse_c_v falls within the given verse reference.
+   * For example, the result for MAT 1:1 can be as follows: [verse_1_1, verse_1_1/p_1, s_1]
+   */
   getVerseSegments(verseRef?: VerseRef): string[] {
     const segmentsInVerseRef: string[] = [];
     if (verseRef == null) {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -288,8 +288,9 @@ export class TextViewModel {
   }
 
   /** Get the segments that fall within a given verse reference. A segment is considered
-   * to be in the reference if its ref is (or the first preceding segment with a ref) of the
-   * format verse_c_v falls within the given verse reference.
+   * to be in the reference if (1) its ref is in the format verse_c_v or verse_c_v-w, and that
+   * ref is within the given verse reference, or (2) its ref is not in that format, but the
+   * first preceding segment with a ref in that format is within the given verse reference.
    * For example, the result for MAT 1:1 can be as follows: [verse_1_1, verse_1_1/p_1, s_1]
    */
   getVerseSegments(verseRef?: VerseRef): string[] {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -295,14 +295,16 @@ export class TextViewModel {
     const verses: VerseRef[] = verseRef.allVerses();
     const startVerseNum: number = verses[0].verseNum;
     const lastVerseNum: number = verses[verses.length - 1].verseNum;
+    let matchStartNum = 0;
+    let matchLastNum = 0;
     for (const segment of this._segments.keys()) {
       const match: RegExpExecArray | null = VERSE_FROM_SEGMENT_REF_REGEX.exec(segment);
-      if (match == null) {
-        continue;
+      if (match != null) {
+        // update numbers for the new verse
+        const verseParts: string[] = match[1].split('-');
+        matchStartNum = +verseParts[0];
+        matchLastNum = +verseParts[verseParts.length - 1];
       }
-      const verseParts: string[] = match[1].split('-');
-      const matchStartNum: number = +verseParts[0];
-      const matchLastNum: number = +verseParts[verseParts.length - 1];
       const matchStartsWithin = matchStartNum >= startVerseNum && matchStartNum <= lastVerseNum;
       const matchEndsWithin = matchLastNum >= startVerseNum && matchLastNum <= lastVerseNum;
       if (matchStartsWithin || matchEndsWithin) {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -1074,43 +1074,44 @@ describe('EditorComponent', () => {
     it('embeds note on verse segments', fakeAsync(() => {
       const env = new TestEnvironment();
       env.addParatextNoteThread(6, 'MAT 1:2', '', { start: 0, length: 0 }, ['user01']);
-      env.addParatextNoteThread(7, 'LUK 1:2-3', '', { start: 0, length: 0 }, ['user01']);
+      env.addParatextNoteThread(7, 'LUK 1:0', 'for chapter', { start: 6, length: 11 }, ['user01']);
+      env.addParatextNoteThread(8, 'LUK 1:2-3', '', { start: 0, length: 0 }, ['user01']);
+      env.addParatextNoteThread(9, 'LUK 1:2-3', 'section heading', { start: 37, length: 15 }, ['user01']);
       env.setProjectUserConfig();
       env.wait();
-      const segment: HTMLElement = env.targetTextEditor.nativeElement.querySelector(
-        'usx-segment[data-segment=verse_1_1]'
-      )!;
-      expect(segment).not.toBeNull();
-
-      const note = segment.querySelector('display-note')! as HTMLElement;
-      expect(note).not.toBeNull();
-      expect(note.getAttribute('style')).toEqual('--icon-file: url(/assets/icons/TagIcons/01flag3.png);');
-      expect(note.getAttribute('title')).toEqual('Note from user01\n--- 2 more note(s) ---');
-      const contents = env.targetEditor.getContents();
+      const verse1Segment: HTMLElement = env.getSegmentElement('verse_1_1')!;
+      const verse1Note = verse1Segment.querySelector('display-note') as HTMLElement;
+      expect(verse1Note).not.toBeNull();
+      expect(verse1Note.getAttribute('style')).toEqual('--icon-file: url(/assets/icons/TagIcons/01flag3.png);');
+      expect(verse1Note.getAttribute('title')).toEqual('Note from user01\n--- 2 more note(s) ---');
+      let contents = env.targetEditor.getContents();
       expect(contents.ops![3].insert).toEqual('target: ');
       expect(contents.ops![4].attributes!['iconsrc']).toEqual('--icon-file: url(/assets/icons/TagIcons/01flag3.png);');
 
       // three notes in the segment on verse 3
-      const noteVerse3: HTMLElement[] = env.targetTextEditor.nativeElement.querySelectorAll(
-        'usx-segment[data-segment="verse_1_3"] display-note'
-      )!;
+      const noteVerse3: NodeListOf<Element> = env.getSegmentElement('verse_1_3')!.querySelectorAll('display-note')!;
       expect(noteVerse3.length).toEqual(3);
 
-      const blankSegmentNote = env.targetTextEditor.nativeElement.querySelector(
-        'usx-segment[data-segment="verse_1_2"] display-note'
-      )! as HTMLElement;
-      expect(blankSegmentNote).not.toBeNull();
+      const blankSegmentNote = env.getSegmentElement('verse_1_2')!.querySelector('display-note') as HTMLElement;
       expect(blankSegmentNote.getAttribute('style')).toEqual('--icon-file: url(/assets/icons/TagIcons/01flag1.png);');
       expect(blankSegmentNote.getAttribute('title')).toEqual('Note from user01');
 
       env.updateParams({ projectId: 'project01', bookId: 'LUK' });
       env.wait();
-      const combinedVerseUsxSegment: HTMLElement = env.targetTextEditor.nativeElement.querySelector(
-        'usx-segment[data-segment="verse_1_2-3"]'
-      );
+      const titleUsxSegment: HTMLElement = env.getSegmentElement('s_1')!;
+      expect(titleUsxSegment.classList).toContain('note-thread-segment');
+      const titleUsxNote: HTMLElement | null = titleUsxSegment.querySelector('display-note');
+      expect(titleUsxNote).not.toBeNull();
+
+      const combinedVerseUsxSegment: HTMLElement = env.getSegmentElement('verse_1_2-3')!;
       expect(combinedVerseUsxSegment.classList).toContain('note-thread-segment');
       const verse2and3CombinedNote: HTMLElement | null = combinedVerseUsxSegment.querySelector('display-note');
       expect(verse2and3CombinedNote).not.toBeNull();
+
+      const sectionHeadingUsxSegment: HTMLElement = env.getSegmentElement('s_2')!;
+      expect(sectionHeadingUsxSegment.classList).toContain('note-thread-segment');
+      const sectionHeadingNote: HTMLElement | null = sectionHeadingUsxSegment.querySelector('display-note');
+      expect(sectionHeadingNote).not.toBeNull();
       env.dispose();
     }));
 
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -1630,15 +1630,7 @@ namespace SIL.XForge.Scripture.Services
         private string GetVerseText(Delta delta, VerseRef verseRef)
         {
             string vref = string.IsNullOrEmpty(verseRef.Verse) ? verseRef.VerseNum.ToString() : verseRef.Verse;
-            string segment = $"verse_{verseRef.ChapterNum}_{vref}";
-            bool segmentFilter(JToken op)
-            {
-                return op.Type == JTokenType.Object &&
-                    op["attributes"] != null && op["attributes"]["segment"] != null &&
-                    (((string)op["attributes"]["segment"]) == segment ||
-                    ((string)op["attributes"]["segment"]).StartsWith(segment + "/"));
-            };
-            return delta.TryConcatenateInserts(out string verseText, segmentFilter) ? verseText : string.Empty;
+            return delta.TryConcatenateInserts(out string verseText, vref) ? verseText : string.Empty;
         }
 
         private TextAnchor GetThreadTextAnchor(CommentThread thread, Dictionary<int, ChapterDelta> chapterDeltas)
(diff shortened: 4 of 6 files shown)


marksvc:
:eyes:

marksvc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1224)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @marksvc and @RaymondLuong3)

___
*[src/SIL.XForge/Realtime/RichText/Delta.cs, line 226 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-MuwNN-iAp_h8-zWv3IA:-MuwNN-iAp_h8-zWv3IB:bq09v1) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7c3f47233d924355fdf6cbeb831afb60a3a094c/src/SIL.XForge/Realtime/RichText/Delta.cs#L226)):*
> ```Smalltalk
>         }
> 
>         public bool TryConcatenateInserts(out string opStr, string verse)
> ```

It's slightly confusing that `verse` is `string`. I'd be inclined to instead have `int verse`, but then we would need to ParseInt. What about using `string verseNum`?



_Code quote:_
```Smalltalk
 string verse
```
___
*[src/SIL.XForge/Realtime/RichText/Delta.cs, line 226 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-MuwNx8S1Z9OCeiF5gcf:-MuwNx8TEu2ZJtKCeZRo:b5u78al) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7c3f47233d924355fdf6cbeb831afb60a3a094c/src/SIL.XForge/Realtime/RichText/Delta.cs#L226)):*
> ```Smalltalk
>         }
> 
>         public bool TryConcatenateInserts(out string opStr, string verse)
> ```

This method just got a bunch more complex. Can we give a method comment that explains what is happening? From my examination so far, it seems that what we do in the revised method is:

With the exception of any ops that insert a chapter number or a newline:
```
If `verse == 0`:
  Select all ops that precede the first op that inserts a verse number
Else:
  Select all ops that follow an op that inserts a verse number matching `verse`, until we get to another op that inserts a verse number that doesn't match 'verse'.
Concat the selected ops and return.
```

I'm not sure that's different than what was before with `filter` except for the `verse==0` case.
But presumably I haven't understood it.



_Code quote:_
```Smalltalk
 TryConcatenateInserts
```
___
*[src/SIL.XForge/Realtime/RichText/Delta.cs, line 229 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-MuwMgRN-rmWdXp9zVWI:-MuwMgRN-rmWdXp9zVWJ:b961qmv) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7c3f47233d924355fdf6cbeb831afb60a3a094c/src/SIL.XForge/Realtime/RichText/Delta.cs#L229)):*
> ```Smalltalk
>         {
>             List<JToken> verseOps = new List<JToken>();
>             bool isTargetVerse = verse == "0";
> ```

Can the purpose or use of this be explained in a method comment? It looks like setting `verse=0` will result in selecting operations that _precede_ any op that inserts a verse number.



_Code quote:_
```Smalltalk
 "0";
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 290 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-MuwSTCT6WKCEo5-XSAL:-MuwSTCT6WKCEo5-XSAM:btjx4x7) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7c3f47233d924355fdf6cbeb831afb60a3a094c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L290)):*
> ```XML
>   }
> 
>   getVerseSegments(verseRef?: VerseRef): string[] {
> ```

It looks like the change causes the following:

Look thru all segments. Once we find a segment that matches our desired range, add it to a list.
Continue adding subsequent segments to the list, until we get to a segment that has a verse ref which is outside of our range. 

So a new behaviour would be that segments with references that don't match the regex, will also be included in the list, if they follow a regex match that relates to our verse specification and precede a regex match that does not relate to our verse specification. 

So, maybe what's desired is to collect segments with ref suffixes like `/p1`? If that's what is desired, maybe we could instead modify the regex from 
```
export const VERSE_FROM_SEGMENT_REF_REGEX = /verse_\d+_(\d+-?\d*)/;
```
to something like
```
export const VERSE_FROM_SEGMENT_REF_REGEX = /verse_\d+_(\d+-?\d*)(\/.*)?/;
```
to also match refs that optionally have the suffix.

Or maybe that's not what's happening here?

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1224)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge/Realtime/RichText/Delta.cs, line 226 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-MuwNx8S1Z9OCeiF5gcf:-Mv-gnvPBu7lj4r3R-LK:b60umw5) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7c3f47233d924355fdf6cbeb831afb60a3a094c/src/SIL.XForge/Realtime/RichText/Delta.cs#L226)):*

This method just got a bunch more complex. Can we give a method comment that explains what is happening? From my examination so far, it seems that what we do in the revised method is:

With the exception of any ops that insert a chapter number or a newline:
```
If `verse == 0`:
  Select all ops that precede the first op that inserts a verse number
Else:
  Select all ops that follow an op that inserts a verse number matching `verse`, until we get to another op that inserts a verse number that doesn't match 'verse'.
Concat the selected ops and return.
```

I'm not sure that's different than what was before with `filter` except for the `verse==0` case.
But presumably I haven't understood it.

</blockquote></details>

Yes, you have it right. And the difference between the previous implementation and the current one is that the previous one would not pick up the section headings since they are do not have a segment attribute that matches 'verse_1_1'. These segments look like `s_1`.
___
*[src/SIL.XForge/Realtime/RichText/Delta.cs, line 226 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-MuwNN-iAp_h8-zWv3IA:-Mv-ic7g3jUwqByyV-D8:b-bb9611) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7c3f47233d924355fdf6cbeb831afb60a3a094c/src/SIL.XForge/Realtime/RichText/Delta.cs#L226)):*

It's slightly confusing that `verse` is `string`. I'd be inclined to instead have `int verse`, but then we would need to ParseInt. What about using `string verseNum`?
</blockquote></details>

Good thinking. I've updated it to be verseRef.
___
*[src/SIL.XForge/Realtime/RichText/Delta.cs, line 229 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-MuwMgRN-rmWdXp9zVWI:-Mv-hMH9C_MJSPt16B-d:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7c3f47233d924355fdf6cbeb831afb60a3a094c/src/SIL.XForge/Realtime/RichText/Delta.cs#L229)):*

Can the purpose or use of this be explained in a method comment? It looks like setting `verse=0` will result in selecting operations that _precede_ any op that inserts a verse number.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 290 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-MuwSTCT6WKCEo5-XSAL:-Mv-hqc5CbMHW3W5GSDA:b-q92w0x) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7c3f47233d924355fdf6cbeb831afb60a3a094c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L290)):*

It looks like the change causes the following:

Look thru all segments. Once we find a segment that matches our desired range, add it to a list.
Continue adding subsequent segments to the list, until we get to a segment that has a verse ref which is outside of our range. 

So a new behaviour would be that segments with references that don't match the regex, will also be included in the list, if they follow a regex match that relates to our verse specification and precede a regex match that does not relate to our verse specification. 

So, maybe what's desired is to collect segments with ref suffixes like `/p1`? If that's what is desired, maybe we could instead modify the regex from 
```
export const VERSE_FROM_SEGMENT_REF_REGEX = /verse_\d+_(\d+-?\d*)/;
```
to something like
```
export const VERSE_FROM_SEGMENT_REF_REGEX = /verse_\d+_(\d+-?\d*)(\/.*)?/;
```
to also match refs that optionally have the suffix.

Or maybe that's not what's happening here?
</blockquote></details>

You mostly have it right. We are trying to get all segments that belong to a verse. This include all verses with the segment attribute `verse_x_x` and any segment with a suffix too. However, since section headings and other headings do not contain the `verse_x_x` part in its segmentRef, we assume that all  text between verses belongs to a given verse. That matches how paratext distinguishes which verse a note on a section heading belong.

marksvc (changes requested):
> Reviewed 5 of 6 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1224)* status: all files reviewed, 3 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge/Realtime/RichText/Delta.cs, line 226 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-MuwNN-iAp_h8-zWv3IA:-Mv07xdR4cCaopdkdRhi:b-a65so6) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7c3f47233d924355fdf6cbeb831afb60a3a094c/src/SIL.XForge/Realtime/RichText/Delta.cs#L226)):*

Good thinking. I've updated it to be verseRef.
</blockquote></details>

Haha, now I'd think it would want "MAT 1:2", except for the clarifying method comment :)
___
*[src/SIL.XForge/Realtime/RichText/Delta.cs, line 226 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-MuwNx8S1Z9OCeiF5gcf:-Mv08nXc5z7fNi0cY48-:byjt1nd) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7c3f47233d924355fdf6cbeb831afb60a3a094c/src/SIL.XForge/Realtime/RichText/Delta.cs#L226)):*

Yes, you have it right. And the difference between the previous implementation and the current one is that the previous one would not pick up the section headings since they are do not have a segment attribute that matches 'verse_1_1'. These segments look like `s_1`.
</blockquote></details>

Oh man. I was going to advocate for including section heads that _precede_ a verse rather than section heads that _follow_ a verse, but I can see how it makes sense to think of section heads as being "in" the preceding verse since they are in fact between `\v 1` and `\v 2` markers. I don't like it, however :)
The jira issue, for example, shows a section header that I think as being _for_ verse 10 and following, but in the structure of the document it's _prior_ to verse 10, and so kind of "in_ verse 9.
___
*[src/SIL.XForge/Realtime/RichText/Delta.cs, line 228 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-Mv0I6hX4-kREllTsz7V:-Mv0I6hX4-kREllTsz7W:b-x90jow) ([raw file](https://github.com/sillsdev/web-xforge/blob/ffab8232191ccadad5f3da99ff780ee2ea2e2437/src/SIL.XForge/Realtime/RichText/Delta.cs#L228)):*
> ```Smalltalk
>         public bool TryConcatenateInserts(out string opStr, Func<JToken, bool> filter = null)
>         {
>             Delta filteredDelta = filter == null ? this : new Delta(this.Ops.Where(op => filter(op)));
> ```

I imagine a bit trying to re-write the method as still using a filter, but using a bigger filter like:

```cs
filter(segmentRef) {
  If segmentRef is of the form `verse_c_v(/foo)`:
    If `c` and `v` are what we are looking for: Return true
    Else: Return false
  Else: (so segmentRef is something else, like `s_1`)
    Determine where in the `this.Ops` array we are, by index.
    Look for the first preceding `this.Ops` item with a segmentRef of the form `verse_c_v(/foo)`
    If that `c` and `v` are what we are looking for: Return true
    Else: Return false
}
```
Then it wouldn't need all these (JObject)op[InsertType]).Property Value Type pieces. We can leave the code like it is, but if you feel inspired you might have a try re-writing it to use a somewhat bigger filter.

Hmm, I guess the original filter code was also written with code like (string)op["attributes"] tho.



_Code quote:_
```Smalltalk
: new Delta(this.Ops.Where(op => filter(op)));
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 300 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-Mv0BYOf8ibIH5mXqzL5:-Mv0BYOg8BeXgfvl98uL:b-i2y3zr) ([raw file](https://github.com/sillsdev/web-xforge/blob/70edb5c2b86a3a8b97f065e788175d009f85966a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L300)):*
> ```XML
>     let matchStartNum = 0;
>     let matchLastNum = 0;
>     for (const segment of this._segments.keys()) {
> ```

Could we drop in an implementation comment? I struggle a bit with the the block of code. How about:

// Collect all segments that overlap with the specified verse range. A segment (such as a section heading, or additional paragraph) is considered to be in the range, if it's ref is, or the first preceding segment with a ref of, the format `verse_c_v` is known to be in the range. 

I.. hoped for conciseness there.



_Code quote:_
```XML
 for 
```
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 2137 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-Mv0c-L_AB2IRTgSmY3t:-Mv0c-L_AB2IRTgSmY3u:b-c1mhar) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7c3f47233d924355fdf6cbeb831afb60a3a094c/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2137)):*
> ```Smalltalk
>                         if (comp.alternateText == SelectionType.RelatedVerse)
>                         {
>                             before = before + text + after + "\n*";
> ```

Is there supposed to be an asterisk there? What's the meaning of that?



_Code quote:_
```Smalltalk
n*
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1224)* status: 4 of 6 files reviewed, 3 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge/Realtime/RichText/Delta.cs, line 228 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-Mv0I6hX4-kREllTsz7V:-Mv4lfZR-_7u0eoLqupc:b-xtgj5l) ([raw file](https://github.com/sillsdev/web-xforge/blob/ffab8232191ccadad5f3da99ff780ee2ea2e2437/src/SIL.XForge/Realtime/RichText/Delta.cs#L228)):*

I imagine a bit trying to re-write the method as still using a filter, but using a bigger filter like:

```cs
filter(segmentRef) {
  If segmentRef is of the form `verse_c_v(/foo)`:
    If `c` and `v` are what we are looking for: Return true
    Else: Return false
  Else: (so segmentRef is something else, like `s_1`)
    Determine where in the `this.Ops` array we are, by index.
    Look for the first preceding `this.Ops` item with a segmentRef of the form `verse_c_v(/foo)`
    If that `c` and `v` are what we are looking for: Return true
    Else: Return false
}
```
Then it wouldn't need all these (JObject)op[InsertType]).Property Value Type pieces. We can leave the code like it is, but if you feel inspired you might have a try re-writing it to use a somewhat bigger filter.

Hmm, I guess the original filter code was also written with code like (string)op["attributes"] tho.
</blockquote></details>

That is a good thought on improving the algorithm. I think I'm happy with what is already here.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 300 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-Mv0BYOf8ibIH5mXqzL5:-Mv4zME9D9RlQX_DCeMY:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/70edb5c2b86a3a8b97f065e788175d009f85966a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L300)):*

Could we drop in an implementation comment? I struggle a bit with the the block of code. How about:

// Collect all segments that overlap with the specified verse range. A segment (such as a section heading, or additional paragraph) is considered to be in the range, if it's ref is, or the first preceding segment with a ref of, the format `verse_c_v` is known to be in the range. 

I.. hoped for conciseness there.
</blockquote></details>

Done.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 2137 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-Mv0c-L_AB2IRTgSmY3t:-Mv4yxWI05UjzdcPir6J:b-cdsoy7) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7c3f47233d924355fdf6cbeb831afb60a3a094c/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2137)):*

Is there supposed to be an asterisk there? What's the meaning of that?
</blockquote></details>

There is a footnote on the following related verse segment that is represented by the asterisk. I  added a comment.

marksvc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1224)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 293 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-MvLJBfA0Ww8xyjDkygC:-MvLJBfA0Ww8xyjDkygD:b4hyd9p) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4d7015a03ee4dc6105dcb44f817169b373f46b1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L293)):*
> ```XML
>    * to be in the reference if its ref is (or the first preceding segment with a ref) of the
>    * format verse_c_v falls within the given verse reference.
>    * For example, the result for MAT 1:1 can be as follows: [verse_1_1, verse_1_1/p_1, s_1]
> ```

Thank you. I think we had just better not be concise. How about

  /** Get the segments that are within a given verse reference. A segment is considered
   * to be in the reference if (1) its ref is in the format verse_c_v or verse_c_v-w, and that ref is within the given verse reference, or (2) its ref is not in that format, but the first preceding segment with a ref in that format, is a ref within the given verse reference.
   * For example, the result for MAT 1:1 can be as follows: [verse_1_1, verse_1_1/p_1, s_1]
   */

I also steered us wrong about `verse_c_v`, and tried to correct it above.
Good job having an example.



_Code quote:_
```XML
  /** Get the segments that fall within a given verse reference. A segment is considered
   * to be in the reference if its ref is (or the first preceding segment with a ref) of the
   * format verse_c_v falls within the given verse reference.
   * For example, the result for MAT 1:1 can be as follows: [verse_1_1, verse_1_1/p_1, s_1]
```
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 2137 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-MvLHi4FBKi6I9Z9mi2D:-MvLHi4FBKi6I9Z9mi2E:b7svaev) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4d7015a03ee4dc6105dcb44f817169b373f46b1/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2137)):*
> ```Smalltalk
>                         if (comp.alternateText == SelectionType.RelatedVerse)
>                         {
>                             // The alternate text is in a proceeding paragraph with a footnote represented by '*'
> ```

(By this do you mean something like "subsequent"? Or maybe what was meant was "preceding"?)



_Code quote:_
```Smalltalk
proceeding
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1224)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 293 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-MvLJBfA0Ww8xyjDkygC:-MvPW0C-0vrOlTBDt_Wt:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4d7015a03ee4dc6105dcb44f817169b373f46b1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L293)):*

Thank you. I think we had just better not be concise. How about

  /** Get the segments that are within a given verse reference. A segment is considered
   * to be in the reference if (1) its ref is in the format verse_c_v or verse_c_v-w, and that ref is within the given verse reference, or (2) its ref is not in that format, but the first preceding segment with a ref in that format, is a ref within the given verse reference.
   * For example, the result for MAT 1:1 can be as follows: [verse_1_1, verse_1_1/p_1, s_1]
   */

I also steered us wrong about `verse_c_v`, and tried to correct it above.
Good job having an example.
</blockquote></details>

Done.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 2137 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1224#-MvLHi4FBKi6I9Z9mi2D:-MvPWAr3A3WiRxUNDnAd:b-9gbie3) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4d7015a03ee4dc6105dcb44f817169b373f46b1/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2137)):*

(By this do you mean something like "subsequent"? Or maybe what was meant was "preceding"?)
</blockquote></details>

I did mean subsequent. In my mind they mean the same thing, but I do like subsequent.

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1224)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1225, opened 2022-01-14
SF-1470 Do not allow backspace or delete on non-text elements

RaymondLuong3:
When hitting backspace on a blank segment, the editor history records information on the event but the user does not see any changes. However, if a note is embedded with the blank segment, hitting delete, and then undo will result in duplicate verses. I've put this change on master branch because this change is an improvement on the behaviour of blank segments as a whole.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1225)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -506,8 +506,10 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
       const text = this._editor.getText(range.index, range.length);
       return text !== '';
     }
+    const text = this._editor.getText(range.index - 1, 1);
+    const isTextDeletion: boolean = text != null && text.length > 0;
 
-    return this._segment != null && range.index !== this._segment.range.index;
+    return isTextDeletion && this._segment != null && range.index !== this._segment.range.index;
   }
 
   private isDeleteAllowed(range: RangeStatic): boolean {
@@ -519,8 +521,12 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
       const text = this._editor.getText(range.index, range.length);
       return text !== '';
     }
+    const text = this._editor.getText(range.index, 1);
+    const isTextDeletion: boolean = text != null && text.length > 0;
 
-    return this._segment != null && range.index !== this._segment.range.index + this._segment.range.length;
+    return (
+      isTextDeletion && this._segment != null && range.index !== this._segment.range.index + this._segment.range.length
+    );
   }
 
   private moveNextSegment(end: boolean = true): void {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -923,6 +923,8 @@ describe('EditorComponent', () => {
       env.setProjectUserConfig();
       env.wait();
       let range = env.component.target!.getSegmentRange('verse_1_2')!;
+      let contents = env.targetEditor.getContents(range.index, 1);
+      expect(contents.ops![0].insert.blank).toBeDefined();
 
       // set selection on a blank segment
       env.targetEditor.setSelection(range.index, 'user');
@@ -934,6 +936,8 @@ describe('EditorComponent', () => {
       expect(env.targetEditor.history['stack']['undo'].length).toEqual(0);
       env.pressKey('delete');
       expect(env.targetEditor.history['stack']['undo'].length).toEqual(0);
+      contents = env.targetEditor.getContents(range.index, 1);
+      expect(contents.ops![0].insert.blank).toBeDefined();
 
       // set selection at the end of a segment
       range = env.component.target!.getSegmentRange('verse_1_4')!;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -918,10 +918,11 @@ describe('EditorComponent', () => {
       env.dispose();
     }));
 
-    it('user cannot backspace or delete blanks', fakeAsync(() => {
+    it('backspace and delete disabled for non-text elements and at segment boundaries', fakeAsync(() => {
       const env = new TestEnvironment();
       env.setProjectUserConfig();
       env.wait();
+      expect(env.targetEditor.history['stack']['undo'].length).withContext('setup').toEqual(0);
       let range = env.component.target!.getSegmentRange('verse_1_2')!;
       let contents = env.targetEditor.getContents(range.index, 1);
       expect(contents.ops![0].insert.blank).toBeDefined();
@@ -929,7 +930,6 @@ describe('EditorComponent', () => {
       // set selection on a blank segment
       env.targetEditor.setSelection(range.index, 'user');
       env.wait();
-      expect(env.targetEditor.hasFocus()).toBe(true);
       expect(env.targetEditor.history['stack']['undo'].length).toEqual(0);
 
       env.pressKey('backspace');
@@ -939,13 +939,28 @@ describe('EditorComponent', () => {
       contents = env.targetEditor.getContents(range.index, 1);
       expect(contents.ops![0].insert.blank).toBeDefined();
 
-      // set selection at the end of a segment
+      // set selection at segment boundaries
       range = env.component.target!.getSegmentRange('verse_1_4')!;
       env.targetEditor.setSelection(range.index + range.length, 'user');
       env.wait();
-      expect(env.targetEditor.hasFocus()).toBe(true);
       env.pressKey('delete');
       expect(env.targetEditor.history['stack']['undo'].length).toEqual(0);
+      env.targetEditor.setSelection(range.index, 'user');
+      env.wait();
+      env.pressKey('backspace');
+      expect(env.targetEditor.history['stack']['undo'].length).toEqual(0);
+
+      // other non-text elements
+      range = env.component.target!.getSegmentRange('verse_1_1')!;
+      env.targetEditor.insertEmbed(range.index, 'note', { caller: 'a', style: 'ft' }, 'api');
+      env.wait();
+      contents = env.targetEditor.getContents(range.index, 1);
+      expect(contents.ops![0].insert.note).toBeDefined();
+      env.targetEditor.setSelection(range.index + 1, 'user');
+      env.pressKey('backspace');
+      expect(env.targetEditor.history['stack']['undo'].length).toEqual(0);
+      contents = env.targetEditor.getContents(range.index, 1);
+      expect(contents.ops![0].insert.note).toBeDefined();
       env.dispose();
     }));
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -918,6 +918,33 @@ describe('EditorComponent', () => {
       env.dispose();
     }));
 
+    it('user cannot backspace or delete blanks', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig();
+      env.wait();
+      let range = env.component.target!.getSegmentRange('verse_1_2')!;
+
+      // set selection on a blank segment
+      env.targetEditor.setSelection(range.index, 'user');
+      env.wait();
+      expect(env.targetEditor.hasFocus()).toBe(true);
+      expect(env.targetEditor.history['stack']['undo'].length).toEqual(0);
+
+      env.pressKey('backspace');
+      expect(env.targetEditor.history['stack']['undo'].length).toEqual(0);
+      env.pressKey('delete');
+      expect(env.targetEditor.history['stack']['undo'].length).toEqual(0);
+
+      // set selection at the end of a segment
+      range = env.component.target!.getSegmentRange('verse_1_4')!;
+      env.targetEditor.setSelection(range.index + range.length, 'user');
+      env.wait();
+      expect(env.targetEditor.hasFocus()).toBe(true);
+      env.pressKey('delete');
+      expect(env.targetEditor.history['stack']['undo'].length).toEqual(0);
+      env.dispose();
+    }));
+
     it('undo/redo', fakeAsync(() => {
       const env = new TestEnvironment();
       env.setProjectUserConfig({ selectedBookNum: 40, selectedChapterNum: 1, selectedSegment: 'verse_1_2' });
@@ -1384,6 +1411,10 @@ class TestEnvironment {
     return this.trainingProgress.query(By.css('#training-close-button'));
   }
 
+  get targetTextEditor(): DebugElement {
+    return this.fixture.debugElement.query(By.css('#target-text-area .ql-container'));
+  }
+
   get sourceTextArea(): DebugElement {
     return this.fixture.debugElement.query(By.css('#source-text-area'));
   }
@@ -1646,6 +1677,15 @@ class TestEnvironment {
     this.targetEditor.setSelection(selection.index, 'user');
     this.wait();
     return selection.index;
+  }
+
+  pressKey(key: string): void {
+    const keyCodes = { backspace: 8, delete: 46 };
+    if (keyCodes[key] == null) {
+      throw new Error('key code does not exist');
+    }
+    this.targetEditor.root.dispatchEvent(new KeyboardEvent('keydown', { keyCode: keyCodes[key] }));
+    this.wait();
   }
 
   triggerUndo(): void {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -930,6 +930,8 @@ describe('EditorComponent', () => {
       // set selection on a blank segment
       env.targetEditor.setSelection(range.index, 'user');
       env.wait();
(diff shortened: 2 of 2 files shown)


marksvc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1225)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[-- commits, line 2 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-MuwYC58AXDNG54NfAzC):*
Is it significant that past the edge of some segments, there is a verse number, which may be what is "not text"? Because if so, you could have a segment that is adjacent not to a verse number, but to another segment of the same verse, or to another segment like a section header.

For example

segment section header: Header text
segment verse 1:  Verse text
segment verse 1/p1: More verse text
segment verse 2: Next verse

If we put the insertion point at  the beginning or end of the "segment verse 1" segment and backspace or delete, do we still hit our guard rails?



_Code quote:_
```
on non-text elements
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 510 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-MuwWdEf9y66EXbW3WTD:-MuwWdEf9y66EXbW3WTE:b40nkyw) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4d821c554310a0bf3948499bfc3ba8e33683d38/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L510)):*
> ```XML
>     }
>     const text = this._editor.getText(range.index - 1, 1);
>     const isTextDeletion: boolean = text != null && text.length > 0;
> ```

Okay. I suspect that this might cause some different behaviour when trying to delete non-text things that are or aren't at the edge of a segment.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 936 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-MuwXC6T0jj-Cgr_bFHy:-MuwXC6T0jj-Cgr_bFHz:bm6xdwk) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4d821c554310a0bf3948499bfc3ba8e33683d38/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L936)):*
> ```XML
>       expect(env.targetEditor.history['stack']['undo'].length).toEqual(0);
>       env.pressKey('delete');
>       expect(env.targetEditor.history['stack']['undo'].length).toEqual(0);
> ```

Can we here also assert that the segment still contains just the blank?
Also, what if we put the user selection at the end of the blank and press backspace? That isn't part of the change to text.component.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 1687 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-MuwXvrI3CDjK7pxczMM:-MuwXvrI3CDjK7pxczMN:b-ul7oq9) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4d821c554310a0bf3948499bfc3ba8e33683d38/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1687)):*
> ```XML
>       throw new Error('key code does not exist');
>     }
>     this.targetEditor.root.dispatchEvent(new KeyboardEvent('keydown', { keyCode: keyCodes[key] }));
> ```

Good job.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1225)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[-- commits, line 2 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-MuwYC58AXDNG54NfAzC):*

Is it significant that past the edge of some segments, there is a verse number, which may be what is "not text"? Because if so, you could have a segment that is adjacent not to a verse number, but to another segment of the same verse, or to another segment like a section header.

For example

segment section header: Header text
segment verse 1:  Verse text
segment verse 1/p1: More verse text
segment verse 2: Next verse

If we put the insertion point at  the beginning or end of the "segment verse 1" segment and backspace or delete, do we still hit our guard rails?
</blockquote></details>

That's right. Since our text-view-model keeps track of every segment within a verse, we won't be able to delete or backspace against the segment boundaries. So this case works, and was working before my changes.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 510 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-MuwWdEf9y66EXbW3WTD:-Mv-WzQG4YROfBk5uSvQ:bvnb5yn) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4d821c554310a0bf3948499bfc3ba8e33683d38/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L510)):*

Okay. I suspect that this might cause some different behaviour when trying to delete non-text things that are or aren't at the edge of a segment.
</blockquote></details>

That's correct. Users will not be able to backspace or delete things like notes icons, cross-references and other embeds. My understanding is that SF is best used to draft translations, but the formatting and more complex operations like cross references and biblical terms should be done in Paratext. So this is reasonable to me.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 936 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-MuwXC6T0jj-Cgr_bFHy:-Mv-dKtR1zpbm-s3LcWT:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4d821c554310a0bf3948499bfc3ba8e33683d38/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L936)):*

Can we here also assert that the segment still contains just the blank?
Also, what if we put the user selection at the end of the blank and press backspace? That isn't part of the change to text.component.
</blockquote></details>

Done.

marksvc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1225)* status: all files reviewed, 3 unresolved discussions (waiting on @RaymondLuong3)

___
*[-- commits, line 2 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-MuwYC58AXDNG54NfAzC):*

That's right. Since our text-view-model keeps track of every segment within a verse, we won't be able to delete or backspace against the segment boundaries. So this case works, and was working before my changes.
</blockquote></details>

Sorry, somehow it didn't sink in that the guard rails are already saying no to backspace at the beginning of a segment, and no to delete at the end of a segment.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 510 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-MuwWdEf9y66EXbW3WTD:-Mv02VT6443PRdvMKEcG:b-x5rry8) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4d821c554310a0bf3948499bfc3ba8e33683d38/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L510)):*

That's correct. Users will not be able to backspace or delete things like notes icons, cross-references and other embeds. My understanding is that SF is best used to draft translations, but the formatting and more complex operations like cross references and biblical terms should be done in Paratext. So this is reasonable to me.
</blockquote></details>

(I mean it may cause a difference in behaviour between:
- A user presses backspace next to a figure or note icon that is in the middle of a segment in SF.
- A user presses backspace next to a figure or note icon that is at the beginning of a segment in SF.
)

Oh, actually, maybe not that, but it looks like it may cause be a difference of behaviour between:
- User selects text "aXb" where X is a figure, and presses backspace. This code looks like the result will be that backspace is allowed.
- User clicks after X in "aXb" and presses backspace. This code looks like the result will be that nothing happens.


And I'm not complaining about the patch, just pointing out something I noticed (or maybe misunderstood :)
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 936 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-MuwXC6T0jj-Cgr_bFHy:-Mv06nM-BVPshXFLMaru:b3jts3r) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4d821c554310a0bf3948499bfc3ba8e33683d38/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L936)):*
> (me) That isn't part of the change to text.component.

Oops, haha, yes it is; I was wrong.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 921 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-Mv05agu2RpRrNSqsMx3:-Mv05agu2RpRrNSqsMx4:b-eo9pzn) ([raw file](https://github.com/sillsdev/web-xforge/blob/0292ddcbfc3c36557da994c3e270da23c977f6e2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L921)):*
> ```XML
>     }));
> 
>     it('user cannot backspace or delete blanks', fakeAsync(() => {
> ```

From the jira issue, it sounds like the issue may be related to backspacing away _note icons_, not blanks. Is it not significant to have in the test that there is a note icon next to the blank, and the user's cursor is just to the right of the note icon and they press backspace?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 947 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-Mv05ucgCTznj4DE7NFq:-Mv05ucgCTznj4DE7NFr:b-he73bm) ([raw file](https://github.com/sillsdev/web-xforge/blob/0292ddcbfc3c36557da994c3e270da23c977f6e2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L947)):*
> ```XML
>       env.wait();
>       expect(env.targetEditor.hasFocus()).toBe(true);
>       env.pressKey('delete');
> ```

I would have expected text.component to disallow this already anyway, from reading its code, since it's at the end of the segment. Not that we shouldn't have a test for it. But it might be significant to also assert what happens if the user here also presses _backspace_ (which one might thing has the potential to delete the blank). 



_Code quote:_
```XML
delete'
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 949 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-Mv06EWlD5NU6Yqtedzh:-Mv06EWlD5NU6Yqtedzi:b9ftjju) ([raw file](https://github.com/sillsdev/web-xforge/blob/0292ddcbfc3c36557da994c3e270da23c977f6e2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L949)):*
> ```XML
>       env.pressKey('delete');
>       expect(env.targetEditor.history['stack']['undo'].length).toEqual(0);
>       env.dispose();
> ```

Let's specify here also what the editor content should be. For example, whether it still has the blank.

Actually, maybe verse_1_4 isn't a segment that contains only a blank anyway? If not, then let's specify what happens when there is a segment, which contains only a blank, and the user clicks to _the right of the blank_ and presses backspace.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1225)* status: all files reviewed, 3 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 510 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-MuwWdEf9y66EXbW3WTD:-Mv0J5j51i6355zW8FlC:b858kyp) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4d821c554310a0bf3948499bfc3ba8e33683d38/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L510)):*

(I mean it may cause a difference in behaviour between:
- A user presses backspace next to a figure or note icon that is in the middle of a segment in SF.
- A user presses backspace next to a figure or note icon that is at the beginning of a segment in SF.
)

Oh, actually, maybe not that, but it looks like it may cause be a difference of behaviour between:
- User selects text "aXb" where X is a figure, and presses backspace. This code looks like the result will be that backspace is allowed.
- User clicks after X in "aXb" and presses backspace. This code looks like the result will be that nothing happens.


And I'm not complaining about the patch, just pointing out something I noticed (or maybe misunderstood :)
</blockquote></details>

That is right, the figure can be deleted if it is part of a larger selection (it cannot be selected on its own). But then the history stack works as expected in that scenario. It's not ideal that we have behaviour that only prevents deleting and backspace but highlight and delete works. But this is the best I could come up with.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 921 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-Mv05agu2RpRrNSqsMx3:-Mv0Jrc02ZbsaZsDkkRL:b-o2sr40) ([raw file](https://github.com/sillsdev/web-xforge/blob/0292ddcbfc3c36557da994c3e270da23c977f6e2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L921)):*

From the jira issue, it sounds like the issue may be related to backspacing away _note icons_, not blanks. Is it not significant to have in the test that there is a note icon next to the blank, and the user's cursor is just to the right of the note icon and they press backspace?
</blockquote></details>

The jira issue was initially about note icons. But it turned out that it applied to blanks and other non-text elements. So this will cover blanks and those other cases too. I added a test for a footnote, but I'm inclined to say that we do not need to test every possible embed.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 947 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-Mv05ucgCTznj4DE7NFq:-Mv0XqZHC8j53uX3e1mF:b-7x23kk) ([raw file](https://github.com/sillsdev/web-xforge/blob/0292ddcbfc3c36557da994c3e270da23c977f6e2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L947)):*

I would have expected text.component to disallow this already anyway, from reading its code, since it's at the end of the segment. Not that we shouldn't have a test for it. But it might be significant to also assert what happens if the user here also presses _backspace_ (which one might thing has the potential to delete the blank). 
</blockquote></details>

Right, this case was about hitting delete at the segment boundary end. I'll add a test for the segment boundary start. 
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 949 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-Mv06EWlD5NU6Yqtedzh:-Mv0ZByu61PuY5pEmkMP:b-2f79cz) ([raw file](https://github.com/sillsdev/web-xforge/blob/0292ddcbfc3c36557da994c3e270da23c977f6e2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L949)):*

Let's specify here also what the editor content should be. For example, whether it still has the blank.

Actually, maybe verse_1_4 isn't a segment that contains only a blank anyway? If not, then let's specify what happens when there is a segment, which contains only a blank, and the user clicks to _the right of the blank_ and presses backspace.
</blockquote></details>

You are right, verse 4 has some text. Verse 2 is the segment with a blank, so that scenario you described is tested a few lines above.

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1225)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 936 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-MvLPvm_-lV17LcMjv9-:-MvLPvm_-lV17LcMjv90:b-t9lvcg) ([raw file](https://github.com/sillsdev/web-xforge/blob/27b6e23becec3d071e173cb77ca4d87d209b1702/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L936)):*
> ```XML
> 
>       env.pressKey('backspace');
>       expect(env.targetEditor.history['stack']['undo'].length).toEqual(0);
> ```

(To be meaningful, this presumably needs to be done at `.setSelection(range.index+1)`, not here at `.setSelection(range.index)`.)



_Code quote:_
```XML
      env.pressKey('backspace');
      expect(env.targetEditor.history['stack']['undo'].length).toEqual(0);
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1225)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts, line 936 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-MvLPvm_-lV17LcMjv9-:-MvPX3Kn8niZF2l7Tyvj:b-aazaa5) ([raw file](https://github.com/sillsdev/web-xforge/blob/27b6e23becec3d071e173cb77ca4d87d209b1702/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L936)):*

(To be meaningful, this presumably needs to be done at `.setSelection(range.index+1)`, not here at `.setSelection(range.index)`.)
</blockquote></details>

Yea good thinking. I've added an expect and a comment since it is not clear that programmatically the selection is always moved after the blank.

Nateowami:
I think "Do no allow" should be "Do not allow"

RaymondLuong3 (commented):
You are right!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1225#-:-MvPiz_p5LeugEz9sfNQ:b-ey7prd)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @marksvc)

PR #1230, opened 2022-01-26
SF-1479a Retry request for Transcelerator questions when network fails

Nateowami:
Previously our strategy for dealing with a lack of a good network connection was to try to detect whether the user was online (using the PwaService), and wait for the user to be online before initiating a request. This works to a degree, but if the request fails we treat it as a hard error and don't retry. The page is then essentially broken. One place where this could happen is in the import questions dialog, which immediately fetches the Transcelerator questions when the dialog opens, so they'll be ready if the user selects to import from Transcelerator.

A much better strategy is to keep trying as long as the error is only a network error. In order to make this as simple as possible for the components that use it, I've created a RetryingRequest class. There are a lot of doc comments for that class, so I would recommend starting there.

The rest of this change is:
- Using the RetryingRequest in the ImportQuestionsDialog
- A TestingRetryingRequestService that creates RetryingRequests that can easily be used in tests. That way components can easily test what happens when different network conditions occur.
- Tests for RetryingRequest and ImportQuestionsDialog
- A change in SubscriptionDisposable to allow subclasses of SubscriptionDisposable (that would be most of our components I think) to access the ngUnsubscribe property, thereby allowing more flexibility in how subcasses perform cleanup. This change allowed the ImportQuestionsDialog to do `projectService.transceleratorQuestions(this.data.projectId, this.ngUnsubscribe)`, where `this.ngUnsubscribe` (a `Subject<void>`) is being used to inform the RetryingRequest when the dialog is being destroyed and the data is no longer needed (and hence it can stop retrying).
 
Here's how part of the import questions dialog looks when the network request is in various states:

Offline | Attempting | Success
--------|------------|--------
 |  | 

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1230)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html
@@ -32,12 +32,17 @@
                 mat-flat-button
                 color="primary"
                 (click)="importFromTranscelerator()"
-                [disabled]="showTransceleratorOfflineMsg"
+                [disabled]="transceleratorRequest.status !== 'complete'"
               >
                 {{ t("import_from_transcelerator") }}
               </button>
               <a mat-button [href]="urls.transceleratorImportHelpPage" target="_blank">{{ t("learn_more") }}</a>
-              <mat-error *ngIf="showTransceleratorOfflineMsg">{{ t("no_transcelerator_offline") }}</mat-error>
+              <mat-error *ngIf="transceleratorRequest.status === 'offline'">
+                {{ t("no_transcelerator_offline") }}
+              </mat-error>
+              <mat-error *ngIf="transceleratorRequest.status === 'trying' && transceleratorRequest.failedAttempts > 0">
+                {{ t("network_error_transcelerator", { count: transceleratorRequest.failedAttempts }) }}
+              </mat-error>
             </mat-card-actions>
           </mat-card>
           <mat-card>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts
@@ -14,13 +14,13 @@ import { TextInfo } from 'realtime-server/lib/esm/scriptureforge/models/text-inf
 import { fromVerseRef } from 'realtime-server/lib/esm/scriptureforge/models/verse-ref-data';
 import { Canon } from 'realtime-server/lib/esm/scriptureforge/scripture-utils/canon';
 import { VerseRef } from 'realtime-server/lib/esm/scriptureforge/scripture-utils/verse-ref';
-import { Observable, of, Subject } from 'rxjs';
+import { BehaviorSubject, Observable, of, throwError } from 'rxjs';
 import { anything, capture, instance, mock, verify, when } from 'ts-mockito';
 import { AuthService } from 'xforge-common/auth.service';
 import { CsvService } from 'xforge-common/csv-service.service';
 import { RealtimeQuery } from 'xforge-common/models/realtime-query';
-import { PwaService } from 'xforge-common/pwa.service';
 import { configureTestingModule, TestTranslocoModule } from 'xforge-common/test-utils';
+import { TestingRetryingRequestService } from 'xforge-common/testing-retrying-request.service';
 import { UICommonModule } from 'xforge-common/ui-common.module';
 import { QuestionDoc } from '../../core/models/question-doc';
 import { TextsByBookId } from '../../core/models/texts-by-book-id';
@@ -37,7 +37,6 @@ const mockedProjectService = mock(SFProjectService);
 const mockedAuthService = mock(AuthService);
 const mockedCookieService = mock(CookieService);
 const mockedMdcDialog = mock(MdcDialog);
-const mockedPwaService = mock(PwaService);
 const mockedCsvService = mock(CsvService);
 
 describe('ImportQuestionsDialogComponent', () => {
@@ -48,7 +47,6 @@ describe('ImportQuestionsDialogComponent', () => {
       { provide: SFProjectService, useMock: mockedProjectService },
       { provide: CookieService, useMock: mockedCookieService },
       { provide: MdcDialog, useMock: mockedMdcDialog },
-      { provide: PwaService, useMock: mockedPwaService },
       { provide: CsvService, useMock: mockedCsvService }
     ]
   }));
@@ -393,7 +391,7 @@ describe('ImportQuestionsDialogComponent', () => {
     const env = new TestEnvironment({ offline: true });
     expect(env.importFromTransceleratorButton.disabled).toBe(true);
     expect(env.errorMessages).toEqual(['Importing from Transcelerator is not available offline.']);
-    env.setOnline();
+    env.setOnline(true);
     expect(env.importFromTransceleratorButton.disabled).toBe(false);
     expect(env.errorMessages).toEqual([]);
     env.click(env.importFromTransceleratorButton);
@@ -447,7 +445,7 @@ class TestEnvironment {
   mockedImportQuestionsConfirmationMdcDialogRef =
     mock<MdcDialogRef<ImportQuestionsConfirmationDialogComponent>>(MdcDialogRef);
   editedTransceleratorQuestionIds: string[] = [];
-  uponOnline$ = new Subject<void>();
+  online$ = new BehaviorSubject<boolean>(true);
 
   private questions: TransceleratorQuestion[] = [
     {
@@ -498,9 +496,8 @@ class TestEnvironment {
   ) {
     this.questions = options.transceleratorQuestions || this.questions;
     this.errorOnFetchQuestions = !!options.errorOnFetchQuestions;
-    when(mockedPwaService.online).thenReturn(this.uponOnline$.toPromise());
-    if (!options.offline) {
-      this.uponOnline$.complete();
+    if (options.offline === true) {
+      this.online$.next(false);
     }
 
     this.fixture = TestBed.createComponent(ChildViewContainerComponent);
@@ -633,8 +630,8 @@ class TestEnvironment {
     );
   }
 
-  setOnline() {
-    this.uponOnline$.complete();
+  setOnline(value: boolean) {
+    this.online$.next(value);
     tick();
     this.fixture.detectChanges();
   }
@@ -704,11 +701,13 @@ class TestEnvironment {
 
   private setupTransceleratorQuestions(): void {
     if (this.errorOnFetchQuestions) {
-      when(mockedProjectService.transceleratorQuestions('project01')).thenReject(
-        new Error('Transcelerator version unsupported')
+      when(mockedProjectService.transceleratorQuestions('project01', anything())).thenCall(() =>
+        TestingRetryingRequestService.createRequest(throwError(new Error('Transcelerator version unsupported')))
       );
     } else {
-      when(mockedProjectService.transceleratorQuestions('project01')).thenCall(() => Promise.resolve(this.questions));
+      when(mockedProjectService.transceleratorQuestions('project01', anything())).thenCall(() =>
+        TestingRetryingRequestService.createRequest(of(this.questions), this.online$)
+      );
     }
     when(mockedProjectService.queryQuestions('project01')).thenResolve({
       ready$: new Observable<void>(subscriber => {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts
@@ -13,8 +13,8 @@ import { objectId } from 'xforge-common/utils';
 import { TranslocoService } from '@ngneat/transloco';
 import { I18nService } from 'xforge-common/i18n.service';
 import { ExternalUrlService } from 'xforge-common/external-url.service';
-import { PwaService } from 'xforge-common/pwa.service';
 import { CsvService } from 'xforge-common/csv-service.service';
+import { RetryingRequest } from 'xforge-common/retrying-request.service';
 import { environment } from '../../../environments/environment';
 import { QuestionDoc } from '../../core/models/question-doc';
 import { TextsByBookId } from '../../core/models/texts-by-book-id';
@@ -78,7 +78,6 @@ export class ImportQuestionsDialogComponent extends SubscriptionDisposable {
   loading = false;
   importClicked: boolean = false;
   maxListItemsToDisplay = 100;
-  showTransceleratorOfflineMsg = true;
 
   importing: boolean = false;
   importedCount: number = 0;
@@ -97,7 +96,8 @@ export class ImportQuestionsDialogComponent extends SubscriptionDisposable {
     filter: this.filterControl
   });
 
-  promiseForTransceleratorQuestions: Promise<TransceleratorQuestion[]>;
+  transceleratorRequest: RetryingRequest<TransceleratorQuestion[]>;
+  promiseForTransceleratorQuestions: Promise<TransceleratorQuestion[] | undefined>;
   promiseForQuestionDocQuery: Promise<RealtimeQuery<QuestionDoc>>;
 
   constructor(
@@ -106,7 +106,6 @@ export class ImportQuestionsDialogComponent extends SubscriptionDisposable {
     private readonly dialogRef: MatDialogRef<ImportQuestionsDialogComponent>,
     private readonly transloco: TranslocoService,
     private readonly mdcDialog: MdcDialog,
-    private readonly pwaService: PwaService,
     private readonly zone: NgZone,
     private readonly csvService: CsvService,
     readonly i18n: I18nService,
@@ -131,19 +130,16 @@ export class ImportQuestionsDialogComponent extends SubscriptionDisposable {
       this.updateSelectAllCheckbox();
     });
 
-    this.promiseForTransceleratorQuestions = this.pwaService.online
-      .then(() => {
-        this.showTransceleratorOfflineMsg = false;
-        return projectService.transceleratorQuestions(this.data.projectId);
-      })
-      .catch(err => {
-        if (/Transcelerator version unsupported/.test(err.message)) {
-          this.transceleratorOutdated = true;
-          return [];
-        } else {
-          throw err;
-        }
-      });
+    this.transceleratorRequest = projectService.transceleratorQuestions(this.data.projectId, this.ngUnsubscribe);
+
+    this.promiseForTransceleratorQuestions = this.transceleratorRequest.promiseForResult.catch((error: unknown) => {
+      if (typeof error === 'object' && /Transcelerator version unsupported/.test(error?.['message'])) {
+        this.transceleratorOutdated = true;
+        return [];
+      } else {
+        throw error;
+      }
+    });
 
     this.promiseForQuestionDocQuery = projectService.queryQuestions(this.data.projectId);
   }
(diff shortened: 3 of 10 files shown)


RaymondLuong3 (changes requested):
Nice work. I have a few suggestions for this PR.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-:-MuX7y8eBaqU71XEDyO-:beylbw9)* status: all files reviewed, 8 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 35 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuX8Gbb8Mp3ZgVIj4s7:-MuX8Gbb8Mp3ZgVIj4s8:b8yh0mw) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L35)):*
> ```HTML
>                 color="primary"
>                 (click)="importFromTranscelerator()"
>                 [disabled]="transceleratorRequest.status !== 'complete'"
> ```

Is there a reason that this component does not follow the normal convention where we do not disable a button, but will show the error message when the user click the Import button. It might be worth it to hide the error message until the user clicks the import button. It would be more intuitive if the error about retrying appears when the user has made an attempt to import from transcelerator.



_Code quote:_
```HTML
         [disabled]="transceleratorRequest.status !== 'complete'"
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 119 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWhaUF-9LrP7tFPOiT:-MuWhaUF-9LrP7tFPOiU:b-j3arai) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L119)):*
> ```JSON
>     "loading": "Loading...",
>     "must_be_valid_reference": "Must be a valid reference",
>     "network_error_transcelerator": "Network error fetching Transcelerator questions. Retrying ({{ count }}).",
> ```

Would it be confusing to users that they get an error fetching questions even before the click import from transcelerator? Perhaps just a simple "There has been a network error. Retrying,"



_Code quote:_
```JSON
    "network_error_transcelerator": "Network error fetching Transcelerator questions. Retrying ({{ count }}).",
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts, line 45 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWoFpi7bVksT5fYKvd:-MuWoFpi7bVksT5fYKve:bheu7w) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts#L45)):*
> ```XML
> }
> 
> export class TestingRetryingRequestService {
> ```

This should probably be in a separate file. Or it could go into the spec file.



_Code quote:_
```XML
export class TestingRetryingRequestService {
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWupEt6VSdEWM-C8gu:-MuWupEt6VSdEWM-C8gv:bxdqr32) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts#L68)):*
> ```XML
>    * Does not throw on network errors.
>    */
>   complete: Promise<T | undefined>;
> ```

How about changing this to be a method. It would be easier to understand that this is used to execute the request and wait for a response. complete() or completeRequest() would be appropriate names.



_Code quote:_
```XML
  complete: Promise<T | undefined>;
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts, line 114 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuX0tiF0sAbVHqjV6gM:-MuX0tiF0sAbVHqjV6gN:b-ly5f96) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts#L114)):*
> ```XML
>     while (!this.canceled && this.status !== 'complete') {
>       const online = await this.online$.pipe(take(1)).toPromise();
>       if (online === false) {
> ```

Nit: This could check !online or online !== true.



_Code quote:_
```XML
      if (online === false) {
        this.status = 'offline';
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-requestion.service.spec.ts, line 1 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWu1ecEYk_mp9lTH7S:-MuWu1eeAlahMINqnPVb:boui8cg) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-requestion.service.spec.ts#L1)):*
> ```XML
> import { fakeAsync, tick } from '@angular/core/testing';
> ```

That's a funny name for this spec file ;)
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-requestion.service.spec.ts, line 82 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWxm6P6aITW6KbkG72:-MuWxm6QCzlismW58d7s:bwcjkf2) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-requestion.service.spec.ts#L82)):*
> ```XML
>     tick(RETRY_DELAY + NEGLIGIBLE_AMOUNT_OF_TIME);
>     expect(request.failedAttempts).toBe(2);
>     expect(request.status).toBe('complete');
> ```

That is great that the request will continue if a user goes offline, and then comes back online.



_Code quote:_
```XML
    expect(request.status).toBe('complete');
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-requestion.service.spec.ts, line 117 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWyN4o9NjlkfenWujV:-MuWyN4o9NjlkfenWujW:b-oai0lu) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-requestion.service.spec.ts#L117)):*
> ```XML
> 
>     tick(RETRY_DELAY - NEGLIGIBLE_AMOUNT_OF_TIME);
>     expect(request.failedAttempts).toBe(1);
> ```

I would have expected failedAttempts to be 2. What am I missing?



_Code quote:_
```XML
 expect(request.failedAttempts).toBe(1);
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1230)* status: 7 of 9 files reviewed, 7 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 35 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuX8Gbb8Mp3ZgVIj4s7:-MuXQiVv7ykYLG5n3tI9:b-f3q359) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L35)):*

Is there a reason that this component does not follow the normal convention where we do not disable a button, but will show the error message when the user click the Import button. It might be worth it to hide the error message until the user clicks the import button. It would be more intuitive if the error about retrying appears when the user has made an attempt to import from transcelerator.
</blockquote></details>

Hmm. I think we follow that convention for form validation. We don't complain to a user that an input is invalid when we haven't given the user the opportunity to finish filling out the form, because that's annoying. We also don't disable the submit button, because that's annoying.

We haven't been handling offline issues that way though. See for example the sync page:


We disable what's unavailable and explain why it's disabled. That said, I can see how it's a tad odd to tell the user there was a network error before they've clicked the button:



It seems like it would be inconsistent with how we do things elsewhere to not show it though. Or maybe the offline message should show immediately, but the "network error" message should only show once the user has clicked the button? If we do that though, then what happens once the request succeeds after the user has already clicked the button?

In general this message should be the exception and not the rule, since it only happens when the user has an established connection, but for some reason the request fails anyway.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts, line 45 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWoFpi7bVksT5fYKvd:-MuXNLkf3VdnAoJD97lW:bcwtxwf) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts#L45)):*

This should probably be in a separate file. Or it could go into the spec file.
</blockquote></details>

It doesn't belong it the spec file, because it's not just for testing the `RetryingRequest` class. It's for any component to use in its tests. Kind of like how we have a `MemoryRealtimeRemoteStore` and a `SharedbRealtimeRemoteStore` that both extend `RealtimeRemoteStore`. In our production code we use `SharedbRealtimeRemoteStore`, but for our tests we use `MemoryRealtimeRemoteStore`. Given that, do you think it should be in this file or a separate one?
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWupEt6VSdEWM-C8gu:-MuXQKPc88e7J9KCTuqi:b20ijot) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts#L68)):*

How about changing this to be a method. It would be easier to understand that this is used to execute the request and wait for a response. complete() or completeRequest() would be appropriate names.
</blockquote></details>

I think perhaps it should be renamed, but I don't understand how it could be a method. The request occurs regardless of whether this promise is awaited. If it is awaited, you can use it to get the result of the request, and catch any non-network errors.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts, line 114 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuX0tiF0sAbVHqjV6gM:-MuXNHpJEVqtCK4URYye:b-qu6n94) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts#L114)):*

Nit: This could check !online or online !== true.
</blockquote></details>

Good point. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-requestion.service.spec.ts, line 1 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWu1ecEYk_mp9lTH7S:-MuXMMEc0u8Rl-sHUL-h:b-dkd6u8) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-requestion.service.spec.ts#L1)):*

That's a funny name for this spec file ;)
</blockquote></details>

Oops, good catch. Fixed.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-requestion.service.spec.ts, line 117 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWyN4o9NjlkfenWujV:-MuXMjr89jsM8mSz2AUk:b-g7fhvx) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-requestion.service.spec.ts#L117)):*

I would have expected failedAttempts to be 2. What am I missing?
</blockquote></details>

With the first `tick()`, almost no time passed (let's say 1ms at most). Then time moves forward by just under 3 seconds (the current retry time) when we do `tick(RETRY_DELAY - NEGLIGIBLE_AMOUNT_OF_TIME)`. At this point the second request should not have been sent yet, because the time is not quite up. And then we cancel the request, right before it tries a second time.

RaymondLuong3 (approved):
Nice work. I have a couple remaining suggestions, but nothing blocking.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-:-MulQirD9myMHA4w_4iW:bxjuoun)* status: all files reviewed, 4 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html, line 35 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuX8Gbb8Mp3ZgVIj4s7:-MulP_hR68l97SReNO6K:bxrv78i) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L35)):*

Hmm. I think we follow that convention for form validation. We don't complain to a user that an input is invalid when we haven't given the user the opportunity to finish filling out the form, because that's annoying. We also don't disable the submit button, because that's annoying.

We haven't been handling offline issues that way though. See for example the sync page:


We disable what's unavailable and explain why it's disabled. That said, I can see how it's a tad odd to tell the user there was a network error before they've clicked the button:



It seems like it would be inconsistent with how we do things elsewhere to not show it though. Or maybe the offline message should show immediately, but the "network error" message should only show once the user has clicked the button? If we do that though, then what happens once the request succeeds after the user has already clicked the button?

In general this message should be the exception and not the rule, since it only happens when the user has an established connection, but for some reason the request fails anyway.
</blockquote></details>

That is a good point that this dialog should be distinct from a form. And it is true that our other pages that need an internet connection immediately notify the user that some features are unavailable offline. So this is ok with me.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts, line 45 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWoFpi7bVksT5fYKvd:-MulA5V96mO3bHvgT2Vg:bt0rjxv) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts#L45)):*

It doesn't belong it the spec file, because it's not just for testing the `RetryingRequest` class. It's for any component to use in its tests. Kind of like how we have a `MemoryRealtimeRemoteStore` and a `SharedbRealtimeRemoteStore` that both extend `RealtimeRemoteStore`. In our production code we use `SharedbRealtimeRemoteStore`, but for our tests we use `MemoryRealtimeRemoteStore`. Given that, do you think it should be in this file or a separate one?
</blockquote></details>

Ok I see that now. I am in favour of putting it in a new file since the RetryingRequestService if fully functional without this TestingRetryingRequestService class.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWupEt6VSdEWM-C8gu:-MulNNH3CBML9vARoRjH:b-ic2eha) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts#L68)):*

I think perhaps it should be renamed, but I don't understand how it could be a method. The request occurs regardless of whether this promise is awaited. If it is awaited, you can use it to get the result of the request, and catch any non-network errors.
</blockquote></details>

I was thinking of using a method so that when you call complete() or getResult() and await the result, it will be obvious that you intend to get the result of the network request. You are right in that it is important to have a way to catch the error. You would still be able to add a .catch() error handler to the method. However, if it still makes more sense to use a property to await, I'm ok with that too.
FYI, the indexeddb-offline-store.ts lets you assign a listener to the different events like onsuccess or onerror. It all gets wrapped in a promise that can be awaited by calling openDB().
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-requestion.service.spec.ts, line 117 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWyN4o9NjlkfenWujV:-MulPR005N-bOJPQe0dm:b-uzqs7q) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-requestion.service.spec.ts#L117)):*

With the first `tick()`, almost no time passed (let's say 1ms at most). Then time moves forward by just under 3 seconds (the current retry time) when we do `tick(RETRY_DELAY - NEGLIGIBLE_AMOUNT_OF_TIME)`. At this point the second request should not have been sent yet, because the time is not quite up. And then we cancel the request, right before it tries a second time.
</blockquote></details>

Right, I missed that minus in the tick right above this.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1230)* status: 6 of 10 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 119 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWhaUF-9LrP7tFPOiT:-Muq5NATASSqXATZNzVt:b-n3a219) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L119)):*

Would it be confusing to users that they get an error fetching questions even before the click import from transcelerator? Perhaps just a simple "There has been a network error. Retrying,"
</blockquote></details>

Is this distinct from the conversation over on `import-questions-dialog.component.html`? I read the two as being the same, but I'm not sure.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts, line 45 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWoFpi7bVksT5fYKvd:-MuqxCWM91owtwosDBL7:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts#L45)):*

Ok I see that now. I am in favour of putting it in a new file since the RetryingRequestService if fully functional without this TestingRetryingRequestService class.
</blockquote></details>

Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWupEt6VSdEWM-C8gu:-MurJk0k-kF9Mo3HeDn9:b-c43qfd) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts#L68)):*

I was thinking of using a method so that when you call complete() or getResult() and await the result, it will be obvious that you intend to get the result of the network request. You are right in that it is important to have a way to catch the error. You would still be able to add a .catch() error handler to the method. However, if it still makes more sense to use a property to await, I'm ok with that too.
FYI, the indexeddb-offline-store.ts lets you assign a listener to the different events like onsuccess or onerror. It all gets wrapped in a promise that can be awaited by calling openDB().
</blockquote></details>

I don't really understand how making it a method makes a difference. If anything I would be concerned that it would give the incorrect idea that calling it causes the request to get sent. Maybe I'm missing something?

Maybe it should be called `promiseForResult`? I called it `complete` because it can also be used just to find out when the request is complete, and it's not the only way to get the result. My hope is that most components will not even need to await anything, and just read the properties of the request in the component's template.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1230)* status: 7 of 10 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1230#-MuWupEt6VSdEWM-C8gu:-MuvzXls7cXwe4L8EW-O:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/3258e9f0aa93d1707beea7e6f1fbb0d3a784ace7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/retrying-request.service.ts#L68)):*

I just mean that I'm not so familiar with how network requests are typically handled, and that I have not seen a property used in the way that `complete` is used. It could be that awaiting a property called `complete` doesn't fit my thinking, even though I know what it does. I do prefer the name `promiseForResult`, it is similar to the property `tryLoginPromise` in `auth.service`.
</blockquote></details>

Done.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1230)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1233, opened 2022-02-01
drop: simplify blank and specify when sibling

marksvc:
---
This PR is some followup after doing SF-1454.
I prefer for this PR to have its individual commits retained.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1233)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
@@ -25,12 +25,9 @@ export class DragAndDrop {
       // Stop the browser from doing any drag-and-drop behaviour itself, such as inserting text with formatting.
       dragEvent.preventDefault();
 
-      const [targetUsxSegmentElement, targetElement, droppingIntoBlankSegment]: [
-        Element | null,
-        Element | null,
-        boolean
-      ] = this.discernTarget(dragEvent);
-      if (targetUsxSegmentElement == null || targetElement == null) {
+      const [targetUsxSegmentElement, droppingIntoBlankSegment]: [Element | null, boolean] =
+        this.discernTarget(dragEvent);
+      if (targetUsxSegmentElement == null) {
         return;
       }
 
@@ -61,18 +58,15 @@ export class DragAndDrop {
         return;
       }
 
-      const insPos: number | undefined = this.determineInsertionPosition(
+      const insPosInSegment: number | undefined = this.getInsertionPositionInSegment(
         targetUsxSegmentElement,
-        targetElement,
-        destinationSegmentRange,
-        droppingIntoBlankSegment,
         dragEvent.clientX,
         dragEvent.clientY
       );
-      if (insPos == null) {
+      if (insPosInSegment == null) {
         return;
       }
-      let insertionPositionInDocument: number = insPos;
+      let insertionPositionInDocument: number = destinationSegmentRange.index + insPosInSegment;
 
       // Identify the selection range, if any. This will be a
       // selection that was already present before the drop, whether it is the text that was dragged, or a selection
@@ -113,7 +107,7 @@ export class DragAndDrop {
           quill.insertText(insertionPositionInDocument, newText, 'user');
           setTimeout(() => {
             // If we inserted into a blank segment, and let SF respond by removing the blank in between our quill
-            // changes, then the position of insertion needs to move back by 1.
+            // changes, then the position needs to move back by 1 to select the inserted text.
             const lengthOfBlank: number = 1;
             if (droppingIntoBlankSegment) {
               insertionPositionInDocument -= lengthOfBlank;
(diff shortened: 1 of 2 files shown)


RaymondLuong3 (changes requested):
Nice work, and perfect text coverage. I just had one minor blocking comment.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1233#-:-MvPiWy-8c8jmeO37n1N:bqr35ap)* status: all files reviewed, 4 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts, line 161 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1233#-MvPdvbD03yl0VuDtSd1:-MvPdvbD03yl0VuDtSd2:b7n5wia) ([raw file](https://github.com/sillsdev/web-xforge/blob/4de721fbbfeea43dd739c832c04d814a5326b524/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts#L161)):*
> ```XML
> 
>   /** Return the editor position in the segment where we want to insert text from a drop. */
>   private determineInsertionPosition(
> ```

Another name could be getInsertionPositionInSegment(), but the method comment makes this work.



_Code quote:_
```XML
  private determineInsertionPosition(
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1688 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1233#-MvPiOQj7t06DsUMIMke:-MvPiOQj7t06DsUMIMkf:bdf1xuk) ([raw file](https://github.com/sillsdev/web-xforge/blob/4de721fbbfeea43dd739c832c04d814a5326b524/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1688)):*
> ```XML
>       // (both with the same data-segment segment ref value). This test is for the case of the
>       // display-text-anchor and usx-blank being in the same usx-segment element. The resulting DOM after
>       // the drop will be for now two usx-segment elements, one with the display-text-anchor and one with the usx-blank.
> ```

The result does not contain a usx-blank as far as I am aware right?



_Code quote:_
```XML
one with the display-text-anchor and one with the usx-blank.
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1777 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1233#-MvPg-xf14ODmMBwpZit:-MvPg-xf14ODmMBwpZiu:b-61gqje) ([raw file](https://github.com/sillsdev/web-xforge/blob/4de721fbbfeea43dd739c832c04d814a5326b524/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1777)):*
> ```XML
>       const env = new TestEnvironment();
>       const textDocId: TextDocId = new TextDocId('project01', 40, chapterNum);
>       const delta = new Delta(textDoc);
> ```

Nit: This could be moved into a helper method to clean this test up a bit.



_Code quote:_
```XML
      const delta = new Delta(textDoc);
      env.realtimeService.addSnapshot<TextData>(TextDoc.COLLECTION, {
        id: textDocId.toString(),
        data: delta,
        type: RichText.type.name
      });
      env.fixture.detectChanges();
      env.id = textDocId;
      tick();
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1892 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1233#-MvPh1QW6-vAExuVoN7M:-MvPh1QW6-vAExuVoN7N:bb7xrsy) ([raw file](https://github.com/sillsdev/web-xforge/blob/4de721fbbfeea43dd739c832c04d814a5326b524/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1892)):*
> ```XML
>     /** Helper method to perform a drag-and-drop and check expectations on resulting data, elements, and
>      *  editor selection. */
>     function performDropTest({
> ```

Nit: This could have been made a method in the TestEnvironment rather than on it's own where it must have the test environment passed into it.



_Code quote:_
```XML
  function performDropTest({
```

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1233)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1688 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1233#-MvPiOQj7t06DsUMIMke:-MvQ51er6VUzzdZB8Dfu:bkobfw7) ([raw file](https://github.com/sillsdev/web-xforge/blob/4de721fbbfeea43dd739c832c04d814a5326b524/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1688)):*

The result does not contain a usx-blank as far as I am aware right?
</blockquote></details>

It does, but in another usx-segment element. See down below where we have:
```ts
      // It would make sense to here expect `['display-text-anchor', '#text']`, but initially, the target usx-segment
      // element will be left with the display-text-anchor element alone, and a new usx-segment will be
      // created with the text node.
      const expectedTopLevelNodeSeriesAfterEvent: string[] = ['display-text-anchor'];
```
The `performDropTest()` nicely lets me assert the nodes in a target usx-segment both before and after a drop occurs. But I didn't make it flexible enough to assert behaviour regarding usx-segments being split apart.

RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1233)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1688 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1233#-MvPiOQj7t06DsUMIMke:-MvQVstr4H1JHUQ8ny4c:b-hkxdzm) ([raw file](https://github.com/sillsdev/web-xforge/blob/4de721fbbfeea43dd739c832c04d814a5326b524/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1688)):*

It does, but in another usx-segment element. See down below where we have:
```ts
      // It would make sense to here expect `['display-text-anchor', '#text']`, but initially, the target usx-segment
      // element will be left with the display-text-anchor element alone, and a new usx-segment will be
      // created with the text node.
      const expectedTopLevelNodeSeriesAfterEvent: string[] = ['display-text-anchor'];
```
The `performDropTest()` nicely lets me assert the nodes in a target usx-segment both before and after a drop occurs. But I didn't make it flexible enough to assert behaviour regarding usx-segments being split apart.
</blockquote></details>

That doesn't sound right. If you drag text to a blank segment, that usx-blank really should not be there after the fact. Unless you mean that the blank is there momentarily before the text-view-model cleans it up.

marksvc:
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1688 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1233#-MvPiOQj7t06DsUMIMke:-MvVQIvP5Kiw9VnB2N2K:bg8xbys) ([raw file](https://github.com/sillsdev/web-xforge/blob/4de721fbbfeea43dd739c832c04d814a5326b524/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1688)):*

That doesn't sound right. If you drag text to a blank segment, that usx-blank really should not be there after the fact. Unless you mean that the blank is there momentarily before the text-view-model cleans it up.
</blockquote></details>

Thank you. I was not thinking. It's a _text node_ that is in the other usx-segment element. I have updated the comment. I also tested in the app, and during the unit tests, to confirm this is true.

RaymondLuong3 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1233#-:-MvVoLIP9yKox3u1X_9P:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1234, opened 2022-02-01
drop: expect and hide console logs in tests

marksvc:
- @triviality/logger is not here used in a way that requires it in
production, and so it is added only to the development set of
dependencies.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1234)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/package-lock.json
+++ b/src/SIL.XForge.Scripture/ClientApp/package-lock.json
@@ -68,6 +68,7 @@
         "@angular/compiler": "~11.2.14",
         "@angular/compiler-cli": "~11.2.14",
         "@angular/language-service": "~11.2.14",
+        "@triviality/logger": "^1.1.6",
         "@types/auth0-js": "^9.14.5",
         "@types/jasmine": "~3.10.3",
         "@types/jasminewd2": "~2.0.10",
@@ -3793,6 +3794,29 @@
         "node": ">= 6"
       }
     },
+    "node_modules/@triviality/core": {
+      "version": "1.2.5",
+      "resolved": "https://registry.npmjs.org/@triviality/core/-/core-1.2.5.tgz",
+      "integrity": "sha512-gON9hw952+91mmHS5WKLsa0iGiyAV1t/1rPVAVIgeIu/QzuQ3/Jpk+8Wac963x0mnCWJ89BDg3xiNcEm2weGOg==",
+      "dev": true,
+      "dependencies": {
+        "@types/lodash": "^4.14.149",
+        "@types/ramda": "0",
+        "lodash": "^4.17.15",
+        "ramda": "0"
+      }
+    },
+    "node_modules/@triviality/logger": {
+      "version": "1.1.6",
+      "resolved": "https://registry.npmjs.org/@triviality/logger/-/logger-1.1.6.tgz",
+      "integrity": "sha512-eImRpBT4auMklBiCbUCLnzRiuf1gTYCj7zsRKircN4T6DjnA/mCLoBqLyUaO9ldtBsMNoCFp4lRZZxfc4VzBaw==",
+      "dev": true,
+      "dependencies": {
+        "@triviality/core": "^1.2.5",
+        "moment": "2",
+        "ts-log": "2"
+      }
+    },
     "node_modules/@trysound/sax": {
       "version": "0.1.1",
       "resolved": "https://registry.npmjs.org/@trysound/sax/-/sax-0.1.1.tgz",
@@ -3923,6 +3947,15 @@
       "dev": true,
       "dependencies": {
         "parchment": "^1.1.2"
+      }
+    },
+    "node_modules/@types/ramda": {
+      "version": "0.27.64",
+      "resolved": "https://registry.npmjs.org/@types/ramda/-/ramda-0.27.64.tgz",
+      "integrity": "sha512-EDf++ss/JoMiDpvT1MuA8oi88OwpvmqVE+o8Ojm5v/5bdJEPZ6eIQd/XYAeQ0imlwG6Tf0Npfq4Z9w3hAKBk9Q==",
+      "dev": true,
+      "dependencies": {
+        "ts-toolbelt": "^6.15.1"
       }
     },
     "node_modules/@types/selenium-webdriver": {
@@ -13068,6 +13101,15 @@
         "obliterator": "^1.5.0"
       }
     },
+    "node_modules/moment": {
+      "version": "2.29.1",
+      "resolved": "https://registry.npmjs.org/moment/-/moment-2.29.1.tgz",
+      "integrity": "sha512-kHmoybcPV8Sqy59DwNDY3Jefr64lK/by/da0ViFcuA4DH0vQg5Q6Ze5VimxkfQNSC+Mls/Kx53s7TjP1RhFEDQ==",
+      "dev": true,
+      "engines": {
+        "node": "*"
+      }
+    },
     "node_modules/move-concurrently": {
       "version": "1.0.1",
       "resolved": "https://registry.npmjs.org/move-concurrently/-/move-concurrently-1.0.1.tgz",
@@ -16207,11 +16249,6 @@
         "safer-buffer": "^2.0.2",
         "tweetnacl": "~0.14.0"
       },
-      "bin": {
-        "sshpk-conv": "bin/sshpk-conv",
-        "sshpk-sign": "bin/sshpk-sign",
-        "sshpk-verify": "bin/sshpk-verify"
-      },
       "engines": {
         "node": ">=0.10.0"
       }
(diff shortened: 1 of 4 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1234)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @marksvc and @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1837 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1234#-Mux8AMn7WnOfQbNGsdO:-Mux8AMn7WnOfQbNGsdP:b1tpulg) ([raw file](https://github.com/sillsdev/web-xforge/blob/9396afa67594dd8f4ad1909eeed5c42067910a1c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1837)):*
> ```XML
>         document.caretRangeFromPoint = (_x: number, _y: number) => ({ startOffset: 0 } as Range);
>       });
>       mockedConsole.verify();
> ```

Stating an expectation at the beginning and then calling `verify()` a the end is kind of odd. Is there a reason you couldn't just call `shouldOnlyGet()` at the end?
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/mock-console.ts, line 28 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1234#-MuwkDZa412HtU_4XtTs:-MuwkDZa412HtU_4XtTt:bgdhmfe) ([raw file](https://github.com/sillsdev/web-xforge/blob/9396afa67594dd8f4ad1909eeed5c42067910a1c/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/mock-console.ts#L28)):*
> ```XML
>   static install(): MockConsole {
>     const instance: MockConsole = new this();
>     (console as any) = instance;
> ```

I'm not really a fan of redefining a global variable. It's not really a big deal though since it's just the tests. But it looks to me like deliberately avoiding something we *should* be doing, namely, having a logger class that is used by each component, rather than having each component directly access a global variable. It seems like a perfect place to use dependency injection. And right now we're already dependency injecting the console in a few places, and it's *super ugly*:
``` ts
// constructor argument
@Inject(CONSOLE) private readonly console: ConsoleInterface
```
If we instead created a LoggingService class that has the console internally, then it would make dependency injecting it easier (presumably we could ditch the `@Inject(CONSOLE)` decorator that looks so out of place). The tests could then provide their own class (MockConsole) that would have the same interface, but would provide helper methods for verifying what was logged. There are only a few classes that actually log to the console, and a couple of them are already injecting the console.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1234)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1837 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1234#-Mux8AMn7WnOfQbNGsdO:-Mv-xbpBBYa_xd13-xzO:biamobc) ([raw file](https://github.com/sillsdev/web-xforge/blob/9396afa67594dd8f4ad1909eeed5c42067910a1c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1837)):*

Stating an expectation at the beginning and then calling `verify()` a the end is kind of odd. Is there a reason you couldn't just call `shouldOnlyGet()` at the end?
</blockquote></details>

Sure, and specifically there are methods in the mock-console.ts for that like `assertHasOnly(expectedMessage)`. A reason to write the expectation first and verify it after the SUT is because anything not expected or muted is shown. If you run `shouldOnlyGet('file not found')` then when the SUT runs it will mute any matching messages. 

I did make provision for muting arbitrary messages (`.muteFor(pattern)`). Using `shouldOnlyGet(pattern)` lets you define an expectation and cause muting for a pattern in one step. 
___
*[src/SIL.XForge.Scripture/ClientApp/src/xforge-common/mock-console.ts, line 28 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1234#-MuwkDZa412HtU_4XtTs:-Mv-yUR0Asg7RyNHPx5S:b-vykm6x) ([raw file](https://github.com/sillsdev/web-xforge/blob/9396afa67594dd8f4ad1909eeed5c42067910a1c/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/mock-console.ts#L28)):*

I'm not really a fan of redefining a global variable. It's not really a big deal though since it's just the tests. But it looks to me like deliberately avoiding something we *should* be doing, namely, having a logger class that is used by each component, rather than having each component directly access a global variable. It seems like a perfect place to use dependency injection. And right now we're already dependency injecting the console in a few places, and it's *super ugly*:
``` ts
// constructor argument
@Inject(CONSOLE) private readonly console: ConsoleInterface
```
If we instead created a LoggingService class that has the console internally, then it would make dependency injecting it easier (presumably we could ditch the `@Inject(CONSOLE)` decorator that looks so out of place). The tests could then provide their own class (MockConsole) that would have the same interface, but would provide helper methods for verifying what was logged. There are only a few classes that actually log to the console, and a couple of them are already injecting the console.
</blockquote></details>

Yeah, we could inject something. I was injecting something like that at first until I found that I could just overwrite the global console and leave the application class unchanged. I like that overriding the console is very _clean_ for the application class, since it's transparent.

If we do inject a console/logger, then we will have two consoles: One that we are supposed to log to (injected), and one that we might quickly log to for debugging but need to remember not to use for anything else (`console.log`). We can do that. 

I remember that we've talked before about rolling console messages into automated bug reports, so whatever console/logger is injected in production would presumably be something that would play into that.

Nateowami (commented):
> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1234)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1837 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1234#-Mux8AMn7WnOfQbNGsdO:-Mv01pzBAYslIxzMVIkY:b-sw1zpy) ([raw file](https://github.com/sillsdev/web-xforge/blob/9396afa67594dd8f4ad1909eeed5c42067910a1c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1837)):*

Sure, and specifically there are methods in the mock-console.ts for that like `assertHasOnly(expectedMessage)`. A reason to write the expectation first and verify it after the SUT is because anything not expected or muted is shown. If you run `shouldOnlyGet('file not found')` then when the SUT runs it will mute any matching messages. 

I did make provision for muting arbitrary messages (`.muteFor(pattern)`). Using `shouldOnlyGet(pattern)` lets you define an expectation and cause muting for a pattern in one step. 
</blockquote></details>

Ah, that makes sense. Maybe that behavior would be more obvious if the method was called something like "expectAndHide"?

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1234)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts, line 1837 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1234#-Mux8AMn7WnOfQbNGsdO:-MxBh7-U-dMYf5Jjt-ia:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/9396afa67594dd8f4ad1909eeed5c42067910a1c/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L1837)):*

Ah, that makes sense. Maybe that behavior would be more obvious if the method was called something like "expectAndHide"?
</blockquote></details>

Done.

Nateowami (commented):
@marksvc Sorry I've taken so long to return to this PR. Could you update it to resolve conflicts?

> Reviewed 2 of 4 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1234#-:-MzIbeMV-y4vDG7p_Y9B:b-wmi290)* status: 2 of 4 files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami (approved):
:lgtm:

> Reviewed 1 of 2 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1234#-:-MzwkNPT2hucjfO0dKoa:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

Nateowami:
Wait, what, this was a PR against develop? I never noticed that. Should it have been changed to master?

PR #1235, opened 2022-02-02
Add script to modify doc, even if it's corrupted

Nateowami:
If a doc is corrupted in such a way that it's not a valid doc, then it's not possible to convert it into a valid doc by applying an op.

This script gets around that problem by deleting the doc and recreating it. It's still the same doc though; the version numbers keep climbing. Basically there's an op that says "delete everything" and then another op that says "create it in this state."

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1235)

### DIFF
--- a/scripts/db_tools/fix-doc.js
+++ b/scripts/db_tools/fix-doc.js
@@ -1,0 +1,51 @@
+#!/usr/bin/env node
+
+// Delete a doc (though it still exists in the DB), then recreate it in another state. This is useful for when a
+// document has been corrupted in such a way that it cannot be fixed by just applying a operation to it. This script is
+// intended to be edited to be used as needed.
+
+const utils = require('./utils.js');
+const RichText = utils.requireFromRealTimeServer('rich-text');
+const ShareDB = utils.requireFromRealTimeServer('sharedb/lib/client');
+const WebSocket = utils.requireFromRealTimeServer('ws');
+
+// Edit these settings to specify which doc to revert
+const docId = '';
+const connectionConfig = utils.devConfig;
+const collection = 'texts';
+const dryRun = true;
+
+ShareDB.types.register(RichText.type);
+
+async function run() {
+  const ws = new WebSocket(connectionConfig.wsConnectionString);
+  const conn = new ShareDB.Connection(ws);
+  try {
+    const doc = conn.get(collection, docId);
+    await utils.fetchDoc(doc);
+
+    const newOps = cleanupOps(doc.data.ops);
+
+    if (!dryRun) {
+      await utils.deleteDoc(doc);
+      console.log('deleted doc');
+    }
+
+    if (dryRun) {
+      console.log('would set doc to:');
+      utils.visualizeOps(newOps, true);
+    } else {
+      await utils.createDoc(doc, newOps, RichText.type.name);
+      console.log('recreated doc');
+    }
+  } finally {
+    conn.close();
+  }
+}
+
+// Edit this function to modify the ops to whatever state they should be in
+function cleanupOps(ops) {
+  return ops.filter(op => op.insert != null && ['object', 'string'].includes(typeof op.insert));
+}
+
+run();


marksvc:
:eyes:

marksvc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1235)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[scripts/db_tools/fix-doc.js, line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1235#-Muw_INc608zKTA68iYs:-Muw_INc608zKTA68iYt:borw68) ([raw file](https://github.com/sillsdev/web-xforge/blob/cd07a22b3638955d4660baa4687e21e1263a42ef/scripts/db_tools/fix-doc.js#L3)):*
> ```JavaScript
> #!/usr/bin/env node
> 
> // Delete a doc (though it still exists in the DB), then recreate it in another state. This is useful for when a
> ```

What's meant by "in another state"? Would it be true to say "then recreate it from just the insert ops." ?



_Code quote:_
```JavaScript
then recreate it in another state.
```
___
*[scripts/db_tools/fix-doc.js, line 45 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1235#-MuwZbHV7XJ5Hq2xs8JD:-MuwZbHV7XJ5Hq2xs8JE:bxhszw6) ([raw file](https://github.com/sillsdev/web-xforge/blob/cd07a22b3638955d4660baa4687e21e1263a42ef/scripts/db_tools/fix-doc.js#L45)):*
> ```JavaScript
> }
> 
> function fixDoc(doc) {
> ```

I expected `fixDoc()` to _do_ something. Maybe a better name might be something like `getInsertOps()`? 
___
*[scripts/db_tools/fix-doc.js, line 46 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1235#-Muw_Uqa3uxfmA46jqRZ:-Muw_Uqa3uxfmA46jqR_:b8yejoq) ([raw file](https://github.com/sillsdev/web-xforge/blob/cd07a22b3638955d4660baa4687e21e1263a42ef/scripts/db_tools/fix-doc.js#L46)):*
> ```JavaScript
> 
> function fixDoc(doc) {
>   return doc.data.ops.filter(op => typeof op.insert !== 'undefined');
> ```

BTW I this will include unpleasant situations like 
```json
{
  insert: null
}
```
What's the thinking behind using `typeof op.insert !== 'undefined'` rather than something like `op.insert != null` or `op.insert != undefined`? Maybe `typeof ....` is _the normal way_ to check if an object has a field?

Nateowami (commented):
Thanks for looking at this. I've made some changes. The purpose of this script is not to "remove all non-insert ops". The purpose is to "apply whatever function is needed to a doc, and then have it deleted and recreated in the new state." In other words, it's general purpose. The script as currently written is what I thought would fix our data corruption issue, but due to the unfortunate loss of verse numbers, we had to do something else.

The way I see this script being used is that we see a problem and want to update a text doc. We can't just edit it, because it's a realtime doc. If it's not corrupted, we should use a different method, that will apply an op. If it is corrupted, and we want to set it to a previous version, revert-doc.js can do that. If we want to apply a function to it, that's what this script is for.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1235#-:-MuwhOzi25OO7NqGZsnN:bcqb88y)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @marksvc)

___
*[scripts/db_tools/fix-doc.js, line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1235#-Muw_INc608zKTA68iYs:-Muwd2sqBZ8h6tz2OrzK:b5y6po7) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b96e6d0ab9560ed8c9b9069237b83fa18348f0b/scripts/db_tools/fix-doc.js#L3)):*

What's meant by "in another state"? Would it be true to say "then recreate it from just the insert ops." ?
</blockquote></details>

Maybe this should be reworeded to be clearer? I've expanded the comment a bit.
___
*[scripts/db_tools/fix-doc.js, line 45 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1235#-MuwZbHV7XJ5Hq2xs8JD:-Muwio0d2--R5IPzGrRB:bh167g5) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b96e6d0ab9560ed8c9b9069237b83fa18348f0b/scripts/db_tools/fix-doc.js#L45)):*

I expected `fixDoc()` to _do_ something. Maybe a better name might be something like `getInsertOps()`? 
</blockquote></details>

I've renamed this to `cleanupOps()`. What do you think?
___
*[scripts/db_tools/fix-doc.js, line 46 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1235#-Muw_Uqa3uxfmA46jqRZ:-MuwdMG_7e-3pAGCsNoZ:b-xyo5hy) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b96e6d0ab9560ed8c9b9069237b83fa18348f0b/scripts/db_tools/fix-doc.js#L46)):*
Good point. Using `typeof` is an old way of checking for undefined; I forgot we now just use the non-strict inequality operator to compare it to null. I got in the habit because older JavaScript engines actually allowed defining undefined, so if undefined had been set to "pizza", and then you did `value !== undefined` you would actually be checking whether the value was "pizza". I'm guessing this wasn't usually a problem, but it's such a horrendous failure mode that the accepted method of checking whether something is undefined was to check whether `typeof something === 'undefined'`. Another way is `var x; console.log(value === x)`, since an uninitialized variable will always be undefined.

Maybe the best option would be `typeof op.insert != null && ['object', 'string'].includes(typeof op.insert)`?

Actually though, this function was intended to be an *example* function that the user of the script could easily redefine to clean up the doc however. I've added a comment.

> Maybe `typeof ....` is *the normal way* to check if an object has a field?

I think the way you check for a field is `'insert' in {insert: null}`, which will evaluate to `true`.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1235)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @marksvc)

___
*[scripts/db_tools/fix-doc.js, line 46 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1235#-Muw_Uqa3uxfmA46jqRZ:-MuwoXGmA-AO0Se2a6xY:btam1y1) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b96e6d0ab9560ed8c9b9069237b83fa18348f0b/scripts/db_tools/fix-doc.js#L46)):*

Good point. Using `typeof` is an old way of checking for undefined; I forgot we now just use the non-strict inequality operator to compare it to null. I got in the habit because older JavaScript engines actually allowed defining undefined, so if undefined had been set to "pizza", and then you did `value !== undefined` you would actually be checking whether the value was "pizza". I'm guessing this wasn't usually a problem, but it's such a horrendous failure mode that the accepted method of checking whether something is undefined was to check whether `typeof something === 'undefined'`. Another way is `var x; console.log(value === x)`, since an uninitialized variable will always be undefined.

Maybe the best option would be `typeof op.insert != null && ['object', 'string'].includes(typeof op.insert)`?

Actually though, this function was intended to be an *example* function that the user of the script could easily redefine to clean up the doc however. I've added a comment.

> Maybe `typeof ....` is *the normal way* to check if an object has a field?

I think the way you check for a field is `'insert' in {insert: null}`, which will evaluate to `true`.
</blockquote></details>

Actually LGTM just caught my mistake in doing `typeof op.insert != null`. It should be `op.insert != null`. Just fixed it.

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1235)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[scripts/db_tools/fix-doc.js, line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1235#-Muw_INc608zKTA68iYs:-Mv-ulEU2BW-lLmBBu2y:bb6wk0y) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b96e6d0ab9560ed8c9b9069237b83fa18348f0b/scripts/db_tools/fix-doc.js#L3)):*

Maybe this should be reworeded to be clearer? I've expanded the comment a bit.
</blockquote></details>

Good
___
*[scripts/db_tools/fix-doc.js, line 45 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1235#-MuwZbHV7XJ5Hq2xs8JD:-Mv-uvXw-HaJaM6MfIuz:bb6wk0y) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b96e6d0ab9560ed8c9b9069237b83fa18348f0b/scripts/db_tools/fix-doc.js#L45)):*

I've renamed this to `cleanupOps()`. What do you think?
</blockquote></details>

Good
___
*[scripts/db_tools/fix-doc.js, line 46 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1235#-Mv-uxjXDpFe4MgmfSEZ:-Mv-uxjXDpFe4MgmfSE_:b-xhyy19) ([raw file](https://github.com/sillsdev/web-xforge/blob/13d80ad283f0fe550f51052be45cd44f4bdc7482/scripts/db_tools/fix-doc.js#L46)):*
> ```JavaScript
> }
> 
> // Edit this function to modify the ops to whatever state they should be in
> ```

That's good. It shows that this method isn't one to just use as is, but it's a place to go make modifications to do something that's desired for a particular case.



_Code quote:_
```JavaScript
/ Edit this function to
```

PR #1238, opened 2022-02-09
SF-1488 Prevent editing of known corrupt text docs

Nateowami:
- Prevent editing of known corrupt text docs
- Inform user doc is corrupt
- Make editor border color consistent with other borders

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1238)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
@@ -1,5 +1,6 @@
 @use '@material/typography/mixins' as typography;
 
+@import 'src/variables';
 @import 'quill/dist/quill.snow.css';
 @import 'usx';
 
@@ -15,6 +16,10 @@ quill-editor {
       text-align: start;
     }
   }
+}
+
+.ql-container.ql-snow {
+  border: 1px solid $border-color;
 }
 
 .read-only-editor > .ql-container {
--- a/src/SIL.XForge.Scripture/ClientApp/src/_variables.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_variables.scss
@@ -16,6 +16,7 @@ $questions: $purpleLight;
 $answers: $greenLight;
 $comments: #ff7300;
 $languageFonts: language_picker, language_picker_fallback, Arial, Helvetica, sans-serif;
+$border-color: rgba(0, 0, 0, 0.12);
 
 $pt-button-green: $greenDark;
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss
@@ -82,7 +82,7 @@
 // Section: Import errors screen
 
 .mat-column-rowNumber {
-  border-right: 1px solid rgba(0, 0, 0, 0.12);
+  border-right: 1px solid $border-color;
   padding-right: 1em;
   text-align: right;
   font-weight: 500;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -3,7 +3,7 @@ import Quill, { DeltaOperation, DeltaStatic, RangeStatic, Sources, StringMap } f
 import { VerseRef } from 'realtime-server/lib/esm/scriptureforge/scripture-utils/verse-ref';
 import { Subscription } from 'rxjs';
 import { Delta, TextDoc } from '../../core/models/text-doc';
-import { VERSE_FROM_SEGMENT_REF_REGEX } from '../utils';
+import { containsInvalidOp, VERSE_FROM_SEGMENT_REF_REGEX } from '../utils';
 import { USFM_STYLE_DESCRIPTIONS } from './usfm-style-descriptions';
 
 const PARA_STYLES: Set<string> = new Set<string>([
@@ -114,6 +114,12 @@ export class TextViewModel {
     }
     const textData = this.textDoc.data;
     return textData.ops == null || textData.ops.length === 0;
+  }
+
+  get areOpsCorrupted(): boolean {
+    return (
+      this.textDoc?.isLoaded === true && this.textDoc.data?.ops != null && containsInvalidOp(this.textDoc.data.ops)
+    );
   }
 
   bind(textDoc: TextDoc, subscribeToUpdates: boolean): void {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -196,6 +196,10 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     return this.displayMessage;
   }
 
+  get areOpsCorrupted(): boolean {
+    return this.viewModel.areOpsCorrupted;
+  }
+
   @Input() set placeholder(value: string) {
     this._placeholder = value;
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.spec.ts
@@ -1,7 +1,8 @@
+import { DeltaOperation } from 'rich-text';
 import { SelectableProject } from '../core/paratext.service';
-import { projectLabel } from './utils';
+import { containsInvalidOp, projectLabel } from './utils';
 
-describe('Utils', () => {
+describe('shared utils', () => {
   describe('projectLabel function', () => {
     const shortName = 'SN';
     const name = 'Name';
@@ -29,6 +30,41 @@ describe('Utils', () => {
     it('should use both if project has both', () => {
       const label = projectLabel({ name, shortName } as SelectableProject);
       expect(label).toEqual(shortName + ' - ' + name);
+    });
+  });
+
+  describe('isDocCorrupted function', () => {
+    it('requires op.insert to be a string or object', () => {
+      expect(containsInvalidOp([{}])).toBeTrue();
+      expect(containsInvalidOp([{ insert: null }])).toBeTrue();
+      expect(containsInvalidOp([{ insert: 1 }])).toBeTrue();
+      // this isn't actually a good op, but isDocCorrupted won't see a problem with it
+      // it's looking for known issues, not proving validity
+      expect(containsInvalidOp([{ insert: {} }])).toBeFalse();
+    });
+
+    it('requires that there be an insert', () => {
+      expect(containsInvalidOp([{ thisIsNotAnInsert: 1 } as DeltaOperation])).toBeTrue();
+    });
+
+    it('rejects delete and retain ops', () => {
+      expect(containsInvalidOp([{ insert: 'text' }, { delete: 100 }])).toBeTrue();
+      expect(containsInvalidOp([{ insert: 'text' }, { retain: 1 }])).toBeTrue();
+      expect(containsInvalidOp([{ insert: 'text' }, { delete: 100 }, { retain: 1 }])).toBeTrue();
+    });
+
+    it('does not allow op.insert.verse to be a non object', () => {
+      // op.insert.verse does not have to exist
+      expect(containsInvalidOp([{ insert: {} }])).toBeFalse();
+      // but if it does, it has to be an object
+      expect(containsInvalidOp([{ insert: { verse: true } }])).toBeTrue();
+      expect(containsInvalidOp([{ insert: { verse: { number: '2', style: 'v' } } }])).toBeFalse();
+    });
+
+    it('requires attributes.segment to not contain null or undefined', () => {
+      expect(containsInvalidOp([{ insert: 'text', attributes: { segment: 'verse_1_1' } }])).toBeFalse();
+      expect(containsInvalidOp([{ insert: 'text', attributes: { segment: 'verse_1_null' } }])).toBeTrue();
+      expect(containsInvalidOp([{ insert: 'text', attributes: { segment: 'verse_1_undefined' } }])).toBeTrue();
     });
   });
 });
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts
@@ -1,4 +1,5 @@
 import { VerseRef } from 'realtime-server/lib/esm/scriptureforge/scripture-utils/verse-ref';
+import { DeltaOperation } from 'rich-text';
 import { SelectableProject } from '../core/paratext.service';
 
 // Regular expression for getting the verse from a segment ref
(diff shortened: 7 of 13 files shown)


RaymondLuong3 (changes requested):
I think we also need to disable sync for users who get into this situation since we really don't know what could happen if a user's text doc gets corrupted, and they try to synchronize their project with Paratext.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-:-Mvt_3MQ4mxgYn3zyJOz:bbjozqf)* status: all files reviewed, 5 unresolved discussions (waiting on @Nateowami)

___
*[-- commits, line 2 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-MvtO_lo3hbs6QwZF5Bn):*
Would we need to do this if we detect invalid ops before they happen?



_Code quote:_
```
SF-1488 Prevent editing of known corrupt text docs
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-MvtQwUJ29RiELqmrlTK:-MvtQwUJ29RiELqmrlTL:b-aor1rw) ([raw file](https://github.com/sillsdev/web-xforge/blob/0d291593802322305b14fd76c77481ce4565afe3/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L68)):*
> ```XML
>  * @param ops An array of ops to check.
>  */
> export function isDocCorrupted(ops: any[]): boolean {
> ```

Let's change this docHasInvalidOp instead of isCorrupted. We are checking whether the ops in this doc passes a series of validity check, so I think that is more appropriate. The comment would need updated too.



_Code quote:_
```XML
export function isDocCorrupted(ops: any[]): boolean {
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts, line 69 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-MvtRMKL9ZZ_8SwH_NpH:-MvtRMKL9ZZ_8SwH_NpI:b-i0jptb) ([raw file](https://github.com/sillsdev/web-xforge/blob/0d291593802322305b14fd76c77481ce4565afe3/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L69)):*
> ```XML
>  */
> export function isDocCorrupted(ops: any[]): boolean {
>   return ops.some(
> ```

Where do we check that the op.retain and op.delete properties are null. I know it is in the test, but I can't see how we do it here.



_Code quote:_
```XML
  return ops.some(
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts, line 120 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-MvtWZ3h2Y5EKrVyyz9z:-MvtWZ3h2Y5EKrVyyzA-:b-8b3wfk) ([raw file](https://github.com/sillsdev/web-xforge/blob/0d291593802322305b14fd76c77481ce4565afe3/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L120)):*
> ```XML
> 
>   get isDocCorrupted(): boolean {
>     return this.textDoc?.isLoaded === true && this.textDoc.data?.ops != null && isDocCorrupted(this.textDoc.data.ops);
> ```

Sometimes I've seen something like `!!this.textDoc?.isLoaded` as a way to test whether the doc is not null and isLoad is true. The current way is more descriptive and I also like it.



_Code quote:_
```XML
this.textDoc?.isLoaded === true
```
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 199 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-MvtX3Pt9gYUq-Q8T2N8:-MvtX3Pt9gYUq-Q8T2N9:bdswbrs) ([raw file](https://github.com/sillsdev/web-xforge/blob/0d291593802322305b14fd76c77481ce4565afe3/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L199)):*
> ```XML
>   }
> 
>   get isCorrupted(): boolean {
> ```

I am not a fan of having production code checking for corruption because it sounds like we are falling a step behind. I know changing this to check if the text isInvalid is just putting another name on the same problem. Although perhaps we want to differentiate invalid and corrupted. Text can be invalid in SF if usfm markers in Paratext is not right or incomplete. So maybe isCorrupted is the best choice of word here.



_Code quote:_
```XML
  get isCorrupted(): boolean {
    return this.viewModel.isDocCorrupted;
  }
```

Nateowami (commented):
Well, the sync will fail, and right now that's how users tend to find out about the problem. I'm not sure the best way to disable the sync though. Maybe a pre-sync check that no docs are corrupted? I think that could be done as a separate new feature.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-:-MvuERYN8xzCtF-QUw6f:bjmgc78)* status: all files reviewed, 4 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[-- commits, line 2 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-MvtO_lo3hbs6QwZF5Bn):*

Would we need to do this if we detect invalid ops before they happen?
</blockquote></details>

No, but I'm not sure we can completely prevent docs from getting corrupt. When Mark and I looked into this we couldn't figure out how it occurred. We noticed that the ops that lead to the problem went something like this:
- Have doc with length 100.
- Delete 50 characters. Doc now has length 50.
- Delete character at position 75, which doesn't exist. Now the doc is corrupted.

While we don't know how this happened, I suspect that it has something to do with operations that don't properly apply getting applied.  The last operation has the appearance of having been created to apply to the document before the deletion happened. That made me concerned that whatever we do to try to prevent invalid operations from being applied to a document might not be successful. After all, editing can be done offline, and then the changes try to merge after a connection is restored, and that could create all sorts of unexpected scenarios. I would have expected something like an "op does not apply error," and then the op just does not get applied.

This is intended as a safety net for when the railing isn't high enough. I'm hoping the function in the utils file can be used for checking ops before they get saved.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-MvtQwUJ29RiELqmrlTK:-MvuCtYj7ZU2uyO4qYMU:b-sdkx86) ([raw file](https://github.com/sillsdev/web-xforge/blob/0d291593802322305b14fd76c77481ce4565afe3/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L68)):*

Let's change this docHasInvalidOp instead of isCorrupted. We are checking whether the ops in this doc passes a series of validity check, so I think that is more appropriate. The comment would need updated too.
</blockquote></details>

I like that idea, but it's technically not a doc, just an array of ops. So maybe we should call it `containsInvalidOp`?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts, line 69 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-MvtRMKL9ZZ_8SwH_NpH:-MvuD2hB0v0T0LYOOmAm:bbevh1n) ([raw file](https://github.com/sillsdev/web-xforge/blob/0d291593802322305b14fd76c77481ce4565afe3/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L69)):*

Where do we check that the op.retain and op.delete properties are null. I know it is in the test, but I can't see how we do it here.
</blockquote></details>

We don't directly do that. Operations are generally inserts, deletes, or retains. We just assume that if it's an insert, it won't also have a delete or retain property.

We could do `'delete' in op || 'retain' in op`; I just didn't see it as necessary.

Nateowami (commented):
> What do you think about adding a more specific message when the sync fails?

A message to "please contact the team" if it fails would be fairly easy to implement. I think we should add that. But not all failures are caused by data corruption, so it might be difficult to get to the point where we can tell a user what the cause is.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-:-Mw2Yp0nBbpj4tYUyw-5:b9rhv4k)* status: 10 of 13 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts, line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-MvtQwUJ29RiELqmrlTK:-Mw2ZCsJ-jpKC_TQnt1M:b-k6gp1e) ([raw file](https://github.com/sillsdev/web-xforge/blob/0d291593802322305b14fd76c77481ce4565afe3/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L68)):*

Good thinking. Let's change it to `containsInvalidOp`. We want to differentiate that this method so we know it is supposed to be used on the ops in a delta that represent a doc, and not just any delta. I see the method comment does that. Is there a reason why we aren't passing in the text-data and get the ops from there? I don't like the `any[]` type of the parameter.
</blockquote></details>

Done. I also updated the documentation for the method. I used `any[]` because I looked somewhere and found that as the definition for an ops array. Looking today I see we can use `DeltaOperation[]`, so I've changed it to that.

Regarding passing the doc itself rather than the ops array, I think that gives this utility method more responsibility than it should have. A component always has to check whether a doc is loaded, and once it is loaded checks whether the doc is corrupted or not. Considering the possibility of the doc not yet being loaded is more the responsibility of the component than of a method like this. There are really three possibilities: corrupted, not corrupted, and doc not yet loaded. Since this method returns a boolean, we would have to decide that "doc not yet loaded" will correspond with "not corrupted" and group them together. And it wouldn't be possible to tell that this is what is happening just by looking at the code that calls this. It also made the tests a lot cleaner.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 199 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-MvtX3Pt9gYUq-Q8T2N8:-Mw2bjNsBS_kOR4PQMAv:b-shzm78) ([raw file](https://github.com/sillsdev/web-xforge/blob/0d291593802322305b14fd76c77481ce4565afe3/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L199)):*

I am not a fan of having production code checking for corruption because it sounds like we are falling a step behind. I know changing this to check if the text isInvalid is just putting another name on the same problem. Although perhaps we want to differentiate invalid and corrupted. Text can be invalid in SF if usfm markers in Paratext is not right or incomplete. So maybe isCorrupted is the best choice of word here.
</blockquote></details>

We could call it something like "areOpsCorrupted", and maybe the other should be "isUsfmInvalid"? That would be a lot easier to distinguish than "corrupted" and "invalid", which while somewhat distinct are close enough to not clearly convey what the difference is.

RaymondLuong3 (changes requested):
Now that I think of it, maybe the best place to put a contact the SF team would be in the scripture view where we should the "you cannot edit corrupted text" message. If a users gets into this state, it would be good for them to know immediately.

> Reviewed 2 of 3 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-:-Mw7bMOE4dgbkXEJ-MHj:b53t17p)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 199 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-MvtX3Pt9gYUq-Q8T2N8:-Mw7aWI11APKrVFiStFv:b-u1npy5) ([raw file](https://github.com/sillsdev/web-xforge/blob/0d291593802322305b14fd76c77481ce4565afe3/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L199)):*

We could call it something like "areOpsCorrupted", and maybe the other should be "isUsfmInvalid"? That would be a lot easier to distinguish than "corrupted" and "invalid", which while somewhat distinct are close enough to not clearly convey what the difference is.
</blockquote></details>

We could add a comment on this getter to help clarify that this checks specifically for invalid ops in a doc, and not usfm formatting problems. But since this getter is so simple I'm ok with it as is.
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 86 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-Mw7cOpo5lr_zoDXL-PI:-Mw7cOpo5lr_zoDXL-PJ:b-a3xzna) ([raw file](https://github.com/sillsdev/web-xforge/blob/613df36493529bb87ee9799ce92c3e6422b9404e/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L86)):*
> ```JSON
>     "project_data_out_of_sync": "Your project data is out of sync. Please synchronize your project to edit this text.",
>     "swap_source_and_target": "Swap source and target",
>     "text_doc_corrupted": "This chapter cannot be edited because the data has become corrupted.",
> ```

We should direct users to contact SF at this point.



_Code quote:_
```JSON
    "text_doc_corrupted": "This chapter cannot be edited because the data has become corrupted.
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1238)* status: 11 of 13 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 86 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-Mw7cOpo5lr_zoDXL-PI:-Mw7p6smA2vsrOPYwu3J:b-khputc) ([raw file](https://github.com/sillsdev/web-xforge/blob/613df36493529bb87ee9799ce92c3e6422b9404e/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L86)):*

We should direct users to contact SF at this point.
</blockquote></details>

Actually, it already does that. Here's a screenshot of what the error messages look like (I edited the template to make them all show up)):


I did this by reusing a string from the error dialog, "To report an issue, email {{ issueEmailLink }}."

This is a questionable move; I've changed it. Now the strings can be edited interdependently. And if one gets translated, I think Crowdin will automatically translate the other for as long as they are identical.

RaymondLuong3 (approved):
Ok, I'm glad to see that users will be notified that something needs attention. Thanks for your work!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-:-Mw83NF7-wJZJbEWqrJu:b-y1am1x)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json, line 86 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1238#-Mw7cOpo5lr_zoDXL-PI:-Mw832iwCDHDgGgeed3S:btojbfd) ([raw file](https://github.com/sillsdev/web-xforge/blob/613df36493529bb87ee9799ce92c3e6422b9404e/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L86)):*

Actually, it already does that. Here's a screenshot of what the error messages look like (I edited the template to make them all show up)):


I did this by reusing a string from the error dialog, "To report an issue, email {{ issueEmailLink }}."

This is a questionable move; I've changed it. Now the strings can be edited interdependently. And if one gets translated, I think Crowdin will automatically translate the other for as long as they are identical.
</blockquote></details>

Great!

PR #1240, opened 2022-02-09
Add SyncUser migration tool

RaymondLuong3:
To hide pt usernames from users that should not have access, the sf-project model had the ParatextUsers property added, but this will be projected away for users who do not have access. Since SyncUsers in the SFProjectSecret model essentially shared the same purpose of mapping paratext usernames to an id used to mark notes, it made sense to merge the PT sync users from the SFProjectSecret model into the SFProject in the ParatextUsers property.

This tool can be run on the QA and Live servers to move the SyncUsers of each project secret to the corresponding project. The link to the run sheet can be found [here](https://docs.google.com/document/d/1sTk0Du8XxebFNbAlqVH-BHbyHGwqpAPTcmIDPPEBBC8/edit?usp=sharing)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1240)

### DIFF
--- a/src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs
+++ b/src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs
@@ -80,12 +80,6 @@ namespace SyncUserToPTUser
                         }
                         if (projectDoc.Data.ParatextUsers == null)
                             op.Set(pd => pd.ParatextUsers, ptUsers);
-                        else
-                        {
-                            _logger.Log($"      {Program.Bullet3} Some Paratext users already exist. Adding to the list.");
-                            foreach (ParatextUserProfile user in ptUsers)
-                                op.Add(pd => pd.ParatextUsers, user);
-                        }
                     });
                     await _projectSecretRepo.UpdateAsync(projectSecret.Id, u =>
                     {
--- a/src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs
+++ b/src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs
@@ -39,7 +39,7 @@ namespace SyncUserToPTUser
             List<SFProject> sfProjects = _realtimeService.QuerySnapshots<SFProject>().ToList<SFProject>();
             if (sfProjectIdsToUpdate != null)
             {
-                sfProjects.RemoveAll((SFProject sfProject) => !sfProjectIdsToUpdate.Contains(sfProject.Id));
+                sfProjects = sfProjects.Where((SFProject p) => sfProjectIdsToUpdate.Contains(p.Id)).ToList();
                 string ids = string.Join(' ', sfProjects.Select((SFProject sfProject) => sfProject.Id));
                 int count = sfProjects.Count;
                 _logger.Log($"Only working on the subset of projects (count {count}) with these SF project ids: {ids}");
@@ -66,16 +66,26 @@ namespace SyncUserToPTUser
                     List<ParatextUserProfile> ptUsers = new List<ParatextUserProfile>();
                     await projectDoc.SubmitJson0OpAsync(op =>
                     {
-                        foreach (SyncUser syncUser in projectSecret.SyncUsers)
+                        if (projectSecret.SyncUsers != null)
                         {
-                            var paratextUser = new ParatextUserProfile
+                            foreach (SyncUser syncUser in projectSecret.SyncUsers)
                             {
-                                Username = syncUser.ParatextUsername,
-                                OpaqueUserId = syncUser.Id,
-                            };
-                            ptUsers.Add(paratextUser);
+                                var paratextUser = new ParatextUserProfile
+                                {
+                                    Username = syncUser.ParatextUsername,
+                                    OpaqueUserId = syncUser.Id,
+                                };
+                                ptUsers.Add(paratextUser);
+                            }
                         }
-                        op.Set(pd => pd.ParatextUsers, ptUsers);
+                        if (projectDoc.Data.ParatextUsers == null)
+                            op.Set(pd => pd.ParatextUsers, ptUsers);
+                        else
+                        {
+                            _logger.Log($"      {Program.Bullet3} Some Paratext users already exist. Adding to the list.");
+                            foreach (ParatextUserProfile user in ptUsers)
+                                op.Add(pd => pd.ParatextUsers, user);
+                        }
                     });
                     await _projectSecretRepo.UpdateAsync(projectSecret.Id, u =>
                     {
--- a/src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs
+++ b/src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs
@@ -1,0 +1,96 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.ExceptionServices;
+using System.Threading.Tasks;
+using Autofac;
+using MongoDB.Driver;
+using SIL.XForge.DataAccess;
+using SIL.XForge.Models;
+using SIL.XForge.Realtime;
+using SIL.XForge.Realtime.Json0;
+using SIL.XForge.Scripture.Models;
+using SIL.XForge.Scripture.Services;
+
+namespace SyncUserToPTUser
+{
+    /// <summary> Code to move sync users to the sf-project model. </summary>
+    public class SyncUserToPTUserService : ISyncUserToPTUserService
+    {
+        private readonly IRealtimeService _realtimeService;
+        private readonly IParatextService _paratextService;
+        private readonly IRepository<SFProjectSecret> _projectSecretRepo;
+        private IConnection _realtimeServiceConnection;
+
+        private readonly IProgramLogger _logger;
+
+        public SyncUserToPTUserService(IRealtimeService realtimeService, IParatextService paratextService,
+            IRepository<SFProjectSecret> projectSecretRepo, IProgramLogger logger)
+        {
+            _realtimeService = realtimeService;
+            _paratextService = paratextService;
+            _projectSecretRepo = projectSecretRepo;
+            _logger = logger;
+        }
+
+        /// <summary> Move sync users to the SF project model. </summary>
+        public async Task MoveSyncUsersToProject(bool dryRun, ISet<string> sfProjectIdsToUpdate = null)
+        {
+            List<SFProject> sfProjects = _realtimeService.QuerySnapshots<SFProject>().ToList<SFProject>();
+            if (sfProjectIdsToUpdate != null)
+            {
+                sfProjects.RemoveAll((SFProject sfProject) => !sfProjectIdsToUpdate.Contains(sfProject.Id));
+                string ids = string.Join(' ', sfProjects.Select((SFProject sfProject) => sfProject.Id));
+                int count = sfProjects.Count;
+                _logger.Log($"Only working on the subset of projects (count {count}) with these SF project ids: {ids}");
+            }
+            _realtimeServiceConnection = await _realtimeService.ConnectAsync();
+
+            _logger.Log($"{Program.Bullet1} Report on projects to move sync users:");
+            foreach (SFProject sfProject in sfProjects)
+            {
+                try
+                {
+                    IDocument<SFProject> projectDoc = await _realtimeServiceConnection.FetchAsync<SFProject>(sfProject.Id);
+                    if (!(await _projectSecretRepo.TryGetAsync(sfProject.Id)).TryResult(out SFProjectSecret projectSecret))
+                    {
+                        _logger.Log($"Could not find project secret for project {sfProject.Id}.");
+                        continue;
+                    }
+                    if (dryRun)
+                    {
+                        int syncUserCount = projectSecret.SyncUsers.Count;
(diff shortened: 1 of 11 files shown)


src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs
```diff
@@ -0,0 +1,96 @@
...
+            _logger = logger;
+        }
+
+        /// <summary> Move sync users to the SF project model. </summary>
+        public async Task MoveSyncUsersToProject(bool dryRun, ISet<string> sfProjectIdsToUpdate = null)
+        {
+            List<SFProject> sfProjects = _realtimeService.QuerySnapshots<SFProject>().ToList<SFProject>();
+            if (sfProjectIdsToUpdate != null)
+            {
+                sfProjects.RemoveAll((SFProject sfProject) => !sfProjectIdsToUpdate.Contains(sfProject.Id));
```
marksvc, line 42: Might be actual problem:

Does RemoveAll() not need to return something that we keep? (Rather than modify the set in place). 

Just saying, but not needed:

It might be a bit better conceptually to look for what you want to keep, rather than remove what you don't want to keep. You might do something like
```suggestion
                sfProjects = sfProjects.Where((SFProject sfProject) => sfProjectIdsToUpdate.Contains(sfProject.Id));
```

In fact, you could even start off the method with something like the following:
```cs
if (sfProjectIdsToUpdate == null) {
  throw new ArgumentNullException(nameof(sfProjectIdsToUpdate));
}
List<SFProject> sfProjects = _realtimeService.QuerySnapshots<SFProject>()
  .ToList<SFProject>()
  .Where((SFProject sfProject) => sfProjectIdsToUpdate.Contains(sfProject.Id));
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1240)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs, line 42 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1240#-MvVGPkdqnoJQUKxZcTR-r1-42:-MynRjWj5MUYelUVJdrS:brp7z7o) ([raw file](https://github.com/sillsdev/web-xforge/blob/71484b9a7c38ba58b17a59578acb5abe341a18b9/src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs#L42)):*

Might be actual problem:

Does RemoveAll() not need to return something that we keep? (Rather than modify the set in place). 

Just saying, but not needed:

It might be a bit better conceptually to look for what you want to keep, rather than remove what you don't want to keep. You might do something like
```suggestion
                sfProjects = sfProjects.Where((SFProject sfProject) => sfProjectIdsToUpdate.Contains(sfProject.Id));
```

In fact, you could even start off the method with something like the following:
```cs
if (sfProjectIdsToUpdate == null) {
  throw new ArgumentNullException(nameof(sfProjectIdsToUpdate));
}
List<SFProject> sfProjects = _realtimeService.QuerySnapshots<SFProject>()
  .ToList<SFProject>()
  .Where((SFProject sfProject) => sfProjectIdsToUpdate.Contains(sfProject.Id));
```


</blockquote></details>

Good thinking. I've updated it to check that a project is in the list and select for those.
___
*[src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs, line 75 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1240#-MvVGPkdqnoJQUKxZcTR-r1-75:-MynS2v0CtxIgQh9UX-L:btaakx7) ([raw file](https://github.com/sillsdev/web-xforge/blob/71484b9a7c38ba58b17a59578acb5abe341a18b9/src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs#L75)):*

And so we won't try setting an SF user id but we'll just let that get filled in during the course of upcoming sync activity, right?
</blockquote></details>

That's correct, user Id's are added at sync time, and do not need to be added during this migration.

marksvc (changes requested):
> Reviewed 11 of 11 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1240)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[-- commits, line 8 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1240#-MynXmEXBD0075Ee4a-i):*
It occurs to me: When we run the migration, won't we really be wanting to run the migration on all projects? Having the capability to run on a subset of the projects could potentially be handy. But aren't we really wanting to run on all projects? Are you planning to specify a big list of ids? 
It hasn't seemed like specifying an empty list of ids would result in processing everything, but rather nothing.



_Code quote:_
```
 Really use the subset of projects when specified
```
___
*[src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs, line 87 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1240#-MynW_8UCpEDaEEIYtT1:-MynW_8UCpEDaEEIYtT2:b-fhh12t) ([raw file](https://github.com/sillsdev/web-xforge/blob/c889f468f6b5f036738d8c4757b87d9d31b42afe/src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs#L87)):*
> ```Smalltalk
>                             _logger.Log($"      {Program.Bullet3} Some Paratext users already exist. Adding to the list.");
>                             foreach (ParatextUserProfile user in ptUsers)
>                                 op.Add(pd => pd.ParatextUsers, user);
> ```

If a project already _has_ paratext users, mightn't it be that some of them will overlap with users in the syncusers list? And so then in that situtaion, you might skip items in `ptUsers` that have a ParatextUsername that is already in `projectDoc.Data.ParatextUsers`. ?
Actually, this might start to cause mis-matches since there might now  be two opaque ids for the same Paratext username. One for the old syncuser list, and one for the new ParatextUsers list that existed right before the migration. Is the reason for the already-existing projectDoc.Data.ParatxtUsers simply because projects that have been manipulated on the develop branch may have this property? And/or because of previous attempts (successfully or not) at doing the sync user migration on a project?
You might just consider that if a project _has_ syncusers, then you are going to overwrite the ParatextUsers list with the SyncUser data.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1240)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[-- commits, line 8 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1240#-MynXmEXBD0075Ee4a-i):*

It occurs to me: When we run the migration, won't we really be wanting to run the migration on all projects? Having the capability to run on a subset of the projects could potentially be handy. But aren't we really wanting to run on all projects? Are you planning to specify a big list of ids? 
It hasn't seemed like specifying an empty list of ids would result in processing everything, but rather nothing.
</blockquote></details>

Actually, if the project ids list is null, then the migration is run on all project
___
*[src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs, line 87 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1240#-MynW_8UCpEDaEEIYtT1:-MynZYb-AA59NvbS71Wu:b-17kkhp) ([raw file](https://github.com/sillsdev/web-xforge/blob/c889f468f6b5f036738d8c4757b87d9d31b42afe/src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs#L87)):*

If a project already _has_ paratext users, mightn't it be that some of them will overlap with users in the syncusers list? And so then in that situtaion, you might skip items in `ptUsers` that have a ParatextUsername that is already in `projectDoc.Data.ParatextUsers`. ?
Actually, this might start to cause mis-matches since there might now  be two opaque ids for the same Paratext username. One for the old syncuser list, and one for the new ParatextUsers list that existed right before the migration. Is the reason for the already-existing projectDoc.Data.ParatxtUsers simply because projects that have been manipulated on the develop branch may have this property? And/or because of previous attempts (successfully or not) at doing the sync user migration on a project?
You might just consider that if a project _has_ syncusers, then you are going to overwrite the ParatextUsers list with the SyncUser data. 
</blockquote></details>

The ParatextUsers property may not be null if the migration tool was previously run. This allows the tool to be run multiple times over the a project without any consequences. The reality is that we really do not want to have the server partially migrated. I've removed some logic that adds sync users to the PT user list. We will only care about the situation where the project has never been migrated, and we then move their sync users.
We clear sync users in the project secrets following this, so it is specifically important that we never clear the Paratext Users property, otherwise we will lose any previously migrated Paratext users.

marksvc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1240)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[-- commits, line 8 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1240#-MynXmEXBD0075Ee4a-i):*

Actually, if the project ids list is null, then the migration is run on all project
</blockquote></details>

Oh. I see that now :)
___
*[src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs, line 87 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1240#-MynW_8UCpEDaEEIYtT1:-Mynk9CtAcqMEAUPQ5SK:b-qtjlpd) ([raw file](https://github.com/sillsdev/web-xforge/blob/c889f468f6b5f036738d8c4757b87d9d31b42afe/src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs#L87)):*

The ParatextUsers property may not be null if the migration tool was previously run. This allows the tool to be run multiple times over the a project without any consequences. The reality is that we really do not want to have the server partially migrated. I've removed some logic that adds sync users to the PT user list. We will only care about the situation where the project has never been migrated, and we then move their sync users.
We clear sync users in the project secrets following this, so it is specifically important that we never clear the Paratext Users property, otherwise we will lose any previously migrated Paratext users.
</blockquote></details>

Right okay. Do you think it would be worthwhile to change to also only _delete_ the SyncUsers if we just wrote them into the ParatextUsers? In other words, 

```cs
alreadyHasParatextUsers = projectDoc.Data.ParatextUsers != null 
await projectDoc.SubmitJson0OpAsync(
  if there are any sync users,
    make a set of paratext users
  if !alreadyHasParatextUsers:
    Write in the list of Paratext users.    
)
if !alreadyHasParatextUsers:
  await projectSecretRepo.UpdateAsync(
    unset(projectSecret.SyncUsers)
  )
```

haha, or really
```cs
alreadyHasParatextUsers = projectDoc.Data.ParatextUsers != null 
if alreadyHasParatextUsers:
  log("Oops, this project already _has_ ParatextUsers! Not modifying it.")
  continue
await projectDoc.SubmitJson0OpAsync(
  if there are any sync users,
    make a set of paratext users
  Write in the list of Paratext users.    
)
await projectSecretRepo.UpdateAsync(
  unset(projectSecret.SyncUsers)
)
```

Because it seems like deleting SyncUsers if we see any ParatextUsers could lead to accidentally deleting something based on a mistake.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1240)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs, line 87 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1240#-MynW_8UCpEDaEEIYtT1:-MyrXnF43poSKjsTJDeO:b-n404bt) ([raw file](https://github.com/sillsdev/web-xforge/blob/c889f468f6b5f036738d8c4757b87d9d31b42afe/src/Migrations/SyncUserToPTUser/SyncUserToPTUserService.cs#L87)):*

Right okay. Do you think it would be worthwhile to change to also only _delete_ the SyncUsers if we just wrote them into the ParatextUsers? In other words, 

```cs
alreadyHasParatextUsers = projectDoc.Data.ParatextUsers != null 
await projectDoc.SubmitJson0OpAsync(
  if there are any sync users,
    make a set of paratext users
  if !alreadyHasParatextUsers:
    Write in the list of Paratext users.    
)
if !alreadyHasParatextUsers:
  await projectSecretRepo.UpdateAsync(
    unset(projectSecret.SyncUsers)
  )
```

haha, or really
```cs
alreadyHasParatextUsers = projectDoc.Data.ParatextUsers != null 
if alreadyHasParatextUsers:
  log("Oops, this project already _has_ ParatextUsers! Not modifying it.")
  continue
await projectDoc.SubmitJson0OpAsync(
  if there are any sync users,
    make a set of paratext users
  Write in the list of Paratext users.    
)
await projectSecretRepo.UpdateAsync(
  unset(projectSecret.SyncUsers)
)
```

Because it seems like deleting SyncUsers if we see any ParatextUsers could lead to accidentally deleting something based on a mistake. 
</blockquote></details>

Hey, at first I thought that this was a good idea. But as I think about it again, we really don't want to keep syncUsers around. The plan is to remove them completely from our data model. We have to assume that if Paratext users exist on a project, then the project has had its sync users correctly migrated, and always remove the sync users.

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1240)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1246, opened 2022-02-16
Linting: add return types

irahopkinson:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1246)

### DIFF
--- a/src/RealtimeServer/common/exception-reporter.ts
+++ b/src/RealtimeServer/common/exception-reporter.ts
@@ -14,7 +14,7 @@ export class ExceptionReporter {
     });
   }
 
-  report(error: NotifiableError) {
+  report(error: NotifiableError): void {
     this.bugsnagClient.notify(error);
   }
 }
--- a/src/RealtimeServer/common/index.ts
+++ b/src/RealtimeServer/common/index.ts
@@ -41,9 +41,9 @@ async function startServer(options: RealtimeServerOptions): Promise<void> {
   }
 
   const exceptionReporter = new ExceptionReporter(options.bugsnagApiKey, options.releaseStage, options.version);
-  function reportError(error: any) {
-    console.error(`Error from ShareDB server: ${error}`);
-    exceptionReporter.report(error);
+  function reportError(...args: unknown[]): void {
+    console.error('Error from ShareDB server: ', ...args);
+    exceptionReporter.report(args.toString());
   }
   // ShareDB sometimes reports errors as warnings
   ShareDB.logger.setMethods({ warn: reportError, error: reportError });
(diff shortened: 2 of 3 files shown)


Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1246)* status: all files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[src/RealtimeServer/common/index.ts, line 45 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1246#-Mw2l7Tx2KnF9Yfy6uLs:-Mw2l7Tx2KnF9Yfy6uLt:b-51dysj) ([raw file](https://github.com/sillsdev/web-xforge/blob/61ac1f9661bec6e837b8d76b567bed8ae51dec3c/src/RealtimeServer/common/index.ts#L45)):*
> ```XML
> 
>   const exceptionReporter = new ExceptionReporter(options.bugsnagApiKey, options.releaseStage, options.version);
>   function reportError(error: NotifiableError): void {
> ```

This gets called by ShareDB, not by us. And ShareDB's declared type for this argument is `any`. What this amounts to then is an implicit assertion that what ShareDB will send us is not just anything, but a NotifiableError, which I don't think is correct.

I looked at ShareDB's source code, and it doesn't look like the number of arguments is even guaranteed to be 1. There's this line in `stream-socket.js`:
``` ts
logger.warn('ShareDB client message stream error', error);
```
It looks like the logger has the same interface as console, so maybe we should check what gets passed in, and if it isn't a string, convert it to one before reporting it?

Maybe something along the lines of:
``` ts
function reportError(...args: unknown[]): void {
  console.error('Error from ShareDB server:', ...args);
  exceptionReporter.report(args.toString());
}
```
Of note, `['thing'].toString()` evaluates to `'thing'`, which surprised me a bit. But it means if we are passed a single string as an argument, `args.toString()` will evaluate to just that string. This doesn't handle objects very well though, so it could be seriously improved. I'm reluctant to use `JSON.stringify()` though, because it can't handle circular objects.

irahopkinson (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1246)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/RealtimeServer/common/index.ts, line 45 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1246#-Mw2l7Tx2KnF9Yfy6uLs:-MwUXQv-E-3UWv59th59:b-6gi5an) ([raw file](https://github.com/sillsdev/web-xforge/blob/61ac1f9661bec6e837b8d76b567bed8ae51dec3c/src/RealtimeServer/common/index.ts#L45)):*

This gets called by ShareDB, not by us. And ShareDB's declared type for this argument is `any`. What this amounts to then is an implicit assertion that what ShareDB will send us is not just anything, but a NotifiableError, which I don't think is correct.

I looked at ShareDB's source code, and it doesn't look like the number of arguments is even guaranteed to be 1. There's this line in `stream-socket.js`:
``` ts
logger.warn('ShareDB client message stream error', error);
```
It looks like the logger has the same interface as console, so maybe we should check what gets passed in, and if it isn't a string, convert it to one before reporting it?

Maybe something along the lines of:
``` ts
function reportError(...args: unknown[]): void {
  console.error('Error from ShareDB server:', ...args);
  exceptionReporter.report(args.toString());
}
```
Of note, `['thing'].toString()` evaluates to `'thing'`, which surprised me a bit. But it means if we are passed a single string as an argument, `args.toString()` will evaluate to just that string. This doesn't handle objects very well though, so it could be seriously improved. I'm reluctant to use `JSON.stringify()` though, because it can't handle circular objects.
</blockquote></details>

Thanks for catching that. Yours is a better solution.

Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1246#-:-MwUjhZGAGzYK9edaH8r:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @irahopkinson)

PR #1249, opened 2022-02-18
SF-1471 Delete have-read note refs when rm note

marksvc:
- At sync time, when a note is removed, also remove refs to it from
project-user-config docs' have-read note refs lists, to not needlessly
build up orphaned refs there.
- In the SFProjectUserConfig model, move NoteRefsRead out of the
Checking area.
- This does not take into account notes being removed while a user is
offline, where they could still read a note and add its ref to their
have-read list, even even tho that note had already been removed from
the server.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1249)

### DIFF
--- a/src/RealtimeServer/scriptureforge/models/sf-project-user-config.ts
+++ b/src/RealtimeServer/scriptureforge/models/sf-project-user-config.ts
@@ -18,8 +18,8 @@ export interface SFProjectUserConfig extends ProjectData {
   numSuggestions: number;
   selectedSegment: string;
   selectedSegmentChecksum?: number;
+  noteRefsRead: string[];
   questionRefsRead: string[];
   answerRefsRead: string[];
   commentRefsRead: string[];
-  noteRefsRead: string[];
 }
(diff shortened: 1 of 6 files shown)


RaymondLuong3 (commented):
It is worth looking into removing note read refs on the real-time server. The benefit would be that these refs are cleaned up as a side-effect of deleting the note, which will work in real-time once that feature comes together.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1249#-:-MwYXbyTCEgye4_m_rN3:bf4w40o)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs, line 866 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1249#-MwDvK8sAjZ3OVUg08TK:-MwDvK8sAjZ3OVUg08TL:b-yec5yf) ([raw file](https://github.com/sillsdev/web-xforge/blob/4f9b5a993fb707ecb0960fb54d560dbed9005b2c/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L866)):*
> ```Smalltalk
>                 await pucDoc.SubmitJson0OpAsync(op =>
>                 {
>                     int index = pucDoc.Data.NoteRefsRead.IndexOf(noteDataId);
> ```

This could be moved outside the submitJson0Op



_Code quote:_
```Smalltalk
                    int index = pucDoc.Data.NoteRefsRead.IndexOf(noteDataId);
                    if (index >= 0)
```

PR #1257, opened 2022-03-04
SF-1505 Fix permission error removing user during sync

Nateowami:
Here's what the problem appears to be:
- Project B is based on Project A
- User 1 is on Project A as an observer, and Project B as an admin
- A user is removed from Project A in Paratext
- User 1 does a sync of Project B, which also syncs Project A
- The sync of Project A involves removing the user from the project that was removed already in Paratext
- The removal of the user fails, because the user is only an observer and doesn't have permission to remove a user

The solution is to just not check permissions when removing the user. If the user has permission to sync, and the project had a user removed, then the user can effectuate the removal of the user from the Scripture Forge project. The user isn't actually making a decision to remove a user from the project, only to update Scripture Forge with what is already in Paratext.

There's another simpler way this issue could probably manifest:
- Project A exists, and has User 1 as a translator
- An admin removes a user from Project A in Paratext
- User 1 syncs Project A
- When the sync process attempts to remove the user, this will fail, since the user is only a translator on the project and not an admin

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1257)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
@@ -285,8 +285,7 @@ namespace SIL.XForge.Scripture.Services
                                 await _paratextService.GetResourcePermissionAsync(sourceParatextId, uid, token);
                             if (permission == TextInfoPermission.None)
                             {
-                                // As resource projects don't have administrators, connect as the user we are to remove
-                                await _projectService.RemoveUserAsync(uid, sourceProjectRef, uid);
+                                await _projectService.RemoveUserWithoutPermissionsCheckAsync(uid, sourceProjectRef, uid);
                             }
                         }
                     }
@@ -846,7 +845,7 @@ namespace SIL.XForge.Scripture.Services
                 }
             });
             foreach (var userId in userIdsToRemove)
-                await _projectService.RemoveUserAsync(_userSecret.Id, _projectDoc.Id, userId);
+                await _projectService.RemoveUserWithoutPermissionsCheckAsync(_userSecret.Id, _projectDoc.Id, userId);
             if (_notesMapper.NewSyncUsers.Count > 0)
             {
                 await _projectSecrets.UpdateAsync(_projectSecret.Id, u =>
--- a/src/SIL.XForge/Services/IProjectService.cs
+++ b/src/SIL.XForge/Services/IProjectService.cs
@@ -8,6 +8,7 @@ namespace SIL.XForge.Services
     {
         Task AddUserAsync(string curUserId, string projectId, string projectRole = null);
         Task RemoveUserAsync(string curUserId, string projectId, string projectUserId);
+        Task RemoveUserWithoutPermissionsCheckAsync(string curUserId, string projectId, string projectUserId);
         Task<string> GetProjectRoleAsync(string curUserId, string projectId);
         Task UpdateRoleAsync(string curUserId, string systemRole, string projectId, string projectRole);
         Task<Uri> SaveAudioAsync(string curUserId, string projectId, string dataId, string extension,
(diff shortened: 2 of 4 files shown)


irahopkinson (approved):
:lgtm: this all sounds fine given your description of the issue.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1257#-:-MxIcIq42O9p1j_3X0I_:b-ugkh51)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[src/Migrations/PTDDCloneAll/PTDDSyncRunner.cs, line 210 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1257#-MxIbsy642c2-ALWE7X3:-MxIbsy642c2-ALWE7X4:b-q93v6y) ([raw file](https://github.com/sillsdev/web-xforge/blob/2a03978bcb5f2b2e0d424a6fd13ea59c9127aa35/src/Migrations/PTDDCloneAll/PTDDSyncRunner.cs#L210)):*
> ```Smalltalk
>                             if (permission == TextInfoPermission.None)
>                                 await _projectService.RemoveUserWithoutPermissionsCheckAsync(uid, sourceProjectRef, uid);
> ```

I'm not sure if we should make this change? Do we expect to use this again? (Non-blocking)

RaymondLuong3 (approved):
I suggest writing a test for this change, but it is so small a change it can go without it too.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1257#-:-MxLAavE5KqQ0L6kddvU:b2qstkk)* status: all files reviewed, 1 unresolved discussion (waiting on @irahopkinson and @Nateowami)

___
*[src/Migrations/PTDDCloneAll/PTDDSyncRunner.cs, line 210 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1257#-MxIbsy642c2-ALWE7X3:-MxLA24_CkjVG5ZVf1X2:b-4tzcvv) ([raw file](https://github.com/sillsdev/web-xforge/blob/2a03978bcb5f2b2e0d424a6fd13ea59c9127aa35/src/Migrations/PTDDCloneAll/PTDDSyncRunner.cs#L210)):*

Actually I was wondering the same thing but forgot to leave a note about it. @raymondluong3, looks like you authored this file. Do you know whether this should be updated?
</blockquote></details>

We do not need to use this file again. You can skip this change.

RaymondLuong3 (commented):
Well, I see the test that exists covers the scenario. So all good from my side.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1257#-:-MxLBc7zATeY9vHHnw7W:bdw93vn)* status: all files reviewed, 1 unresolved discussion (waiting on @irahopkinson and @Nateowami)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1257)* status: 4 of 5 files reviewed, all discussions resolved (waiting on @irahopkinson)

___
*[src/Migrations/PTDDCloneAll/PTDDSyncRunner.cs, line 210 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1257#-MxIbsy642c2-ALWE7X3:-MxLghVt2WWsaOgysYDZ:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/2a03978bcb5f2b2e0d424a6fd13ea59c9127aa35/src/Migrations/PTDDCloneAll/PTDDSyncRunner.cs#L210)):*

We do not need to use this file again. You can skip this change.
</blockquote></details>

Done.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1257)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1263, opened 2022-03-11
SF-1508a Show projects the user cannot sync on connect page

RaymondLuong3:
* determine if a user has a license to a shared repository
* set a can synchronize flag on paratext projects
* show message on settings and connect project pages
* disable options for projects that users cannot sync


RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1263)* status: 0 of 15 files reviewed, all discussions resolved

___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1091 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1263#-MxvLXYm0o-cpwH8gWOb:-MxvLXYm0o-cpwH8gWOc:b4a17s2) ([raw file](https://github.com/sillsdev/web-xforge/blob/e961609e2bd8d1231a79262db4cf81893bfabdc7/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1091)):*
> ```Smalltalk
>                     IsConnectable = ptProjectIsConnectable,
>                     IsConnected = sfProjectExists && sfUserIsOnSfProject,
>                     CanSynchronize = remotePtProject.License != null && !remotePtProject.License.IsLimited
> ```

I wanted to write a test for this, but project licenses are difficult to setup since they are signed using RSA cryto provider service, and this change is so small I didn't think it was worth the effort.

Nateowami (commented):
It was my mistake to not investigate more closely the underlying cause and whether this is the right path to pursue. It very well may be, but we should probably put this on hold for the moment until we can figure out whether it's the right way or not. #1264 on the other hand is almost certainly the right solution, and it's what users are up against at the moment, so it's the priority.

> Reviewed 1 of 15 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1263#-:-MyDplZl-Oz_XRh9gWyU:bac95ol)* status: 2 of 15 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/Models/ParatextProject.cs, line 25 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1263#-My94JEa7WZdnUnQ2mAS:-My94JEa7WZdnUnQ2mAT:bga6e7i) ([raw file](https://github.com/sillsdev/web-xforge/blob/4273516bdebd734028b0efbf9af16eb26515fdd1/src/SIL.XForge.Scripture/Models/ParatextProject.cs#L25)):*
> ```Smalltalk
>         public bool IsConnected { get; set; }
>         /// <summary> Indicates whether a user can send/receive the project based having a valid license. </summary>
>         public bool CanSynchronize { get; set; }
> ```

To me these variable names don't seem very descriptive. `IsConnectable` and `CanSynchronize` would seem to mean about the same thing. When there was only one that was probably fine, but it seems like they could be more descriptive. I don't think they're in the DB, so presumably we could rename them pretty easily? (Actually there's the issue that our front end can be older than our back end, so we do have to be careful with `IsConnectable`). 
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1091 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1263#-MxvLXYm0o-cpwH8gWOb:-My8yBQG2ahxG2Bslqw-:b8b25xs) ([raw file](https://github.com/sillsdev/web-xforge/blob/e961609e2bd8d1231a79262db4cf81893bfabdc7/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1091)):*

I wanted to write a test for this, but project licenses are difficult to setup since they are signed using RSA cryto provider service, and this change is so small I didn't think it was worth the effort.
</blockquote></details>

I think this could also be `remotePtProject.License?.IsLimited == false`, but in this case what you've written is probably better. I'm just leaving this note here because this line prompted me to look up C#'s optional chaining.

PR #1265, opened 2022-03-15
Removed all references to beta migration code

nigel-wells:
- Removes all beta migration checks within the main SF app
- Removes any ability to start up an alternative beta instance

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1265)


Nateowami (changes requested):
Looks good; just one change needed I think.

> Reviewed 59 of 59 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1265#-:-MyAQ1hX4sk4-DrD3oi8:bcotz2h)* status: all files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @nigel-wells)

___
*[src/SIL.XForge/Controllers/UsersRpcController.cs, line 86 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1265#-MyAM0IHApOsadybXZaq:-MyAM0IHApOsadybXZar:b-65bwfh) ([raw file](https://github.com/sillsdev/web-xforge/blob/16c6010607e9916b2f7d086517170a2aaa7d5010/src/SIL.XForge/Controllers/UsersRpcController.cs#L86)):*
> ```Smalltalk
>         }
> 
>         public async Task<IRpcMethodResult> CheckUserNeedsMigrating(string userId)
> ```

This needs to not be removed, because old clients will still be calling it until the clients get updated. Otherwise users get error messages until they update. It could be changed to just return false though.

This would be a good reason to make the client forcibly update the user immediately. I'm not sure what the downsides of that would be.
___
*[src/SIL.XForge/Controllers/UsersRpcController.cs, line 91 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1265#-MyAMhee8rA1pXVX-JUu:-MyAMhee8rA1pXVX-JUv:b-voxr6i) ([raw file](https://github.com/sillsdev/web-xforge/blob/16c6010607e9916b2f7d086517170a2aaa7d5010/src/SIL.XForge/Controllers/UsersRpcController.cs#L91)):*
> ```Smalltalk
>         }
> 
>         public async Task<IRpcMethodResult> UserMigrationComplete(string userId)
> ```

This probably *can* be removed, because unlike the one above, I *think* (I didn't check) it only gets called by the front end if the one above returned true and a migration occurred. If the one above is always returning false, then this one would never be reached. Or this could be made a no-op for now, until we remove both together.
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_az.json, line 29 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1265#-MyAOP_x1W8edDFic7z2:-MyAOP_yB4cUTU7DAMpp:bm3ppkk) ([raw file](https://github.com/sillsdev/web-xforge/blob/16c6010607e9916b2f7d086517170a2aaa7d5010/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_az.json#L29)):*
> ```JSON
>     "update_is_available": "Yeniləmə mümkündür Xahiş edirəm səhifəni yeniləyin."
>   },
>   "beta_migration": {
> ```

This is fine, but generally if a change is made to a localization file, it only needs to be made to the English file. When the English file is sent to Crowdin and then we pull translations from Crowdin, all the non-English files get updated automatically. In this case the change was simple, so going through and doing it manually probably didn't take much time. But I just wanted to point out that it would happen automatically the next time translation files are updated, because that's probably not obvious.

Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1265)* status: all files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @nigel-wells)

___
*[src/SIL.XForge/Controllers/UsersRpcController.cs, line 86 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1265#-MyAM0IHApOsadybXZaq:-MyAQDsQ9W7V_imNPw0o:b-bgdsg4) ([raw file](https://github.com/sillsdev/web-xforge/blob/16c6010607e9916b2f7d086517170a2aaa7d5010/src/SIL.XForge/Controllers/UsersRpcController.cs#L86)):*

This needs to not be removed, because old clients will still be calling it until the clients get updated. Otherwise users get error messages until they update. It could be changed to just return false though.

This would be a good reason to make the client forcibly update the user immediately. I'm not sure what the downsides of that would be.
</blockquote></details>

I forgot to mention, a good way of marking this would be with an annotation:
```
[Obsolete("Only here for clients still running a front end that still calls it")]
```
This is what we currently do for `SFProjectsRpcController.HasTransceleratorQuestions`. I initially removed that method and it produced errors on QA for the testers, because until the front end code is actually updated it would automatically call that endpoint.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1265)* status: 56 of 59 files reviewed, 2 unresolved discussions (waiting on @irahopkinson and @Nateowami)

___
*[src/SIL.XForge/Controllers/UsersRpcController.cs, line 86 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1265#-MyAM0IHApOsadybXZaq:-MydKukv4V-U87DHqPb4:bioyobh) ([raw file](https://github.com/sillsdev/web-xforge/blob/16c6010607e9916b2f7d086517170a2aaa7d5010/src/SIL.XForge/Controllers/UsersRpcController.cs#L86)):*

I forgot to mention, a good way of marking this would be with an annotation:
```
[Obsolete("Only here for clients still running a front end that still calls it")]
```
This is what we currently do for `SFProjectsRpcController.HasTransceleratorQuestions`. I initially removed that method and it produced errors on QA for the testers, because until the front end code is actually updated it would automatically call that endpoint.
</blockquote></details>

Ah yes good point - I've restored it, added that comment, and also forced it to return `false`
___
*[src/SIL.XForge/Controllers/UsersRpcController.cs, line 91 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1265#-MyAMhee8rA1pXVX-JUu:-MydLDbj0_lu-QNR_cZ3:buqsu8n) ([raw file](https://github.com/sillsdev/web-xforge/blob/16c6010607e9916b2f7d086517170a2aaa7d5010/src/SIL.XForge/Controllers/UsersRpcController.cs#L91)):*

This probably *can* be removed, because unlike the one above, I *think* (I didn't check) it only gets called by the front end if the one above returned true and a migration occurred. If the one above is always returning false, then this one would never be reached. Or this could be made a no-op for now, until we remove both together.
</blockquote></details>

Yes safe to remove - it would only get called if we happened to restart the server while someone was in the middle of a migration which is very unlikely.
___
*[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_az.json, line 29 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1265#-MyAOP_x1W8edDFic7z2:-MydLM7kEVdbvY8BWnib:b-4guora) ([raw file](https://github.com/sillsdev/web-xforge/blob/16c6010607e9916b2f7d086517170a2aaa7d5010/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_az.json#L29)):*

This is fine, but generally if a change is made to a localization file, it only needs to be made to the English file. When the English file is sent to Crowdin and then we pull translations from Crowdin, all the non-English files get updated automatically. In this case the change was simple, so going through and doing it manually probably didn't take much time. But I just wanted to point out that it would happen automatically the next time translation files are updated, because that's probably not obvious.
</blockquote></details>

Thanks for clarifying. I knew it created new entries but wasn't sure about deleting - now I know.

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1265)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @irahopkinson)

Nateowami (approved):
@nigel-wells Do you want Ira to review this, or should we go ahead and merge it?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1265#-:-Myhx96_EjuEn5ELooxr:b9vaw1z)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @irahopkinson)

Nateowami (approved):
I'm going to wait to merge this when there aren't as many moving parts getting merged to master and shipped. If possible I'd like this to be the only significant change when it gets shipped, just because I don't have a ton of confidence in my ability to get it right the first time (I think it's going to be easy, but there's no point in making it any more complicated than it needs to be).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1265#-:-MyzJg842-8j0PUdYwn0:b-yn1i1g)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @irahopkinson)

PR #1273, opened 2022-03-21
SF-1501 Clean up old SyncUser references

RaymondLuong3:
* also add tests for when no user exists for a note


marksvc (changes requested):
> Reviewed 7 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1273)* status: 7 of 8 files reviewed, 7 unresolved discussions (waiting on @marksvc and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/Models/SFProjectSecret.cs, line 22 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGCz-_5jc4l2ulzI0b:-MzGCz-_5jc4l2ulzI0c:blde0nx) ([raw file](https://github.com/sillsdev/web-xforge/blob/72a0596010b44af527caa8b9bb790f79e5840c59/src/SIL.XForge.Scripture/Models/SFProjectSecret.cs#L22)):*
> ```Smalltalk
>         /// Keeps track of all Paratext usernames that have been used to sync notes with Paratext
>         /// </summary>
>         public List<SyncUser> SyncUsers { get; set; } = new List<SyncUser>();
> ```

So, this code can't really be merged until the migration has been run, right? So when are you thinking this will be merged?



_Code quote:_
```Smalltalk
 public List<SyncUser> SyncUsers { get; set; } = new List<SyncUser>();
```
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1742 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGE3Pv4yRKTfQH6JJM:-MzGE3Pv4yRKTfQH6JJN:b-m9uxob) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad7640dd453cc2816bb4d601d3466aa80e45ae9/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1742)):*
> ```Smalltalk
>         {
>             if (string.IsNullOrEmpty(paratextUsername))
>                 return null;
> ```

Was this happening?



_Code quote:_
```Smalltalk
            if (string.IsNullOrEmpty(paratextUsername))
                return null;
```
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 774 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGP0z6AuVwkb2N859K:-MzGP0z6AuVwkb2N859L:b-14p5ec) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad7640dd453cc2816bb4d601d3466aa80e45ae9/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L774)):*
> ```Smalltalk
>                 NoteThreadChange change06 = changes.Where(c => c.ThreadId == "thread6").Single();
>                 Assert.That(change06.ThreadChangeToString(),
>                     Is.EqualTo("Context before Text selected thread6 context after.-Start:15-Length:21-MAT 1:6-conflict1"));
> ```

What's the meaning of the expected string here? 

If the change to `new ThreadComponents` above, where `username = env.Username01` is removed, which I assume means the thread has no username set, then it seems all the more likely that thread 6 is going to have a syncuser used to sync the note since we don't even know who authored the note. 



_Code quote:_
```Smalltalk
string expected6 = "thread6--user02-<p>thread6 note 1.</p>-conflict1";
```
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 791 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGPwEXD7CnwqmpSD-f:-MzGPwEXD7CnwqmpSD-g:b-xtwt2e) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad7640dd453cc2816bb4d601d3466aa80e45ae9/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L791)):*
> ```Smalltalk
> 
>                 // User 02 is added to the list of Paratext Users
>                 Assert.That(ptProjectUsers.Keys, Is.EquivalentTo(new[] { env.Username01, env.Username02 }));
> ```

I see that a dozen lines above, this same assert appears to be performed. And I also notice that before the SUT we didn't assert that something contrary was so, which would be good to do, to show that something changed.



_Code quote:_
```Smalltalk
                // User 02 is added to the list of Paratext Users
                Assert.That(ptProjectUsers.Keys, Is.EquivalentTo(new[] { env.Username01, env.Username02 }));
```
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 930 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGQMse6ALjZo63aMBY:-MzGQMse6ALjZo63aMBZ:b5i5phu) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad7640dd453cc2816bb4d601d3466aa80e45ae9/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L930)):*
> ```Smalltalk
>                 new ThreadComponents { threadNum = 7, noteCount = 1, username = env.Username01, notes = threadNotes7 },
>                 new ThreadComponents { threadNum = 8, noteCount = 1, username = env.Username01 },
>                 new ThreadComponents { threadNum = 9, noteCount = 1 }
> ```

What is the significance of adding this line? 



_Code quote:_
```Smalltalk
 new ThreadComponents { threadNum = 9, noteCount = 1 }
```
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 991 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGQVBj7so99UPS6knn:-MzGQVBj7so99UPS6kno:b-pc9cbj) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad7640dd453cc2816bb4d601d3466aa80e45ae9/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L991)):*
> ```Smalltalk
> 
>                 // User 02 is added to the list of Paratext Users
>                 Assert.That(ptProjectUsers.Keys, Is.EquivalentTo(new[] { env.Username01, env.Username02 }));
> ```

Is this related to the new line `  new ThreadComponents { threadNum = 9, noteCount = 1 }`? If so, can you explain in the comment "because [thus and such reason why we should  expect user 2 to be in the list now]"?

Also, can you assert before the SUT that ptProjectUsers.Keys is some other value. Then we can see that it changed.



_Code quote:_
```Smalltalk
                // User 02 is added to the list of Paratext Users
                Assert.That(ptProjectUsers.Keys, Is.EquivalentTo(new[] { env.Username01, env.Username02 }));
```
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 1283 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGSXjv0CfdlSTe5U-s:-MzGSXjv0CfdlSTe5U-t:b-syka7n) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad7640dd453cc2816bb4d601d3466aa80e45ae9/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L1283)):*
> ```Smalltalk
> 
>             // Add a conflict note
>             env.SetupNewConflictNoteThreadChange("conflictthread01");
> ```

What is the significance of this? 

Is this a conflict between two notes that have irreconcilable changes or data? Is it a special kind of note that signals a conflict in the chapter text?

Is there a significance that we are calling the SUT a second time in the same test? (eg are we building on something up above from the previous call to the SUT?)



_Code quote:_
```Smalltalk
 // Add a conflict note
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1273)* status: 7 of 8 files reviewed, 7 unresolved discussions (waiting on @marksvc)

___
*[src/SIL.XForge.Scripture/Models/SFProjectSecret.cs, line 22 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGCz-_5jc4l2ulzI0b:-MzH3OewDCMY-C44wddO:b-inela5) ([raw file](https://github.com/sillsdev/web-xforge/blob/72a0596010b44af527caa8b9bb790f79e5840c59/src/SIL.XForge.Scripture/Models/SFProjectSecret.cs#L22)):*

So, this code can't really be merged until the migration has been run, right? So when are you thinking this will be merged?
</blockquote></details>

Good spotting. We will be able to merge this code, but our migration tool needs to be built before this commit. Then when we got to migrate QA, we right the tool and ship `master` branch. From my perspective, that is the best way to introduce this change - do the migration and remove SyncUsers from our model all at the same time.
___
*[src/SIL.XForge.Scripture/Services/ParatextService.cs, line 1742 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGE3Pv4yRKTfQH6JJM:-MzH3wd82tDAy37H4-m5:buriw6d) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad7640dd453cc2816bb4d601d3466aa80e45ae9/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1742)):*

Was this happening?
</blockquote></details>

The paratextUsername can be the empty string if the note isn't associated with a user (this is true for conflict notes, and likely others).
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 774 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGP0z6AuVwkb2N859K:-MzH5jbkBVHVS5wZbncK:b7ph6m6) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad7640dd453cc2816bb4d601d3466aa80e45ae9/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L774)):*

What's the meaning of the expected string here? 

If the change to `new ThreadComponents` above, where `username = env.Username01` is removed, which I assume means the thread has no username set, then it seems all the more likely that thread 6 is going to have a syncuser used to sync the note since we don't even know who authored the note. 
</blockquote></details>

When we do not assign the username to a note, our test will give the note the empty string for the note's author. This is perfectly valid since some notes (conflict notes for instance) do not have a Paratext user associated with it. We do not expect these notes to have an form of sync user associated with it either. That field will be left null.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 791 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGPwEXD7CnwqmpSD-f:-MzH7BJT9UXDn8zxn7aD:b7i0eox) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad7640dd453cc2816bb4d601d3466aa80e45ae9/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L791)):*

I see that a dozen lines above, this same assert appears to be performed. And I also notice that before the SUT we didn't assert that something contrary was so, which would be good to do, to show that something changed.
</blockquote></details>

Good eye. I'll remove the first instance of the assert. Regarding asserting what ptProjectUsers is before the SUT, I don't think that is necessary since we specifically assign its value in this test.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 930 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGQMse6ALjZo63aMBY:-MzHCEy46MFSS7s61EgS:bj4j1hx) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad7640dd453cc2816bb4d601d3466aa80e45ae9/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L930)):*

What is the significance of adding this line? 
</blockquote></details>

Looks like I forgot to add some asserts. I've added those now.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 991 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGQVBj7so99UPS6knn:-MzHAI6d9-w3101VkQvx:b-2ssiw3) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad7640dd453cc2816bb4d601d3466aa80e45ae9/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L991)):*

Is this related to the new line `  new ThreadComponents { threadNum = 9, noteCount = 1 }`? If so, can you explain in the comment "because [thus and such reason why we should  expect user 2 to be in the list now]"?

Also, can you assert before the SUT that ptProjectUsers.Keys is some other value. Then we can see that it changed.
</blockquote></details>

This was supposed to show that no sync user was introduced when a note was created but not associated with a paratext user. I added those asserts. I don't see a reason to assert that ptProjectUsers has just "User 01" in it since it is clearly assigned above.
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs, line 1283 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGSXjv0CfdlSTe5U-s:-MzHBJjG-uQOvzEjyn_F:bl79ama) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad7640dd453cc2816bb4d601d3466aa80e45ae9/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L1283)):*

What is the significance of this? 

Is this a conflict between two notes that have irreconcilable changes or data? Is it a special kind of note that signals a conflict in the chapter text?

Is there a significance that we are calling the SUT a second time in the same test? (eg are we building on something up above from the previous call to the SUT?)
</blockquote></details>

This test simply shows that a conflict note gets added to our note thread docs at sync time. I was using TDD to build this behaviour. I could have set up a separate unit test for this, but it was so easy to just reused the setup that was already done for this test.

marksvc (changes requested):
> Reviewed 1 of 8 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1273)* status: all files reviewed, 4 unresolved discussions (waiting on @RaymondLuong3)

___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 774 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGP0z6AuVwkb2N859K:-MzHTZq_CgV0JlIVqlsw:b-g2kqlh) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad7640dd453cc2816bb4d601d3466aa80e45ae9/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L774)):*

When we do not assign the username to a note, our test will give the note the empty string for the note's author. This is perfectly valid since some notes (conflict notes for instance) do not have a Paratext user associated with it. We do not expect these notes to have an form of sync user associated with it either. That field will be left null.
</blockquote></details>

Hmm. So if there is no associated PT user, why does  "user02" appear in the  string? I see that's from note ExtUserId. And that it may come from a comment ExternalUser. 
Oh is that maybe an _SF_ user id then?
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 791 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGPwEXD7CnwqmpSD-f:-MzHW-279XZsl4EKo3x5:b-h3yeiw) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad7640dd453cc2816bb4d601d3466aa80e45ae9/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L791)):*

Good eye. I'll remove the first instance of the assert. Regarding asserting what ptProjectUsers is before the SUT, I don't think that is necessary since we specifically assign its value in this test.
</blockquote></details>

Thank you, I see that now. One thing that isn't very clear from this being down here is _why_ env.Username02 should now be in the ptProjectUsers list. Is it that   thread 3's changes (and _not_ thread 6, right?) would  have caused this? If that's so, maybe the comment could say something like "User 02 is added to the list of Paratext Users, because of the changes from  thread 3."

Oh and if this, too, is to show that a change didn't add something to the list, it would be good to specific about that as well, like adding "And no other user is added as a result of the new thread 6 because it  is not associated with a PT user."
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 991 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzGQVBj7so99UPS6knn:-MzHXzAYAO8F8sEBsCUR:b-j7qkkr) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad7640dd453cc2816bb4d601d3466aa80e45ae9/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L991)):*
> assigned above 

Thank you, I have found that now. 

Why is env.Username02 now in the list of ptProjectUsers? Is it because of the thread 9 change? Oh that wouldn't make sense, given your statement above. 
Hmm. Is it because the **thread note components** specify a username02 assignment? But that's assignment, not related to a sync user. 

> supposed to show that no sync user was introduced when a note was created but not associated with a paratext user

That's good. But I think it will be better if we write something about that in the test. If the SUT started to add a new user to the list, I want to know "this new addition is wrong" rather than "oops; a change means I need to update this test's assert". Maybe it could say "User 02 is added to the list of Paratext Users because of ???. And no other user is added as a result of the new thread 9 because it  is not associated with a PT user."
___
*[test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs, line 2344 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1273#-MzHdmjV468M146MSnVc:-MzHdmjV468M146MSnVd:b5wvlse) ([raw file](https://github.com/sillsdev/web-xforge/blob/8f455becc738df3be94342493b37c162796db479/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2344)):*
> ```Smalltalk
>                 {
>                     string threadId = "thread" + comp.threadNum;
>                     var associatedPtUser = new SFParatextUser(comp.username ?? "");
> ```

What's going on here? We make a ParatextUser with an empty username? And that's what's used to construct a PT Comment rather than a null?



_Code quote:_
```Smalltalk
   var associatedPtUser = new SFParatextUser(comp.username ?? "");
```

PR #1282, opened 2022-03-25
SF-1516a Use IndexedDB getAll instead of looping

Nateowami:
This is an initial fix to improve performance. The change was originally suggested by Damien when we were investigating slow question import, but didn't get made at the time. As I've been investigating slowness in community checking, my attention was directed to this method by the debugger (if you pause the debugger when the page is slow, it often shows this method in the stack, which made it look like almost all of the time was being spent here).

Manual testing shows when you load a page from scratch, the time to first paint for 1500 questions is about 85% of what it was before this change. That's not massive, though for a change this small, it's surprising it makes even that much of a difference.

Here's my testing data (time for questions to show on screen, after entering URL in new tab, listed in seconds).

Test # | Before | After
-------|--------|------
1 | 42.5 | 35.9
2 | 42.7 | 36.96
3 | 42.6 | 35.54
4 | 42.95 | 35.85
5 | 42.34 | 36.08

Again, it's not a huge change, but it certainly seems to make a difference. I've found ways of affecting performance way more significantly than this, but this is low hanging fruit and I wanted to get it out of the way.

I also updated the browser version requirements. The getAll method is not available in older IndexedDB implementations. Sources for this are [MDN](https://developer.mozilla.org/en-US/docs/Web/API/IDBObjectStore/getAll) and [Can I Use](https://caniuse.com/indexeddb2).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1282)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/indexeddb-offline-store.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/indexeddb-offline-store.ts
@@ -12,18 +12,9 @@ function getAllFromCursor<T extends OfflineData>(
   query?: IDBValidKey | IDBKeyRange
 ): Promise<T[]> {
   return new Promise<T[]>((resolve, reject) => {
-    const results: T[] = [];
-    const request = store.openCursor(query);
+    const request = store.getAll(query);
     request.onerror = () => reject(request.error);
-    request.onsuccess = () => {
-      const cursor = request.result;
-      if (cursor != null) {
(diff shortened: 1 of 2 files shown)


irahopkinson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1282#-:-MyzJTFR9_3vvwsun7Go:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1284, opened 2022-03-28
SF-1521 Only highlight translate target editor when user can edit

RaymondLuong3:
When a user clicks on the target editor, the focused event is fired even though the target editor never truly gets focused when it is read only. This fix is a work-around that sets the text component to not add yellow highlight to verses when the editor is read only. This restores the previous behaviour before we regressed this behaviour with another work-around to set the targetFocused field on the editor component.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1284)


marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1284)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1291, opened 2022-03-30
SF-1527 Trim out unproductive ops when deleting duplicate notes

RaymondLuong3:
Sometimes a delta with a note insert means that the user is undoing previously deleting a note icon. Since the note gets recreated when deleted, the recreated note is marked for deletion. However, if a user deletes the final character of a segment, any note immediately following the next verse embed gets marked for deletion because Quill diff mechanism doesn't accurately represent the change applied to the text.

For instance, if the true delta applied to a text is [ retain: 10, insert: blank ], then the delta emitted by the Quill.onContentChanged will be [ retain: 10, insert: blank, insert: verse, insert: note, delete: 2 ]. This change trims the unproductive false ops from the delta and uses that to determine if a note needs to be marked for deletion.

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -10,7 +10,7 @@ import {
   ViewChild
 } from '@angular/core';
 import { TranslocoService } from '@ngneat/transloco';
-import { clone } from 'lodash-es';
+import { clone, cloneDeep } from 'lodash-es';
 import isEqual from 'lodash-es/isEqual';
 import merge from 'lodash-es/merge';
 import Quill, { DeltaOperation, DeltaStatic, RangeStatic, Sources } from 'quill';
@@ -967,10 +967,14 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     }
     // Delta for the removal of notes that were re-created
     let notesDeletionDelta: DeltaStatic | undefined;
-    for (const op of delta.ops) {
+    const productiveOps = this.trimUnproductiveOps(delta);
+    if (productiveOps.ops == null) {
+      return;
+    }
+    for (const op of productiveOps.ops) {
       if (op.insert != null && op.insert['note-thread-embed'] != null) {
         const embedId: string = op.insert['note-thread-embed']['threadid'];
-        let deletePosition = this.embeddedElements.get(embedId);
+        const deletePosition = this.embeddedElements.get(embedId);
         if (deletePosition != null) {
           const noteDeleteOps: DeltaOperation[] = [{ retain: deletePosition }, { delete: 1 }];
           const noteDeleteOpDelta = new Delta(noteDeleteOps);
@@ -987,6 +991,42 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
         this.editor?.updateContents(deleteDelta, 'api');
       });
     }
+  }
+
+  /**
+   * Trims out the unproductive ops from a delta emitted by Quill's onContentChanged event.
+   */
+  private trimUnproductiveOps(delta: DeltaStatic): DeltaStatic {
+    // The quill way of determining changes applied to its content doesn't work perfectly on non-text objects
+    // and the result in the delta is a series of object inserts followed by a delete op that removes
+    // the original object inserts. This trims out those ops from this delta so the delta is a true representation
+    // of the changes applied to the text.
+    // For example, a delta may contain [ retain: 10, insert: blank, insert: verse, insert: note, delete: 2 ]
+    // where the delete: 2 op deletes the original verse and note inserts that get replaced in this delta
+    const productiveOps: DeltaStatic = cloneDeep(delta);
+    if (productiveOps.ops == null || productiveOps.ops.length < 2) {
+      return delta;
+    }
+    const opCount: number = productiveOps.ops.length;
+    // find the trailing delete op
+    const lastDeleteOp: number | undefined = productiveOps.ops[opCount - 1].delete;
+    let deleteCount: number = lastDeleteOp == null ? 0 : lastDeleteOp;
+    if (deleteCount === 0) {
+      return delta;
+    }
+
+    for (let i = 0; i < deleteCount; i++) {
+      const curIndex: number = opCount - i - 2;
+      const insertObj = productiveOps.ops[curIndex].insert;
+      if (insertObj != null && typeof insertObj !== 'string') {
+        // the object insert may be a false op
+        continue;
+      }
+      // the op is productive, so keep the entire delta
+      return delta;
+    }
+    productiveOps.ops.splice(opCount - 1 - deleteCount);
+    return productiveOps;
   }
 
   private setHighlightMarkerPosition(): void {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -1972,6 +1972,30 @@ describe('EditorComponent', () => {
       env.dispose();
     }));
 
+    it('can backspace the last character in a segment', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig();
+      env.wait();
+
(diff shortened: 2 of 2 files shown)


nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1291)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 1013 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1291#-N-LnGxD0O1Eb2Jfjjbr:-N-LnGxD0O1Eb2Jfjjbs:b-2mvopa) ([raw file](https://github.com/sillsdev/web-xforge/blob/d94ab73602f089935f1fe55b5a8a57c4b44f096b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1013)):*
> ```XML
>     // find the trailing delete op
>     const lastDeleteOp: number | undefined = productiveOps.ops[opCount - 1].delete;
>     let deleteCount: number = lastDeleteOp == null ? 0 : lastDeleteOp;
> ```

Could this be `lastDeleteOp ?? 0` ?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 1019 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1291#-N-Lnz-eCl-RkfQstONz:-N-Lnz-eCl-RkfQstOO-:b-fngz7d) ([raw file](https://github.com/sillsdev/web-xforge/blob/d94ab73602f089935f1fe55b5a8a57c4b44f096b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1019)):*
> ```XML
> 
>     for (let i = 0; i < deleteCount; i++) {
>       const curIndex: number = opCount - i - 2;
> ```

Is this specifically setup to only ever check the last couple of ops - will that always be the case?
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 1028 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1291#-N-Lp4AMBGed-2a8VwRV:-N-Lp4AMBGed-2a8VwRW:b-n0w3qj) ([raw file](https://github.com/sillsdev/web-xforge/blob/d94ab73602f089935f1fe55b5a8a57c4b44f096b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1028)):*
> ```XML
>       return delta;
>     }
>     productiveOps.ops.splice(opCount - 1 - deleteCount);
> ```

So, we're saying that regardless of inserts being found we will trim them off assuming they're not required? I guess if this is only ever going to be called by the `deleteDuplicateNoteIcon` method then we can make some assumptions - perhaps this method should be renamed to better imply that or change the method description to state that this currently only applies to notes

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1291)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 1013 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1291#-N-LnGxD0O1Eb2Jfjjbr:-N-POPNq-uYlOm2qrOz7:b-9fmn3c) ([raw file](https://github.com/sillsdev/web-xforge/blob/d94ab73602f089935f1fe55b5a8a57c4b44f096b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1013)):*

Could this be `lastDeleteOp ?? 0` ?
</blockquote></details>

Good suggestion. Done.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 1019 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1291#-N-Lnz-eCl-RkfQstONz:-N-POdbKEIlFFbN-nwAC:blhpzpn) ([raw file](https://github.com/sillsdev/web-xforge/blob/d94ab73602f089935f1fe55b5a8a57c4b44f096b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1019)):*

Is this specifically setup to only ever check the last couple of ops - will that always be the case?
</blockquote></details>

It is setup to check the insert ops from the second-to-last position and moving forward 1 position at a time for the length of the delete op that occupies the final op position. Generally the delete will have a length of 2 - 3. It could be a blank-verse-note series, or a blank-verse series. In some situations, it could be much longer if the text is mostly blank.
___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 1028 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1291#-N-Lp4AMBGed-2a8VwRV:-N-PS4HqE7Uv5uTJNNV0:bswdshn) ([raw file](https://github.com/sillsdev/web-xforge/blob/d94ab73602f089935f1fe55b5a8a57c4b44f096b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1028)):*

So, we're saying that regardless of inserts being found we will trim them off assuming they're not required? I guess if this is only ever going to be called by the `deleteDuplicateNoteIcon` method then we can make some assumptions - perhaps this method should be renamed to better imply that or change the method description to state that this currently only applies to notes
</blockquote></details>

It is not ideal to have to handle delta's in this way, but at least we are not depending on this method for anything other than detecting if a note embed should result in deleting a duplicate one that already exists in the text editor. This does check that each of the ops that it 'trims' away is an object and not text. So we can be confident that it will return that right thing because a user cannot manually insert objects, only texts, so this is unlikely to trim real productive ops.

nigel-wells (commented):
Happy with this - I've not run tests myself but it sounds like it should do what it is supposed to do.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1291#-:-N-ZuvQjAjgMa9H9xvBj:b-l9ih1y)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 1028 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1291#-N-Lp4AMBGed-2a8VwRV:-N-ZuQZs4XqQKojBgBR-:b4t2xfo) ([raw file](https://github.com/sillsdev/web-xforge/blob/d94ab73602f089935f1fe55b5a8a57c4b44f096b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1028)):*

It is not ideal to have to handle delta's in this way, but at least we are not depending on this method for anything other than detecting if a note embed should result in deleting a duplicate one that already exists in the text editor. This does check that each of the ops that it 'trims' away is an object and not text. So we can be confident that it will return that right thing because a user cannot manually insert objects, only texts, so this is unlikely to trim real productive ops.
</blockquote></details>

What will happen in the future when we can insert notes and thereby a user can insert objects? I guess we can address anything needed then so long as our tests are in place for these scenarios.

nigel-wells (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1291#-:-N-ZvE_A9bNTTMYibBoE:bnfp4nl)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @nigel-wells)

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1291)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts, line 1028 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1291#-N-Lp4AMBGed-2a8VwRV:-N-_0HJXCfdjYk5wK4fU:byfa0g8) ([raw file](https://github.com/sillsdev/web-xforge/blob/d94ab73602f089935f1fe55b5a8a57c4b44f096b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1028)):*

What will happen in the future when we can insert notes and thereby a user can insert objects? I guess we can address anything needed then so long as our tests are in place for these scenarios.
</blockquote></details>

That's right, we will need to address that then when a user can insert a note in Scripture Forge.

PR #1293, opened 2022-03-31
SF-1516b Improve performance by enabling ngZoneEventCoalescing

Nateowami:
This is not the road I have been going down for improving performance, but I came across it tonight and it makes a massive difference, likely greater than my more significant changes. So I briefly threw my other changes on the back burner to test this out first. There's still significant slowness in the checking component when there are lots of questions, but it's a lot better.

Essentially we're telling Angular that if an event occurs, and it knows another one is about to fire, to just wait until they've both fired before running change detection (coalescing events).

See this article by Netanel Basal (the guy behind Transloco and a bunch of other helpful Angular articles I've read) for a deeper explanation of what this does: https://netbasal.com/reduce-change-detection-cycles-with-event-coalescing-in-angular-c4037199859f

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/main.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/main.ts
@@ -18,5 +18,5 @@ if (environment.production || environment.pwaTest) {
 ExceptionHandlingService.initBugsnag();
 
 platformBrowserDynamic(providers)
-  .bootstrapModule(AppModule)
+  .bootstrapModule(AppModule, { ngZoneEventCoalescing: true })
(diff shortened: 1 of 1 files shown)


RaymondLuong3 (approved):
You found a really good feature of Angular. I'm glad that they thought of this, and I can see why it helps so much in our situation. I approve!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1293#-:-MzVfF2843mHL-Qx_sCE:bp71rkl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1140, opened 2021-09-10
SF-1400 Sort questions by canonical order

Nateowami:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1140)

PR #1294, opened 2022-03-31
SF-1400 Sort questions in query rather than component

Nateowami:
While working on improving performance in the Angular app, I discovered that the way I had implemented sorting questions was non-ideal for several reasons. We already had a way to sort questions in a query, and rather than use that, I had rolled my own. Questions would be sorted every time the `get questionDocs()` getter was used, which would be multiple times per round of change detection. This is not ideal, since sorting large numbers of things is slow, and there's no real reason to re-sort them. It also means if another component wanted sorted questions the same sorting logic could not be easily re-used.

Also I was filtering for questions that are not archived, which should have been done in the query itself.

Unfortunately, my original change involved somewhat confusing changes to the tests. So I just reverted that entire commit, and then made my changes afterwards. I think it will be easiest to review by looking at the commits separately. The PR I'm reverting is #1140.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1294)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
@@ -237,7 +237,7 @@ describe('CheckingComponent', () => {
     it('responds to remote community checking disabled when observer', fakeAsync(() => {
       // User with access to translate app should get redirected there
       const env = new TestEnvironment(OBSERVER_USER, 'ALL');
-      env.selectQuestion(2);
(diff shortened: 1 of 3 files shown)


RaymondLuong3 (approved):
Thanks for the explanation. I'm glad we found this which will make the sorted questions easily accessible to all components.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1294#-:-MzWzzh10u8q6w6fa-4q:b-uwr4kp)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1299, opened 2022-04-08
SF-1531 - The error appears when opening the note icon if a note is created inside the special characters

nigel-wells:
- Adjust the order of quill inlines to ensure text-anchor is deeper than char
- Added styling for usx-char `em` style

The issue was how Quill ordered elements in the DOM and it placed the `<display-text-anchor>` above `<usx-char>` elements:
image

After adjusting the order it now appears correctly and the `usx-char` can be styled correctly as well:
image

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1299)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
@@ -329,7 +329,8 @@ usx-segment:first-child usx-blank {
 }
 
 usx-char {
-  &[data-style='it'] {
+  &[data-style='it'],
+  &[data-style='em'] {
     font-style: italic;
   }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
@@ -509,7 +509,10 @@ export function registerScripture(): string[] {
   }
   formats.push(TextAnchorInline);
 
+  // Lower index means deeper in the DOM tree i.e. text-anchor will be nested inside of char. If char doesn't exist
+  // then it will nest inside the next available element higher up the DOM
   (Inline as any).order.push('text-anchor');
+  (Inline as any).order.push('char');
   (Inline as any).order.push('segment');
   (Inline as any).order.push('para-contents');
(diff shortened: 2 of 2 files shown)


RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1299)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts, line 512 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1299#-N-OmrO26B1VaeWEnhEJ:-N-OmrO26B1VaeWEnhEK:boiis8p) ([raw file](https://github.com/sillsdev/web-xforge/blob/39cd083404645b153c4a6df969a04b8e12ff5e81/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L512)):*
> ```XML
>   formats.push(TextAnchorInline);
> 
>   // Lower index means deeper in the DOM tree i.e. text-anchor will be nested inside of char. If char doesn't exist
> ```

Great job figuring this out. I wasn't aware of this feature of quill, but I'm glad we know about it now.



_Code quote:_
```XML
  // Lower index means deeper in the DOM tree i.e. text-anchor will be nested inside of char. If char doesn't exist
```

Nateowami (commented):
@nigel-wells For PR titles, I think it would be helpful to reword the issue to more clearly explain what the actual problem and/or solution is, rather than just using the title from the Jira issue, which is often somewhat difficult to understand (and may not fully reflect the actual problem).

For example, I think this PR could be described as "Fix error opening note inside usx-char tag". This makes it a lot easier to figure out what each commit changes when looking at a commit log. I *think* this description is correct, so I'll go ahead with this message and merge it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1299#-:-N-PGJ1nD1g17LZqmvao:bd95901)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

nigel-wells:
Yes fair comment @Nateowami  - I'll take that on board for my next PR.

PR #1303, opened 2022-04-12
notes: don't remove the wrong closing tag

marksvc:
- The wrong closing tag can be removed. We may find success removing a
closing tag that matches an opening tag, but we can just remove all
unknown opening and closing tags, as with this change.

commit-id:ce153c40

---

**Stack**:
- #1311
- #1308
- #1307
- #1306
- #1305
- #1304
- #1303 ⮜

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts
@@ -82,6 +82,10 @@ describe('NoteDialogComponent', () => {
         expected: 'turn <i>text italic</i>'
       },
       {
+        text: 'Alpha <unknown><bold>Bravo</bold></unknown> Charlie',
+        expected: 'Alpha <b>Bravo</b> Charlie'
+      },
+      {
         text: '<p>this is a paragraph</p>',
         expected: 'this is a paragraph<br />'
       },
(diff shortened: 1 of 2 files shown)


marksvc:
*[src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts, line 137 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1303#-N-Yn85oBTyeSnYlZLnd:-N-ZmP7z2Y2yDbjYj_Wb:bx4970o) ([raw file](https://github.com/sillsdev/web-xforge/blob/ea8357ab95e2162728ace9aa701d2ba5891ae67a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L137)):*

Where do you see <span> used where we did not account for it before?
</blockquote></details>

I think the problem may have related to processing content with a series of diff indicators. If an unknown tag began before a tag that we wanted to keep, we would remove the wrong closing tag. A simple example is something like the following, where `language` is an unknown tag, but `bold` is something to keep and process.
`<language><bold>blah</bold></language> -> <bold>blah</language>`
(And a closer to real-data example would be `Bob modified this on two computers. <p><language><p><strikethru><color>ABC</color></strikethru></p></language></p>`.)

We could make the regex be more clever and only remove closing tags _that match the opening unknown tag_. I had a go at doing that but found that just removing all unknown starting and closing tags was working.
Nathaniel and I also started thinking about other ways to process the data that didn't involve regex, such as using an HTML parser.

PR #1327, opened 2022-04-26
Remove XRegExp dependency

Nateowami:
We've been using XRegExp for two things:
1. Escaping strings to use them as part of a regex
2. [Unicode property escapes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Unicode_Property_Escapes)

The former can be done using Lodash, and the latter is now supported in all modern browsers with the `u` flag added to a regex.

XRegExp is fairly sizeable (41.08 KB gzipped out of our total gzipped bundle size of 705.45 KB, making it 5.82% of our bundle) and we're not using it for much, so getting rid of it seemed like a good idea.

Unfortunately it's also used by @sillsdev/machine, which is one of our dependencies, so there's no actual reduction in bundle size. I didn't realize that until I'd made the change though, and it seems like something that's likely to eventually be removed from @sillsdev/machine, so I think at some point in the future we will see the slight size reduction, and there's no real advantage to continuing to use it.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1327)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/package-lock.json
+++ b/src/SIL.XForge.Scripture/ClientApp/package-lock.json
@@ -54,7 +54,6 @@
         "sharedb": "1.0.0-beta.31",
         "tslib": "^2.4.0",
         "uuid": "^8.3.2",
-        "xregexp": "^5.1.0",
         "zone.js": "^0.11.5"
       },
       "devDependencies": {
--- a/src/SIL.XForge.Scripture/ClientApp/package.json
+++ b/src/SIL.XForge.Scripture/ClientApp/package.json
@@ -65,7 +65,6 @@
     "sharedb": "1.0.0-beta.31",
     "tslib": "^2.4.0",
     "uuid": "^8.3.2",
-    "xregexp": "^5.1.0",
     "zone.js": "^0.11.5"
   },
   "devDependencies": {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
@@ -39,7 +39,6 @@ import { NoticeService } from 'xforge-common/notice.service';
 import { PwaService } from 'xforge-common/pwa.service';
 import { UserService } from 'xforge-common/user.service';
 import { getLinkHTML, issuesEmailTemplate } from 'xforge-common/utils';
-import XRegExp from 'xregexp';
 import { SFProjectProfileDoc } from '../../core/models/sf-project-profile-doc';
 import { environment } from '../../../environments/environment';
 import { NoteThreadDoc } from '../../core/models/note-thread-doc';
@@ -63,7 +62,7 @@ import { TranslateMetricsSession } from './translate-metrics-session';
 
 export const UPDATE_SUGGESTIONS_TIMEOUT = 100;
 
-const PUNCT_SPACE_REGEX = XRegExp('^(\\p{P}|\\p{S}|\\p{Cc}|\\p{Z})+$');
+const PUNCT_SPACE_REGEX = /^(?:\p{P}|\p{S}|\p{Cc}|\p{Z})+$/u;
 
 /** Scripture editing area. Used for Translate task. */
 @Component({
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts
@@ -1,9 +1,8 @@
-import merge from 'lodash-es/merge';
+import { escapeRegExp, merge } from 'lodash-es';
 import { Project } from 'realtime-server/lib/esm/common/models/project';
 import { ProjectRole } from 'realtime-server/lib/esm/common/models/project-role';
 import { combineLatest, Observable, Subject } from 'rxjs';
 import { debounceTime, distinctUntilChanged, switchMap } from 'rxjs/operators';
-import XRegExp from 'xregexp';
 import { CommandService } from './command.service';
 import { ProjectDoc } from './models/project-doc';
 import { NONE_ROLE, ProjectRoleInfo } from './models/project-role-info';
@@ -49,7 +48,7 @@ export abstract class ProjectService<
 
     return combineLatest([debouncedTerm$, queryParameters$]).pipe(
       switchMap(([term, queryParameters]) => {
-        term = XRegExp.escape(term.trim());
+        term = escapeRegExp(term.trim());
         let filters: Filters = {};
         if (term.length > 0) {
           filters = {
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/system-administration/sa-projects.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/system-administration/sa-projects.component.spec.ts
@@ -4,6 +4,7 @@ import { MatCheckbox } from '@angular/material/checkbox';
 import { By } from '@angular/platform-browser';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { RouterTestingModule } from '@angular/router/testing';
+import { escapeRegExp } from 'lodash-es';
 import merge from 'lodash-es/merge';
 import { Project } from 'realtime-server/lib/esm/common/models/project';
 import { obj } from 'realtime-server/lib/esm/common/utils/obj-path';
@@ -12,7 +13,6 @@ import { switchMap } from 'rxjs/operators';
 import { anything, mock, verify, when } from 'ts-mockito';
 import { FileType } from 'xforge-common/models/file-offline-data';
 import { TestRealtimeModule } from 'xforge-common/test-realtime.module';
-import XRegExp from 'xregexp';
 import { ProjectDoc } from '../models/project-doc';
 import { NONE_ROLE, ProjectRoleInfo } from '../models/project-role-info';
 import { NoticeService } from '../notice.service';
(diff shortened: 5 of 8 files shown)


RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1327)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 67 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1327#-N0ggKgx9I63P8dUcf0V:-N0ggKgx9I63P8dUcf0W:b-9b41) ([raw file](https://github.com/sillsdev/web-xforge/blob/4a588520d373b51ed1c23d592f1e664988c7d309/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L67)):*
> ```XML
> 
> const PUNCT_SPACE_REGEX = /^(?:\p{P}|\p{S}|\p{Cc}|\p{Z})+$/u;
> ```

What does the `?:` do that we weren't doing before?



_Code quote:_
```XML
const PUNCT_SPACE_REGEX = /^(?:\p{P}|\p{S}|\p{Cc}|\p{Z})+$/u;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1327#-N0gg1Oa5vOdr_LGkIf_:-N0gg1Oa5vOdr_LGkIfa:b-w17oeh) ([raw file](https://github.com/sillsdev/web-xforge/blob/4a588520d373b51ed1c23d592f1e664988c7d309/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts#L23)):*
> ```XML
> 
> export function supportedBrowser(): boolean {
>   // See https://caniuse.com/#feat=indexeddb2 for browsers supporting IndexedDB 2.0
> ```

Where did you see these new values. This website says those previous values are still ok.



_Code quote:_
```XML
https://caniuse.com/#feat=indexeddb2
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1327)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 67 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1327#-N0ggKgx9I63P8dUcf0V:-N0gk9qo8odQ1LtjeIpu:bn0jk0e) ([raw file](https://github.com/sillsdev/web-xforge/blob/4a588520d373b51ed1c23d592f1e664988c7d309/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L67)):*

What does the `?:` do that we weren't doing before?
</blockquote></details>

It makes it a non-capturing group. It shouldn't change whether it matches any string. Theoretically it might be _slightly_ faster. Generally you don't want a repeating capturing group, since it only captures the last iteration. In some cases that might be the desired behavior.

By the way, if you put a regex into regex101.com (make sure to select the programming language you're using) it will explain what each thing in the regex does.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1327#-N0gg1Oa5vOdr_LGkIf_:-N0gl9M48gMvNDsf3Mjw:b9j06tl) ([raw file](https://github.com/sillsdev/web-xforge/blob/4a588520d373b51ed1c23d592f1e664988c7d309/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts#L23)):*

Where did you see these new values. This website says those previous values are still ok.
</blockquote></details>

https://caniuse.com/mdn-javascript_builtins_regexp_property_escapes

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1327)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1327#-N0gg1Oa5vOdr_LGkIf_:-N0glhCb1JVGR6o9ml0x:b5n1k7k) ([raw file](https://github.com/sillsdev/web-xforge/blob/4a588520d373b51ed1c23d592f1e664988c7d309/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts#L23)):*

https://caniuse.com/mdn-javascript_builtins_regexp_property_escapes
</blockquote></details>

I've removed the comment since it sort of implies (incorrectly) that the values chosen here are only chosen based on IndexdDB 2.0 support. In reality they've been modified over time for various purposes, always increasing the minimum required version number.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1327)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1327#-N0gg1Oa5vOdr_LGkIf_:-N0lgiyQ-WfP4-ESHgTC:b-jyo8ox) ([raw file](https://github.com/sillsdev/web-xforge/blob/4a588520d373b51ed1c23d592f1e664988c7d309/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts#L23)):*

I've removed the comment since it sort of implies (incorrectly) that the values chosen here are only chosen based on IndexdDB 2.0 support. In reality they've been modified over time for various purposes, always increasing the minimum required version number.
</blockquote></details>

Thanks for updating the supported browsers and removing the comment. I prefer leaving a comment here so we can immediately have a reference for these browser values, but we should be able to search for this comment when the time comes.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1327)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1327#-N0gg1Oa5vOdr_LGkIf_:-N0lltce-VQlc3cGU7f_:bmq23mo) ([raw file](https://github.com/sillsdev/web-xforge/blob/4a588520d373b51ed1c23d592f1e664988c7d309/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts#L23)):*

Thanks for updating the supported browsers and removing the comment. I prefer leaving a comment here so we can immediately have a reference for these browser values, but we should be able to search for this comment when the time comes.
</blockquote></details>

I don't quite understand what you mean about what we should do with the comments. Maybe each time we adjust these values we should add another comment explaining why they were increased (e.g. leave a link to each caniuse feature we're basing the decision on)?

RaymondLuong3 (approved):
> Reviewed 2 of 2 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1327)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1327#-N0gg1Oa5vOdr_LGkIf_:-N0lvTv2BeSNSpT8fBao:b5v1tri) ([raw file](https://github.com/sillsdev/web-xforge/blob/4a588520d373b51ed1c23d592f1e664988c7d309/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts#L23)):*

I don't quite understand what you mean about what we should do with the comments. Maybe each time we adjust these values we should add another comment explaining why they were increased (e.g. leave a link to each caniuse feature we're basing the decision on)?
</blockquote></details>

That would be my preference. Let's add that in just in case it becomes a useful note.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1327)* status: 7 of 8 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1327#-N0gg1Oa5vOdr_LGkIf_:-N0ly91T9FIdHpyPihRK:b5biiue) ([raw file](https://github.com/sillsdev/web-xforge/blob/4a588520d373b51ed1c23d592f1e664988c7d309/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts#L23)):*

That would be my preference. Let's add that in just in case it becomes a useful note.
</blockquote></details>

Done. I went back and looked at when we changed this and why. I thought there were more changes, but it looks like these two were the only features that dictated the version numbers.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1327)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1327#-N0gg1Oa5vOdr_LGkIf_:-N0mF2IZ8JD7N6pjGBvB:b5dofh7) ([raw file](https://github.com/sillsdev/web-xforge/blob/4a588520d373b51ed1c23d592f1e664988c7d309/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts#L23)):*

Done. I went back and looked at when we changed this and why. I thought there were more changes, but it looks like these two were the only features that dictated the version numbers.
</blockquote></details>

I like it. Thanks.

PR #1330, opened 2022-04-27
Search for projects by shortName on sysadmin page

Nateowami:
Very often we have a project shortname, rather than a project name. The current system admin page only allows searching by name. This commit enables searching by shortname.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1330)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts
@@ -76,7 +76,7 @@ export class CheckingQuestionsComponent extends SubscriptionDisposable {
     this.projectProfileDocChangesSubscription?.unsubscribe();
     this.project = projectProfileDoc?.data;
     if (projectProfileDoc != null) {
-      this.projectProfileDocChangesSubscription = this.subscribe(projectProfileDoc?.changes$, () => {
+      this.projectProfileDocChangesSubscription = this.subscribe(projectProfileDoc.changes$, () => {
         this.changeDetector.markForCheck();
       });
     }
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts
@@ -1,7 +1,6 @@
 import merge from 'lodash-es/merge';
 import { Project } from 'realtime-server/lib/esm/common/models/project';
 import { ProjectRole } from 'realtime-server/lib/esm/common/models/project-role';
-import { obj } from 'realtime-server/lib/esm/common/utils/obj-path';
 import { combineLatest, Observable, Subject } from 'rxjs';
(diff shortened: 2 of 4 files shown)


RaymondLuong3 (commented):
Is there an easy way to test this behaviour? I see that the changes in `project.service.ts` are not covered.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1330#-:-N0ghbiP8y9YwJ1iY24u:b-i8tuoa)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts` line 56 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1330#-N0ggzEC4o9cVrFmPLJ4:-N0ggzEC4o9cVrFmPLJ5:bxtpj83) ([raw file](https://github.com/sillsdev/web-xforge/blob/572944b9f15b2acb4a70f96eb88c8ea33403cb6f/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts#L56)):*
> ```XML
>         if (term.length > 0) {
>           filters = {
>             $or: termMatchProperties.map(prop => ({ [prop]: { $regex: term, $options: 'i' } }))
> ```

So we no longer need to specify that the match is anywhere in the string. Any idea why we had that in the first place?



_Code quote:_
```XML
$regex: term,
```

Nateowami (commented):
I thought there was a test covering this. Looking into it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1330#-:-N0gnixO2TEV3wlOxPKx:b-5fznpw)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts` line 56 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1330#-N0ggzEC4o9cVrFmPLJ4:-N0gjfLfE1Wt_9e-wnB9:bdcm257) ([raw file](https://github.com/sillsdev/web-xforge/blob/572944b9f15b2acb4a70f96eb88c8ea33403cb6f/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/project.service.ts#L56)):*

So we no longer need to specify that the match is anywhere in the string. Any idea why we had that in the first place?
</blockquote></details>

A regular expression can match anywhere in a string. Putting `.*` before and after doesn't do anything. It would if we had a `^` marker at the beginning and a `$` marker at the end to assert position at the beginning and end of the string respectively, but we're not doing that.

Nateowami (commented):
Oh, I see, there is a test covering this area, but the ProjectService gets mocked. We don't have any tests at all for the project service on the front end.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1330#-:-N0gqY6v7xDfAfk_-Aom:bkkb3oo)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

PR #1333, opened 2022-04-27
Update package-lock.json in ClientApp

Nateowami:
This is the result of running `npm i` in ClientApp using npm@8.8.0 (yes, the version of npm makes a difference, unfortunately).

Of course it seems odd that running `npm i` should update `package-lock.json`. According to [this issue comment](https://github.com/npm/npm/issues/17979#issuecomment-332701215) the logic is as follows:
> 1. If you have a `package.json` and you run `npm i` we generate a `package-lock.json` from it.
> 2. If you run `npm i` against that `package.json` and `package-lock.json`, the latter will never be updated, even if the `package.json` would be happy with newer versions.
> 3. If you manually edit your `package.json` to have different ranges and run `npm i` and those ranges aren't compatible with your `package-lock.json` then the latter will be updated with version that _are_ compatible with your `package.json`. Further runs of `npm i` will be as with 2 above.

So the explanation for why it would update the package lock is that there was a contradiction between `package.json` and `package-lock.json`. I don't know why that so often happens. I suspect it may be that we're all using different versions of npm, and different versions may have different ideas of what is or isn't a conflict.

This hypothesis is supported by the following two facts:
1. Running `npm ci` is supposed to fail if there's a conflict between `package.json` and `package-lock.json` (this is clearly documented).
2. Newer versions of npm fail when running `npm ci` against our current files, stating that they conflict.

My guess is that npm fixed a bug in a newer version, and that's triggering a change in what it outputs.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1333)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/package-lock.json
+++ b/src/SIL.XForge.Scripture/ClientApp/package-lock.json
@@ -97,7 +97,7 @@
         "karma-jasmine-html-reporter": "~1.7.0",
         "karma-teamcity-reporter": "^1.1.0",
         "lint-staged": "^11.2.0",
-        "npm": "^8.1.0",
+        "npm": "^8.8.0",
         "prettier": "^2.4.1",
         "ts-mockito": "^2.6.1",
         "ts-node": "^10.2.1",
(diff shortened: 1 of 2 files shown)


irahopkinson (changes requested):
Oh, I like this change that seems to effectively be doing `npm prune`. I've run that previously with little effect (although perhaps not recently). Since you are effectively saying the official npm version is now v8.8.0, update that in the ClientApp `package,json` file.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1333#-:-N0heV8N4RMEWrSWkNWN:b-h44jv4)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami (commented):
Done.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1333#-:-N0hhDYU0fMFekE6pGmP:b-896fix)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @irahopkinson)

irahopkinson (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1333#-:-N0hhz3i0W8BSaqvc0Sq:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1334, opened 2022-04-28
Update dependencies in RealtimeServer

Nateowami:
- Update dependencies in RealtimeServer
- Also run npm install in ClientApp which depends on RealtimeServer

This is accomplished by running `npm update`. What it does is look at the versions specified in `package.json` and then update them to the most recent version satisfied by the range specified in `package.json`. So for example, `@bugsnag/js` had version `^7.11.0` specified, which means the highest available version that is still in the 7.x range. npm installed 7.16.2 and then updated the file to show `^7.16.2` as the required version. The lockfile is then re-generated. Since RealtimeServer changed I also ran `npm install` in the ClientApp directory.

I intend to update dependencies in the ClientApp in a separate step. I'm trying to keep each of these changes as separate atomic actions in order to keep things simple, since we've had a lot of trouble getting consistent results from npm install. It also means if there's a regression caused by updating a dependency it will be easier to figure out what caused the regression.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1334)


irahopkinson (approved):
:lgtm: I had done similar changes on my branch so this was easy to review. Thanks for letting me know you will be doing the same to ClientApp. I normally keep the same items in sync like prettier and eslint packages between RTS and ClientApp, but this way works too.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1334#-:-N0hoq8o1406WEQRJSrC:b-wedoa5)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1341, opened 2022-05-09
SF-1553 Sort projects by shortName

Nateowami:


RaymondLuong3 (changes requested):
Just curious how this change came up? I just have the one blocking comment.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1341#-:-N1daD0U-v3honlsTjHi:bfd4ss9)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 578 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1341#-N1d_yer3Q4KDWdE6s2E:-N1d_yer3Q4KDWdE6s2F:b-9xzrek) ([raw file](https://github.com/sillsdev/web-xforge/blob/19527b0fb7a27944fd25c7b87c2d679bee97ab00/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L578)):*
> ```XML
>     this.loadingFinished();
>     return projectDocs.sort((a, b) =>
>       a.data == null || b.data == null ? 0 : SFProjectService.compareProjectsForSorting(a.data, b.data)
> ```

The zero here simply means that the order will not be affected if a.data or b.data is null. Is that correct?



_Code quote:_
```XML
a.data == null || b.data == null ? 0 
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts` line 207 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1341#-N1d_amr-HWvOXiH8BBW:-N1d_amr-HWvOXiH8BBX:brtf4sg) ([raw file](https://github.com/sillsdev/web-xforge/blob/19527b0fb7a27944fd25c7b87c2d679bee97ab00/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts#L207)):*
> ```XML
>   }
> 
>   static compareProjectsForSorting(a: { shortName: string }, b: { shortName: string }): 1 | -1 {
> ```

Would it make more sense to have this type of function in utils? I generally see services as a helper class to access information about projects, and less so for manipulating project data.



_Code quote:_
```XML
  static compareProjectsForSorting(a: { shortName: string }, b: { shortName: string }): 1 | -1 {
    return a.shortName < b.shortName ? -1 : 1;
  }
```

Nateowami (commented):
This came up because I was talking with a user that has 200 or so projects. We currently don't sort by anything, and he told me that Paratext sorts by short name and that that is the order he would expect. I have't delved into Paratext's source code to check, but as far as I can tell this is the same order as I am seeing in Paratext, which is what we would want to copy, unless there were a compelling reason to do otherwise.

Without sorting, users with lots of projects have a very bad user experience trying to connect their first project.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1341#-:-N1dd46L2P77r4ik9MvK:bxxnve2)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 578 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1341#-N1d_yer3Q4KDWdE6s2E:-N1dedCb5HKCj-Pv1Ey9:b-c6aaqh) ([raw file](https://github.com/sillsdev/web-xforge/blob/19527b0fb7a27944fd25c7b87c2d679bee97ab00/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L578)):*

The zero here simply means that the order will not be affected if a.data or b.data is null. Is that correct?
</blockquote></details>

From MDN for [`Array.prototype.sort()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort):



So yes, that's correct. It seemed like the best way to handle the unlikely but technically possible case in which data is nullish.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts` line 207 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1341#-N1d_amr-HWvOXiH8BBW:-N1dfgKH7VjKCxb9xdAs:bjhoaax) ([raw file](https://github.com/sillsdev/web-xforge/blob/19527b0fb7a27944fd25c7b87c2d679bee97ab00/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts#L207)):*

Would it make more sense to have this type of function in utils? I generally see services as a helper class to access information about projects, and less so for manipulating project data.
</blockquote></details>

I thought about it and wasn't sure. You're right this isn't where we would normally put a plain function. I'll move it.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1341)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts` line 207 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1341#-N1d_amr-HWvOXiH8BBW:-N1djNNDA_AVxteZSRGZ:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/19527b0fb7a27944fd25c7b87c2d679bee97ab00/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts#L207)):*

I thought about it and wasn't sure. You're right this isn't where we would normally put a plain function. I'll move it.
</blockquote></details>

Done.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1341)* status: 2 of 8 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts` line 127 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1341#-N1i_0Gu1O-mI_8lXc5e:-N1iaudz7vm7brNiow2k:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/8adcc70e9a82c7044b4c81c005ba13edd4d0f3cf/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts#L127)):*

Sorry to make you move it again, but it probably belongs in ClientApp/src/app/shared/utils.ts since it is specific to paratext project. What do you think?
</blockquote></details>

Done.

RaymondLuong3 (approved):
All good now.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1341#-:-N1ic1e9EeFoV5pzBiY5:b4ta47r)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1342, opened 2022-05-09
SF-1557 Fix share dialog menus appearing behind dialog

Nateowami:
Start by reading the issue and steps to reproduce: https://jira.sil.org/browse/SF-1557

This was caused by an interaction between Angular Material and Angular MDC. The mix of a MDC dialog and a Material mat-select within it caused the menu to appear behind the dialog. This problem only manifests itself if a Material dialog was opened prior to this dialog being opened.

I solved the problem by changing the dialog to use Angular Material. I'm fairly confident there are other workarounds, but this gets us closer to using only Angular Material.

Here's a before and after for the normal case, where everything is fine:
Before | After
----------|-------
 | 

As you can see, the main visual change is width. We didn't have a width set before; it was being chosen by Angular MDC as far as I can tell. So I picked a width that felt about right, and it's a little less wide than before.

And here's how it looks when you follow the steps to reproduce the bug. As a write this the screenshots look blurry, but if I click on them they clear up. That's not an actual issue with the UI; it seems to be something weird GitHub is doing.

Before | After
---------|--------
 | 

Also, very fun to fix a bug like this by adding negative 10 lines of code.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1342)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.html
@@ -1,19 +1,12 @@
 <ng-container *transloco="let t; read: 'share_dialog'">
-  <mdc-dialog>
-    <mdc-dialog-container>
-      <mdc-dialog-surface class="dialog-size">
-        <mdc-dialog-title>{{ t("share_with_others") }}</mdc-dialog-title>
-        <mdc-dialog-content>
-          <app-share-control
-            class="share-dialog"
-            [projectId]="data.projectId"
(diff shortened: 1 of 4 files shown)


irahopkinson (approved):
:lgtm: I added the other scenario in a comment on the issue. Both are now working. Thanks for this.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1342#-:-N1g8zTFCHEjk4bnYrm5:b-90yaqt)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1347, opened 2022-05-11
SF-1556b Reuse ShareDB connections on back end to improve performance

Nateowami:
Previously we were creating a new ShareDB connection every single time we check whether a user has permission to access a document (in order to determine whether a user can access e.g. a text, we have to look up the associated project and see if the user is on it). The [documentation for `Backend.prototype.connect()`](https://share.github.io/sharedb/api/backend#connect) says:

> Connects to ShareDB and returns an instance of a [`Connection`](https://share.github.io/sharedb/api/connection) client for interacting with ShareDB. This is the server-side equivalent of `new Connection(socket)` in the browser.

While not explicitly stated, this makes me believe creating thousands of connections for a single end user may cause performance issues. In my testing it appears this change reduces the amount of time the process spends to about 1/2 to 1/3 of what it was previously. That's not nearly as significant of a change as I would like, but I suspect the difference may grow as the number of connections grow, meaning the difference may be greater on live.

I don't consider this the end of the matter, but I think it will improve performance a bit for now.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1347)

### DIFF
--- a/src/RealtimeServer/common/diagnostics.ts
+++ b/src/RealtimeServer/common/diagnostics.ts
@@ -2,7 +2,7 @@ import fs from 'fs';
 import inspector from 'inspector';
 import path from 'path';
 
-const secondsToProfile = 5;
+const secondsToProfile = 30;
 
 // POSIX defines SIGUSR1 and SIGUSR2 as user-defined signals, but SIGUSR1 is reserved by Node.js to start the debugger.
 // Send the signal using e.g. kill -USR2 <node process id>
--- a/src/RealtimeServer/common/realtime-server.ts
+++ b/src/RealtimeServer/common/realtime-server.ts
@@ -110,6 +110,7 @@ export interface RealtimeServer extends ShareDB, shareDBAccess.AccessControlBack
  */
 export class RealtimeServer extends ShareDB {
   private readonly docServices = new Map<string, DocService>();
+  private defaultConnection?: Connection;
 
   constructor(
     private readonly siteId: string,
@@ -166,6 +167,8 @@ export class RealtimeServer extends ShareDB {
       delete context.op.c;
       done();
     });
+
+    this.defaultConnection = this.connect();
   }
 
   async createIndexes(db: Db): Promise<void> {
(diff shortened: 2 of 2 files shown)


RaymondLuong3 (approved):
Good job finding this improvement. It will show a lot to see how this affects QA and Live.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1347#-:-N1oLIqu9ecDwlZC2ayX:br080f9)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/RealtimeServer/common/diagnostics.ts` line 5 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1347#-N1oL0MC6ZIbpvMdRNjV:-N1oL0MC6ZIbpvMdRNjW:b-hazy4k) ([raw file](https://github.com/sillsdev/web-xforge/blob/2f24ec88ce9ad045828a3c44484bb9511573fa2a/src/RealtimeServer/common/diagnostics.ts#L5)):*
> ```XML
> import path from 'path';
> 
> const secondsToProfile = 30;
> ```

Was this on purpose?



_Code quote:_
```XML
const secondsToProfile = 30;
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1347)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/RealtimeServer/common/diagnostics.ts` line 5 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1347#-N1oL0MC6ZIbpvMdRNjV:-N1oMAQs3d3N_vgaEkRE:bn5g1ec) ([raw file](https://github.com/sillsdev/web-xforge/blob/2f24ec88ce9ad045828a3c44484bb9511573fa2a/src/RealtimeServer/common/diagnostics.ts#L5)):*

Was this on purpose?
</blockquote></details>

Yes. Sorry; should have left some sort of comment about it. I found 5 seconds was not enough (like you originally suggested!) and that 30 seconds works fine.

PR #1348, opened 2022-05-11
SF-1557b Prevent scrollbar in share dialog on mobile

Nateowami:
This was broken in #1342 "SF-1557 Fix share dialog menus appearing behind dialog" (cb494c620519e0bcc6313166558976db71b38f11)

Unfortunately I didn't test this as well as I thought I did.

Before this fix | After this fix
-------------------|-----------------
 | 

Disabling view encapsulation and using `!important` isn't the nicest solution, but the main alternative was to move the width specification to the component that opens the dialog and pass it in the config when calling `dialog.open()`. That seems like the wrong way to go, since we could easily start opening the dialog from a different place, and we don't want to have to write:
``` ts
this.dialog.open(ShareDialogComponent, {
  width: '30em',
  maxWidth: '90vw',
  ...
```
every time. In reality we would probably notice if the dialog width was completely wrong, but it still doesn't seem like something that should be duplicated. It should be possible to change it by changing it in one place.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1348)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.scss
@@ -1,3 +1,7 @@
-.mat-dialog-content {
+// Note: ViewEncapsulation is disabled for this file, so all styles can apply to any element globally
+
+.share-dialog-component-panel {
   width: 30em;
+  // Angular Material sets max-width on the element itself, so this is the only way to override in CSS
+  max-width: 90vw !important;
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.ts
@@ -1,5 +1,5 @@
-import { Component, Inject } from '@angular/core';
-import { MAT_DIALOG_DATA } from '@angular/material/dialog';
+import { Component, Inject, ViewEncapsulation } from '@angular/core';
+import { MatDialogRef, MAT_DIALOG_DATA } from '@angular/material/dialog';
 import { SFProjectRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-role';
 
 export interface ShareDialogData {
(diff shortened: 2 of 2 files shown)


irahopkinson (approved):
:lgtm: thanks for noticing and fixing. One non-blocking question.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1348#-:-N1qCdfe0_S020v8XWc0:b-n4dhx)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.ts` line 20 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1348#-N1qCPNN3C9HL80eKPvT:-N1qCPNN3C9HL80eKPvU:b-a2ji6k) ([raw file](https://github.com/sillsdev/web-xforge/blob/33ab78ee5488aa48dc88884a3109c01a6a96f11a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.ts#L20)):*
> ```XML
>     @Inject(MAT_DIALOG_DATA) public readonly data: ShareDialogData
>   ) {
>     dialogRef.addPanelClass('share-dialog-component-panel');
> ```

nit: is there not a way to add this class in HTML?



_Code quote:_
```XML
dialogRef.addPanelClass('share-dialog-component-panel');
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1348)* status: all files reviewed, 1 unresolved discussion (waiting on @irahopkinson)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.ts` line 20 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1348#-N1qCPNN3C9HL80eKPvT:-N1qJvHjAdChHnM6Rotq:b-m3r8cy) ([raw file](https://github.com/sillsdev/web-xforge/blob/33ab78ee5488aa48dc88884a3109c01a6a96f11a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.ts#L20)):*

nit: is there not a way to add this class in HTML?
</blockquote></details>

Unfortunately it needs to get added to the panel, which is what the component is inside of. The element is part of the dialog, not the component residing within it.

PR #1353, opened 2022-05-14
SF-1564a Mark texts with duplicate chapters as invalid

Nateowami:
A project on live managed to duplicate the chapter numbers in a text, which causes sync problems (it's not valid to have duplicate chapters).

This makes the text not editable, and shows a message to the user indicating that it has become corrupted.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1353)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -9,7 +9,7 @@ import tinyColor from 'tinycolor2';
 import { objectId } from 'xforge-common/utils';
 import { Delta, TextDoc } from '../../core/models/text-doc';
 import { MultiCursorViewer } from '../../translate/editor/multi-viewer/multi-viewer.component';
-import { containsInvalidOp, VERSE_FROM_SEGMENT_REF_REGEX } from '../utils';
+import { isBadDelta, VERSE_FROM_SEGMENT_REF_REGEX } from '../utils';
 import { getAttributesAtPosition } from './quill-scripture';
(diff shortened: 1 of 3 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1353)* status: 0 of 3 files reviewed, all discussions resolved

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts` line 98 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1353#-N2-Fl1W4BF3vnbGdg0q:-N2-Fl1W4BF3vnbGdg0r:b-5gdlo2) ([raw file](https://github.com/sillsdev/web-xforge/blob/b8e6b164847abda29d6893159fd213da80209964/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L98)):*
> ```XML
>  * @param ops An array of ops to check.
>  */
> export function isBadDelta(ops: DeltaOperation[]): boolean {
> ```

`isBadDelta` is not a very good name. When Mark and I chose it we were considering the following:
- We don't want it to be confused with the `isValid` attribute for each text, since since that's set at sync time by a USFM validator on the back end.
- Before it was `containsInvalidOp`, but now the logic doesn't just look at ops individually.

If you have a better name please suggest it.

RaymondLuong3 (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1353#-:-N2N2TLjCJ0icIpus4Ff:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1359, opened 2022-05-17
Add Mongo query to count users by interface language

Nateowami:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1359)

### DIFF
--- a/mongodb/Users/CountUsersByInterfaceLanguage.mongodb
+++ b/mongodb/Users/CountUsersByInterfaceLanguage.mongodb
@@ -1,0 +1,15 @@
+use('xforge')
+
+db.users.aggregate([
+  {$project: {
+    interfaceLanguage: 1,
+    _id: 0
+  }},
+  {$group: {
+    _id: '$interfaceLanguage',
+    count: { $sum: 1 }
+  }},
+  {$sort: {
+    count: -1
+  }}
+]);


RaymondLuong3 (approved):
Just one question out of curiousity.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1359#-:-N2Mv5qF6mYjRbUrg8J7:b63gsvy)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`mongodb/Users/CountUsersByInterfaceLanguage.mongodb` line 6 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1359#-N2Mum-S56jsRZdT8o0Q:-N2Mum-S56jsRZdT8o0R:bjkuue) ([raw file](https://github.com/sillsdev/web-xforge/blob/ecd3b9fe20f238417b766eceb58826d3637cecfb/mongodb/Users/CountUsersByInterfaceLanguage.mongodb#L6)):*
> ```
>   {$project: {
>     interfaceLanguage: 1,
>     _id: 0
> ```

Is this for the purpose of having an `_id` property to have a uniquely identifiable property present to be grouped?



_Code quote:_
```
   _id: 0
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1359)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`mongodb/Users/CountUsersByInterfaceLanguage.mongodb` line 6 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1359#-N2Mum-S56jsRZdT8o0Q:-N2MvzpsC6Jb3bqkHquL:b-o35uby) ([raw file](https://github.com/sillsdev/web-xforge/blob/ecd3b9fe20f238417b766eceb58826d3637cecfb/mongodb/Users/CountUsersByInterfaceLanguage.mongodb#L6)):*

Is this for the purpose of having an `_id` property to have a uniquely identifiable property present to be grouped?
</blockquote></details>

In a projection you can add or remove properties from a document. The numbers 1 and 0 are used like true and false here. Generally I think that's to be avoided, but it's the standard way of specifying a projection in a Mongo query.

For an explanation see https://www.mongodb.com/docs/v4.0/tutorial/project-fields-from-query-results/#return-the-specified-fields-and-the-id-field-only

By default the `_id` field is kept, so setting `_id: 0` here is specifying to not include it. So after this stage in the aggregation query, we just have a list of documents with a single property, `interfaceLanguage`.

PR #1368, opened 2022-05-18
SF-1566 Wrap assigned users on mobile viewports

RaymondLuong3:
The note dialog would try to condense the information about the verse and assigned user on the same line on mobile viewports. This didn't look good when names and verse references were longer. This wraps the assigned user onto the second line, and moves the selected text menu button on the line with the selected text.

Before
image

After
image

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1368)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html
@@ -4,18 +4,20 @@
     <div *ngIf="noteThreadAssignedUserRef != null" id="assignedUser" class="assigned-user">
       >{{ getAssignedUserString(noteThreadAssignedUserRef) }}
     </div>
-    <button mat-icon-button [matMenuTriggerFor]="menu"><mat-icon>more_vert</mat-icon></button>
-    <mat-menu #menu="matMenu">
-      <button mat-menu-item (click)="toggleSegmentText()">
-        <mat-icon>compare</mat-icon>
-        <span>{{ showSegmentText ? t("hide_changes") : t("show_changes") }}</span>
-      </button>
-    </mat-menu>
   </h1>
   <mat-dialog-content class="content-padding" [ngClass]="{ rtl: isRtl, ltr: !isRtl }">
-    <div class="text">
-      <div class="note-text" [innerHTML]="noteContextText"></div>
-      <div *ngIf="showSegmentText" class="segment-text" [innerHTML]="segmentText"></div>
+    <div class="text-row">
+      <div class="text">
+        <div class="note-text" [innerHTML]="noteContextText"></div>
+        <div *ngIf="showSegmentText" class="segment-text" [innerHTML]="segmentText"></div>
+      </div>
+      <button mat-icon-button [matMenuTriggerFor]="menu"><mat-icon>more_vert</mat-icon></button>
+      <mat-menu #menu="matMenu">
+        <button mat-menu-item (click)="toggleSegmentText()">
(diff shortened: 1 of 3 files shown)


Nateowami (commented):
This is a good improvement, but it still has wraps in some basic cases. With a 492px wide screen, a reference of 1 Thessalonians 2:20 and assignment text of ">(Unassigned)", it still wraps:


It looks really bad at 437px:


There are two potential simple fixes I see:
1. Set `white-space: nowrap;` on the reference. This works great down to very narrow screens, but could be an issue with even longer references, where maybe we do want them to wrap.
2. Use a media query to set `flex-direction: column;` on smaller screens. That ends up looking something like this (dashed line is from the dev tools visualizing the flexbox layout):

The problem with this solution is that it relies only on the width of the screen, and if a user has a very long name, or the reference is very long, it won't wrap. That makes me think solution 1 is probably better.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1368#-:-N2ND4p54PzRK8Nl6HCk:b34ylns)* status: 0 of 2 files reviewed, all discussions resolved

Nateowami (approved):
Nice work. :lgtm: 

> Reviewed 1 of 2 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1368#-:-N2NVvcI5oO7TIO5zdit:b-a5e3ql)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html` line 11 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1368#-N2NVjN01KHDw1-wSWew:-N2NVjN01KHDw1-wSWex:bey6c98) ([raw file](https://github.com/sillsdev/web-xforge/blob/ed9ba33e318dc0d17addc606d0fa304dbce865b9/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html#L11)):*
> ```HTML
>     <div class="text-row">
>       <div class="text">
>         <div class="note-text" [innerHTML]="noteContextText"></div>
> ```

Is it normal that I have a note where this is basically blank (`<b></b>`, which isn't visible)?

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1368)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html` line 11 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1368#-N2NVjN01KHDw1-wSWew:-N2NxHvu1QYKmYpT_N9c:bj3eotl) ([raw file](https://github.com/sillsdev/web-xforge/blob/7442596978d5e0e3355c51298429649402f14079/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html#L11)):*

Is it normal that I have a note where this is basically blank (`<b></b>`, which isn't visible)?
</blockquote></details>

I've seen a few strange cases where notes are essentially blank. The `content` property is null. This happens for reattached notes. In your case, I'm not sure whether is it expected or not.

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1368)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1372, opened 2022-05-19
SF-1492 Use default font size for translate tool specified in Paratext

RaymondLuong3:
* add default font size and font properties to model
* add migration for model in realtime server SF project service
* set font size on translation editor init
* convert text component to accept a font size string as an input
* clean up test SFProject instances to include new properties

Note that the Arial font is the default font in Paratext. We discussed using Charis SIL because it supports most scripts, but I decided to go with Arial so that we are not overriding the Paratext default.
The default font is not currently being used, but is now available when it is needed.
I did a quick search to learn how a font size is determined. I didn't find an exact conversion from font size in points to em, although some things I read suggested that 12 point font converts to 1 em. I roughly implemented it that way.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1372)


Nateowami (changes requested):
> Note that the Arial font is the default font in Paratext.

I created a new project a couple days ago and didn't do anything special with the font settings that I can recall. I just checked and the font is set to Charis SIL. Where are you seeing that the default is Arial? One of the issues with Arial is that based on my limited research I don't think we can distribute it.

I don't think a migration is necessary, or the right way to implement a default. Not all Paratext projects have a font set, and if they don't have it set, we should not have it set in the Scripture Forge database either. The property should be optional, since it only exists on some projects. The default will then be whatever we choose to use when the the project doesn't have a font specified. This also means a lot of the TypeScript test files don't need the properties added. I added a note about not needing a migration in the issue; I wonder if perhaps you overlooked it? I think I've sometimes forgotten to read through issues before working on them.

Thanks for your work on this; I think some upcoming projects are going to really need the font size set since some script need a much larger size in order to be legible.

> Reviewed 6 of 36 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1372#-:-N2PDHXJE-nIAyycfdJh:b-79xom)* status: 6 of 36 files reviewed, 3 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.ts` line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1372#-N2PFdCR0MyEV2tsiFzr:-N2PFdCR0MyEV2tsiFzs:b7bktz2) ([raw file](https://github.com/sillsdev/web-xforge/blob/cb8dc5a406370f827168070c5072f2c2bc0940ab/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.ts#L78)):*
> ```XML
> 
>   @Input() set fontSize(fontSize: string) {
>     this._fontSize = fontSize;
> ```

Why do we need a getter and setter here? Couldn't we just use a public property `fontSize`?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 319 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1372#-N2PG2tWDVW2xJHBBZU5:-N2PG2tWDVW2xJHBBZU6:b-pjux5x) ([raw file](https://github.com/sillsdev/web-xforge/blob/cb8dc5a406370f827168070c5072f2c2bc0940ab/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L319)):*
> ```XML
> 
>   @Input() set fontSize(size: string) {
>     this.editorStyles = { fontSize: size };
> ```

Isn't this going to overwrite any other styles that are set? I would think this would be something like:
``` ts
this.editorStyles.fontSize = size;
this.applyEditorStyles();
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 312 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1372#-N2PGeDeE4x481quSO4D:-N2PGeDeE4x481quSO4E:b-lakt9f) ([raw file](https://github.com/sillsdev/web-xforge/blob/cb8dc5a406370f827168070c5072f2c2bc0940ab/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L312)):*
> ```XML
> 
>   get fontSize(): string {
>     // Paratext allows a font size between 8 and 32. We roughly convert that to a scale between 1-3rem
> ```

I know using pixels as a unit is generally discouraged in CSS, but I don't understand the issue well enough to know whether this is an improvement or has all the same drawbacks. Is there a compelling reason not to just use the font size as pixels (which I assume is the unit Paratext is using)?

RaymondLuong3 (commented):
It was fun to work on this one. I've made some changes as requested. Making the default font size and default font optional made this PR a lot leaner.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1372#-:-N2rNMiz0AJ390qpoSku:b-wu00ds)* status: 6 of 36 files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.ts` line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1372#-N2PFdCR0MyEV2tsiFzr:-N2rIni72PYM0kQlksO7:bkyz9lm) ([raw file](https://github.com/sillsdev/web-xforge/blob/cb8dc5a406370f827168070c5072f2c2bc0940ab/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.ts#L78)):*

Why do we need a getter and setter here? Couldn't we just use a public property `fontSize`?
</blockquote></details>

I'll clean that up. Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 319 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1372#-N2PG2tWDVW2xJHBBZU5:-N2rJCQ6--OMGtoHRd89:btaihgv) ([raw file](https://github.com/sillsdev/web-xforge/blob/cb8dc5a406370f827168070c5072f2c2bc0940ab/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L319)):*

Isn't this going to overwrite any other styles that are set? I would think this would be something like:
``` ts
this.editorStyles.fontSize = size;
this.applyEditorStyles();
```
</blockquote></details>

That is a good question. Right now we only use the font size editor style, but I'll update this so that we can handle when we choose to add in other styles.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 312 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1372#-N2PGeDeE4x481quSO4D:-N2rKYU31KWZ42lXaKIa:b-fbecf7) ([raw file](https://github.com/sillsdev/web-xforge/blob/cb8dc5a406370f827168070c5072f2c2bc0940ab/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L312)):*

I know using pixels as a unit is generally discouraged in CSS, but I don't understand the issue well enough to know whether this is an improvement or has all the same drawbacks. Is there a compelling reason not to just use the font size as pixels (which I assume is the unit Paratext is using)?
</blockquote></details>

I chatted with Tim S, and he mentioned that Paratext uses the standard pt units for their font size. We could copy that, but after reading https://kyleschaeffer.com/css-font-size-em-vs-px-vs-pt-vs-percent, I propose we stick to using em as our font size value.

Nateowami (approved):
> Reviewed 7 of 36 files at r1, 27 of 27 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1372)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 312 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1372#-N2PGeDeE4x481quSO4D:-N2sG-HOBliFDdMk1rTV:b-vs1pr4) ([raw file](https://github.com/sillsdev/web-xforge/blob/cb8dc5a406370f827168070c5072f2c2bc0940ab/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L312)):*

I chatted with Tim S, and he mentioned that Paratext uses the standard pt units for their font size. We could copy that, but after reading https://kyleschaeffer.com/css-font-size-em-vs-px-vs-pt-vs-percent, I propose we stick to using em as our font size value.
</blockquote></details>

Looks good.

PR #1375, opened 2022-05-23
Convert auth service to use the new Auth0 SPA

nigel-wells:
- Converted `AuthService` to use the new Auth0 SPA package
- Enable use of refresh tokens which then automatically uses the offline scope
- Fixed tests setup to better replicate different logging in flows
- Improved handling of return URL to avoid redirect loops
- Used a new dedicated callback URL to know when to check with auth0 on handling the callback code

Updates made to auth0 configuration
- Added `http://localhost:5000/callback` as a new callback URL
- Removed old beta callback URLs
- Refresh token rotation enabled
- API access settings | enabled Allow Offline Access

There appears to be a few different circumstances to trigger an update to tokens:
- API token settings expiry is hit - this appear to be our main use of the `expires_at` setting
- Application ID Token expiration

Both of those are stored and handled but the auth0 package and it will automatically renew them as required. We currently have a renewal timer set on the API expiration so that we can force a refresh early enough before we send the access token to the backend. The default for the ID Token expiration is 36000 or 10 hours. I currently have both of these set to 300 or 5 minutes. 30 seconds prior to this a silent token exchange will take place and the expiration will be reset to another 5 minutes. All going well there will be no redirect take place as it was at this point browsers that blocked 3rd party cookies would return a `login_required` response.

The expected behavior after switching to the new auth flow is:
- If the browser is already running the app then nothing will change. On an expired token if the browser supports 3rd party cookies then it will continue to authenticate fine. If the browser does not support 3rd party cookies it will get the usual `login_required` response sending you back to auth0. Auth0 will likely assume you're already logged in and send you back to SF. SF will not have the new token data required from the new auth0 package and will attempt to log the user in again but this time with the correct callback and scope. Auth0 will now prompt for the `offline_access` scope and then redirect back to SF where you left off.
- If the browser is reloaded, and you're already logged in, then it will be redirected to auth0 for the new `offline_access` scope

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/package-lock.json
+++ b/src/SIL.XForge.Scripture/ClientApp/package-lock.json
@@ -22,6 +22,7 @@
         "@angular/pwa": "~12.2.17",
         "@angular/router": "~12.2.16",
         "@angular/service-worker": "~12.2.16",
+        "@auth0/auth0-spa-js": "^1.21.0",
         "@bugsnag/js": "^7.16.5",
         "@bugsnag/plugin-angular": "^7.16.5",
         "@ngneat/transloco": "^3.2.0",
@@ -816,6 +817,30 @@
       "resolved": "https://registry.npmjs.org/@assemblyscript/loader/-/loader-0.10.1.tgz",
       "integrity": "sha512-H71nDOOL8Y7kWRLqf6Sums+01Q5msqBW2KhDUTemh1tvY04eSkSXrK0uj/4mmY0Xr16/3zyZmsrxN7CKuRbNRg==",
       "dev": true
+    },
+    "node_modules/@auth0/auth0-spa-js": {
+      "version": "1.21.0",
+      "resolved": "https://registry.npmjs.org/@auth0/auth0-spa-js/-/auth0-spa-js-1.21.0.tgz",
+      "integrity": "sha512-a+8+onJdOIUSEEjdIzP/BGd731DZjBn2Q0tH+v7H+bN9ldeuBOifnjPbFwQAriL+94SwkaqhoFx3L6uGDk3+vg==",
+      "dependencies": {
+        "abortcontroller-polyfill": "^1.7.3",
+        "browser-tabs-lock": "^1.2.15",
+        "core-js": "^3.20.3",
+        "es-cookie": "^1.3.2",
+        "fast-text-encoding": "^1.0.3",
+        "promise-polyfill": "^8.2.1",
+        "unfetch": "^4.2.0"
+      }
+    },
+    "node_modules/@auth0/auth0-spa-js/node_modules/core-js": {
+      "version": "3.22.2",
+      "resolved": "https://registry.npmjs.org/core-js/-/core-js-3.22.2.tgz",
+      "integrity": "sha512-Z5I2vzDnEIqO2YhELVMFcL1An2CIsFe9Q7byZhs8c/QxummxZlAHw33TUHbIte987LkisOgL0LwQ1P9D6VISnA==",
+      "hasInstallScript": true,
+      "funding": {
+        "type": "opencollective",
+        "url": "https://opencollective.com/core-js"
+      }
     },
     "node_modules/@babel/code-frame": {
       "version": "7.16.7",
@@ -4853,6 +4878,11 @@
       "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
       "dev": true
     },
+    "node_modules/abortcontroller-polyfill": {
+      "version": "1.7.3",
+      "resolved": "https://registry.npmjs.org/abortcontroller-polyfill/-/abortcontroller-polyfill-1.7.3.tgz",
+      "integrity": "sha512-zetDJxd89y3X99Kvo4qFx8GKlt6GsvN3UcRZHwU6iFA/0KiOmhkTVhe8oRoTBiTVPZu09x3vCra47+w8Yz1+2Q=="
+    },
     "node_modules/accepts": {
       "version": "1.3.8",
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.8.tgz",
@@ -5703,6 +5733,15 @@
       },
       "engines": {
         "node": ">=8"
+      }
+    },
+    "node_modules/browser-tabs-lock": {
+      "version": "1.2.15",
+      "resolved": "https://registry.npmjs.org/browser-tabs-lock/-/browser-tabs-lock-1.2.15.tgz",
+      "integrity": "sha512-J8K9vdivK0Di+b8SBdE7EZxDr88TnATing7XoLw6+nFkXMQ6sVBh92K3NQvZlZU91AIkFRi0w3sztk5Z+vsswA==",
+      "hasInstallScript": true,
+      "dependencies": {
+        "lodash": ">=4.17.21"
       }
     },
     "node_modules/browserslist": {
@@ -8016,6 +8055,11 @@
         "url": "https://github.com/sponsors/ljharb"
       }
     },
+    "node_modules/es-cookie": {
+      "version": "1.3.2",
+      "resolved": "https://registry.npmjs.org/es-cookie/-/es-cookie-1.3.2.tgz",
+      "integrity": "sha512-UTlYYhXGLOy05P/vKVT2Ui7WtC7NiRzGtJyAKKn32g5Gvcjn7KAClLPWlipCtxIus934dFg9o9jXiBL0nP+t9Q=="
+    },
     "node_modules/es-module-lexer": {
       "version": "0.7.1",
       "resolved": "https://registry.npmjs.org/es-module-lexer/-/es-module-lexer-0.7.1.tgz",
@@ -9376,6 +9420,11 @@
       "version": "2.1.1",
       "resolved": "https://registry.npmjs.org/fast-safe-stringify/-/fast-safe-stringify-2.1.1.tgz",
       "integrity": "sha512-W+KJc2dmILlPplD/H4K9l9LcAHAfPtP6BY84uVLXQ6Evcz9Lcg33Y2z1IVblT6xdY54PXYVHEv+0Wpq8Io6zkA=="
+    },
+    "node_modules/fast-text-encoding": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/fast-text-encoding/-/fast-text-encoding-1.0.3.tgz",
+      "integrity": "sha512-dtm4QZH9nZtcDt8qJiOH9fcQd1NAgi+K1O2DbE6GG1PPCK/BWfOH3idCTRQ4ImXRUOyopDEgDEnVEE7Y/2Wrig=="
     },
     "node_modules/fast-url-parser": {
       "version": "1.1.3",
@@ -18911,6 +18960,11 @@
       "integrity": "sha1-mEcocL8igTL8vdhoEputEsPAKeM=",
       "dev": true
     },
+    "node_modules/promise-polyfill": {
+      "version": "8.2.3",
+      "resolved": "https://registry.npmjs.org/promise-polyfill/-/promise-polyfill-8.2.3.tgz",
+      "integrity": "sha512-Og0+jCRQetV84U8wVjMNccfGCnMQ9mGs9Hv78QFe+pSDD3gWTpz0y+1QCuxy5d/vBFuZ3iwP2eycAkvqIMPmWg=="
+    },
     "node_modules/promise-retry": {
       "version": "2.0.1",
       "resolved": "https://registry.npmjs.org/promise-retry/-/promise-retry-2.0.1.tgz",
@@ -23854,6 +23908,27 @@
       "integrity": "sha512-H71nDOOL8Y7kWRLqf6Sums+01Q5msqBW2KhDUTemh1tvY04eSkSXrK0uj/4mmY0Xr16/3zyZmsrxN7CKuRbNRg==",
       "dev": true
     },
+    "@auth0/auth0-spa-js": {
+      "version": "1.21.0",
+      "resolved": "https://registry.npmjs.org/@auth0/auth0-spa-js/-/auth0-spa-js-1.21.0.tgz",
+      "integrity": "sha512-a+8+onJdOIUSEEjdIzP/BGd731DZjBn2Q0tH+v7H+bN9ldeuBOifnjPbFwQAriL+94SwkaqhoFx3L6uGDk3+vg==",
+      "requires": {
+        "abortcontroller-polyfill": "^1.7.3",
+        "browser-tabs-lock": "^1.2.15",
+        "core-js": "^3.20.3",
+        "es-cookie": "^1.3.2",
+        "fast-text-encoding": "^1.0.3",
+        "promise-polyfill": "^8.2.1",
+        "unfetch": "^4.2.0"
+      },
+      "dependencies": {
+        "core-js": {
+          "version": "3.22.2",
+          "resolved": "https://registry.npmjs.org/core-js/-/core-js-3.22.2.tgz",
+          "integrity": "sha512-Z5I2vzDnEIqO2YhELVMFcL1An2CIsFe9Q7byZhs8c/QxummxZlAHw33TUHbIte987LkisOgL0LwQ1P9D6VISnA=="
+        }
+      }
+    },
     "@babel/code-frame": {
       "version": "7.16.7",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.16.7.tgz",
@@ -27116,6 +27191,11 @@
       "integrity": "sha512-nne9/IiQ/hzIhY6pdDnbBtz7DjPTKrY00P/zvPSm5pOFkl6xuGrGnXn/VtTNNfNtAfZ9/1RtehkszU9qcTii0Q==",
       "dev": true
     },
+    "abortcontroller-polyfill": {
+      "version": "1.7.3",
+      "resolved": "https://registry.npmjs.org/abortcontroller-polyfill/-/abortcontroller-polyfill-1.7.3.tgz",
+      "integrity": "sha512-zetDJxd89y3X99Kvo4qFx8GKlt6GsvN3UcRZHwU6iFA/0KiOmhkTVhe8oRoTBiTVPZu09x3vCra47+w8Yz1+2Q=="
+    },
     "accepts": {
       "version": "1.3.8",
       "resolved": "https://registry.npmjs.org/accepts/-/accepts-1.3.8.tgz",
@@ -27758,6 +27838,14 @@
       "dev": true,
       "requires": {
         "fill-range": "^7.0.1"
+      }
+    },
+    "browser-tabs-lock": {
+      "version": "1.2.15",
+      "resolved": "https://registry.npmjs.org/browser-tabs-lock/-/browser-tabs-lock-1.2.15.tgz",
+      "integrity": "sha512-J8K9vdivK0Di+b8SBdE7EZxDr88TnATing7XoLw6+nFkXMQ6sVBh92K3NQvZlZU91AIkFRi0w3sztk5Z+vsswA==",
+      "requires": {
+        "lodash": ">=4.17.21"
       }
     },
     "browserslist": {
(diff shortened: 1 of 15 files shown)


RaymondLuong3 (changes requested):
Good work! And the test coverage is good too. I have some suggested improvements, but the critical parts appear to be there.

> Reviewed 4 of 10 files at r1, 7 of 7 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-:-N2xB90F1KUDMBFR3mzq:b-9d5u3o)* status: 10 of 11 files reviewed, 5 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 64 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N2wIggoE6sddpfAGgaI:-N2wIggp8BKKf1atcSzZ:b55msu0) ([raw file](https://github.com/sillsdev/web-xforge/blob/0625f2654f8542793a4ec3e9e3bb2301c1a5df63/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L64)):*
> ```XML
>   private renewTokenPromise?: Promise<void>;
>   private checkSessionPromise?: Promise<GetTokenSilentlyVerboseResponse | null>;
>   private readonly auth0 = this.auth0Service.init({
> ```

Can you add a type for this?



_Code quote:_
```XML
 private readonly auth0 = this.auth0Service.init({
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 160 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N2wET9QErqTDu6dds4O:-N2wET9QErqTDu6dds4P:bc8kats) ([raw file](https://github.com/sillsdev/web-xforge/blob/0625f2654f8542793a4ec3e9e3bb2301c1a5df63/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L160)):*
> ```XML
>     try {
>       return await this.auth0!.getTokenSilently();
>     } catch {
> ```

We typically want to avoid using the bang to assert not null. We could return undefined if `this.auth0` is undefined. Something like `return await this.auth0?.getTokenSilently()` 
This and below.



_Code quote:_
```XML
      return await this.auth0!.getTokenSilently();
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 329 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N2wLq6L-ERpXiKymQgw:-N2wLq6L-ERpXiKymQgx:b-2lzl1i) ([raw file](https://github.com/sillsdev/web-xforge/blob/0625f2654f8542793a4ec3e9e3bb2301c1a5df63/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L329)):*
> ```XML
>           )
>           .then(async () => {
>             const withoutState = clone(authDetails);
> ```

It would be helpful if you leave a comment here explaining why you are removing the state.



_Code quote:_
```XML
       const withoutState = clone(authDetails);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 348 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N2wM59rD8ADYDTyT3QD:-N2wM59rD8ADYDTyT3QE:b-8s707x) ([raw file](https://github.com/sillsdev/web-xforge/blob/0625f2654f8542793a4ec3e9e3bb2301c1a5df63/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L348)):*
> ```XML
>     // Ensure the return URL is one we want to return the user to
>     if (this.isValidReturnUrl(state.returnUrl)) {
>       this.router.navigateByUrl(state.returnUrl!, { replaceUrl: true });
> ```

Instead of asserting not null with the bang, you can add a null check in the iff above.



_Code quote:_
```XML
      this.router.navigateByUrl(state.returnUrl!, { replaceUrl: true });
    } else {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 356 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N2wN-mEClmPF66bdoXw:-N2wN-mFBuBBhSPPT4zc:b98zots) ([raw file](https://github.com/sillsdev/web-xforge/blob/0625f2654f8542793a4ec3e9e3bb2301c1a5df63/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L356)):*
> ```XML
> 
>   private isValidReturnUrl(returnUrl: string | undefined): boolean {
>     return !(returnUrl == null || this.isCallbackUrl(returnUrl) || returnUrl === '/login');
> ```

Why is it necessary to check that the the return url is not the callback url? Why is that invalid?



_Code quote:_
```XML
|| this.isCallbackUrl(returnUrl) || 
```

nigel-wells (commented):
The coverage report above is interesting - when I view it locally `auth.service.ts` is at 90%

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-:-N2y1jb248a4HImm2RLt:b-5l4dxc)* status: 10 of 11 files reviewed, 5 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 64 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N2wIggoE6sddpfAGgaI:-N2y0AF33OeWqRZbq1nw:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/0625f2654f8542793a4ec3e9e3bb2301c1a5df63/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L64)):*

Can you add a type for this?
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 160 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N2wET9QErqTDu6dds4O:-N2y0R1_7soZHL7O_7oW:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/0625f2654f8542793a4ec3e9e3bb2301c1a5df63/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L160)):*

We typically want to avoid using the bang to assert not null. We could return undefined if `this.auth0` is undefined. Something like `return await this.auth0?.getTokenSilently()` 
This and below.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 329 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N2wLq6L-ERpXiKymQgw:-N2y1Ta98HEy_oicO-r8:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/0625f2654f8542793a4ec3e9e3bb2301c1a5df63/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L329)):*

It would be helpful if you leave a comment here explaining why you are removing the state.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 348 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N2wM59rD8ADYDTyT3QD:-N2y0bDT-ZX5h_84yizN:be8i0ml) ([raw file](https://github.com/sillsdev/web-xforge/blob/0625f2654f8542793a4ec3e9e3bb2301c1a5df63/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L348)):*

Instead of asserting not null with the bang, you can add a null check in the iff above.
</blockquote></details>

This one is probably ok as it has already been verified that it isn't null by `isValidReturnUrl`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 356 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N2wN-mEClmPF66bdoXw:-N2y0lMnAFo6OR7wi9pJ:b5jlbdq) ([raw file](https://github.com/sillsdev/web-xforge/blob/0625f2654f8542793a4ec3e9e3bb2301c1a5df63/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L356)):*

Why is it necessary to check that the the return url is not the callback url? Why is that invalid?
</blockquote></details>

I believe there was a bug in the current implementation where the return URL sometimes ends up as the same URL returned from auth0 which then gets the user stuck in a loop. This makes sure that doesn't happen and we can also expand upon it if we find other situations that might cause it. It may be that we need to reverse the logic and only have a set of valid paths we allow to be return urls.

nigel-wells (commented):
Perhaps the one noted above is from the original commit though before I added some extra tests in though?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-:-N2y2kpM6ahVxqV5cx9n:bizcuzm)* status: 10 of 11 files reviewed, 5 unresolved discussions (waiting on @RaymondLuong3)

Nateowami:
@nigel-wells While working on other things with my work based off this branch, I've run into the following after coming back from Auth0 (GitHub shows it blurry; click it to see it clearly):



This happened after I used the "log in with Paratext" button in SF that is shown when SF thinks the user does not have access to Paratext. After I come back from Auth0 I get this screen. The tokens apparently don't update, because I'm still greeted with the "log in with Paratext" buttons. Those buttons trigger `linkParatext(returnUrl: string)` in `AuthService`.

RaymondLuong3 (changes requested):
I tried a few more scenarios in chrome incognito mode. As for what Nathaniel pointed out, when I could proceed to my Paratext-linked account when I used paratext credentials linked to an existing SF account. The page reloaded after the pop-up, but was smooth otherwise.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-:-N3617nt4Va6lvCfreTo:b-9zp699)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 348 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N2wM59rD8ADYDTyT3QD:-N35y8jODXzlA4pHiThU:b8sec31) ([raw file](https://github.com/sillsdev/web-xforge/blob/0625f2654f8542793a4ec3e9e3bb2301c1a5df63/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L348)):*

This one is probably ok as it has already been verified that it isn't null by `isValidReturnUrl`
</blockquote></details>

That is fine. There are some places we use the not null assertion in production code, but I want to keep that to a minimal.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 356 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N2wN-mEClmPF66bdoXw:-N35yRgvEJMLfKYF8wxE:b-igjdvb) ([raw file](https://github.com/sillsdev/web-xforge/blob/0625f2654f8542793a4ec3e9e3bb2301c1a5df63/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L356)):*

I believe there was a bug in the current implementation where the return URL sometimes ends up as the same URL returned from auth0 which then gets the user stuck in a loop. This makes sure that doesn't happen and we can also expand upon it if we find other situations that might cause it. It may be that we need to reverse the logic and only have a set of valid paths we allow to be return urls.
</blockquote></details>

Ok, that is god to keep in mind. I would add a comment so that it is easy to see that.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 195 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N35zVWu763sdaCTCxfH:-N35zVWu763sdaCTCxfI:b4sofsq) ([raw file](https://github.com/sillsdev/web-xforge/blob/f35050698819d3b6abbd9d4b20af5ac56e3fa1b4/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L195)):*
> ```XML
>     }
>     this.unscheduleRenewal();
>     await this.auth0!.loginWithRedirect(authOptions);
> ```

I'm wondering if we even need the bang at all for `this.auth0`. Since we declare it to be a defined type, it shouldn't be necessary. Here and below.



_Code quote:_
```XML
    await this.auth0!.loginWithRedirect(authOptions);
```

Nateowami (commented):
> As for what Nathaniel pointed out, when I could proceed to my Paratext-linked account when I used paratext credentials linked to an existing SF account. The page reloaded after the pop-up, but was smooth otherwise.

Nigel fixed the issue I ran into. We talked about it in a meeting but didn't comment on the PR itself.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-:-N36K4Kl9huuNeNzQ1LT:b-wneujx)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

nigel-wells (commented):
@raymondluong3 the page reload is expected behavior - at least that is what it has always done anyway if a user ID changes which is what can happen when linking PT accounts. We could probably make that smoother if we felt it occurred enough but I'd do that as a separate task.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-:-N3Hg3321BFpDj75n30k:blk2hqj)* status: 10 of 11 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 348 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N2wM59rD8ADYDTyT3QD:-N3HdGGv17vnJXzesx6f:b-q693a5) ([raw file](https://github.com/sillsdev/web-xforge/blob/0625f2654f8542793a4ec3e9e3bb2301c1a5df63/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L348)):*

That is fine. There are some places we use the not null assertion in production code, but I want to keep that to a minimal.
</blockquote></details>

As above, I've now removed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 356 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N2wN-mEClmPF66bdoXw:-N3Hdnky8ukZ0BoTPhEg:bh25w3m) ([raw file](https://github.com/sillsdev/web-xforge/blob/0625f2654f8542793a4ec3e9e3bb2301c1a5df63/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L356)):*

Ok, that is god to keep in mind. I would add a comment so that it is easy to see that.
</blockquote></details>

Have added a comment to the method
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 195 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N35zVWu763sdaCTCxfH:-N3Hd-7qDDBg-qwZpv8o:b8yb9q) ([raw file](https://github.com/sillsdev/web-xforge/blob/f35050698819d3b6abbd9d4b20af5ac56e3fa1b4/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L195)):*

I'm wondering if we even need the bang at all for `this.auth0`. Since we declare it to be a defined type, it shouldn't be necessary. Here and below.
</blockquote></details>

You're absolutely right. When I first started refactoring in the new auth0 package I had it setup as possibly undefined. Towards the end I was able to get it setup defined in the constructor. I've gone ahead and removed all the bangs.

RaymondLuong3 (approved):
Thanks for your work. From my testing this is working as expected. I approve.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-:-N3Kj3Y1EeIvdGCKKmKL:b9ihhw8)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

Nateowami (changes requested):
Let's make the needed changes to the Auth0 tenants before merging this. @nigel-wells Can you list the changes that need to be made to the qa and live tenants?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-:-N3PzxdI97PTH_aX_Bqb:bxjwrvt)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

nigel-wells (commented):
At a guess, as I don't have access to the repo, would be:
## Resource server
```
{
   ...
   "allow_offline_access": true,
   ...
}
```

## Client
 ```
{
     "callbacks": [
         "http://localhost/callback/auth0" // Just needs the path updated from "projects" to "callback/auth0"
     ],
     "refresh_token": {
      ...
        "rotation_type": "rotating",
      ...
}
```

Happy to do a screenshare on this if that is easier for us both to look at how it is setup for QA/Prod.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-:-N3QqTIw4YLaeaTo6yyo:b-pvdyjb)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

**Review by nigel-wells — COMMENTED** [review]
I think there is an issue with the linking to PT which I've discovered when testing the passwordless login - hold off merging until I investigate more

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-:-N3S3_J71YcvcZBXtObI:b-g6jzp7)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

**Review by nigel-wells — COMMENTED** [review]
I've just pushed through the latest updates for the PT linking fix. I'll start looking at the auth0 CLI config stuff now

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-:-N4-9_2J6bqFMT3sUC86:b-mkfeow)* status: 9 of 15 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

**Review by RaymondLuong3 — DISMISSED** [review]
> Reviewed 6 of 6 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 333 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N43AKQbE1NUWnOxJDXK:-N43AKQcBu2TQB4JA4rV:bm99gb6) ([raw file](https://github.com/sillsdev/web-xforge/blob/9783587ae6f0fe787a28f28f08d14cb078f4acca/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L333)):*
> ```XML
>       try {
>         await this.commandService.onlineInvoke(USERS_URL, 'linkParatextAccount', { primaryId, secondaryId });
>         // Trigger the login phase again so that local auth0 switches back its understanding to the primary account
> ```

What is going on here? Can this be removed? since we are not triggering the login phase and simply checking the session?



_Code quote:_
```XML
switches back its understanding to the primary 
```

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375)* status: 14 of 15 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 333 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-N43AKQbE1NUWnOxJDXK:-N441OG2-dwz2-k40lfk:b-dm6cy) ([raw file](https://github.com/sillsdev/web-xforge/blob/9783587ae6f0fe787a28f28f08d14cb078f4acca/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L333)):*

What is going on here? Can this be removed? since we are not triggering the login phase and simply checking the session?
</blockquote></details>

Whoops - and that is why we have code review :)

I've updated the comment as well to help it better reflect what is happening as well.

nigel-wells (commented):
Just noting here that I have a feature branch ready to go for the auth0 config when it is time to start testing on QA

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-:-N44lrh98lFKKuxcjth4:bq01wrf)* status: 14 of 15 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

Nateowami (approved):
Thanks @nigel-wells. Let's :shipit: to qa.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1375#-:-N45HHex59j2tMLjDEYD:byoh1h4)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

PR #1377, opened 2022-05-25
SF-1579 Implement rtl ui

Nateowami:
This is initial work to get a right-to-left UI working. It is not intended to fully implement everything, just to have enough working so that when we start to pull in Arabic translations in the very near future, we'll have the interface mostly working correctly and we can meaningfully test the translations.

A few things to note:
- Angular Material seems to be able to do direction fine via the `BidiModule`. I couldn't find anything similar for Angular MDC. The dialogs that use MDC have to have the direction set on them in order to work properly. I only did this to the most important dialogs; I didn't actually go through every single dialog we have. For the simple dialogs it doesn't make much difference anyway.
- I had some challenges getting the app menu to work properly and ended up converting it to Angular Material. I figure we need to do that at some point anyway.
- `::ng-deep` is deprecated and I used it anyway. There are ways around it, but it seemed like the simplest option for now.
- If you don't know what `margin-inline-start` means (I didn't) you'll want to read about CSS logical properties: https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-end. There are similar properties for padding and positioning. When the text is LTR, margin-inline-start is the same as margin-left. When the text is RTL margin-inline-start is the same as margin-right. Browser support isn't the greatest for these properties, so we will need to consider the trade-offs. See https://caniuse.com/mdn-css_properties_scroll-margin-inline-start
- I have yet to fix the tests. Right now there are 28 failures.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html
@@ -116,14 +116,8 @@
                 </div>
                 <mdc-list-divider></mdc-list-divider>
                 <div class="online-status">
-                  <div *ngIf="isAppOnline" fxLayout="row" fxLayoutAlign="center center">
-                    <mdc-icon>cloud</mdc-icon>
-                    {{ t("online") }}
-                  </div>
-                  <div *ngIf="!isAppOnline" fxLayout="row" fxLayoutAlign="center center">
-                    <mdc-icon mdcListItemGraphic>cloud_off</mdc-icon>
-                    {{ t("offline") }}
-                  </div>
+                  <ng-container *ngIf="isAppOnline"><mdc-icon>cloud</mdc-icon> {{ t("online") }}</ng-container>
+                  <ng-container *ngIf="!isAppOnline"><mdc-icon>cloud_off</mdc-icon> {{ t("offline") }}</ng-container>
                 </div>
               </mdc-list-group>
             </mdc-menu>
@@ -165,100 +159,108 @@
       </mdc-select>
     </mdc-drawer-header>
     <mdc-drawer-content>
-      <mdc-list id="menu-list">
-        <mdc-list-item
+      <mat-nav-list id="menu-list">
+        <a
+          mat-list-item
           class="translate-nav-item"
           *ngIf="isTranslateEnabled"
-          (selectionChange)="translateVisible = !translateVisible"
+          (click)="translateVisible = !translateVisible"
         >
-          <mdc-icon svgIcon="translate"></mdc-icon>
+          <mdc-icon svgIcon="translate" class="translate-icon"></mdc-icon>
           {{ t("translate") }}
-          <mdc-icon *ngIf="!hasSingleAppEnabled" mdcListItemMeta>
-            keyboard_arrow_{{ translateVisible ? "down" : "right" }}
-          </mdc-icon>
-        </mdc-list-item>
+          <mat-icon *ngIf="!hasSingleAppEnabled">
+            keyboard_arrow_{{ translateVisible ? "down" : i18n.forwardDirectionWord }}
+          </mat-icon>
+        </a>
         <div *ngIf="isTranslateEnabled && (translateVisible || hasSingleAppEnabled)">
-          <mdc-list-item [appRouterLink]="getRouterLink('translate')" (selectionChange)="itemSelected()">
-            <mdc-icon mdcListItemGraphic>apps</mdc-icon>
+          <a mat-list-item [appRouterLink]="getRouterLink('translate')" (click)="itemSelected()">
+            <mat-icon matListIcon>apps</mat-icon>
             {{ t("overview") }}
-          </mdc-list-item>
-          <mdc-list-item
+          </a>
+          <a
+            mat-list-item
             *ngFor="let text of texts"
             [appRouterLink]="getRouterLink('translate', getBookId(text))"
-            (selectionChange)="itemSelected()"
+            (click)="itemSelected()"
           >
-            <mdc-icon mdcListItemGraphic>book</mdc-icon>
+            <mat-icon matListIcon>book</mat-icon>
             {{ getBookName(text) }}
-          </mdc-list-item>
+          </a>
         </div>
-        <mdc-list-divider *ngIf="isTranslateEnabled"></mdc-list-divider>
-        <mdc-list-item
+        <mat-divider *ngIf="isTranslateEnabled"></mat-divider>
+        <a
+          mat-list-item
           class="community-checking-nav-item"
           *ngIf="isCheckingEnabled"
-          (selectionChange)="checkingVisible = !checkingVisible"
+          (click)="checkingVisible = !checkingVisible"
         >
-          <mdc-icon mdcListItemGraphic>question_answer</mdc-icon>
+          <mat-icon mat-list-icon>question_answer</mat-icon>
           {{ t("community_checking") }}
-          <mdc-icon *ngIf="!hasSingleAppEnabled" mdcListItemMeta>
-            keyboard_arrow_{{ checkingVisible ? "down" : "right" }}
-          </mdc-icon>
-        </mdc-list-item>
+          <mat-icon *ngIf="!hasSingleAppEnabled">
+            keyboard_arrow_{{ checkingVisible ? "down" : i18n.forwardDirectionWord }}
+          </mat-icon>
+        </a>
         <div *ngIf="isCheckingEnabled && (checkingVisible || hasSingleAppEnabled)">
-          <mdc-list-item [appRouterLink]="getRouterLink('checking')" (selectionChange)="itemSelected()">
-            <mdc-icon mdcListItemGraphic>apps</mdc-icon>
+          <a mat-list-item [appRouterLink]="getRouterLink('checking')" (click)="itemSelected()">
+            <mat-icon mat-list-icon>apps</mat-icon>
             {{ t("overview") }}
-          </mdc-list-item>
-          <mdc-list-item
+          </a>
+          <a
+            mat-list-item
             *ngIf="showAllQuestions"
             [appRouterLink]="getRouterLink('checking', 'ALL')"
-            (selectionChange)="itemSelected()"
+            (click)="itemSelected()"
           >
-            <mdc-icon mdcListItemGraphic>bookmarks</mdc-icon>
+            <mat-icon mat-list-icon>bookmarks</mat-icon>
             {{ t("all_questions") }}
-          </mdc-list-item>
+          </a>
           <ng-container *ngFor="let text of texts">
-            <mdc-list-item
+            <a
+              mat-list-item
               *ngIf="hasQuestions(text)"
               [appRouterLink]="getRouterLink('checking', getBookId(text))"
-              (selectionChange)="itemSelected()"
+              (click)="itemSelected()"
             >
-              <mdc-icon mdcListItemGraphic>book</mdc-icon>
+              <mat-icon mat-list-icon>book</mat-icon>
               {{ getBookName(text) }}
-            </mdc-list-item>
+            </a>
           </ng-container>
         </div>
         <div *ngIf="canSync$ | async">
-          <mdc-list-divider *ngIf="isCheckingEnabled"></mdc-list-divider>
-          <mdc-list-item
+          <mat-divider *ngIf="isCheckingEnabled"></mat-divider>
+          <a
+            mat-list-item
             [appRouterLink]="getRouterLink('sync')"
             [class.list-item-disabled]="!isAppOnline"
-            (selectionChange)="itemSelected()"
+            (click)="itemSelected()"
           >
-            <mdc-icon mdcListItemGraphic>sync</mdc-icon>
+            <mat-icon mat-list-icon>sync</mat-icon>
             {{ t("synchronize") }}
-          </mdc-list-item>
+          </a>
         </div>
         <div *ngIf="canSeeSettings$ | async">
-          <mdc-list-item
+          <a
+            mat-list-item
             [appRouterLink]="getRouterLink('settings')"
             [class.list-item-disabled]="!isAppOnline"
-            (selectionChange)="itemSelected()"
+            (click)="itemSelected()"
           >
-            <mdc-icon mdcListItemGraphic>settings</mdc-icon>
+            <mat-icon mat-list-icon>settings</mat-icon>
             {{ t("settings") }}
-          </mdc-list-item>
+          </a>
         </div>
         <div *ngIf="canSeeUsers$ | async">
-          <mdc-list-item
+          <a
+            mat-list-item
             [appRouterLink]="getRouterLink('users')"
             [class.list-item-disabled]="!isAppOnline"
-            (selectionChange)="itemSelected()"
+            (click)="itemSelected()"
           >
-            <mdc-icon mdcListItemGraphic>people</mdc-icon>
+            <mat-icon mat-list-icon>people</mat-icon>
             {{ t("users") }}
-          </mdc-list-item>
+          </a>
         </div>
-      </mdc-list>
+      </mat-nav-list>
     </mdc-drawer-content>
   </mdc-drawer>
   <!-- The cdkScrollable attribute is needed so the CDK can listen to scroll events within this container -->
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
@@ -35,32 +35,6 @@
           height: unset;
           min-height: 48px;
         }
-        .mdc-list-divider {
-          margin: 8px 0;
-        }
-      }
-    }
-  }
-
-  .mdc-list-item {
-    cursor: pointer;
-    &:hover,
-    &.mdc-list-item--activated {
-      mdc-icon {
-        color: $purpleLight;
-        fill: $purpleLight;
-      }
-    }
-    mdc-icon {
-      color: $purpleMedium;
-      fill: $purpleMedium;
-      transition: color 0.15s;
-    }
-    mdc-icon:first-child {
-      margin-right: 24px;
-      &.ngx-mdc-icon {
-        width: 24px;
-        height: 24px;
       }
     }
   }
@@ -79,7 +53,7 @@
     mdc-top-app-bar-section {
       mdc-menu {
         min-width: 200px;
-        text-align: left; // Needed for some menu items in Chrome but not Firefox
+        text-align: start; // Needed for some menu items in Chrome but not Firefox
         a {
           width: 100%;
           color: var(--mdc-theme-text-primary-on-background);
@@ -172,21 +146,20 @@ a {
   mdc-list-item:not(.active-locale) mdc-icon {
     visibility: hidden;
   }
-  mdc-icon {
-    margin-right: 0.5em;
-  }
 }
 
 .locale-menu .mdc-list-item {
   white-space: nowrap;
   font-family: $languageFonts;
+  column-gap: 0.5em;
 }
 
 .online-status {
+  display: flex;
+  column-gap: 10px;
+  justify-content: center;
+  align-items: center;
   padding-bottom: 8px;
-  mdc-icon {
-    margin-right: 10px;
-  }
 }
 
 :host,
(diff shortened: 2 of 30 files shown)


nigel-wells (changes requested):
This must have been a real interesting one to work on - few new CSS learnings in there for me as well which was nice :)

I've not done a visual test/comparison yet as I'll wait for you to finish everything first.

> Reviewed 26 of 26 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-:-N2yM8KWERlzItrGC92x:b-ljyi9y)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N2yIUyl2rYDWqVgcWIs:-N2yIUyl2rYDWqVgcWIt:b5t0l1w) ([raw file](https://github.com/sillsdev/web-xforge/blob/f9702f524efbbad27e8baa93639c9e2252e11991/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L54)):*
> ```SCSS
>     }
>     mat-icon {
>       &.ngx-mdc-icon {
> ```

Is this selector still targeting what it should as it was previously only applied if it was also a `:first-child`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 126 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N2yLV1c7Qqg4CmcUKL_:-N2yLV1c7Qqg4CmcUKLa:b-el3yug) ([raw file](https://github.com/sillsdev/web-xforge/blob/f9702f524efbbad27e8baa93639c9e2252e11991/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L126)):*
> ```XML
> 
>   get forwardDirectionWord(): 'right' | 'left' {
>     return this.currentLocale.direction === 'ltr' ? 'right' : 'left';
> ```

I know this is correct but it still looks confusing!

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N2yIUyl2rYDWqVgcWIs:-N2yRYtt2rz0GLGKk3Dx:bclohan) ([raw file](https://github.com/sillsdev/web-xforge/blob/f9702f524efbbad27e8baa93639c9e2252e11991/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L54)):*

Is this selector still targeting what it should as it was previously only applied if it was also a `:first-child`
</blockquote></details>

Thanks for mentioning it. I'll take a look. My guess is that it's no longer targeting anything and I forgot to rip it out (but it's too late in my day to look into it). I think the selector near the bottom of the file for `.translate-icon` replaced it.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377)* status: 25 of 26 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N2yIUyl2rYDWqVgcWIs:-N37Wghp2fWF91HVzslu:b-5ozozc) ([raw file](https://github.com/sillsdev/web-xforge/blob/f9702f524efbbad27e8baa93639c9e2252e11991/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L54)):*

Thanks for mentioning it. I'll take a look. My guess is that it's no longer targeting anything and I forgot to rip it out (but it's too late in my day to look into it). I think the selector near the bottom of the file for `.translate-icon` replaced it.
</blockquote></details>

Thanks. I remove a bunch more styles that are no longer being used. From what I can tell everything that is left still does something.

Nateowami (commented):
@nigel-wells The tests are now fixed.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-:-N381tJW6JAxd6mRJfYn:b9dwvam)* status: 23 of 28 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

nigel-wells (changes requested):
I've visually looked over it now - looking good.

I've made one note on the menu icon that uses an SVG. I also noticed this one on the share dialog which will need to change to a `margin-inline-end` instead of the current `margin-right`:


I found this really good write up here on tackling RTL especially for Arabic: https://rtlstyling.com/posts/rtl-styling. I don't think you need to change anything but could be useful to refer back to when we do implement the translation.

> Reviewed 4 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-:-N3HlitHElAtIRgq97zM:b-2dukab)* status: 27 of 28 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N2yIUyl2rYDWqVgcWIs:-N3Hi09vA0GQdBNhTiRs:bs90qxj) ([raw file](https://github.com/sillsdev/web-xforge/blob/f9702f524efbbad27e8baa93639c9e2252e11991/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L54)):*

Thanks. I remove a bunch more styles that are no longer being used. From what I can tell everything that is left still does something.
</blockquote></details>

Some of this is required. The translate icon is actually a SVG image rather than being rendered from a font.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377)* status: 27 of 28 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N2yIUyl2rYDWqVgcWIs:-N3Hv1Nu72lrVTpDbIO_:bc2ngtv) ([raw file](https://github.com/sillsdev/web-xforge/blob/f9702f524efbbad27e8baa93639c9e2252e11991/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L54)):*

Some of this is required. The translate icon is actually a SVG image rather than being rendered from a font.
</blockquote></details>

I know it's an SVG image, but I'm not sure how it follows that removed styles are still required. I think the selector near the bottom of the file for `.translate-icon` replaced it by implementing the same rules. Is there a problem functionally or visually?

nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377)* status: 27 of 28 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N2yIUyl2rYDWqVgcWIs:-N3I1M2l5-0wceXQ5QQI:b-em1jrp) ([raw file](https://github.com/sillsdev/web-xforge/blob/f9702f524efbbad27e8baa93639c9e2252e11991/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L54)):*

I know it's an SVG image, but I'm not sure how it follows that removed styles are still required. I think the selector near the bottom of the file for `.translate-icon` replaced it by implementing the same rules. Is there a problem functionally or visually?
</blockquote></details>

There was a `fill: $purpleLight` style applied when the item in the list was activated/hovered to make the SVG color change - this isn't happening any more so the SVG color isn't changing on hover.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377)* status: 27 of 28 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N2yIUyl2rYDWqVgcWIs:-N3I2nWX7H02Vyj0pBGF:boj1qgt) ([raw file](https://github.com/sillsdev/web-xforge/blob/f9702f524efbbad27e8baa93639c9e2252e11991/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L54)):*

There was a `fill: $purpleLight` style applied when the item in the list was activated/hovered to make the SVG color change - this isn't happening any more so the SVG color isn't changing on hover.
</blockquote></details>

Good catch. Done.

nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377)* status: 27 of 28 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N2yIUyl2rYDWqVgcWIs:-N3I7MnS3aSRHmTVwCQZ:bf3qnyl) ([raw file](https://github.com/sillsdev/web-xforge/blob/f9702f524efbbad27e8baa93639c9e2252e11991/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L54)):*

Good catch. Done.
</blockquote></details>

Thanks - that SVG icon is fixed.

The activated state appears to be broken as well i.e. when you're viewing a book, settings, users, etc. then it should be that purple as well

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377)* status: 27 of 28 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N2yIUyl2rYDWqVgcWIs:-N3I93ZvBbeOnM2vaq1e:b4m7vl4) ([raw file](https://github.com/sillsdev/web-xforge/blob/f9702f524efbbad27e8baa93639c9e2252e11991/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L54)):*

Thanks - that SVG icon is fixed.

The activated state appears to be broken as well i.e. when you're viewing a book, settings, users, etc. then it should be that purple as well
</blockquote></details>

I think I just broke it with the last fix. Try now.

nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377)* status: 27 of 28 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N2yIUyl2rYDWqVgcWIs:-N3IH8M57MZsaUMDf4tb:b-80zvdd) ([raw file](https://github.com/sillsdev/web-xforge/blob/f9702f524efbbad27e8baa93639c9e2252e11991/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L54)):*

I think I just broke it with the last fix. Try now.
</blockquote></details>

Still not fixed for me

Settings is the selected/activated menu item but has the off state for the icon whereas Matthew has the hover state and the light purple. The light purple is the colour settings is supposed to be displaying.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377)* status: 27 of 28 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N2yIUyl2rYDWqVgcWIs:-N3KOk8yDc6L8S3xcBk7:bve4507) ([raw file](https://github.com/sillsdev/web-xforge/blob/f9702f524efbbad27e8baa93639c9e2252e11991/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L54)):*

Still not fixed for me

Settings is the selected/activated menu item but has the off state for the icon whereas Matthew has the hover state and the light purple. The light purple is the colour settings is supposed to be displaying.
</blockquote></details>

I pushed my branch but forgot to commit my changes. Here are my actual changes.

nigel-wells (changes requested):
Happy with the overall changes. Just the update to the share component as mentioned above.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-:-N3Ly8aI89iU0-YGsVEC:blk26kw)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N2yIUyl2rYDWqVgcWIs:-N3Ly0o-AbgdqkBYfFnI:b7kb222) ([raw file](https://github.com/sillsdev/web-xforge/blob/f9702f524efbbad27e8baa93639c9e2252e11991/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L54)):*

I pushed my branch but forgot to commit my changes. Here are my actual changes.
</blockquote></details>

That is working now thanks.

Nateowami (commented):
The share component should now be fixed.

I also updated utils.ts to require newer browsers. In particular it now requires Safari 14.1 on Mac, and Safari 14.5 on iOS. I wasn't sure we wanted to do that, especially since it's possible for us to implement things by writing styles like:
``` scss
.ltr .thing {
  padding-left: 10px;
}
.rtl .thing {
  padding-right: 10px;
}
```
But then I discovered that the flexbox gap property (and row-gap and column-gap) require those versions of Safari as well. Based on data from caniuse, it looks like we would be dropping support for just around 2% of users (that's only considering Safari). That's not ideal, but it's a 2% that will continue to shrink.

Alternatively, we can go back to using margins to create gaps in our flexboxes, but it increases the complexity, and the probability that we'll get it wrong.

Also, we haven't been doing much testing in those browsers, so we don't have any good reason to believe Scripture Forge actually works in those browsers. And it doesn't look like we've set up polyfills, which Angular requires in order to support all but the most recent browsers. Probably we should change that.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-:-N3SC4698MNELFMv1J_G:b-e5wjza)* status: 28 of 30 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @nigel-wells)

nigel-wells (changes requested):
I assume the 2% data you got from Google Analytics? I'm ok with that so long as we don't have any big users in that 2% i.e. someone who is using SF all the time.  It looks like Chrome had a big jump as well. I assume, as you didn't mention, that these changes won't affect many? Chrome is pretty good at self updating compared to Safari which gets tied a lot more closely to the OS version.

In regards to Safari and how well it works, I think we've just kind of supported it in the sense that it is a browser than can access the app and should in theory work. We've never been too proactive to test or announce that we support it. In saying that I know we've had people reporting issues around logging in on Safari so there are definitely people who use it. 

> Reviewed 1 of 1 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-:-N405A0U-EJOjln-C_Dq:bogvtnv)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.scss` line 22 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N404uwH8l840pVcqWcH:-N404uwI5nD1jxvi9tNy:b-2f2skd) ([raw file](https://github.com/sillsdev/web-xforge/blob/a9635e2aaaab057377606bf805fb632171798682/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.scss#L22)):*
> ```SCSS
> }
> 
> mdc-text-field {
> ```

Found another issue related to the popup share component from the translate app and the outline on the share link textbox not being ordered correctly:

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.scss` line 22 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N404uwH8l840pVcqWcH:-N40ClfI5KtBfVoh8TxT:bsz0x4q) ([raw file](https://github.com/sillsdev/web-xforge/blob/a9635e2aaaab057377606bf805fb632171798682/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.scss#L22)):*

Found another issue related to the popup share component from the translate app and the outline on the share link textbox not being ordered correctly:


</blockquote></details>

@nigel-wells I'm not seeing that. Do you have the latest version of the branch with npm dependencies installed? What browser?

Nateowami (commented):
The 2% figure came from looking at https://caniuse.com/flexbox-gap and the usage of Safari browsers that we don't support. According to that, the unsupported Safari browsers would make up about 2.43% of global usage. But we already don't support some of them, so I rounded down to 2%. It's a rough figure, but it would definitely be shrinking over time.

Unsupported desktop versions of Chrome make up 1.35% of global usage, after this change. I don't know what it would have been before, or what it is on mobile.

> Chrome is pretty good at self updating compared to Safari which gets tied a lot more closely to the OS version.

Yes, and for that reason I'm not too concerned about supporting older versions of Chrome. Safari versions on iOS are equal to iOS versions if I understand correctly. And I guess it's too much to expect a company that has $28 billion in cash on the balance sheet to have the resources to ship an evergreen browser, or at least let other browser vendors do so.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-:-N40EMOn3qdVA384QrF4:bwyjk0h)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

nigel-wells (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.scss` line 22 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-N404uwH8l840pVcqWcH:-N40P0xa99-PNAzx41ER:bq4qs4p) ([raw file](https://github.com/sillsdev/web-xforge/blob/74a88c497fb717d509810e179474c524461b3ce1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.scss#L22)):*

@nigel-wells I'm not seeing that. Do you have the latest version of the branch with npm dependencies installed? What browser?
</blockquote></details>

You're right, false alarm. I updated dependencies for your branch and that resolved the issue.

nigel-wells (approved):
In that case I'd do a sanity check in our Google Analytics profile just to see what percentage of our users will be hit.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-:-N40PD3U-ZK2tkhWxWig:b-gpzuen)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami (commented):
Looking at our analytics, I can only find one instance where someone visited a URL that wasn't the home page or about page or something, using an older browser that would no longer be supported. It wasn't an exhaustive search; I focused mostly on iOS, Chrome, and Firefox. That was using iOS 10.3.1, which is already unsupported.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1377#-:-N45HNwl8NxqTGNf3SD5:b-kxjay2)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1381, opened 2022-05-26
SF-1443 Handle char style array

marksvc:
quill-scripture.ts CharInline.create(value) was receiving a UsxStyle[]
array in situations where the textdoc ops had an array for `char`
instead of just a single style. Process this incoming array, storing
it in __usx_value, and storing a space-delimited list of its styles in
the data-style attribute.
_usx.scss can match the style items in the space-delimited list using
`~=`.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1381)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
@@ -329,12 +329,14 @@ usx-segment:first-child usx-blank {
 }
 
 usx-char {
-  &[data-style='it'],
-  &[data-style='em'] {
+  // The data-style attribute is a space-delimited list of styles.
+
+  &[data-style~='it'],
+  &[data-style~='em'] {
     font-style: italic;
   }
 
-  &[data-style='va'] {
+  &[data-style~='va'] {
     font-size: 70%;
     color: #5faa5f;
     top: -0.5em;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
@@ -32,6 +32,7 @@ function getUsxValue<T>(node: HTMLElement): T {
   return node[USX_VALUE];
 }
 
+/** Record textdoc data that was used to make this node, to help later write back to the textdoc. */
 function setUsxValue(node: HTMLElement, value: any): void {
   node[USX_VALUE] = value;
 }
@@ -211,12 +212,29 @@ export function registerScripture(): string[] {
     static blotName = 'char';
     static tagName = 'usx-char';
 
-    static create(value: UsxStyle): Node {
+    static create(value: UsxStyle | UsxStyle[]): Node {
       const node = super.create(value) as HTMLElement;
-      if (value != null && value.style != null) {
-        node.setAttribute(customAttributeName('style'), value.style);
-        setUsxValue(node, value);
+      if (value == null) {
+        return node;
       }
+
+      let characterStyles: string = '';
+      if (Array.isArray(value)) {
+        // Transform an array of styles to a space-delimited list
+        characterStyles = value.map((styleItem: UsxStyle) => styleItem.style).join(' ');
+        if (characterStyles.trim() === '') {
+          return node;
+        }
+      } else {
+        if (value.style == null) {
+          return node;
+        }
+        characterStyles = value.style;
+      }
+
+      node.setAttribute(customAttributeName('style'), characterStyles);
+      setUsxValue(node, value);
+
       return node;
     }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts
@@ -2224,6 +2224,51 @@ describe('TextComponent', () => {
 
     // End drag-and-drop section of tests.
   });
+
+  it('lists multiple character formatting', fakeAsync(() => {
+    const chapterNum = 2;
+    const originSegmentRef = `verse_${chapterNum}_1`;
+    const textDocOps: RichText.DeltaOperation[] = [
+      { insert: { chapter: { number: chapterNum.toString(), style: 'c' } } },
+      { insert: { verse: { number: '1', style: 'v' } } },
+      {
+        insert: `quick brown fox`,
+        attributes: {
+          segment: originSegmentRef,
+          char: {
+            style: 'wj',
+            cid: '1111'
+          }
+        }
+      },
+      {
+        insert: 'jumped over',
+        attributes: {
+          segment: originSegmentRef,
+          char: [
+            {
+              style: 'wj',
+              cid: '1111'
+            },
+            {
+              style: 'w',
+              cid: '2222'
+            }
+          ]
+        }
+      }
+    ];
+
+    const env = new TestEnvironment({ chapterNum, textDoc: textDocOps });
+    const usxCharElements: NodeListOf<Element> = env.component.editor!.container.querySelectorAll('usx-char');
+    expect(usxCharElements.length).withContext('unexpected number of formatted character runs').toEqual(2);
+    expect(usxCharElements[0].attributes.getNamedItem('data-style')?.value)
+      .withContext('first run should just have style wj')
+      .toEqual('wj');
+    expect(usxCharElements[1].attributes.getNamedItem('data-style')?.value)
+      .withContext('second run should have two styles')
+      .toEqual('wj w');
+  }));
 });
 
 /** Represents both what the TextComponent understand to be the text in a segment, and what the editor


RaymondLuong3 (changes requested):
Nice work figuring this out. I tested this and I can see when I edit the nested formatting, the edit and formatting gets synced to Paratext as expected.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1381#-:-N36Abf3A0HqHO7013OA:b-jd987p)* status: all files reviewed, 3 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts` line 35 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1381#-N3661SQ-odgquHXAkYj:-N3661SQ-odgquHXAkYk:be6tmz8) ([raw file](https://github.com/sillsdev/web-xforge/blob/69908c0985a16b761970bf741c3496d154ad3a09/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L35)):*
> ```XML
> }
> 
> /** Make a note about what to later use when writing back to usfm. */
> ```

Do you think this would be an improvement? `Record the value of a usx node that can be used to write usfm`



_Code quote:_
```XML
/** Make a note about what to later use when writing back to usfm. */
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts` line 229 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1381#-N367Gkn29hj7U_lCgSN:-N367Gkn29hj7U_lCgSO:b-virswg) ([raw file](https://github.com/sillsdev/web-xforge/blob/69908c0985a16b761970bf741c3496d154ad3a09/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L229)):*
> ```XML
>         }
>       } else {
>         if (value.style == null) {
> ```

Could this not be at the top. `if (value == null || value.style == null) return node`



_Code quote:_
```XML
        if (value.style == null) {
          return node;
        }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts` line 2251 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1381#-N36AHEm53U_FAc55QJF:-N36AHEm53U_FAc55QJG:b-d52pvu) ([raw file](https://github.com/sillsdev/web-xforge/blob/69908c0985a16b761970bf741c3496d154ad3a09/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L2251)):*
> ```XML
>             {
>               style: 'wj',
>               cid: '6746d56a-77eb-4b72-8d38-4b87f3af32e2'
> ```

Was this on purpose to say that this string of characters is separate from the previous string? I would have expected that the `cid` for the `wj` formatting would be the same.



_Code quote:_
```XML
           cid: '6746d56a-77eb-4b72-8d38-4b87f3af32e2'
```

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1381)* status: all files reviewed, 3 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts` line 35 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1381#-N3661SQ-odgquHXAkYj:-N36Ewkb0c2RaEtWX-AZ:b-6tm95b) ([raw file](https://github.com/sillsdev/web-xforge/blob/69908c0985a16b761970bf741c3496d154ad3a09/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L35)):*

Do you think this would be an improvement? `Record the value of a usx node that can be used to write usfm`
</blockquote></details>

Thank you. I had a further look at this today. I see it's being used in several places. I printed out what's being passed to it from verses, char, para, and note. We are passing it various data from the textdoc, such as formatting information, or the data that is getting inserted. eg, when it is a note, we pass a value of 
```js
{
  caller: "+",
  contents: {
    ...
  },
  style: "f"
}
```
I recall that changing what is passed to this method changes what gets written back to the textdoc. I guess mentioning "usfm" is a bit further down the line than where this data is coming from, or where it's first going (which is the textdoc). 
I'm thinking "Record textdoc data that was used to make this node, to help later write back to the textdoc." 

What do you think?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts` line 229 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1381#-N367Gkn29hj7U_lCgSN:-N36G5JBBOlDE2q-tR0i:b-37llma) ([raw file](https://github.com/sillsdev/web-xforge/blob/69908c0985a16b761970bf741c3496d154ad3a09/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L229)):*

Could this not be at the top. `if (value == null || value.style == null) return node`
</blockquote></details>

Good thinking. When `value` is an Array, `value.style==null`  would be true, tho. So we would first need to see if value is an array.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts` line 2251 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1381#-N36AHEm53U_FAc55QJF:-N36G_O33l3twT3yk9Xs:bfuij9s) ([raw file](https://github.com/sillsdev/web-xforge/blob/69908c0985a16b761970bf741c3496d154ad3a09/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts#L2251)):*

Was this on purpose to say that this string of characters is separate from the previous string? I would have expected that the `cid` for the `wj` formatting would be the same.
</blockquote></details>

Thank you. That was not on purpose. Fixed.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1381)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts` line 35 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1381#-N3661SQ-odgquHXAkYj:-N3LG-D3B-2W5ztb3z3R:b-y50zrv) ([raw file](https://github.com/sillsdev/web-xforge/blob/69908c0985a16b761970bf741c3496d154ad3a09/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L35)):*

Thank you. I had a further look at this today. I see it's being used in several places. I printed out what's being passed to it from verses, char, para, and note. We are passing it various data from the textdoc, such as formatting information, or the data that is getting inserted. eg, when it is a note, we pass a value of 
```js
{
  caller: "+",
  contents: {
    ...
  },
  style: "f"
}
```
I recall that changing what is passed to this method changes what gets written back to the textdoc. I guess mentioning "usfm" is a bit further down the line than where this data is coming from, or where it's first going (which is the textdoc). 
I'm thinking "Record textdoc data that was used to make this node, to help later write back to the textdoc." 

What do you think?

</blockquote></details>

That's a good point that we really are using this for the text doc data. I like this.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts` line 229 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1381#-N367Gkn29hj7U_lCgSN:-N3LGXJ47208BPZrHSx0:b-k42g9a) ([raw file](https://github.com/sillsdev/web-xforge/blob/69908c0985a16b761970bf741c3496d154ad3a09/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L229)):*

Good thinking. When `value` is an Array, `value.style==null`  would be true, tho. So we would first need to see if value is an array.
</blockquote></details>

Oh, you mean that we cannot check `value.style` when `value` is an array. It is only if value is a `usxStyle` that we can check the `style` property. I'm not sure I understand what you were saying above.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1381)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts` line 229 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1381#-N367Gkn29hj7U_lCgSN:-N3PYc_lEbWS-I4QEJzQ:b-3roxeo) ([raw file](https://github.com/sillsdev/web-xforge/blob/69908c0985a16b761970bf741c3496d154ad3a09/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L229)):*

Oh, you mean that we cannot check `value.style` when `value` is an array. It is only if value is a `usxStyle` that we can check the `style` property. I'm not sure I understand what you were saying above.
</blockquote></details>

Right. Well, we can check the `style` property in either case, but in the array case we need to be looking at the `style` property on items _in_ the array, not at a `style` property of the array itself.

If `value` is a `UsxStyle[]`, then `value.style` is undefined, because the array itself won't have a `style` field/property. But `value[0].style` might be defined, because the `UsxStyle` elements of the array might have `.style` defined. 

If I'm still not explaining it well maybe we can chat or call about it.

PR #1383, opened 2022-05-27
Set default font size for source text

RaymondLuong3:
* keep font size up-to-date

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1383)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts
@@ -116,3 +116,8 @@ export function isBadDelta(ops: DeltaOperation[]): boolean {
 export function compareProjectsForSorting(a: { shortName: string }, b: { shortName: string }): 1 | -1 {
   return a.shortName.toLowerCase() < b.shortName.toLowerCase() ? -1 : 1;
 }
+
+export function formatFontSizeToRems(fontSize: number | undefined): string | undefined {
+  // Paratext allows a font size between 8 and 32. 12pt font is equivalent to 1rem
+  return fontSize == null ? undefined : `${fontSize / 12}rem`;
+}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
@@ -45,6 +45,7 @@
             (loaded)="onTextLoaded('source')"
             (updated)="onSourceUpdated($event.delta != null)"
             [isRightToLeft]="isSourceRightToLeft"
+            [fontSize]="sourceFontSize"
           ></app-text>
         </div>
       </div>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -908,12 +908,21 @@ describe('EditorComponent', () => {
 
     it('uses default font size', fakeAsync(() => {
       const env = new TestEnvironment();
-      env.setupProject({ defaultFontSize: 16 });
+      env.setupProject({ defaultFontSize: 18 });
       env.setProjectUserConfig();
       env.wait();
 
       const ptToRem = 12;
-      expect(env.component.fontSize).toEqual(16 / ptToRem + 'rem');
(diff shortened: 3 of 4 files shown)


Nateowami (changes requested):
Sorry I've taken so long to review this. Just a couple things I think should be addressed. Thanks for your work on this.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1383#-:-N3vz-zTDay7sOediHEJ:b44ppja)* status: all files reviewed, 3 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 299 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1383#-N3vz6oU5m09MSb_EnEX:-N3vz6oU5m09MSb_EnEY:b-anvcc1) ([raw file](https://github.com/sillsdev/web-xforge/blob/090c8137e6d51c406328f34d1d1fa588fb4b9a84/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L299)):*
> ```XML
>   get sourceFontSize(): string | undefined {
>     const fontSizeSource: number | undefined = this.sourceProjectDoc?.data?.defaultFontSize;
>     return fontSizeSource == null ? undefined : `${fontSizeSource / 12}rem`;
> ```

These feel a little redundant. Could we create a util method `convertPixelsToRems(value: number | undefined): string | undefined`?

Then we can just do `return convertPixelsToRems(this.sourceProjectDoc?.data?.defaultFontSize)`.

We'll probably want to use this logic outside of the editor anyway, such as in the note dialog (Paratext uses the font size in the note dialog, so we probably will want to do likewise).
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 916 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1383#-N3vwoVAD3MmfI9lDigS:-N3vwoVAD3MmfI9lDigT:btb5nhy) ([raw file](https://github.com/sillsdev/web-xforge/blob/090c8137e6d51c406328f34d1d1fa588fb4b9a84/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L916)):*
> ```XML
> 
>       const ptToRem = 12;
>       expect(env.targetTextEditor.nativeElement.style.fontSize).toEqual(18 / ptToRem + 'rem');
> ```

Every time `env.targetTextEditor` is referenced, it's referenced as `env.targetTextEditor.nativeElement`. Maybe just have it return the element, to simplify lines like this?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 924 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1383#-N3vxys34l1XuvwTC5tp:-N3vxys34l1XuvwTC5tq:bgpdww8) ([raw file](https://github.com/sillsdev/web-xforge/blob/090c8137e6d51c406328f34d1d1fa588fb4b9a84/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L924)):*
> ```XML
>       expect(env.targetTextEditor.nativeElement.style.fontSize).toEqual(24 / ptToRem + 'rem');
>       env.updateFontSize('project02', 24);
>       sourceTextEditor = env.sourceTextArea.nativeElement.querySelector('.ql-container')!;
> ```

This line is superfluous. The tests pass with it commented out, since the element itself is not replaced.

Maybe it was intentional though, with the idea being to look at that as an implementation detail, and not something the test should be concerned with? If so, then I would recommend making `sourceTextEditor` a getter right along side `targetTextEditor`. Generally I think we want to avoid query selectors (and the likes) in the tests themselves, though sometimes it's just simpler, especially when only one test needs it, and it's only needed in one place. So I'm not saying never do this, but we have the same query selector two different places, and that's probably something we generally should avoid.

Nateowami (approved):
:lgtm: Nice work.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1383#-:-N4De2Tq6h6Fl6dAwvop:bcmw7t5)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1383)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1385, opened 2022-05-31
SF-1528 Improve note text selection anchor update after drag and drop event

RaymondLuong3:
In the original issue, the note text anchor was not getting correctly reset if the text was fully removed via a drag and drop event. After experimenting with Paratext, it appears that Paratext does not reset note icons after a similar event, so I've chosen to not try to implement that behaviour.
Along the way, I discovered that we do not update note thread text anchors correctly if the edit delta contained multiple points of edit (such as a drag and drop). Two scenarios I identified are:
- The user drags within a verse
- The user drags text from one verse to another

I've added tests to show that this works now. I've also improved the detection and removal of duplicate note thread icons which appear when a user deletes an icon either through drag and drop, or simply backspace a selection. Now note thread icons are removed anytime duplicates are detected, then they get redrawn.

Here is a summary of the changes
* Note thread icons are reset when the text anchor changes
* Drag and drop events are handled when updating note thread text anchors
* Duplicate note icon positions explicitly recorded
* Duplicate note icons reset when explicitly identified
* Emit snapshot of delta and embedded elements when quill editor is updated
* Remove reference to embedded elements in the segment class

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1385)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -368,6 +368,10 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     return this.viewModel.embeddedElements;
   }
 
+  get embeddedPositions(): number[] {
+    return this.viewModel.embeddedPositions;
+  }
+
   get localPresence(): LocalPresence<PresenceData> | undefined {
     return this.viewModel.localPresence;
   }
@@ -690,7 +694,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     }
     let previousEmbedIndex = -1;
     const deleteDelta = new Delta();
-    for (const embedIndex of this.viewModel.embeddedElements.values()) {
+    for (const embedIndex of this.viewModel.embeddedPositions) {
       const lengthBetweenEmbeds: number = embedIndex - (previousEmbedIndex + 1);
       if (lengthBetweenEmbeds > 0) {
         // retain elements other than notes between the previous and current embed
@@ -1151,6 +1155,10 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
           newStart++;
         }
       }
+
+      // while (embedIndices.includes(newStart - 1)) {
+      //   newStart--;
+      // }
       newSel = { index: newStart, length: Math.max(0, newEnd - newStart) };
     } else {
       return null;
@@ -1177,7 +1185,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
    * positions to check.
    */
   private getEmbedCountInRange(editorStartPos: number, length: number): number {
-    const embedPositions: number[] = Array.from(this.embeddedElements.values());
+    const embedPositions: number[] = this.viewModel.embeddedPositions;
     return embedPositions.filter((pos: number) => pos >= editorStartPos && pos < editorStartPos + length).length;
   }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -368,10 +368,6 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     return this.viewModel.embeddedElements;
   }
 
-  get embeddedPositions(): number[] {
-    return this.viewModel.embeddedPositions;
-  }
-
   get localPresence(): LocalPresence<PresenceData> | undefined {
     return this.viewModel.localPresence;
   }
@@ -694,7 +690,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     }
     let previousEmbedIndex = -1;
     const deleteDelta = new Delta();
-    for (const embedIndex of this.viewModel.embeddedPositions) {
+    for (const embedIndex of this.viewModel.embeddedElements.values()) {
       const lengthBetweenEmbeds: number = embedIndex - (previousEmbedIndex + 1);
       if (lengthBetweenEmbeds > 0) {
         // retain elements other than notes between the previous and current embed
@@ -1194,7 +1190,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
    * positions to check.
    */
   private getEmbedCountInRange(editorStartPos: number, length: number): number {
-    const embedPositions: number[] = this.viewModel.embeddedPositions;
+    const embedPositions: number[] = Array.from(this.viewModel.embeddedElements.values());
     return embedPositions.filter((pos: number) => pos >= editorStartPos && pos < editorStartPos + length).length;
   }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -709,6 +709,18 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     }
   }
 
+  removeEmbeddedElement(embedId: string, source: Sources): void {
+    if (this.editor == null) {
+      return;
+    }
+
+    const position: EmbedPosition | undefined = this.embeddedElements.get(embedId);
+    if (position != null) {
+      const deltaOps: DeltaOperation[] = [{ retain: position.position }, { delete: 1 }];
+      this.editor.updateContents(new Delta(deltaOps), source);
+    }
+  }
+
   isSegmentBlank(ref: string): boolean {
     const segmentDelta: DeltaStatic | undefined = this.getSegmentContents(ref);
     if (segmentDelta?.ops == null) {
@@ -1156,9 +1168,6 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
         }
       }
 
-      // while (embedIndices.includes(newStart - 1)) {
-      //   newStart--;
-      // }
       newSel = { index: newStart, length: Math.max(0, newEnd - newStart) };
     } else {
       return null;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -12,7 +12,7 @@ import {
 import { TranslocoService } from '@ngneat/transloco';
 import isEqual from 'lodash-es/isEqual';
 import merge from 'lodash-es/merge';
-import Quill, { DeltaStatic, RangeStatic, Sources } from 'quill';
+import Quill, { DeltaOperation, DeltaStatic, RangeStatic, Sources } from 'quill';
 import QuillCursors from 'quill-cursors';
 import { AuthType, getAuthType } from 'realtime-server/lib/esm/common/models/user';
 import { TextAnchor } from 'realtime-server/lib/esm/scriptureforge/models/text-anchor';
@@ -714,9 +714,9 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
       return;
     }
 
-    const position: EmbedPosition | undefined = this.embeddedElements.get(embedId);
+    const position: number | undefined = this.embeddedElements.get(embedId);
     if (position != null) {
-      const deltaOps: DeltaOperation[] = [{ retain: position.position }, { delete: 1 }];
+      const deltaOps: DeltaOperation[] = [{ retain: position }, { delete: 1 }];
       this.editor.updateContents(new Delta(deltaOps), source);
     }
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -2310,7 +2310,7 @@ describe('EditorComponent', () => {
       env.targetEditor.setSelection(note1Position - beforeNoteLength, deleteLength, 'user');
       env.deleteCharacters();
       newNotePosition = env.getNoteThreadEditorPosition('thread01');
-      const range = env.component.target!.getSegmentRange('verse_1_1')!;
+      let range = env.component.target!.getSegmentRange('verse_1_1')!;
       // note moves to the beginning of the verse
       expect(newNotePosition).toEqual(range.index);
       env.triggerUndo();
@@ -2331,6 +2331,8 @@ describe('EditorComponent', () => {
       expect(noteThread6.data!.position).toEqual(noteThread6Anchor);
       expect(noteThread1.data!.position).toEqual(noteThread1Anchor);
       expect(textDoc.data!.ops![3].insert).toEqual('target: chapter 1, verse 1.');
+      range = env.component.target!.getSegmentRange('verse_1_1')!;
+      expect(env.targetEditor.getText(range.index, range.length)).toEqual('target: chapter 1, verse 1.');
 
       // undo deleting multiple notes
       const noteThread3: NoteThreadDoc = env.getNoteThreadDoc('project01', 'thread03');
@@ -2357,6 +2359,9 @@ describe('EditorComponent', () => {
       expect(noteThread3.data!.position).toEqual(noteThread3Anchor);
       expect(noteThread4.data!.position).toEqual(noteThread4Anchor);
       expect(textDoc.data!.ops![8].insert).toEqual('target: chapter 1, verse 3.');
+      range = env.component.target!.getSegmentRange('verse_1_3')!;
+      expect(env.targetEditor.getText(range.index, range.length)).toEqual('target: chapter 1, verse 3.');
+
       env.dispose();
     }));
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -305,20 +305,19 @@ describe('EditorComponent', () => {
       env.deleteCharacters();
       segmentRange = env.component.target!.segment!.range;
       segmentContents = env.targetEditor.getContents(segmentRange.index, segmentRange.length);
-
-      // The note remains, the blank returns
+      // the blank returns
       op = segmentContents.ops![0];
-      expect(op.insert).toEqual({
+      expect(op.insert.blank).toBe(true);
+      expect(op.attributes!.segment).toEqual('verse_1_4/p_1');
+      // the note is reset to the start of the verse
+      const verseSegment = env.component.target!.getSegmentContents('verse_1_4')!;
+      expect(verseSegment.ops![0].insert).toEqual({
         'note-thread-embed': {
           iconsrc: '--icon-file: url(/assets/icons/TagIcons/01flag1.png);',
           preview: 'Note from user01',
           threadid: 'thread05'
         }
       });
-      op = segmentContents.ops![1];
-      expect(op.insert.blank).toBe(true);
-      expect(op.attributes!.segment).toEqual('verse_1_4/p_1');
-
       env.dispose();
     }));
 
@@ -1225,7 +1224,7 @@ describe('EditorComponent', () => {
     }));
   });
 
-  fdescribe('Note threads', () => {
+  describe('Note threads', () => {
     it('embeds note on verse segments', fakeAsync(() => {
       const env = new TestEnvironment();
       env.addParatextNoteThread(6, 'MAT 1:2', '', { start: 0, length: 0 }, ['user01']);
@@ -2361,7 +2360,7 @@ describe('EditorComponent', () => {
       env.dispose();
     }));
 
-    fit('note dialog appears after undo delete-a-note', fakeAsync(() => {
+    it('note dialog appears after undo delete-a-note', fakeAsync(() => {
       const env = new TestEnvironment();
       env.setProjectUserConfig();
       env.wait();
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -1225,7 +1225,7 @@ describe('EditorComponent', () => {
     }));
   });
 
-  describe('Note threads', () => {
+  fdescribe('Note threads', () => {
     it('embeds note on verse segments', fakeAsync(() => {
       const env = new TestEnvironment();
       env.addParatextNoteThread(6, 'MAT 1:2', '', { start: 0, length: 0 }, ['user01']);
@@ -1377,7 +1377,7 @@ describe('EditorComponent', () => {
       const noteStart5 = env.component.target!.getSegmentRange('verse_1_4')!.index + doc.data!.position.start;
       // positions are 11, 34, 55, 56, 94
       const expected = [noteStart1, noteStart2, noteStart3, noteStart4, noteStart5];
-      expect(Array.from(env.component.target!.embeddedElements.values())).toEqual(expected);
+      expect(Array.from(env.component.target!.embeddedPositions)).toEqual(expected);
       env.dispose();
     }));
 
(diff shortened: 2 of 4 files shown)


marksvc (changes requested):
> Reviewed 1 of 6 files at r1, 3 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1385)* status: 4 of 6 files reviewed, 14 unresolved discussions (waiting on @marksvc and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 112 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5q7Bpr-g1R5j1NciZm:-N5q7Bpr-g1R5j1NciZn:b-avlxrr) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L112)):*
> ```XML
> }
> 
> export interface EmbedPosition {
> ```

It looks like this is still exported out and used in other files, like editor.component.ts. Can we isolate the duplicate-position information to just text-view-model.ts, so it's not known about by text.component and editor.component?

Hmm, it looks like embeddedElementPositions() would be something that wouldn't be being exported either. Maybe after I study more of the PR I'll think it's good to export them.



_Code quote:_
```XML
export 
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 170 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5q829S4xtd5hFNvFu8:-N5q829S4xtd5hFNvFu9:b52to9n) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L170)):*
> ```XML
> 
>   get segments(): IterableIterator<[string, RangeStatic]> {
>     return cloneDeep(this._segments).entries();
> ```

What's the thinking for returning a clone? Ohh, it's because in text.component.ts you want to grab a clone. Hmmmm. 
It seems like it would be useful to expose the actual segments here. Tho in a quick search, it wasn't clear that we use this in a bunch of places. 



_Code quote:_
```XML
cloneDeep(this._segments).entries();
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 688 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qT_ns7Cr21mBvzjuV:-N5qT_ns7Cr21mBvzjuW:b5jbodh) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L688)):*
> ```XML
>             this._embeddedElements.set(id, embedPosition);
>           } else {
>             embedPosition.duplicatePosition = position;
> ```

What do you think about adding a comment here like,  "Record both positions to later clean up a redundant note embed, such as introduced by undo."
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 756 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5q_qAa3do-VF179KrD:-N5q_qAa3do-VF179KrE:b-ogiezl) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L756)):*
> ```XML
>       e => e.duplicatePosition != null
>     );
>     const deletePositions: number[] = embeddedElementPositions(duplicatePositions).sort((a, b) => a - b);
> ```

(Okay, if this makes it do it in order from those latest in the text, to those earliest in the text.)



_Code quote:_
```XML
 a - b);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 761 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qbGrs7rnu_PMEhsm3:-N5qbGrs7rnu_PMEhsm4:bnurhqu) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L761)):*
> ```XML
>       delta = deleteDelta.compose(delta);
>     }
>     return delta.chop();
> ```

I haven't found documentation for .chop() today. It looks like maybe it removes the last item? 
What is the purpose of chop here?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 85 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qd1Nw1O0TeQixAvp3:-N5qd1Nw1O0TeQixAvp4:b-jsndx8) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L85)):*
> ```XML
> export interface EditedVerseEmbeds {
>   embeds: Map<string, EmbedPosition>;
>   verseRange: RangeStatic;
> ```

I wonder if this packs in more meaning than the data structure ought to be having. Tho I might agree with it when I finish reading the PR.
But it seems that maybe this should be something like
```ts
/** A verse's range, and which embeds are located there. */
export interface EmbedsInVerse {
  verseRange: RangeStatic;
  embeds: Map<string, EmbedPosition>;
}
```
, and then that data structure could be _used_ with the meaning of "for a given verse that was edited", or "the data that was so, _prior_ to a change being applied."



_Code quote:_
```XML
/** Represents information about embeds in a verse prior to a change being applied. */
export interface EditedVerseEmbeds {
  embeds: Map<string, EmbedPosition>;
  verseRange: RangeStatic;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 370 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qeIyZDdpEgVNsoGQM:-N5qeIyZDdpEgVNsoGQN:bz8aor9) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L370)):*
> ```XML
>   get embeddedPositions(): number[] {
>     return this.viewModel.embeddedPositions;
>   }
> ```

Is it important that text.component.ts be able to know about the redundant note positions? Can we keep it all isolated in text-view-model.ts?



_Code quote:_
```XML
  get embeddedPositions(): number[] {
    return this.viewModel.embeddedPositions;
  }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 606 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qedrrB3g1wH15zoVU:-N5qedrrB3g1wH15zoVV:bni9kvn) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L606)):*
> ```XML
>       this.editor.formatText(ep.position, 1, embedName, format, 'api');
>     }
>   }
> ```

But, so doesn't this just take one of the redundant embeds, format it, and leave the other alone?



_Code quote:_
```XML
  formatEmbed(embedId: string, embedName: string, format: any) {
    const ep: EmbedPosition | undefined = this.embeddedElements.get(embedId);
    if (ep != null && this.editor != null) {
      this.editor.formatText(ep.position, 1, embedName, format, 'api');
    }
  }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 694 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qfmKv7pEuJ_KqfLNn:-N5qfmKv7pEuJ_KqfLNo:bf1f1jr) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L694)):*
> ```XML
>     let previousEmbedIndex = -1;
>     const deleteDelta = new Delta();
>     for (const embedIndex of this.viewModel.embeddedPositions) {
> ```

(Is this because we don't clean up the redundant notes in time, before this method is run?)



_Code quote:_
```XML
  for (const embedIndex of this.viewModel.embeddedPositions) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 716 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qgBZn83zVS1gCGoQD:-N5qgBZn83zVS1gCGoQE:b691cm0) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L716)):*
> ```XML
>     const position: EmbedPosition | undefined = this.embeddedElements.get(embedId);
>     if (position != null) {
>       const deltaOps: DeltaOperation[] = [{ retain: position.position }, { delete: 1 }];
> ```

It seems concerning that for a multi-position data structure of EmbedPosition, that when we remove it, we just remove _one_ of the positions. (Not that I think we should remove both; I would desire that when text.component.ts is running its methods, there is only ever one position per embed.)



_Code quote:_
```XML
    const position: EmbedPosition | undefined = this.embeddedElements.get(embedId);
    if (position != null) {
      const deltaOps: DeltaOperation[] = [{ retain: position.position }, { delete: 1 }];
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 915 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qhMS03T_ygcWN7KkW:-N5qhMS03T_ygcWN7KkX:b-x59eus) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L915)):*
> ```XML
>     const affectedEmbeds: EditedVerseEmbeds[] = !!isUserEdit
>       ? this.getEmbedsByEditPosition(delta, segmentsBeforeDelta, embedsBeforeDelta)
>       : [];
> ```

I think remote incoming changes coming in from other users editing are api / non-user changes; would they potentially still need announced, rather than using []?



_Code quote:_
```XML
[]
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 984 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qj4gY8TdYhDihuexM:-N5qj4gY8TdYhDihuexN:bthx8e4) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L984)):*
> ```XML
>     }
>     return true;
>   }
> ```

I was still reviewing here down in this file.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 1846 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5pqVky1JvUX62IPnqU:-N5pqVky1JvUX62IPnqV:bfqr5uy) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1846)):*
> ```XML
>         { retain: spaceBetweenEdits },
>         { insert: dragText }
>       ];
> ```

Because a drag and drop is implemented by distinct delete and insert actions, separated by `setTimeout`, this might not be testing something that actually happens in production. Are you sure this is testing something that happens in practice? Maybe it's something that happens from an update from a remote edit? Or something that is an update from the server SF DB?



_Code quote:_
```XML
      const changes: DeltaOperation[] = [
        { retain: dragStart },
        { delete: dragText.length },
        { retain: spaceBetweenEdits },
        { insert: dragText }
      ];
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 1853 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5prERJ3i9XlE81hfoQ:-N5prERJ3i9XlE81hfoR:b-h54sud) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1853)):*
> getNoteThreadEditorPosition('thread01')).toEqual(verse1Range.index);

I wouldn't have expected the icon to jump to the beginning. Can you specify in a comment that that is what we're looking for, if that's so?



_Code quote:_
```XML
 { start: 0,
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 1892 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5psi794lxdcx30AF0B:-N5psi7A9440wDUQ9IWo:bkwz7vg) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1892)):*
> ```XML
>       // target: $chapter 1, verse 1.
>       //    drag  -------   drop ^
>       const updateOps: DeltaOperation[] = [
> ```

(I have a similar concern here about whether the drag is one set of ops, or distinct ops.)

RaymondLuong3 (commented):
I've made some changes according to the suggestion made. It is not possible to restrict the knowledge of duplicate note embeds to the text-view-model since the editor needs to be aware of it to correctly calculate the update to text anchors.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-:-N69hZdk9byFdqsKIaI9:b-hi0auo)* status: 4 of 6 files reviewed, 13 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 112 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5q7Bpr-g1R5j1NciZm:-N69M4rE3Ln2hDxjYksm:bnkyili) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L112)):*

It looks like this is still exported out and used in other files, like editor.component.ts. Can we isolate the duplicate-position information to just text-view-model.ts, so it's not known about by text.component and editor.component?

Hmm, it looks like embeddedElementPositions() would be something that wouldn't be being exported either. Maybe after I study more of the PR I'll think it's good to export them.
</blockquote></details>

That is correct. It is not possible to limit the duplicate notes information to the text-view-model since the editor.component needs to know what the state of the duplicate embeds are to accurately determine how to update the note anchor (since the delta that is emitted applies to the state of the text editor with the duplicate notes)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 170 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5q829S4xtd5hFNvFu8:-N69NS2R5tDjYoD1LoWP:b-v8a662) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L170)):*

What's the thinking for returning a clone? Ohh, it's because in text.component.ts you want to grab a clone. Hmmmm. 
It seems like it would be useful to expose the actual segments here. Tho in a quick search, it wasn't clear that we use this in a bunch of places. 
</blockquote></details>

This allows us to operate on the text data at a specific instance. But after thinking about it, I decided to introduce a new getter in view-model called segmentsSnapshot which give a snapshot of the segments at a point in time.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 688 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qT_ns7Cr21mBvzjuV:-N69QNI38NB0l-OT-Rgv:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L688)):*

What do you think about adding a comment here like,  "Record both positions to later clean up a redundant note embed, such as introduced by undo."
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 756 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5q_qAa3do-VF179KrD:-N69QRO6E6J2tCjKNuCJ:bp2ouvw) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L756)):*

(Okay, if this makes it do it in order from those latest in the text, to those earliest in the text.)
</blockquote></details>

Actually, this sorts it so that the earlier values come first.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 761 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qbGrs7rnu_PMEhsm3:-N69QfmmBwuvs3rzi65c:b-kf88yq) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L761)):*

I haven't found documentation for .chop() today. It looks like maybe it removes the last item? 
What is the purpose of chop here?
</blockquote></details>

This removes the ops at the end that are unnecessary, such as `retain: <the-length-to-the-end-of-the-doc>`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 370 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qeIyZDdpEgVNsoGQM:-N69RdIzDxU_oV9nGwHP:b-9egask) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L370)):*

Is it important that text.component.ts be able to know about the redundant note positions? Can we keep it all isolated in text-view-model.ts?
</blockquote></details>

This is actually only used in tests. I can adjust it so the tests can access the embed positions in a different way.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 606 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qedrrB3g1wH15zoVU:-N69_Q9u7QyMcOkhW_en:bj7jugt) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L606)):*

But, so doesn't this just take one of the redundant embeds, format it, and leave the other alone?
</blockquote></details>

Since the duplicate embeds get fixed almost instantaneously, we can assume that if we are looking up an embed, its position property is correct. I'll leave a comment on the EmbedPosition class.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 694 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qfmKv7pEuJ_KqfLNn:-N69c8Ut8--deXCZhKZV:b4g4iq8) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L694)):*

(Is this because we don't clean up the redundant notes in time, before this method is run?)
</blockquote></details>

No, I do not expect that duplicate notes will exist at this stage. But we will clean them up if for any reason they do exist.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 716 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qgBZn83zVS1gCGoQD:-N69cnDODDAO1oSt7G5n:b9ge7do) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L716)):*

It seems concerning that for a multi-position data structure of EmbedPosition, that when we remove it, we just remove _one_ of the positions. (Not that I think we should remove both; I would desire that when text.component.ts is running its methods, there is only ever one position per embed.)
</blockquote></details>

You are right, I would prefer if we could guarantee that there would only ever be one embed that text component is aware of. And this will be the case since the duplicate embeds will be cleaned up almost instantaneously. 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 915 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5qhMS03T_ygcWN7KkW:-N69e-cv4PLG7Ikzv9g9:bligwej) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L915)):*

I think remote incoming changes coming in from other users editing are api / non-user changes; would they potentially still need announced, rather than using []?
</blockquote></details>

They do not need to be announced if they are remote edits applied locally. Affected embeds means these embeds may need note anchor updates (which is not the case here since that is handled on the view of the editing user)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 1846 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5pqVky1JvUX62IPnqU:-N69eYN_4GPptEfsbzKn:b-bm0tgy) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1846)):*

Because a drag and drop is implemented by distinct delete and insert actions, separated by `setTimeout`, this might not be testing something that actually happens in production. Are you sure this is testing something that happens in practice? Maybe it's something that happens from an update from a remote edit? Or something that is an update from the server SF DB?
</blockquote></details>

This happens when a user does an undo, all of the drag and drop events are condensed into a single action. I'll update the test title to reflect what we are testing.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 1853 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5prERJ3i9XlE81hfoQ:-N69gpGPDNRnvggR6HEp:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1853)):*

> getNoteThreadEditorPosition('thread01')).toEqual(verse1Range.index);

I wouldn't have expected the icon to jump to the beginning. Can you specify in a comment that that is what we're looking for, if that's so?
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 1892 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1385#-N5psi794lxdcx30AF0B:-N69gr-d1GUdtdTEDu0M:bqkusjm) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba71dbe5a0a5939561b639885f7ae884fc2a4adf/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1892)):*

(I have a similar concern here about whether the drag is one set of ops, or distinct ops.)
</blockquote></details>

I've updated the test to reflect that this is not necessarily a drag and drop, but that it is multiple edit events in a single op.

PR #1390, opened 2022-06-04
SF-1591 Prevent error when switching to book with fewer chapters

Nateowami:
- Previously, switching to a book with n chapters caused an error if the currently active chapter was n+1 or greater.
- This was solved by switching the chapter select from Angular MDC to Angular Material, which we were going to need to do at some point anyway.

One of the tests now uses a component harness, which I don't recall using in the past. See https://material.angular.io/cdk/test-harnesses/overview

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.html
@@ -1,34 +1,26 @@
 <ng-container *transloco="let t; read: 'chapter_nav'">
-  <mdc-select
-    #chapterSelect
-    class="chapter-select"
-    [(ngModel)]="chapterString"
-    (selectionChange)="chapterSelect._selectedText.elementRef.nativeElement.blur()"
-  >
-    <mdc-menu>
-      <mdc-list>
-        <mdc-list-item *ngFor="let cs of chapterStrings" [value]="cs">{{ cs }}</mdc-list-item>
-      </mdc-list>
-    </mdc-menu>
-  </mdc-select>
+  <mat-form-field appearance="fill" id="chapter-select">
+    <mat-select [(value)]="chapter" (valueChange)="chapterChanged()" panelClass="chapter-select-menu">
+      <mat-option *ngFor="let c of chapters" [value]="c">{{ bookName }} {{ c }}</mat-option>
+    </mat-select>
+  </mat-form-field>
   <button
-    mdc-icon-button
-    appBlurOnClick
-    icon="keyboard_arrow_left"
-    type="button"
+    mat-icon-button
     (click)="prevChapter()"
     [disabled]="isPrevChapterDisabled()"
     title="{{ t('previous_chapter') }}"
     fxHide.xs
-  ></button>
+  >
+    <mat-icon>keyboard_arrow_left</mat-icon>
+  </button>
   <button
-    mdc-icon-button
-    appBlurOnClick
+    mat-icon-button
     icon="keyboard_arrow_right"
-    type="button"
     (click)="nextChapter()"
     [disabled]="isNextChapterDisabled()"
     title="{{ t('next_chapter') }}"
     fxHide.xs
-  ></button>
+  >
+    <mat-icon>keyboard_arrow_right</mat-icon>
+  </button>
 </ng-container>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.scss
@@ -3,44 +3,34 @@
 
 :host {
   display: flex;
-  flex-direction: row;
-  justify-content: flex-start;
-  align-items: center;
 }
 
-.mdc-icon-button {
-  @include icon-button.icon-size(24px, 24px, 8px);
-}
-
-// Unfortunately, using Sass mixins to style the select was not successful at the current version of @angular-mdc/web
-// (5.1.1) and @angular/material (9.2.4). Consider reviewing this after upgrading dependencies.
-.chapter-select {
-  ::ng-deep {
-    &,
-    .mdc-select__anchor,
-    .mdc-select__selected-text {
-      height: 40px;
+// Unfortunately there doesn't appear to be a good way to style a mat-select
+// These styles were written by inspecting the elements using developer and then adding rules until it looked right
+#chapter-select ::ng-deep {
+  .mat-form-field-wrapper {
+    padding-bottom: 0;
+    .mat-form-field-flex {
+      padding-top: 0;
     }
-
-    .mdc-select__anchor {
+    .mat-form-field-flex:not(:hover) {
       background-color: transparent;
     }
-
-    .mdc-line-ripple {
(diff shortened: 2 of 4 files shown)


RaymondLuong3 (approved):
I did some testing and it looks to be working. Remember to add acceptance tests since the issue doesn't really explain what the issue is.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1390#-:-N3v3R54EbJtef7PbG8c:b-dtg2rz)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.scss` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1390#-N3v3M3QDrEEKhfo5i7M:-N3v3M3QDrEEKhfo5i7N:b-claxla) ([raw file](https://github.com/sillsdev/web-xforge/blob/0b9cc009bf91fb1eeeddf3c2d6aaed5e635f9a31/src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.scss#L23)):*
> ```SCSS
>     display: none;
>   }
>   .mat-select-arrow-wrapper {
> ```

Good job getting it to look just right.



_Code quote:_
```SCSS
  .mat-select-arrow-wrapper {
    transform: unset;
  }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.spec.ts` line 53 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1390#-N3v0uIe6UCJEDVxCir_:-N3v0uIe6UCJEDVxCira:bvvq4qd) ([raw file](https://github.com/sillsdev/web-xforge/blob/0b9cc009bf91fb1eeeddf3c2d6aaed5e635f9a31/src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.spec.ts#L53)):*
> ```XML
> 
>     // Hacky way to clean up the overlay container
>     document.querySelector('.cdk-overlay-container')!.remove();
> ```

I would put this in a dispose method, or as a tearDown() test lifecycle hook if one exists in jasmine.



_Code quote:_
```XML
    document.querySelector('.cdk-overlay-container')!.remove();
```

Nateowami (commented):
Thanks for the reminder. I tend to forget to add acceptance tests.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1390#-:-N3v5Vw16sIVJeRFT7rB:b7x5ts6)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.scss` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1390#-N3v3M3QDrEEKhfo5i7M:-N3v5_E-Ab6apLrMHBS1:b-5n9lvx) ([raw file](https://github.com/sillsdev/web-xforge/blob/0b9cc009bf91fb1eeeddf3c2d6aaed5e635f9a31/src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.scss#L23)):*

Good job getting it to look just right.
</blockquote></details>

The width isn't dynamic like it was before, but that's not a problem so far as I can tell. We might try to change that in the future, but I thought it was more important to try to get this error squashed, as it's pretty common on live.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.spec.ts` line 53 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1390#-N3v0uIe6UCJEDVxCir_:-N3v5jDLDF5D1_13gqNf:btqog79) ([raw file](https://github.com/sillsdev/web-xforge/blob/0b9cc009bf91fb1eeeddf3c2d6aaed5e635f9a31/src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.spec.ts#L53)):*

I would put this in a dispose method, or as a tearDown() test lifecycle hook if one exists in jasmine.
</blockquote></details>

Good idea. Jasmine has afterEach, but it would require rewriting our tests so it can access env. That would require something like:
``` ts
describe('thing',
  const env: TestEnvironment;
  beforeEach(() => {
    env = new TestEnvironment();
  });
  afterEach(() => cleanUp());

  it('works like it should', () => {
    expect(env.thingsAreFine).toBeTrue();
  });
}
```
And then we couldn't pass custom arguments to the TestEnvironment constructor for each test. And all this for one test that needs to clean up after itself, so it didn't seem worth it.

Nateowami:
I've rebased, fixed conflicts, and this is ready for re-review.

RaymondLuong3 (approved):
> Reviewed 1 of 1 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1390)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.html` line 14 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1390#-N4DM0jzAA0ztsi_mKTu:-N4DM0k-4WzqdoYZ216-:b-8408kt) ([raw file](https://github.com/sillsdev/web-xforge/blob/abcb1b1033ac2814e1e58402ffb54384f76b7680/src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.html#L14)):*
> ```HTML
>     fxHide.xs
>   >
>     <mat-icon>keyboard_arrow_{{ i18n.backwardDirectionWord }}</mat-icon>
> ```

What's the advantage of making this a variable? Does this have an impact when the ui is rtl?



_Code quote:_
```HTML
<mat-icon>keyboard_arrow_{{ i18n.backwardDirectionWord }}</mat-icon>
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1390)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.html` line 14 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1390#-N4DM0jzAA0ztsi_mKTu:-N4DQvtmBlBtmLot3pC9:b2jspy9) ([raw file](https://github.com/sillsdev/web-xforge/blob/abcb1b1033ac2814e1e58402ffb54384f76b7680/src/SIL.XForge.Scripture/ClientApp/src/app/shared/chapter-nav/chapter-nav.component.html#L14)):*

What's the advantage of making this a variable? Does this have an impact when the ui is rtl?
</blockquote></details>

Yes. It's defined in the i18n service as `this.currentLocale.direction === 'ltr' ? 'left' : 'right'`.

You might think the right and left arrows wouldn't need to be changed for a right-to-left script, because you still want one pointing right, and one pointing left. However, the order of the icons gets reversed when we set rtl, so they end up pointing inward. Since the order of the icons is swapped we don't have to change what the buttons do, only how they look.

`forwardDirectionWord` is used 8 other places across 3 other components. It was added to this PR after rebasing after the rtl work got merged.

PR #1396, opened 2022-06-08
Prompt user to log in again when PT tokens are rejected

Nateowami:
This is not finished or polished. Names of methods were just the first thing to come to mind. However, I think it can start to be reviewed.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1396)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/paratext.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/paratext.service.ts
@@ -21,10 +21,18 @@ export class ParatextService {
     this.authService.linkParatext(returnUrl);
   }
 
+  logInWithParatext(returnUrl: string): void {
+    this.authService.getNewTokens(returnUrl);
+  }
+
   getParatextUsername(): Observable<string | undefined> {
     return this.http
       .get<string | null>('paratext-api/username', { headers: this.headers })
       .pipe(map(r => r ?? undefined));
+  }
+
+  async hasValidParatextConnection(): Promise<boolean> {
+    return (await this.getProjects()) != null;
   }
 
   getProjects(): Promise<ParatextProject[] | undefined> {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts
@@ -24,7 +24,7 @@ export class SyncComponent extends DataLoadingComponent implements OnInit {
 
   private _syncActive: boolean = false;
   private isSyncCancelled = false;
-  private paratextUsername?: string;
+  private hasValidParatextConnection: boolean | undefined;
   private previousLastSyncDate?: Date;
 
   constructor(
@@ -39,7 +39,7 @@ export class SyncComponent extends DataLoadingComponent implements OnInit {
   }
 
   get isLoggedIntoParatext(): boolean {
-    return this.paratextUsername != null && this.paratextUsername.length > 0;
+    return this.hasValidParatextConnection === true;
   }
 
   // Todo: This may not be return the correct data on reconnect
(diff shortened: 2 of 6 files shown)


nigel-wells (changes requested):
Are you confident those are the only 2 instances where we need to handle the response from PT around the tokens? I've not looked too much in to what is in bugsnag but I know we have a lot of code interacting with PT. I guess as long as there is a check at the start of any interaction then we should be ok going forward.

Did you want to capture any more data in to Bugsnag as well when throwing the error?

Do you also think we should allow `pullAuthUserProfile` for all logins? The auth0 rule should work but there is no fallback if it doesn't and, if it doesn't, then the tokens stored in mongo will be out of date.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1396#-:-N44giRz2-Fed7ZNDjMF:b-x8f7kh)* status: all files reviewed, 4 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/paratext.service.ts` line 35 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1396#-N44fUX-26bdSRe3EA4n:-N44fUX-26bdSRe3EA4o:birkgzo) ([raw file](https://github.com/sillsdev/web-xforge/blob/5bdb33f4bdf4c3a8d566456bec0058c18f132b4a/src/SIL.XForge.Scripture/ClientApp/src/app/core/paratext.service.ts#L35)):*
> ```XML
> 
>   async hasValidParatextConnection(): Promise<boolean> {
>     return (await this.getProjects()) != null;
> ```

Is this the best way? What if someone legitimately doesn't have any projects or will that be returned as an empty array and thereby not be NULL?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts` line 110 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1396#-N44ex182wm_iTnetDtr:-N44ex197uoYgSg7mMe6:b-uz1j4z) ([raw file](https://github.com/sillsdev/web-xforge/blob/5bdb33f4bdf4c3a8d566456bec0058c18f132b4a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts#L110)):*
> ```XML
>     this.subscribe(this.pwaService.onlineStatus, async isOnline => {
>       this.isAppOnline = isOnline;
>       if (this.isAppOnline && this.hasValidParatextConnection == null) {
> ```

This won't account for the connection breaking later on if it was initially true - probably not a big deal as it would require someone sticking around on the component for a while which isn't likely to happen unless it breaks during a sync for some reason.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 181 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1396#-N44dXuL2MJAhVV3NBBt:-N44dXuL2MJAhVV3NBBu:b-qfcwme) ([raw file](https://github.com/sillsdev/web-xforge/blob/5bdb33f4bdf4c3a8d566456bec0058c18f132b4a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L181)):*
> ```XML
>   }
> 
>   getNewTokens(returnUrl: string): void {
> ```

I'd name this something else as it isn't clear which tokens it is referring to. Is it even needed - why not just call `authWithParatext()` direct unless you felt the need to make that `private`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 185 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1396#-N44e1012Z9KKJK0ckTY:-N44e1012Z9KKJK0ckTZ:b6po8fl) ([raw file](https://github.com/sillsdev/web-xforge/blob/5bdb33f4bdf4c3a8d566456bec0058c18f132b4a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L185)):*
> ```XML
>   }
> 
>   authWithParatext(returnUrl: string, linking: boolean) {
> ```

Perhaps name this `loginWithParatext()` as that is what is happening

josephmyers:
Users have been seeing the 500 error a lot recently. Maybe we should revive this?

Nateowami:
Closing in favor of #2506

PR #1399, opened 2022-06-13
SF-1607 Provide the src property to backend op acknowledge message

RaymondLuong3:
It looks like it will be difficult to figure out how to hook into the realtime server to produce a valid test. I was not able to find the right way to set the src on the connection agent at the handshake stage.
There may still be issues involving using the dev tools to go offline, I see that when we use dev tools, the websocket reports that it is still open even though it cannot deliver the request. I propose that we get this fix to QA, and do some testing by disabling the network adaptor or unplugging the cable. I don't know if we can rely on using the dev tools to simulate the real world.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1399)

### DIFF
--- a/src/RealtimeServer/common/realtime-server.ts
+++ b/src/RealtimeServer/common/realtime-server.ts
@@ -50,6 +50,8 @@ class MigrationConnection extends Connection {
 
 /**
  * This class extends the ShareDB agent class to preserve the migration version property from the request.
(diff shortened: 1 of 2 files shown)


marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1399)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1411, opened 2022-06-24
SF-1626 Prompt for contact if last sync was failure

marksvc:
commit-id:b4ff72a3

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1411)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/_variables.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_variables.scss
@@ -17,6 +17,10 @@ $answers: $greenLight;
 $comments: #ff7300;
 $languageFonts: language_picker, language_picker_fallback, Arial, Helvetica, sans-serif;
 $border-color: rgba(0, 0, 0, 0.12);
+/* A text color that is somewhat muted compared to the primary text color. Note that #666666 is used to match
+ * how mat-hint looks on a white background.
+ */
+$lighterTextColor: #666666;
 
 $pt-button-green: $greenDark;
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html
@@ -2,13 +2,11 @@
   <div fxLayout="column" class="base-container">
     <h2 id="title">{{ t("synchronize_project", { projectName: projectName }) }}</h2>
     <span *ngIf="!isAppOnline" class="offline-text"> {{ t("connect_network_to_synchronize") }} </span>
-    <span *ngIf="syncDisabled" id="syncDisabled-message" class="syncDisabled-text" [innerHTML]="syncDisabledMessage">
-    </span>
+    <span *ngIf="syncDisabled" id="sync-disabled-message" [innerHTML]="syncDisabledMessage"> </span>
     <mat-card class="sync-card">
       <div fxLayout="column" class="card-content">
         <button
           *ngIf="showParatextLogin && isAppOnline"
-          class="action-button"
           mat-flat-button
           type="button"
           (click)="logInWithParatext()"
@@ -21,7 +19,6 @@
         <ng-container *ngIf="isLoggedIntoParatext || isLoading || !isAppOnline">
           <button
             *ngIf="!syncActive"
-            class="action-button"
             mat-flat-button
             color="primary"
             type="button"
@@ -32,9 +29,7 @@
             <mat-icon>sync</mat-icon>
             {{ t("synchronize") }}
           </button>
-          <mat-hint *ngIf="syncActive" id="sync-message" class="sync-info">{{
-            t("your_project_is_being_synchronized")
-          }}</mat-hint>
+          <mat-hint *ngIf="syncActive" id="sync-message">{{ t("your_project_is_being_synchronized") }}</mat-hint>
           <app-sync-progress
             *ngIf="syncActive"
             [projectDoc]="projectDoc"
@@ -42,7 +37,6 @@
           ></app-sync-progress>
           <button
             *ngIf="syncActive"
-            class="action-button"
             mat-button
             type="button"
             (click)="cancelSync()"
@@ -51,11 +45,16 @@
           >
             {{ t("cancel") }}
           </button>
-          <mat-hint *ngIf="!syncActive" [title]="lastSyncDisplayDate" id="date-last-sync" class="sync-info">
+          <mat-hint *ngIf="!syncActive" [title]="lastSyncDisplayDate" id="date-last-sync">
             {{ lastSyncNotice }}
           </mat-hint>
         </ng-container>
       </div>
     </mat-card>
+    <span
+      *ngIf="showSyncFailureSupportMessage"
+      id="sync-failure-support-message"
+      [innerHTML]="syncFailureSupportMessage"
+    ></span>
   </div>
 </ng-container>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss
@@ -1,24 +1,18 @@
-@import 'src/variables';
+@use 'src/variables';
+
+.base-container {
+  align-items: center;
+}
 
 .sync-card {
   padding: 0;
   max-width: 37rem;
   min-width: 18rem;
-  align-self: center;
-}
-
-h2 {
-  align-self: center;
-  text-align: center;
 }
 
 .card-content {
   padding: 1rem;
-  align-self: center;
-}
-
-.action-button {
-  align-self: center;
+  align-items: center;
 }
 
 #btn-sync,
@@ -27,7 +21,7 @@ h2 {
 }
 
 #btn-log-in {
-  background-color: $pt-button-green;
+  background-color: variables.$pt-button-green;
   color: white;
   height: 2.5rem;
 }
@@ -42,16 +36,17 @@ h2 {
   margin-bottom: 1rem;
 }
 
-.sync-info {
-  align-self: center;
-}
-
 .offline-text {
   padding-bottom: 1rem;
-  text-align: center;
 }
 
-.syncDisabled-text {
+#sync-disabled-message {
   padding-bottom: 1rem;
-  text-align: center;
+  max-width: 40rem;
+}
+
+#sync-failure-support-message {
+  padding: 1rem;
+  color: variables.$lighterTextColor;
+  max-width: 40rem;
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.spec.ts
@@ -50,7 +50,7 @@ describe('SyncComponent', () => {
   }));
 
   it('should display Log In to Paratext', fakeAsync(() => {
-    const env = new TestEnvironment(false);
+    const env = new TestEnvironment({ isParatextAccountConnected: false });
     expect(env.title.textContent).toContain('Synchronize Sync Test Project with Paratext');
     expect(env.logInButton.nativeElement.textContent).toContain('Log in to Paratext');
     expect(env.syncButton).toBeNull();
@@ -61,7 +61,7 @@ describe('SyncComponent', () => {
   }));
 
   it('should redirect the user to Log In to Paratext', fakeAsync(() => {
-    const env = new TestEnvironment(false);
+    const env = new TestEnvironment({ isParatextAccountConnected: false });
 
     env.clickElement(env.logInButton);
 
@@ -78,7 +78,7 @@ describe('SyncComponent', () => {
   }));
 
   it('should disable button when offline', fakeAsync(() => {
-    const env = new TestEnvironment(true, false, false);
+    const env = new TestEnvironment({ isParatextAccountConnected: true, isInProgress: false, isOnline: false });
     expect(env.logInButton).toBeNull();
     expect(env.syncButton.nativeElement.disabled).toBe(true);
     expect(env.lastSyncDate.textContent).toContain('Last synced on');
@@ -126,13 +126,39 @@ describe('SyncComponent', () => {
   }));
 
   it('should show progress if in-progress when loaded', fakeAsync(() => {
-    const env = new TestEnvironment(true, true);
+    const env = new TestEnvironment({ isParatextAccountConnected: true, isInProgress: true });
     expect(env.component.syncActive).toBe(true);
     expect(env.progressBar).not.toBeNull();
   }));
 
+  it('should direct to support if last sync was failure', fakeAsync(() => {
+    const env = new TestEnvironment({ lastSyncWasSuccessful: false });
+    expect(env.syncFailureSupportMessage).not.toBeNull();
+  }));
+
+  it('should not direct to support if last sync was not failure', fakeAsync(() => {
+    const env = new TestEnvironment({ lastSyncWasSuccessful: true });
+    expect(env.syncFailureSupportMessage).toBeNull();
+  }));
+
+  it('should not direct to support if last sync success has no record', fakeAsync(() => {
+    const env = new TestEnvironment();
+    env.component.projectDoc!.data!.sync.lastSyncSuccessful = undefined;
+    tick();
+    env.fixture.detectChanges();
+    expect(env.component.projectDoc?.data?.sync.lastSyncSuccessful).withContext('setup').toBeUndefined();
+    expect(env.syncFailureSupportMessage)
+      .withContext('do not show support message when the last sync success record is absent')
+      .toBeNull();
+  }));
+
   it('should explain and disable button when syncDisabled', fakeAsync(() => {
-    const env = new TestEnvironment(true, false, true, true);
+    const env = new TestEnvironment({
+      isParatextAccountConnected: true,
+      isInProgress: false,
+      isOnline: true,
+      isSyncDisabled: true
+    });
     expect(env.logInButton).toBeNull();
     expect(env.syncButton.nativeElement.disabled).toBe(true);
     expect(env.lastSyncDate.textContent).toContain('Last synced on');
@@ -140,7 +166,12 @@ describe('SyncComponent', () => {
   }));
 
   it('should not explain or disable button when not syncDisabled', fakeAsync(() => {
-    const env = new TestEnvironment(true, false, true, false);
+    const env = new TestEnvironment({
+      isParatextAccountConnected: true,
+      isInProgress: false,
+      isOnline: true,
+      isSyncDisabled: false
+    });
     expect(env.syncButton.nativeElement.disabled).toBe(false);
     expect(env.syncDisabledMessage).toBeNull();
   }));
@@ -180,6 +211,14 @@ describe('SyncComponent', () => {
   }));
 });
 
+interface SyncComponentTestConstructorArgs {
+  isParatextAccountConnected?: boolean;
+  isInProgress?: boolean;
+  isOnline?: boolean;
+  isSyncDisabled?: boolean;
+  lastSyncWasSuccessful?: boolean;
+}
+
 class TestEnvironment {
   readonly fixture: ComponentFixture<SyncComponent>;
   readonly component: SyncComponent;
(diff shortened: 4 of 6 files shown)


RaymondLuong3 (approved):
Good thinking. That means we should be able to get active projects going again quicker.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-:-N5aRNCe9RtOvV6R_TQb:b-bggom2)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1411)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts` line 108 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5abUtXDB7gMfiMR5Ie:-N5abUtXDB7gMfiMR5If:bkoiqgi) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4cd3a2373c38a9494931ac22c5e106a326f8f57/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts#L108)):*
> ```XML
> 
>   get showSyncFailureSupportMessage(): boolean {
>     return !this.projectDoc?.data?.sync.lastSyncSuccessful ?? false;
> ```

`this.projectDoc?.data?.sync.lastSyncSuccessful` can be `true`, `false`, or `undefined`. It's then negated using the logical not operator (`!`) which leaves either `true` or `false`. But then the nullish coalescing operator (`??`) is used, as though it could be undefined still.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1411)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts` line 108 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5abUtXDB7gMfiMR5Ie:-N5bKkP6-vDLhvUcxb6f:b-wxydnu) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4cd3a2373c38a9494931ac22c5e106a326f8f57/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts#L108)):*

`this.projectDoc?.data?.sync.lastSyncSuccessful` can be `true`, `false`, or `undefined`. It's then negated using the logical not operator (`!`) which leaves either `true` or `false`. But then the nullish coalescing operator (`??`) is used, as though it could be undefined still.
</blockquote></details>

Thank you. I think what I want here for the desired logic is actually
```ts
return !(this.projectDoc?.data?.sync.lastSyncSuccessful ?? true);
```
But I've had to think about it enough that I'm going to just spell it out.

Nateowami (changes requested):
Sorry to leave a second review with lots more requested changes than my first one. My first review was rushed because I thought it was going to show the message whenever the project data wasn't fully loaded, resulting in a flash of the message before it gets hidden. And the merge button was already green, and I was concerned it would get merged before I had time to finish a review. 

> Reviewed 4 of 5 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-:-N5cHqJbCgkEnQckBGUV:b-ckrt49)* status: all files reviewed, 4 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html` line 60 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5ae9at9X5RXKE7cc4V:-N5ae9at9X5RXKE7cc4W:bj8gb1) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4cd3a2373c38a9494931ac22c5e106a326f8f57/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html#L60)):*
> ```HTML
>       </div>
>     </mat-card>
>     <span
> ```



This really looks out of place aesthetically. Can you change the color (perhaps use a `mat-hint`), left-align the text, set max width to 40 or so ems, and centralize the container (e.g. by setting `align-items: center` on the flex container)?

That makes it look more like this:

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html` line 63 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5afX4MAal-wGdkpgw7:-N5afX4MAal-wGdkpgw8:b-h1w3nf) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4cd3a2373c38a9494931ac22c5e106a326f8f57/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html#L63)):*
> ```HTML
>       *ngIf="showSyncFailureSupportMessage"
>       id="syncFailureSupport-message"
>       class="syncFailureSupport-text"
> ```

Using both a class and an ID is kind of redundant. If you need an ID for the sake of the tests, you might as well just use it in the CSS selector too. Unless there's some reason I'm not thinking of.

Also, I think the mix of camel case and kebab case doesn't follow our typical style. There's one other place in this file that we do that, but besides that the only place where this is done is on the settings page for the IDs of the radio buttons, where it arguably makes more sense. The typical style for this would be all kebab case.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts` line 108 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5abUtXDB7gMfiMR5Ie:-N5abzaSAdFE9ptf9nC7:b5w4e68) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4cd3a2373c38a9494931ac22c5e106a326f8f57/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts#L108)):*

Thank you. I think what I want here for the desired logic is actually
```ts
return !(this.projectDoc?.data?.sync.lastSyncSuccessful ?? true);
```
But I've had to think about it enough that I'm going to just spell it out.
</blockquote></details>

Wouldn't `this.projectDoc?.data?.sync.lastSyncSuccessful === false` be simpler? I think we use that form a lot of places. When it's nullish it's going to be false. 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.spec.ts` line 231 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5cH8QH0EhS-WkU7F2Y:-N5cH8QH0EhS-WkU7F2Z:b-a9ofi2) ([raw file](https://github.com/sillsdev/web-xforge/blob/dc445c6dec890af29cedd7df9bd1553cb4cd7c19/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.spec.ts#L231)):*
> ```XML
>   private isOnline: BehaviorSubject<boolean>;
> 
>   constructor(args: syncComponentTestConstructorArgs = {}) {
> ```

The style we're using is to always capitalize types, even if they have a minor use. That just makes them easier to identify. It took me a bit to figure out this wasn't a variable name.

Making the arguments named is an important improvement though. Calling methods with lots of unnamed arguments gets very confusing.

marksvc (commented):
It's okay :). Thanks for reviewing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-:-N5lhAK_DIo1jIMoUjye:b-4awygn)* status: all files reviewed, 4 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html` line 60 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5ae9at9X5RXKE7cc4V:-N5lh8OtB3GzSZeoOZKn:bmg28hl) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4cd3a2373c38a9494931ac22c5e106a326f8f57/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html#L60)):*



This really looks out of place aesthetically. Can you change the color (perhaps use a `mat-hint`), left-align the text, set max width to 40 or so ems, and centralize the container (e.g. by setting `align-items: center` on the flex container)?

That makes it look more like this:


</blockquote></details>

How about this?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html` line 63 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5afX4MAal-wGdkpgw7:-N5gXyyv2jWaVt0eCohz:b-53fe90) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4cd3a2373c38a9494931ac22c5e106a326f8f57/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html#L63)):*

Using both a class and an ID is kind of redundant. If you need an ID for the sake of the tests, you might as well just use it in the CSS selector too. Unless there's some reason I'm not thinking of.

Also, I think the mix of camel case and kebab case doesn't follow our typical style. There's one other place in this file that we do that, but besides that the only place where this is done is on the settings page for the IDs of the radio buttons, where it arguably makes more sense. The typical style for this would be all kebab case.
</blockquote></details>

Thank you. I guess I copied the one wrong example :). Fixed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts` line 108 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5abUtXDB7gMfiMR5Ie:-N5gPNkz7gN4SpWz0gDF:b-3p8ebc) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4cd3a2373c38a9494931ac22c5e106a326f8f57/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts#L108)):*

Wouldn't `this.projectDoc?.data?.sync.lastSyncSuccessful === false` be simpler? I think we use that form a lot of places. When it's nullish it's going to be false. 
</blockquote></details>

I'm not sure I'm comfortable comparing a left-side null or undefined to a right-side boolean. I've probably done it in other places without thinking about it. I'll give it a go. 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.spec.ts` line 231 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5cH8QH0EhS-WkU7F2Y:-N5gWnTG9d0GF-RMgHus:b-pmq36x) ([raw file](https://github.com/sillsdev/web-xforge/blob/dc445c6dec890af29cedd7df9bd1553cb4cd7c19/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.spec.ts#L231)):*

The style we're using is to always capitalize types, even if they have a minor use. That just makes them easier to identify. It took me a bit to figure out this wasn't a variable name.

Making the arguments named is an important improvement though. Calling methods with lots of unnamed arguments gets very confusing.
</blockquote></details>

Oops. Thanks. Done.

Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1411)* status: all files reviewed, 4 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html` line 60 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5ae9at9X5RXKE7cc4V:-N5mEdCF4y-z-YrJkxOU:beh4vhj) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4cd3a2373c38a9494931ac22c5e106a326f8f57/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html#L60)):*

How about this?
</blockquote></details>

This looks a lot better.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss` line 61 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5ltc5p10AL8zWjZFVO:-N5ltc5p10AL8zWjZFVP:b-aasgy4) ([raw file](https://github.com/sillsdev/web-xforge/blob/3cbd9f5d09517098d13c86c69680eba5e0852bb8/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss#L61)):*
> ```SCSS
> #sync-failure-support-message {
>   padding: 1rem;
>   color: grey;
> ```

I don't think we want to set this to a hard-coded color. For consistency it would be better to use a variable and specify a color we're already using. This color is slightly lighter than the color `mat-hint` uses.

For what it's worth, setting the color to gray is actually better than using an RGBA color like `rgba(0,0,0,.6)`, which is what's used by `mat-hint`, because the transparency can cause problems with some scripts, such as Arabic, which can have overlapping glyphs that start to look really bad if you use a transparent color. I'm not sure when specifying grey colors via transparency became an accepted thing. In the near future we're going to want change all our transparent text colors to not use transparency. That's one more reason why we would want to define our colors in as few places as possible and then just reference them.

I think you're right in thinking this is misusing mat-hint though, as it looks like it's intended to be used with a form field.

But it doesn't look like we've defined a variable for this anywhere, so this is kind of a pre-existing issue, and doesn't have to be solved in this PR.

(It took actually writing this comment for me to realize mat-hint is actually probably the wrong thing to use).
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss` line 62 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5lt9P09Bc7FpSI_SCh:-N5lt9P09Bc7FpSI_SCi:beu3tx3) ([raw file](https://github.com/sillsdev/web-xforge/blob/3cbd9f5d09517098d13c86c69680eba5e0852bb8/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss#L62)):*
> ```SCSS
>   padding: 1rem;
>   color: grey;
>   text-align: left;
> ```

This is superfluous. I did say "left-align the text", but that is what happens by default. Sorry; I could have been more clear. When a RTL UI language is selected this needs to not be set in order to have it automatically change text direction.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss` line 64 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5lt0HbBTuCnoAWml08:-N5lt0Hc1R2W_aFgZoJl:b-ike5py) ([raw file](https://github.com/sillsdev/web-xforge/blob/3cbd9f5d09517098d13c86c69680eba5e0852bb8/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss#L64)):*
> ```SCSS
>   text-align: left;
>   max-width: 40rem;
>   align-self: center;
> ```

Setting `align-self` on all the children doesn't make much sense. It would make a lot more sense to just set `align-items` on the parent element.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts` line 108 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5abUtXDB7gMfiMR5Ie:-N5mEW0g3x3HtcP9KzGG:b-u6i4yf) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4cd3a2373c38a9494931ac22c5e106a326f8f57/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.ts#L108)):*

I'm not sure I'm comfortable comparing a left-side null or undefined to a right-side boolean. I've probably done it in other places without thinking about it. I'll give it a go. 
</blockquote></details>

I don't see a problem with it. I think this is a pattern we use a lot of places. The condition will always be false if it's nullish. If we were using weak equality (`==`) then it would be somewhat concerning because it can lead to type coercion.

marksvc:
Screenshots of the component when thin and wide. (With pieces on that are normally conditional.)

sync-page-thin

sync-page-wide

Nateowami (changes requested):
A few changes requested. I think they should be fairly simple. Thanks for your work on this.

> Reviewed 4 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-:-N5pqoOt4pGarftlL7Hj:b-brg3ew)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/_variables.scss` line 21 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5pq1DlAoCl4dDVQ1tz:-N5pq1DlAoCl4dDVQ1u-:b-3c5qn5) ([raw file](https://github.com/sillsdev/web-xforge/blob/048ac001c22f3aedd83f7cb86fb0e59ced4dc607/src/SIL.XForge.Scripture/ClientApp/src/_variables.scss#L21)):*
> ```SCSS
> $border-color: rgba(0, 0, 0, 0.12);
> /* A text color that stands out less. */
> $subtleTextColor: $greyLight;
> ```

Could we go with `#666666` instead? That will exactly match what mat-hint does (`rgba(0,0,0,.6)`) on a white background, which is nice for consistency. More importantly, it meets contrast rules for visual accessibility: https://webaim.org/resources/contrastchecker/, which `$greyLight` does not. I have good eyes and even I feel this color is a little too light.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss` line 61 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5ltc5p10AL8zWjZFVO:-N5poyXI7uZzY7KnM-qp:b8pr492) ([raw file](https://github.com/sillsdev/web-xforge/blob/3cbd9f5d09517098d13c86c69680eba5e0852bb8/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss#L61)):*

I made a variable in _variables.scss for the subtle text colour. I am having it use another colour that is already used, tho for things like border lines ($greyLight). I'm not sure if this is fitting what you are thinking. We can chat more about it.

BTW I see that _variable.scss has a mix of snake_case, kebab-case, and lowerCamelCase. I went with the more dominant lowerCamelCase, tho I wonder if kebab-case would have been the appropriate choice?
</blockquote></details>

It would be good for us to clean up the style of the variables at some point, but they're already inconsistent, so I wouldn't worry about it too much.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss` line 64 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5lt0HbBTuCnoAWml08:-N5pleZP-BtUKtD8gEM6:bo5s2dh) ([raw file](https://github.com/sillsdev/web-xforge/blob/3cbd9f5d09517098d13c86c69680eba5e0852bb8/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss#L64)):*

Ok I re-worked items to fit into that. How's this? 
</blockquote></details>

I think unnecessary nesting of rules in SCSS greatly harms maintainability. I know it's common in our project, but I don't think it's a good idea (I've experienced trouble interpreting it and refactoring it). I've also read articles that argue it's an anti-pattern (though I'm hesitant to hang my argument on "I found people who agree with me" since that's not much of an argument).

In CSS the way you would write this (i.e., basically the only way you *can* do it) is:
``` css
.base-container {
  align-items: center;
}
.sync-card {
  /* more styles go here */
}
```
SCSS *allows* you to nest the rules, like this:
``` scss
.base-container {
  align-items: center;
  .sync-card {
    /* more styles go here */
  }
}
```
Which compiles to:
``` css
.base-container {
  align-items: center;
}
.base-container .sync-card {
  /* more styles go here */
}
```
Sometimes this is *very* useful and makes it easier to write rules without being so redundant. But there's zero benefit that I know of to using it if you don't have a reason to do so. It does make it harder to tell what the full rule is though, harder to refactor, and produces larger output that is *slightly* slower for the CSS engine to match to elements (that last point is probably more academic than practical).

It also doesn't make much sense to nest an ID selector under something, since IDs are unique anyway.

marksvc (commented):
I changed the variable name to lighterTextColor, since #666666 seems close to the normal text colour.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-:-N5qQd2mEdFUXHFFdA9r:b-xabohi)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/_variables.scss` line 21 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5pq1DlAoCl4dDVQ1tz:-N5qOGYh5b2Q_hNT7qFM:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/048ac001c22f3aedd83f7cb86fb0e59ced4dc607/src/SIL.XForge.Scripture/ClientApp/src/_variables.scss#L21)):*

Could we go with `#666666` instead? That will exactly match what mat-hint does (`rgba(0,0,0,.6)`) on a white background, which is nice for consistency. More importantly, it meets contrast rules for visual accessibility: https://webaim.org/resources/contrastchecker/, which `$greyLight` does not. I have good eyes and even I feel this color is a little too light.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss` line 64 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1411#-N5lt0HbBTuCnoAWml08:-N5qOsmT2jKIPqLaXt1S:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/3cbd9f5d09517098d13c86c69680eba5e0852bb8/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.scss#L64)):*

I think unnecessary nesting of rules in SCSS greatly harms maintainability. I know it's common in our project, but I don't think it's a good idea (I've experienced trouble interpreting it and refactoring it). I've also read articles that argue it's an anti-pattern (though I'm hesitant to hang my argument on "I found people who agree with me" since that's not much of an argument).

In CSS the way you would write this (i.e., basically the only way you *can* do it) is:
``` css
.base-container {
  align-items: center;
}
.sync-card {
  /* more styles go here */
}
```
SCSS *allows* you to nest the rules, like this:
``` scss
.base-container {
  align-items: center;
  .sync-card {
    /* more styles go here */
  }
}
```
Which compiles to:
``` css
.base-container {
  align-items: center;
}
.base-container .sync-card {
  /* more styles go here */
}
```
Sometimes this is *very* useful and makes it easier to write rules without being so redundant. But there's zero benefit that I know of to using it if you don't have a reason to do so. It does make it harder to tell what the full rule is though, harder to refactor, and produces larger output that is *slightly* slower for the CSS engine to match to elements (that last point is probably more academic than practical).

It also doesn't make much sense to nest an ID selector under something, since IDs are unique anyway.
</blockquote></details>

Done.

PR #1418, opened 2022-07-01
Rework styling of sync component

Nateowami:
- Move sync failure support message above sync controls, where other messages are
- Make color of support messages consistent
- Use row-gap for spacing elements in flex containers, rather than margin

I've also avoided using fxLayout when it's just as easy to use CSS to do the styling. It seems odd and confusing to write something like:
``` css
.base-container {
  align-items: center;
}
```
because there's no indication in our CSS that `.base-container` is a flex container, because we set that using `fxLayout="column"` in the template. There may be times that fxLayout makes producing the layout we want easier, and it makes sense to use it then. But I don't see the benefit in this particular case.

Here's how it looks with all elements visible:

And here's how that looks when the flex layout is visualized:

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1418)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync.component.html
@@ -1,10 +1,15 @@
 <ng-container *transloco="let t; read: 'sync'">
-  <div fxLayout="column" class="base-container">
+  <div class="base-container">
     <h2 id="title">{{ t("synchronize_project", { projectName: projectName }) }}</h2>
     <span *ngIf="!isAppOnline" class="offline-text"> {{ t("connect_network_to_synchronize") }} </span>
     <span *ngIf="syncDisabled" id="sync-disabled-message" [innerHTML]="syncDisabledMessage"> </span>
+    <span
+      *ngIf="showSyncFailureSupportMessage"
+      id="sync-failure-support-message"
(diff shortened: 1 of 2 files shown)


Nateowami (commented):
I agree. It's worth noting that in the past the browsers we claimed to support didn't all support the `row-gap` property. Only recently when working on the RTL UI support I wanted to use some properties that required changing the list of supported browsers. Once that was done, all supported browsers support the `row-gap` property. (Keep in mind that "supported browsers" aren't necessarily actually supported; Angular doesn't officially support all versions of browsers we claim to support...).

There's also the `column-gap` property, which is even more useful, because it will create a horizontal gap between elements regardless of the direction of the UI. Previously we'd do something like `padding-right: 10px`, which would need to be `padding-left: 10px` for a RTL UI. You can do `padding-inline-end: 10px`, which should theoretically set the padding on the right for LTR, and on the left for RTL, but it can be more finicky, since the text direction isn't necessarily the same everywhere on the page. Also, you generally want to set the padding for all elements except the last one, which is trickier than just specifying a gap (especially if some elements are conditional).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1418#-:-N6AWcjM4i_NRaTpYyPk:b-i23ho2)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #1419, opened 2022-07-01
Update all translation files from Crowdin

Nateowami:
I went through the books of the Bible in Arabic and suggested the translation Paratext uses. There are a couple reasons for this.

One of them is that our testers keep seeing places where the behavior is wrong because the text is not really right-to-left. It looks like a bug, but once you pull in the translations, everything starts behaving correctly. Here's how it looks for 2 Samuel 1:

Before | After
----------|-------
 | 

The second reason is that the translator is unlikely to know to trust and use the Paratext translations. And we probably want to match Paratext as closely as possible most of the time.

However, I did see that for Ecclesiastes, there was a " 1" (space followed by 1) at the end of the translation. I did not include that. My guess is it is an incorrect translation in Paratext. I'm going to try to find out for sure. Bible Gateway doesn't include a "1" in the name, and it seems a bit odd that it would be added.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1419)


RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1419)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/Resources/Pages.LearnMore.zh-CN.resx` line 244 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1419#-N691-GxAERJMOGOsoHw:-N691-GxAERJMOGOsoHx:b2fhitt) ([raw file](https://github.com/sillsdev/web-xforge/blob/0b255d9393653d122e1715028c987a8e4a6aeca9/src/SIL.XForge.Scripture/Resources/Pages.LearnMore.zh-CN.resx#L244)):*
> ```XML
>   </data>
>   <data name="ReadyToTry" xml:space="preserve">
>     <value>准备好尝试《经文锻造》了吗？</value>
> ```

I looked to see where this came from. I was surprised to learn that Ken knew simplified Chinese!



_Code quote:_
```XML
准备好尝试《经文锻造》了吗？
```

PR #1420, opened 2022-07-02
SF-1555 Prevent admin pages from scrolling away

Nateowami:
- Only part of the menu is now scrollable.
- Community checking no longer wraps in English when the scrollbar appears.
- Some incorrect behavior regarding when dividers are shown in the menu has been fixed (for example, for resources, previously a divider was shown under the list of books).
- I removed a few pixels of what I think was unnecessary padding at the top of the menu.

Before | After
----------|-------
 | 

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1420)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html
@@ -159,7 +159,7 @@
       </mdc-select>
     </mdc-drawer-header>
     <mdc-drawer-content>
-      <mat-nav-list id="menu-list">
+      <mat-nav-list id="tools-menu-list">
         <a
           mat-list-item
           class="translate-nav-item"
@@ -187,7 +187,7 @@
             {{ getBookName(text) }}
           </a>
         </div>
-        <mat-divider *ngIf="isTranslateEnabled"></mat-divider>
+        <mat-divider *ngIf="isCheckingEnabled && isTranslateEnabled"></mat-divider>
         <a
           mat-list-item
           class="community-checking-nav-item"
@@ -226,8 +226,10 @@
             </a>
           </ng-container>
         </div>
+      </mat-nav-list>
+      <mat-nav-list id="admin-pages-menu-list" *ngIf="canSeeAdminPages$ | async">
+        <mat-divider></mat-divider>
         <div *ngIf="canSync$ | async">
-          <mat-divider *ngIf="isCheckingEnabled"></mat-divider>
           <a
             mat-list-item
             [appRouterLink]="getRouterLink('sync')"
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
@@ -205,6 +205,7 @@ mat-nav-list .mat-list-item {
 
   ::ng-deep .mat-list-item-content {
     column-gap: 16px;
+    padding-inline-end: 8px;
   }
 }
 
@@ -212,6 +213,17 @@ mat-nav-list .mat-list-item {
   background-color: #e7e8e7;
 }
 
-#menu-list {
-  padding-top: 6px;
+mdc-drawer-content {
+  display: flex;
+  flex-direction: column;
+  justify-content: space-between;
+}
+
+#tools-menu-list,
+#admin-pages-menu-list {
+  padding-top: 0;
+}
+
+#tools-menu-list {
+  overflow-y: auto;
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -634,7 +634,7 @@ class TestEnvironment {
   }
 
   get menuListItems(): DebugElement[] {
-    return this.fixture.debugElement.queryAll(By.css('#menu-list .mat-list-item'));
+    return this.fixture.debugElement.queryAll(By.css('#menu-drawer .mat-list-item'));
   }
 
   get helpMenuList(): DebugElement {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
@@ -69,6 +69,8 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
   canSeeSettings$?: Observable<boolean>;
   canSeeUsers$?: Observable<boolean>;
   canSync$?: Observable<boolean>;
+  /** Whether the user can see at least one of settings, users, or sync page */
+  canSeeAdminPages$?: Observable<boolean>;
   hasUpdate: boolean = false;
 
   projectLabel = projectLabel;
@@ -334,6 +336,9 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
           this.canSeeSettings$ = this.settingsAuthGuard.allowTransition(projectId);
           this.canSeeUsers$ = this.usersAuthGuard.allowTransition(projectId);
           this.canSync$ = this.syncAuthGuard.allowTransition(projectId);
+          this.canSeeAdminPages$ = combineLatest([this.canSeeSettings$, this.canSeeUsers$, this.canSync$]).pipe(
+            map(([settings, users, sync]) => settings || users || sync)
+          );
           // the project deleted dialog should be closed by now, so we can reset its ref to null
           if (projectId == null) {
             this.projectDeletedDialogRef = null;


RaymondLuong3 (changes requested):
I'll mention that it looks a little odd to have so much white space in navigation menu when setting mdc-drawer-content { justify-content: space-between; }


> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1420#-:-N695rgT5FpcWJGv5RMq:b-5hvqhj)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html` line 232 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1420#-N693go27NnjKY_SbYrf:-N693go27NnjKY_SbYrg:bfm71hr) ([raw file](https://github.com/sillsdev/web-xforge/blob/236b0b283dbaa12c9c7c78dea8763456e88309db/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L232)):*
> ```HTML
>       <mat-nav-list
>         id="admin-pages-menu-list"
>         *ngIf="(canSync$ | async) || (canSeeSettings$ | async) || (canSeeUsers$ | async)"
> ```

this could all be condensed in the ts file into one property that this template can read. It may be cleaner that way.



_Code quote:_
```HTML
        *ngIf="(canSync$ | async) || (canSeeSettings$ | async) || (canSeeUsers$ | async)"
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 223 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1420#-N696NtcADCG5Km8Z5zc:-N696NtcADCG5Km8Z5zd:b-uy1f5n) ([raw file](https://github.com/sillsdev/web-xforge/blob/236b0b283dbaa12c9c7c78dea8763456e88309db/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L223)):*
> ```SCSS
> 
> #tools-menu-list,
> #admin-pages-menu-list {
> ```

It doesn't appear that this line affects anything that I can see.



_Code quote:_
```SCSS
#admin-pages-menu-list {
```

Nateowami (commented):
It does look somewhat odd, but I'm not sure it looks more odd than previously, when it could also have a lot of blank space:


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1420#-:-N6AZuju3WWSAAKHvwT1:bs1c17y)* status: 1 of 4 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html` line 232 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1420#-N693go27NnjKY_SbYrf:-N6ELqumCeCmb3nO0kmQ:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/236b0b283dbaa12c9c7c78dea8763456e88309db/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L232)):*

this could all be condensed in the ts file into one property that this template can read. It may be cleaner that way.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 223 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1420#-N696NtcADCG5Km8Z5zc:-N6AZ5wY-iDibS49kxJX:biztj4) ([raw file](https://github.com/sillsdev/web-xforge/blob/236b0b283dbaa12c9c7c78dea8763456e88309db/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L223)):*

It doesn't appear that this line affects anything that I can see.
</blockquote></details>

Without it you get this weird look. Note the gap between the divider and the bottom of the scroll element.

RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1420)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html` line 232 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1420#-N693go27NnjKY_SbYrf:-N6E_0rE0X0yuz4MXQ43:bmfeaba) ([raw file](https://github.com/sillsdev/web-xforge/blob/236b0b283dbaa12c9c7c78dea8763456e88309db/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L232)):*

Done.
</blockquote></details>

I wonder if this should have a async pipe? And I do not believe that the "this" is necessary.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 223 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1420#-N696NtcADCG5Km8Z5zc:-N6EWrGz5Z-0zMm5x0uE:b-a9pqfg) ([raw file](https://github.com/sillsdev/web-xforge/blob/236b0b283dbaa12c9c7c78dea8763456e88309db/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L223)):*

Without it you get this weird look. Note the gap between the divider and the bottom of the scroll element.


</blockquote></details>

Ok, I see it now.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1420)* status: 3 of 4 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html` line 232 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1420#-N693go27NnjKY_SbYrf:-N6EaEOz9tZSZyFF4q7a:b-yqzk0g) ([raw file](https://github.com/sillsdev/web-xforge/blob/236b0b283dbaa12c9c7c78dea8763456e88309db/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L232)):*

I wonder if this should have a async pipe? And I do not believe that the "this" is necessary.
</blockquote></details>

Oh, good catch. Presumably this would always be true.

I didn't catch it in my manual testing because another mistake I just noticed was masking the incorrect logic. The `mat-divider` should be a direct descendant of `mat-nav-list`, rather than included with the sync list item. In fact, the *only* reason we need to make the `#admin-pages-menu-list` element conditional is because we don't want it to include the `mat-divider` when there are no items below the divider. It just so happens that in every case where a user can see at least one admin page, the user can see the sync page.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1420)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html` line 232 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1420#-N693go27NnjKY_SbYrf:-N6EvLtx8zjUmWpfR9vk:b-d1hx2p) ([raw file](https://github.com/sillsdev/web-xforge/blob/236b0b283dbaa12c9c7c78dea8763456e88309db/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L232)):*

Hmm, I'm surprised that we did not catch this in a test. I see there is a test in `app.component.spec.ts`. Are you aware of why this test didn't catch that? Maybe the observable does emit a value that the *ngIf can read without the async pipe.
</blockquote></details>

@raymondluong3 That's because the tests are concerned only with which menu items are shown, and they were always properly shown when they should be shown. The only way you would catch this is by noticing that the divider was shown when it shouldn't have been shown. But that wasn't actually happening, since the way the divider was previously made conditional was to show it when the sync menu item is shown.

Put another way:
This conditional is not very important. I added it to fix incorrect behavior regarding when the divider is shown. But it turns out that because I forgot to move the divider, the divider was controlled by the old logic as well. They combined to produce (by accident) the correct behavior, even though this condition was always evaluating to true.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1420)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html` line 232 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1420#-N693go27NnjKY_SbYrf:-N6FEY-Y-MHzq8pv1Tc0:b-kkq7z1) ([raw file](https://github.com/sillsdev/web-xforge/blob/236b0b283dbaa12c9c7c78dea8763456e88309db/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L232)):*

@raymondluong3 That's because the tests are concerned only with which menu items are shown, and they were always properly shown when they should be shown. The only way you would catch this is by noticing that the divider was shown when it shouldn't have been shown. But that wasn't actually happening, since the way the divider was previously made conditional was to show it when the sync menu item is shown.

Put another way:
This conditional is not very important. I added it to fix incorrect behavior regarding when the divider is shown. But it turns out that because I forgot to move the divider, the divider was controlled by the old logic as well. They combined to produce (by accident) the correct behavior, even though this condition was always evaluating to true.
</blockquote></details>

Ok, thanks for digging into that.

PR #1421, opened 2022-07-04
SF-1615 Allow disabling suggestions on a project-wide basis

nigel-wells:
- Always show based on project on settings page
- Hide translate suggestions checkbox until a based on project/resource has been selected

**New default settings screen for translations**
image

**Translate suggestions will appear once a based on project has been set**
image

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1421)


RaymondLuong3 (changes requested):
Nice work. I have some relatively minor suggestions.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-:-N69IrZi4BgNKwpJvjK7:bgya6v9)* status: all files reviewed, 7 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts` line 250 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N69ASNH0SZhkzKh2GHv:-N69ASNH0SZhkzKh2GHw:bb8qho4) ([raw file](https://github.com/sillsdev/web-xforge/blob/d56a97058a4dbc4373949f50ae0c366bfd27c9eb/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts#L250)):*
> ```XML
>     if (sourceProjectChanged) {
>       const settings: SFProjectSettings = {
>         sourceParatextId: newValue.sourceParatextId != null ? newValue.sourceParatextId : 'unset'
> ```

What about assigning 'unset' to a class member variable so it doesn't have to be hard coded?



_Code quote:_
```XML
'unset'
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts` line 194 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N69Ac_RCqasavjsqY3y:-N69Ac_S5YWd4FuboVZF:bew4nbs) ([raw file](https://github.com/sillsdev/web-xforge/blob/d56a97058a4dbc4373949f50ae0c366bfd27c9eb/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts#L194)):*
> ```XML
>       }));
> 
>       it('should hide Translation Suggestions when Based On  is not set', fakeAsync(() => {
> ```

nit: looks like there is an extra space here.



_Code quote:_
```XML
when Based On  is not set'
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts` line 303 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N69ECng6vnjVhcIw-cg:-N69ECng6vnjVhcIw-ch:bbobiay) ([raw file](https://github.com/sillsdev/web-xforge/blob/d56a97058a4dbc4373949f50ae0c366bfd27c9eb/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts#L303)):*
> ```XML
>       }));
> 
>       it('should not save Translation Suggestions when Based On is changed', fakeAsync(() => {
> ```

This test does not show that if the checkbox was checked previously, selecting a new based on project resets the checkbox. I think that is missing from this test.



_Code quote:_
```XML
'should not save Translation Suggestions when Based On is changed'
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts` line 320 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N69E1mVCh377DHdufrA:-N69E1mVCh377DHdufrB:bwn1smf) ([raw file](https://github.com/sillsdev/web-xforge/blob/d56a97058a4dbc4373949f50ae0c366bfd27c9eb/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts#L320)):*
> ```XML
> 
>         env.setBasedOnValue('paratextId02');
>         expect(env.inputElement(env.translationSuggestionsCheckbox).checked).toBe(false);
> ```

nit: The order for these expects would ideally match



_Code quote:_
```XML
        expect(env.inputElement(env.translationSuggestionsCheckbox).checked).toBe(false);
        expect(env.basedOnSelectValue).toContain('ParatextP2');
        expect(env.statusDone(env.translationSuggestionsStatus)).toBeNull();
```
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 242 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N69G4HpA0_0I5XvBfnu:-N69G4HpA0_0I5XvBfnv:b-4cx43b) ([raw file](https://github.com/sillsdev/web-xforge/blob/d56a97058a4dbc4373949f50ae0c366bfd27c9eb/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L242)):*
> ```Smalltalk
>                 // Get the source - any creation or permission updates are handled in GetTranslateSourceAsync
>                 TranslateSource source = null;
>                 bool unsetSourceProject = settings.SourceParatextId == "unset";
> ```

How about using a member variable to defined the 'unset' string?



_Code quote:_
```Smalltalk
   bool unsetSourceProject = settings.SourceParatextId == "unset";
```
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 286 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N69GhHO63qucFoJto1i:-N69GhHO63qucFoJto1j:bc1tohw) ([raw file](https://github.com/sillsdev/web-xforge/blob/d56a97058a4dbc4373949f50ae0c366bfd27c9eb/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L286)):*
> ```Smalltalk
>                             && projectDoc.Data.TranslateConfig.Source != null)
>                         {
>                             // translate task was enabled or source project changed
> ```

This comment could be updated. Something like "translation suggestions was enabled"



_Code quote:_
```Smalltalk
  // translate task was enabled or source project changed
```
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 302 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N69GjaJ7KafpdX3X9xz:-N69GjaJ7KafpdX3X9y-:b-aw43lg) ([raw file](https://github.com/sillsdev/web-xforge/blob/d56a97058a4dbc4373949f50ae0c366bfd27c9eb/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L302)):*
> ```Smalltalk
>                         else
>                         {
>                             // translate task was disabled or source project set to null
> ```

This comment could be updated. Maybe "translation suggestions was disabled or the source project was unset.



_Code quote:_
```Smalltalk
    // translate task was disabled or source project set to null
```

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1421)* status: 6 of 9 files reviewed, 5 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts` line 303 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N69ECng6vnjVhcIw-cg:-N6BMWIdB0faUHRpMawa:brloyq2) ([raw file](https://github.com/sillsdev/web-xforge/blob/d56a97058a4dbc4373949f50ae0c366bfd27c9eb/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts#L303)):*

This test does not show that if the checkbox was checked previously, selecting a new based on project resets the checkbox. I think that is missing from this test.
</blockquote></details>

Does this happen for you? Changing the based on project to another project shouldn't touch the translation suggestions so if it was checked then it should still be checked after changing projects. I've tried it again now and that is what I observe. Are you seeing something else happen?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts` line 320 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N69E1mVCh377DHdufrA:-N6BKMCj9Jw0E_s300zb:b4hgsii) ([raw file](https://github.com/sillsdev/web-xforge/blob/d56a97058a4dbc4373949f50ae0c366bfd27c9eb/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts#L320)):*

nit: The order for these expects would ideally match
</blockquote></details>

I'm not really sure what you want me to match? I assume you mean the previous set of expects but they are slightly different in terms of what is being checked.
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 242 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N69G4HpA0_0I5XvBfnu:-N6BJTkUBrTmiFyYgnPl:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/d56a97058a4dbc4373949f50ae0c366bfd27c9eb/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L242)):*

How about using a member variable to defined the 'unset' string?
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 286 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N69GhHO63qucFoJto1i:-N6BHiOT2xbWiOTdi0yQ:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/d56a97058a4dbc4373949f50ae0c366bfd27c9eb/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L286)):*

This comment could be updated. Something like "translation suggestions was enabled"
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 302 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N69GjaJ7KafpdX3X9xz:-N6BHioY-BLB4Ri70x3o:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/d56a97058a4dbc4373949f50ae0c366bfd27c9eb/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L302)):*

This comment could be updated. Maybe "translation suggestions was disabled or the source project was unset.
</blockquote></details>

Done.

RaymondLuong3 (changes requested):
Almost good. Just a couple more suggestions.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-:-N6E5wwD0mzk3yhLExpq:b-7w0jpb)* status: all files reviewed, 2 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts` line 303 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N69ECng6vnjVhcIw-cg:-N6E4JLgC_jbxO5hnIcX:b-nc6u6g) ([raw file](https://github.com/sillsdev/web-xforge/blob/d56a97058a4dbc4373949f50ae0c366bfd27c9eb/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts#L303)):*

Does this happen for you? Changing the based on project to another project shouldn't touch the translation suggestions so if it was checked then it should still be checked after changing projects. I've tried it again now and that is what I observe. Are you seeing something else happen?
</blockquote></details>

I must have read the test description wrong. I understood it as testing that the translation suggestion gets reset if based on is changed. But looking at the test again, it says that we expect that selecting the based on project does not enable translation suggestions by default. Would that be a better name for this test?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts` line 320 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N69E1mVCh377DHdufrA:-N6E4BSp5sXtXl6vBM-X:biya7e9) ([raw file](https://github.com/sillsdev/web-xforge/blob/d56a97058a4dbc4373949f50ae0c366bfd27c9eb/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts#L320)):*

I'm not really sure what you want me to match? I assume you mean the previous set of expects but they are slightly different in terms of what is being checked.
</blockquote></details>

That's a good point, there are slight differences. Ok, this is reasonable.
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 33 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N6E4jUA6T109Wdq617G:-N6E4jUA6T109Wdq617H:b-xrv7w) ([raw file](https://github.com/sillsdev/web-xforge/blob/68faf24b1bd9df4b394ea3ede8ac7d64b698e0c0/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L33)):*
> ```Smalltalk
>             new DictionaryComparer<string, string>();
>         public static readonly string ErrorAlreadyConnectedKey = "error-already-connected";
>         public static readonly string ProjectSettingValueUnset = "unset";
> ```

Since this is not used outside this class, lets make it private



_Code quote:_
```Smalltalk
   public static readonly string ProjectSettingValueUnset = "unset";
```

RaymondLuong3 (commented):
Nathaniel pointed out that the connect project page should be reworked too.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-:-N6ENBzA83MdbD8_hKKD:brn1f3)* status: all files reviewed, 2 unresolved discussions (waiting on @nigel-wells)

nigel-wells (commented):
I've update the connect project component now as well.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-:-N6GZnklEXCYkcYZz6bN:b9j6zas)* status: 6 of 14 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts` line 303 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N69ECng6vnjVhcIw-cg:-N6GDlnLA9xZrsBtIIIV:bomlctr) ([raw file](https://github.com/sillsdev/web-xforge/blob/d56a97058a4dbc4373949f50ae0c366bfd27c9eb/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.spec.ts#L303)):*

I must have read the test description wrong. I understood it as testing that the translation suggestion gets reset if based on is changed. But looking at the test again, it says that we expect that selecting the based on project does not enable translation suggestions by default. Would that be a better name for this test?
</blockquote></details>

Ah, I see how you came to that conclusion now. I've renamed the test as well as added a couple more expects to ensure the translate suggestions checkbox remains true after the based on as changed as previously it only checked that it remained false which is the default state anyway.
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 33 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N6E4jUA6T109Wdq617G:-N6GCha--HOwj7oP3R16:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/68faf24b1bd9df4b394ea3ede8ac7d64b698e0c0/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L33)):*

Since this is not used outside this class, lets make it private
</blockquote></details>

Done.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1421)* status: 11 of 14 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.html` line 61 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N6K158Z6S8ET9MfrczG:-N6LKnB18Blrddm7QZMO:b-7wjinh) ([raw file](https://github.com/sillsdev/web-xforge/blob/0cc6ac1012598e26fe41da93f638cbbe5027c9bc/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.html#L61)):*

On my machine, I do not even see the disabled checkbox visually, even though it is in the dom. I think we can removed this checkbox entirely. Or maybe this is to make the text and alignment the same as the community checking option?
</blockquote></details>

I copied this from the settings page - I agree, it isn't needed but I wanted to keep it in line with what was done. I suspect it was done to aid with alignment issues. I actually think we need to come up with a more consistent framework for things like this when the layout is outside of the Angular Material framework. I'd leave it as part of that discussion and work.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 65 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N6K4Kyv4DoUMqj26JSx:-N6LO8VC7g6NTSL4i-Pc:b-v0y1ex) ([raw file](https://github.com/sillsdev/web-xforge/blob/0cc6ac1012598e26fe41da93f638cbbe5027c9bc/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L65)):*

Instead of putting these translations here, you can keep them in the original place. Then in the connect project component, you can create getters that read these translations from `settings.translate` or `settings.translate_tool_is_alwys_available` and then read them in the template like {{ translateToolHelperText }}.
What do you think of that?
</blockquote></details>

I've changed them to all use the same strings from the settings strings now - this does assume we don't want different instructions at all for the connect page.

RaymondLuong3 (approved):
This looks good to go.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-:-N6Oc2DiBwN_WZP1rtPk:basotln)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.html` line 61 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N6K158Z6S8ET9MfrczG:-N6ObasZC7y9JnTBuRsT:b2lpix6) ([raw file](https://github.com/sillsdev/web-xforge/blob/0cc6ac1012598e26fe41da93f638cbbe5027c9bc/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.html#L61)):*

I copied this from the settings page - I agree, it isn't needed but I wanted to keep it in line with what was done. I suspect it was done to aid with alignment issues. I actually think we need to come up with a more consistent framework for things like this when the layout is outside of the Angular Material framework. I'd leave it as part of that discussion and work.
</blockquote></details>

Ok I see. I didn't realize that it was a copy over from the settings page. That makes sense now.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 65 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1421#-N6K4Kyv4DoUMqj26JSx:-N6Obt36Ews4ul5lFZrl:b-bqjq15) ([raw file](https://github.com/sillsdev/web-xforge/blob/0cc6ac1012598e26fe41da93f638cbbe5027c9bc/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L65)):*

I've changed them to all use the same strings from the settings strings now - this does assume we don't want different instructions at all for the connect page.
</blockquote></details>

Nice. For the time being, I do not see a reason to have different translations on the connect project page.

nigel-wells (commented):
> Reviewed 5 of 9 files at r1, 1 of 3 files at r2, 5 of 8 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1421)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

Nateowami:
I think the wording needs to be reworked, especially on the connect project page, but I didn't want to hold up the change. That can be a separate task.

PR #1425, opened 2022-07-06
SF-1462b Ignore length of paragraph breaks that cannot contain verse

RaymondLuong3:
* refactor canContainVerseText into a static method
* check that a paragraph break insert should be included in verse text

This is how the note threads were being affected before and after this change.

Before
Poetry Line Break (Before)

After
Poetry Line Break

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1425)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/DeltaUsxMapper.cs
+++ b/src/SIL.XForge.Scripture/Services/DeltaUsxMapper.cs
@@ -101,6 +101,17 @@ namespace SIL.XForge.Scripture.Services
             ExceptionHandler = exceptionHandler;
         }
 
+        public static bool CanParaContainVerseText(string style)
+        {
+            // an empty style indicates an improperly formatted paragraph which could contain verse text
+            if (style == string.Empty)
+                return true;
+            if (char.IsDigit(style[style.Length - 1]))
+                style = style.Substring(0, style.Length - 1);
+            // paragraph, poetry, and list styles are the only types of valid paras that can contain verse text
+            return ParagraphPoetryListStyles.Contains(style);
+        }
+
         /// <summary>
         /// Create list of ChapterDelta objects from USX.
         ///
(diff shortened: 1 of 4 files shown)


RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1425)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs` line 909 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1425#-N7HmjFc8Ik_LGc0OTg4:-N7RNs5o8Iqb8PQO7vQW:b-pls7mh) ([raw file](https://github.com/sillsdev/web-xforge/blob/84e004c89699323a7fc0495849c72b8c6d764ebe/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L909)):*

I find it confusing the order of q and b. But I'm assuming you did it the right way. I notice, for example, that the `comment.ContextBefore` up above has `\b` before `\q`, whereas here q is before b in the order of inserts.
I should proably be thinking about this as telling me 
- Finish a paragraph of style q. 
- Finish a paragraph of style b.
Would that be accurate?

</blockquote></details>

Yep, that is the correct way to think of it.
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs` line 921 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1425#-N7HnQax-u1-JG7dy39U:-N7ROrqaAY-CbBq2CBco:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/84e004c89699323a7fc0495849c72b8c6d764ebe/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L921)):*

How about adding a comment above this line that says something like, "Include the length of the newline for q but not the length of the newline for b" ?
</blockquote></details>

Done.
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs` line 924 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1425#-N7HoDH3AQb05DZ9CBMS:-N7ROsOLBbjheGDR0AF8:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/84e004c89699323a7fc0495849c72b8c6d764ebe/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L924)):*

(And let's omit the consoles so the tests are quieter.)
</blockquote></details>

Done.

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1425)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1428, opened 2022-07-07
SF-1596 Enable passwordless login

nigel-wells:
- Pass additional auth config options to auth0 to enable passwordless login
- Use passwordless login by default when using a shared key

Default login/sign up screen shows an additional phone number option:
image

When choosing to login with phone number the user only sees an option to enter their phone number:
image

When using a shared link, and not currently logged in, passwordless is enabled and you see phone number login instead of email/password and PT is hidden:
image

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1428)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.spec.ts
@@ -1,5 +1,5 @@
 import { ActivatedRouteSnapshot, RouterStateSnapshot } from '@angular/router';
-import { anything, instance, mock, resetCalls, verify, when } from 'ts-mockito';
+import { anything, deepEqual, instance, mock, resetCalls, verify, when } from 'ts-mockito';
 import { AuthGuard } from './auth.guard';
 import { AuthService } from './auth.service';
 import { LocationService } from './location.service';
@@ -28,7 +28,7 @@ describe('AuthGuard', () => {
 
     canActivate$.subscribe(canActivate => {
       expect(canActivate).toBe(true);
-      verify(mockedAuthService.logIn(anything(), anything(), anything())).never();
+      verify(mockedAuthService.logIn({ returnUrl: anything() })).never();
       done();
     });
   });
@@ -47,7 +47,7 @@ describe('AuthGuard', () => {
 
     canActivate$.subscribe(canActivate => {
       expect(canActivate).toBe(false);
-      verify(mockedAuthService.logIn(anything(), false, undefined)).once();
+      verify(mockedAuthService.logIn(anything())).once();
       done();
     });
   });
@@ -58,6 +58,8 @@ describe('AuthGuard', () => {
     expect(authGuard).toBeDefined();
     when(mockedAuthService.isLoggedIn).thenResolve(false);
     when(mockedActivatedRouteSnapshot.queryParams).thenReturn({ sharing: 'true', 'sign-up': 'false' });
+    when(mockedLocationService.pathname).thenReturn('/');
+    when(mockedLocationService.search).thenReturn('');
 
     const canActivate$ = authGuard.canActivate(
       instance(mockedActivatedRouteSnapshot),
@@ -66,7 +68,16 @@ describe('AuthGuard', () => {
 
     canActivate$.subscribe(canActivate => {
       expect(canActivate).toBe(false);
-      verify(mockedAuthService.logIn(anything(), true, undefined)).once();
+      verify(
+        mockedAuthService.logIn(
+          deepEqual({
+            returnUrl: anything(),
+            signUp: true,
+            locale: anything(),
+            promptPasswordlessLogin: true
+          })
+        )
+      ).once();
       done();
     });
   });
@@ -85,7 +96,16 @@ describe('AuthGuard', () => {
 
     canActivate$.subscribe(canActivate => {
       expect(canActivate).toBe(false);
-      verify(mockedAuthService.logIn(anything(), true, undefined)).once();
+      verify(
+        mockedAuthService.logIn(
+          deepEqual({
+            returnUrl: anything(),
+            signUp: true,
+            locale: undefined,
+            promptPasswordlessLogin: false
+          })
+        )
+      ).once();
       done();
     });
   });
@@ -105,7 +125,11 @@ describe('AuthGuard', () => {
 
     canActivate$.subscribe(canActivate => {
       expect(canActivate).toBe(false);
-      verify(mockedAuthService.logIn(anything(), false, locale)).once();
+      verify(
+        mockedAuthService.logIn(
+          deepEqual({ returnUrl: anything(), signUp: false, locale, promptPasswordlessLogin: false })
+        )
+      ).once();
       done();
     });
   });
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.guard.ts
@@ -17,7 +17,12 @@ export class AuthGuard implements CanActivate {
         if (!isLoggedIn) {
           const signUp = route.queryParams['sharing'] === 'true' || route.queryParams['sign-up'] === 'true';
           const locale: string = route.queryParams['locale'];
-          this.authService.logIn(this.locationService.pathname + this.locationService.search, signUp, locale);
+          this.authService.logIn({
+            returnUrl: this.locationService.pathname + this.locationService.search,
+            signUp,
+            locale,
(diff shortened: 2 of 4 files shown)


RaymondLuong3 (changes requested):
I'll note here that the login page does not say Scripture Forge, and instead says Auth0. The testers saw this and another issue will be opened for it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1428#-:-N7STeR190Mbh_FLnfPk:b2g56zc)* status: all files reviewed, 3 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 189 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1428#-N7SUEej6yncLGkPQcHm:-N7SUEej6yncLGkPQcHn:bl6bnjq) ([raw file](https://github.com/sillsdev/web-xforge/blob/b5676991fb0d5f3e10ab34999d237ec1c84836cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L189)):*
> ```XML
>   }
> 
>   async logIn({ returnUrl, signUp, locale, promptPasswordlessLogin }: LoginParams = { returnUrl: '' }): Promise<void> {
> ```

When do we call `logIn()` without parameters? Is that still appropriate to not include a return url? I assume that means we are returning to the app start.component



_Code quote:_
```XML
 LoginParams = { returnUrl: '' })
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts` line 300 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1428#-N7SS8fVCj8CsAhqx7YR:-N7SS8fVCj8CsAhqx7YS:bwkl4s0) ([raw file](https://github.com/sillsdev/web-xforge/blob/b5676991fb0d5f3e10ab34999d237ec1c84836cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L300)):*
> ```XML
>     expect(authOptions).toBeDefined();
>     if (authOptions != null) {
>       expect(authOptions.enablePasswordless).toEqual(true);
> ```

Nathaniel and I had a conversation about instances in tests where we want to assert that something is not null and not undefined. We concluded that expect().toBeTruthy() is adequate to do this.
If we expect authOptions to be defined and not null, we could do
```
expect(authOptions).toBeTruthy();
expect(authOptions.enablePasswordless!).toEqual(true);
```
What do you think of that.



_Code quote:_
```XML
    if (authOptions != null) {
      expect(authOptions.enablePasswordless).toEqual(true);
      expect(authOptions.promptPasswordlessLogin).toEqual(false);
    }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts` line 316 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1428#-N7STZsO025V_aIBmNvU:-N7STZsO025V_aIBmNvV:b2nreh6) ([raw file](https://github.com/sillsdev/web-xforge/blob/b5676991fb0d5f3e10ab34999d237ec1c84836cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L316)):*
> ```XML
>     ).last()[0];
>     expect(authOptions).toBeDefined();
>     if (authOptions != null) {
> ```

Same as above.



_Code quote:_
```XML
 if (authOptions != null) {
```

nigel-wells (commented):
Thanks - this is fixed now.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1428#-:-N7n5oKD6GjAoN-ufXcJ:bwfn43h)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 189 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1428#-N7SUEej6yncLGkPQcHm:-N7n4GZTEjXHhJTKObT8:beq8uds) ([raw file](https://github.com/sillsdev/web-xforge/blob/b5676991fb0d5f3e10ab34999d237ec1c84836cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L189)):*

When do we call `logIn()` without parameters? Is that still appropriate to not include a return url? I assume that means we are returning to the app start.component
</blockquote></details>

It's more so we can create some kind of state to pass along to auth0. We also expect something to be available to check on the callback from auth0. I see one of interfaces said this could be `undefined` so I've removed that to ensure it is always defined. FYI there is a validation check in the callback that will send users to `/projects` if it fails for whatever reason.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts` line 300 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1428#-N7SS8fVCj8CsAhqx7YR:-N7n5AWK7pKWNf1qj3bX:bv2w07h) ([raw file](https://github.com/sillsdev/web-xforge/blob/b5676991fb0d5f3e10ab34999d237ec1c84836cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L300)):*

Nathaniel and I had a conversation about instances in tests where we want to assert that something is not null and not undefined. We concluded that expect().toBeTruthy() is adequate to do this.
If we expect authOptions to be defined and not null, we could do
```
expect(authOptions).toBeTruthy();
expect(authOptions.enablePasswordless!).toEqual(true);
```
What do you think of that.
</blockquote></details>

Yes, makes sense. I suspect these were written this way in the past (I copied and pasted the format for consistency across all tests) to make the IDE happy. If you remove the specific `null` check then the IDE complains about the proceeding statements saying that `authOptions` may actually be `null`. We could add in some `!` to make it happy. What are your thoughts? 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts` line 316 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1428#-N7STZsO025V_aIBmNvU:-N7n5lfjD7Z9fAF6bDk8:bqi7dbg) ([raw file](https://github.com/sillsdev/web-xforge/blob/b5676991fb0d5f3e10ab34999d237ec1c84836cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L316)):*

Same as above.
</blockquote></details>

As per previous comment.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1428)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 189 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1428#-N7SUEej6yncLGkPQcHm:-N7qURty-qUuiR80Fh2F:b-tiwng7) ([raw file](https://github.com/sillsdev/web-xforge/blob/b5676991fb0d5f3e10ab34999d237ec1c84836cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L189)):*

It's more so we can create some kind of state to pass along to auth0. We also expect something to be available to check on the callback from auth0. I see one of interfaces said this could be `undefined` so I've removed that to ensure it is always defined. FYI there is a validation check in the callback that will send users to `/projects` if it fails for whatever reason.
</blockquote></details>

Ok, that's good to know.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts` line 300 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1428#-N7SS8fVCj8CsAhqx7YR:-N7qUrDS7cb8kvRa2NXZ:b-enni0s) ([raw file](https://github.com/sillsdev/web-xforge/blob/b5676991fb0d5f3e10ab34999d237ec1c84836cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts#L300)):*

Yes, makes sense. I suspect these were written this way in the past (I copied and pasted the format for consistency across all tests) to make the IDE happy. If you remove the specific `null` check then the IDE complains about the proceeding statements saying that `authOptions` may actually be `null`. We could add in some `!` to make it happy. What are your thoughts? 
</blockquote></details>

Hmm, that's a good thought that this keeps it consistent with the other tests in the spec. I think I am happy to keep it consistent since the benefit is so minor.

Nateowami:
We'll wait until Twilio is ready before merging this, since once it's on master it's in the queue to go to qa and live.

marksvc:
This is working great in local testing on master. 
I think we should change the text "sign in" to "log in".

PR #1433, opened 2022-07-11
Renew PT access tokens 2 minutes before expiration

Nateowami:
PT access tokens are valid for 20 minutes. Previously we would check if they were going to expire in the next 1 minute and renew if they were. This changes it to 2 minutes, since the registry server's clock has drifted in the past and resulted in errors.

This doesn't fully fix the problem, but I think it's a reasonable way to permanently reduce the frequency with which we run into this problem.


RaymondLuong3 (approved):
I wish we didn't have to do this...

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1433#-:-N6nYqbx9r26b7d6N7tL:bhvjyk7)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1440, opened 2022-07-18
SF-1694 Enclose duplicate note removal logic in text view model

RaymondLuong3:
* explicitly record duplicate notes in text view model
* remove duplicate note thread icons earlier
* set icon removal source to api so it is not saved in history

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1440)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -190,7 +190,7 @@ export class TextViewModel {
   }
 
   private get embedPositions(): number[] {
-    return this.embeddedElementPositions(this._embeddedElements);
+    return this.embeddedElementPositions(Array.from(this._embeddedElements.values()));
   }
 
   /** Associate the existing editor to a (single) specific textdoc. */
@@ -687,6 +687,11 @@ export class TextViewModel {
             embedPosition = { position };
             this._embeddedElements.set(id, embedPosition);
           } else {
+            if (embedPosition.duplicatePosition != null) {
+              console.warn(
+                'Warning: text-view-model.updateSegments() did not expect to encounter an embed with >2 positions'
+              );
+            }
             embedPosition.duplicatePosition = position;
           }
           curSegment.notesCount++;
@@ -746,7 +751,7 @@ export class TextViewModel {
     return [fixDelta, fixOffset];
   }
 
-  private embeddedElementPositions(embeds: Map<string, EmbedPosition> | EmbedPosition[]): number[] {
+  private embeddedElementPositions(embeds: EmbedPosition[]): number[] {
     let result: number[] = [];
     for (const ep of embeds.values()) {
       result.push(ep.position);
@@ -873,24 +878,19 @@ export class TextViewModel {
   }
 
   /** Gets the number of embeds in a given range displayed in the quill editor. */
-  private getEmbedsInEditorRange(startIndex: number, length: number): number {
-    const embedPositions: IterableIterator<EmbedPosition> = this._embeddedElements.values();
-    const opEndIndex: number = startIndex + length;
+  private getEmbedsInEditorRange(startPos: number, length: number): number {
+    const embedPositions = this.embedPositions;
+    const opEndIndex: number = startPos + length;
     let embeddedElementsCount: number = 0;
-    for (const pos of embedPositions) {
-      if (pos.position < startIndex && (pos.duplicatePosition == null || pos.duplicatePosition < startIndex)) {
+    for (const embedPos of embedPositions) {
+      if (embedPos < startPos) {
         continue;
-      } else if (pos.position >= startIndex && pos.position < opEndIndex) {
-        embeddedElementsCount++;
-      } else if (
-        pos.duplicatePosition != null &&
-        pos.duplicatePosition >= startIndex &&
-        pos.duplicatePosition < opEndIndex
-      ) {
-        embeddedElementsCount++;
-      } else {
-        break;
       }
+      if (embedPos >= startPos && embedPos < opEndIndex) {
+        embeddedElementsCount++;
+        continue;
+      }
+      break;
     }
     return embeddedElementsCount;
   }


marksvc (changes requested):
Before you process the other comments, I suggest to first look at the comment starting off "What do you think about the following?", since it considers a change with broader impact.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-:-N8A2w4B9p3bXgQJZY9v:b-uawp4z)* status: all files reviewed, 4 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 171 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N89piD48aIbHNfK4Rl9:-N89piD48aIbHNfK4RlA:b-oyv39i) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L171)):*
> ```XML
>     const embeddedElements: Map<string, number> = new Map<string, number>();
>     for (const [embedId, embedPosition] of this._embeddedElements.entries()) {
>       embeddedElements.set(embedId, embedPosition.position);
> ```

I am glad we are just returning the .position, not also the location of the duplicate embed.



_Code quote:_
```XML
n.position);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 173 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N89oVoy9Eu-fKWRjL_y:-N89oVoy9Eu-fKWRjL_z:bhpdro5) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L173)):*
> ```XML
>       embeddedElements.set(embedId, embedPosition.position);
>     }
>     return embeddedElements;
> ```

It seems like we ought to be able to do this by mapping/transforming, but in playing with it for a bit I didn't quickly see how we would do that :)



_Code quote:_
```XML
  get embeddedElements(): Map<string, number> {
    const embeddedElements: Map<string, number> = new Map<string, number>();
    for (const [embedId, embedPosition] of this._embeddedElements.entries()) {
      embeddedElements.set(embedId, embedPosition.position);
    }
    return embeddedElements;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 192 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N89w5PH9-Fh7ovX8glR:-N89w5PH9-Fh7ovX8glS:b6y3dvj) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L192)):*
> ```XML
>   }
> 
>   private get embedPositions(): number[] {
> ```

I think I may start to get confused about which methods return positions or information that is accounting for duplicate positions, and which aren't :)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 686 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N8A1Q7Y5yUpF872iM1X:-N8A1Q7Y5yUpF872iM1Y:b-waxjyh) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L686)):*
> ```XML
>             embedPosition = { position };
>             this._embeddedElements.set(id, embedPosition);
>           } else {
> ```

What do you think about the following? 

The proposed change has 
- updateSegments() detect and record the presence of duplicate embeds, 
- prepares various methods in text-view-model to account for and return correct information if there are duplicate embeds, and 
- adjusts update() so that it cleans up the existence of duplicate embeds.

Instead of this, what do you think about modifying the updateSegments() method so that when it detects a duplicate embed, it adds some ops to `fixDelta` that will delete the duplicate embed. Then when update() applies the result of updateSegments(), we clean up the duplicate embeds. This might result in not requiring the various methods in text-view-model.ts to account for and return correct information in the face of the existence of duplicate embeds. And we may not even need an `EmbedPosition` type. 
This approach may have trouble if during the `Promise.resolve().then` delay, text-view-model needs to do various processing that needs to account for duplicate embeds existing, before the Promise is processed/resolved. 



_Code quote:_
```XML
  } else {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 687 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N89sQyn2whT95iotrx2:-N89sQyn2whT95iotrx3:b-4yjj3q) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L687)):*
> ```XML
>             this._embeddedElements.set(id, embedPosition);
>           } else {
>             embedPosition.duplicatePosition = position;
> ```

What do you think about having something like
```ts
console.warn(`Warning: text-view-model updateSegments did not expect to encounter an embed with >2 positions.`);
```
The code as is will silently ignore the situation. And maybe it will never happen. 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 746 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N89tMZPAzleEMbzZUJz:-N89tMZPAzleEMbzZUK-:b-nkxe98) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L746)):*
> ```XML
>   }
> 
>   private embeddedElementPositions(embeds: Map<string, EmbedPosition> | EmbedPosition[]): number[] {
> ```

Seeing this type makes me wonder if we would benefit from adding the embed id to the type, like 
```ts
interface EmbedInfo {
  embedId: string;  
  position: number;
  duplicatePosition?: number;
}
```
I haven't played with it to know if that would result in messiness.



_Code quote:_
```XML
(embeds: Map<string, EmbedPosition> | EmbedPosition[]
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 763 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N89u7oSAeAUnOldT6v9:-N89u7oSAeAUnOldT6vA:b-owrddr) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L763)):*
> ```XML
>     );
> 
>     const deletePositions: number[] = this.embeddedElementPositions(duplicatePositions).sort((a, b) => a - b);
> ```

I'm concerned about this sort order. Won't the wrong thing happen if we delete from the start, kind of like if you modify an array while processing it?

For example, if I have data 
```
A,B,C,D,E,F,G
```
and I want to delete items B, D, and F, and so I want to delete items in (zero-indexed) positions 1, 3, and 5, I can't do this:
```ts
for pos_to_delete in [1, 3, 5] {
  my_array = my_array.delete_item_at_position(pos_to_delete)
}
```
because although the first iteration will work, removing B from position 1, the data is now 
```
A,C,D,E,F,G
```
and so D is no longer in position 3 but in position 2. 
But if I instead do:
```ts
for pos_to_delete in reverse_this_array([1, 3, 5]) {
  my_array = my_array.delete_item_at_position(pos_to_delete)
}
```
then I delete the item in position 5, so the data becomes 
```
A,B,C,D,E,G
```
then I delete the item in position 3, so the data becomes
```
A,B,C,E,G
```
then 1,
```
A,C,E,G
```

So if the delta and related code works in a similar way, then we will want to process the positions from latest to earliest. Or maybe something in the code is taking account of things so it still works anyway?



_Code quote:_
```XML
sort((a, b) => a - b);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 877 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N89v5_hDdGGDElzYkuC:-N89v5_hDdGGDElzYkuD:b-3d8hcq) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L877)):*
> ```XML
>     const opEndIndex: number = startIndex + length;
>     let embeddedElementsCount: number = 0;
>     for (const pos of embedPositions) {
> ```

Hmm. Won't this return a count that doesn't include hits of a duplicatePosition being within range, when the non-duplicatePosition is also in range? In other words, if range is 0 to 10, and we have an item with position 5 and duplicatePosition 7, we want this method to return 2, not 1, right? But the code as written will only look at the EmbedPosition item once, count the instance of position==5, not consider that duplicatePosition==7, and move on to the next item in the EmbedPosition list, thus skipping counting the duplicatePosition==7 hit. 

Probably we want to use `const embedPositions=this.embedPositions()` and not worry about processing specifically duplicate-positions here, and then this method will return the count of all embeds in the range. 

Or maybe you specifically don't want it to include duplicate positions in the range?



_Code quote:_
```XML
    for (const pos of embedPositions) {
      if (pos.position < startIndex && (pos.duplicatePosition == null || pos.duplicatePosition < startIndex)) {
        continue;
      } else if (pos.position >= startIndex && pos.position < opEndIndex) {
        embeddedElementsCount++;
      } else if (
        pos.duplicatePosition != null &&
        pos.duplicatePosition >= startIndex &&
        pos.duplicatePosition < opEndIndex
      ) {
        embeddedElementsCount++;
      } else {
        break;
      }
```

RaymondLuong3 (commented):
Thanks for the suggestion. I added my comment below.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-:-N8_Me-j9yhCFeVddxBT:b-btafsj)* status: all files reviewed, 4 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 171 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N89piD48aIbHNfK4Rl9:-N8_Mj1O9A9dpPkDdbZK:b-z8lhi7) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L171)):*

I am glad we are just returning the .position, not also the location of the duplicate embed.
</blockquote></details>

I hope it stays this way!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 173 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N89oVoy9Eu-fKWRjL_y:-N8_Mt1P8-WWclUCg_Dh:b-dq6ens) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L173)):*

It seems like we ought to be able to do this by mapping/transforming, but in playing with it for a bit I didn't quickly see how we would do that :)
</blockquote></details>

Yea, I couldn't think of something better.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 192 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N89w5PH9-Fh7ovX8glR:-N8_N3fX26AHOSKCtRzi:buu86rz) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L192)):*

I think I may start to get confused about which methods return positions or information that is accounting for duplicate positions, and which aren't :)
</blockquote></details>

The return type does helps to differentiate that.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 686 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N8A1Q7Y5yUpF872iM1X:-N8_LtF_3SYOYgVWaIye:b48p4dt) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L686)):*

What do you think about the following? 

The proposed change has 
- updateSegments() detect and record the presence of duplicate embeds, 
- prepares various methods in text-view-model to account for and return correct information if there are duplicate embeds, and 
- adjusts update() so that it cleans up the existence of duplicate embeds.

Instead of this, what do you think about modifying the updateSegments() method so that when it detects a duplicate embed, it adds some ops to `fixDelta` that will delete the duplicate embed. Then when update() applies the result of updateSegments(), we clean up the duplicate embeds. This might result in not requiring the various methods in text-view-model.ts to account for and return correct information in the face of the existence of duplicate embeds. And we may not even need an `EmbedPosition` type. 
This approach may have trouble if during the `Promise.resolve().then` delay, text-view-model needs to do various processing that needs to account for duplicate embeds existing, before the Promise is processed/resolved. 
</blockquote></details>

I did try to go down that route originally. Then I realized that it has to be slit apart primarily because `this.updateSegments()` produces a fixDelta that corrects blanks which is considered a user operation. But removing duplicate ops is considered an api operation. This allows the undo stack to not account for ops submitted that remove duplicate ops.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 687 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N89sQyn2whT95iotrx2:-N8_NlzC5aBhzb5cmnjI:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L687)):*

What do you think about having something like
```ts
console.warn(`Warning: text-view-model updateSegments did not expect to encounter an embed with >2 positions.`);
```
The code as is will silently ignore the situation. And maybe it will never happen. 

</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 746 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N89tMZPAzleEMbzZUJz:-N8_OFnA1ZnRJXWq7OMP:b-ogx2ey) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L746)):*

Seeing this type makes me wonder if we would benefit from adding the embed id to the type, like 
```ts
interface EmbedInfo {
  embedId: string;  
  position: number;
  duplicatePosition?: number;
}
```
I haven't played with it to know if that would result in messiness.
</blockquote></details>

Actually, the `embedId` is not even important for this method. So I reduced the method signature to only accept `EmbedPosition[]` type
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 763 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N89u7oSAeAUnOldT6v9:-N8_VkwM8hmK5z80lMqh:b2e1hqr) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L763)):*

I'm concerned about this sort order. Won't the wrong thing happen if we delete from the start, kind of like if you modify an array while processing it?

For example, if I have data 
```
A,B,C,D,E,F,G
```
and I want to delete items B, D, and F, and so I want to delete items in (zero-indexed) positions 1, 3, and 5, I can't do this:
```ts
for pos_to_delete in [1, 3, 5] {
  my_array = my_array.delete_item_at_position(pos_to_delete)
}
```
because although the first iteration will work, removing B from position 1, the data is now 
```
A,C,D,E,F,G
```
and so D is no longer in position 3 but in position 2. 
But if I instead do:
```ts
for pos_to_delete in reverse_this_array([1, 3, 5]) {
  my_array = my_array.delete_item_at_position(pos_to_delete)
}
```
then I delete the item in position 5, so the data becomes 
```
A,B,C,D,E,G
```
then I delete the item in position 3, so the data becomes
```
A,B,C,E,G
```
then 1,
```
A,C,E,G
```

So if the delta and related code works in a similar way, then we will want to process the positions from latest to earliest. Or maybe something in the code is taking account of things so it still works anyway?
</blockquote></details>

Actually, when we compose the delta, starting from the first embed is the right thing (in my testing). What you described would be the case if we were applying the delete operation to the data one delta at a time. However, this explicitly composes the delta so that all the delete ops exist in the single delta which we apply all at once.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 877 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N89v5_hDdGGDElzYkuC:-N8_cW6n4QvoVOc2aUgt:bw4xr3n) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L877)):*

Hmm. Won't this return a count that doesn't include hits of a duplicatePosition being within range, when the non-duplicatePosition is also in range? In other words, if range is 0 to 10, and we have an item with position 5 and duplicatePosition 7, we want this method to return 2, not 1, right? But the code as written will only look at the EmbedPosition item once, count the instance of position==5, not consider that duplicatePosition==7, and move on to the next item in the EmbedPosition list, thus skipping counting the duplicatePosition==7 hit. 

Probably we want to use `const embedPositions=this.embedPositions()` and not worry about processing specifically duplicate-positions here, and then this method will return the count of all embeds in the range. 

Or maybe you specifically don't want it to include duplicate positions in the range?

</blockquote></details>

That's right. Good catch. I did another trace of the code, and it looks like at this stage we do not expect duplicate positions to be present since it runs before updateSegments(). I've cleaned it up.

marksvc (changes requested):
I'm wondering if we should be trying to be more strict about what can be cut, deleted, or dragged, so that we don't allow an embed to be manipulated in such a way that would/could result in duplicate embeds later coming in (such as later thru an undo). But I would need to look more closely at what's happening.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-:-N9caspp9yfnqlCiVqWo:bqut4zu)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @marksvc and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 686 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N8A1Q7Y5yUpF872iM1X:-N9b_2_dAqO7VbbFXNth:b-h417ti) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L686)):*

I did try to go down that route originally. Then I realized that it has to be slit apart primarily because `this.updateSegments()` produces a fixDelta that corrects blanks which is considered a user operation. But removing duplicate ops is considered an api operation. This allows the undo stack to not account for ops submitted that remove duplicate ops.
</blockquote></details>

Hmm, okay. I think text-view-model.ts can be slimmer tho. What do you think about modifying updateSegments() to return both a user delta and an api delta, which update() can then apply with the appropriate sources specified? This addition to updateSegments() and update() would allow other parts of text-view-model.ts to not need to become knowledgeable about or account for duplicate embed positions. I have a change that illustrates this in branch task/delete-multiple-notes-idea.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 763 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N89u7oSAeAUnOldT6v9:-N9bdyu9EKzKRIHRllZO:b-9755fa) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L763)):*

Actually, when we compose the delta, starting from the first embed is the right thing (in my testing). What you described would be the case if we were applying the delete operation to the data one delta at a time. However, this explicitly composes the delta so that all the delete ops exist in the single delta which we apply all at once.
</blockquote></details>

Thank you.

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1440)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts` line 686 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1440#-N8A1Q7Y5yUpF872iM1X:-N9lum2A6zamu8zuIYOk:b-2g7we2) ([raw file](https://github.com/sillsdev/web-xforge/blob/ca45b7d9a74116b82692761dcfb8f1130de399a1/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L686)):*

Hmm, okay. I think text-view-model.ts can be slimmer tho. What do you think about modifying updateSegments() to return both a user delta and an api delta, which update() can then apply with the appropriate sources specified? This addition to updateSegments() and update() would allow other parts of text-view-model.ts to not need to become knowledgeable about or account for duplicate embed positions. I have a change that illustrates this in branch task/delete-multiple-notes-idea.
</blockquote></details>

We discussed this. Having two deltas and sending both to editor.updateContents() can result in update() being re-entered between the calls, and so the second call to editor.updateContents(), with a delta to remove duplicate embeds, may have out of date position information by the time it is run. We looked at the quilljs API but it did not appear to have a "pause and don't respond to updates for a moment" setting, during which we could apply both deltas.

PR #1442, opened 2022-07-20
SF-1653 Update text anchor when multiple edits in a delta

RaymondLuong3:
This change enables the ability to update a text anchor when multiple edit positions exist in a delta. Currently, this only happens when a user edits text via a drag and drop, and then undoes the drag and drop. Previously, this would cause the text anchor to be updated incorrectly because we only handled the first edit (insert or delete) that we saw in the delta.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1442)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/segment.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/segment.ts
@@ -8,8 +8,6 @@ export class Segment {
   private _range: RangeStatic = { index: 0, length: 0 };
   private _checksum?: number;
   private initialTextLen: number = -1;
-  /** Mapping of embed ids to their position in the editor. Enables processing changes to embeds after text edits. */
-  private _segmentVerseEmbeddedElements: Map<string, number> = new Map<string, number>();
 
   constructor(public readonly bookNum: number, public readonly chapter: number, public readonly ref: string) {}
 
@@ -36,19 +34,14 @@ export class Segment {
     return this._text.length - this.initialTextLen;
   }
 
-  get embeddedElements(): Map<string, number> {
-    return this._segmentVerseEmbeddedElements;
-  }
-
   acceptChanges(): void {
     this.initialTextLen = this._text.length;
     this.initialChecksum = this.checksum;
   }
 
-  update(text: string, range: RangeStatic, embeddedElementIndices: Map<string, number>): void {
+  update(text: string, range: RangeStatic): void {
     this._text = text;
     this._range = range;
-    this._segmentVerseEmbeddedElements = embeddedElementIndices;
     this._checksum = undefined;
     if (this.initialTextLen === -1) {
       this.initialTextLen = text.length;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -165,12 +165,20 @@ export class TextViewModel {
     return this._segments.entries();
   }
 
+  get segmentsSnapshot(): IterableIterator<[string, RangeStatic]> {
+    return cloneDeep(this._segments).entries();
+  }
+
   get embeddedElements(): Map<string, number> {
     const embeddedElements: Map<string, number> = new Map<string, number>();
     for (const [embedId, embedPosition] of this._embeddedElements.entries()) {
       embeddedElements.set(embedId, embedPosition.position);
     }
     return embeddedElements;
+  }
+
+  get embeddedElementsSnapshot(): Map<string, number> {
+    return cloneDeep(this.embeddedElements);
   }
 
   get isLoaded(): boolean {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -1032,9 +1032,11 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     let embedsByVerse = new Map<string, number>();
     const editPositions: number[] = this.getEditPositionsInDelta(delta);
     const embedsByEditedVerse: EmbedsByVerse[] = [];
+    // content before verse 1 are considered to be in verse 0
+    let baseVerse: string = '0';
     for (const [segment, range] of preDeltaSegmentCache) {
-      // TODO: if a section heading, continue in previous verse
-      const baseVerse: string = this.getBaseVerse(segment);
+      // if we cannot determine the base verse, consider it as part of the previous verse
+      baseVerse = this.getBaseVerse(segment) ?? baseVerse;
       if (currentVerse === '') {
         // set the current verse and range on the first pass
         currentVerse = baseVerse;
@@ -1096,9 +1098,13 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     return editPositions;
   }
 
-  private getBaseVerse(segmentRef: string): string {
+  /**
+   * Returns the base verse of the segment ref. e.g. 'verse_1_5'
+   * @return The segment ref of the first segment in the verse, or undefined if the segment does not belong to a verse.
+   */
+  private getBaseVerse(segmentRef: string): string | undefined {
     const matchArray: RegExpExecArray | null = VERSE_REGEX.exec(segmentRef);
-    return matchArray == null ? segmentRef : matchArray[0];
+    return matchArray == null ? undefined : matchArray[0];
   }
 
   private getVerseEndIndex(baseRef: string): number | undefined {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -606,14 +606,14 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
 
   /** Respond to text changes in the quill editor. */
   onContentChanged(delta: DeltaStatic, source: string): void {
-    const segmentsBeforeDelta: IterableIterator<[string, RangeStatic]> = this.viewModel.segmentsSnapshot;
-    const embedsBeforeDelta: Readonly<Map<string, number>> = this.viewModel.embeddedElementsSnapshot;
+    const preDeltaSegmentCache: IterableIterator<[string, RangeStatic]> = this.viewModel.segmentsSnapshot;
+    const preDeltaEmbedCache: Readonly<Map<string, number>> = this.viewModel.embeddedElementsSnapshot;
     this.viewModel.update(delta, source as Sources);
     this.updatePlaceholderText();
     // skip updating when only formatting changes occurred
     if (delta.ops != null && delta.ops.some(op => op.insert != null || op.delete != null)) {
       const isUserEdit: boolean = source === 'user';
-      this.update(delta, segmentsBeforeDelta, embedsBeforeDelta, isUserEdit);
+      this.update(delta, preDeltaSegmentCache, preDeltaEmbedCache, isUserEdit);
     }
   }
 
@@ -860,8 +860,8 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
 
   private update(
     delta?: DeltaStatic,
-    segmentsBeforeDelta?: IterableIterator<[string, RangeStatic]>,
-    embedsBeforeDelta?: Readonly<Map<string, number>>,
+    preDeltaSegmentCache?: IterableIterator<[string, RangeStatic]>,
+    preDeltaEmbedCache?: Readonly<Map<string, number>>,
     isUserEdit?: boolean
   ): void {
     let segmentRef: string | undefined;
@@ -921,7 +921,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
 
     Promise.resolve().then(() => this.adjustSelection());
     const affectedEmbeds: EmbedsByVerse[] =
-      isUserEdit === true ? this.getEmbedsAffectedByDelta(delta, segmentsBeforeDelta, embedsBeforeDelta) : [];
+      isUserEdit === true ? this.getEmbedsAffectedByDelta(delta, preDeltaSegmentCache, preDeltaEmbedCache) : [];
     this.updated.emit({
       delta,
       prevSegment,
@@ -1020,19 +1020,20 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
   /** Gets the embeds affected */
   private getEmbedsAffectedByDelta(
     delta?: DeltaStatic,
-    segmentsBeforeDelta?: IterableIterator<[string, RangeStatic]>,
-    embedsBeforeDelta?: Readonly<Map<string, number>>
+    preDeltaSegmentCache?: IterableIterator<[string, RangeStatic]>,
+    preDeltaEmbedCache?: Readonly<Map<string, number>>
   ): EmbedsByVerse[] {
-    if (delta?.ops == null || segmentsBeforeDelta == null || embedsBeforeDelta == null) {
+    if (delta?.ops == null || preDeltaSegmentCache == null || preDeltaEmbedCache == null) {
       return [];
     }
     let verseIsEdited = false;
     let currentVerse: string = '';
     let currentVerseRange: RangeStatic = { index: 0, length: 0 };
     let embedsByVerse = new Map<string, number>();
-    const editPositions: number[] = this.getEditPositionInDelta(delta);
+    const editPositions: number[] = this.getEditPositionsInDelta(delta);
     const embedsByEditedVerse: EmbedsByVerse[] = [];
-    for (const [segment, range] of segmentsBeforeDelta) {
+    for (const [segment, range] of preDeltaSegmentCache) {
+      // TODO: if a section heading, continue in previous verse
       const baseVerse: string = this.getBaseVerse(segment);
       if (currentVerse === '') {
         // set the current verse and range on the first pass
@@ -1040,7 +1041,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
         currentVerseRange = range;
       }
       if (currentVerse !== baseVerse) {
-        // this segment belongs to a new verse, add the embeds if the  previous verse has been edited
+        // this segment belongs to a new verse, add the embeds if the previous verse has been edited
         if (verseIsEdited) {
           embedsByEditedVerse.push({ embeds: embedsByVerse, verseRange: currentVerseRange });
         }
@@ -1049,7 +1050,8 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
         currentVerseRange = range;
         verseIsEdited = false;
       } else {
-        currentVerseRange.length += range.length;
+        const lengthFromVerseStart: number = range.index + range.length - currentVerseRange.index;
+        currentVerseRange.length = lengthFromVerseStart;
       }
 
       const editedPositionsWithinRange: number[] = editPositions.filter(
@@ -1059,7 +1061,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
         verseIsEdited = true;
       }
 
-      for (const [embedId, embedPosition] of embedsBeforeDelta.entries()) {
+      for (const [embedId, embedPosition] of preDeltaEmbedCache.entries()) {
         if (embedPosition >= range.index && embedPosition < range.index + range.length) {
           embedsByVerse.set(embedId, embedPosition);
         }
@@ -1073,7 +1075,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     return embedsByEditedVerse;
   }
 
-  private getEditPositionInDelta(delta: DeltaStatic): number[] {
+  private getEditPositionsInDelta(delta: DeltaStatic): number[] {
     let curIndex = 0;
     const editPositions: number[] = [];
     if (delta.ops == null) {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -997,24 +997,13 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
       return;
     }
 
-    const matchArray: RegExpExecArray | null = VERSE_REGEX.exec(this._segment.ref);
-    const baseVerse = matchArray == null ? this._segment.ref : matchArray[0];
-    const startSegmentRange: RangeStatic | undefined = this.viewModel.getSegmentRange(baseVerse);
     const segmentRange: RangeStatic | undefined = this.viewModel.getSegmentRange(this._segment.ref);
     if (segmentRange == null) {
       return;
     }
 
-    const endIndex: number = this.getVerseEndIndex(baseVerse) ?? segmentRange.index + segmentRange.length;
-    const startIndex: number = startSegmentRange?.index ?? segmentRange.index;
     const text = this.viewModel.getSegmentText(this._segment.ref);
-    const verseEmbeddedElements: Map<string, number> = new Map<string, number>();
-    for (const [threadId, index] of this.embeddedElements.entries()) {
-      if (index >= startIndex && index < endIndex) {
-        verseEmbeddedElements.set(threadId, index);
-      }
-    }
-    this._segment.update(text, segmentRange, verseEmbeddedElements);
+    this._segment.update(text, segmentRange);
   }
 
   /** Gets the embeds affected */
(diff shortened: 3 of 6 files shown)


marksvc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1442)* status: all files reviewed, 11 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1001 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7rZ3Nf1DPpkWzpSr8x:-N7rZ3Nf1DPpkWzpSr8y:b-q5cnxo) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1001)):*
> ```XML
>     delta?: DeltaStatic,
>     segmentsBeforeDelta?: IterableIterator<[string, RangeStatic]>,
>     embedsBeforeDelta?: Readonly<Map<string, number>>
> ```

To avoid any misunderstanding from the ambiguity of "before" meaning in time or position, let's use something like
```
segmentsTemporallyBeforeDelta
embedsTemporallyBeforeDelta
```
I'm not very happy with those suggested names, but I'm not sure what's best to use. 



_Code quote:_
```XML
    segmentsBeforeDelta?: IterableIterator<[string, RangeStatic]>,
    embedsBeforeDelta?: Readonly<Map<string, number>>
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1017 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7rOh8c2TEVzANgHjGS:-N7rOh8c2TEVzANgHjGT:bmzyxu5) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1017)):*
> ```XML
>         // set the current verse and range on the first pass
>         currentVerse = baseVerse;
>         currentVerseRange = range;
> ```

Oh no, won't this be storing the range of a `segment` and so conceivably not of the whole verse? 

Same for a dozen lines later, with `currentVerseRange = range;`



_Code quote:_
```XML
 currentVerseRange = range;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1029 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7rO9gp0PKVKFDznOxj:-N7rO9gp0PKVKFDznOxk:bh32fn0) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1029)):*
> ```XML
>         verseIsEdited = false;
>       } else {
>         currentVerseRange.length += range.length;
> ```

Oh, didn't we maybe not want to do this on the first iteration of the loop, since a dozen lines above, we do `currentVerseRange = range;`? Maybe if we omit the `currentVerseRange = range;` above, then we can safely do this line here without incrementing currentVerseRange.length more than desired?



_Code quote:_
```XML
      } else {
        currentVerseRange.length += range.length;
      }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1034 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7rPO0n9NZzAZ_d0As_:-N7rPO0n9NZzAZ_d0Asa:baebauz) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1034)):*
> ```XML
>       const editedPositionsWithinRange: number[] = editPositions.filter(
>         pos => pos >= range.index && pos <= range.index + range.length
>       );
> ```

I know that we have handling sometimes for when an edit occurs at the end of a range, where we include it in that range, and so we  would use `<=` in `pos <= range.index + range.length` rather than the `<` operator. But it occurs to me that by doing this, we potentially will find a position "in" two different segments, if the segments are adjacent. I don't know yet if there will be a problem about it, but it's something I was just thinking.



_Code quote:_
```XML
      const editedPositionsWithinRange: number[] = editPositions.filter(
        pos => pos >= range.index && pos <= range.index + range.length
      );
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1055 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7rTjm_0HTFhsXKLgeB:-N7rTjm_0HTFhsXKLgeC:b7vc0o8) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1055)):*
> ```XML
>   private getEditPositionInDelta(delta: DeltaStatic): number[] {
>     let curIndex = 0;
>     const editPositions: number[] = [];
> ```

How about `s` in `getEditPositionsInDelta`?



_Code quote:_
```XML
getEditPositionInDelta
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1076 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7rUkME3t4HsnHwwEcn:-N7rUkMF0C0YkvNhYExc:b-m14ge1) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1076)):*
> ```XML
>   private getBaseVerse(segmentRef: string): string {
>     const matchArray: RegExpExecArray | null = VERSE_REGEX.exec(segmentRef);
>     return matchArray == null ? segmentRef : matchArray[0];
> ```

What's the thinking  behind returning `segmentRef` if there was no match? 



_Code quote:_
```XML
null ? segmentRef
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1157 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7qomI62-fapDBnT_fc:-N7qomI62-fapDBnT_fd:bmln88f) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1157)):*
> ```XML
>     const textDeleteOps: DeltaOperation[] = delta.filter(ops => ops.delete != null);
>     const hasTextEditOp: boolean = textInsertOps.length > 0 || textDeleteOps.length > 0;
>     for (const affected of affectedEmbeds) {
> ```

So there is now an outer loop that runs the inner loop over multiple sets of embeds, that may correspond to different verse ranges. And the `editPosition` will be one within the given range, unlike perhaps `editOpIndex` above.

Are we limiting to within specific ranges for performance, or because it would be incorrect behaviour to process outside of those ranges? It looks like from `getAnchorChanges()` that the verse range is being used to only  consider and apply some insert and delete information when we are in the range. I'm unsure so far if we would be getting to getAnchorChanges() and working when we are _out_ of the range, tho. 



_Code quote:_
```XML
for (const affected of affectedEmbeds) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1176 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7qnes4EwTs0xudaQMf:-N7qnes4EwTs0xudaQMg:brfxozr) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1176)):*
> ```XML
>         );
>         const noteAnchorEndIndex: number = notePos + noteThreadDoc.data.position.length + noteCountInTextAnchor;
>         const isUndoOperation: boolean = reinsertedNoteEmbeds.length > 0;
> ```

Introducing this variable is helpful for communicating the meaning.

Haha, actually, while debugging and doing an undo, I found that `isUndoOperation` was false. I wonder if there is another variable name we could use that communicates but is not at times incorrect. Maybe `isUndoingDeleteNoteEmbed`?



_Code quote:_
```XML
const isUndoOperation: boolean =
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1191 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7r3D0s52J3FcgprdZV:-N7r3D0t1ckT4k3VU7Qg:b-xr9t1y) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1191)):*
> ```XML
> 
>         const oldNotePosition: TextAnchor = noteThreadDoc.data.position ?? { start: 0, length: 0 };
>         const newTextAnchor: TextAnchor | undefined = this.getUpdatedTextAnchor(
> ```

Okay, it looks like the fix is in here.



_Code quote:_
```XML
this.getUpdatedTextAnchor(
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1292 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7r-cbY3BXS3ZjbY8mo:-N7r-cbY3BXS3ZjbY8mp:baky7kj) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1292)):*
> ```XML
>       }
> 
>       const editInVerseRange: boolean =
> ```

This might be better named `editIsInVerseRange` or `isEditInVerseRange`.



_Code quote:_
```XML
editInVerseRange:
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1295 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7r4thZ9davJTSKe9Lg:-N7r4thZ9davJTSKe9Lh:b-3y0r0g) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1295)):*
> ```XML
>         curIndex >= verseRange.index && curIndex <= verseRange.index + verseRange.length;
>       if (insertOp != null) {
>         if (!editInVerseRange) continue;
> ```

Hey, couldn't we omit `verseRange` from the method arguments, and instead query for the information we want here? 
In fact, couldn't we do that in updatedVerseNoteThreadAnchors() as well, and not need to pass around the verse range with the potentially-affected-embeds list?



_Code quote:_
```XML
editInVerseRange
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1317 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7r0glc5UrFrkv95Yj1:-N7r0glc5UrFrkv95Yj2:b-xl976i) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1317)):*
> ```XML
>         }
>       } else if (deleteOp != null) {
>         if (editInVerseRange) {
> ```

So, are there going to be edits that are coming to this method that are for other verses, that we don't want to be processing, then?



_Code quote:_
```XML
     if (editInVerseRange) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 1788 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7qcTCz0LB_B49s2zoC:-N7qcTCz0LB_B49s2zoD:b11sav2) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1788)):*
> ```XML
>       tick();
>       env.fixture.detectChanges();
>       const insertStart: number = notePosition + ' 1, ver'.length;
> ```

I see it doesn't necessarily matter, for the test :). But I think this is a mistake and should be `'ter 1, ver'.length;`



_Code quote:_
```XML
const insertStart: number = notePosition + ' 1, ver'.length;
```

marksvc:
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1001 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7rZ3Nf1DPpkWzpSr8x:-N8AluVsC-fS_qc_WlKo:bdk55ka) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1001)):*

To avoid any misunderstanding from the ambiguity of "before" meaning in time or position, let's use something like
```
segmentsTemporallyBeforeDelta
embedsTemporallyBeforeDelta
```
I'm not very happy with those suggested names, but I'm not sure what's best to use. 
</blockquote></details>

Some colleagues liked `preDeltaSegmentCache`.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1442)* status: all files reviewed, 11 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1001 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7rZ3Nf1DPpkWzpSr8x:-N8Z_kEG37XizoWglmtL:bg9dsy) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1001)):*

Some colleagues liked `preDeltaSegmentCache`.
</blockquote></details>

Good suggestion. I renamed it hear and in other places.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1017 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7rOh8c2TEVzANgHjGS:-N8Zb61DBdczq9BQrMdH:b-xkk72p) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1017)):*

Oh no, won't this be storing the range of a `segment` and so conceivably not of the whole verse? 

Same for a dozen lines later, with `currentVerseRange = range;`
</blockquote></details>

This will set `currentVerseRange = range` when `currentVerseRange == null` on the very first pass. Then on subsequent passes, the `currentVerseRange` length gets updated. So we only reset `currentVerseRange` when the total verse range has calculated and added to the `embedByEditedVerse` list.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1029 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7rO9gp0PKVKFDznOxj:-N8ZeyiV6MLMcrd6lKX_:b-osm7rp) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1029)):*

Oh, didn't we maybe not want to do this on the first iteration of the loop, since a dozen lines above, we do `currentVerseRange = range;`? Maybe if we omit the `currentVerseRange = range;` above, then we can safely do this line here without incrementing currentVerseRange.length more than desired?
</blockquote></details>

Good thinking. I've adjusted the logic so that even if this is the first iteration, the length is correct.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1034 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7rPO0n9NZzAZ_d0As_:-N8ZfZDm1YIug8TktIB3:b-fp02rj) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1034)):*

I know that we have handling sometimes for when an edit occurs at the end of a range, where we include it in that range, and so we  would use `<=` in `pos <= range.index + range.length` rather than the `<` operator. But it occurs to me that by doing this, we potentially will find a position "in" two different segments, if the segments are adjacent. I don't know yet if there will be a problem about it, but it's something I was just thinking.
</blockquote></details>

We are actually ok, since there will always be a length of 1 between segments. So if the edit occurred at position 10 at the end of a segment. The start of the next segment could not be at position 10. Typically, segments are separated by a verse embed or a newline character which have a length of 1.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1055 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7rTjm_0HTFhsXKLgeB:-N8ZgNsFDjnpIDBneoi5:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1055)):*

How about `s` in `getEditPositionsInDelta`?
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1076 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7rUkME3t4HsnHwwEcn:-N8Zg_0s7dDsIRjpAdj3:be3tdeq) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1076)):*

What's the thinking  behind returning `segmentRef` if there was no match? 
</blockquote></details>

Thanks for asking. Originally I did not expect the match to fail. But after working on other issues, I've discovered that paragraph breaks such as section headings do not have an associated verse segment. Paratext allows you to create notes on paragraph headings, and considers the notes on those segments as part of the previous verse. I've adjusted the logic and added a test.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1157 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7qomI62-fapDBnT_fc:-N8ZjRHy-gHVEftDhr24:b-gyd9d6) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1157)):*

So there is now an outer loop that runs the inner loop over multiple sets of embeds, that may correspond to different verse ranges. And the `editPosition` will be one within the given range, unlike perhaps `editOpIndex` above.

Are we limiting to within specific ranges for performance, or because it would be incorrect behaviour to process outside of those ranges? It looks like from `getAnchorChanges()` that the verse range is being used to only  consider and apply some insert and delete information when we are in the range. I'm unsure so far if we would be getting to getAnchorChanges() and working when we are _out_ of the range, tho. 



</blockquote></details>

We need to know what edits happened in a verse range to decide how to update the text anchor, since the text anchor is relative to the verse. So if edits occurred outside of the verse range, they can be ignored
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1176 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7qnes4EwTs0xudaQMf:-N8Zkeh07qwAeaAwKH5R:boj1qgt) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1176)):*

Introducing this variable is helpful for communicating the meaning.

Haha, actually, while debugging and doing an undo, I found that `isUndoOperation` was false. I wonder if there is another variable name we could use that communicates but is not at times incorrect. Maybe `isUndoingDeleteNoteEmbed`?
</blockquote></details>

Good catch. Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1292 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7r-cbY3BXS3ZjbY8mo:-N8Zkx0ZDeFG5NMOpznT:b-i21bbo) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1292)):*

This might be better named `editIsInVerseRange` or `isEditInVerseRange`.
</blockquote></details>

Thanks. Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1295 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7r4thZ9davJTSKe9Lg:-N8Zlvro7wkpwiAjFVyv:b1xbd9l) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1295)):*

Hey, couldn't we omit `verseRange` from the method arguments, and instead query for the information we want here? 
In fact, couldn't we do that in updatedVerseNoteThreadAnchors() as well, and not need to pass around the verse range with the potentially-affected-embeds list?

</blockquote></details>

I'm not sure what you mean by query for the information we want. We need to know the verse range to know which edits in the delta occur in the verse, then we can update note text anchors accordingly. other method arguments do not provide info about verse ranges.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1317 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7r0glc5UrFrkv95Yj1:-N8ZmO4SDiXEWG9x0vMo:b1hdrly) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1317)):*

So, are there going to be edits that are coming to this method that are for other verses, that we don't want to be processing, then?
</blockquote></details>

Yes, that is correct. We process one embed at a time for each verse that was edited.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 1788 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7qcTCz0LB_B49s2zoC:-N8ZnkcYCjitLhFuHeWW:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L1788)):*

I see it doesn't necessarily matter, for the test :). But I think this is a mistake and should be `'ter 1, ver'.length;`
</blockquote></details>

Done.

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1442)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1034 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7rPO0n9NZzAZ_d0As_:-NA5Q05z-gVqkpPi4Grp:b-cz7bae) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1034)):*

We are actually ok, since there will always be a length of 1 between segments. So if the edit occurred at position 10 at the end of a segment. The start of the next segment could not be at position 10. Typically, segments are separated by a verse embed or a newline character which have a length of 1.
</blockquote></details>

Ok, thank you.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1295 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1442#-N7r4thZ9davJTSKe9Lg:-NA5iKRFCwMOdVIMT9b2:b-20xq2k) ([raw file](https://github.com/sillsdev/web-xforge/blob/357c14fabbf33e2abd799ea41641c2d4193257dd/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1295)):*

I'm not sure what you mean by query for the information we want. We need to know the verse range to know which edits in the delta occur in the verse, then we can update note text anchors accordingly. other method arguments do not provide info about verse ranges.
</blockquote></details>

Yeah I don't think I was thinking about this the right way.

PR #1450, opened 2022-07-25
SF-1656 Do not build machine project when target has no source books

RaymondLuong3:
* detect and create suggestion engine only if conditions met on front end
* notify user when source project does not contain target books
* do not create translation engine if source books not available on back end

The recent update to Machine 2.5.11 has revealed some existing problems with how we handle building our translation engine using Machine. This corrects how we were using Machine by detecting whether we have texts to generate a translation suggestion engine with. In the event that no source texts are available for any book in our project, we skip over the step of creating a translation suggestion engine.

image

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1450)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/translation-engine.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/translation-engine.service.ts
@@ -7,6 +7,7 @@ import {
   RemoteTranslationEngine
 } from '@sillsdev/machine';
 import * as crc from 'crc-32';
+import { SFProjectProfile } from 'realtime-server/lib/esm/scriptureforge/models/sf-project';
 import { SFProjectUserConfig } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-user-config';
 import { getTextDocId } from 'realtime-server/lib/esm/scriptureforge/models/text-data';
 import { Canon } from 'realtime-server/lib/esm/scriptureforge/scripture-utils/canon';
@@ -50,6 +51,10 @@ export class TranslationEngineService extends SubscriptionDisposable {
 
   createTranslationEngine(projectId: string): RemoteTranslationEngine {
     return new RemoteTranslationEngine(projectId, this.machineHttp);
+  }
+
+  checkHasSourceBooks(project: SFProjectProfile): boolean {
+    return project.texts.filter(t => t.hasSource).length > 0;
   }
 
   /**
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -643,6 +643,19 @@ describe('EditorComponent', () => {
       env.dispose();
     }));
 
+    it('does not build machine project if no source books exists', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig();
+      when(mockedTranslationEngineService.checkHasSourceBooks(anything())).thenReturn(false);
+      env.wait();
+      verify(mockedTranslationEngineService.createTranslationEngine(anything())).never();
+      env.updateParams({ projectId: 'project02', bookId: 'MAT' });
+      env.wait();
+      verify(mockedTranslationEngineService.createTranslationEngine(anything())).never();
+      expect().nothing();
+      env.dispose();
+    }));
+
     it('change texts', fakeAsync(() => {
       const env = new TestEnvironment();
       env.setProjectUserConfig({ selectedBookNum: 40, selectedChapterNum: 1, selectedSegment: 'verse_1_1' });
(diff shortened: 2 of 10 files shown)


Nateowami (changes requested):
> Reviewed 8 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1450)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.html` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1450#-N8eQ72jD4F61IIe4uLc:-N8eQ72jD4F61IIe4uLd:bbvz194) ([raw file](https://github.com/sillsdev/web-xforge/blob/426c365cc86e9bcfb89d366816a736ee4135052b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.html#L3)):*
> ```HTML
> <ng-container *transloco="let t; read: 'translate_overview'">
>   <div class="title mat-display-1">{{ t("translate_overview") }}</div>
>   <div *ngIf="showCannotTrainEngineMessage" class="translation-suggestions-info">
> ```

Maybe hide the suggestion engine card too, in this case?

Or maybe the suggestion engine card should be shown, but this should be part of the card itself, rather than a separate disjointed message?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts` line 96 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1450#-N8eQVhdEulXFwSlPKpt:-N8eQVhdEulXFwSlPKpu:bme3ami) ([raw file](https://github.com/sillsdev/web-xforge/blob/426c365cc86e9bcfb89d366816a736ee4135052b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts#L96)):*
> ```XML
>   get showCannotTrainEngineMessage(): boolean {
>     if (this.projectDoc?.data == null) {
>       return true;
> ```

Won't this result in it being shown wrongly while the page is still loading? It seems like it would be better to not show the message when we don't yet know that this is a problem.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1450)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.html` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1450#-N8eQ72jD4F61IIe4uLc:-N8eriz99_Y-uH4zBuSF:b8przi5) ([raw file](https://github.com/sillsdev/web-xforge/blob/426c365cc86e9bcfb89d366816a736ee4135052b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.html#L3)):*

Maybe hide the suggestion engine card too, in this case?

Or maybe the suggestion engine card should be shown, but this should be part of the card itself, rather than a separate disjointed message?
</blockquote></details>

Good thinking. I'll add that onto the engine card.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts` line 96 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1450#-N8eQVhdEulXFwSlPKpt:-N8es4T01mpWBaTamACE:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/426c365cc86e9bcfb89d366816a736ee4135052b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts#L96)):*

Won't this result in it being shown wrongly while the page is still loading? It seems like it would be better to not show the message when we don't yet know that this is a problem.
</blockquote></details>

Done.

Nateowami (commented):
@raymondluong3 I can't get the message to show, and it looks like there's still a front-end test failing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1450#-:-N8f0Qbe2saswGLz0Xei:beeycmq)* status: 2 of 10 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

RaymondLuong3 (commented):
Whoops, I had left a bang on the template that let me see the message. It should be good now.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1450#-:-N8iqSlK6iKJs3DtnMbZ:b-gtqbw6)* status: 2 of 10 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

Nateowami (approved):
> Reviewed 2 of 10 files at r1, 5 of 6 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1450)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1451, opened 2022-07-26
Try splitting up GHA job

marksvc:
This PR demonstrates breaking up the GHA so that frontend tests (and sometimes the frontend build) happen in parallel with other parts of the build. 

This does speed up the whole, perhaps taking 7 minutes instead of perhaps 11. It should be noted that comparing times between GHA job runs needs to account for quite a variance in the amount of time a job can take from run to run.

This PR tries to still run the right things to accomplish testing the same situation as prior to breaking out some steps. 

When the build steps are not broken up, it may be more clear that the build process is specified correctly. 

The time saved by breaking up the build steps may not help us if we are still waiting for other aspects of a PR to become ready.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1451)

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -9,6 +9,47 @@ on:
   workflow_dispatch:
 
 jobs:
+  test-frontend:
+    name: "Build and test frontend"
+    strategy:
+      matrix:
+        os: ["ubuntu-20.04"]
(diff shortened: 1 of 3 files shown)


marksvc:
Closing. Am glad to not break up the GHA job at this time. We can revisit in the future if desired.

PR #1454, opened 2022-08-02
SF-1660 Allow re-marking node as invalid

marksvc:
- A node may be marked as "invalid-inline". It can happen again to the
same node, but crashes, due to the API method used. Mark the node as
invalid in a way that can be re-done without crashing.
- A situation of "invalid-block" double-marking did not trigger from
the app or tests, but I applied the same adjustment to the
"invalid-block" marking.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1454)


Nateowami (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1454#-:-N8VUwu34RJ_AH74RcEw:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1458, opened 2022-08-04
SF-1677 Fix suggestions settings dialog for RTL UI

Nateowami:
Before | After
---------|--------
 | 

You can't see it in the screenshots, but the main issue was that the slider causes a scrollbar. The other issue was the toggle was too close to the text.

This _doesn't_ fix the issue of the dialog flowing the wrong direction if the language has been changed, and the page hasn't been reloaded. A number of dialogs had that problem, and it's easier to fix them all at once.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1458)


RaymondLuong3 (approved):
I like this approach of putting the form elements on their own rows using flex-direction: column.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1458#-:-N8dnwo1ACg5Phdlpdpm:b-9krog0)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami:
Note: The branch name says SF-1633, but really it is SF-1677. The branch name is in error.

PR #1459, opened 2022-08-04
SF-1633 SF-1634 SF-1631 Fix direction of dialogs when UI is set to RTL

Nateowami:
The problem for all of these issues is that a MatDialog doesn't get the correct direction set if the direction changed after the page loaded (everything works fine if the locale hasn't changed since the page loaded).

Right now most of our dialogs are still MdcDialog, so the issue only affected a few dialogs. However, the number is likely to grow as we move towards Material, so I wanted a general solution to the problem. Hence the creation of a DialogService that sets the direction when the dialog is opened.

There are a some settings that can be set when opening a dialog that all seem like they should (generally) be the concern of the dialog itself. These include:
- Whether to autofocus the first focusable element
- The default width of the dialog
- Whether the user can use escape or clicking on the backdrop to close the dialog

However, these have to be set by the component that opens the dialog. One of my ideas in creating the service is that perhaps (as a future step) we can have dialog component specify default config values, and the dialog service can read those values from the component, so the calling component doesn't have to concern itself with any of this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1459)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts
@@ -255,17 +255,17 @@ describe('CheckingOverviewComponent', () => {
       env.simulateRowClick(1, id);
       // Edit a question with no answers
       env.clickElement(env.questionEditButtons[3]);
-      verify(mockedMdcDialog.open(QuestionAnsweredDialogComponent)).never();
+      verify(mockedMdcDialog.open(QuestionAnsweredDialogComponent, anything())).never();
       resetCalls(mockedMdcDialog);
       when(env.mockedAnsweredDialogRef.afterClosed()).thenReturn(of('close'));
       // Edit a question with answers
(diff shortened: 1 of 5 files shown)


RaymondLuong3 (approved):
Testing this on this branch, I see the RTL dialogs are showing as expected.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1459#-:-N98Kd9T-gUIGioYHwMa:b-56fosv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami:
@RaymondLuong3 Just to clarify, the issue was _only_ when you switch the page to a RTL language and then don't reload the page. If you reload the page then it works, even without this change.

RaymondLuong3 (approved):
Yep, I saw that in the issue.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1459#-:-N98XTKC8MDBCT42Ki5j:bv69v47)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1461, opened 2022-08-05
SF-1665 Clear edit history before chapter embeds are duplicated

RaymondLuong3:
Texts were being corrupted when a user typed in the first section heading in a blank chapter with no introductory texts or headings. This is a side effect of a problem when Quill diffs the delta before and after a change is applied. The diff algorithm that Quill users to compare deltas cannot accurately determine if embeds have changed. Most of the time this is not an issue, and quill simply assumes that embeds have changed, reintroduces them, and deletes the originals.

This change clears the history stack if it finds a chapter embed is being inserted as a result of an undo operation. The user will not be able to undo their edits to the section heading.

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/test-utils.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/test-utils.ts
@@ -76,10 +76,12 @@ export function getPoetryVerseTextDoc(id: TextDocId): TextData {
   return delta;
 }
 
-export function getEmptyChapterDoc(id: TextDocId): TextData {
+export function getEmptyChapterDoc(id: TextDocId, includeHeading: boolean = true): TextData {
   const delta = new Delta();
-  delta.insert({ blank: true }, { segment: 's_1' });
-  delta.insert('\n', { para: { style: 's' } });
+  if (includeHeading) {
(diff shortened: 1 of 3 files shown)


RaymondLuong3 (commented):
Here are the steps that lead to the bad behaviour:

1. User enters a word in the first segment after the chapter.
2. User does ctrl + z to undo the text changes.
3. The undo op looks like { insert: chapter, insert: blank, delete: 2 }
4. The text is not fully deleted (the blot that was inserted with the blank was mistakenly considered to have length 2) so only the duplicate chapter is deleted, when it should also have deleted the text.
5. The text-view-model makes adjustments, deleting the blank since text still exists.
6. A delta gets added to the undo stack that looks like { insert: chapter, insert: blank }
7. The user can undo, and corrupt their text.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1461#-:-NAPyyCj-_SVDjzxDioB:b-2dfrjj)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @Nateowami)

pmachapman:
@RaymondLuong3 Does this still happen with the Quill upgrade? I can no longer recreate this bug using master.

RaymondLuong3:
Amazing! I am also not able to reproduce the issue. I will close this.

PR #1468, opened 2022-08-12
SF-1668 Store the current project id and navigate on login

RaymondLuong3:
* add currentProjectId property on the Site class
* store the project id on the user's site in the data
* use the stored id when no project id exists in local storage
* modify user service currentProjectId to accept a user parameter
* add user service test

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1468)

### DIFF
--- a/src/RealtimeServer/common/models/site.ts
+++ b/src/RealtimeServer/common/models/site.ts
@@ -1,3 +1,6 @@
 export interface Site {
+  // The most recent project a user has navigated to, which may be different than the current open project
+  // if a user is logged in on multiple browsers
+  currentProjectId?: string;
   projects: string[];
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -114,7 +114,7 @@ describe('AppComponent', () => {
     expect(env.isDrawerVisible).toEqual(true);
     expect(env.selectedProjectId).toEqual('project01');
     expect(env.menuLength).toEqual(5);
-    verify(mockedUserService.setCurrentProjectId('project01')).once();
+    verify(mockedUserService.setCurrentProjectId(anything(), 'project01')).once();
   }));
 
   it('navigate to different project', fakeAsync(() => {
@@ -128,7 +128,7 @@ describe('AppComponent', () => {
     expect(env.menuLength).toEqual(5);
     expect(env.component.isCheckingEnabled).toEqual(true);
     expect(env.component.isTranslateEnabled).toEqual(false);
-    verify(mockedUserService.setCurrentProjectId('project02')).once();
+    verify(mockedUserService.setCurrentProjectId(anything(), 'project02')).once();
   }));
 
   it('hide translate tool for community checkers', fakeAsync(() => {
@@ -142,7 +142,7 @@ describe('AppComponent', () => {
     expect(env.menuLength).toEqual(5);
     expect(env.component.isCheckingEnabled).toEqual(true);
     expect(env.component.isTranslateEnabled).toEqual(false);
-    verify(mockedUserService.setCurrentProjectId('project03')).once();
+    verify(mockedUserService.setCurrentProjectId(anything(), 'project03')).once();
 
     // Does not collapse Community Checking item when translate is disabled
     env.selectItem(0);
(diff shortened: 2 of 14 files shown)


Nateowami (changes requested):
> Reviewed 12 of 14 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1468)* status: 12 of 14 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/user.service.ts` line 51 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1468#-NAAvZNYC9gUiXot4vO9:-NAAvZNYC9gUiXot4vOA:b-8qo4fd) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eb8bc515370c60b814fcc3ec4b07a3f8a73f7a9/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/user.service.ts#L51)):*
> ```XML
>   async setCurrentProjectId(user?: UserDoc, value?: string): Promise<void> {
>     this.localSettings.set(CURRENT_PROJECT_ID_SETTING, value);
>     await user?.submitJson0Op(update => {
> ```

`user` needs to be non-optional, otherwise this is a no-op when it should clear the setting.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1468)* status: 8 of 14 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/user.service.ts` line 51 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1468#-NAAvZNYC9gUiXot4vO9:-NABZN3X69K03-IxoT0Q:b-ilwfnd) ([raw file](https://github.com/sillsdev/web-xforge/blob/0eb8bc515370c60b814fcc3ec4b07a3f8a73f7a9/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/user.service.ts#L51)):*

`user` needs to be non-optional, otherwise this is a no-op when it should clear the setting.
</blockquote></details>

Thanks. I've made both non-optional.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1468)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 418 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1468#-NACSzwQ0Cb7aFWPqbeL:-NACSzwR1a9EdGJ0_bi7:b-12myw) ([raw file](https://github.com/sillsdev/web-xforge/blob/dca5c50a672620d4067d8c210375eff9dffb2fb4/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L418)):*
> ```XML
>         }
>         if (this.currentUserDoc != null) {
>           await this.userService.setCurrentProjectId(this.currentUserDoc, this.selectedProjectDoc.id);
> ```

Is there any reason to think `this.currentUserDoc` might be undefined? If not, then I think it makes more sense to assert that it's not null. While it's generally preferable to avoid non-null assertions, the advantage of an assertion is that if it turns out that you're wrong, you'll eventually run into an error and find out that you were wrong and can address it. In a case like this, if we're wrong it will just silently not behave correctly.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/user.service.spec.ts` line 50 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1468#-NACZ8IMCSPYmDlyDiaR:-NACZ8INBlRJty7guCvx:bwwnkc8) ([raw file](https://github.com/sillsdev/web-xforge/blob/dca5c50a672620d4067d8c210375eff9dffb2fb4/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/user.service.spec.ts#L50)):*
> ```XML
>     expect(user.data!.sites['sf'].currentProjectId).toEqual('project01');
>     env.service.setCurrentProjectId(user, 'project02');
>     expect(user.data!.sites['sf'].currentProjectId).toEqual('project02');
> ```

It looks like there's no test to make sure it saves locally as well?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1468)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1468#-:-NAFo9O5CsIyg4a8UYB9:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1476, opened 2022-08-17
SF-1674 Inform user when lacking permission to edit chapter

Nateowami:
It's happened a couple of times that a user has written to us because the editor was in read-only mode. In both cases I'm thinking of it was because while the user had permission to edit in general (admin or translator), the permissions were set in such a way that the user didn't have permission for that particular chapter.

This is user error, but it tends to happen when a user is just getting started with Scripture Forge, and the whole Paratext permissions system is complicated, so it makes sense to tell the user why the chapter can't be edited.

Here's how it looks:

It's not as elegant as it could be, but it follow the same design as our other warnings. If we want to improve the appearance we can go back and change the style of all of them. Right not it's just about functionality.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1476)


RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1476)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 275 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1476#-N9hMsxy0CNcEUT9MeXo:-N9hMsxz5kIpY5uGGIa2:b-k8nqp4) ([raw file](https://github.com/sillsdev/web-xforge/blob/16febcc07fbf581563c3d9b0e48420b7ef2f0bef/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L275)):*
> ```XML
>   }
> 
>   get userRole(): string {
> ```

I don't think it is necessary to include the user role. It may be extra information. What the user really needs to know is  that their permission to edit the chapter needs to be updated. Is there a reason you want to include the role?



_Code quote:_
```XML
 get userRole(): string {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 86 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1476#-N9hNOePCX6qdxa_rx2p:-N9hNOePCX6qdxa_rx2q:b-nzdd0q) ([raw file](https://github.com/sillsdev/web-xforge/blob/16febcc07fbf581563c3d9b0e48420b7ef2f0bef/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L86)):*
> ```JSON
>     "configure_translation_suggestions": "Configure translation suggestions",
>     "more_notes": "--- {{ count }} more note(s) ---",
>     "no_permission_edit_chapter": "You don't have permission to edit this chapter, even though you have the {{ userRole }} role. Permissions can only be changed in Paratext.",
> ```

I would simply say "You do not have permission to edit this chapter. Contact your project administrator to update your edit permissions in Paratext.



_Code quote:_
```JSON
    "no_permission_edit_chapter": "You don't have permission to edit this chapter, even though you have the {{ userRole }} role. Permissions can only be changed in Paratext.",
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1476)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 86 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1476#-N9hNOePCX6qdxa_rx2p:-N9ho9pa0Mg3SZI_-vFo:bkshbvf) ([raw file](https://github.com/sillsdev/web-xforge/blob/16febcc07fbf581563c3d9b0e48420b7ef2f0bef/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L86)):*

I would simply say "You do not have permission to edit this chapter. Contact your project administrator to update your edit permissions in Paratext.
</blockquote></details>

The reason I included the role is that multiple times users have written to us saying it won't let them edit, thinking it was an issue with Scripture Forge. Maybe just saying "you don't have permission to edit; admins can change permissions in Paratext" is sufficient, but I thought even then users might think it's a mistake on the part of SF.

In the most recent example David Baines (who admittedly is probably not a frequent user of Paratext) didn't have permission to edit, despite being an admin on the project. In another, Beth was using a translator account and was confused why it wouldn't let her edit.

In some sense disallowing editing (which we're currently doing) is telling the use they don't have permission to edit, and that's not clear enough to the user. Maybe explicitly stating you don't have permission is enough, and implies more than just not allowing editing? My hope was to direct the user away from thinking "Scripture Forge isn't pulling in the right permissions, or user role, or editing is broken, or something didn't sync right, or I set it up wrong" and towards "oh, I forgot admins and translators don't necessarily have permission to edit any given chapter." It does make it kind of wordy though.

RaymondLuong3 (commented):
Hmm, Ok that is fair. I think you have convinced me that the wordy-ness is acceptable. Being over-wordy is better than user confusion as I think this through. Could we bold the role so that it is immediately clear to the user?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1476#-:-N9ilbTc4QIYuBuKiFKk:bd0oirl)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1476)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 86 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1476#-N9hNOePCX6qdxa_rx2p:-N9ilz0-A17RrdZSKJNv:bfvel4d) ([raw file](https://github.com/sillsdev/web-xforge/blob/16febcc07fbf581563c3d9b0e48420b7ef2f0bef/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L86)):*

The reason I included the role is that multiple times users have written to us saying it won't let them edit, thinking it was an issue with Scripture Forge. Maybe just saying "you don't have permission to edit; admins can change permissions in Paratext" is sufficient, but I thought even then users might think it's a mistake on the part of SF.

In the most recent example David Baines (who admittedly is probably not a frequent user of Paratext) didn't have permission to edit, despite being an admin on the project. In another, Beth was using a translator account and was confused why it wouldn't let her edit.

In some sense disallowing editing (which we're currently doing) is telling the use they don't have permission to edit, and that's not clear enough to the user. Maybe explicitly stating you don't have permission is enough, and implies more than just not allowing editing? My hope was to direct the user away from thinking "Scripture Forge isn't pulling in the right permissions, or user role, or editing is broken, or something didn't sync right, or I set it up wrong" and towards "oh, I forgot admins and translators don't necessarily have permission to edit any given chapter." It does make it kind of wordy though.
</blockquote></details>

Could we bold the role?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1476)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 86 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1476#-N9hNOePCX6qdxa_rx2p:-N9ni9LWAQYROuRTH6j9:bojz2fm) ([raw file](https://github.com/sillsdev/web-xforge/blob/16febcc07fbf581563c3d9b0e48420b7ef2f0bef/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L86)):*

Could we bold the role?
</blockquote></details>

We could, though to me it looks a bit odd. It's also kind of odd to implement. See the branch `origin/sf-1674-inform-user-no-edit-permission-role-bold`

RaymondLuong3 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1476#-:-NA5RNpi735yvFfL8dSK:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 86 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1476#-N9hNOePCX6qdxa_rx2p:-NA5Qafl61hHHWbKN4XH:betnfd8) ([raw file](https://github.com/sillsdev/web-xforge/blob/16febcc07fbf581563c3d9b0e48420b7ef2f0bef/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L86)):*

We could, though to me it looks a bit odd. It's also kind of odd to implement. See the branch `origin/sf-1674-inform-user-no-edit-permission-role-bold`



</blockquote></details>

Hmm, I actually like the bold format for the role. I quickly looked through our app, and it doesn't look like we bold the roles anywhere I can see. So I am ok leaving it normal font.

PR #1478, opened 2022-08-19
SF-1693 Fix question icons not showing up

nigel-wells:
- Moved filtering questions to where questions are toggled
- Fix to remove the active verse if the text changes
- Added missing selector ] in tests

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1478)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.spec.ts
@@ -91,6 +91,22 @@ describe('CheckingTextComponent', () => {
     expect(env.isSegmentHighlighted(1, 4)).toBe(true);
   }));
 
+  it('should show related questions when the text changes ', fakeAsync(() => {
+    const env = new TestEnvironment();
+    env.wait();
+
+    expect(env.segmentHasQuestion(1, 1)).toBe(true);
+    expect(env.segmentHasQuestion(1, 3)).toBe(true);
+    expect(env.segmentHasQuestion(2, 1)).toBe(false);
+
+    env.component.id = new TextDocId('project01', 40, 2);
+    env.wait();
+
+    expect(env.segmentHasQuestion(1, 1)).toBe(false);
+    expect(env.segmentHasQuestion(1, 3)).toBe(false);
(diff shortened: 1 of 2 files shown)


RaymondLuong3 (commented):
It looks like it is working. There looks to be excessive space between the question icon and the text. What do you think?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1478#-:-NA5VwfX1nk1Q_MiykdY:b-ukovz6)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

nigel-wells (commented):
Now that you mention it I see what you mean. Looking back at the history it has been this way since 2019 so nothing new. I haven't looked extensively across multiple languages or fonts to see if it is best to have the gap like it is. It looks bigger when the question is highlighted but looks fine when it isn't IMO. I'd say if we've been happy with it for so long then let's leave it as is.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1478#-:-NBASYccBZhoBcPflOo0:bg097fd)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (approved):
Thanks for looking into this. A little extra space is not a big issue. If it has been like this for 3 years now without any complaints, I am satisfied.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1478#-:-NBIY-c55dDbHJt07SSK:b9oasuw)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

nigel-wells:
I think in this context it doesn't need to be "complete" valid syntax - think of it more of a regex, so long as it matches something then that is fine rather than having to be technically correct. For example, you can remove the double quotes for the class and it still works - well, not in the above example as it is looking for multiple classes.

Nateowami (commented):
@nigel-wells I just did a test, and to my surprise, this expression does return results: `document.querySelectorAll('[class=heading-with-icon')`

So it appears you are correct.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1478#-:-NBQ0hp-6A-6zWZRT6dY:b-6iby84)* status: all files reviewed, 1 unresolved discussion

PR #1483, opened 2022-08-31
SF-1703 Show editable content only for users with write permissions

RaymondLuong3:
* determine user's permission based on their project role
* show insert note button for users with permission
* show the save button for users with permission

### DIFF
--- a/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts
+++ b/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts
@@ -54,6 +54,10 @@ export class SFProjectRights extends ProjectRights {
       { projectDomain: SFProjectDomain.Likes, operation: Operation.Create },
       { projectDomain: SFProjectDomain.Likes, operation: Operation.DeleteOwn },
 
+      { projectDomain: SFProjectDomain.NoteThreads, operation: Operation.Create },
+      { projectDomain: SFProjectDomain.NoteThreads, operation: Operation.Edit },
+      { projectDomain: SFProjectDomain.NoteThreads, operation: Operation.Delete },
+
       { projectDomain: SFProjectDomain.Notes, operation: Operation.Create },
       { projectDomain: SFProjectDomain.Notes, operation: Operation.EditOwn },
       { projectDomain: SFProjectDomain.Notes, operation: Operation.DeleteOwn }
@@ -67,15 +71,7 @@ export class SFProjectRights extends ProjectRights {
     this.addRights(SFProjectRole.ParatextConsultant, ptReviewerRights);
 
     const translatorRights: ProjectRight[] = ptReviewerRights.concat([
-      { projectDomain: SFProjectDomain.Texts, operation: Operation.Edit },
-
-      { projectDomain: SFProjectDomain.NoteThreads, operation: Operation.Create },
-      { projectDomain: SFProjectDomain.NoteThreads, operation: Operation.Edit },
-      { projectDomain: SFProjectDomain.NoteThreads, operation: Operation.Delete },
-
-      { projectDomain: SFProjectDomain.Notes, operation: Operation.Create },
-      { projectDomain: SFProjectDomain.Notes, operation: Operation.EditOwn },
-      { projectDomain: SFProjectDomain.Notes, operation: Operation.DeleteOwn }
+      { projectDomain: SFProjectDomain.Texts, operation: Operation.Edit }
     ]);
     this.addRights(SFProjectRole.ParatextTranslator, translatorRights);
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts
@@ -1,3 +1,6 @@
+import { Operation } from 'realtime-server/lib/esm/common/models/project-rights';
+import { SFProjectProfile } from 'realtime-server/lib/esm/scriptureforge/models/sf-project';
+import { SFProjectDomain, SF_PROJECT_RIGHTS } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-rights';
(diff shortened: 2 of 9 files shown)


Nateowami (changes requested):
Looks good. Just one LGTM fix needed, and one minor complaint.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1483#-:-NBrjB4y3q9A2AzA1VkF:b-ada62l)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 2335 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1483#-NBrZuoUDvRG44dRJYf-:-NBrZuoUDvRG44dRJYf0:b-vrbrjo) ([raw file](https://github.com/sillsdev/web-xforge/blob/fc8a0d0059b838ea287a93185f7e3af82917bbe3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L2335)):*
> ```XML
>       // user02 does not have read permission on the text
>       const usersWhoCanInsertNotes = ['user01', 'user03', 'user04', 'user05'];
>       usersWhoCanInsertNotes.forEach(user => {
> ```

Can we do:
``` ts
for (const user of usersWhoCanInsertNotes) {
```

I'm pretty sure in practice it's going to behave exactly the same, but I'm always a tad reluctant to use callbacks to create assertions in tests, since if a callback doesn't actually get called before the test completes, the assertions don't get made, and the test silently passes without actually asserting what it was supposed to assert (though again, I don't think that would be a real problem in this case).

Also, I find using `forEach()` to loop kind of odd, since when I see it I tend to assume there was a reason it needed to be done that way, and start looking closer to see if I overlooked something.

Unless perhaps there's a reason I'm overlooking?

The same thing would apply to `usersWhoCannotInsertNotes` below.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts` line 20 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1483#-NBrZd6wAnjD-LutQpmc:-NBrZd6wAnjD-LutQpmd:beyn0pl) ([raw file](https://github.com/sillsdev/web-xforge/blob/fc8a0d0059b838ea287a93185f7e3af82917bbe3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L20)):*
> ```XML
> import { TextAnchor } from 'realtime-server/lib/esm/scriptureforge/models/text-anchor';
> import { SFProjectRights, SF_PROJECT_RIGHTS } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-rights';
> import { canInsertNote } from 'src/app/shared/utils';
> ```

This is the line LGTM is complaining about. We typically use relative imports when possible.

Nateowami (commented):
When a verse is blank, this looks kind of odd.


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1483#-:-NBrpmn0E-G2eEKx9dfB:byn5b5f)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
It does look weird. I gave the element a min height, so it looks better, but will likely adjust in to look better in the future.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1483#-:-NCS_Zoi1rNZw_ahS8yD:brkn1tj)* status: 3 of 10 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 2335 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1483#-NBrZuoUDvRG44dRJYf-:-NCSSK9j3TuDGd15v8Ng:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/fc8a0d0059b838ea287a93185f7e3af82917bbe3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L2335)):*

Can we do:
``` ts
for (const user of usersWhoCanInsertNotes) {
```

I'm pretty sure in practice it's going to behave exactly the same, but I'm always a tad reluctant to use callbacks to create assertions in tests, since if a callback doesn't actually get called before the test completes, the assertions don't get made, and the test silently passes without actually asserting what it was supposed to assert (though again, I don't think that would be a real problem in this case).

Also, I find using `forEach()` to loop kind of odd, since when I see it I tend to assume there was a reason it needed to be done that way, and start looking closer to see if I overlooked something.

Unless perhaps there's a reason I'm overlooking?

The same thing would apply to `usersWhoCannotInsertNotes` below.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts` line 20 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1483#-NBrZd6wAnjD-LutQpmc:-NCSRjt53MvaSkJLMOJd:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/fc8a0d0059b838ea287a93185f7e3af82917bbe3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L20)):*

This is the line LGTM is complaining about. We typically use relative imports when possible.
</blockquote></details>

Done.

Nateowami (approved):
:lgtm:

> Reviewed 4 of 7 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1483#-:-NCWOe9p6bgtOEj2_7KO:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1486, opened 2022-09-02
SF-1704 Allow user to add a note and edit notes in existing thread

RaymondLuong3:
* make note dialog opens with text area by default
* show edit buttons on last note if user is the author
* detect when existing note content changes
* add note to thread if editing existing note is not selected
* get edit icons to display in mobile view

Note-dialog-edit-actions

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1486)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -45,7 +45,7 @@ import { Canon } from 'realtime-server/lib/esm/scriptureforge/scripture-utils/ca
 import { VerseRef } from 'realtime-server/lib/esm/scriptureforge/scripture-utils/verse-ref';
 import * as RichText from 'rich-text';
 import { BehaviorSubject, defer, Observable, of, Subject } from 'rxjs';
-import { anything, capture, deepEqual, instance, mock, resetCalls, verify, when } from 'ts-mockito';
+import { anything, capture, deepEqual, instance, mock, objectContaining, resetCalls, verify, when } from 'ts-mockito';
 import { AuthService } from 'xforge-common/auth.service';
 import { CONSOLE } from 'xforge-common/browser-globals';
 import { BugsnagService } from 'xforge-common/bugsnag.service';
@@ -74,6 +74,7 @@ import { EditorComponent, SF_NOTE_THREAD_PREFIX, UPDATE_SUGGESTIONS_TIMEOUT } fr
 import { NoteDialogComponent, NoteDialogResult } from './note-dialog/note-dialog.component';
 import { SuggestionsComponent } from './suggestions.component';
 import { ACTIVE_EDIT_TIMEOUT } from './translate-metrics-session';
+import { NoteDialogData } from './note-dialog/note-dialog.component';
 
 const mockedAuthService = mock(AuthService);
 const mockedSFProjectService = mock(SFProjectService);
@@ -2399,28 +2400,16 @@ describe('EditorComponent', () => {
       const env = new TestEnvironment();
       env.setProjectUserConfig();
       env.wait();
-      const date: string = '2022-08-01T01:00:000Z';
-      const note: Note = {
-        dataId: 'notenew01',
-        threadId: '',
-        extUserId: 'user01',
-        ownerRef: 'user01',
-        status: NoteStatus.Todo,
-        content: 'New note thread',
-        conflictType: NoteConflictType.DefaultValue,
-        dateCreated: date,
-        dateModified: date,
-        deleted: false,
-        type: NoteType.Normal
-      };
+
       const position: TextAnchor = { start: 0, length: 0 };
       const selectedText = 'target: chapter 1, verse 4.';
       const verseRef: VerseRefData = fromVerseRef(VerseRef.parse('Mat 1:4'));
       env.setSelectionAndInsertNote('verse_1_4');
 
+      const emptyThreadId = '';
+      const note: Note = env.getNoteTemplate(emptyThreadId);
       env.mockNoteDialogRef.close({ verseRef, note, position, selectedText });
-      tick();
-      env.fixture.detectChanges();
+      env.wait();
       verify(mockedMatDialog.open(NoteDialogComponent, anything())).once();
       const [, config] = capture(mockedMatDialog.open).last();
       const noteVerseRef: VerseRef = (config as MatDialogConfig).data!.verseRef;
@@ -2430,6 +2419,40 @@ describe('EditorComponent', () => {
       const noteInserted = noteThread.notes[0];
       expect(noteThread.dataId).toEqual(noteInserted.threadId);
       expect(noteThread.dataId.startsWith(SF_NOTE_THREAD_PREFIX)).toBe(true);
+      env.dispose();
+    }));
+
+    it('can add a note to an existing thread', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig();
+      env.wait();
+
+      const threadId = 'thread01';
+      const existingThread: NoteThreadDoc = env.getNoteThreadDoc('project01', threadId)!;
+      expect(existingThread.data!.notes.length).toEqual(3);
+
+      const note: Note = env.getNoteTemplate(threadId);
+      const content = 'new note on thread';
+      env.openNoteDialogAndEdit('verse_1_1', existingThread.data!, note, content);
+      expect(existingThread!.data!.notes.length).toEqual(4);
+      env.dispose();
+    }));
+
+    it('can edit an existing note', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig();
+      env.wait();
+
+      const threadId = 'thread01';
+      const existingThread: NoteThreadDoc = env.getNoteThreadDoc('project01', threadId)!;
+      expect(existingThread.data!.notes.length).toEqual(3);
+
+      const lastNote: Note = cloneDeep(existingThread.data!.notes[2]);
+      const lastNoteId: string = lastNote.dataId;
+      const content = 'edit content';
+      env.openNoteDialogAndEdit('verse_1_1', existingThread.data!, lastNote, content);
+      expect(existingThread.data!.notes.length).toEqual(3);
+      expect(existingThread.data!.notes.find(n => n.dataId === lastNoteId)!.content).toEqual(content);
       env.dispose();
     }));
   });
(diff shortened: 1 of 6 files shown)


Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1486)* status: all files reviewed, 4 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html` line 43 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1486#-NBwFifWClOYTrJCoaXR:-NBwFifX8oLpY1pC83MB:bp6929s) ([raw file](https://github.com/sillsdev/web-xforge/blob/66ac78c7953aae0dbc6dcf5233eb32eadb99c4be/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html#L43)):*
> ```HTML
>           </div>
>         </div>
>         <div fxShow.xs fxHide.gt-xs *ngIf="isNoteEditable(note)" class="edit-actions">
> ```

This duplication of buttons can be easily avoided with a little CSS. I'll open a PR against this branch, since it's easier to make a change across a couple files than to suggest one.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.scss` line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1486#-NBtRWlFELpOM4m2ELMP:-NBtRWlFELpOM4m2ELMQ:b-yx6k0x) ([raw file](https://github.com/sillsdev/web-xforge/blob/66ac78c7953aae0dbc6dcf5233eb32eadb99c4be/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.scss#L78)):*
I don't think this rule is doing anything. `hidden` is not a valid value for the `display` property. I think you may be mixing up `display: none` and `visibility: hidden`.

[`display: none`](https://developer.mozilla.org/en-US/docs/Web/CSS/display)
> Turns off the display of an element so that it has no effect on layout (the document is rendered as though the element did not exist). All descendant elements also have their display turned off. To have an element take up the space that it would normally take, but without actually rendering anything, use the visibility property instead.

[`visibility: hidden`](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility)
> The element box is invisible (not drawn), but still affects layout as normal. Descendants of the element will be visible if they have visibility set to visible. The element cannot receive focus (such as when navigating through tab indexes).
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts` line 527 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1486#-NBw9jmJ8-8jwzwhX0EL:-NBw9jmJ8-8jwzwhX0EM:bn2p2wv) ([raw file](https://github.com/sillsdev/web-xforge/blob/66ac78c7953aae0dbc6dcf5233eb32eadb99c4be/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts#L527)):*
> ```XML
>     expect(env.notes.length).toEqual(3);
>     noteNumbers.forEach(n => expect(env.hasEditActions(n)).toBe(false));
>     expect(env.component.currentNoteContent).toEqual('note05');
> ```

I'm confused. I thought there were 4 notes, and the last one is being edited. Where is the number 5 coming from?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts` line 863 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1486#-NBw8oRJC6B_iDXhpVox:-NBw8oRJC6B_iDXhpVoy:b-r3zlve) ([raw file](https://github.com/sillsdev/web-xforge/blob/66ac78c7953aae0dbc6dcf5233eb32eadb99c4be/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts#L863)):*
> ```XML
> 
>   hasEditActions(noteNumber: number): boolean {
>     const noteElement: DebugElement = this.overlayContainerElement.query(
> ```

This seems like a slightly odd way to implement this method. Couldn't it just be:
``` ts
const noteElements: DebugElement[] = this.overlayContainerElement.queryAll(By.css('.notes .note'));
return noteElements[noteNumber]?.query(By.css('.edit-actions')) != null;
```
Or even simpler:
``` ts
return this.overlayContainerElement.query(By.css(`.notes .note:nth-child(${noteNumber}) .edit-actions`)) != null;
```

Nateowami (commented):
I've opened a PR to suggest changes to this branch: https://github.com/sillsdev/web-xforge/pull/1498

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1486#-:-NBwI7PW3fnY4NJL8ISl:b-20o18h)* status: all files reviewed, 4 unresolved discussions (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1486)* status: 0 of 6 files reviewed, 4 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html` line 43 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1486#-NBwFifWClOYTrJCoaXR:-NCWDpYA2QCOTVP4hspu:b-q4qpxi) ([raw file](https://github.com/sillsdev/web-xforge/blob/66ac78c7953aae0dbc6dcf5233eb32eadb99c4be/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html#L43)):*

This duplication of buttons can be easily avoided with a little CSS. I'll open a PR against this branch, since it's easier to make a change across a couple files than to suggest one.
</blockquote></details>

Great, I'll get that in as soon as I can.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.scss` line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1486#-NBtRWlFELpOM4m2ELMP:-NBxvl61842NLBE-4l2E:bjbzdwg) ([raw file](https://github.com/sillsdev/web-xforge/blob/66ac78c7953aae0dbc6dcf5233eb32eadb99c4be/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.scss#L78)):*

I don't think this rule is doing anything. `hidden` is not a valid value for the `display` property. I think you may be mixing up `display: none` and `visibility: hidden`.

[`display: none`](https://developer.mozilla.org/en-US/docs/Web/CSS/display)
> Turns off the display of an element so that it has no effect on layout (the document is rendered as though the element did not exist). All descendant elements also have their display turned off. To have an element take up the space that it would normally take, but without actually rendering anything, use the visibility property instead.

[`visibility: hidden`](https://developer.mozilla.org/en-US/docs/Web/CSS/visibility)
> The element box is invisible (not drawn), but still affects layout as normal. Descendants of the element will be visible if they have visibility set to visible. The element cannot receive focus (such as when navigating through tab indexes).
</blockquote></details>

Thanks for looking into this. It looks like I did have the rule incorrect. It also appears that I do not need that rule anymore since `isNoteEditable` accounts for whether the note is the last in the list, so I have removed it.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts` line 527 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1486#-NBw9jmJ8-8jwzwhX0EL:-NBxw4SP0NN-1hqFNYjw:bs4x6mf) ([raw file](https://github.com/sillsdev/web-xforge/blob/66ac78c7953aae0dbc6dcf5233eb32eadb99c4be/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts#L527)):*

I'm confused. I thought there were 4 notes, and the last one is being edited. Where is the number 5 coming from?
</blockquote></details>

the note called `note0
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts` line 863 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1486#-NBw8oRJC6B_iDXhpVox:-NCWDui54ZGMRY8PglJO:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/66ac78c7953aae0dbc6dcf5233eb32eadb99c4be/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts#L863)):*

This seems like a slightly odd way to implement this method. Couldn't it just be:
``` ts
const noteElements: DebugElement[] = this.overlayContainerElement.queryAll(By.css('.notes .note'));
return noteElements[noteNumber]?.query(By.css('.edit-actions')) != null;
```
Or even simpler:
``` ts
return this.overlayContainerElement.query(By.css(`.notes .note:nth-child(${noteNumber}) .edit-actions`)) != null;
```
</blockquote></details>

Done.

Nateowami (commented):
> Reviewed 2 of 6 files at r2, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1486)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3149 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1486#-NCWEpMTDfFv6k_go4h_:-NCWEpMTDfFv6k_go4ha:b-psfss6) ([raw file](https://github.com/sillsdev/web-xforge/blob/19822c76b0a725d8b668dbe701f2f27b3a831354/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3149)):*
> ```XML
>       selectedText: thread.originalSelectedText
>     });
>     this.wait();
> ```

Maybe the wait should be in the `close()` method?

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1486)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3149 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1486#-NCWEpMTDfFv6k_go4h_:-NCX77eC9Yym0EcY5WGL:bywgnj3) ([raw file](https://github.com/sillsdev/web-xforge/blob/19822c76b0a725d8b668dbe701f2f27b3a831354/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3149)):*

Maybe the wait should be in the `close()` method?
</blockquote></details>

We could, but often I get tripped up with the message that 'tick' can only be called in a fake async. In this case, it would work, but not every case.

Nateowami (approved):
:lgtm:

> Reviewed 4 of 6 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1486#-:-NCvTajj1ip_45mU7Ai8:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3149 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1486#-NCWEpMTDfFv6k_go4h_:-NCvREjgB53B0DwCroJl:bdcv6l8) ([raw file](https://github.com/sillsdev/web-xforge/blob/19822c76b0a725d8b668dbe701f2f27b3a831354/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3149)):*

We could, but often I get tripped up with the message that 'tick' can only be called in a fake async. In this case, it would work, but not every case.
</blockquote></details>

Good point.

Nateowami (approved):
> Reviewed 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1486)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1496, opened 2022-09-13
SF-1724 Allow a user to delete a note from a thread

RaymondLuong3:
* show a confirmation dialog when a user clicks to delete a note
* delete the note from the thread if the user confirms
* delete the note thread doc if the note is the last one in the thread

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1496)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -2455,6 +2455,34 @@ describe('EditorComponent', () => {
       expect(existingThread.data!.notes.find(n => n.dataId === lastNoteId)!.content).toEqual(content);
       env.dispose();
     }));
+
+    it('allows user to delete a thread', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig();
+      env.wait();
+
+      const threadId = 'thread02';
+      const existingThread: NoteThreadDoc = env.getNoteThreadDoc('project01', threadId);
+      expect(existingThread.data!.notes.length).toEqual(1);
+      const note: Note = existingThread.data!.notes[0];
+      env.openNoteDialogAndEdit('verse_1_3', existingThread.data!, note, 'deleted note', true);
+      expect(existingThread.data).toBeUndefined();
+      env.dispose();
+    }));
+
+    it('deletes the note if notes still exist in the thread', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig();
+      env.wait();
+
+      const threadId = 'thread01';
+      const existingThread: NoteThreadDoc = env.getNoteThreadDoc('project01', threadId);
+      expect(existingThread.data!.notes.length).toEqual(3);
+      const note: Note = existingThread.data!.notes[2];
+      env.openNoteDialogAndEdit('verse_1_1', existingThread.data!, note, 'deleted note', true);
+      expect(existingThread.data!.notes.length).toEqual(2);
+      env.dispose();
+    }));
   });
 
   describe('Translation Suggestions disabled', () => {
@@ -3116,7 +3144,7 @@ class TestEnvironment {
     return thread != null;
   }
 
-  openNoteDialogAndEdit(segmentRef: string, thread: NoteThread, note: Note, content: string): void {
+  openNoteDialogAndEdit(segmentRef: string, thread: NoteThread, note: Note, content: string, deleted?: boolean): void {
     const iconElement: HTMLElement = this.getNoteThreadIconElement(segmentRef, thread.dataId)!;
     iconElement.click();
     this.wait();
@@ -3126,7 +3154,8 @@ class TestEnvironment {
       verseRef: thread.verseRef,
       note,
       position: thread.position,
-      selectedText: thread.originalSelectedText
+      selectedText: thread.originalSelectedText,
+      deleted
     });
     this.wait();
     const noteDialogData: NoteDialogData = {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
@@ -855,7 +855,8 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
 
   private async handleNoteDialogResult(result: NoteDialogResult): Promise<void> {
     const projectId: string = this.projectDoc!.id;
-    if (result.note.threadId === '') {
+
+    if (result.note.threadId === '' && result.verseRef != null) {
       // create a new thread
       const threadId: string = SF_NOTE_THREAD_PREFIX + objectId();
       result.note.threadId = threadId;
@@ -879,16 +880,33 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
     const noteThreadDoc: NoteThreadDoc | undefined = this.noteThreadQuery?.docs.find(
       nt => nt.data?.dataId === result.note.threadId
     );
-    if (noteThreadDoc?.data != null) {
-      const noteIndex: number = noteThreadDoc.data.notes.findIndex(n => n.dataId === result.note.dataId);
-      if (noteIndex >= 0) {
-        noteThreadDoc.submitJson0Op(op => {
-          op.set(t => t.notes[noteIndex].content, result.note.content);
-          op.set(t => t.notes[noteIndex].dateModified, result.note.dateModified);
-        });
+    if (noteThreadDoc?.data == null) {
+      return;
+    }
+
+    if (result.deleted) {
+      // delete the note or the entire thread
+      if (noteThreadDoc.data.notes.length === 1 && noteThreadDoc.data.notes[0].dataId === result.note.dataId) {
+        // only delete the thread if deleting the last note in the thread
+        await noteThreadDoc.delete();
       } else {
-        noteThreadDoc.submitJson0Op(op => op.add(t => t.notes, result.note));
+        const noteIndex: number = noteThreadDoc.data.notes.findIndex(n => n.dataId === result.note.dataId);
+        if (noteIndex >= 0) {
+          noteThreadDoc.submitJson0Op(op => op.remove(nt => nt.notes, noteIndex));
+        }
       }
+      return;
+    }
+
+    // updated the existing note
+    const noteIndex: number = noteThreadDoc.data.notes.findIndex(n => n.dataId === result.note.dataId);
+    if (noteIndex >= 0) {
+      noteThreadDoc.submitJson0Op(op => {
+        op.set(t => t.notes[noteIndex].content, result.note.content);
+        op.set(t => t.notes[noteIndex].dateModified, result.note.dateModified);
+      });
+    } else {
+      noteThreadDoc.submitJson0Op(op => op.add(t => t.notes, result.note));
     }
   }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
@@ -824,12 +824,6 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
       await this.handleNoteDialogResult(result);
       this.toggleNoteThreadVerses(true);
     }
-    // Ensure cursor selection remains at the position of the note in case the focus was lost when the dialog was open
-    const currentSegment: string | undefined = this.target!.currentSegmentOrDefault;
-    if (currentSegment != null) {
-      const selectIndex = this.target!.getSegmentRange(currentSegment)!.index;
-      this.target!.editor!.setSelection(selectIndex, 0, 'silent');
-    }
   }
 
   private updateReadNotes(threadId: string) {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.spec.ts
@@ -520,10 +520,12 @@ describe('NoteDialogComponent', () => {
   it('allows user to delete the last note in the thread', fakeAsync(() => {
     env = new TestEnvironment({ noteThread: TestEnvironment.getNoteThread() });
     expect(env.notes.length).toEqual(4);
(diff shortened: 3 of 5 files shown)


Nateowami (commented):
I looked into why the test is failing, and it's the Chromium 105 failure. If you rebase `feature/create-note` on master, then the builds should start to pass.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-:-NBsS3vL9Sexb3kb-dNg:bmb27f1)* status: 0 of 8 files reviewed, all discussions resolved

Nateowami (changes requested):
I reviewed everything except the tests. Just a couple questions and suggestions.

I tested this locally, and you can delete the latest note, and then there's no more delete button. But if you then close the dialog and open it again, there's an option to delete the last note again. Is that just an incomplete feature that will be addressed in a future PR, or is that something that's missing from this one?

> Reviewed 5 of 8 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-:-NDU4RKTBHvnzRXn_Sed:b-xp8x4d)* status: 5 of 8 files reviewed, 5 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts` line 229 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-NDV2Ft582sysfWKEOg1:-NDV2Ft582sysfWKEOg2:bf8c3iu) ([raw file](https://github.com/sillsdev/web-xforge/blob/e0e3c1d5008f163106e21798d23c640947d8c182/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L229)):*
> ```XML
>     const confirm: Observable<string> = this.i18n.translate('note_dialog.confirm');
>     const confirmed: boolean = await this.dialogService.confirm(message, confirm);
>     if (confirmed && this.threadDoc?.data != null) {
> ```

When reading this line, one of the first things I wonder is what happens if this condition is false. In order to find out, I have to find the matching curly brace, and see if anything comes after it (such as an `else if`, `else`, or just more lines of code). Doing so, I see that nothing happens if this condition fails.

If this was written as:
``` ts
if (!confirmed || this.threadDoc?.data == null) {
  return;
}
```
it would tell the reader immediately that this condition failing means the entire method completes, and everything that comes after this is dependent on this condition being true. There are no additional lines that will execute when this is false.

If this were a very short method like:
``` ts
if (a) {
  doB();
}
```
I probably wouldn't want it to be written as:
``` ts
if (!a) {
  return;
}
doB();
```
since that's less readable.

Maybe there's a countervailing argument I'm not thinking of? I just know that when I read this, I can't immediately comprehend it because I have to search for the matching curly brace, which is hard.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts` line 229 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-NDV6mK09cfpxLt56tQY:-NDV6mK09cfpxLt56tQZ:b-bivu05) ([raw file](https://github.com/sillsdev/web-xforge/blob/e0e3c1d5008f163106e21798d23c640947d8c182/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L229)):*
> ```XML
>     const confirm: Observable<string> = this.i18n.translate('note_dialog.confirm');
>     const confirmed: boolean = await this.dialogService.confirm(message, confirm);
>     if (confirmed && this.threadDoc?.data != null) {
> ```

What about when `this.threadDoc?.data == null`? From what I can tell, this should never happen, so it would make more sense to assert that, than to silently do nothing.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts` line 232 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-NDV8KM9CWTISeARHgOY:-NDV8KM9CWTISeARHgOZ:b-8ylb3p) ([raw file](https://github.com/sillsdev/web-xforge/blob/e0e3c1d5008f163106e21798d23c640947d8c182/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L232)):*
> ```XML
>       if (this.notesToDisplay.length === 1) {
>         // This may be the last note in the thread, so the thread may need to be deleted
>         this.dialogRef.close({ note, selectedText: '', position: { start: 0, length: 0 }, deleted: true });
> ```

I'm guessing there's some reason all these thread actions (creating, editing, deleting) need to be delegated to the editor component, rather than being done within the note dialog component itself? It seems slightly odd to essentially send messages to the opener of the dialog to tell it what it needs to do to the note thread.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts` line 233 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-NDV15B71eOn1N-95nby:-NDV15B8CdqsXOE8FL7J:b-trbk1n) ([raw file](https://github.com/sillsdev/web-xforge/blob/e0e3c1d5008f163106e21798d23c640947d8c182/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L233)):*
> ```XML
>         // This may be the last note in the thread, so the thread may need to be deleted
>         this.dialogRef.close({ note, selectedText: '', position: { start: 0, length: 0 }, deleted: true });
>         return;
> ```

This is approximately the opposite complaint as above. How about doing:
``` ts
if (condition) {
  doThing();
}
else  {
  doOtherThing()
}
```
Instead of:
``` ts
if (condition) {
  doThing();
  return;
}
doOtherThing();
```
With an `if`/`else` it's pretty clear that there are two mutually exclusive, collectively exhaustive, possibilities. It's still possible to understand this by reading line by line, but the higher-level logic (do thing a, or do thing b) is partially hidden within thing a.

As above, since this is a stylistic argument that comes up because of how I read through it, there could be a countervailing argument I'm not thinking of. I just know how this looks to me when I read and try to understand the logic.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 151 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-ND4Vc0E2ILliBRKOxUA:-ND4Vc0E2ILliBRKOxUB:b-662ue6) ([raw file](https://github.com/sillsdev/web-xforge/blob/594050119d4c3921180125ce382a951ca6bdbc5d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L151)):*
> ```JSON
>     "acceptance_change_heading": "Changes Paratext Accepted (current text):",
>     "close": "Close",
>     "confirm": "Confirm",
> ```

How about "Delete"?

RaymondLuong3 (commented):
Good catch, I'll fix it here in this PR.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-:-NDZ-bgu4eSvnDOo1Arw:bag8kz2)* status: 3 of 8 files reviewed, 5 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts` line 229 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-NDV2Ft582sysfWKEOg1:-NDZ0f9q7ODETmd8ipeT:b-ystrx0) ([raw file](https://github.com/sillsdev/web-xforge/blob/e0e3c1d5008f163106e21798d23c640947d8c182/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L229)):*

When reading this line, one of the first things I wonder is what happens if this condition is false. In order to find out, I have to find the matching curly brace, and see if anything comes after it (such as an `else if`, `else`, or just more lines of code). Doing so, I see that nothing happens if this condition fails.

If this was written as:
``` ts
if (!confirmed || this.threadDoc?.data == null) {
  return;
}
```
it would tell the reader immediately that this condition failing means the entire method completes, and everything that comes after this is dependent on this condition being true. There are no additional lines that will execute when this is false.

If this were a very short method like:
``` ts
if (a) {
  doB();
}
```
I probably wouldn't want it to be written as:
``` ts
if (!a) {
  return;
}
doB();
```
since that's less readable.

Maybe there's a countervailing argument I'm not thinking of? I just know that when I read this, I can't immediately comprehend it because I have to search for the matching curly brace, which is hard.
</blockquote></details>

Good thinking. I typically use that logic, but I missed it here.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts` line 229 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-NDV6mK09cfpxLt56tQY:-NDZ0lrm2SlBO-Z3YP0W:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/e0e3c1d5008f163106e21798d23c640947d8c182/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L229)):*

What about when `this.threadDoc?.data == null`? From what I can tell, this should never happen, so it would make more sense to assert that, than to silently do nothing.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts` line 232 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-NDV8KM9CWTISeARHgOY:-NDZ0pFEE9r4SNBO3lRJ:b2s2gdo) ([raw file](https://github.com/sillsdev/web-xforge/blob/e0e3c1d5008f163106e21798d23c640947d8c182/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L232)):*

I'm guessing there's some reason all these thread actions (creating, editing, deleting) need to be delegated to the editor component, rather than being done within the note dialog component itself? It seems slightly odd to essentially send messages to the opener of the dialog to tell it what it needs to do to the note thread.
</blockquote></details>

I plan to change that in a following PR. I wanted to get the functionality in, which seems to be almost done. And now we can do some refactoring.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts` line 233 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-NDV15B71eOn1N-95nby:-NDZ173A9fJmCccpiS2d:b-o9dby6) ([raw file](https://github.com/sillsdev/web-xforge/blob/e0e3c1d5008f163106e21798d23c640947d8c182/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L233)):*

This is approximately the opposite complaint as above. How about doing:
``` ts
if (condition) {
  doThing();
}
else  {
  doOtherThing()
}
```
Instead of:
``` ts
if (condition) {
  doThing();
  return;
}
doOtherThing();
```
With an `if`/`else` it's pretty clear that there are two mutually exclusive, collectively exhaustive, possibilities. It's still possible to understand this by reading line by line, but the higher-level logic (do thing a, or do thing b) is partially hidden within thing a.

As above, since this is a stylistic argument that comes up because of how I read through it, there could be a countervailing argument I'm not thinking of. I just know how this looks to me when I read and try to understand the logic.
</blockquote></details>

I like using else for the readability too, and the mutually exclusive argument is valid. Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 151 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-ND4Vc0E2ILliBRKOxUA:-NDZ1Fm0DED260ewoJ_J:bhqqiqv) ([raw file](https://github.com/sillsdev/web-xforge/blob/594050119d4c3921180125ce382a951ca6bdbc5d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L151)):*

How about "Delete"?
</blockquote></details>

Yes, I like it.

Nateowami (commented):
One question. Still reviewing the tests. 

> Reviewed 1 of 2 files at r4, 2 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-:-NDZ7n5j05ieb2pSHkAE:beikf6n)* status: 6 of 8 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 827 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-NDZ7c8s3hI-stuw8rtM:-NDZ7c8s3hI-stuw8rtN:b-hyoy0j) ([raw file](https://github.com/sillsdev/web-xforge/blob/594050119d4c3921180125ce382a951ca6bdbc5d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L827)):*
> ```XML
>       this.toggleNoteThreadVerses(true);
>     }
>     // Ensure cursor selection remains at the position of the note in case the focus was lost when the dialog was open
> ```

Why was this removed?

Nateowami (commented):
> Reviewed 1 of 2 files at r4, 1 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1496)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1496)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 827 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1496#-NDZ7c8s3hI-stuw8rtM:-NDZR2-_2o9fVkntQSP2:bn78acr) ([raw file](https://github.com/sillsdev/web-xforge/blob/594050119d4c3921180125ce382a951ca6bdbc5d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L827)):*

Why was this removed?
</blockquote></details>

This was unfortunately moving the selection to the beginning of the segment, and it didn't appear that it was needed since the cursor reappears in the correct position when the user closes the dialog.

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1496)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1499, opened 2022-09-16
SF-1702 Improve support for other footnote callers

nigel-wells:
- Allow any time of note to be clicked
- Added support for hidden notes
- Use cross character for cross-references when a custom caller is not used
- Improved support for customer callers that are more than 1 character

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1499)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
@@ -142,10 +142,6 @@ quill-editor.rtl {
   > .ql-editor {
     background-color: #f5f3ef;
 
-    *[contenteditable='false'] {
-      color: #7c7c7c;
-    }
-
     usx-para,
     p {
       margin-bottom: 1em;
--- a/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
@@ -351,6 +351,7 @@ usx-chapter {
   font-size: 1.5em;
   margin-top: 8pt;
   margin-bottom: 4pt;
+  color: #7c7c7c;
 }
 
 usx-verse {
@@ -366,6 +367,7 @@ usx-verse {
     border-radius: 8px;
     white-space: nowrap;
     margin: 0 0.2em 0 0.3em;
+    color: #7c7c7c;
     span[data-style='va'] {
       font-size: 65%;
       color: #5faa5f;
--- a/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
@@ -385,48 +385,41 @@ usx-note {
     vertical-align: baseline;
   }
 
-  &[data-caller='+'] > span::before,
-  &[data-caller]:not([data-caller='+']):not([data-caller='-']) > span {
+  > span {
     background: #000;
     color: white;
-    border-radius: 50%;
-    width: 1.25em;
+    border-radius: 1em;
+    min-width: 1.25em;
     height: 1.25em;
     display: inline-block;
     line-height: 1.25em;
     text-align: center;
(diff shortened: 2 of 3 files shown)


Nateowami (changes requested):
Just a couple issues. The color is grayish on my machine, unlike your screenshots. The border radius makes it into an oval, rather than rounded corners. If you set the border radius to a fixed value (e.g. 5em) rather than a percentage you can get consistent rounded corners. The height is set to `1.25em`, so we could set the the border radius to `0.625em`. Or we could just set it to `5em`, and the browser will limit the actual border radius to 50% of the smaller of the width or height (in order case, height is equal or less than width, so the border radius will end up being half the height).



> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1499#-:-NC6jUXvCYIZNkgWk54M:b-8zk81q)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`-- commits` line 6 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1499#-NC7UX55Aqlr6AkKKJWI):*
"Type", not "time"



_Code quote:_
```
time
```

nigel-wells (commented):
Good call, I thought it didn't look too bad in my screenshot but it does get worse if there are more characters. I've changed to `1em` and it looks better:


Regarding the gray colour, didn't we conclude that there was a setting either in your browser or your monitor that was causing colors to be off? I can't recall what the other issue was but I'm pretty sure it was only happening to you :) My experience with that happening to me is the incorrect color profile has been set for my monitor in Windows settings - not sure what or where it would be in Linux.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1499#-:-NC8JM6-59JUP4E9kyMd:b-5kqrw)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @nigel-wells)

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1499)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`-- commits` line 6 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1499#-NC7UX55Aqlr6AkKKJWI):*

"Type", not "time"
</blockquote></details>

Done.

Nateowami (changes requested):
The text color still seems to be off: 

I don't know why we wouldn't both be seeing the same thing. It clearly doesn't look the same in your screenshots.

Here's the rule applying the text color: 
``` scss
// _text.scss
.template-editor > .ql-container > .ql-editor *[contenteditable=false] {
    color: #7c7c7c;
}
```

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1499#-:-NCLMoTRD2WuVNio6KTx:b-j2zflk)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

nigel-wells (commented):
I've removed that style and added the colour to the chapter and verse as discussed. The testers can take a better look to see if it might need to be applied to other elements as well.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1499#-:-NCNmCJbBdPKwzQaw6qY:bwgcqgw)* status: 1 of 3 files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1499)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

Nateowami (approved):
> Reviewed 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1499)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

PR #1502, opened 2022-09-16
Fix type guards in connect project component

Nateowami:
We recently updated the version of TypeScript we're using. There is now an option called `useUnknownInCatchVariables`. See https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables

> In TypeScript 4.0, support was added to allow changing the type of the variable in a catch clause from `any` to `unknown`. Allowing for code like:
``` ts
try {
  // ...
} catch (err) {
  // We have to verify err is an
  // error before using it as one.
  if (err instanceof Error) {
    console.log(err.message);
  }
}
```
> This pattern ensures that error handling code becomes more comprehensive because you cannot guarantee that the object being thrown _is_ a Error subclass ahead of time. With the flag `useUnknownInCatchVariables` enabled, then you do not need the additional syntax (`: unknown`) nor a linter rule to try enforce this behavior.

In strict mode, which we have enabled, `useUnknownInCatchVariables` defaults to true. I've temporarily set it to false in `tsconfig.json` in order to make the upgrade process smoother. This PR fixes one place that would fail to build once `useUnknownInCatchVariables` is set to true.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1502)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts
@@ -6,6 +6,7 @@ import { DataLoadingComponent } from 'xforge-common/data-loading-component';
 import { I18nService } from 'xforge-common/i18n.service';
 import { NoticeService } from 'xforge-common/notice.service';
 import { PwaService } from 'xforge-common/pwa.service';
+import { hasStringProp } from '../../type-utils';
 import { ParatextProject } from '../core/models/paratext-project';
 import { SFProjectCreateSettings } from '../core/models/sf-project-create-settings';
 import { SFProjectDoc } from '../core/models/sf-project-doc';
@@ -188,8 +189,8 @@ export class ConnectProjectComponent extends DataLoadingComponent implements OnI
       let projectId: string = '';
       try {
         projectId = await this.projectService.onlineCreate(settings);
-      } catch (err) {
-        if (!err.message?.includes(ConnectProjectComponent.errorAlreadyConnectedKey)) {
+      } catch (err: unknown) {
+        if (!hasStringProp(err, 'message') || !err.message.includes(ConnectProjectComponent.errorAlreadyConnectedKey)) {
           throw err;
         }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/drag-and-drop.ts
@@ -1,5 +1,5 @@
 import Quill, { DeltaOperation, RangeStatic, StringMap } from 'quill';
-import { CaretPosition, hasFunctionProp } from '../../../utils';
+import { CaretPosition, hasFunctionProp } from '../../../type-utils';
 import { Delta } from '../../core/models/text-doc';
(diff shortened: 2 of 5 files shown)


RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1502)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts` line 193 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1502#-NCRqD-7-LalejSDSdDo:-NCRqD-7-LalejSDSdDp:bb2ab3c) ([raw file](https://github.com/sillsdev/web-xforge/blob/cdcdc987abfa3517184abf6142950a7305b999c9/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts#L193)):*
> ```XML
>         projectId = await this.projectService.onlineCreate(settings);
>       } catch (err: unknown) {
>         if (!hasStringProp(err, 'message') || !err.message.includes(ConnectProjectComponent.errorAlreadyConnectedKey)) {
> ```

Was there a reason we need to detect that `message` is a property of type string? Or is that simply the preferred way to do this?



_Code quote:_
```XML
!hasStringProp(err, 'message')
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1502)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts` line 193 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1502#-NCRqD-7-LalejSDSdDo:-NCRqwlB5DuOi4lAZn8R:b-elg3re) ([raw file](https://github.com/sillsdev/web-xforge/blob/cdcdc987abfa3517184abf6142950a7305b999c9/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts#L193)):*

Was there a reason we need to detect that `message` is a property of type string? Or is that simply the preferred way to do this?
</blockquote></details>

If we only checked for the existence of the property, and not that it was a string, then `err.message.includes()` could potentially fail, since we couldn't be sure there would be an `includes()` method.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1502)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts` line 193 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1502#-NCRqD-7-LalejSDSdDo:-NCRwAY0AvmWrExTFHQ0:b-4cj0dj) ([raw file](https://github.com/sillsdev/web-xforge/blob/cdcdc987abfa3517184abf6142950a7305b999c9/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts#L193)):*

If we only checked for the existence of the property, and not that it was a string, then `err.message.includes()` could potentially fail, since we couldn't be sure there would be an `includes()` method.
</blockquote></details>

Right, good thinking then.

PR #1506, opened 2022-09-23
SF-1610 Larger cursor appears when presence is on a blank heading

nigel-wells:
- Change usx-blank to inline-flex to avoid stacking child elements

This issue is only obvious if the Paratext font size has been increased to something large i.e. 20

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1506)

PR #1507, opened 2022-09-23
SF-1750 Fix multiple browsers causing infinite writing of currentProjectId

Nateowami:
This is probably only a partial fix.

I can't even test #1506 because of how pervasive this bug is. I don't know what all it takes to reproduce it, but logging in as the same user in two different browsers does it for me. Memory usage skyrockets as each browser keeps fighting to write currentProjectId to the user.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1507)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/user.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/user.service.ts
@@ -48,6 +48,9 @@ export class UserService {
 
   async setCurrentProjectId(user: UserDoc, value: string | undefined): Promise<void> {
     this.localSettings.set(CURRENT_PROJECT_ID_SETTING, value);
+    if (user.data?.sites[this.siteId].currentProjectId === value) {
+      return;
+    }
     await user.submitJson0Op(update => {
       if (value != null) {
         update.set(u => u.sites[this.siteId].currentProjectId, value);


nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1507)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/user.service.ts` line 50 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1507#-NCcrZWG7xA5GSSY_daM:-NCcrZWG7xA5GSSY_daN:b6vhdml) ([raw file](https://github.com/sillsdev/web-xforge/blob/d93b73aac91251100c111ad6154569f8c39bdad3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/user.service.ts#L50)):*
> ```XML
> 
>   async setCurrentProjectId(user: UserDoc, value: string | undefined): Promise<void> {
>     this.localSettings.set(CURRENT_PROJECT_ID_SETTING, value);
> ```

Should the local storage be ignored as well?

nigel-wells (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1507)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1507)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/user.service.ts` line 50 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1507#-NCcrZWG7xA5GSSY_daM:-NCe_N1F8P1s9sPlButx:b-g5qte9) ([raw file](https://github.com/sillsdev/web-xforge/blob/d93b73aac91251100c111ad6154569f8c39bdad3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/user.service.ts#L50)):*

Should the local storage be ignored as well?
</blockquote></details>

I don't think it's necessary, and it would need to be in its own separate check, since the local storage value is not necessarily the same as the value on the user object.

The reason for the recursiveness bug is because the app component calls this method when the user is updated, even when nothing needs to be changed.

PR #1510, opened 2022-09-26
SF-1753 Replace message dialog with generic dialog

Nateowami:
I also changed `setLocale` and `trySetLocale` in `I18nService` to take an `AuthService`. Now the `AuthService` is not needed at the time the constructor is called. This fixed a circular dependency where `AuthService` needed `DialogService`, `DialogService` needed `I18nService`, and `I18nService` needed `AuthService`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1510)


RaymondLuong3 (changes requested):
Nice work. I just have the one blocking comment

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1510#-:-ND4Y96A7yKyUTdMQIzm:bq9jpw9)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 353 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1510#-ND4WfAG4q19x9cq1Ffx:-ND4WfAG4q19x9cq1Ffy:bjjt349) ([raw file](https://github.com/sillsdev/web-xforge/blob/1f9ca23788120cafbacc6ef6a9be022cd8d5042e/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L353)):*
> ```XML
>         this.dialogService
>           .message(
>             this.i18n.translate('connect_project.paratext_account_linked_to_another_user', {
> ```

Is there a reason you like `i18n.translate` over `translate` from `@ngneat/transloco`? Mostly out of curiousity.



_Code quote:_
```XML
            this.i18n.translate('connect_project.paratext_account_linked_to_another_user', {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 142 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1510#-ND4XHGB8hee1FWnujL8:-ND4XHGB8hee1FWnujL9:bk7ieqw) ([raw file](https://github.com/sillsdev/web-xforge/blob/1f9ca23788120cafbacc6ef6a9be022cd8d5042e/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L142)):*
> ```XML
>   }
> 
>   setLocale(tag: string, authService: AuthService) {
> ```

Instead of passing in the auth service, you could just pass in the `updateInterfaceLanguage` method. Theoretically that would work.



_Code quote:_
```XML
  setLocale(tag: string, authService: AuthService) {
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1510)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 353 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1510#-ND4WfAG4q19x9cq1Ffx:-ND4cJ1i9aaa1GfZdRnn:b-uxo3o7) ([raw file](https://github.com/sillsdev/web-xforge/blob/1f9ca23788120cafbacc6ef6a9be022cd8d5042e/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L353)):*

Is there a reason you like `i18n.translate` over `translate` from `@ngneat/transloco`? Mostly out of curiousity.
</blockquote></details>

Dependency injection, mainly. We've had some issues with testing components that use the `translate()` function directly. Also, the I18nService extends the capabilities of Transloco. For example, it can localize a book, or a reference, or a role, or format a date, and more. If a component needs that, then Transloco isn't sufficient.

In general I think it's preferable to abstract away systems like Transloco. I don't foresee moving away from Transloco, but if we ever did, having the I18nService used throughout our app would make it easier*. Before too long we're planning to improve our analytics, by moving to Google Analytics 4. Rather than calling Google Analytics throughout our app, I'm guessing we'll use an analytics service. That way if we ever decide to switch to another system, or also log some analytics to our database, or Bugsnag, we only have to change the logic in one place.

<sup>* Well, we use `t(...)` all over out templates, so maybe not that much easier.</sup>
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 142 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1510#-ND4XHGB8hee1FWnujL8:-ND4f164CE7HhWqb4c7f:b6ah0ff) ([raw file](https://github.com/sillsdev/web-xforge/blob/1f9ca23788120cafbacc6ef6a9be022cd8d5042e/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L142)):*

Instead of passing in the auth service, you could just pass in the `updateInterfaceLanguage` method. Theoretically that would work.
</blockquote></details>

You can't generally pass around methods in JavaScript like you can functions. If you update the method to take `updateInterfaceLanguage` itself, and then do:

``` ts
this.i18n.trySetLocale(languageTag, this.authService.updateInterfaceLanguage);
```
You'll get an error:
```
TypeError: Cannot read properties of undefined (reading 'isLoggedIn')
    at auth.service.ts:230:20
```
That error happens here:
``` ts
async updateInterfaceLanguage(language: string): Promise<void> {
  if (await this.isLoggedIn) { // on this line, this is undefined
    await this.commandService.onlineInvoke(USERS_URL, 'updateInterfaceLanguage', { language });
  }
}
```
You can work around this by binding the function to the object:
``` ts
 this.i18n.trySetLocale(languageTag, this.authService.updateInterfaceLanguage.bind(this.authService));
```
This is a good workaround if you have to pass a method to code you don't control, but in our case it's the more complicated and error-prone option.

Side note: Really JavaScript only has functions, not methods. If an object has a function property, and you call `obj.someFunction()`, then within that function call, the value of `this` will be `obj`. If you do `const x = obj.someFunction(); x()`, then the function is called with `this` set to `undefined`. This is only true in strict mode (which TypeScript turns on for us). Non-strict mode behaves differently and only exists for compatibility with old code.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1510)* status: 14 of 16 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 142 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1510#-ND4XHGB8hee1FWnujL8:-ND4tcD9A4tKrK6Aj3Dj:brrfjy) ([raw file](https://github.com/sillsdev/web-xforge/blob/1f9ca23788120cafbacc6ef6a9be022cd8d5042e/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L142)):*
I made a mistake in one of the last code examples. This:
> If you do `const x = obj.someFunction(); x()`, then the function is called with `this` set to `undefined`.

Should say:
> If you do `const x = obj.someFunction; x()`, then the function is called with `this` set to `undefined`.

RaymondLuong3 (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1510#-:-ND4zumS-hl5wM7S5NUM:bg7igkv)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 353 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1510#-ND4WfAG4q19x9cq1Ffx:-ND4ze6-3WC_ScUgnRm2:b44183h) ([raw file](https://github.com/sillsdev/web-xforge/blob/1f9ca23788120cafbacc6ef6a9be022cd8d5042e/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L353)):*

Dependency injection, mainly. We've had some issues with testing components that use the `translate()` function directly. Also, the I18nService extends the capabilities of Transloco. For example, it can localize a book, or a reference, or a role, or format a date, and more. If a component needs that, then Transloco isn't sufficient.

In general I think it's preferable to abstract away systems like Transloco. I don't foresee moving away from Transloco, but if we ever did, having the I18nService used throughout our app would make it easier*. Before too long we're planning to improve our analytics, by moving to Google Analytics 4. Rather than calling Google Analytics throughout our app, I'm guessing we'll use an analytics service. That way if we ever decide to switch to another system, or also log some analytics to our database, or Bugsnag, we only have to change the logic in one place.

<sup>* Well, we use `t(...)` all over out templates, so maybe not that much easier.</sup>
</blockquote></details>

That's a good point that abstracting `translate` within the `i18n.service` is a good way to go in case we ever do want to change translation libraries.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 142 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1510#-ND4XHGB8hee1FWnujL8:-ND4zOpVCEUbu7_EJCc5:b-g2uagj) ([raw file](https://github.com/sillsdev/web-xforge/blob/1f9ca23788120cafbacc6ef6a9be022cd8d5042e/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L142)):*

I made a mistake in one of the last code examples. This:
> If you do `const x = obj.someFunction(); x()`, then the function is called with `this` set to `undefined`.

Should say:
> If you do `const x = obj.someFunction; x()`, then the function is called with `this` set to `undefined`.
</blockquote></details>

Thanks for the thorough explanation. I was getting functions and methods mixed up. This would work if `updateInterfaceLanguage` was strictly a function, which it is not.

PR #1515, opened 2022-09-27
SF-1743a Stop sync if repo mismatch

marksvc:
- SF cleans up at the end of the sync process. If the sync process is
abruptly terminated without being able to clean up, the sync process
makes poor assumptions next time about its state. Namely, it can think
that the data in sync.
- A better and future change can address checking and cleaning up at
the beginning of a sync. This change merely checks if we are in a bad
state and stops the sync.
- If the PT hg repo commit revision does not match the SF DB record of
what PT hg repo commit revision the project was syncronized to, stop
the sync. Except for a special case.
- Clearly there are code cleanup activities that can be done here.
This change prioritizes getting the behaviour changed.

Tests file:
- The test BackupRestoredPreviouslyRevNotMatching_WritesToPT seems to
request the bad behaviour. The implementation code may have moved on
since this time. I spoke with the author. I removed the test.
- Some tests were specifically requesting that we _do_ run SendReceive
even in a situation where we are out of sync. This expectation is
reversed with this change, where the sync should quit before calling
SendReceive.

commit-id:a5069a6a

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1515)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
@@ -138,6 +138,24 @@ namespace SIL.XForge.Scripture.Services
                 var targetTextDocsByBook = new Dictionary<int, SortedList<int, IDocument<TextData>>>();
                 var questionDocsByBook = new Dictionary<int, IReadOnlyList<IDocument<Question>>>();
                 string lastSharedVersion = _paratextService.GetLatestSharedVersion(_userSecret, targetParatextId);
+                if (lastSharedVersion != _projectDoc.Data.Sync.SyncedToRepositoryVersion)
+                {
+                    if (
+                        _projectDoc.Data.Sync.SyncedToRepositoryVersion == null
+                        && _projectDoc.Data.Sync.DataInSync == null
+                    )
+                    {
+                        // This project pre-dates when we started tracking this information, and does not yet have a record of it.
+                    }
+                    else
+                    {
+                        Log(
+                            $"RunAsync: The reported PT hg repo latest shared version '{lastSharedVersion}' does not match record of project SyncedToRepositoryVersion '{_projectDoc.Data.Sync.SyncedToRepositoryVersion}'. Refusing to continue, to protect data."
+                        );
+                        await CompleteSync(successful: false, canRollbackParatext, trainEngine, token);
+                        return;
+                    }
+                }
 
                 bool isDataInSync = _projectDoc.Data.Sync.DataInSync ?? false;
                 if (lastSharedVersion == null)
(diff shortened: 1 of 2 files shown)


jasonleenaylor (commented):
Maybe I missed it but it might be good to clarify in the code or tests how the first sync of a project after this update will manage to succeed.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1515#-:-ND-pZA5C6htdwXKYk6M:bnsogzl)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @jasonleenaylor)

___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs` line 1677 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1515#-ND-pviwDvuQCzcF4KK7:-ND-pviwDvuQCzcF4KK8:bw858rm) ([raw file](https://github.com/sillsdev/web-xforge/blob/26cf6e077c21c3c7501c0d6ba77b2367601c9957/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L1677)):*
> ```Smalltalk
>         {
>             // Note that there are multiple constructions of TestEnvironment, since some mocks return different
>             // values when called more than once.
> ```

an alternative would be to clear the mocks, not sure which is better.



_Code quote:_
```Smalltalk
values when called more than once.
```

Nateowami (approved):
:lgtm:

> Reviewed 1 of 2 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1515#-:-ND-zaV09-5h26PO4dd4:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

marksvc (commented):
Thanks Jason. That will be good to include :). When a project is first Connected, the ParatxtService.GetLatestSharedVersion() is null, and the SF DB SyncedToRepositoryVersion and DataInSync are also null. The new code checks and sees that null==null and so allows the Connect to continue. I agree that will be good to be more explanatory about in code. That can be included in some in-progress changes to improve this area.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1515#-:-ND00W9IETl2Qu-u8AGE:bfmchxv)* status: :shipit: complete! all files reviewed, all discussions resolved

___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs` line 1677 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1515#-ND-pviwDvuQCzcF4KK7:-ND036AF2etvjxIJNX8_:b-kc3tt2) ([raw file](https://github.com/sillsdev/web-xforge/blob/26cf6e077c21c3c7501c0d6ba77b2367601c9957/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L1677)):*

an alternative would be to clear the mocks, not sure which is better.
</blockquote></details>

Thank you. I didn't think of that. The constructor does 
```cs
ParatextService.When(someone calls SendReceiveAsync()).Do(Change the return value of GetLatestSharedVersion() (from what we were returning before));
```
And so it's a bit more complex than just needing to reset the mocks. I had a look but am planning to keep it as is for now. I am updating the comment in a follow up that makes it more clear what the need is.

PR #1520, opened 2022-09-29
Add some nullable annotations

marksvc:
**Stack**:
- #1521
- #1520 ⮜

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1520)

### DIFF
--- a/src/SIL.XForge.Scripture/Models/Sync.cs
+++ b/src/SIL.XForge.Scripture/Models/Sync.cs
@@ -1,3 +1,5 @@
+#nullable enable annotations
+
 using System;
 
 namespace SIL.XForge.Scripture.Models
@@ -13,7 +15,7 @@ namespace SIL.XForge.Scripture.Models
         /// The SF project was last successfully synchronized with PT project data at this repository
         /// commit on the PT project send/receive server.
         /// </summary>
-        public string SyncedToRepositoryVersion { get; set; }
+        public string? SyncedToRepositoryVersion { get; set; }
 
         /// <summary>
         /// If the local PT repo has been imported into SF DB.
--- a/src/SIL.XForge.Scripture/Services/IJwtTokenHelper.cs
+++ b/src/SIL.XForge.Scripture/Services/IJwtTokenHelper.cs
@@ -1,3 +1,4 @@
+#nullable enable annotations
 using System.Net.Http;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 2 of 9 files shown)


RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1520)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs` line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1520#-NDUatRiBfrAWnSMIZjJ:-NDUatRiBfrAWnSMIZjK:bgtro8o) ([raw file](https://github.com/sillsdev/web-xforge/blob/aa03439b3de15a0ca0a43c18fc7a809629364ae7/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L27)):*
> ```Smalltalk
> 
>         /// <summary> Get the Paratext username from the access token stored in the UserSecret. </summary>
>         public string? GetParatextUsername(UserSecret userSecret)
> ```

This makes it explicit whether a string is nullable? Is that correct. So in this case, the Paratext Username can be null.



_Code quote:_
```Smalltalk
 public string? GetParatextUsername(UserSecret userSecret)
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1520)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

Nateowami:
@marksvc And just after merging I realized this was created with spr and wasn't supposed to be merged.

marksvc (commented):
@nateowami Actually in this case it is fine. The second PR in the series was dependent on this one being merged first, but they weren't associated in such a way that they should be merged together, or not merged until all reviewed. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1520#-:-NEbOazwCto69kmMJk4C:b-3paip)* status: all files reviewed, 1 unresolved discussion

___
*[`src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs` line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1520#-NDUatRiBfrAWnSMIZjJ:-NEbO2qL6sXGiMU54T5N:b-rt2elv) ([raw file](https://github.com/sillsdev/web-xforge/blob/aa03439b3de15a0ca0a43c18fc7a809629364ae7/src/SIL.XForge.Scripture/Services/JwtTokenHelper.cs#L27)):*

This makes it explicit whether a string is nullable? Is that correct. So in this case, the Paratext Username can be null.
</blockquote></details>

Right! So type `string` means it will be a string but not null, and `string?` means it could be either a string or null.
The method returns `null` if, say, `userSecret == null`, and so we can specify that `GetParatextUsername()` returns `string?` to help callers know what to do with the potentially-null result.

PR #1521, opened 2022-09-29
SF-1743c sync: ensure correct revision is checked out when sync starts

marksvc:
- The SF synchronize process records if 'DataInSync' at the end. The
next time SF synchronizes, it considers 'DataInSync' as part of
determining if it is safe to write data. If SF is abruptly terminated
during synchronize, we may not set 'DataInSync' to false when we would
have, resulting in the next sync incorrectly thinking it can safely
write data.
- This commit changes the behaviour. At the beginning of each
synchronize, set the local hg repository to the revision that we know
we last successfully imported from, and consider that we are always
'in sync', and so can safely write data. This is to make the
synchronize process more resilient against problems that occur during
its operation.
- The current hg repo revision is now being logged, in addition to the
latest shared/public revision, to identify possible problems in how we
use these.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1521)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/HgWrapper.cs
+++ b/src/SIL.XForge.Scripture/Services/HgWrapper.cs
@@ -1,4 +1,6 @@
+#nullable enable annotations
 using System;
+using System.IO;
 using System.Linq;
 using Paratext.Data.Repository;
 
@@ -53,6 +55,19 @@ namespace SIL.XForge.Scripture.Services
         }
 
         /// <summary>
+        /// Returns the currently checked out revision of an hg repository.
+        /// </summary>
+        public string GetRepoRevision(string repositoryPath)
+        {
+            string rev = RunCommand(repositoryPath, "log --limit 1 --rev . --template {node}");
+            if (string.IsNullOrWhiteSpace(rev))
+            {
+                throw new InvalidDataException($"Unable to determine repo revision for hg repo at {repositoryPath}");
+            }
+            return rev;
+        }
+
+        /// <summary>
         /// Restores the repository.
         /// </summary>
         /// <param name="destination">The destination to restore to.</param>
@@ -94,6 +109,16 @@ namespace SIL.XForge.Scripture.Services
         public void Update(string repository)
         {
             Hg.Default.Update(repository);
+        }
+
+        /// <summary>
+        /// Set a hg repo's files to a particular revision in history. Similar to running
+        /// `git checkout --force --detach COMMITTISH`
+        /// Changes to tracked files will be discarded. Untracked files are left in place without being cleaned up.
+        /// </summary>
+        public void Update(string repositoryPath, string rev)
+        {
+            Hg.Default.Update(repositoryPath, rev);
         }
     }
 }
--- a/src/SIL.XForge.Scripture/Services/IHgWrapper.cs
+++ b/src/SIL.XForge.Scripture/Services/IHgWrapper.cs
@@ -7,9 +7,11 @@ namespace SIL.XForge.Scripture.Services
         void SetDefault(Hg hgDefault);
         void Init(string repository);
         void Update(string repository);
+        void Update(string repository, string rev);
         void BackupRepository(string repository, string backupFile);
         void RestoreRepository(string destination, string backupFile);
         string GetLastPublicRevision(string repository);
+        string GetRepoRevision(string repositoryPath);
         void MarkSharedChangeSetsPublic(string repository);
     }
 }
--- a/src/SIL.XForge.Scripture/Services/IParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/IParatextService.cs
@@ -69,9 +69,12 @@ namespace SIL.XForge.Scripture.Services
             Dictionary<string, ParatextUserProfile> ptProjectUsers
         );
         string? GetLatestSharedVersion(UserSecret userSecret, string paratextId);
+        string GetRepoRevision(UserSecret userSecret, string projectPTId);
+        void SetRepoToRevision(UserSecret userSecret, string projectPTId, string desiredRevision);
         bool BackupExists(UserSecret userSecret, string paratextId);
         bool BackupRepository(UserSecret userSecret, string paratextId);
         bool RestoreRepository(UserSecret userSecret, string paratextId);
+        bool LocalProjectDirExists(string projectPTId);
 
         Task<ParatextProject> SendReceiveAsync(
             UserSecret userSecret,
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -1328,6 +1328,38 @@ namespace SIL.XForge.Scripture.Services
             }
         }
 
+        /// <summary>Returns the current revision of the local hg repo for a given project PT id,
+        /// accessed using a given user.</summary>
+        public string GetRepoRevision(UserSecret userSecret, string projectPTId)
+        {
+            if (IsResource(projectPTId))
+            {
+                throw new InvalidOperationException("Cannot query a resource for an hg repo revision.");
+            }
+            using ScrText scrText = GetScrText(userSecret, projectPTId);
+            return _hgHelper.GetRepoRevision(scrText.Directory);
+        }
+
+        /// <summary>Set a project local hg repo to a given Mercurial revision. Accessed by a given user.</summary>>
+        public void SetRepoToRevision(UserSecret userSecret, string projectPTId, string desiredRevision)
+        {
+            if (IsResource(projectPTId))
+            {
+                throw new InvalidOperationException("Cannot query a resource for an hg repo revision.");
+            }
+            using ScrText scrText = GetScrText(userSecret, projectPTId);
+            // Act
+            _hgHelper.Update(scrText.Directory, desiredRevision);
+            // Verify
+            string currentRepoRev = GetRepoRevision(userSecret, projectPTId);
+            if (currentRepoRev != desiredRevision)
+            {
+                throw new Exception(
+                    $"SetRepoToRevision failed to set repo for project PT id {projectPTId} to revision {desiredRevision}, as the resulting revision is {currentRepoRev}."
+                );
+            }
+        }
+
         /// <summary>
         /// Checks whether a backup exists for the Paratext project repository.
         /// </summary>
@@ -1505,10 +1537,34 @@ namespace SIL.XForge.Scripture.Services
             return false;
         }
 
+        /// <summary>Does a local directory exist for the project? (i.e. in /var/lib/...)</summary>
+        public bool LocalProjectDirExists(string projectPTId)
+        {
+            string dir = LocalProjectDir(projectPTId);
+            return _fileSystemService.DirectoryExists(dir);
+        }
+
         protected override void DisposeManagedResources()
         {
             _registryClient.Dispose();
             _httpClientHandler.Dispose();
+        }
+
+        private ScrText GetScrText(UserSecret userSecret, string projectPTId)
+        {
+            string? ptUsername = GetParatextUsername(userSecret);
+            if (ptUsername == null)
+            {
+                throw new DataNotFoundException($"Failed to get username for UserSecret id {userSecret.Id}.");
+            }
+            ScrText? scrText = ScrTextCollection.FindById(GetParatextUsername(userSecret), projectPTId);
+            if (scrText == null)
+            {
+                throw new DataNotFoundException(
+                    $"Could not find project for UserSecret id {userSecret.Id}, project PT id {projectPTId}"
+                );
+            }
+            return scrText;
         }
 
         /// <summary>
@@ -1747,7 +1803,7 @@ namespace SIL.XForge.Scripture.Services
                 // Extract the resource to the source directory
                 if (needsToBeCloned)
                 {
-                    string path = Path.Combine(SyncDir, targetParatextId, "target");
+                    string path = LocalProjectDir(targetParatextId);
                     _fileSystemService.CreateDirectory(path);
                     resource.InstallableResource.ExtractToDirectory(path);
                 }
@@ -1788,9 +1844,19 @@ namespace SIL.XForge.Scripture.Services
             };
         }
 
-        private void CloneProjectRepo(IInternetSharedRepositorySource source, string projectId, SharedRepository repo)
+        /// <summary>Path for project directory on local filesystem. Whether or not that directory
+        /// exists. </summary>
+        private string LocalProjectDir(string projectPTId)
         {
-            string clonePath = Path.Combine(SyncDir, projectId, "target");
+            // Historically, SF used both "target" and "source" projects in adjacent directories. Then
+            // moved to just using "target".
+            string subDirForMainProject = "target";
+            return Path.Combine(SyncDir, projectPTId, subDirForMainProject);
+        }
+
+        private void CloneProjectRepo(IInternetSharedRepositorySource source, string projectPTId, SharedRepository repo)
+        {
+            string clonePath = LocalProjectDir(projectPTId);
             if (!_fileSystemService.DirectoryExists(clonePath))
             {
                 _fileSystemService.CreateDirectory(clonePath);
--- a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
@@ -24,9 +24,9 @@ namespace SIL.XForge.Scripture.Services
     /// <summary>
     /// This class syncs real-time text and question docs with the Paratext S/R server. This class coordinates the data
     /// flow between three sources:
-    /// 1. The real-time docs stored in Mongo.
+    /// 1. The real-time docs stored in Mongo, the SF DB.
     /// 2. The local Paratext project repo.
-    /// 3. The remote Paratext project repo.
+    /// 3. The remote Paratext project repo, aka PT Archives, aka the Send and Receive server.
     ///
     /// Algorithm:
     /// 1. The text deltas from the real-time docs are converted to USX.
@@ -43,6 +43,16 @@ namespace SIL.XForge.Scripture.Services
     ///
     /// Target and source refer to daughter and mother translation data. Not to be confused with a target or source for
     /// where data is coming from or going to when fetching or syncing.
+    /// <code>
+    /// Diagram showing a high-level look at the order of data being transmitted:
+    /// ┌─────┐  ┌─────┐  ┌───────────┐
+    /// │SF DB├─>│Local│  │PT Archives│
+    /// │     │  │PT hg├─>│           │
+    /// │     │  │repo │  │           │
+    /// │     │  │     │<─┤           │
+    /// │     │<─┤     │  │           │
+    /// └─────┘  └─────┘  └───────────┘
+    /// </code>
     /// </summary>
     public class ParatextSyncRunner : IParatextSyncRunner
     {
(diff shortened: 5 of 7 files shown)


RaymondLuong3 (changes requested):
Essentially, we set the local hg repo to where we know we have synchronized to, so even if we crash before we commit changes to the text docs (any becoming out of sync), we still always write data to the repo before send/receive, and this is ok because as long as the base is the expected one, Paratext will be able to merge local changes during send/receives so that we never get into the scenario of accidentally overwrite data. Is that correct.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-:-NEl3iErCTvSIXHn0pfx:b8p3l9r)* status: all files reviewed, 9 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/Services/HgWrapper.cs` line 120 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl6mhy7TvbvDWRuvA2:-NEl6mhy7TvbvDWRuvA3:b-3c2x95) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/HgWrapper.cs#L120)):*
> ```Smalltalk
>         public void Update(string repositoryPath, string rev)
>         {
>             Hg.Default.Update(repositoryPath, rev);
> ```

This will update even if there are uncommitted changes, is that correct?



_Code quote:_
```Smalltalk
            Hg.Default.Update(repositoryPath, rev);
```
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1333 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl7Px08kM56Iu6fPE8:-NEl7Px08kM56Iu6fPE9:b1vs4ae) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1333)):*
> ```Smalltalk
>         }
> 
>         /// <summary>Returns the currently revision of the hg repo that a user accesses for a project PT id.</summary>
> ```

Maybe a better name is Returns the current revision of the local hg repo for a given project id.



_Code quote:_
```Smalltalk
mary>Returns the currently revision of the hg repo that a user accesses for a project PT id.
```
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1338 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl7nHtBkYsnYhi9gxS:-NEl7nHtBkYsnYhi9gxT:b-dr6091) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1338)):*
> ```Smalltalk
>             if (IsResource(projectPTId))
>             {
>                 throw new InvalidOperationException("Can not query a resource for an hg repo revision.");
> ```

Is this because the hg repo is different than a project repo?
Typo: Cannot is one word.



_Code quote:_
```Smalltalk
"Can not query a resource for an hg repo revision.")
```
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1349 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl89Zp0hJbH46LPImp:-NEl89Zp0hJbH46LPImq:b-uh4lge) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1349)):*
> ```Smalltalk
>             if (IsResource(projectPTId))
>             {
>                 throw new InvalidOperationException("Can not query a resource for an hg repo revision.");
> ```

Typo: Cannot is one word.



_Code quote:_
```Smalltalk
         throw new InvalidOperationException("Can not query a resource for an hg repo revision.");
```
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1359 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl8M-8A914oAIaAR9M:-NEl8M-8A914oAIaAR9N:bbmv7o2) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1359)):*
> ```Smalltalk
>             {
>                 throw new Exception(
>                     $"SetRepoToRevision failed to set repo for project PT id {projectPTId} to revision {desiredRevision}, as resulting revision is {currentRepoRev}."
> ```

Do you mean "the" resulting revision?



_Code quote:_
```Smalltalk
}, as resulting revision 
```
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 49 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl9LoOD80tUU__2Hk7:-NEl9LoOD80tUU__2Hk8:b-4fwmme) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L49)):*
> ```Smalltalk
>     /// where data is coming from or going to when fetching or syncing.
>     /// <code>
>     /// Diagram showing a high-level look at the order of data being transmitted:
> ```

Nice diagram :)



_Code quote:_
```Smalltalk
   /// Diagram showing a high-level look at the order of data being transmitted:
```
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 178 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NElA3JxB0SBWyrG_N9U:-NElA3JxB0SBWyrG_N9V:brigeou) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L178)):*
> ```Smalltalk
>                         $"Setting hg repo to last imported hg repo rev of {_projectDoc.Data.Sync.SyncedToRepositoryVersion}."
>                     );
>                     this._paratextService.SetRepoToRevision(
> ```

Nit: No need for `this`



_Code quote:_
```Smalltalk
this._paratextService
```
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 1504 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NElA_jo5rl5bu8Z51MG:-NElA_jo5rl5bu8Z51MH:b-2xy9xe) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L1504)):*
> ```Smalltalk
>                 Log($"{prefix}In-sync info: Is resource. {dbInfo}");
>             }
>             else if (!this._paratextService.LocalProjectDirExists(projectPTId))
> ```

Nit: no need for `this`



_Code quote:_
```Smalltalk
this._paratextService.
```
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs` line 3165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NElBJna621CYYhBl3zd:-NElBJna621CYYhBl3ze:ba35bht) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L3165)):*
> ```Smalltalk
>             string projectPTId = env.SetupProject(env.Project01, associatedPtUser);
>             string rev = "some-desired-revision";
>             env.MockHgWrapper.GetRepoRevision(Arg.Any<string>()).Returns(rev);
> ```

You could use the repository path instead of Arg.Any<string> if you want to be extra sure.



_Code quote:_
```Smalltalk
 env.MockHgWrapper.GetRepoRevision(Arg.Any<string>()).Returns(rev);
```

marksvc (commented):
That's right. Each time we are going to synchronize data between SF and PT Archives, which happens thru the local hg repo, we first set the local hg repo to the known-good state. The known-good state is the state that it was in when we last imported from it. Because at the beginning of a Sync, the local hg repo is in the state that it was when we last imported from it, we will always start out our Sync as an  "in sync" situation, and can dump the SF DB text right into the local hg repo, which is effectively applying just SF changes. ParatextData.dll then uses hg to push the repo to PT Archives, and then it's up to PT Archives to decide how to merge and flag PT-conflicts. 

We'll find more things to fix, but I think this will be less likely to have problems.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-:-NElGo0T4fc3obkZzqQo:bm2pg1b)* status: all files reviewed, 7 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/Services/HgWrapper.cs` line 120 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl6mhy7TvbvDWRuvA2:-NElJ-IE3TPpr9uhTVF9:b-ezkl53) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/HgWrapper.cs#L120)):*
Right. And the uncommitted, tracked changes will be discarded. The API doc says it "Does an update, using --clean option". 
The hg docs say of `hg update` `--clean`, 
> With the -C/--clean option, uncommitted changes are discarded and the working directory is updated to the requested changeset.

Here is an example:
```bash
$ hg status
M existing-file.txt
? new-file.txt
$ hg update --clean 'head()'
1 files updated, 0 files merged, 0 files removed, 0 files unresolved
$ hg status
? new-file.txt
```
In this example, we see that the changes to the modified file, `existing-file.txt`, were discarded. We also see that the untracked file, new-file.txt, was left alone and not removed. 

It may be worth also deleting untracked files. I suspect that the potential for a problem from that is low, but it may be a good thing to do in followup.
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1333 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl7Px08kM56Iu6fPE8:-NElM-054i4iA3yMGG_o:bhhpfct) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1333)):*

Maybe a better name is Returns the current revision of the local hg repo for a given project id.
</blockquote></details>

Yes! I want to give a clue as to why  `userSecret` is one of the parameters. How about "Returns the current revision of the local hg repo for a given project PT id, accessed using a given user."?
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1338 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl7nHtBkYsnYhi9gxS:-NElOMNo7xU2gY0Gsi3U:bwrixjo) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1338)):*

Is this because the hg repo is different than a project repo?
Typo: Cannot is one word.
</blockquote></details>

It's because a resource _isn't_ stored in a local hg repo! There is no .hg folder. I chose to make `GetRepoRevision()` reject requests for resources so that the method can always return a hg revision, rather than returning special case results.
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1349 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl89Zp0hJbH46LPImp:-NElP3gW1cGdBgse69kK:b-vuj8ut) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1349)):*

Typo: Cannot is one word.
</blockquote></details>

Thank you :)
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1359 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl8M-8A914oAIaAR9M:-NElPGR_DeRltwIvIvts:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1359)):*

Do you mean "the" resulting revision?
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 49 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl9LoOD80tUU__2Hk7:-NElPJBU09eaoKFbsrgp:b-vuj8ut) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L49)):*

Nice diagram :)
</blockquote></details>

Thank you :)
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs` line 3165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NElBJna621CYYhBl3zd:-NElUHxt4BoTeYmOjwrH:b-jpslh8) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L3165)):*

You could use the repository path instead of Arg.Any<string> if you want to be extra sure.
</blockquote></details>

Done. I did the same to the following test as well, but then decided to leave it as is so that it more likely continues returning the specific bad value even if other parts of the system change (like if we we start using "blah" instead of "target" in the directory path).

RaymondLuong3 (approved):
How can we make it easy for the users to test this? In the past, I created a branch that contains code to make the sync fail in a particular way. Then I tell the testers to fail a sync using the branch. After the failure, I get the testers to checking out good branch. What do you think of that?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-:-NEvkqlyEcMlkCHgtgtS:b-3nm09g)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/Services/HgWrapper.cs` line 120 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl6mhy7TvbvDWRuvA2:-NEvlBHT9cpO373Ldup4:b-3jmj82) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/HgWrapper.cs#L120)):*

Right. And the uncommitted, tracked changes will be discarded. The API doc says it "Does an update, using --clean option". 
The hg docs say of `hg update` `--clean`, 
> With the -C/--clean option, uncommitted changes are discarded and the working directory is updated to the requested changeset.

Here is an example:
```bash
$ hg status
M existing-file.txt
? new-file.txt
$ hg update --clean 'head()'
1 files updated, 0 files merged, 0 files removed, 0 files unresolved
$ hg status
? new-file.txt
```
In this example, we see that the changes to the modified file, `existing-file.txt`, were discarded. We also see that the untracked file, new-file.txt, was left alone and not removed. 

It may be worth also deleting untracked files. I suspect that the potential for a problem from that is low, but it may be a good thing to do in followup.
</blockquote></details>

Nice, that is important that we can update to a new revision, even with a dirty working tree. I think the untracked files could be an issue, but right now I can't think of how. Maybe leave a comment that this does not clean up untracked files.
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1333 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl7Px08kM56Iu6fPE8:-NEvkMDq8vXbtMz1HY_c:b-83sjo4) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1333)):*

Yes! I want to give a clue as to why  `userSecret` is one of the parameters. How about "Returns the current revision of the local hg repo for a given project PT id, accessed using a given user."?
</blockquote></details>

Nice.
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1338 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl7nHtBkYsnYhi9gxS:-NEvkQAe4F78ajVktwX0:bkddfm8) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1338)):*

It's because a resource _isn't_ stored in a local hg repo! There is no .hg folder. I chose to make `GetRepoRevision()` reject requests for resources so that the method can always return a hg revision, rather than returning special case results.
</blockquote></details>

Ah, ok. I didn't know that. Great.
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs` line 3165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NElBJna621CYYhBl3zd:-NEvkjj79GOHW7X2PXW2:b-83sjo4) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L3165)):*

Done. I did the same to the following test as well, but then decided to leave it as is so that it more likely continues returning the specific bad value even if other parts of the system change (like if we we start using "blah" instead of "target" in the directory path).
</blockquote></details>

Nice.

marksvc (commented):
I put some more work into writing an acceptance tests to simulate the problem. I wrote a test that pulls content from ~/Paratext8Projects/... into the repo in /var/lib/scriptureforge/... . In this way, it simulates advancing SF's Paratext project repo, but without the content being imported into SF. 

Good point about using another branch. A similar test to the above can be done by having SF kill its process right after running SendReceive (and then Sync in SF again without having SF killing its process). I also wrote up how to do that. but I directed testers to the other method, which is something they should be able to do on future versions of SF without needing to use a then-old branch. These are written up in a jira comment.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-:-NEwdjo57uVSumKSlSvj:b-2aeckg)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/Services/HgWrapper.cs` line 120 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1521#-NEl6mhy7TvbvDWRuvA2:-NEw2jWmBtYk1-SO0R_Z:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/9f5006768b1110da55a5e23440b556bc2cb62c3f/src/SIL.XForge.Scripture/Services/HgWrapper.cs#L120)):*

Nice, that is important that we can update to a new revision, even with a dirty working tree. I think the untracked files could be an issue, but right now I can't think of how. Maybe leave a comment that this does not clean up untracked files.
</blockquote></details>

Done.

Nateowami:
@marksvc Can you change the commit message to something easier to understand than "set repo rev"? Maybe something about what problem is solved? When I read this I think "Wait, don't we already store the revision of the Paratext repo?" If the answer is "not always" then maybe the description should be something like "Fix repo revision not always being set during sync" or something.

marksvc:
@Nateowami How about this one? Also, don't miss the "sync: " part of the commit message, which indicates what area this patch is for.

marksvc:
@Nateowami That has advantages. Changing it to that.

Nateowami (commented):
> Reviewed 1 of 1 files at r3, 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1521)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1538, opened 2022-10-13
SF-1769 Fix direction of project names in project select

Nateowami:
Before | After
---------|--------
 | 

The menu itself:
LTR | RTL
------|-------
project_select_ltr | project_select_rtl

This screenshot is here to show that while project name when select is closed is LTR, and the project names in the menu are LTR, the add project item is still rendered RTL when using a RTL locale.


nigel-wells (changes requested):
Why would the list in the menu not be RTL as well?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1538#-:-NEIQuga-KXauk1Mx2hS:b7a6r53)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html` line 158 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1538#-NEIQnHG1X9F6DXYDJmh:-NEIQnHH60Gf6bRIcpq4:b6is5f4) ([raw file](https://github.com/sillsdev/web-xforge/blob/ef840544a29cb39732798f6dea8142abdd4b0916/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L158)):*
> ```HTML
>       <mdc-select
>         *ngIf="projectDocs != null"
>         dir="auto"
> ```

Is it possible to have the down arrow on the other side?

Nateowami (commented):
> Why would the list in the menu not be RTL as well?

Because the project short names and most project names (maybe all?) use the Latin script.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1538#-:-NEJSGiN0uFp7SWPJqxq:bxg007c)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html` line 158 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1538#-NEIQnHG1X9F6DXYDJmh:-NEJVgOf8kGO3RQizWti:bvydes2) ([raw file](https://github.com/sillsdev/web-xforge/blob/ef840544a29cb39732798f6dea8142abdd4b0916/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L158)):*

Is it possible to have the down arrow on the other side?
</blockquote></details>

That's a good point. I tried, and it's hard to do. Turns out MDC determines the direction to use by having a rule like:
``` css
[dir=rtl] .mdc-select__selected-text {
  // styles for RTL
}
```
The `[dir=rtl]` part matches the body element, and will continue to match even if we set `dir=ltr` on the select itself. Angular Material has a much more sophisticated way of handling directionality. With some carefully crafted CSS rules we could force this to do what we want, but we want to move to Angluar Material anyway. I experimented with switching the select to Material, but it isn't handling overflow or directionality quite how I expect it to.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1538)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html` line 158 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1538#-NEIQnHG1X9F6DXYDJmh:-NEJXlfFC3e-ctTixP4_:bkqvns) ([raw file](https://github.com/sillsdev/web-xforge/blob/ef840544a29cb39732798f6dea8142abdd4b0916/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L158)):*

That's a good point. I tried, and it's hard to do. Turns out MDC determines the direction to use by having a rule like:
``` css
[dir=rtl] .mdc-select__selected-text {
  // styles for RTL
}
```
The `[dir=rtl]` part matches the body element, and will continue to match even if we set `dir=ltr` on the select itself. Angular Material has a much more sophisticated way of handling directionality. With some carefully crafted CSS rules we could force this to do what we want, but we want to move to Angluar Material anyway. I experimented with switching the select to Material, but it isn't handling overflow or directionality quite how I expect it to.
</blockquote></details>

All that to say it should be addressed, but maybe be fixed as a separate task since I anticipate it taking a lot more time to fix.

nigel-wells (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1538#-:-NEXp0H47SBDJD9oF1yY:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1539, opened 2022-10-14
SF-1577 Opening nav drawer doesn't close keyboard on mobile

nigel-wells:
- Introduce new package to help calculator on-screen keyboard
- If keyboard has been open then don't try to re-focus the target editor

The package used in this solution has some limitations where the detection event is ~1 second before being emitted - they acknowledge this in their readme and code with reasons behind it. Still, it is better than what we currently experience on mobile devices.

My testing finds that it works well as long as you you don't have your tester hat on where you immediately try to close the keyboard in <1s which of course will trigger it to open again. If you tap it a second time I find that it stays away.

I've tested on a real Android phone and a simulated iOS phone.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1539)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/package-lock.json
+++ b/src/SIL.XForge.Scripture/ClientApp/package-lock.json
@@ -43,6 +43,7 @@
         "ngx-avatar": "^4.1.0",
         "ngx-cookie-service": "^13.2.1",
         "ngx-quill": "14.1.1",
+        "on-screen-keyboard-detector": "^2.3.0",
         "ot-json0": "^1.1.0",
         "papaparse": "^5.3.2",
         "process": "^0.11.10",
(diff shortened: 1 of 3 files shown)


RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1539)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1010 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1539#-NElGYwx6FGklBWwIlM6:-NElGYwx6FGklBWwIlM7:b-vws0ml) ([raw file](https://github.com/sillsdev/web-xforge/blob/fa257c091de83865001baadafbaf536c52e08d52/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1010)):*
> ```XML
>         // reset focus, which causes Quill to scroll to the selection
>         this.target!.blur();
>         if (!this.isKeyboardVisible) {
> ```

What is the logic behind how this fixes the problem? Is it because we call `setTextHeight()` anytime a user clicks outside of the editor?



_Code quote:_
```XML
        if (!this.isKeyboardVisible) {
          this.target!.focus();
        }
```

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1539)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1010 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1539#-NElGYwx6FGklBWwIlM6:-NFBUN_p-0vzwOvYwZfs:b1nh7gp) ([raw file](https://github.com/sillsdev/web-xforge/blob/fa257c091de83865001baadafbaf536c52e08d52/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1010)):*

What is the logic behind how this fixes the problem? Is it because we call `setTextHeight()` anytime a user clicks outside of the editor?
</blockquote></details>

The underlaying scenario is we recalculate the height of the quill editor on the resize event. The resize event occurs when the keyboard starts disappearing and because the focus was initially set we then ensure the cursor is still set after the resize event. This simply checks to see if the keyboard was there (which is only triggered for the mobile keyboards) then don't set the focus again after the resize event.

RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1539)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1010 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1539#-NElGYwx6FGklBWwIlM6:-NFEwOkc3Sh2Pp3DQXuM:bcd6kbc) ([raw file](https://github.com/sillsdev/web-xforge/blob/fa257c091de83865001baadafbaf536c52e08d52/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1010)):*

The underlaying scenario is we recalculate the height of the quill editor on the resize event. The resize event occurs when the keyboard starts disappearing and because the focus was initially set we then ensure the cursor is still set after the resize event. This simply checks to see if the keyboard was there (which is only triggered for the mobile keyboards) then don't set the focus again after the resize event.
</blockquote></details>

Ok, let's add a comment, since it isn't obvious from reading the code.

nigel-wells (commented):
I'm putting this on hold for now as this change might not be needed any more of the change I made for SF-1733 passes testing then that also resolves this issue without the need for an additional package.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1539#-:-NFGe0T_C92LHTikp460:b9ijvv8)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

nigel-wells:
Closing this PR as I found a different way to improve what we already have seeming it is our code causing the issue.

PR #1549, opened 2022-10-21
SF-1778 Add reviewer role for invitation and share links

RaymondLuong3:
* enable inviting users as a reviewer
* only allow administrators to invite reviewers
* update endpoint to accept the reviewer role in the invite

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1549)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/sf-project-role-info.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/sf-project-role-info.ts
@@ -10,7 +10,7 @@ export const SF_PROJECT_ROLES: ProjectRoleInfo[] = [
     canBeShared: false
   },
   { role: SFProjectRole.ParatextObserver, displayName: 'Observer', canBeShared: false },
-  { role: SFProjectRole.Reviewer, displayName: 'Reviewer', canBeShared: false },
+  { role: SFProjectRole.Reviewer, displayName: 'Reviewer', canBeShared: true },
   { role: SFProjectRole.CommunityChecker, displayName: 'Community Checker', canBeShared: true },
   { role: SFProjectRole.Observer, displayName: 'View Translation', canBeShared: true }
 ];
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts
@@ -10,6 +10,7 @@ import { SFProjectRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-
 import { TranslateShareLevel } from 'realtime-server/lib/esm/scriptureforge/models/translate-config';
 import { BehaviorSubject } from 'rxjs';
 import { anything, capture, mock, verify, when } from 'ts-mockito';
+import { FeatureFlag, FeatureFlagService } from 'xforge-common/feature-flags/feature-flag.service';
 import { I18nService } from 'xforge-common/i18n.service';
 import { LocationService } from 'xforge-common/location.service';
 import { NoticeService } from 'xforge-common/notice.service';
@@ -31,6 +32,7 @@ const mockedPwaService = mock(PwaService);
 const mockedI18nService = mock(I18nService);
 const mockedLocationService = mock(LocationService);
 const mockedUserService = mock(UserService);
+const mockedFeatureFlagService = mock(FeatureFlagService);
 
 describe('ShareControlComponent', () => {
   configureTestingModule(() => ({
@@ -42,7 +44,8 @@ describe('ShareControlComponent', () => {
       { provide: PwaService, useMock: mockedPwaService },
       { provide: I18nService, useMock: mockedI18nService },
       { provide: LocationService, useMock: mockedLocationService },
-      { provide: UserService, useMock: mockedUserService }
+      { provide: UserService, useMock: mockedUserService },
+      { provide: FeatureFlagService, useMock: mockedFeatureFlagService }
     ]
   }));
 
@@ -389,6 +392,7 @@ class TestEnvironment {
     when(mockedProjectService.onlineIsAlreadyInvited(anything(), 'unknown-address@example.com')).thenResolve(false);
     when(mockedProjectService.onlineIsAlreadyInvited(anything(), 'already@example.com')).thenResolve(true);
     when(mockedLocationService.origin).thenReturn('https://scriptureforge.org');
+    when(mockedFeatureFlagService.allowAddingNotes).thenReturn({ enabled: true } as FeatureFlag);
 
     this.fixture.detectChanges();
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts
@@ -217,13 +217,23 @@ describe('ShareControlComponent', () => {
   }));
 
   it('changing user role refreshes the share key', fakeAsync(() => {
-    const env = new TestEnvironment({ isLinkSharingEnabled: true, projectId: 'myProject1' });
+    const env = new TestEnvironment({
+      projectId: 'project01',
+      userId: 'user02',
+      isLinkSharingEnabled: true
+    });
     env.wait();
-    env.hostComponent.component.roleControl.setValue(SFProjectRole.Observer);
+    const roles: string[] = env.component.availableRolesInfo.map(info => info.role);
+    expect(roles).toContain(SFProjectRole.CommunityChecker);
+    expect(roles).toContain(SFProjectRole.Observer);
+    expect(roles).toContain(SFProjectRole.Reviewer);
+    env.component.roleControl.setValue(SFProjectRole.Observer);
     env.wait();
-    verify(mockedProjectService.onlineGetLinkSharingKey('myProject1', anything())).twice();
-    expect().nothing();
-    flush();
+    verify(mockedProjectService.onlineGetLinkSharingKey('project01', anything())).twice();
+    env.component.roleControl.setValue(SFProjectRole.Reviewer);
+    env.wait();
+    verify(mockedProjectService.onlineGetLinkSharingKey('project01', anything())).thrice();
+    expect(env.component.isLinkSharingEnabled).toBe(true);
   }));
 
   it('role should be visible for administrators', fakeAsync(() => {
@@ -242,15 +252,14 @@ describe('ShareControlComponent', () => {
     const env = new TestEnvironment({ userId: 'user02', projectId: 'project01' });
     env.wait();
     expect(env.hostComponent.component.roleControl.value).toEqual(SF_DEFAULT_SHARE_ROLE);
-    expect(env.hostComponent.component.availableRolesInfo.length).toBe(2);
+    expect(env.hostComponent.component.availableRolesInfo.length).toBe(3);
   }));
 
   it('default share role should be translation observer when checking is disabled', fakeAsync(() => {
     const env = new TestEnvironment({ userId: 'user02', projectId: 'project01', checkingEnabled: false });
     env.wait();
     expect(env.hostComponent.component.roleControl.value).toEqual(SF_DEFAULT_TRANSLATE_SHARE_ROLE);
-    expect(env.hostComponent.component.availableRolesInfo.length).toBe(1);
-    expect(env.hostComponent.component.availableRolesInfo[0].role).toBe(SF_DEFAULT_TRANSLATE_SHARE_ROLE);
+    expect(env.hostComponent.component.availableRolesInfo.length).toBe(2);
   }));
 
   it('should hide link sharing if checking is unavailable', fakeAsync(() => {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts
@@ -16,7 +16,11 @@ import { SubscriptionDisposable } from 'xforge-common/subscription-disposable';
 import { UserService } from 'xforge-common/user.service';
 import { XFValidators } from 'xforge-common/xfvalidators';
 import { SFProjectProfileDoc } from '../../core/models/sf-project-profile-doc';
-import { SF_DEFAULT_SHARE_ROLE, SF_PROJECT_ROLES } from '../../core/models/sf-project-role-info';
+import {
+  SF_DEFAULT_SHARE_ROLE,
+  SF_DEFAULT_TRANSLATE_SHARE_ROLE,
+  SF_PROJECT_ROLES
+} from '../../core/models/sf-project-role-info';
 import { SFProjectService } from '../../core/sf-project.service';
 
 /** UI to share project access with new users, such as by sending an invitation email. */
@@ -81,14 +85,6 @@ export class ShareControlComponent extends SubscriptionDisposable {
     this.projectId$.next(id);
   }
 
-  get defaultShareRole(): string | undefined {
-    const roles = this.userShareableRoles;
-    if (this.defaultRole != null && roles.some(role => role === this.defaultRole)) {
-      return this.defaultRole;
-    }
-    return roles.some(role => role === SF_DEFAULT_SHARE_ROLE) ? SF_DEFAULT_SHARE_ROLE : roles[0];
-  }
-
   get availableRolesInfo(): ProjectRoleInfo[] {
     return SF_PROJECT_ROLES.filter(info => info.canBeShared && this.userShareableRoles.includes(info.role));
   }
@@ -119,6 +115,8 @@ export class ShareControlComponent extends SubscriptionDisposable {
       [SFProjectRole.CommunityChecker]:
         project.checkingConfig.shareEnabled && project.checkingConfig.shareLevel === CheckingShareLevel.Anyone,
       [SFProjectRole.Observer]:
+        project.translateConfig.shareEnabled && project.translateConfig.shareLevel === TranslateShareLevel.Anyone,
+      [SFProjectRole.Reviewer]:
         project.translateConfig.shareEnabled && project.translateConfig.shareLevel === TranslateShareLevel.Anyone
     };
     return linkSharingSettings[this.shareRole] === true && this.userShareableRoles.includes(this.shareRole);
@@ -149,10 +147,26 @@ export class ShareControlComponent extends SubscriptionDisposable {
           project.translateConfig.shareEnabled &&
           SF_PROJECT_RIGHTS.hasRight(project, this.userService.currentUserId, SFProjectDomain.Texts, Operation.View) &&
           userRole !== SFProjectRole.CommunityChecker
+      },
+      {
+        role: SFProjectRole.Reviewer,
+        available: true,
+        permission: this.isProjectAdmin
       }
     ]
       .filter(info => info.available && (info.permission || this.isProjectAdmin))
       .map(info => info.role as string);
+  }
+
+  private get defaultShareRole(): string | undefined {
+    const roles = this.userShareableRoles;
+    if (this.defaultRole != null && roles.some(role => role === this.defaultRole)) {
+      return this.defaultRole;
+    }
+    if (roles.some(role => role === SF_DEFAULT_SHARE_ROLE)) {
+      return SF_DEFAULT_SHARE_ROLE;
+    }
+    return roles.some(role => role === SF_DEFAULT_TRANSLATE_SHARE_ROLE) ? SF_DEFAULT_TRANSLATE_SHARE_ROLE : roles[0];
   }
 
   copyShareLink(): void {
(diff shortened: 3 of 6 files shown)


RaymondLuong3 (commented):
I will need to factor this into the feature flag. Should be a quick change for Monday.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-:-NEwe2W15lBjHHQD7M2c:b-cjqclq)* status: 0 of 5 files reviewed, all discussions resolved

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1549)* status: 0 of 5 files reviewed, 1 unresolved discussion

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts` line 157 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NFA1AgdDtNRPp-sUlp1:-NFA1AgdDtNRPp-sUlp2:b2zookg) ([raw file](https://github.com/sillsdev/web-xforge/blob/9d32577cc3594d1505aa5bd29c5c32ff197791fc/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts#L157)):*
> ```XML
>         role: SFProjectRole.Reviewer,
>         available: this.featureFlags.allowAddingNotes.enabled,
>         permission: this.isProjectAdmin
> ```

I'm not entirely certain if we want any user to be able to invite other reviewers. Maybe we do, since we already have the ability to create a share link.



_Code quote:_
```XML
  permission: this.isProjectAdmin
```

marksvc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1549)* status: all files reviewed, 3 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts` line 155 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NFG438ADn9AWInTV_72:-NFG438ADn9AWInTV_73:bdq3zrv) ([raw file](https://github.com/sillsdev/web-xforge/blob/9d32577cc3594d1505aa5bd29c5c32ff197791fc/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts#L155)):*
> ```XML
>       },
>       {
>         role: SFProjectRole.Reviewer,
> ```

Hmm, I'm a bit concerned that seems that the thing that prevents someone from getting a shareable link, which makes Joining users have the Reviewer role, is allowed or disallowed by the frontend. It appears that the C# code will make a shareable link with Reviewer role, if asked by the frontend, as long as TranslateConfig.ShareEnabled and sharelevel is Anyone. Hmm, it appears that there may be similar problems with the permissions check for the CommunityChecker and Observer, a few lines above, where the permissions are specified on the client, but the backend just checks that the user is on the project at all. 
Let's talk about this some more.

I also wonder if maybe the backend should not include the capability for a shareable link to be created for people to Join as Reviewers.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts` line 172 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NFG8ciBEHCfVKZuCvRu:-NFG8ciBEHCfVKZuCvRv:b-vt647a) ([raw file](https://github.com/sillsdev/web-xforge/blob/9d32577cc3594d1505aa5bd29c5c32ff197791fc/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts#L172)):*
> ```XML
>       return SF_DEFAULT_SHARE_ROLE;
>     }
>     return roles.some(role => role === SF_DEFAULT_TRANSLATE_SHARE_ROLE) ? SF_DEFAULT_TRANSLATE_SHARE_ROLE : roles[0];
> ```

I can't help but think that `roles` could be an empty list some time, such as if `project.translateConfig.shareEnabled===false` and `project.checkingConfig.checkingEnabled===false` and `this.isProjectAdmin===false`. But so far, looking at the code and fiddling with the frontend hasn't led me to it. 



_Code quote:_
```XML
[0];
```
___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 103 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NFG0Wtq1sz8kSmLocQs:-NFG0Wtq1sz8kSmLocQt:b-dwa9mj) ([raw file](https://github.com/sillsdev/web-xforge/blob/9d32577cc3594d1505aa5bd29c5c32ff197791fc/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L103)):*
> ```Smalltalk
>             env.SecurityService.GenerateKey().Returns("sfobserverkey");
>             await env.Service.InviteAsync(User01, Project04, observerEmail, "en", SFProjectRole.SFObserver);
>             await env.EmailService
> ```

I think a significant thing to check is that when InviteAsync() is called in a valid way, the project secrets are modified such that there is now a new share key with specific values for the  email address, share key code, and in this situation, the role. So checking that an email is sent is good, but some of our 'security' of how sharing is handled relies on the values of the share key in the project secrets. 

eg, maybe something like
```cs
shareKey = GetProjectShareKeyByKeyCode("sfobserverkey")
Assert.That(shareKey.Email, Is.EqualTo("sf_observer@example.com"));
Assert.That(shareKey.Role, Is.EqualTo(SFProjectRole.SFObserver));
Assert.That(shareKey.Key, Is.EqualTo("sfobserverkey")); // which may or may not be very necessary depending on how GetProjectShareKeyByKeyCode() works
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1549)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts` line 155 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NFG438ADn9AWInTV_72:-NFtvpg12pnYfVIlKVnv:b-i2oarc) ([raw file](https://github.com/sillsdev/web-xforge/blob/9d32577cc3594d1505aa5bd29c5c32ff197791fc/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts#L155)):*

Hmm, I'm a bit concerned that seems that the thing that prevents someone from getting a shareable link, which makes Joining users have the Reviewer role, is allowed or disallowed by the frontend. It appears that the C# code will make a shareable link with Reviewer role, if asked by the frontend, as long as TranslateConfig.ShareEnabled and sharelevel is Anyone. Hmm, it appears that there may be similar problems with the permissions check for the CommunityChecker and Observer, a few lines above, where the permissions are specified on the client, but the backend just checks that the user is on the project at all. 
Let's talk about this some more.

I also wonder if maybe the backend should not include the capability for a shareable link to be created for people to Join as Reviewers.
</blockquote></details>

I've added a check that a user must be an admin in order to create a link. I also added some logic to check whether a user should have the right to invite a specific role to the backend. This will restrict who has permission to get a share link.
___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 103 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NFG0Wtq1sz8kSmLocQs:-NFuAkdS0wsDKNb-mkOM:b-yjfgon) ([raw file](https://github.com/sillsdev/web-xforge/blob/9d32577cc3594d1505aa5bd29c5c32ff197791fc/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L103)):*

I think a significant thing to check is that when InviteAsync() is called in a valid way, the project secrets are modified such that there is now a new share key with specific values for the  email address, share key code, and in this situation, the role. So checking that an email is sent is good, but some of our 'security' of how sharing is handled relies on the values of the share key in the project secrets. 

eg, maybe something like
```cs
shareKey = GetProjectShareKeyByKeyCode("sfobserverkey")
Assert.That(shareKey.Email, Is.EqualTo("sf_observer@example.com"));
Assert.That(shareKey.Role, Is.EqualTo(SFProjectRole.SFObserver));
Assert.That(shareKey.Key, Is.EqualTo("sfobserverkey")); // which may or may not be very necessary depending on how GetProjectShareKeyByKeyCode() works
```
</blockquote></details>

Good thinking. I'll update the test to also check the project secrets were updated.

marksvc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1549)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts` line 155 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NFG438ADn9AWInTV_72:-NFubej64CeivuYcchcH:b-6mhyx0) ([raw file](https://github.com/sillsdev/web-xforge/blob/9d32577cc3594d1505aa5bd29c5c32ff197791fc/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.ts#L155)):*

I've added a check that a user must be an admin in order to create a link. I also added some logic to check whether a user should have the right to invite a specific role to the backend. This will restrict who has permission to get a share link.
</blockquote></details>

Great, thanks!
___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 103 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NFG0Wtq1sz8kSmLocQs:-NFuYXmdDJRP7nM1QTyN:b-6t2132) ([raw file](https://github.com/sillsdev/web-xforge/blob/9d32577cc3594d1505aa5bd29c5c32ff197791fc/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L103)):*

Good thinking. I'll update the test to also check the project secrets were updated.
</blockquote></details>

This is good, but it didn't check a key part here which is that the _role_ is what is intended :)
___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 2592 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NFu_Al40k_pNeGcvPrK:-NFu_Al40k_pNeGcvPrL:b-nt8ydb) ([raw file](https://github.com/sillsdev/web-xforge/blob/337acf9c4688d788d92eb0257ebe05933d9ef60e/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L2592)):*
> ```Smalltalk
>                                     }
>                                 },
>                                 CheckingConfig = new CheckingConfig { CheckingEnabled = true, ShareEnabled = false },
> ```

I wonder if it would help what is being tested if 
```cs
ShareEnabled=true,
ShareLevel = Checking...Anyone
```
It may not make a difference, but seems like it should be relevant.  (?)



_Code quote:_
```Smalltalk
ShareEnabled = false
```
___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 2697 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NFu_hyp8CZhAT4swvtG:-NFu_hyp8CZhAT4swvtH:b4safy1) ([raw file](https://github.com/sillsdev/web-xforge/blob/337acf9c4688d788d92eb0257ebe05933d9ef60e/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L2697)):*
> ```Smalltalk
>                                     ShareLevel = TranslateShareLevel.Anyone
>                                 },
>                                 UserRoles = { { User01, SFProjectRole.Administrator } }
> ```

This sort of change  makes me wonder if it will make some other test start passing now but for the wrong reason :)
For example, suppose the prior code was actually
```cs
UserRoles = { { User01, SFProjectRole.Translator } }
```
and we had a test saying
```cs
MyTest() 
{
  Assert.That(CheckUserPermissions(User01), Is.True, "Translators have adequate access");
}
```
which runs thru SUT code of
```cs
CheckUserPermissions(User01)
{
  if (user is a translator) 
    return true;
  if (user is an admin)
    return true;
  return false;
}
```
Then if we change the TestEnvironment so that User01 is now an Administrator, we are no longer really testing the point of `MyTest()`, which is that a _translator_ has adequate access. So then if we later change the SUT code in `CheckUserPermissions()` so that only, say, admininstrators can do the activity, we don't break the corresponding test. 
That.. maybe wasn't a super example. But anyway, it's something that can be mitigated if in `MyTest()` we specify the environment we are meaning to test, for example
```cs
Assert.That(User01 has the Translator role, "setup");
```
Something to keep in mind.



_Code quote:_
```Smalltalk
UserRoles = { { User01, SFProjectRole.Administrator } }
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1549)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 103 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NFG0Wtq1sz8kSmLocQs:-NGIgIK3E7sD6WYyH9KV:b-mbf4vj) ([raw file](https://github.com/sillsdev/web-xforge/blob/9d32577cc3594d1505aa5bd29c5c32ff197791fc/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L103)):*

This is good, but it didn't check a key part here which is that the _role_ is what is intended :)
</blockquote></details>

Ah yes. Good catch.
___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 2592 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NFu_Al40k_pNeGcvPrK:-NGIX6xZCNXESsxrnbDi:b-r7kn8a) ([raw file](https://github.com/sillsdev/web-xforge/blob/337acf9c4688d788d92eb0257ebe05933d9ef60e/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L2592)):*

I wonder if it would help what is being tested if 
```cs
ShareEnabled=true,
ShareLevel = Checking...Anyone
```
It may not make a difference, but seems like it should be relevant.  (?)
</blockquote></details>

This looks to be intentional. I get test failures if I update this to link sharing.
___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 2697 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NFu_hyp8CZhAT4swvtG:-NGIg6WjBaHWD0Z9L54V:b-yox7ka) ([raw file](https://github.com/sillsdev/web-xforge/blob/337acf9c4688d788d92eb0257ebe05933d9ef60e/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L2697)):*

This sort of change  makes me wonder if it will make some other test start passing now but for the wrong reason :)
For example, suppose the prior code was actually
```cs
UserRoles = { { User01, SFProjectRole.Translator } }
```
and we had a test saying
```cs
MyTest() 
{
  Assert.That(CheckUserPermissions(User01), Is.True, "Translators have adequate access");
}
```
which runs thru SUT code of
```cs
CheckUserPermissions(User01)
{
  if (user is a translator) 
    return true;
  if (user is an admin)
    return true;
  return false;
}
```
Then if we change the TestEnvironment so that User01 is now an Administrator, we are no longer really testing the point of `MyTest()`, which is that a _translator_ has adequate access. So then if we later change the SUT code in `CheckUserPermissions()` so that only, say, admininstrators can do the activity, we don't break the corresponding test. 
That.. maybe wasn't a super example. But anyway, it's something that can be mitigated if in `MyTest()` we specify the environment we are meaning to test, for example
```cs
Assert.That(User01 has the Translator role, "setup");
```
Something to keep in mind.
</blockquote></details>

Good thinking. I added a new user to the project with the administrator role instead.

marksvc (changes requested):
Great! One more thing :)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-:-NGIqaxw8O24edLHWFMB:bi71yi3)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 2697 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NFu_hyp8CZhAT4swvtG:-NGInAkLBMtbN18BwgAY:b4c9lyr) ([raw file](https://github.com/sillsdev/web-xforge/blob/337acf9c4688d788d92eb0257ebe05933d9ef60e/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L2697)):*

Good thinking. I added a new user to the project with the administrator role instead.
</blockquote></details>

:thumbsup: 
___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 2857 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NGIqIjgC7PvifURbtDQ:-NGIqIjgC7PvifURbtDR:b-duwse3) ([raw file](https://github.com/sillsdev/web-xforge/blob/6f921367661680706f3bb0aaa682e3743b7a4744/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L2857)):*
> ```Smalltalk
>                             new SFProjectUserConfig { Id = SFProjectUserConfig.GetDocId(Project02, User02) },
>                             new SFProjectUserConfig { Id = SFProjectUserConfig.GetDocId(Project03, User01) },
>                             new SFProjectUserConfig { Id = SFProjectUserConfig.GetDocId(Project03, User02) },
> ```

Obviously this doesn't matter for the tests that are running at present, but it may be helpful in the future if we now add
```cs
new SFProjectUserConfig { Id = SFProjectUserConfig.GetDocId(Project04, User02) },
```
to correspond to adding the user to the project.



_Code quote:_
```Smalltalk
new MemoryRepository
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1549)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 2857 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NGIqIjgC7PvifURbtDQ:-NGJLCpl345pkZqI4zLC:br4occ9) ([raw file](https://github.com/sillsdev/web-xforge/blob/6f921367661680706f3bb0aaa682e3743b7a4744/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L2857)):*

Obviously this doesn't matter for the tests that are running at present, but it may be helpful in the future if we now add
```cs
new SFProjectUserConfig { Id = SFProjectUserConfig.GetDocId(Project04, User02) },
```
to correspond to adding the user to the project.
</blockquote></details>

Done. And added a few more missing configs.

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1549)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1549)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 1127 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NGy6Ana83BbWqRaR4tJ:-NGy6Ana83BbWqRaR4tK:bmxkili) ([raw file](https://github.com/sillsdev/web-xforge/blob/ea64d9f885b7693528824ed6c79d1cd3d4c3fd93/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L1127)):*
> ```Smalltalk
>                 return false;
>             if (role == SFProjectRole.CommunityChecker)
>                 return true;
> ```

This logic seems wrong. What if the user is an observer?

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1549)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 1127 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1549#-NGy6Ana83BbWqRaR4tJ:-NH0TZp-EeYInm1glZpS:b-f3gnto) ([raw file](https://github.com/sillsdev/web-xforge/blob/ea64d9f885b7693528824ed6c79d1cd3d4c3fd93/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L1127)):*

This logic seems wrong. What if the user is an observer?
</blockquote></details>

That is a good question. As it is now, an observer is allowed to invite community checkers before these changes. And as it stands, an observer has access to the community checking tool, so observers inviting community checkers is not overreaching what they should be able to share. It is worth revisiting the sharing workflow.

Nateowami:
> As it is now, an observer is allowed to invite community checkers before these changes. And as it stands, an observer has access to the community checking tool

Surely observers can't answer questions, right? If they can that's a bug.

RaymondLuong3:
Observers cannot answer questions or comment. They only have view access. I just checked, and the add answer button is not available to observer users.

PR #1552, opened 2022-10-21
Turn on TS option useUnknownInCatchVariables

Nateowami:
See the documentation for useUnknownInCatchVariables: https://www.typescriptlang.org/tsconfig#useUnknownInCatchVariables

When you have this:
``` ts
catch (error) {
  // what is the type of error here?
}
```
It used to be `any`. If you turn on `useUnknownInCatchVariables`, it's `unknown`. A recent TS update made `useUnknownInCatchVariables` the default in strict mode. Rather than going around and changing a bunch of code when updating TS, I just turned the setting off temporarily. Now I'm turning it back on so any future code we write will have to either explicitly write `catch (error: any)`, or (preferable) write type-safe code, using type guards from `src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts`

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1552)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
@@ -274,7 +274,7 @@ export class AuthService {
       }
       await this.remoteStore.init(() => this.getAccessToken());
       return { loggedIn: true, newlyLoggedIn: false };
-    } catch (error) {
+    } catch (error: any) {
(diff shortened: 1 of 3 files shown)


RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1552)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1557, opened 2022-10-26
SF-1628 Fix src and seq not being correct after reconnection

Nateowami:
- Add ability to interfere with ShareDB connection
- Store src and seq before sending to server
- Interfere with ShareDB connection to ensure src and seq are set

This PR is not polished. It probably calls for a *lot* of comments, and some other changes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1557)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts
@@ -67,7 +67,7 @@ export class QuestionDoc extends ProjectDataDoc<Question> {
     return undefined;
   }
 
-  protected async updateOfflineData(force: boolean = false): Promise<void> {
+  async updateOfflineData(force: boolean = false): Promise<void> {
     // Check to see if any answers have been removed by comparing with current offline data
     const fileService: FileService | undefined = this.realtimeService.fileService;
     if (fileService != null) {
--- a/src/SIL.XForge.Scripture/ClientApp/src/typings/sharedb.d.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/typings/sharedb.d.ts
@@ -40,6 +40,7 @@ declare module 'sharedb/lib/client' {
   export type Callback = (err: Error) => void;
 
   export interface Connection {
+    id: string | null;
     get(collection: string, id: string): Doc;
     createFetchQuery(
       collection: string,
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts
@@ -23,7 +23,7 @@ class LocalStorageFlagStore implements FeatureFlagStore {
     this._enabled = typeof valueFromStore === 'boolean' ? valueFromStore : defaultValue;
   }
 
-  get enabled() {
+  get enabled(): boolean {
     return this._enabled;
   }
 
@@ -64,6 +64,16 @@ export class FeatureFlagService {
   allowAddingNotes: FeatureFlag = new FeatureFlag(
     new LocalStorageFlagStore('ALLOW_ADDING_NOTES'),
     'Allow adding notes'
+  );
+
+  preventOpSubmission: FeatureFlag = new FeatureFlag(
+    new LocalStorageFlagStore('PREVENT_OP_SUBMISSION'),
+    'Prevent op submission (intentionally breaks things)'
+  );
+
+  preventOpAcknowledgement: FeatureFlag = new FeatureFlag(
+    new LocalStorageFlagStore('PREVENT_OP_ACKNOWLEDGEMENT'),
+    'Prevent op acknowledgement (intentionally breaks things)'
   );
 
   get featureFlags(): FeatureFlag[] {
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/memory-realtime-remote-store.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/memory-realtime-remote-store.ts
@@ -55,6 +55,8 @@ export class MemoryRealtimeRemoteStore extends RealtimeRemoteStore {
   createQueryAdapter(collection: string, parameters: QueryParameters): RealtimeQueryAdapter {
     return new MemoryRealtimeQueryAdapter(this, collection, parameters);
   }
+
+  subscribeToBeforeSendOp(_listener: (collection: string, docId: string) => Promise<void>): void {}
 }
 
 /**
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/project-data-doc.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/project-data-doc.ts
@@ -75,7 +75,7 @@ export abstract class ProjectDataDoc<T extends ProjectData = ProjectData> extend
     return undefined;
   }
 
-  protected async updateOfflineData(force: boolean = false): Promise<void> {
+  async updateOfflineData(force: boolean = false): Promise<void> {
     await super.updateOfflineData(force);
     await this.updateFileCache();
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/models/realtime-doc.ts
@@ -197,7 +197,7 @@ export abstract class RealtimeDoc<T = any, Ops = any, P = any> {
    *
    * @param {boolean} [force=false] Indicates whether force the update to occur even if not subscribed.
    */
-  protected async updateOfflineData(force: boolean = false): Promise<void> {
+  async updateOfflineData(force: boolean = false): Promise<void> {
     if (this.adapter.type == null) {
       return;
     }
@@ -213,7 +213,12 @@ export abstract class RealtimeDoc<T = any, Ops = any, P = any> {
 
     const pendingOps = this.adapter.pendingOps
       .filter(opInfo => opInfo.op != null)
-      .map(opInfo => this.prepareDataForStore(opInfo.op));
+      .map(opInfo => {
+        const data = { op: this.prepareDataForStore(opInfo.op) };
+        if (opInfo.hasOwnProperty('src')) data['src'] = opInfo.src;
+        if (opInfo.hasOwnProperty('seq')) data['seq'] = opInfo.seq;
+        return data;
+      });
 
     this.offlineSnapshotVersion = this.adapter.version;
     const offlineData: RealtimeOfflineData = {
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime-remote-store.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime-remote-store.ts
@@ -12,6 +12,7 @@ export abstract class RealtimeRemoteStore {
   abstract init(getAccessToken: () => Promise<string | undefined>): Promise<void>;
   abstract createDocAdapter(collection: string, id: string): RealtimeDocAdapter;
   abstract createQueryAdapter(collection: string, parameters: QueryParameters): RealtimeQueryAdapter;
+  abstract subscribeToBeforeSendOp(listener: (collection: string, docId: string) => Promise<void>): void;
 }
 
 /**
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts
@@ -35,6 +35,13 @@ export class RealtimeService {
     if (this.fileService != null) {
       this.fileService.init(this);
     }
+
+    // Subscribe to before-send-op events so the pending ops can be updated in IndexedDB with the src and seq on the
+    // inflight op, before it is sent to the server.
+    this.remoteStore.subscribeToBeforeSendOp(async (collection: string, docId: string) => {
+      const doc: RealtimeDoc | undefined = this.docs.get(getDocKey(collection, docId));
+      await doc?.updateOfflineData();
+    });
   }
 
   get<T extends RealtimeDoc>(collection: string, id: string): T {
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts
@@ -1,6 +1,7 @@
 import { Injectable } from '@angular/core';
 import cloneDeep from 'lodash-es/cloneDeep';
-import ReconnectingWebSocket from 'reconnecting-websocket';
+import ReconnectingWebSocket, { Message } from 'reconnecting-websocket';
+import Events from 'reconnecting-websocket/dist/events';
 import * as RichText from 'rich-text';
 import { fromEvent, Observable, Subject } from 'rxjs';
 import { filter, map } from 'rxjs/operators';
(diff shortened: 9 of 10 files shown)


marksvc:
@Nateowami Is there something I can do to help you get this finished up so we can fix this bug?

RaymondLuong3 (changes requested):
I haven't given it a full testing, but just wanted to leave comments before doing that later.

> Reviewed 6 of 8 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-:-NQ1Iwxm0fKNqB7ff3BX:b-94hjne)* status: all files reviewed, 4 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts` line 42 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-NQ18pSL17qJj6JdKsK4:-NQ18pSL17qJj6JdKsK5:b-fmm3et) ([raw file](https://github.com/sillsdev/web-xforge/blob/67a0e1909c57ddc7bc9bd608a514984e497677a3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts#L42)):*
> ```XML
>     // inflight op, before it is sent to the server.
>     this.remoteStore.subscribeToBeforeSendOp(async (collection: string, docId: string) => {
>       const doc = this.docs.get(getDocKey(collection, docId));
> ```

Add a type here.



_Code quote:_
```XML
const doc = this.docs.get(getDocKey(collection, docId));
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts` line 62 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-NQ1Ijzn2wdtdIUjqZS3:-NQ1Ijzn2wdtdIUjqZS4:byvbntc) ([raw file](https://github.com/sillsdev/web-xforge/blob/67a0e1909c57ddc7bc9bd608a514984e497677a3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts#L62)):*
> ```XML
>  * - This class is implementing a websocket, but only the methods that are used by ShareDB are implemented.
>  */
> class ShareDBWebsocketAdapter {
> ```

Was there no way to just extend the ReconnectingWebSocket class? I assume that there was not, but just checking.



_Code quote:_
```XML
class ShareDBWebsocketAdapter {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts` line 89 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-NQ1A73I7_bW-c_AP5XR:-NQ1A73I7_bW-c_AP5XS:bh6dy22) ([raw file](https://github.com/sillsdev/web-xforge/blob/67a0e1909c57ddc7bc9bd608a514984e497677a3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts#L89)):*
> ```XML
>     this.socket.onmessage = (event: MessageEvent) => {
>       const sendingOp = isMessageSendingOp(tryParseJSON(event.data));
>       if (handler == null || (sendingOp && this.featureFlags.preventOpAcknowledgement.enabled)) {
> ```

You could also reverse the statement and only call handler() if handler != null && (sendingOp || disabled)



_Code quote:_
```XML
      if (handler == null || (sendingOp && this.featureFlags.preventOpAcknowledgement.enabled)) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts` line 129 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-NQ1BXPf5UUT54MdhIH-:-NQ1BXPf5UUT54MdhIH0:bn9hyqb) ([raw file](https://github.com/sillsdev/web-xforge/blob/67a0e1909c57ddc7bc9bd608a514984e497677a3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts#L129)):*
> ```XML
>    */
>   async send(data: Message): Promise<void> {
>     const message = tryParseJSON(data);
> ```

Add types here.



_Code quote:_
```XML
    const message = tryParseJSON(data);
    const sendingOp = isMessageSendingOp(message);
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1557)* status: 8 of 10 files reviewed, 4 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts` line 42 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-NQ18pSL17qJj6JdKsK4:-NQ1roXIC2i4HJqOUzEI:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/67a0e1909c57ddc7bc9bd608a514984e497677a3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts#L42)):*

Add a type here.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts` line 62 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-NQ1Ijzn2wdtdIUjqZS3:-NQ1sb1s8pZdMA3cpkwa:bysgtpo) ([raw file](https://github.com/sillsdev/web-xforge/blob/67a0e1909c57ddc7bc9bd608a514984e497677a3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts#L62)):*

Was there no way to just extend the ReconnectingWebSocket class? I assume that there was not, but just checking.
</blockquote></details>

I did not investigate extending ReconnectingWebSocket, preferring composition over inheritance. JavaScript isn't object oriented, so even starting to think about what it means to extend a class and what the ramifications of that might be is difficult. This solution is verbose, but there are fewer gotchas.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts` line 89 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-NQ1A73I7_bW-c_AP5XR:-NQ1sJuG91H5Dl-0kfG3:b-haz60z) ([raw file](https://github.com/sillsdev/web-xforge/blob/67a0e1909c57ddc7bc9bd608a514984e497677a3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts#L89)):*

You could also reverse the statement and only call handler() if handler != null && (sendingOp || disabled)
</blockquote></details>

I don't think that's the same. I think it could be `handler != null && (!sendingOp || !this.featureFlags.preventOpAcknowledgement.enabled)`, but I find that much harder to reason about.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts` line 129 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-NQ1BXPf5UUT54MdhIH-:-NQ1s_xI2XA1lMC33nSJ:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/67a0e1909c57ddc7bc9bd608a514984e497677a3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts#L129)):*

Add types here.
</blockquote></details>

Done.

marksvc:
@Nateowami Do you have some suggested steps to cause the problem to happen, so we can test the fix? Ideally on a workstation.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1557)* status: 8 of 10 files reviewed, 5 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts` line 135 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-NQ25dMT1xQsrSzqW_R6:-NQ25dMT1xQsrSzqW_R7:bbyi1g) ([raw file](https://github.com/sillsdev/web-xforge/blob/18d8d0319699b802bbf322bd7822731c689f66c9/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts#L135)):*
> ```XML
>     }
>     if (sendingOp && hasStringProp(message, 'c') && hasStringProp(message, 'd')) {
>       await this.remoteStore.beforeSendOp(message['c'], message['d']);
> ```

@marksvc To test this:

1. Comment out this line which makes sure the op gets updated in IndexedDB before being sent
2. Use the feature flags to interfere with the connection. Specifically, allow the ops to reach the server, but don't allow the acknowledgement of the ops to return to the client. Then when you refresh the page the op will get submitted a second time and won't be identified as the one that was already submitted.

I suggest starting by playing with the feature flags and two browser windows side by side. This doesn't just affect texts; all realtime docs are affected. Make changes in one window and see whether they propagate to the other. I'd be happy to chat about this if you want more details.

RaymondLuong3 (commented):
I'll spend some time doing some testing today before I give it the approval.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-:-NQVlfI34NkzvihE1dR2:boq7u0a)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami and @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts` line 62 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-NQ1Ijzn2wdtdIUjqZS3:-NQVl44M0U53d-4McQTo:bhv4715) ([raw file](https://github.com/sillsdev/web-xforge/blob/67a0e1909c57ddc7bc9bd608a514984e497677a3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts#L62)):*

I did not investigate extending ReconnectingWebSocket, preferring composition over inheritance. JavaScript isn't object oriented, so even starting to think about what it means to extend a class and what the ramifications of that might be is difficult. This solution is verbose, but there are fewer gotchas.
</blockquote></details>

I agree, by creating a new class to interface with the remote store will keep the code more transparent. I'm happy with that approach.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts` line 89 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-NQ1A73I7_bW-c_AP5XR:-NQVlTYX5kPY636SFrpE:bimb0h) ([raw file](https://github.com/sillsdev/web-xforge/blob/67a0e1909c57ddc7bc9bd608a514984e497677a3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts#L89)):*

I don't think that's the same. I think it could be `handler != null && (!sendingOp || !this.featureFlags.preventOpAcknowledgement.enabled)`, but I find that much harder to reason about.
</blockquote></details>

Let's keep it as-is.

RaymondLuong3 (approved):
I can see the fix working to prevent the duplicate op from being submitted when op acknowledgement has been turned off. Good job!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-:-NQVuy8r2xtTZn_mIlhl:ba2m04n)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami and @pmachapman)

pmachapman (approved):
The code looks good, and I couldn't see any obvious bugs or API miscalls when comparing to ShareDB. I ran a few tests locally checking for concurrency issues, and all worked as expected.

marksvc:
I did some testing of this today. I sometimes commented out sharedb-realtime-remote-store.ts line
```ts
await this.remoteStore.beforeSendOp(message['c'], message['d']);
```
and sometimes did not.
I found that I needed to do quite a bit of browser Reloading in order for ops to go/come thru sometimes, but that may be because of getting the feature flags to fully turn on and off. 
With the line of code commented out, and "Prevent op acknowledgement" enabled, I can type a letter into a verse, turn off "Prevent op acknowledgement", and Reload, and then I see that the typed letter has been doubled up. 
This behaviour is not present when I uncomment the line of code.
Experimenting with the "Prevent op acknowledgement" and "Prevent op submission" feature flags, moving back and forth between chapters, and reloading, has failed to reveal any problems when the line of code is uncommented. 

So that's a good test result from what I was trying.

marksvc (commented):
I did not carefully scrutinize, but did read over the code. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-:-NQafRnsAyYmaXgIIiyN:b-dcuo65)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts` line 402 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-NQacCo-8EMWSfaHf6Hh:-NQacCo-8EMWSfaHf6Hi:b-q72a5o) ([raw file](https://github.com/sillsdev/web-xforge/blob/18d8d0319699b802bbf322bd7822731c689f66c9/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts#L402)):*
> ```XML
>     this.doc.pendingOps.push(
>       ...ops.map(component => {
>         const data = { op: component.op, type: this.doc.type, callbacks: [] };
> ```

Here and in realtime-doc.ts, you seem to be changing the structure of what is returned, as here you set `data.op` to `component.op`, rather than set `data.op` to `component`. 
Similarly in realtime-doc.ts, you change the return from an array of `this.prepareDataForStore(opInfo.op)` to an array of `{ op: this.prepareDataForStore(opInfo.op) }`.
What's with that?
I'm also a little confused why you set src and seq both here and in realtime-doc.ts.



_Code quote:_
```XML
t.op, 
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1557)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts` line 402 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-NQacCo-8EMWSfaHf6Hh:-NQlBZqV2-s6nTayQxwj:b-l2sdo9) ([raw file](https://github.com/sillsdev/web-xforge/blob/18d8d0319699b802bbf322bd7822731c689f66c9/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts#L402)):*

Here and in realtime-doc.ts, you seem to be changing the structure of what is returned, as here you set `data.op` to `component.op`, rather than set `data.op` to `component`. 
Similarly in realtime-doc.ts, you change the return from an array of `this.prepareDataForStore(opInfo.op)` to an array of `{ op: this.prepareDataForStore(opInfo.op) }`.
What's with that?
I'm also a little confused why you set src and seq both here and in realtime-doc.ts.

</blockquote></details>

If you look at what calls `updatePendingOps()`, it's only called by `RealtimeDoc.loadFromOfflineStore()`, when we're loading data out of IndexedDB into ShareDB.

ShareDB is meant to work in a single session, by  we're cryogenically freezing information about its state in IndexeDB, and then in a new session we're creating a new instance and giving it memories from its past life.

So we call `this.doc.pendingOps.push(...)` in this method to restore memory of pending ops. However, when we stored the data in IndexeDB we didn't store every single property. Notice that the doc type is inferred from `this.doc.type`, rather than from the data being restored. We don't need to store it on every doc. And callbacks can't be serialized.

The component may be something like (real example):
``` json
{
  "op": {
    "ops": [
      { "retain": 26 },
      {
        "insert": "a",
        "attributes": { "segment": "verse_1_4" }
      },
      { "delete": 1 }
    ]
  },
  "src": "338dc861291274a32db961b3857530f6",
  "seq": 24
}
```
This is what we stored when a user typed `a`. To "inject" it into ShareDB's in-memory state, we need to add the type of the doc and an empty array of callbacks (ShareDB might crash if the callback array isn't there, since it expects it to exist).

Rather than modify the object we've been passed (mutating objects we don't "own" could have bad consequences), we create a new one and copy over the ops, add the type and callbacks, and if the `src` and `seq` exist, add them as well. After this code change they really should always exist, but I was trying to be cautious. And actually if someone did an update of SF and still had ops in IndexedDB, they wouldn't have the `src` and `seq`. So there could be cases where `src` and/or `seq` don't exist, and we wouldn't want to accidentally set them to `undefined`.

I'm guessing with this explanation `prepareDataForStore()` in `realtime-doc.ts` now makes more sense as well? That's where we're trimming down the data to save (and adding the `src` if it's not on) without mutating the original object.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1557)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts` line 402 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-NQacCo-8EMWSfaHf6Hh:-NQlGaY37O5nLAbTM2AY:b1eewer) ([raw file](https://github.com/sillsdev/web-xforge/blob/18d8d0319699b802bbf322bd7822731c689f66c9/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts#L402)):*

If you look at what calls `updatePendingOps()`, it's only called by `RealtimeDoc.loadFromOfflineStore()`, when we're loading data out of IndexedDB into ShareDB.

ShareDB is meant to work in a single session, by  we're cryogenically freezing information about its state in IndexeDB, and then in a new session we're creating a new instance and giving it memories from its past life.

So we call `this.doc.pendingOps.push(...)` in this method to restore memory of pending ops. However, when we stored the data in IndexeDB we didn't store every single property. Notice that the doc type is inferred from `this.doc.type`, rather than from the data being restored. We don't need to store it on every doc. And callbacks can't be serialized.

The component may be something like (real example):
``` json
{
  "op": {
    "ops": [
      { "retain": 26 },
      {
        "insert": "a",
        "attributes": { "segment": "verse_1_4" }
      },
      { "delete": 1 }
    ]
  },
  "src": "338dc861291274a32db961b3857530f6",
  "seq": 24
}
```
This is what we stored when a user typed `a`. To "inject" it into ShareDB's in-memory state, we need to add the type of the doc and an empty array of callbacks (ShareDB might crash if the callback array isn't there, since it expects it to exist).

Rather than modify the object we've been passed (mutating objects we don't "own" could have bad consequences), we create a new one and copy over the ops, add the type and callbacks, and if the `src` and `seq` exist, add them as well. After this code change they really should always exist, but I was trying to be cautious. And actually if someone did an update of SF and still had ops in IndexedDB, they wouldn't have the `src` and `seq`. So there could be cases where `src` and/or `seq` don't exist, and we wouldn't want to accidentally set them to `undefined`.

I'm guessing with this explanation `prepareDataForStore()` in `realtime-doc.ts` now makes more sense as well? That's where we're trimming down the data to save (and adding the `src` if it's not on) without mutating the original object.
</blockquote></details>

If you wanted to see this for yourself when stepping through the debugger, all you need to do is:
1. Set a breakpoint here
2. Make sure some unsubmitted ops are in IndexdDB to load back into memory (Enable the preventOpSubmission feature flag, or go offline, and then do some edits)
3. Refresh the page
It should hit the breakpoint as it loads from IndexedDB.
The method that saves to IndexedDB should be hit without any feature flags enabled.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1557)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts` line 402 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1557#-NQacCo-8EMWSfaHf6Hh:-NR024ma1kxCYKidKyS9:b-lkkqfe) ([raw file](https://github.com/sillsdev/web-xforge/blob/18d8d0319699b802bbf322bd7822731c689f66c9/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/sharedb-realtime-remote-store.ts#L402)):*

If you wanted to see this for yourself when stepping through the debugger, all you need to do is:
1. Set a breakpoint here
2. Make sure some unsubmitted ops are in IndexdDB to load back into memory (Enable the preventOpSubmission feature flag, or go offline, and then do some edits)
3. Refresh the page
It should hit the breakpoint as it loads from IndexedDB.
The method that saves to IndexedDB should be hit without any feature flags enabled.
</blockquote></details>

Thank you. That's helpful. I wasn't understanding the relationship between the other code storing the data, and this code fetching and reconstructing the data.

Nateowami:
While this needs further testing by the test team, they can't test it until it's on QA, and the developers have tested it locally already.

PR #1566, opened 2022-10-27
format: ignore angular cache files

marksvc:
commit-id:65813f17

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1566)

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,4 +2,4 @@
 *.min.*
 
 # mostly 3rd-party code
-src/SIL.XForge*/wwwroot/lib/**/*
+/src/SIL.XForge*/wwwroot/lib
--- a/src/RealtimeServer/.prettierignore
+++ b/src/RealtimeServer/.prettierignore
@@ -1,2 +1,2 @@
-lib/**
-coverage/**
+/lib
+/coverage
--- a/src/SIL.XForge.Scripture/ClientApp/.prettierignore
+++ b/src/SIL.XForge.Scripture/ClientApp/.prettierignore
@@ -1,2 +1,3 @@
-coverage/**
-dist/**
+/coverage
+/dist
+/.angular/cache


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1566)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/.prettierignore` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1566#-NFROxrW-zwFQl_D_rp2:-NFROxrW-zwFQl_D_rp3:b-y1ri74) ([raw file](https://github.com/sillsdev/web-xforge/blob/d36c654c6f430d36bd314a8b353435a4c3efa83e/src/SIL.XForge.Scripture/ClientApp/.prettierignore#L3)):*
I realize you're following the pattern above, but according to the docs (https://prettier.io/docs/en/ignore.html):
> To exclude files from formatting, create a .prettierignore file in the root of your project. .prettierignore uses gitignore syntax.

So I think this whole file should be:
```
coverage
dist
.angular/cache
```

My guess is that the double asterisks were originally used because of this example from the linked documentation:
```
**/.git
**/.svn
**/.hg
**/node_modules
```
I just checked and this affects other .prettierignore files as well. Do you want to fix them all, or just this file, or just merge as-is?

marksvc:
*[`src/SIL.XForge.Scripture/ClientApp/.prettierignore` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1566#-NFROxrW-zwFQl_D_rp2:-NH0B5LY-GBOQuq2SL_7:b-jta089) ([raw file](https://github.com/sillsdev/web-xforge/blob/d36c654c6f430d36bd314a8b353435a4c3efa83e/src/SIL.XForge.Scripture/ClientApp/.prettierignore#L3)):*

I realize you're following the pattern above, but according to the docs (https://prettier.io/docs/en/ignore.html):
> To exclude files from formatting, create a .prettierignore file in the root of your project. .prettierignore uses gitignore syntax.

So I think this whole file should be:
```
coverage
dist
.angular/cache
```

My guess is that the double asterisks were originally used because of this example from the linked documentation:
```
**/.git
**/.svn
**/.hg
**/node_modules
```
I just checked and this affects other .prettierignore files as well. Do you want to fix them all, or just this file, or just merge as-is?
</blockquote></details>

I'm fixing them all. Hmm. After having a look over the git ignore syntax documentation, I have another adjustment to propose as well. When we write something like
```
coverage
dist
```
that doesn't just match `coverage` or `dist` in the current directory, but also in subdirectories. So if we ever introduced, say, `ClientApp/src/app/checking/coverage/coverage.component.ts`, that file would not be scrutinized by prettier. I prepended `/` to a number of items, since what we really mean is thus and such directory _in the current directory_.

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1566)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/.prettierignore` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1566#-NFROxrW-zwFQl_D_rp2:-NH0wX1S-R_uo0PiH4A0:b-pnbf3z) ([raw file](https://github.com/sillsdev/web-xforge/blob/d36c654c6f430d36bd314a8b353435a4c3efa83e/src/SIL.XForge.Scripture/ClientApp/.prettierignore#L3)):*

I'm fixing them all. Hmm. After having a look over the git ignore syntax documentation, I have another adjustment to propose as well. When we write something like
```
coverage
dist
```
that doesn't just match `coverage` or `dist` in the current directory, but also in subdirectories. So if we ever introduced, say, `ClientApp/src/app/checking/coverage/coverage.component.ts`, that file would not be scrutinized by prettier. I prepended `/` to a number of items, since what we really mean is thus and such directory _in the current directory_. 
</blockquote></details>

Interesting. Good to know. I'm pretty sure that means every .gitignore I've written was kind of wrong...

PR #1569, opened 2022-10-28
Add Node 18.12.0 to build configuration

Nateowami:
The goal of this is to find out if tests will pass with an updated version of Node. Actually updating will require changes to TeamCity, QA, live, etc. This is just an experiment.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1569)


marksvc:
Looks promising.

Nateowami:
Closing this. It was originally opened to see if the build would pass. I think we could move to Node 18 now, but unfortunately we don't have a good process for updating major dependencies so it's time consuming.

PR #1573, opened 2022-10-31
Clarify source of NullReferenceException

marksvc:
Such as from SF-1776.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1573)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -889,8 +889,32 @@ namespace SIL.XForge.Scripture.Services
                     {
                         continue;
                     }
+
+                    if (remotePtProject.SourceUsers is null)
+                    {
+                        throw new InvalidDataException(
+                            $"Unexpected null SourceUsers when working with project PT id {remotePtProject.SendReceiveId}."
+                        );
+                    }
+                    if (remotePtProject.SourceUsers.Users is null)
+                    {
+                        throw new InvalidDataException(
+                            $"Unexpected null SourceUsers.Users when working with project PT id {remotePtProject.SendReceiveId}."
+                        );
+                    }
                     string role = ConvertFromUserRole(
-                        remotePtProject.SourceUsers.Users.SingleOrDefault(u => u.UserName == projectUserName)?.Role
+                        remotePtProject.SourceUsers.Users
+                            .SingleOrDefault(u =>
+                            {
+                                if (u is null)
+                                {
+                                    _logger.LogWarning(
+                                        $"An element of SourceUsers.Users was null when working with project PT id {remotePtProject.SendReceiveId}."
+                                    );
+                                }
+                                return u?.UserName == projectUserName;
+                            })
+                            ?.Role
                     );
 
                     // Get the PT user ID


RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1573)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 895 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1573#-NFoeFxXE79ocFF_B-P7:-NFoeFxXE79ocFF_B-P8:bn92w36) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4f1853e8f36dccd6209a4784450362ffa6de997/src/SIL.XForge.Scripture/Services/ParatextService.cs#L895)):*
> ```Smalltalk
>                     if (remotePtProject.SourceUsers is null)
>                     {
>                         throw new InvalidDataException(
> ```

Is there a reason we need to throw an error here? Is it because if SourceUsers is null, then the project is not one that we can handle? Previously, no exceptions were thrown, and I imagine this would simply return an empty array of user mappings.



_Code quote:_
```Smalltalk
                    throw new InvalidDataException(
```

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1573)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 895 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1573#-NFoeFxXE79ocFF_B-P7:-NFoqNMmCCL6PqUxdyGS:b58j2ty) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4f1853e8f36dccd6209a4784450362ffa6de997/src/SIL.XForge.Scripture/Services/ParatextService.cs#L895)):*

Is there a reason we need to throw an error here? Is it because if SourceUsers is null, then the project is not one that we can handle? Previously, no exceptions were thrown, and I imagine this would simply return an empty array of user mappings.
</blockquote></details>

If `remotePtProject.SourceUsers` or `remotePtProject.SourceUsers.Users` are null, then a few lines below with `remotePtProject.SourceUsers.Users.SingleOrDefault(`, there would be a NullReferenceException. 
As to whether we should have an exception at all in this situation, or return an empty list, I lean toward exception because
  - If either of those properties are null, it seems that we received some data that is broken somehow. (Unless it's correct or expected behaviour for that to happen, then I might think that we should handle it without throwing.)
  - When we call GetProjectRolesAsync(), if we somehow fail to actually query the PT role list, an exception more clearly passes that error back than a statement of no roles.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1573)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1573)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1574, opened 2022-10-31
SF-1782 Enable insert note FAB for reviewer users

RaymondLuong3:
* add insert note FAB when reviewer selects a verse
* make FAB appear at the same y-coordinates as the segment
* underline a selected verse when a user clicks the text
* remove note thread styling for reviewer users
* TODO: improve experience on mobile view

Insert Note Fab

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1574)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
@@ -22,6 +22,10 @@ quill-editor {
   border: 1px solid $border-color;
 }
 
+.read-only-editor.selectable-verses usx-segment {
+  cursor: pointer;
+}
+
 .read-only-editor > .ql-container {
   > .ql-editor {
     usx-segment {
@@ -32,14 +36,17 @@ quill-editor {
       background-color: #ffeb3b;
     }
 
-    .question-segment {
+    .question-segment,
+    usx-segment.reviewer-selection {
       background-image: linear-gradient(to right, gray 75%, transparent 75%);
       background-position: bottom 0.2em left;
       background-repeat: repeat-x;
       background-size: 7px 1px;
       position: relative;
       cursor: pointer;
+    }
 
+    .question-segment {
       &[data-question-count] {
         &:before {
           content: '?';
@@ -136,15 +143,6 @@ quill-editor.rtl {
     background-size: 7px 1px;
     position: relative;
   }
-}
-
-.ql-editor .reviewer-selection {
-  background-image: linear-gradient(to right, gray 75%, transparent 75%);
-  background-position: bottom 0.2em left;
-  background-repeat: repeat-x;
-  background-size: 7px 1px;
-  position: relative;
-  cursor: pointer;
 }
 
 .template-editor > .ql-container {
--- a/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
@@ -138,6 +138,15 @@ quill-editor.rtl {
   }
 }
 
+.ql-editor .reviewer-selection {
+  background-image: linear-gradient(to right, gray 75%, transparent 75%);
+  background-position: bottom 0.2em left;
+  background-repeat: repeat-x;
+  background-size: 7px 1px;
+  position: relative;
+  cursor: pointer;
+}
+
 .template-editor > .ql-container {
   > .ql-editor {
     background-color: #f5f3ef;
--- a/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
@@ -382,7 +382,7 @@ usx-verse {
 span.reviewer-selection > usx-verse {
   span > span > span {
     color: white;
-    background-color: $purpleLight;
+    background-color: $greyDark;
   }
 }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss
@@ -1,3 +1,5 @@
+@import 'src/variables';
+
 .ql-editor {
   counter-reset: note;
 }
@@ -374,6 +376,13 @@ usx-verse {
       top: -0.5em;
       position: relative;
     }
+  }
+}
+
+span.reviewer-selection > usx-verse {
+  span > span > span {
+    color: white;
+    background-color: $purpleLight;
   }
 }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
@@ -646,6 +646,10 @@ export function registerScripture(): string[] {
     scope: Parchment.Scope.INLINE
   });
   formats.push(NoteThreadHighlightClass);
+  const ReviewerSelectedSegmentClass = new ClassAttributor('reviewer-selection', 'reviewer-selection', {
+    scope: Parchment.Scope.INLINE
+  });
+  formats.push(ReviewerSelectedSegmentClass);
 
   const InvalidBlockClass = new ClassAttributor('invalid-block', 'invalid-block', {
     scope: Parchment.Scope.BLOCK
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
@@ -645,6 +645,7 @@ export function registerScripture(): string[] {
   const NoteThreadHighlightClass = new ClassAttributor('note-thread-highlight', 'note-thread-highlight', {
     scope: Parchment.Scope.INLINE
   });
+
   formats.push(NoteThreadHighlightClass);
   const ReviewerSelectedSegmentClass = new ClassAttributor('reviewer-selection', 'reviewer-selection', {
     scope: Parchment.Scope.INLINE
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.html
@@ -14,7 +14,8 @@
     'template-editor': !readOnlyEnabled,
     ltr: isLtr,
     rtl: isRtl,
-    'mark-invalid': markInvalid
+    'mark-invalid': markInvalid,
+    'selectable-verses': selectableVerses
   }"
   [dir]="$any(textDirection)"
   [lang]="lang"
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.spec.ts
@@ -25,6 +25,7 @@ import { UserDoc } from 'xforge-common/models/user-doc';
 import { UserService } from 'xforge-common/user.service';
 import { DialogService } from 'xforge-common/dialog.service';
 import { LocalPresence } from 'sharedb/lib/sharedb';
+import { SFProjectRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-role';
 import { CaretPosition } from '../../../type-utils';
 import { SFProjectProfileDoc } from '../../core/models/sf-project-profile-doc';
 import { SF_TYPE_REGISTRY } from '../../core/models/sf-type-registry';
@@ -3160,14 +3161,14 @@ class TestEnvironment {
   }
 
   /** Where reference is like 'MAT 1:2'. */
-  embedThreadAt(reference: string, textAnchor: TextAnchor): void {
+  embedThreadAt(reference: string, textAnchor: TextAnchor, role: string = SFProjectRole.ParatextTranslator): void {
     const verseRef: VerseRef = VerseRef.parse(reference);
     const uniqueSuffix: string = Math.random().toString();
     const id: string = `embedid${reference}${uniqueSuffix}`;
     const iconSource: string = '--icon-file: url(/assets/icons/TagIcons/01flag1.png)';
     const text: string = `text message on ${id}`;
     const format = { iconsrc: iconSource, preview: text, threadid: id };
-    this.component.embedElementInline(verseRef, id, textAnchor, 'note-thread-embed', format);
+    this.component.embedElementInline(verseRef, id, role, textAnchor, 'note-thread-embed', format);
     this.component.toggleFeaturedVerseRefs(true, [verseRef], 'note-thread');
   }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -112,6 +112,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
   @Input() markInvalid: boolean = false;
   @Input() multiSegmentSelection = false;
   @Input() subscribeToUpdates = true;
+  @Input() selectableVerses: boolean = false;
   @Output() updated = new EventEmitter<TextUpdatedEvent>(true);
   @Output() segmentRefChange = new EventEmitter<string>();
   @Output() loaded = new EventEmitter(true);
@@ -425,7 +426,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
   }
 
   get firstVerseSegment(): string | undefined {
-    for (const [segmentRef] of this.viewModel.segments) {
+    for (const [segmentRef] of this.segments) {
       if (getBaseVerse(segmentRef) != null) {
         return segmentRef;
       }
@@ -696,11 +697,16 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
       selectionValue = formats['reviewer-selection'] ? null : true;
     }
 
-    const format: any = { ['reviewer-selection']: selectionValue };
+    const format: StringMap = { ['reviewer-selection']: selectionValue };
+    let verseEmbedFormatted: boolean = false;
     for (const segment of verseSegments) {
       const range: RangeStatic | undefined = this.getSegmentRange(segment);
       if (range != null) {
-        this.editor?.formatText(range.index, range.length, format, 'api');
+        if (!verseEmbedFormatted) {
+          this.editor.formatText(range.index - 1, 1, format, 'api');
+          verseEmbedFormatted = true;
+        }
+        this.editor.formatText(range.index, range.length, format, 'api');
       }
     }
     return selectionValue === true;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -12,7 +12,7 @@ import {
 import { TranslocoService } from '@ngneat/transloco';
 import isEqual from 'lodash-es/isEqual';
 import merge from 'lodash-es/merge';
-import Quill, { DeltaStatic, RangeStatic, Sources } from 'quill';
+import Quill, { DeltaStatic, RangeStatic, Sources, StringMap } from 'quill';
 import QuillCursors from 'quill-cursors';
 import { AuthType, getAuthType } from 'realtime-server/lib/esm/common/models/user';
 import { TextAnchor } from 'realtime-server/lib/esm/scriptureforge/models/text-anchor';
@@ -28,12 +28,13 @@ import { DialogService } from 'xforge-common/dialog.service';
 import { objectId } from 'xforge-common/utils';
 import tinyColor from 'tinycolor2';
 import { takeUntil } from 'rxjs/operators';
+import { SFProjectRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-role';
 import { Delta, TextDoc, TextDocId } from '../../core/models/text-doc';
 import { SFProjectService } from '../../core/sf-project.service';
 import { NoteThreadIcon } from '../../core/models/note-thread-doc';
 import { attributeFromMouseEvent, getBaseVerse } from '../utils';
 import { MultiCursorViewer } from '../../translate/editor/multi-viewer/multi-viewer.component';
-import { registerScripture } from './quill-scripture';
+import { getAttributesAtPosition, registerScripture } from './quill-scripture';
 import { Segment } from './segment';
 import { EditorRange, TextViewModel } from './text-view-model';
 import { TextNoteDialogComponent, NoteDialogData } from './text-note-dialog/text-note-dialog.component';
@@ -359,6 +360,10 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     return this._segment;
   }
 
+  get segments(): IterableIterator<[string, RangeStatic]> {
+    return this.viewModel.segments;
+  }
+
   get segmentText(): string {
     return this._segment == null ? '' : this._segment.text;
   }
@@ -416,6 +421,10 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     if (this._segment != null) {
       return this._segment.ref;
     }
+    return this.firstVerseSegment;
+  }
+
+  get firstVerseSegment(): string | undefined {
     for (const [segmentRef] of this.viewModel.segments) {
       if (getBaseVerse(segmentRef) != null) {
         return segmentRef;
@@ -597,6 +606,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
   embedElementInline(
     verseRef: VerseRef,
     id: string,
+    role: string,
     textAnchor: TextAnchor,
     formatName: string,
     format: any
@@ -660,8 +670,11 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     this.editor.insertEmbed(embedInsertPos, formatName, format, 'api');
     const textAnchorRange = this.viewModel.getEditorContentRange(embedInsertPos, textAnchor.length);
     const formatLength: number = textAnchorRange.editorLength;
-    insertFormat['text-anchor'] = 'true';
-    this.editor.formatText(embedInsertPos, formatLength, insertFormat, 'api');
+
+    if (role !== SFProjectRole.Reviewer) {
+      insertFormat['text-anchor'] = 'true';
+      this.editor.formatText(embedInsertPos, formatLength, insertFormat, 'api');
+    }
     this.updateSegment();
     return embedSegmentRef;
   }
@@ -671,6 +684,26 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     if (position != null && this.editor != null) {
       this.editor.formatText(position, 1, embedName, format, 'api');
     }
+  }
+
+  toggleVerseSelection(verseRef: VerseRef): boolean {
+    if (this.editor == null) return false;
+    const verseSegments: string[] = this.getVerseSegments(verseRef);
+    const verseRange: RangeStatic | undefined = this.getSegmentRange(verseSegments[0]);
+    let selectionValue: true | null = true;
+    if (verseRange != null) {
+      const formats: StringMap = getAttributesAtPosition(this.editor, verseRange.index);
+      selectionValue = formats['reviewer-selection'] ? null : true;
+    }
+
+    const format: any = { ['reviewer-selection']: selectionValue };
+    for (const segment of verseSegments) {
+      const range: RangeStatic | undefined = this.getSegmentRange(segment);
+      if (range != null) {
+        this.editor?.formatText(range.index, range.length, format, 'api');
+      }
+    }
+    return selectionValue === true;
   }
 
   /** Respond to text changes in the quill editor. */
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -32,7 +32,7 @@ import { SFProjectRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-
 import { Delta, TextDoc, TextDocId } from '../../core/models/text-doc';
 import { SFProjectService } from '../../core/sf-project.service';
 import { NoteThreadIcon } from '../../core/models/note-thread-doc';
-import { attributeFromMouseEvent, getBaseVerse } from '../utils';
+import { attributeFromMouseEvent, getBaseVerse, VERSE_REGEX } from '../utils';
 import { MultiCursorViewer } from '../../translate/editor/multi-viewer/multi-viewer.component';
 import { getAttributesAtPosition, registerScripture } from './quill-scripture';
 import { Segment } from './segment';
@@ -700,9 +700,12 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     const format: StringMap = { ['reviewer-selection']: selectionValue };
     let verseEmbedFormatted: boolean = false;
     for (const segment of verseSegments) {
+      // only underline the selection if it is part of the verse text i.e. not a section heading
+      if (!VERSE_REGEX.test(segment)) continue;
       const range: RangeStatic | undefined = this.getSegmentRange(segment);
       if (range != null) {
         if (!verseEmbedFormatted) {
+          // add the formatting to the verse embed on the first iteration
           this.editor.formatText(range.index - 1, 1, format, 'api');
           verseEmbedFormatted = true;
         }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts
@@ -75,7 +75,8 @@ export function verseRefFromMouseEvent(event: MouseEvent, bookNum: number): Vers
     return;
   }
   const segmentParts = clickSegment.split('_', 3);
-  return new VerseRef(bookNum, segmentParts[1], segmentParts[2]);
+  const versePart = segmentParts[2].split('/')[0];
+  return new VerseRef(bookNum, segmentParts[1], versePart);
 }
 
 export function threadIdFromMouseEvent(event: MouseEvent): string | undefined {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
@@ -86,7 +86,12 @@
         <div *ngIf="userHasGeneralEditRight && !hasChapterEditPermission" class="no-edit-permission-message">
           <mat-icon>info</mat-icon> {{ t("no_permission_edit_chapter", { userRole: userRoleStr }) }}
         </div>
-        <div #targetContainer class="text-container" [style.height]="textHeight">
+        <div
+          #targetContainer
+          class="text-container"
+          [style.height]="textHeight"
+          [dir]="isTargetRightToLeft ? 'rtl' : 'ltr'"
+        >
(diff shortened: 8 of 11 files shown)


nigel-wells (changes requested):
Nice to see this coming together. 
Here are some initial notes from testing:
- The FAB is always appearing at the top right of the screen rather than the segment as per your notes

- Can the text in a segment have `cursor: pointer` as well to indicate that it can be clicked on. I see the cursor is set once it is highlighted which is a good indication you can click again to unselect it.
- I'm not 100% sold on underlining a blank verse as it seems odd but I'm not sure of an alternative. Perhaps we could also have the verse number background color highlight as well - that would be a real nice indication you have the correct verse selected.
- As a translator, clicking on a verse that already has a note creates an additional underline

- As a translator, clicking a verse underlines the verse, clicking somewhere else in the same verse then removes the underline - I imagine we'd want it to remain unless you focused on another verse or the editor lost focus
- If the FAB is showing and you change chapters then it remains visible 

And a generic question around the reviewer role and whether or not this should be a permission assigned to roles? Will we have other roles in the future that might need the functionality?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-:-NFpgL6tATH4tt9WAOWy:b-9vxx6j)* status: all files reviewed, 10 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/_text.scss` line 142 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFpi7yZ5WKWsKUIPuZ_:-NFpi7yZ5WKWsKUIPuZa:b-57o6vx) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/_text.scss#L142)):*
> ```SCSS
> 
> .ql-editor .reviewer-selection {
>   background-image: linear-gradient(to right, gray 75%, transparent 75%);
> ```

Change to one of our theme variables for gray
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts` line 649 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFpiPHB9F_QV_Bl3dIX:-NFpiPHC4LUNmFO44YXO:bftwcjn) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts#L649)):*
> ```XML
>   });
>   formats.push(NoteThreadHighlightClass);
>   const ReviewerSelectedSegmentClass = new ClassAttributor('reviewer-selection', 'reviewer-selection', {
> ```

nit: needs a line break above it to separate it from the highlight class
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 428 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFplrIrB0IGnUy0TaiG:-NFplrIrB0IGnUy0TaiH:b-by0w7v) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L428)):*
> ```XML
> 
>   get firstVerseSegment(): string | undefined {
>     for (const [segmentRef] of this.viewModel.segments) {
> ```

Could take advantage of your new `segments` getter created
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 696 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFppHdG2zxDnte9Pur0:-NFppHdG2zxDnte9Pur1:b-f1jj10) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L696)):*
> ```XML
>     if (verseRange != null) {
>       const formats: StringMap = getAttributesAtPosition(this.editor, verseRange.index);
>       selectionValue = formats['reviewer-selection'] ? null : true;
> ```

As noted in the summary, this causes an issue for a translator who might click in the same verse multiple times and I don't think it should unselect the verse
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 699 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFpnenb8MRiUbhGMWcX:-NFpnenb8MRiUbhGMWcY:b-mbyfu9) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L699)):*
> ```XML
>     }
> 
>     const format: any = { ['reviewer-selection']: selectionValue };
> ```

Should be `StringMap` type instead of `any`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss` line 145 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFprKMXAZnIQZNmDWyW:-NFprKMXAZnIQZNmDWyX:bkx1oqg) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss#L145)):*
> ```SCSS
> .insert-note-fab {
>   position: fixed;
>   right: 5px;
> ```

Will need some work to support RTL
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 151 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFptlVTDv25vDwXpsjx:-NFptlVTDv25vDwXpsjy:b-w3gdez) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L151)):*
> ```XML
>   private toggleNoteThreadSub?: Subscription;
>   private shouldNoteThreadsRespondToEdits: boolean = false;
>   private showFabOnClick: boolean = false;
> ```

I wonder if this could be a getter instead that checked if `target === true` and not currently inserting a note i.e. the dialog is open. It will save having to remember to set this to true and false each time a particular event happens and allows us to define its visibility in one place.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 775 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFpuaf5ATznYp6sMLIs:-NFpuaf5ATznYp6sMLIt:bvuq1j) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L775)):*
> ```XML
>     if (this.userRole === SFProjectRole.Reviewer) {
>       let verseRef: VerseRef | undefined = this.reviewerSelectedVerseRef;
>       if (this.reviewerSelectedVerseRef == null) {
> ```

This can check the now defined `verseRef`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1217 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFpyn1K8DLaMrBozZYv:-NFpyn1K8DLaMrBozZYw:b-73gsqy) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1217)):*
> ```XML
>               this.target.toggleVerseSelection(this.reviewerSelectedVerseRef);
>               this.reviewerSelectedVerseRef = verseRef;
>             }
> ```

Is this here to unselect any other verse currently selected?



_Code quote:_
```XML
          if (this.reviewerSelectedVerseRef != null) {
            if (verseRef.equals(this.reviewerSelectedVerseRef)) {
              this.reviewerSelectedVerseRef = undefined;
            } else {
              this.target.toggleVerseSelection(this.reviewerSelectedVerseRef);
              this.reviewerSelectedVerseRef = verseRef;
            }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1372 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFpvp8q3TDjU7sCoAdd:-NFpvp8q3TDjU7sCoAde:bxpgkdx) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1372)):*
> ```XML
>   }
> 
>   private setupInsertNoteFab(): void {
> ```

As mentioned, something in here isn't working for me and the top position isn't being set - the method isn't even called

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1574)* status: all files reviewed, 10 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss` line 145 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFprKMXAZnIQZNmDWyW:-NFq122o--OuyIKtlfXz:b-62ovxr) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss#L145)):*

Will need some work to support RTL
</blockquote></details>

I think the CSS logical property you are looking for is [inset-inline-end](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline-end):
``` css
inset-inline-end: 5px;
```
Instead of `right: 5px;`

RaymondLuong3 (commented):
Nice suggestions. I incorporated those into this second commit. Translator users were not supposed to be able to select verses in the way that reviewers are supposed to, so I added a guard for that. Thanks for the thorough testing!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-:-NGIRAdY-ss9810p_epd:b-j78vvz)* status: all files reviewed, 9 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/_text.scss` line 142 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFpi7yZ5WKWsKUIPuZ_:-NGIBz2ADDVk0OiXFBaW:b-8p7j6j) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/_text.scss#L142)):*

Change to one of our theme variables for gray
</blockquote></details>

It looks like the `gray 75%` part does something so the background image isn't actually a gradient. When I swap that out for `$greyDark` the gradient is visible, and it doesn't look as nice.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 428 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFplrIrB0IGnUy0TaiG:-NGICYh3CkeTFVeVQ9uk:bo5cpot) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L428)):*

Could take advantage of your new `segments` getter created
</blockquote></details>

Nice :)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 696 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFppHdG2zxDnte9Pur0:-NGICbSb6tZBgxiIt3o7:b-so71vw) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L696)):*

As noted in the summary, this causes an issue for a translator who might click in the same verse multiple times and I don't think it should unselect the verse
</blockquote></details>

Thanks. I have put a guard in so that this formatting only applies to reviewers
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 699 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFpnenb8MRiUbhGMWcX:-NGIDbPrEhBaXnqbXyI6:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L699)):*

Should be `StringMap` type instead of `any`
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss` line 145 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFprKMXAZnIQZNmDWyW:-NGIDjImCN8shFfy1las:b8go2vq) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss#L145)):*

I think the CSS logical property you are looking for is [inset-inline-end](https://developer.mozilla.org/en-US/docs/Web/CSS/inset-inline-end):
``` css
inset-inline-end: 5px;
```
Instead of `right: 5px;`
</blockquote></details>

Good catch. Thanks.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 151 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFptlVTDv25vDwXpsjx:-NGIL6lQ266HQN6XS2D8:bljtp2o) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L151)):*

I wonder if this could be a getter instead that checked if `target === true` and not currently inserting a note i.e. the dialog is open. It will save having to remember to set this to true and false each time a particular event happens and allows us to define its visibility in one place.
</blockquote></details>

Good thinking. I introduced a member called `isNoteDialogOpen`, and the `canShowInsertNoteFab` getter checks `isNoteDialogOpen` and `targetLoaded`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 775 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFpuaf5ATznYp6sMLIs:-NGIMKIkCuvcD4pgWkDY:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L775)):*

This can check the now defined `verseRef`
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1217 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFpyn1K8DLaMrBozZYv:-NGIMMcW50_aa8VW4BOT:bo7fnp8) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1217)):*

Is this here to unselect any other verse currently selected?
</blockquote></details>

That's right. I'll leave a comment since it isn't as obvious.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1372 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFpvp8q3TDjU7sCoAdd:-NGIMkhk0tKXZ8e5rnyR:b-lcqptq) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1372)):*

As mentioned, something in here isn't working for me and the top position isn't being set - the method isn't even called
</blockquote></details>

Hmm, I wonder if it has to do with the document query for the `reviewer-selection`. It may be a race condition. I updated it so that the event listener on the verse will make this calculation.

Nateowami (commented):
Great job. This looks really good!

It looks like when you switch to a different chapter the fab disappears, but not when you switch to a different book. My guess is that this is related to the fact that the segment identifier (e.g. `verse_2_2`) specifies the chapter, but not the book.

Another thing I noticed while testing (which is much more minor) this is that if you scroll after clicking a verse, the fab icon stays in the same place. I'm not sure there's a great solution to this. I just wanted to mention it, though we don't need to address it in this PR.

Also, you can select things and attempt to add a note to something that isn't a verse.


If I'm not mistaken, at this point we're only supporting adding a note to a specific verse (not text within in a verse, or a range of verses).

Later on in the chapter see the selection including both the verse and the title that comes after it, though this case is better since it will correctly add the note to verse 14.


I'm also seeing behavior where sometimes I can select multiple verses, though this is harder to do and isn't consistent.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-:-NGItILHDKLPCnAzD3JB:b-wihpr0)* status: 2 of 10 files reviewed, 9 unresolved discussions (waiting on @nigel-wells)

RaymondLuong3 (commented):
I updated this change so that when switching books, the fab icon disappears. I also needed to clear the reviewerSelectedVerseRef property which then fixed the issue with multiple verses selected problem.
I changed the behaviour so that segments that are not part of verses cannot be selected, and they do not show the underline when the verse before is clicked.
As for the floating icon, I'm not sure what is better, to have it scroll with the page, or left at the y-coordinate where it first shows on screen. I can experiment with it more in a separate PR.

@nigel-wells I think I know why the button was not moving on your screen as expected. You might not have had the insert note feature flag turned on. Previously I was not hiding the FAB if the feature was turned off, so the fact that you could see it was a mistake. If you turn the feature on now, you should see the FAB moving as expected (I think).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-:-NGTGHbr341LsOFs_SYv:bhgfyel)* status: all files reviewed, 6 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/_text.scss` line 142 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFpi7yZ5WKWsKUIPuZ_:-NGTH5zX9rrfQNZdl2Ww:b-pu4jp9) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/_text.scss#L142)):*

Ok
</blockquote></details>

It looks like using the gradient function was an easy way to generate a background image but setting the `gray 75%` somehow voided the gradient.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/_usx.scss` line 385 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NGOlK1QAUX8oq7DMcOu:-NGTI5EU2hXV3w6IVd3l:b-wzcd70) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4bda687e1749266c7feb5a28aec0e8a8462538f/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss#L385)):*

Love this! Did you want to change the background color as suggested the other day - perhaps just reverse the colors i.e. dark background with light text?
</blockquote></details>

I used a dark gray, and it looks good.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.html` line 18 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NGOu8-qDr1_CG9xJuMk:-NGTIdLn-VEt670FWUvz:b-of2w67) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4bda687e1749266c7feb5a28aec0e8a8462538f/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.html#L18)):*

Just noting this here in case it is worth considering. Are we better to have an overarching CSS rule where we define what "should" be clickable or should we rather set this per element that can be clickable? I raise the question in case there might be some elements that can't/shouldn't be clicked on which would make it better for each element to state if it can be clicked on or not. That would add a heap more styles for sure.
</blockquote></details>

That is a good thought. Right now, a usx-segment is clickable in the translate editor for reviewers, and generally in the checking editor. So far this seems to work well. But there may be a time when we want to set a more general rule as to which segments are clickable.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 708 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NGOy1Oq6weayfpCnZHZ:-NGTK6jvB8sGkZZ4mH1R:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4bda687e1749266c7feb5a28aec0e8a8462538f/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L708)):*

A comment on why this needs to happen would be nice
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss` line 145 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFprKMXAZnIQZNmDWyW:-NGTYTyoDP6LWMG1L1nx:b-poe8hx) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss#L145)):*

This works for the UI in RTL but not if the project text is RTL which I think also needs to trigger the FAB on the other side.
</blockquote></details>

I have made it work with rtl now. It took some tweaking, and now I calculate the FAB element's left value based the target container's bounding client rectangle.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 151 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFptlVTDv25vDwXpsjx:-NGTXqYp39UP57hT2TYH:b-kifexx) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L151)):*

Thanks, this works better. I wonder if we could take it one step further where the `dialogService` controls the state of a dialog open or not. This means we can have the same logic regardless of what dialog might open i.e. share dialog, error dialog, some other dialog in the future, etc. `canShowInserNoteFab()` can then check the service if a dialog is open or not.
</blockquote></details>

It looks like dialog service already has a property for the number of open dialogs. I have used that now!

nigel-wells (changes requested):
I think my underlaying issue was I wasn't viewing it as a Reviewer - I'm pretty sure I had turned on the feature flag. It is much better now.

Personally, I'd expect the FAB to be stuck to the selected verse in the same way we have the green line indicated on a selected verse when editing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-:-NGYLXAg6GfLEnfwqrzW:b1uq11w)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/_usx.scss` line 385 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NGOlK1QAUX8oq7DMcOu:-NGYHVQACgxDI2S0lT-e:bb70w76) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4bda687e1749266c7feb5a28aec0e8a8462538f/src/SIL.XForge.Scripture/ClientApp/src/_usx.scss#L385)):*

I used a dark gray, and it looks good.
</blockquote></details>

Nice
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss` line 145 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFprKMXAZnIQZNmDWyW:-NGYJ9eK7aDs5LoN1Z8i:b-he6vt9) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss#L145)):*

I have made it work with rtl now. It took some tweaking, and now I calculate the FAB element's left value based the target container's bounding client rectangle.
</blockquote></details>

Might need to tweak this some more - possibly just changed the position to `absolute` will help. If there is a source project it currently sits all the way on the left of the source rather than the target

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 151 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFptlVTDv25vDwXpsjx:-NGYIdRZ74THptGEobzC:b-hxjyw2) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L151)):*

It looks like dialog service already has a property for the number of open dialogs. I have used that now!
</blockquote></details>

Awesome
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1271 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NGYJXdoAbQAqP459Q7Y:-NGYJXdp47sZn1vcTHMB:bz9lqau) ([raw file](https://github.com/sillsdev/web-xforge/blob/55402ccdd888764fdcbb667a4b16e060c85abb8b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1271)):*
> ```XML
>   }
> 
>   private resetInsertNoteFab(): void {
> ```

You may need to tweak this based on my other comment around the CSS being `fixed` instead of `absolute`. A quick test changing it to `absolute` brought it closer to where I'd expect but now the `left` pixels are probably too much.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 2458 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NGYKRMS3ZIhAkrMZDlD:-NGYKRMS3ZIhAkrMZDlE:b-pp285i) ([raw file](https://github.com/sillsdev/web-xforge/blob/55402ccdd888764fdcbb667a4b16e060c85abb8b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L2458)):*
> ```XML
>       expect(verse2Elem.classList).not.toContain('reviewer-selection');
> 
>       verse2Elem.click();
> ```

Why the need to click verse 2 before changing params and then immediately clicking verse 3? Should there be a check before clicking verse 3? I'm guessing this is to address Nathanial's point in which case we should check to ensure verse_1_2 isn't still selected before clicking verse 3.

RaymondLuong3 (commented):
I think I also agree that it would be nice if the FAB was fixed in the text too, but I think that will need some additional work in another PR.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-:-NGre51cCBxPIU4_5J1h:brnq027)* status: all files reviewed, 3 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss` line 145 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFprKMXAZnIQZNmDWyW:-NGrlZU06ZhbIWoRTrXG:b-ame4m1) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss#L145)):*

Might need to tweak this some more - possibly just changed the position to `absolute` will help. If there is a source project it currently sits all the way on the left of the source rather than the target


</blockquote></details>

I did do some tweaking, it seems that it was working in my firefox browser, but in chrome I was more likely to see the behaviour you were seeing. I implemented a 10ms delay for when we calculate the target container coordinates, and it seems to fix the issue. The `position: absolute` suggestion made the icon offscreen in firefox, but what I did seems to work for both.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1271 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NGYJXdoAbQAqP459Q7Y:-NGrm2hE-11oRsTYMxhn:b-q64pkt) ([raw file](https://github.com/sillsdev/web-xforge/blob/55402ccdd888764fdcbb667a4b16e060c85abb8b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1271)):*

You may need to tweak this based on my other comment around the CSS being `fixed` instead of `absolute`. A quick test changing it to `absolute` brought it closer to where I'd expect but now the `left` pixels are probably too much.
</blockquote></details>

See above.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 2458 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NGYKRMS3ZIhAkrMZDlD:-NGrXLOl--FB1wg8bZNy:b-sg9ufq) ([raw file](https://github.com/sillsdev/web-xforge/blob/55402ccdd888764fdcbb667a4b16e060c85abb8b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L2458)):*

Why the need to click verse 2 before changing params and then immediately clicking verse 3? Should there be a check before clicking verse 3? I'm guessing this is to address Nathanial's point in which case we should check to ensure verse_1_2 isn't still selected before clicking verse 3.
</blockquote></details>

The issue was that if we selected a verse, then moved to a new book, the previous verse was still selected, and selecting a new verse results in both verses being underlined. Good point about checking that verse 2 is unselected, I'll add that in.

nigel-wells (changes requested):
> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1574)* status: all files reviewed, 4 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html` line 126 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NGs1hDz01AbYoCEhqrm:-NGs1hDz01AbYoCEhqrn:b76jm8s) ([raw file](https://github.com/sillsdev/web-xforge/blob/378fb308d14f0a454f5e2d11396fb2d283d78e14/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html#L126)):*
> ```HTML
>           ></app-suggestions>
>           <div #fabButton class="insert-note-fab" [style.left]="insertNoteFabLeft">
>             <button mat-mini-fab title="{{ t('insert_note') }}" (click)="insertNote()">
> ```

Good usability addition there.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss` line 145 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NFprKMXAZnIQZNmDWyW:-NGs0FAdCHnuYfV3B5NF:bmgrs9n) ([raw file](https://github.com/sillsdev/web-xforge/blob/08ac9a33066041f8c139f126626daeb3349c9379/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss#L145)):*

I did do some tweaking, it seems that it was working in my firefox browser, but in chrome I was more likely to see the behaviour you were seeing. I implemented a 10ms delay for when we calculate the target container coordinates, and it seems to fix the issue. The `position: absolute` suggestion made the icon offscreen in firefox, but what I did seems to work for both.
</blockquote></details>

Thanks - that has resolved the issue for me.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 209 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NGs0tVj20p2kovnMtCT:-NGs0tVkA_UhK0YuTBLQ:b8d089g) ([raw file](https://github.com/sillsdev/web-xforge/blob/378fb308d14f0a454f5e2d11396fb2d283d78e14/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L209)):*
> ```XML
>     if (this.projectUserConfigDoc != null && this.isTargetTextRight !== value) {
>       this.projectUserConfigDoc.submitJson0Op(op => op.set(puc => puc.isTargetTextRight, value));
>       this.resetInsertNoteFab();
> ```

Good thinking. It looks like the verse still stays selected so you'll need to unselect that as well.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 425 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NGs1B9G2gSF-4mQRRuw:-NGs1B9G2gSF-4mQRRux:b-hrbq9p) ([raw file](https://github.com/sillsdev/web-xforge/blob/378fb308d14f0a454f5e2d11396fb2d283d78e14/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L425)):*
> ```XML
>   ngAfterViewInit(): void {
>       this.setTextHeight();
> ```

Same issue with target text changing from left to right where the FAB is reset but the verse stays selected.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1574)* status: all files reviewed, 4 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html` line 126 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NGs1hDz01AbYoCEhqrm:-NH0ipyi2qIHLQPrRWhl:b-zcjvte) ([raw file](https://github.com/sillsdev/web-xforge/blob/378fb308d14f0a454f5e2d11396fb2d283d78e14/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html#L126)):*

Good usability addition there.
</blockquote></details>

Thanks
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 209 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NGs0tVj20p2kovnMtCT:-NH0o1rH9-O8jyzcGhzS:b-i21bbo) ([raw file](https://github.com/sillsdev/web-xforge/blob/378fb308d14f0a454f5e2d11396fb2d283d78e14/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L209)):*

Good thinking. It looks like the verse still stays selected so you'll need to unselect that as well.
</blockquote></details>

Thanks. Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 425 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-NGs1B9G2gSF-4mQRRuw:-NH0nv269YVING967d5F:b-sexisl) ([raw file](https://github.com/sillsdev/web-xforge/blob/378fb308d14f0a454f5e2d11396fb2d283d78e14/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L425)):*

Same issue with target text changing from left to right where the FAB is reset but the verse stays selected.
</blockquote></details>

Oh yes, you're right. That makes sense to me, to have the verse unselected if you modify the window.

nigel-wells (approved):
I'm ok with that - let's keep it moving along.

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1574#-:-NH1I_zz5OI1oAY8QebH:b-1ejbty)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1576, opened 2022-11-02
SF-1556 Prevent excessive queries from the front end

Nateowami:
As far as I can tell, the main issue is that connecting a basic project (Stp22 in this case) created about 85,000+ question subscribe queries* for questions and 1,500,000+ text fetches

<sup>*A subscribe query is distinct from a fetch query in that the subscriber is asking for all future changes to the query in real time, which is presumably a much more expensive operation.</sup>

The problem is that rather than update the page with whatever changes come over the wire, we tend to completely recalculate everything from scratch every time something is updated. As an example, in the app component, every time the project doc was updated it would unsubscribe from any existing question queries and then create a new query for each book in the project.

A simple fix, which I have applied here, is to throttle incoming changes, but I don't think it's the ideal solution. I think it would be better to react to what actually changed, rather than re-initializing everything, though that's harder to do.

The `feature/realtime-server-stats` branch can be used to analyze realtime service usage and better see what's going on.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1576)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -171,7 +171,7 @@ describe('AppComponent', () => {
     expect(env.menuLength).toEqual(5);
     const projectDoc = env.component.projectDocs![0];
     projectDoc.submitJson0Op(op => op.set<boolean>(p => p.checkingConfig.checkingEnabled, false));
-    env.wait();
+    env.waitForProjectDocChanges();
     // Expect: Translate | Overview | Matthew | Mark | Synchronize | Settings | Users
     expect(env.menuLength).toEqual(7);
     // No affect when clicking Translate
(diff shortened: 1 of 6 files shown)


nigel-wells (approved):
:lgtm:

> Reviewed 3 of 6 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1576#-:-NG-egYpAGivs2e_AX4Y:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami:
I've changed this to `testing-not-required`. I plan to merge it and send it to QA to be thoroughly tested there. If for some reason it turns out out's terrible, we can revert it. We don't normally do this, but it's a fairly big problem, and there isn't anything else significant to send to QA for testing.

PR #1577, opened 2022-11-04
SF-1805 Improve performance of questions query in app component

Nateowami:
- Reverts temporary app component changes from #1576 (the `throttleTime`, as well as changes to the tests)
- Improved the performance of determining which books have questions

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1577)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -171,7 +171,7 @@ describe('AppComponent', () => {
     expect(env.menuLength).toEqual(5);
     const projectDoc = env.component.projectDocs![0];
     projectDoc.submitJson0Op(op => op.set<boolean>(p => p.checkingConfig.checkingEnabled, false));
-    env.waitForProjectDocChanges();
+    env.wait();
     // Expect: Translate | Overview | Matthew | Mark | Synchronize | Settings | Users
     expect(env.menuLength).toEqual(7);
     // No affect when clicking Translate
@@ -251,7 +251,7 @@ describe('AppComponent', () => {
     env.init();
 
     expect(env.selectedProjectId).toEqual('project01');
-    env.removesUserFromProject('project01');
+    env.removeUserFromProject('project01');
     verify(mockedMdcDialog.open(ProjectDeletedDialogComponent, anything())).once();
     env.confirmProjectDeletedDialog();
     // Get past setTimeout to navigation
@@ -511,7 +511,7 @@ describe('AppComponent', () => {
 
       const projectId = 'project01';
       expect(env.selectedProjectId).toEqual(projectId);
-      env.removesUserFromProject(projectId);
+      env.removeUserFromProject(projectId);
       verify(mockedSFProjectService.localDelete(projectId)).once();
     }));
   });
@@ -564,10 +564,9 @@ class TestEnvironment {
 
     this.realtimeService.addSnapshots<Question>(QuestionDoc.COLLECTION, []);
     when(mockedSFProjectService.queryQuestions(anything(), anything())).thenCall((_projectId, options) => {
-      const parameters: QueryParameters = {
-        [obj<Question>().pathStr(q => q.verseRef.bookNum)]: options.bookNum,
-        [obj<Question>().pathStr(q => q.isArchived)]: false
-      };
+      const parameters: QueryParameters = {};
+      if (options.bookNum != null) parameters[obj<Question>().pathStr(q => q.verseRef.bookNum)] = options.bookNum;
+      if (options.activeOnly) parameters[obj<Question>().pathStr(q => q.isArchived)] = false;
       return this.realtimeService.subscribeQuery(QuestionDoc.COLLECTION, parameters);
     });
 
@@ -819,12 +818,6 @@ class TestEnvironment {
     flush(70);
   }
 
-  // Project doc changes are throttled by 1000 ms, so we have to wait for them
-  waitForProjectDocChanges(): void {
-    tick(1000);
-    this.wait();
-  }
-
   deleteProject(projectId: string, isLocal: boolean): void {
     if (isLocal) {
       when(mockedUserService.currentProjectId(anything())).thenReturn(undefined);
@@ -836,23 +829,22 @@ class TestEnvironment {
     this.wait();
   }
 
-  removesUserFromProject(projectId: string): void {
+  removeUserFromProject(projectId: string): void {
     const projectDoc = this.realtimeService.get<SFProjectProfileDoc>(SFProjectProfileDoc.COLLECTION, projectId);
     projectDoc.submitJson0Op(op => op.unset<string>(p => p.userRoles['user01']), false);
-    this.waitForProjectDocChanges();
+    this.wait();
   }
 
   setLastSyncSuccessful(projectId: string, lastSyncSuccessful: boolean): void {
     const projectDoc = this.realtimeService.get<SFProjectProfileDoc>(SFProjectProfileDoc.COLLECTION, projectId);
     projectDoc.submitJson0Op(op => op.set<boolean>(p => p.sync.lastSyncSuccessful!, lastSyncSuccessful));
-    this.waitForProjectDocChanges();
+    this.wait();
   }
 
   addUserToProject(projectId: string): void {
     const projectDoc = this.realtimeService.get<SFProjectProfileDoc>(SFProjectProfileDoc.COLLECTION, projectId);
     projectDoc.submitJson0Op(op => op.set<string>(p => p.userRoles['user01'], SFProjectRole.CommunityChecker), false);
     this.currentUserDoc.submitJson0Op(op => op.add<string>(u => u.sites['sf'].projects, 'project04'), false);
-    this.waitForProjectDocChanges();
   }
 
   confirmProjectDeletedDialog() {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
@@ -14,8 +14,8 @@ import { AuthType, getAuthType, User } from 'realtime-server/lib/esm/common/mode
 import { SFProjectRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-role';
 import { TextInfo } from 'realtime-server/lib/esm/scriptureforge/models/text-info';
 import { Canon } from 'realtime-server/lib/esm/scriptureforge/scripture-utils/canon';
-import { combineLatest, merge, Observable, Subscription } from 'rxjs';
-import { distinctUntilChanged, filter, map, startWith, tap, throttleTime } from 'rxjs/operators';
+import { combineLatest, Observable, Subscription } from 'rxjs';
+import { distinctUntilChanged, filter, map, startWith, tap } from 'rxjs/operators';
 import { AuthService } from 'xforge-common/auth.service';
 import { DataLoadingComponent } from 'xforge-common/data-loading-component';
 import { DialogService } from 'xforge-common/dialog.service';
@@ -88,7 +88,8 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
   private selectedProjectDeleteSub?: Subscription;
   private removedFromProjectSub?: Subscription;
   private _isDrawerPermanent: boolean = true;
-  private readonly questionCountQueries = new Map<number, RealtimeQuery<QuestionDoc>>();
+  private communityCheckingBooks: number[] = [];
+  private questionsQuery?: RealtimeQuery<QuestionDoc>;
 
   constructor(
     private readonly router: Router,
@@ -399,21 +400,15 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
         if (this.removedFromProjectSub != null) {
           this.removedFromProjectSub.unsubscribe();
         }
-        // TODO Find a better solution than merely throttling remote changes
-        this.removedFromProjectSub = this.selectedProjectDoc.remoteChanges$.pipe(throttleTime(1000)).subscribe(() => {
+        this.removedFromProjectSub = this.selectedProjectDoc.remoteChanges$.subscribe(() => {
           if (
-            this.selectedProjectDoc != null &&
-            this.selectedProjectDoc.data != null &&
+            this.selectedProjectDoc?.data != null &&
             this.currentUserDoc != null &&
             !(this.currentUserDoc.id in this.selectedProjectDoc.data.userRoles)
           ) {
             // The user has been removed from the project
             this.showProjectDeletedDialog();
             this.projectService.localDelete(this.selectedProjectDoc.id);
-          }
-          // See if we need to enable any books in the checking app
-          if (this.isCheckingEnabled && !this.checkingVisible) {
-            this.checkCheckingBookQuestions();
           }
         });
 
@@ -427,14 +422,13 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
           this._projectSelect.value = this.selectedProjectDoc.id;
         }
 
-        this.checkCheckingBookQuestions();
         this.checkDeviceStorage();
       });
 
-      this.subscribe(
-        projectId$.pipe(filter(id => id != null)),
-        async projectId => await this.userService.setCurrentProjectId(this.currentUserDoc!, projectId)
-      );
+      this.subscribe(projectId$.pipe(filter(id => id != null)), projectId => {
+        this.refreshQuestionsQuery(projectId);
+        this.userService.setCurrentProjectId(this.currentUserDoc!, projectId);
+      });
     }
     this.loadingFinished();
   }
@@ -447,7 +441,7 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
     if (this.removedFromProjectSub != null) {
       this.removedFromProjectSub.unsubscribe();
     }
-    this.disposeQuestionQueries();
+    this.questionsQuery?.dispose();
   }
 
   setLocale(locale: string) {
@@ -540,8 +534,7 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
   }
 
   hasQuestions(text: TextInfo): boolean {
-    const query = this.questionCountQueries.get(text.bookNum);
-    return query != null && query.count > 0;
+    return this.communityCheckingBooks.includes(text.bookNum);
   }
 
   reloadWithUpdates(): void {
@@ -556,40 +549,24 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
     return this.selectedProjectDoc?.data?.sync.lastSyncSuccessful === false;
   }
 
-  private async checkCheckingBookQuestions(): Promise<void> {
-    this.disposeQuestionQueries();
-    if (!this.isCheckingEnabled || this.selectedProjectDoc === undefined) {
-      return;
-    }
-    const promises: Promise<any>[] = [];
-    for (const text of this.texts) {
-      promises.push(
-        this.projectService
-          .queryQuestions(this.selectedProjectDoc.id, {
-            bookNum: text.bookNum,
-            activeOnly: true
-          })
-          .then(query => this.questionCountQueries.set(text.bookNum, query))
-      );
-    }
-    await Promise.all(promises);
-    this.questionCountQueries.forEach((query: RealtimeQuery<QuestionDoc>, bookNum: number) => {
-      this.subscribe(merge(query.remoteChanges$, query.localChanges$, query.ready$), () => {
-        if (this.selectedProjectDoc == null) {
-          return;
-        }
-        if (query.count > 0) {
-          this.selectedProjectDoc.loadTextDocs(bookNum);
-        }
-      });
-    });
-  }
+  private async refreshQuestionsQuery(projectId: string): Promise<void> {
+    this.questionsQuery?.dispose();
 
-  private disposeQuestionQueries(): void {
-    for (const questionQuery of this.questionCountQueries.values()) {
-      questionQuery.dispose();
-    }
-    this.questionCountQueries.clear();
+    this.questionsQuery = await this.projectService.queryQuestions(projectId, { activeOnly: true });
+    this.questionsQuery.docs$.subscribe(docs => {
+      const books = new Set<number>();
+      for (const question of docs) {
+        const bookNum = question.data?.verseRef.bookNum;
+        if (bookNum != null) books.add(bookNum);
+      }
+      // Subscribe to the texts of any added book, so it will be available offline. This really shouldn't be the concern
+      // of the app component, but for now it is. There isn't an easy way to unsubscribe from a book that was removed,
+      // and it's not extremely important to do so, so we won't bother doing that.
+      for (const bookNum of books) {
+        if (!this.communityCheckingBooks.includes(bookNum)) this.selectedProjectDoc?.loadTextDocs(bookNum);
+      }
+      this.communityCheckingBooks = Array.from(books).sort((a, b) => a - b);
+    });
   }
 
   private async showProjectDeletedDialog(): Promise<void> {


nigel-wells (changes requested):
Great job on this - much more efficient and compact.

Just one query around offline support to consider.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-:-NGDvgTr0bHFUVI8CO-I:b62sb0l)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc and @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 556 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGDv8QBAdSfvQDuSvRt:-NGDv8QBAdSfvQDuSvRu:b3ms0br) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L556)):*
> ```XML
> 
>     const query = await this.projectService.queryQuestions(projectId, { activeOnly: true });
>     this.questionsQuery = query;
> ```

nit: `query` is not needed and you can just set straight onto `this.questionsQuery`



_Code quote:_
```XML
    const query = await this.projectService.queryQuestions(projectId, { activeOnly: true });
    this.questionsQuery = query;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 582 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGDw2mA6I-TbFwwPqH0:-NGDw2mA6I-TbFwwPqH1:bm507ce) ([raw file](https://github.com/sillsdev/web-xforge/blob/a9d7caa44f4d05df67cd7149c82dc83025f36989/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L582)):*
> ```XML
>         }
>         if (query.count > 0) {
>           this.selectedProjectDoc.loadTextDocs(bookNum);
> ```

Did we still want to do this as it was a nice side affect to help support community checking in offline mode while we don't have an interface to choose what gets downloaded.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1577)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 556 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGDv8QBAdSfvQDuSvRt:-NGEk-SQCzKD6cRSLnDc:btp4ym2) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L556)):*

nit: `query` is not needed and you can just set straight onto `this.questionsQuery`
</blockquote></details>

Good point. I'll fix it.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 582 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGDw2mA6I-TbFwwPqH0:-NGEjrvB60vy9Etv6dXx:bctgg4w) ([raw file](https://github.com/sillsdev/web-xforge/blob/a9d7caa44f4d05df67cd7149c82dc83025f36989/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L582)):*

Did we still want to do this as it was a nice side affect to help support community checking in offline mode while we don't have an interface to choose what gets downloaded.
</blockquote></details>

Oh, good point. I missed this somehow. What do you think about just querying for all texts on the project?

nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1577)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc and @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 582 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGDw2mA6I-TbFwwPqH0:-NGExm2BDRZ5SgRk_8Bp:bj42wya) ([raw file](https://github.com/sillsdev/web-xforge/blob/a9d7caa44f4d05df67cd7149c82dc83025f36989/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L582)):*

Oh, good point. I missed this somehow. What do you think about just querying for all texts on the project?
</blockquote></details>

From a community checkers POV they only need the text for the books that have questions. Main consideration if we do all text is how much data that would download to a phone and if that would create an issue that they're unable to resolve on their own. I think go with minimal effort to achieve the same/similar functionality we already have as we'll no doubt address this more when we properly address offline support next year.

marksvc:
:eyes:

marksvc (changes requested):
Thanks for your work on this. :clap: 
I approve, except for my blocking comment, and offline texts that Nigel brought up.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-:-NGIIt6S9VkedWXaxBHg:bcc00hm)* status: all files reviewed, 6 unresolved discussions (waiting on @marksvc, @Nateowami, and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 295 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGI3JEJ-ZD3etlkdyKK:-NGI3JEJ-ZD3etlkdyKL:b-nbrjbb) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L295)):*
> ```XML
>   async ngOnInit(): Promise<void> {
>     this.loadingStarted();
>     if (await this.isLoggedIn) {
> ```

Wow, it looks like this could make use of a serious if-inversion. 



_Code quote:_
```XML
if (await this.isLoggedIn) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 407 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGI26LH9O_YYGIJQwkE:-NGI26LH9O_YYGIJQwkF:b-tsx4c0) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L407)):*
> ```XML
>             this.selectedProjectDoc?.data != null &&
>             this.currentUserDoc != null &&
>             !(this.currentUserDoc.id in this.selectedProjectDoc.data.userRoles)
> ```

It seems like if there is a remote change and the current user is _no longer on the project_, then that user wouldn't have permission to _receive_ the remote change about it :-). 



_Code quote:_
```XML
   !(this.currentUserDoc.id in this.selectedProjectDoc.data.userRoles)
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 428 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGI8QGv5LscGQWOI3hf:-NGI8QGv5LscGQWOI3hg:b80ys1n) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L428)):*
> ```XML
>       });
> 
>       this.subscribe(projectId$.pipe(filter(id => id != null)), async projectId => {
> ```

Previously, this code had a lambda saying `async projectId => await this...setCurrentProjectId(`. 
For my own understanding:  **In the context at hand** (where this is getting called from a subscribe), was the `await` without any influence on behaviour, and the behaviour would be the same without the `await`?



_Code quote:_
```XML
async 
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 428 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGII42P2A0ly4SSj5Y5:-NGII42P2A0ly4SSj5Y6:bn4z8f1) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L428)):*
> ```XML
>       });
> 
>       this.subscribe(projectId$.pipe(filter(id => id != null)), async projectId => {
> ```

At present, the lambda has no `await`s in it, but we do specify the lambda as `async`. Is there any use or significance of specifying that the lambda is `async`? Is it going to be the same as if we omitted `async`?



_Code quote:_
```XML
async projectId
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 430 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGI3jbUC2h1-xvW0y1p:-NGI3jbUC2h1-xvW0y1q:bknmzml) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L430)):*
> ```XML
>       this.subscribe(projectId$.pipe(filter(id => id != null)), async projectId => {
>         this.refreshQuestionsQuery(projectId);
>         this.userService.setCurrentProjectId(this.currentUserDoc!, projectId);
> ```

Hmm. I wonder if there would be any significance for a situation where a user is removed from their last SF project, that we set their current project id to null/undefined, instead of leaving it as the no-longer-really-valid id of the last project that they were on. 
Just thinking out loud. 

(This is not related to something your patch is doing.)



_Code quote:_
```XML
      this.subscribe(projectId$.pipe(filter(id => id != null)), async projectId => {
        this.refreshQuestionsQuery(projectId);
        this.userService.setCurrentProjectId(this.currentUserDoc!, projectId);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 557 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGI4hOLA1L8cXxj0tUx:-NGI4hOLA1L8cXxj0tUy:b-hkmc1r) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L557)):*
> ```XML
>     const query = await this.projectService.queryQuestions(projectId, { activeOnly: true });
>     this.questionsQuery = query;
>     query.docs$.subscribe(docs => {
> ```

I'm so glad this had an API comment over in realtime-query.ts :-)



_Code quote:_
```XML
.docs$
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 559 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGI7QQK9mMfgZcnkHeL:-NGI7QQK9mMfgZcnkHeM:besxg2a) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L559)):*
> ```XML
>     query.docs$.subscribe(docs => {
>       const books = new Set<number>();
>       for (const question of docs) {
> ```

I think this could benefit from some throttling, because if I add 1000 questions to a project, then this might run 1000 times, each time iterating over all questions in the project and building the book number set (and so run the inner part of the loop  1000+999+998+...+1 times). This seems like something that could benefit from a debounce or throttling. Do you think?
Not that, as is, it's going to be slower than the previous version of the  code. 



_Code quote:_
```XML
    const query = await this.projectService.queryQuestions(projectId, { activeOnly: true });
    this.questionsQuery = query;
    query.docs$.subscribe(docs => {
      const books = new Set<number>();
      for (const question of docs) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts` line 569 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGICKQzDS40dBGYQ0qi:-NGICKQzDS40dBGYQ0qj:btt7zwp) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts#L569)):*
> ```XML
>       const parameters: QueryParameters = {};
>       if (options.bookNum != null) parameters[obj<Question>().pathStr(q => q.verseRef.bookNum)] = options.bookNum;
>       if (options.activeOnly) parameters[obj<Question>().pathStr(q => q.isArchived)] = false;
> ```

I had to think about this one for a bit to decide if it was doing the right thing :)



_Code quote:_
```XML
     if (options.activeOnly) parameters[obj<Question>().pathStr(q => q.isArchived)] = false;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts` line 855 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGIEfqS6BzeakWYBA2a:-NGIEfqS6BzeakWYBA2b:b-8f2je9) ([raw file](https://github.com/sillsdev/web-xforge/blob/a9d7caa44f4d05df67cd7149c82dc83025f36989/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts#L855)):*
> ```XML
>     projectDoc.submitJson0Op(op => op.set<string>(p => p.userRoles['user01'], SFProjectRole.CommunityChecker), false);
>     this.currentUserDoc.submitJson0Op(op => op.add<string>(u => u.sites['sf'].projects, 'project04'), false);
>     this.waitForProjectDocChanges();
> ```

I see that in your throttling commit, you introduced this line, where there was no `this.wait()`. But this sort of method seems like something that would meaningfully benefit from a detectChanges() (such as via `this.wait()`), as being added to  a project should result in a change to the UI (such as a new project in the drop-down projects list). Do you think?



_Code quote:_
```XML
   this.waitForProjectDocChanges();
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1577)* status: 1 of 2 files reviewed, 5 unresolved discussions (waiting on @marksvc and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 295 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGI3JEJ-ZD3etlkdyKK:-NGIOUPYCimn1IfNAqr8:bvftvb2) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L295)):*

Wow, it looks like this could make use of a serious if-inversion. 
</blockquote></details>

Good point. Probably this shouldn't be a conditional, but `await this.uponUserLoggedIn`.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 407 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGI26LH9O_YYGIJQwkE:-NGIjkfJ44gBIre7MXP7:btugd1i) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L407)):*

It seems like if there is a remote change and the current user is _no longer on the project_, then that user wouldn't have permission to _receive_ the remote change about it :-). 
</blockquote></details>

Yeah, that is kind of weird. It would be good to investigate this at some point.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 428 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGI8QGv5LscGQWOI3hf:-NGIjx2d5VZjg0LKDxOq:b-5ired1) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L428)):*

Previously, this code had a lambda saying `async projectId => await this...setCurrentProjectId(`. 
For my own understanding:  **In the context at hand** (where this is getting called from a subscribe), was the `await` without any influence on behaviour, and the behaviour would be the same without the `await`?



</blockquote></details>

As we discussed, I don't think it had any effect except possibly on stack traces.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 428 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGII42P2A0ly4SSj5Y5:-NGIk35y62teSCcDqR4m:bbq2mkz) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L428)):*

At present, the lambda has no `await`s in it, but we do specify the lambda as `async`. Is there any use or significance of specifying that the lambda is `async`? Is it going to be the same as if we omitted `async`?
</blockquote></details>

I just forgot to remove it. Fixed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 430 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGI3jbUC2h1-xvW0y1p:-NGIkAeeE2a8rYcB-M9D:b-o7dvvd) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L430)):*

Hmm. I wonder if there would be any significance for a situation where a user is removed from their last SF project, that we set their current project id to null/undefined, instead of leaving it as the no-longer-really-valid id of the last project that they were on. 
Just thinking out loud. 

(This is not related to something your patch is doing.)
</blockquote></details>

As discussed on the call, I don't think this will really be an issue.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 556 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGDv8QBAdSfvQDuSvRt:-NGIk8XI0VwRYSE_Ac3U:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L556)):*

Good point. I'll fix it.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 557 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGI4hOLA1L8cXxj0tUx:-NGIT73w-fN94OjruPrf:b5xdmrv) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L557)):*

I'm so glad this had an API comment over in realtime-query.ts :-)
</blockquote></details>

Glad it was helpful.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 559 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGI7QQK9mMfgZcnkHeL:-NGIVybiDrRy_NA-rBdy:b7nea0k) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L559)):*

I think this could benefit from some throttling, because if I add 1000 questions to a project, then this might run 1000 times, each time iterating over all questions in the project and building the book number set (and so run the inner part of the loop  1000+999+998+...+1 times). This seems like something that could benefit from a debounce or throttling. Do you think?
Not that, as is, it's going to be slower than the previous version of the  code. 
</blockquote></details>

Well, possibly. At present adding questions is fairly slow, so debouncing wouldn't tend to do very much in my opinion. At present import speeds (which just increased dramatically due to these efficiency improvements), this would be firing an average of every 22ms when importing 1000 questions.

As a side note, the number of operations can be generalized as `(n*(n+1))/2`, so `(1000*(1000+1))/2` = 500,500. (I always forget the exact formula, and had way too much fun deriving it geometrically on paper). 

The checking overview component, on the other hand, is extraordinarily inefficient by comparison, sorting and filtering the question list many times _for every digest cycle_. It used to be much worse, because the number of digest cycles was out of control.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 582 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGDw2mA6I-TbFwwPqH0:-NGIkHn95mbo-TC2h8zK:b68u1ng) ([raw file](https://github.com/sillsdev/web-xforge/blob/a9d7caa44f4d05df67cd7149c82dc83025f36989/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L582)):*

From a community checkers POV they only need the text for the books that have questions. Main consideration if we do all text is how much data that would download to a phone and if that would create an issue that they're unable to resolve on their own. I think go with minimal effort to achieve the same/similar functionality we already have as we'll no doubt address this more when we properly address offline support next year.
</blockquote></details>

Done. The new solution I came up with is hopefully temporary.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts` line 855 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGIEfqS6BzeakWYBA2a:-NGIOO-CB5tOWndkuLOd:b1o9uzo) ([raw file](https://github.com/sillsdev/web-xforge/blob/a9d7caa44f4d05df67cd7149c82dc83025f36989/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts#L855)):*

I see that in your throttling commit, you introduced this line, where there was no `this.wait()`. But this sort of method seems like something that would meaningfully benefit from a detectChanges() (such as via `this.wait()`), as being added to  a project should result in a change to the UI (such as a new project in the drop-down projects list). Do you think?
</blockquote></details>

Yes, though it didn't appear to be necessary.

nigel-wells (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-:-NGJ-PQR-0bhnUtK12pb:bnfp4nl)* status: all files reviewed, 5 unresolved discussions (waiting on @marksvc and @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 559 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGI7QQK9mMfgZcnkHeL:-NGIxQ1NB8fVyBoF4uYg:b3p2ilu) ([raw file](https://github.com/sillsdev/web-xforge/blob/654adbb98dd09ef72bff2f4be68edd2bae2d01b2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L559)):*

Well, possibly. At present adding questions is fairly slow, so debouncing wouldn't tend to do very much in my opinion. At present import speeds (which just increased dramatically due to these efficiency improvements), this would be firing an average of every 22ms when importing 1000 questions.

As a side note, the number of operations can be generalized as `(n*(n+1))/2`, so `(1000*(1000+1))/2` = 500,500. (I always forget the exact formula, and had way too much fun deriving it geometrically on paper). 

The checking overview component, on the other hand, is extraordinarily inefficient by comparison, sorting and filtering the question list many times _for every digest cycle_. It used to be much worse, because the number of digest cycles was out of control.
</blockquote></details>

It's actually may be a bit more. My observation from testing is a change to a question (archiving and un-archiving) results in 2x remote changes. I haven't look closely at importing notes - I think it just triggered it once but I wasn't testing multiple browsers either. We could consider only setting/unsetting books but that may take more compute time then just clearing it out and adding them back in. What we could look for importing is some kind of "pause" on change events until the import is complete.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 582 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGDw2mA6I-TbFwwPqH0:-NGIzjglCVNGmXtlu7Bl:b-u5wrpo) ([raw file](https://github.com/sillsdev/web-xforge/blob/a9d7caa44f4d05df67cd7149c82dc83025f36989/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L582)):*

Done. The new solution I came up with is hopefully temporary.
</blockquote></details>

Looks good and nice supporting comment.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1577)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 582 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1577#-NGDw2mA6I-TbFwwPqH0:-NGN-COJ8UFkhTEqosSp:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/a9d7caa44f4d05df67cd7149c82dc83025f36989/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L582)):*

I am amused by "but for now it is" :)
</blockquote></details>

Done.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1577)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1518, opened 2022-09-27
SF-1750 SF-1017 Fix infinite loop when user doc updated

Nateowami:
This should also fix SF-1017.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1518)

PR #1578, opened 2022-11-05
SF-1816 Fix app component observable type and behavior

Nateowami:
- A type should have been declared as `string | undefined`, but was instead declared as `string`.
- The observable for the project ID was subscribed two twice, making all its internal logic run twice, when it should have only run once.

More explanations for the changes have been commented on their respective lines.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1578)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
@@ -14,7 +14,7 @@ import { AuthType, getAuthType, User } from 'realtime-server/lib/esm/common/mode
 import { SFProjectRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-role';
 import { TextInfo } from 'realtime-server/lib/esm/scriptureforge/models/text-info';
 import { Canon } from 'realtime-server/lib/esm/scriptureforge/scripture-utils/canon';
-import { combineLatest, Observable, Subscription } from 'rxjs';
+import { combineLatest, Observable, of, Subscription } from 'rxjs';
 import { distinctUntilChanged, filter, map, startWith, tap } from 'rxjs/operators';
 import { AuthService } from 'xforge-common/auth.service';
 import { DataLoadingComponent } from 'xforge-common/data-loading-component';
@@ -319,7 +319,7 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
       // retrieve the projectId from the current route. Since the nav menu is outside of the router outlet, it cannot
       // use ActivatedRoute to get the params. Instead the nav menu, listens to router events and traverses the route
       // tree to find the currently activated route
-      const projectId$ = this.router.events.pipe(
+      const projectId$: Observable<string | undefined> = this.router.events.pipe(
         filter(e => e instanceof NavigationEnd),
         startWith(null),
         map(() => {
@@ -342,18 +342,21 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
             }
           }
         }),
-        map(r => r.params['projectId'] as string),
+        map(r => r.params['projectId'] as string | undefined),
         distinctUntilChanged(),
         tap(projectId => {
-          this.canSeeSettings$ = this.settingsAuthGuard.allowTransition(projectId);
-          this.canSeeUsers$ = this.usersAuthGuard.allowTransition(projectId);
-          this.canSync$ = this.syncAuthGuard.allowTransition(projectId);
+          this.canSeeSettings$ = projectId == null ? of(false) : this.settingsAuthGuard.allowTransition(projectId);
+          this.canSeeUsers$ = projectId == null ? of(false) : this.usersAuthGuard.allowTransition(projectId);
+          this.canSync$ = projectId == null ? of(false) : this.syncAuthGuard.allowTransition(projectId);
           this.canSeeAdminPages$ = combineLatest([this.canSeeSettings$, this.canSeeUsers$, this.canSync$]).pipe(
             map(([settings, users, sync]) => settings || users || sync)
           );
           // the project deleted dialog should be closed by now, so we can reset its ref to null
           if (projectId == null) {
             this.projectDeletedDialogRef = null;
+          } else {
+            this.userService.setCurrentProjectId(this.currentUserDoc!, projectId);
+            this.refreshQuestionsQuery(projectId);
           }
         })
       );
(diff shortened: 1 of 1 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1578)* status: 0 of 1 files reviewed, all discussions resolved

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 344 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1578#-NG4H9IW-3LlqFxgDMHX:-NG4H9IW-3LlqFxgDMHY:b-el6zr4) ([raw file](https://github.com/sillsdev/web-xforge/blob/521fbb892dd7796f682616f29486715c255ca568/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L344)):*
> ```XML
>           }
>         }),
>         map(r => r.params['projectId'] as string | undefined),
> ```

The previous type assertion here was wrong. A lot of places after here depend on the fact that the value might be nullish. 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 347 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1578#-NG4Hd-CC1nEwOzU5vlq:-NG4Hd-CC1nEwOzU5vlr:bhtxqkn) ([raw file](https://github.com/sillsdev/web-xforge/blob/521fbb892dd7796f682616f29486715c255ca568/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L347)):*
> ```XML
>         distinctUntilChanged(),
>         tap(projectId => {
>           this.canSeeSettings$ = projectId == null ? of(false) : this.settingsAuthGuard.allowTransition(projectId);
> ```

Previously we were passing `undefined` to `this.settingsAuthGuard.allowTransition()`, which apparently actually worked. It's not supposed to take `undefined` as an argument though, and once we stopped misleading the type system (a few lines above), this line (and a few following) had errors.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 357 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1578#-NG4I1wi1VxhjXr8HbgN:-NG4I1wi1VxhjXr8HbgO:bdnhj6j) ([raw file](https://github.com/sillsdev/web-xforge/blob/521fbb892dd7796f682616f29486715c255ca568/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L357)):*
> ```XML
>             this.projectDeletedDialogRef = null;
>           } else {
>             this.userService.setCurrentProjectId(this.currentUserDoc!, projectId);
> ```

Previously we had `this.subscribe(projectId$, ... )` further down in the file, and within that subscription called `userService.setCurrentProjectId()`. It's been moved up here because the previous "solution" was causing all of this code to run twice.

[Observables are like functions, and subscribing to one is like calling a function](https://rxjs.dev/guide/observable#observables-as-generalizations-of-functions), something I apparently forgot about a month ago in #1518, when I added an additional subscribe below.

When we do `const projectId$ = this.router.events.pipe( ... )`, we are creating an Observable, but we are not doing anything with it. None of the things in `pipe( ... )` are run until we call `this.subscribe(combineLatest([projectDocs$, projectId$]), ... )` a little later. However, subscribing to it twice makes all of it run twice. 

If this is at all confusing, I strongly recommend reading the linked article from the official RxJS documentation.

RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1578)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 348 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1578#-NGIhRsc9SMxIoWKpzet:-NGIhRsc9SMxIoWKpzeu:b-a18zw) ([raw file](https://github.com/sillsdev/web-xforge/blob/521fbb892dd7796f682616f29486715c255ca568/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L348)):*
> ```XML
>         tap(projectId => {
>           this.canSeeSettings$ = projectId == null ? of(false) : this.settingsAuthGuard.allowTransition(projectId);
>           this.canSeeUsers$ = projectId == null ? of(false) : this.usersAuthGuard.allowTransition(projectId);
> ```

Would it not make more sense to add a null guard to `allowTransition`?



_Code quote:_
```XML
this.usersAuthGuard.allowTransition(projectId);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 357 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1578#-NG4I1wi1VxhjXr8HbgN:-NGIifuR2MzPupGBVxYZ:b-5rzr0x) ([raw file](https://github.com/sillsdev/web-xforge/blob/521fbb892dd7796f682616f29486715c255ca568/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L357)):*

Previously we had `this.subscribe(projectId$, ... )` further down in the file, and within that subscription called `userService.setCurrentProjectId()`. It's been moved up here because the previous "solution" was causing all of this code to run twice.

[Observables are like functions, and subscribing to one is like calling a function](https://rxjs.dev/guide/observable#observables-as-generalizations-of-functions), something I apparently forgot about a month ago in #1518, when I added an additional subscribe below.

When we do `const projectId$ = this.router.events.pipe( ... )`, we are creating an Observable, but we are not doing anything with it. None of the things in `pipe( ... )` are run until we call `this.subscribe(combineLatest([projectDocs$, projectId$]), ... )` a little later. However, subscribing to it twice makes all of it run twice. 

If this is at all confusing, I strongly recommend reading the linked article from the official RxJS documentation.
</blockquote></details>

Interesting. I wouldn't have thought that subscribing to an observable was like calling a function, but the article you gave was simple to understand. Thanks for the tip.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1578)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 348 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1578#-NGIhRsc9SMxIoWKpzet:-NGIlJfJ0cSMnNg__Omg:b-lvfsr7) ([raw file](https://github.com/sillsdev/web-xforge/blob/521fbb892dd7796f682616f29486715c255ca568/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L348)):*

Would it not make more sense to add a null guard to `allowTransition`?
</blockquote></details>

My solution here is definitely odd and not ideal, but it seems odd to me to modify `allowTransition` to force it to take a nullish value. But maybe it's preferable to this mess?

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1578)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami:
I'm going to wait for #1577 to be tested and merged before sending this for testing. The changes will need to be reconciled, and that will actually produce a functional difference. So testing them in parallel is not actually a true test.

PR #1582, opened 2022-11-07
dev: clarify and adjust variable names

marksvc:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1582)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts
@@ -150,9 +150,9 @@ export class SFProjectService extends ProjectService<SFProject, SFProjectDoc> {
     await this.realtimeService.create<NoteThreadDoc>(NoteThreadDoc.COLLECTION, docId, noteThread);
   }
 
-  queryNoteThreads(id: string): Promise<RealtimeQuery<NoteThreadDoc>> {
+  queryNoteThreads(sfProjectId: string): Promise<RealtimeQuery<NoteThreadDoc>> {
     const queryParams: QueryParameters = {
-      [obj<NoteThread>().pathStr(t => t.projectRef)]: id,
+      [obj<NoteThread>().pathStr(t => t.projectRef)]: sfProjectId,
       [obj<NoteThread>().pathStr(t => t.status)]: NoteStatus.Todo
     };
     return this.realtimeService.subscribeQuery(NoteThreadDoc.COLLECTION, queryParams);
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
@@ -1236,9 +1236,9 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
     }
   }
 
-  private async loadNoteThreadDocs(projectId: string): Promise<void> {
+  private async loadNoteThreadDocs(sfProjectId: string): Promise<void> {
     this.noteThreadQuery?.dispose();
-    this.noteThreadQuery = await this.projectService.queryNoteThreads(projectId);
+    this.noteThreadQuery = await this.projectService.queryNoteThreads(sfProjectId);
     this.toggleNoteThreadSub?.unsubscribe();
     this.toggleNoteThreadSub = this.subscribe(
       merge(this.toggleNoteThreadVerseRefs$, this.noteThreadQuery.ready$, this.noteThreadQuery.remoteChanges$),
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -324,15 +324,15 @@ namespace SIL.XForge.Scripture.Services
         /// "the Send and Receive server".
         /// False if there is a problem with authorization or connecting to the PT Archives.
         /// </returns>
-        public async Task<bool> CanUserAuthenticateToPTArchivesAsync(string userSFId)
+        public async Task<bool> CanUserAuthenticateToPTArchivesAsync(string sfUserId)
         {
-            if (string.IsNullOrWhiteSpace(userSFId))
+            if (string.IsNullOrWhiteSpace(sfUserId))
             {
-                throw new ArgumentException(nameof(userSFId));
+                throw new ArgumentException(nameof(sfUserId));
             }
 
             IInternetSharedRepositorySource ptRepoSource = await GetInternetSharedRepositorySource(
-                userSFId,
+                sfUserId,
                 CancellationToken.None
             );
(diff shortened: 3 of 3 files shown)


Nateowami (commented):
I'm curious how complete this change was. Did you try to update every potentially ambiguous project and user variable name?

> Reviewed 2 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1582#-:-NGJZdlA0z8hgix0U5lX:b-xn64a)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 495 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1582#-NGJZUvf0LEZapDvep8b:-NGJZUvf0LEZapDvep8c:bejyqg8) ([raw file](https://github.com/sillsdev/web-xforge/blob/17f3588ef5e0ff6b69adb27c6b2eccc9b970952d/src/SIL.XForge.Scripture/Services/ParatextService.cs#L495)):*
> ```Smalltalk
>         /// </summary>
>         /// <param name="paratextId">The paratext resource identifier.</param>
>         /// <param name="userSFId">The user identifier.</param>
> ```

Can you update this description as well?

marksvc (commented):
No. I tend to clarify the variable names when I am frustrated that I don't know what kind of id something is without going and reading the source. Specific instances of ambiguity, or perhaps nearby items, got attention, rather than sweeping changes. I seem to recall that Ira may not have been excited about making a bunch of changes for variable names to specify which kind of id they were, and so I have not been very aggressive in making the changes.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1582#-:-NGOT65y611FQvkh_kgu:bxn4lz3)* status: 2 of 3 files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 495 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1582#-NGJZUvf0LEZapDvep8b:-NGOT4Eo-Jq2EDWnsOeG:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/17f3588ef5e0ff6b69adb27c6b2eccc9b970952d/src/SIL.XForge.Scripture/Services/ParatextService.cs#L495)):*

Can you update this description as well?
</blockquote></details>

Done.

Nateowami (commented):
Makes sense. I can see in some contexts it seems like it would be redundant. Actually a couple of them seem that way (e.g. `sf-project.service.ts`), especially on the front end, I would expect the ID to be the SF ID, not the Paratext project ID.

What do you think? Where do we draw the line? Certainly I think when we pass around Paratext project IDs it needs to be very clear.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1582#-:-NGOVJwmCwXJwG7bAGSU:b-gxdf4h)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

marksvc:
Here's the thing with sf-project.service.ts. This file has a method called queryNoteThreads(id: string). Is `id` a note id? A thread id? From the plural `s` in `queryNoteThreads` I might be able to guess that I should be passing a project id, to receive the note threads for that project. But when I'm working with code and I have real puzzles to solve, I'm not very helped by needing to make guesses or go read our methods just to understand our own API, in cases where we could have just specified what a parameter is to begin with. 

My frustrations, detail-oriented nature, and displeasure of ambiguity don't need to result in this particular change being applied to the repository :). But it's where the motivation for it comes from.

For editor.component.ts loadNoteThreadDocs(projectId: string), it's true that a consideration of its context, and the plausibility of whether a particular kind of project id would be used, could lead to a good assumption that it is the project SF id. 
I'm looking at my git history to see what I was doing at the time. I was working in C# to add timer code for performance measuring of note fetching, and seeing how I could add something to the frontend that could trigger a stress test of this backend code, and also looking into pausing realtime server to simulate being offline. I probably first puzzled over the `id` in sf-project.service.ts `queryNoteThreads(id)`.  The editor.component.ts `loadNoteThreadDocs(projectId)` is connected to queryNoteThreads(id), calling it. I may have actually used editor.component.ts loadNoteThreadDocs(projectId) to help me understand what the `id` in `queryNoteThreads(id)` meant, as `loadNoteThreadDocs` was passing a project id. Being in the presence of need for making method arguments plain, and working in the backend where project PT ids more commonly mix with project SF ids, were probably contributing to further clarifying the argument in editor.component.ts loadNoteThreadDocs(projectId) , being connected to queryNoteThreads(id). 

> What do you think? Where do we draw the line? 

I say _make it plain_. I too often come across classes or methods that have no comments and I don't really understand what they are doing. Clarifying ids is a low-hanging-fruit way to make the meaning of parts of our codebase more plan, where it may not have been as quickly evident before. 
I think there is reason to use `id` as an argument name. And it may not be obvious what will be confusing to future-me or another developer. Tho maybe when another developer (or future-me) thinks the code could use illumination, it could be applied. Then we don't all need to check (perhaps more than once over time) if something like the `id` in `queryNoteThreads(id)` is a thread id or a project SF id.

The PR doesn't need to go thru like this if you're not persuaded :-).

nigel-wells (commented):
I'm happy with this approach @marksvc - make it clear. I just need to remind future me to make sure I do that :) 
One thing we did on another PR recently was also ensure consistency with variables names. We found about 8 variations for the exact same type of variable across many methods - very confusing. I do see in your PR that perhaps we could make a similar decision around how we label these kind of variables. For example, there is `projectSFId` where `SF` is after `project`. In other places there is `sfProject` where `sf` is lowercase and before `Project`. We know what they mean but it would still be nice to be consistent. 

Thoughts?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1582#-:-NGUc7qk92v3FI4th5wh:b-s3dda0)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

Nateowami (commented):
I think Nigel has a good point about consistency. Personally I prefer `sfProjectId`.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1582#-:-NGUt1ql-0iRBHs8HdYQ:b5e2iid)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

marksvc:
I changed to `sfProjectId`, `ptProjectId`, and `sfUserId` in the PR. For ParatextService.cs, I did a bunch of renaming of existing `projectSFId` and `projectPTId`.

Nateowami (approved):
:lgtm:

> Reviewed 1 of 3 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1582#-:-NHVcvOV2OPag85-aBcy:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1586, opened 2022-11-07
log: provide more information when failing to refresh PT token

marksvc:
Currently this can throw without revealing what user it threw for.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1586)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -2488,12 +2488,23 @@ namespace SIL.XForge.Scripture.Services
 
                 if (!userSecret.ParatextTokens.ValidateLifetime())
                 {
-                    Tokens refreshedUserTokens = await _jwtTokenHelper.RefreshAccessTokenAsync(
-                        _paratextOptions.Value,
-                        userSecret.ParatextTokens,
-                        _registryClient,
-                        token
-                    );
+                    Tokens refreshedUserTokens;
+                    try
+                    {
+                        refreshedUserTokens = await _jwtTokenHelper.RefreshAccessTokenAsync(
+                            _paratextOptions.Value,
+                            userSecret.ParatextTokens,
+                            _registryClient,
+                            token
+                        );
+                    }
+                    catch
+                    {
+                        _logger.LogWarning(
+                            $"ParatextService.GetParatextAccessLock for sfUserId {sfUserId} is throwing from call RefreshAccessTokenAsync()."
+                        );
+                        throw;
+                    }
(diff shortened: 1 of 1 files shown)


Nateowami (commented):
It seems like the real solution would be adding the userId to the exception, but that's probably not something we can easily do, unless we catch and throw a new exception that wraps the old one?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1586#-:-NGJcIMc1EEJ136wnhCc:b-zi20oi)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

marksvc:
(Oops, the CI shows a compile error.)

marksvc:
Haha, the CI error may be because this relies on an earlier commit in another PR. So we'll just need to wait for that one to be merged. Oops.

Nateowami:
@marksvc Which one?

marksvc (commented):
Hmm. We could catch the exception and throw a new one that wraps the original exception. But (not doing that and instead) throwing the original exception is nice because we preserve what kind of exception the original was (which presumably could be various things, even if often something like HttpException). 
An alternative is to add an item to the exception's `Data` Dictionary, giving more information. I haven't played with that much to know how easily that would be exposing the information. 
But I know that logging it on the way works.

> Which one?

PR 1582. I'm not in a rush.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1586#-:-NGOR-Bj140wP1seZcsW:b-3ffd10)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1586)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2504 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1586#-NHW1N6DAuoT64yLqu5Z:-NHW1N6E5yhkWh29oJM9:b18n47v) ([raw file](https://github.com/sillsdev/web-xforge/blob/d20ced594cf939047686cca7199d5277304742ac/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2504)):*
> ```Smalltalk
>                     {
>                         _logger.LogWarning(
>                             $"ParatextService.GetParatextAccessLock for sfUserId {sfUserId} is throwing from call RefreshAccessTokenAsync()."
> ```

I made a minor change here to get things to compile.

PR #1590, opened 2022-11-07
doc: update production build steps to use script

marksvc:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1590)

### DIFF
--- a/README.md
+++ b/README.md
@@ -254,14 +254,13 @@ Delete the PR branch after merge.
 You can do a production build of the backend and frontend locally by running the following in the repository root:
 
 ```bash
-dotnet publish src/SIL.XForge.Scripture/SIL.XForge.Scripture.csproj -c Release \
-  -r linux-x64 /p:Version=9.9.9 /p:AngularConfig=production --self-contained
+scripts/build-production
 ```
 
 Then you can attempt to run it with:
 
 ```bash
-cd src/SIL.XForge.Scripture/bin/Release/net6.0/linux-x64/publish
+cd artifacts/app
 ./SIL.XForge.Scripture
 ```
 


Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1590)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`README.md` line 257 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1590#-NGK6vqICSnLpzw5wY0t:-NGK6vqICSnLpzw5wY0u:bz0acd7) ([raw file](https://github.com/sillsdev/web-xforge/blob/2b1e40be9cc26e53a4ce64c639308b5eee753b35/README.md#L257)):*
> ```Markdown
> 
> ```bash
> scripts/build-production
> ```

I think this needs to be `./scripts/build-production`.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1590)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`README.md` line 257 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1590#-NGK6vqICSnLpzw5wY0t:-NGxHxtF2SL1TTuUoMBj:b-ypyt99) ([raw file](https://github.com/sillsdev/web-xforge/blob/2b1e40be9cc26e53a4ce64c639308b5eee753b35/README.md#L257)):*

I think this needs to be `./scripts/build-production`.
</blockquote></details>

Good eye! In this situation, we don't need to start with `./`.  Here's the reason:
If you were to 
```bash
$ cd scripts
```
and then try to run 
```bash
$ build-production
```
You would be told `command not found`. The reason is because `build-production` is not in a directory in the `PATH` environment variable. If you were to add `.` to `PATH`, then you could successfully run `build-production` in the way shown above. (However, it's not good security practice to do this, because you could find yourself in a directory where there is a malicious `ls` script, which you might accidentally end up running.)

When we run commands in bash, we specify them either as commands that are in one of the directories in `PATH`, or by path, where the specified path contains a `/`. So in the example above, we would instead write 
```bash
$ ./build-production
```
which causes us to request to run the script _by path_. Prepending `./` is a handy way to get a `/` character into our command. We could instead have written `../scripts/build-production` or `db_tools/../build-production` which would also work.

If you were in the directory _above_ `scripts` (eg 
```bash
cd ~/src/web-xforge
```
), then you could run the `build-production` script by running
```bash
scripts/build-production
```
because, again, it is specifying to run the command _by path_. Prepending `./` is still valid, but not needed.

Nateowami (approved):
Ah, right! Looks good then.

Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1590#-:-NGy5BdQBSfKCt20lI06:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1590)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1591, opened 2022-11-08
SF-1652 - Allow marking responses for export

nigel-wells:
- Admins can now set the status of an answer to exportable or resolved
- Added a new project setting for exporting answers
- Added UI for filtering questions

**UI for setting status**
image

**UI for filtering (admin)**
image

**UI for filtering (non admin)**
image

**UI for export settings**
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)


Nateowami (commented):
This is great. Before reviewing anything I checked it out locally. I have some suggested wording changes.

Obviously a lot of this is highly subjective, so push back if you have a different opinion.

---

"Exporting answers" instead of just "Exporting"
"Answers marked for export" feels simpler to me than "Answers set as exportable" and is more consistent with wording elsewhere.



---

Keeping with the theme of making sure users understand what the setting is actually doing "Marked for export" makes it more clear that this is an attribute the user is setting, rather than something about the answer itself (e.g. it can't be exported because it has audio, and audio doesn't export to Paratext). "Marked resolved" would go along with it.



---

"Mark for export to Paratext" is just shorter than "Mark answer to be exported into Paratext" and feels somewhat less wordy to me.



---

I prefer "Answered" and "Not answered" over "Not answered" and "Have answered" (order being significant as well).



---

Good work, and enough bike shedding from me. I've reached the end of my day though so reviewing this will have to be another time.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-:-NGKHgPY-ccu4_2GmzD8:bnmmuc9)* status: 0 of 53 files reviewed, all discussions resolved

nigel-wells (commented):
Happy to make those changes - it's no secret that I'm not the best wordsmith :)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-:-NGOA6RsAA4dKZkEVgwR:bso35yp)* status: 0 of 53 files reviewed, all discussions resolved

RaymondLuong3 (changes requested):
Nice work! I think this will be a useful feature, and now Paratext projects won't be overwhelmed by answers from community checking! One question.
If an answer is marked resolved, it does not get exported if I understand correctly. When we sync with Paratext, that answer is not synced. What happens if the answer was first exported, and then later that answer is resolved in SF. I imagine it just disappears from Paratext. Is that what you were intending?

> Reviewed 48 of 53 files at r1, 5 of 5 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-:-NGP5jIG8tSV2yzioIbW:bxj14m5)* status: all files reviewed, 18 unresolved discussions (waiting on @nigel-wells)

___
*[`src/RealtimeServer/scriptureforge/models/sf-project-rights.ts` line 11 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOYy3R3gA_ZHgAOcQg:-NGOYy3S9zUjOFH5L5_I:b-o9o8j9) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts#L11)):*
> ```XML
>   Answers = 'answers',
>   AnswerComments = 'answer_comments',
>   AnswerStatus = 'answer_status',
> ```

That's right, this needs to be its own domain in order for a user to edit it.



_Code quote:_
```XML
  AnswerStatus = 'answer_status',
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/index.html` line 14 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOZrWD2C7k56-gYuRS:-NGOZrWD2C7k56-gYuRT:b-eurkta) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/index.html#L14)):*
> ```HTML
>     </script>
> 
>     <link href="https://fonts.googleapis.com/icon?family=Material+Icons|Material+Icons+Outlined" rel="stylesheet" />
> ```

What does this do?



_Code quote:_
```HTML
  <link href="https://fonts.googleapis.com/icon?family=Material+Icons|Material+Icons+Outlined" rel="stylesheet" />
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 28 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOaMyE3FMfWYBgMSo4:-NGOaMyE3FMfWYBgMSo5:bx965ix) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L28)):*
> ```SCSS
>     #question-filter-menu {
>       width: calc(100% - 32px);
>       margin-left: 16px;
> ```

Do we have to account for RTL text margin too?



_Code quote:_
```SCSS
   margin-left: 16px;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts` line 581 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOcify2DU4EqVP6RQg:-NGOcify2DU4EqVP6RQh:b-q2yig0) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts#L581)):*
> ```XML
>     }));
> 
>     it('show no filters questions response', fakeAsync(() => {
> ```

What does this mean? Is it supposed to say `show no questions message for filter`?



_Code quote:_
```XML
  it('show no filters questions response', fakeAsync(() => {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts` line 1480 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOdJnZ0nZbziJWSfVG:-NGOdJn_BLNkMAZwBAnz:bkble8e) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts#L1480)):*
> ```XML
>         env.selectQuestion(6);
>         if (USER === ADMIN_USER) {
>           expect(env.getExportAnswerButton(0)).withContext(`${USER.role} can see export button`).not.toBeNull();
> ```

In some situations, there is a benefit to using `toBeTruthy()` if what you are testing for is that it is not null and not undefined. In this case, I think it is sufficient to test that it is not null, since we are confident that `not.toBeNull()` adequately demonstrates that the button is visible.



_Code quote:_
```XML
          expect(env.getExportAnswerButton(0)).withContext(`${USER.role} can see export button`).not.toBeNull();
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts` line 1493 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOeGLa6NGnEhJrnSIo:-NGOeGLa6NGnEhJrnSIp:btjo713) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts#L1493)):*
> ```XML
>       env.selectQuestion(6);
>       const buttonIndex = 0;
> ```

It is worth showing that the `status-exportable` class doesn't exist on the button to begin with.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts` line 1495 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOebqL8_tCzLNCv1TR:-NGOebqL8_tCzLNCv1TS:bnk5iyv) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts#L1495)):*
> ```XML
> 
>       env.clickButton(env.getExportAnswerButton(buttonIndex));
>       expect(env.getExportAnswerButton(buttonIndex).classes['status-exportable']).toBe(true);
> ```

If a user clicks on the button a second time, does it remove the `status-exportable` class? I see that exportable and resolved are mutually exclusive. If I click resolve, we can show that the status is now resolved and the `status-resolve` class is visible.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts` line 1504 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOeUjB5J3lK36_0-Y4:-NGOeUjB5J3lK36_0-Y5:b-po7jvb) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts#L1504)):*
> ```XML
>       env.selectQuestion(6);
>       const buttonIndex = 0;
> ```

It is worth showing that the `status-resolved` class isn't on the button to begin with.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts` line 1506 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOejmhE3BOzx0qE0EQ:-NGOejmhE3BOzx0qE0ER:b25nsr8) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts#L1506)):*
> ```XML
> 
>       env.clickButton(env.getResolveAnswerButton(buttonIndex));
>       expect(env.getResolveAnswerButton(buttonIndex).classes['status-resolved']).toBe(true);
> ```

If the user clicks on the button a second time, does it remove the `status-resolved` class?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html` line 166 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGP4pbSE_PqQ6MR402g:-NGP4pbSE_PqQ6MR402h:b-4lhhea) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html#L166)):*
> ```HTML
>                     matTooltipPosition="above"
>                   >
>                     {{ t(isAnswerExportable(answer) ? "marked_for_export" : "mark_for_export") }}
> ```

Should this be visible if the user has not selected "answers marked for export" on the settings page? It is better to hide it, so a user knows that if they select this, it will be exported, instead of it also depending on the setting on the settings page.



_Code quote:_
```HTML
                  >
                    {{ t(isAnswerExportable(answer) ? "marked_for_export" : "mark_for_export") }}
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html` line 190 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOxZg_C0EBc7RbroVh:-NGOxZga8pZYG7ZY8Am_:bxse4jg) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html#L190)):*
> ```HTML
>                     {{ t("edit") }}
>                   </button>
>                   <div *ngIf="canDeleteAnswer(answer)" class="delete-divider">|</div>
> ```

This devider would not be necessary if `canChangeAnswerStatus()` is false, I think



_Code quote:_
```HTML
<div *ngIf="canDeleteAnswer(answer)" class="delete-divider">|</div>
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts` line 409 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOyElXC8sufEkXlL4V:-NGOyElXC8sufEkXlL4W:b-3kgkvd) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L409)):*
> ```XML
>   }
> 
>   isAnswerExportable(answer: Answer): boolean {
> ```

What do you think about `isMarkedForExport` instead. isAnswerExportable sounds we are testing for a certain condition of whether it can or cannot be exported.



_Code quote:_
```XML
  isAnswerExportable(answer: Answer): boolean {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts` line 431 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOyay3-Vl74iTHLhrr:-NGOyay3-Vl74iTHLhrs:b8wn4e8) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L431)):*
> ```XML
>   }
> 
>   markAnswerAsExportable(answer: Answer) {
> ```

Then this would be `markAnswerForExport`



_Code quote:_
```XML
  markAnswerAsExportable(answer: Answer) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 31 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOyyf98zDg3fIi1oHQ:-NGOyyf98zDg3fIi1oHR:b746fk1) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L31)):*
> ```XML
> 
> export enum QuestionFilter {
>   All,
> ```

It seems strange that we are filtering for `All`. What do you think about `None` to represent no filter, i.e. `QuestionFilter.None`.



_Code quote:_
```XML
  All,
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 102 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOzxzDAe8S95ZDGIoY:-NGOzxzDAe8S95ZDGIoZ:b-ukv4f7) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L102)):*
> ```XML
>       const hasExportableAnswers: boolean =
>         q.data.answers.filter((a: Answer) => a.status === AnswerStatus.Exportable).length > 0;
>       if (this._filter === QuestionFilter.All) {
> ```

I wonder if a switch clause might make this a little cleaner?



_Code quote:_
```XML
      if (this._filter === QuestionFilter.All) {
```
___
*[`src/SIL.XForge.Scripture/Models/AnswerStatus.cs` line 12 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGP1BOdEZJBg5Cxlad2:-NGP1BOdEZJBg5Cxlad3:bo935fm) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/Models/AnswerStatus.cs#L12)):*
> ```Smalltalk
>         public const string Resolved = "resolve";
> 
>         /// <summary>Answer can be exported during a sync if allowed.</summary>
> ```

Do you mean answer is exported if configured instead of is allowed? Since it is the setting that is configured to export or not.



_Code quote:_
```Smalltalk
   /// <summary>Answer can be exported during a sync if allowed.</summary>
```
___
*[`src/SIL.XForge.Scripture/Models/CheckingConfig.cs` line 9 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGP4IFY-YYJQt3iwiUn:-NGP4IFY-YYJQt3iwiUo:b-q2htpe) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/Models/CheckingConfig.cs#L9)):*
> ```Smalltalk
>         public bool ShareEnabled { get; set; } = false;
>         public string ShareLevel { get; set; } = CheckingShareLevel.Specific;
>         public string AnswerExport { get; set; } = CheckingAnswerExport.MarkedForExport;
> ```

What about AnswerExportMethod as a more descriptive name? Or add a comment so we know what AnswerExport means.



_Code quote:_
```Smalltalk
 public string AnswerExport { get; set; } = CheckingAnswerExport.MarkedForExport;
```
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextNotesMapperTests.cs` line 778 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGP3u-1-9oUbeUhDXV8:-NGP3u-1-9oUbeUhDXV9:b-jruwz5) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/test/SIL.XForge.Scripture.Tests/Services/ParatextNotesMapperTests.cs#L778)):*
> ```Smalltalk
>                                         OwnerRef = "user04",
>                                         SyncUserRef = answerSyncUserId2,
>                                         DateCreated = new DateTime(2019, 1, 2, 8, 0, 0, DateTimeKind.Utc),
> ```

Change this date slightly. The date is important to be able to uniquely identify a comment. You can see how the other notes have slightly different dates.



_Code quote:_
```Smalltalk
    DateCreated = new DateTime(2019, 1, 2, 8, 0, 0, DateTimeKind.Utc),
```

Nateowami (commented):
> I imagine it just disappears from Paratext.

In my testing it didn't disappear. I expected that perhaps it would be marked as resolved.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-:-NGP6wIQBLYIuMm-1D6d:btwnusn)* status: all files reviewed, 18 unresolved discussions (waiting on @nigel-wells)

nigel-wells (commented):
If your export configuration is set to only export answers marked for export then any previously exported answers that are no longer marked for export will be soft deleted and removed from PT after the next sync and S/R. I've tested this just now and it works as expected.

Currently (I think) our sync for answers to PT is one way and any changes made in PT are ignored.

It is a good question about if a resolved answer in SF should still be exported but marked as resolved in PT - is that still clogging up their notes? What is the likely workflow admins are going to be doing here as that should guide is to the right decision.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-:-NGTQg_e1RmPJGSho925:bykimy2)* status: 21 of 53 files reviewed, 17 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/index.html` line 14 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOZrWD2C7k56-gYuRS:-NGPPY9i9N5OENMPNHWR:b1erap8) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/index.html#L14)):*

What does this do?
</blockquote></details>

This enables a new outline version of all the material icons so that I could have a toggled off/on state for the filter icon

**No filter applied**


** Filter applied**

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 28 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOaMyE3FMfWYBgMSo4:-NGPQ2Xz37LaSxh-IhlB:b2jsidk) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L28)):*

Do we have to account for RTL text margin too?
</blockquote></details>

Doh! Not quite a habit yet to check RTL on everything. Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts` line 581 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOcify2DU4EqVP6RQg:-NGPQeK9B-_h_D1oJNhE:b-9uikfb) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts#L581)):*

What does this mean? Is it supposed to say `show no questions message for filter`?
</blockquote></details>

Yes - have updated to your suggestion which reads much better.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts` line 1480 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOdJnZ0nZbziJWSfVG:-NGPQrLlElPvc9FyMQBe:b89gd5g) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts#L1480)):*

In some situations, there is a benefit to using `toBeTruthy()` if what you are testing for is that it is not null and not undefined. In this case, I think it is sufficient to test that it is not null, since we are confident that `not.toBeNull()` adequately demonstrates that the button is visible.
</blockquote></details>

Correct - good to remember, thanks.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts` line 1493 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOeGLa6NGnEhJrnSIo:-NGTGVn2DyxIn2B8Wq-H:bct2865) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts#L1493)):*

It is worth showing that the `status-exportable` class doesn't exist on the button to begin with.
</blockquote></details>

Wouldn't hurt - done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts` line 1495 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOebqL8_tCzLNCv1TR:-NGTJURS769epkRP49y6:b-itq0qz) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts#L1495)):*

If a user clicks on the button a second time, does it remove the `status-exportable` class? I see that exportable and resolved are mutually exclusive. If I click resolve, we can show that the status is now resolved and the `status-resolve` class is visible.
</blockquote></details>

I've added in another test for this.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts` line 1504 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOeUjB5J3lK36_0-Y4:-NGTGXX2AeuNz9-cGK4E:bct2865) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts#L1504)):*

It is worth showing that the `status-resolved` class isn't on the button to begin with.
</blockquote></details>

Wouldn't hurt - done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts` line 1506 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOejmhE3BOzx0qE0EQ:-NGTJYF64igvwAfB-Y2z:b-kr8zis) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts#L1506)):*

If the user clicks on the button a second time, does it remove the `status-resolved` class?
</blockquote></details>

Have included a new test.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html` line 166 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGP4pbSE_PqQ6MR402g:-NGPQzcV-L2FtkhZyAAi:b-z75z8j) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html#L166)):*

Should this be visible if the user has not selected "answers marked for export" on the settings page? It is better to hide it, so a user knows that if they select this, it will be exported, instead of it also depending on the setting on the settings page.
</blockquote></details>

I believe so as there was talk of a future need to allow different kinds of exporting rather than just PT i.e. CSV. @nateowami what do you think on this?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html` line 190 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOxZg_C0EBc7RbroVh:-NGTMslE0aBurLAsjUmC:b-19lggu) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html#L190)):*

This devider would not be necessary if `canChangeAnswerStatus()` is false, I think
</blockquote></details>

Good point. I've added in some CSS to make it `display: none` and only show if a button is before it. We could get a lot more technical about showing the `|` but seems like a heap of extra work that will require tests to be written when we can simply hide it - there is no value or security concerns for this visual element.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts` line 409 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOyElXC8sufEkXlL4V:-NGPRU-AAmkUo1B0dlgE:b-xu6f4d) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L409)):*

What do you think about `isMarkedForExport` instead. isAnswerExportable sounds we are testing for a certain condition of whether it can or cannot be exported.
</blockquote></details>

Happy with that and keeps it inline with other text suggestions that have been made as well.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts` line 431 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOyay3-Vl74iTHLhrr:-NGPRe2Y-HEXZiUihZZm:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L431)):*

Then this would be `markAnswerForExport`
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 31 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOyyf98zDg3fIi1oHQ:-NGPRrfw-eDj0Y2ib2lm:b-n2f1fn) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L31)):*

It seems strange that we are filtering for `All`. What do you think about `None` to represent no filter, i.e. `QuestionFilter.None`.
</blockquote></details>

Good question. I'd probably interpret `None` differently i.e. show none. What about `Show All`? I guess the enum could be `None` as that would make sense but the label could be `Show All` as that might read better to the user?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 102 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOzxzDAe8S95ZDGIoY:-NGTNMLDCTvyxv_CiIyN:byury3x) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L102)):*

I wonder if a switch clause might make this a little cleaner?
</blockquote></details>

Thanks for the reminder as I meant to come back to this. I've refactored.
___
*[`src/SIL.XForge.Scripture/Models/AnswerStatus.cs` line 12 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGP1BOdEZJBg5Cxlad2:-NGPSSTiE79FBe5w1H2D:b-hs5c47) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/Models/AnswerStatus.cs#L12)):*

Do you mean answer is exported if configured instead of is allowed? Since it is the setting that is configured to export or not.
</blockquote></details>

Good point - I've updated
___
*[`src/SIL.XForge.Scripture/Models/CheckingConfig.cs` line 9 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGP4IFY-YYJQt3iwiUn:-NGTEr2v9zCPD1yFjqq7:bmvozkl) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/Models/CheckingConfig.cs#L9)):*

What about AnswerExportMethod as a more descriptive name? Or add a comment so we know what AnswerExport means.
</blockquote></details>

I've updated to your suggestion as it is used in a number of place backend and frontend so this will save commenting it in multiple places
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextNotesMapperTests.cs` line 778 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGP3u-1-9oUbeUhDXV8:-NGTFqpe7AoF1RZbQoyT:b-chij2u) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/test/SIL.XForge.Scripture.Tests/Services/ParatextNotesMapperTests.cs#L778)):*

Change this date slightly. The date is important to be able to uniquely identify a comment. You can see how the other notes have slightly different dates.
</blockquote></details>

Ah ok. Done.

Nateowami (changes requested):
> It is a good question about if a resolved answer in SF should still be exported but marked as resolved in PT

I think I would avoid doing that, since the user has indicated that it shouldn't be exported. However, if an answer has been exported and then is marked resolved, I think it makes sense to resolve it in Paratext rather than deleting it, so that there is still some record of it (we're not trying to rewrite history).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-:-NGUpTkH9--Rdodf4lAt:bs9n3k1)* status: 21 of 53 files reviewed, 18 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/models/sf-project-rights.ts` line 95 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGTPFBH8B0P-nrizj9v:-NGTPFBH8B0P-nrizj9w:blwgulg) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts#L95)):*
> ```XML
>       { projectDomain: SFProjectDomain.AnswerComments, operation: Operation.Delete },
> 
>       { projectDomain: SFProjectDomain.AnswerStatus, operation: Operation.Edit },
> ```

There's a potential problem here. We have a setting called "Can add & edit questions" for each Paratext user on the project. It's intended that anyone with this enabled has full power to manage questions (basically 

(In this screenshot, the user is an admin, so it can't be turned off, since it's part of an admin's inerrant permissions, which is why it's shown disabled).



I see two possible solutions to this:
1. Remove the "AnswerStatus" domain, and make modifying the answer status part of the question domain (it's on the answer in the structure of the data, but conceptually it could be thought of as being a property of the question, and anyone with permission to edit questions would be able to edit it). I'm not sure if that would work.
2. Add this permission at the same time that we add the other permissions (see `CollaboratorsComponent.toggleQuestionPermission` in `collaborators.component.ts`).
___
*[`src/SIL.XForge.Scripture/ClientApp/src/index.html` line 14 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOZrWD2C7k56-gYuRS:-NGUpxR9-Q_KJgAZnIl9:b2fshbs) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/index.html#L14)):*

This enables a new outline version of all the material icons so that I could have a toggled off/on state for the filter icon

**No filter applied**


** Filter applied**


</blockquote></details>

How much more data is transferred over the network due to this? Just wondering since we're serving users that typically have slow/limited internet. I know our bundle isn't exactly that small right now, but my guess is that at some point in the future (when features are a little more complete) we may put a bigger focus on having a small bundle.

Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)* status: 21 of 53 files reviewed, 19 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/models/checking-config.ts` line 8 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGUralq0hz9Dl3u3FQI:-NGUralrEJb2LkDPoIo6:bjfvjlq) ([raw file](https://github.com/sillsdev/web-xforge/blob/760be84f551a2a724c8b885dcdad14b23da31043/src/RealtimeServer/scriptureforge/models/checking-config.ts#L8)):*
> ```XML
> export enum CheckingAnswerExport {
>   All = 'all',
>   MarkedForExport = 'export',
> ```

This seems to me like a slightly odd choice of value. Seeing in a project document `"answerExportMethod": "export"`, I think an uninitiated developer is likely to find that confusing (I would read it as "yes, do indeed export things"). I think the values "all" and "none" make sense. I would expect the third option to be something like "only_marked_answers" or "marked_for_export".

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)* status: 21 of 53 files reviewed, 19 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/index.html` line 14 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOZrWD2C7k56-gYuRS:-NGYuGzm6fGIBeM6ybEM:b-li8cbj) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/index.html#L14)):*

How much more data is transferred over the network due to this? Just wondering since we're serving users that typically have slow/limited internet. I know our bundle isn't exactly that small right now, but my guess is that at some point in the future (when features are a little more complete) we may put a bigger focus on having a small bundle.
</blockquote></details>

I just tested this and found our current icon file is 129KB, and this loads a second file that is 155KB. That's not awful, but it's not ideal either. I just checked the network tab on live, and our bundle is (unsurprisingly) the largest thing to go over the network, at 756KB. The second largest thing is the icon file at 129KB, followed 43.8KB styles, and everything else falling off in size very quickly.

Maybe the right answer is to leave this for now and at some point in the future try to serve up a very small icon font with just the icons we're actually using?

nigel-wells (commented):
We'd then need to keep track of what exported answers actually made its way through the sync process as a  real `marked_for_export` (rather than export everything which I assume you wouldn't apply this rule to) so that we can then determine if it then got resolved we can then export it again. Writing this out out makes me wonder if there is going to be too much hidden logic going on that the administrator won't be able to figure out why some resolved questions are in PT and not others. Would a PT admin find real value in having it resolved in PT or would they be content to look those up in SF? Given that it is currently a one way sync I'd be inclined to leave the full management and lookup in SF.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-:-NGn_ywmDC6GzhMm8g3Q:b-iqqhxw)* status: 19 of 53 files reviewed, 19 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/models/checking-config.ts` line 8 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGUralq0hz9Dl3u3FQI:-NGnTnoR5ymEBLBfC-Fl:b-sxk8ol) ([raw file](https://github.com/sillsdev/web-xforge/blob/760be84f551a2a724c8b885dcdad14b23da31043/src/RealtimeServer/scriptureforge/models/checking-config.ts#L8)):*

This seems to me like a slightly odd choice of value. Seeing in a project document `"answerExportMethod": "export"`, I think an uninitiated developer is likely to find that confusing (I would read it as "yes, do indeed export things"). I think the values "all" and "none" make sense. I would expect the third option to be something like "only_marked_answers" or "marked_for_export".
</blockquote></details>

Fair enough. I was trying to keep it short as it is only seen in the DB. I've updated to `marked_for_export`
___
*[`src/RealtimeServer/scriptureforge/models/sf-project-rights.ts` line 95 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGTPFBH8B0P-nrizj9v:-NGnUKtV0xFjsSUcaVcg:bgoy9g2) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts#L95)):*

There's a potential problem here. We have a setting called "Can add & edit questions" for each Paratext user on the project. It's intended that anyone with this enabled has full power to manage questions (basically 

(In this screenshot, the user is an admin, so it can't be turned off, since it's part of an admin's inerrant permissions, which is why it's shown disabled).



I see two possible solutions to this:
1. Remove the "AnswerStatus" domain, and make modifying the answer status part of the question domain (it's on the answer in the structure of the data, but conceptually it could be thought of as being a property of the question, and anyone with permission to edit questions would be able to edit it). I'm not sure if that would work.
2. Add this permission at the same time that we add the other permissions (see `CollaboratorsComponent.toggleQuestionPermission` in `collaborators.component.ts`).
</blockquote></details>

I'd see exporting and resolving answers from "can I add & edit questions". If the intention of that setting is beyond just managing questions then we should probably rename it to some kind of "Community Checking Admin" permission to cover a wide use or permissions which covers questions, answers, anything else. I'm not sure of the workflow here as to what is best. I can see both an admin wanting a translator to help with both questions and answers as a "admin" type role but I can also see wanting help with setting up questions but leave it to the main admin to manage the responses to those questions. 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/index.html` line 14 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOZrWD2C7k56-gYuRS:-NGnZWfF99Ud5lW4972T:biz74b0) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/index.html#L14)):*

I just tested this and found our current icon file is 129KB, and this loads a second file that is 155KB. That's not awful, but it's not ideal either. I just checked the network tab on live, and our bundle is (unsurprisingly) the largest thing to go over the network, at 756KB. The second largest thing is the icon file at 129KB, followed 43.8KB styles, and everything else falling off in size very quickly.

Maybe the right answer is to leave this for now and at some point in the future try to serve up a very small icon font with just the icons we're actually using?
</blockquote></details>

Yes, that is the ideal to only serve up icons that are actually used. It's a bit of a process to get setup but then is relatively trivial after that. We basically need to get all the material SVG icons and compile a font file of the ones we actually want to use. There is a repo available (not maintained by Google by the sounds) where we can access the SVG files. We could either work from that (to keep up-to-date with changes to icons) or extract and store the icons we want and store them in our repo.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)* status: 49 of 53 files reviewed, 6 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 31 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGOyyf98zDg3fIi1oHQ:-NGs4Z4Y0sWKaHekQLIJ:bdellsx) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L31)):*

I think I would prefer that option, have the enum value as `none` but the user sees Show all.
</blockquote></details>

Great - done.

Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)* status: 49 of 53 files reviewed, 8 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 101 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NHBxIJZ32lYjzTvH1iI:-NHBxIJZ32lYjzTvH1iJ:b-8yl1zt) ([raw file](https://github.com/sillsdev/web-xforge/blob/6e1f664918997e0be62357daef4b8849fa38776f/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L101)):*
> ```XML
>         q.data.answers.filter((a: Answer) => a.status === AnswerStatus.Resolved).length > 0;
>       const hasExportableAnswers: boolean =
>         q.data.answers.filter((a: Answer) => a.status === AnswerStatus.Exportable).length > 0;
> ```

It seems to me that this is less efficient than it could be, since we are running filters that aren't actually needed. I'll open a PR against this branch.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 109 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NHBlhfC61YGiTj-ATHs:-NHBlhfC61YGiTj-ATHt:b9z0yj6) ([raw file](https://github.com/sillsdev/web-xforge/blob/6e1f664918997e0be62357daef4b8849fa38776f/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L109)):*
> ```XML
>         .set(QuestionFilter.HasAnswers, hasAnswers)
>         .set(QuestionFilter.NoAnswers, !hasAnswers)
>         .set(QuestionFilter.StatusNone, hasAnswers && !hasExportableAnswers && !hasResolvedAnswers)
> ```

This logic looks wrong to me. It looks like if there are there were two answers, and one was marked for export (or marked resolved), and one was not marked at all this would not match it.

Nateowami (changes requested):
> I'd be interested to know if the change detection logic is still required in the questions component any more? 

It's likely, but I'll do some testing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-:-NHSW55kC3MYoNIITPxo:ba1ko20)* status: 43 of 65 files reviewed, 10 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 216 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NHSWE-r96AcnyXcyexa:-NHSWE-r96AcnyXcyexb:b-7rl62h) ([raw file](https://github.com/sillsdev/web-xforge/blob/5adb6e5ff16e6edc41eee1297f5b1bb780e5306d/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L216)):*
> ```SCSS
>   }
>   &.activated-nav-item mat-icon {
>     animation: spin 1s linear infinite;
> ```

I assume this was for testing?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 24 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NHCFWSW3vL8BR8vpBMG:-NHCFWSW3vL8BR8vpBMH:bga49og) ([raw file](https://github.com/sillsdev/web-xforge/blob/6e1f664918997e0be62357daef4b8849fa38776f/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L24)):*
> ```HTML
>                 </h2>
>                 <button mdcIconButton (click)="questionFilterMenu.open = !questionFilterMenu.open">
>                   <mdc-icon [ngClass]="{ 'material-icons-outlined': !isQuestionFilterApplied }">filter_alt</mdc-icon>
> ```

Is there a reason all of this was done with MDC instead of Angular Material? We're trying to move away from MDC sicne it's no longer maintained.

(Sorry, I wrote this comment a few days ago and thought I had posted it).
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 1054 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NHCCJGy1kgZ-WR4wzrg:-NHCCJGy1kgZ-WR4wzrh:b-qmdsfh) ([raw file](https://github.com/sillsdev/web-xforge/blob/6e1f664918997e0be62357daef4b8849fa38776f/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L1054)):*
> ```XML
>     this.questionFilters.clear();
>     this.questionFilters.set(QuestionFilter.None, translate('checking.question_filter_none'));
>     if (this.isProjectAdmin) {
> ```

With some exceptions, in general, we should be checking permissions, not roles. If someone has `answer_status.edit` permission, they should have the tools to filter by that status, since that's the main purpose behind filtering (the other uses are likely useful, but nowhere near as important).

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)* status: 42 of 65 files reviewed, 10 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 24 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NHCFWSW3vL8BR8vpBMG:-NHW9_yD7bv3POC52ca8:balnbs0) ([raw file](https://github.com/sillsdev/web-xforge/blob/6e1f664918997e0be62357daef4b8849fa38776f/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L24)):*

Is there a reason all of this was done with MDC instead of Angular Material? We're trying to move away from MDC sicne it's no longer maintained.

(Sorry, I wrote this comment a few days ago and thought I had posted it).
</blockquote></details>

Have converted to Material. The Material works slightly different from MDC as it figures out the best position to display it on its own - which is good I guess.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 1054 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NHCCJGy1kgZ-WR4wzrg:-NHWDOa17bC3W4R_T0X_:b-i8go8h) ([raw file](https://github.com/sillsdev/web-xforge/blob/6e1f664918997e0be62357daef4b8849fa38776f/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L1054)):*

With some exceptions, in general, we should be checking permissions, not roles. If someone has `answer_status.edit` permission, they should have the tools to filter by that status, since that's the main purpose behind filtering (the other uses are likely useful, but nowhere near as important).
</blockquote></details>

Have updated. Initially I thought only admins could access settings to actually export but you're right, if others have access to set the status then they should be able to filter on the status.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 216 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NHSWE-r96AcnyXcyexa:-NHW46PZ-yd41rEqKb1F:bommnly) ([raw file](https://github.com/sillsdev/web-xforge/blob/5adb6e5ff16e6edc41eee1297f5b1bb780e5306d/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L216)):*

I assume this was for testing?
</blockquote></details>

Ha! Yes I quickly added that in for the screenshot in slack and forgot about it. Removed.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)* status: 42 of 65 files reviewed, 10 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 24 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NHCFWSW3vL8BR8vpBMG:-NHWYH8s9e-L4dOAXJiH:b-j0erda) ([raw file](https://github.com/sillsdev/web-xforge/blob/6e1f664918997e0be62357daef4b8849fa38776f/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L24)):*
> it figures out the best position to display it on its own 

Not sure what you mean by this. If you add `xPosition="before"` to `mat-menu`, you can get this:

Nateowami (commented):
> Reviewed 5 of 53 files at r1, 1 of 5 files at r2, 5 of 32 files at r3, 1 of 3 files at r4, 12 of 21 files at r6, 2 of 6 files at r7.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)* status: 56 of 65 files reviewed, 9 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 24 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NHCFWSW3vL8BR8vpBMG:-NHW_X3D1ZqqKLg-fjuN:b-16xq03) ([raw file](https://github.com/sillsdev/web-xforge/blob/6e1f664918997e0be62357daef4b8849fa38776f/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L24)):*

The API docs made it sound much more precise
> Position of the menu in the X axis.
I interpreted that as needing to be very specific so ignored it. Do you have an opinion on its position or just let Material do its thing?
</blockquote></details>

I got this by looking at the API examples. And the type definitions say:
``` ts
type MenuPositionX = 'before' | 'after';
type MenuPositionY = 'above' | 'below';
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)* status: 56 of 65 files reviewed, 9 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 24 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NHCFWSW3vL8BR8vpBMG:-NHWaJAB4dmA5vUpoxp8:bxch1j6) ([raw file](https://github.com/sillsdev/web-xforge/blob/6e1f664918997e0be62357daef4b8849fa38776f/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L24)):*

I got this by looking at the API examples. And the type definitions say:
``` ts
type MenuPositionX = 'before' | 'after';
type MenuPositionY = 'above' | 'below';
```
</blockquote></details>

Also, the nice thing about `'before' | 'after'` is that it handles things correctly when switching from LTR to RTL.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)* status: 56 of 65 files reviewed, 9 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 24 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NHCFWSW3vL8BR8vpBMG:-NHWb33o7NLMuTxjtLiv:b-zds4gf) ([raw file](https://github.com/sillsdev/web-xforge/blob/6e1f664918997e0be62357daef4b8849fa38776f/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L24)):*

Also, the nice thing about `'before' | 'after'` is that it handles things correctly when switching from LTR to RTL.
</blockquote></details>

I think I would go with `'before'` (sorry; I missed the question part of your comment when I originally replied).

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)* status: 55 of 65 files reviewed, 9 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 24 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NHCFWSW3vL8BR8vpBMG:-NHWrjHtBDigSl6E1MV1:bx8js5f) ([raw file](https://github.com/sillsdev/web-xforge/blob/6e1f664918997e0be62357daef4b8849fa38776f/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L24)):*

I think I would go with `'before'` (sorry; I missed the question part of your comment when I originally replied).
</blockquote></details>

Good to know that it nicely handles RTL switching. Done.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)* status: 51 of 65 files reviewed, 10 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 62 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NHzZUl57LfL5RWOo89E:-NHzZUl57LfL5RWOo89F:b-ldyfz3) ([raw file](https://github.com/sillsdev/web-xforge/blob/12416353da02d44f55fb619a6210df6cbcbedac3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L62)):*
> ```HTML
>           <div class="panel">
>             <div class="panel-heading">
>               <div class="panel-nav">
> ```

I'll just mention here that the book icon was removed. It looks like that was deliberate, and cleaner, but I'll mention it in case that was an accident.



_Code quote:_
```HTML
       <div class="panel-nav">
```

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)* status: 51 of 65 files reviewed, 10 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 62 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NHzZUl57LfL5RWOo89E:-NI-0XPFEo1sCGsO2pwd:b2i4tcz) ([raw file](https://github.com/sillsdev/web-xforge/blob/12416353da02d44f55fb619a6210df6cbcbedac3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L62)):*

I'll just mention here that the book icon was removed. It looks like that was deliberate, and cleaner, but I'll mention it in case that was an accident.
</blockquote></details>

Yes,  I removed 2 icons as part of this as it cleaned up the interface and removed any confusion around decorative icons vs button icons which all looked the same.

nigel-wells (commented):
@nateowami I've made those updates discussed if you want to finish your review now.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-:-NI-lZRACS3PkgfvuvE1:b-cz500d)* status: 51 of 65 files reviewed, 10 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

Nateowami (commented):
> Reviewed 2 of 21 files at r6, 2 of 7 files at r9, 3 of 7 files at r10.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)* status: 58 of 65 files reviewed, 5 unresolved discussions (waiting on @RaymondLuong3)

src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts
```diff
@@ -52,6 +52,17 @@ interface Summary {
...
+  None,
+  CurrentUserHasAnswered,
+  CurrentUserHasNotAnswered,
+  HasAnswers,
+  NoAnswers,
+  StatusNone,
+  StatusExport,
+  StatusResolved
+}
+
```
Nateowami, line 65: Numeric enums behave weirdly in TypeScript, the most surprising thing being that any number is assignable to it (even `-Infinity` and  `NaN`). It's also a little harder to debug if you log a value or are looking at it in the debugger, since you'll just see a number without a clear meaning.

Another situation where it makes a difference is if we forgot to add `[QuestionFilter.None]: () => true` to `questionFilterFunctions` lower in this class. As a numeric enum, the compilation error will be "Property '0' is missing in type ... ", whereas with strings the message becomes "Property 'none' is missing in type ...".

Here's my suggested change (you should be able to apply it within GitHub):

``` suggestion
export enum QuestionFilter {
  None = 'none',
  CurrentUserHasAnswered = 'current_user_has_answered',
  CurrentUserHasNotAnswered = 'current_user_has_not_answered',
  HasAnswers = 'has_answers',
  NoAnswers = 'no_answers',
  StatusNone = 'status_none',
  StatusExport = 'status_export',
  StatusResolved = 'status_resolved'
}
```

src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts
```diff
@@ -695,14 +732,24 @@ export class CheckingComponent extends DataLoadingComponent implements OnInit, O
...
+    this.questionFilterSelected = filter;
+    this.updateVisibleQuestions();
+  }
+
   totalQuestions(): number {
     return this.questionDocs.length;
   }
 
+  totalVisibleQuestions(): number {
+    return (this.visibleQuestions ?? []).length;
```
Nateowami, line 750: How about:
``` suggestion
    return this.visibleQuestions?.length ?? 0;
```

Nateowami (commented):
I've added a couple comments here in GitHub because it allows me to add changes that can be committed directly

src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts
Nateowami: I just discovered this has the undesired side effect of making entries be in the wrong order in the UI, since the template uses `let f of questionFilters | keyvalue`, and [the documentation for KeyValuePipe](https://angular.io/api/common/KeyValuePipe) says:
> The output array will be ordered by keys. By default the comparator will be by Unicode point value. You can optionally pass a compareFn if your keys are complex types.

Nateowami:
@nigel-wells I've created a branch feature/SF-1652-i18n-keys with some changes.

Right now we're trying to map values to keys, which produces things like this:
``` ts
      this.questionFilters
        .set(QuestionFilter.CurrentUserHasAnswered, 'question_filter_answered')
        .set(QuestionFilter.CurrentUserHasNotAnswered, 'question_filter_not_answered');
```
You have to make sure all your filter mappings are correct, and that all the i18n keys actually exist. The change I've made is to just use a single set of strings to keep track of which filter is selected, and for i18n. This removes the need to map values to each other. It also means more type safety. Instead of a map, you just use an array. I still created a mapping using an object:
``` ts
export const QuestionFilter = {
  None: 'none',
  CurrentUserHasAnswered: 'answered',
  CurrentUserHasNotAnswered: 'not_answered',
  HasAnswers: 'has_answers',
  NoAnswers: 'no_answers',
  StatusNone: 'not_reviewed',
  StatusExport: 'exportable',
  StatusResolved: 'resolved'
} as const;
```
However, this mapping is just a convenient way of accessing the string values, and if one of the values didn't match what's in `checking_en.json`, there will be a type error.

src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts
Nateowami: I'm retracting this suggestion in favor of a more complete solution (see other comments).

Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)* status: 58 of 65 files reviewed, 8 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/models/answer.ts` line 6 at r10](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NI8pDD90J1LoaCcThLg:-NI8pDD90J1LoaCcThLh:b-h9f8wi) ([raw file](https://github.com/sillsdev/web-xforge/blob/0cae2fa5b324bf406274f02f8fd516700d15a542/src/RealtimeServer/scriptureforge/models/answer.ts#L6)):*
> ```XML
> 
> export enum AnswerStatus {
>   None = '',
> ```

What's the rationale behind the empty string? Is the idea that initially there will be no value? It looks like when we save an answer we set the status to this, but there are old answers that I don't think we'll be migrating.

nigel-wells (commented):
Thanks for going ahead and making these changes to support the new i18n type improvements - that's very handy. How did you want to handle this new branch - did you want to merge it into the existing one or keep working from that one? For now I've rebased again off `master` with some other changes.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-:-NI9SD2x5pjBTs2acR0s:b-n3j2bx)* status: 45 of 65 files reviewed, 8 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/models/answer.ts` line 6 at r10](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NI8pDD90J1LoaCcThLg:-NI9TW932vlE4d7fGzak:bf3ukz5) ([raw file](https://github.com/sillsdev/web-xforge/blob/0cae2fa5b324bf406274f02f8fd516700d15a542/src/RealtimeServer/scriptureforge/models/answer.ts#L6)):*

What's the rationale behind the empty string? Is the idea that initially there will be no value? It looks like when we save an answer we set the status to this, but there are old answers that I don't think we'll be migrating.
</blockquote></details>

I didn't see much value in saving "none" or anything to mongo when an empty strings says the same thing - there is no status.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 65 at r10](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGK6z2XQaFtlx4Iukbq-r10-65:-NI9SzXV6SqodVk6qF1j:b2nhll9) ([raw file](https://github.com/sillsdev/web-xforge/blob/0cae2fa5b324bf406274f02f8fd516700d15a542/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L65)):*

I'm retracting this suggestion in favor of a more complete solution (see other comments).
</blockquote></details>

Ok.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 750 at r10](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGK6z2XQaFtlx4Iukbq-r10-750:-NI9THdvC7YtVDNM_Qqx:b-rma598) ([raw file](https://github.com/sillsdev/web-xforge/blob/0cae2fa5b324bf406274f02f8fd516700d15a542/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L750)):*

How about:
``` suggestion
    return this.visibleQuestions?.length ?? 0;
```
</blockquote></details>

I'm sure I tried that but got a type error - perhaps my IDE was just slow in validating it. I've changed it now as that is the better approach.

Nateowami (commented):
> Reviewed 1 of 2 files at r8, 1 of 7 files at r9, 2 of 7 files at r10, 15 of 16 files at r11, 12 of 12 files at r12.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591)* status: 64 of 65 files reviewed, 4 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html` line 166 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-NGP4pbSE_PqQ6MR402g:-NIB2Qbj2j7eWsQM_RY7:b-5xfc6f) ([raw file](https://github.com/sillsdev/web-xforge/blob/df13caba00b662c972d8d141c2652591e562290c/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html#L166)):*

I believe so as there was talk of a future need to allow different kinds of exporting rather than just PT i.e. CSV. @nateowami what do you think on this?
</blockquote></details>

I think let's leave it for now. Theoretically a user may want to be able to set the status before deciding for certain to export the answers (e.g. in testing the software on a project before using it)

Nateowami (approved):
:lgtm:

> Dismissed @RaymondLuong3 from 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1591#-:-NIG8Wdb1CrSpF4qeB3j:bnfp4nl)* status: 64 of 65 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1594, opened 2022-11-08
dev: reduce nesting in app component ngOnInit

marksvc:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1594)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
@@ -292,142 +292,146 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
 
   async ngOnInit(): Promise<void> {
     this.loadingStarted();
-    if (await this.isLoggedIn) {
-      this.currentUserDoc = await this.userService.getCurrentUser();
-      const userData = cloneDeep(this.currentUserDoc.data);
-      if (userData != null) {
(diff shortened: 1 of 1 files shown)


marksvc:
Inverting the `if` to reduce nesting came up during PR 1577. There is some debate as to whether it even makes sense to have `await this.isLoggedIn` in an `if`.

Whether or not `isLoggedIn` is or is not being used in the most accurate way, this patch does at least reduce the nesting of `ngOnInit()`.

Nateowami (commented):
This is a good change, though we may want to hold off on it and let #1577 and possibly #1578 through first, since this is a really simple change and it seems likely to produce some conflicts (it would be very easy to fix up any conflicts if we merge this second, since it's so easy to just redo the change.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1594#-:-NGMhJl13BX7xKKpB3Mx:bbazb22)* status: 0 of 1 files reviewed, all discussions resolved

marksvc:
I agree that this should take a back seat and let other PRs in the same area go thru first. 
I just wanted to apply action to our discussion while it was fresh.

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1594)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1598, opened 2022-11-09
SF-1807 Filter note threads to SF notes only for reviewer users

RaymondLuong3:
* query note threads with an extra parameter for the prefix
* add guard to allow reading from the realtime server
* move SF note prefix to realtime server
* add tests for reviewer view of note threads in the editor

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1598)

### DIFF
--- a/src/RealtimeServer/common/services/json-doc-service.ts
+++ b/src/RealtimeServer/common/services/json-doc-service.ts
@@ -1,5 +1,6 @@
 import ShareDB from 'sharedb';
 import { ObjProxyArg } from 'ts-object-path';
+import { OwnedData } from '../models/owned-data';
 import { obj, ObjPathTemplate } from '../utils/obj-path';
 import { DocService } from './doc-service';
 
@@ -16,20 +17,20 @@ export abstract class JsonDocService<T> extends DocService<T> {
     return obj<T>().pathTemplate(field, inherit);
   }
 
-  protected checkImmutableProps(ops: ShareDB.Op[] | ShareDB.Op): boolean {
+  protected checkImmutableProps(ops: ShareDB.Op[] | ShareDB.Op, entity?: OwnedData): boolean {
     if (ops instanceof Array) {
       for (const op of ops) {
-        if (this.getMatchingPathTemplate(this.immutableProps, op.p) !== -1) {
+        if (this.getMatchingPathTemplate(this.immutableProps, op.p, entity) !== -1) {
           return false;
         }
       }
       return true;
     }
 
-    return this.getMatchingPathTemplate(this.immutableProps, ops.p) === -1;
+    return this.getMatchingPathTemplate(this.immutableProps, ops.p, entity) === -1;
   }
 
-  protected getMatchingPathTemplate(pathTemplates: ObjPathTemplate[], path: ShareDB.Path): number {
+  protected getMatchingPathTemplate(pathTemplates: ObjPathTemplate[], path: ShareDB.Path, _entity?: OwnedData): number {
     for (let i = 0; i < pathTemplates.length; i++) {
       if (pathTemplates[i].matches(path)) {
         return i;
--- a/src/RealtimeServer/common/services/project-data-service.ts
+++ b/src/RealtimeServer/common/services/project-data-service.ts
@@ -81,7 +81,7 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
       throw new Error('The doc service has not been initialized.');
     }
     const project = await this.server.getProject(doc.projectRef);
-    const domain = this.getUpdatedDomain([]);
+    const domain = this.getUpdatedDomain([], doc);
     return project != null && domain != null && this.hasRight(project, domain, Operation.Create, session.userId, doc);
   }
 
@@ -94,7 +94,7 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
       throw new Error('The doc service has not been initialized.');
     }
     const project = await this.server.getProject(doc.projectRef);
-    const domain = this.getUpdatedDomain([]);
+    const domain = this.getUpdatedDomain([], doc);
     return project != null && domain != null && this.hasRight(project, domain, Operation.Delete, session.userId, doc);
   }
 
@@ -111,7 +111,7 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
       return false;
     }
 
-    for (const domain of this.domains) {
+    for (const domain of this.getApplicableDomains(doc)) {
       if (!this.hasRight(project, domain, Operation.View, session.userId, doc)) {
         return false;
       }
@@ -122,7 +122,7 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
   protected async allowUpdate(
     _docId: string,
     oldDoc: T,
-    _newDoc: T,
+    newDoc: T,
     ops: ShareDB.Op[],
     session: ConnectSession
   ): Promise<boolean> {
@@ -139,7 +139,7 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
     }
 
     for (const op of ops) {
-      const domain = this.getUpdatedDomain(op.p);
+      const domain: ProjectDomainConfig | undefined = this.getUpdatedDomain(op.p, newDoc);
       if (domain == null) {
         return false;
       }
@@ -245,15 +245,25 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
     return Promise.resolve();
   }
 
-  private getUpdatedDomain(path: ShareDB.Path): ProjectDomainConfig | undefined {
-    const index = this.getMatchingPathTemplate(
-      this.domains.map(dc => dc.pathTemplate),
-      path
+  /**
+   * Gets the applicable domains based on the properties in the entity.
+   * @param _entity A noteThread or note.
+   * @returns
+   */
+  protected getApplicableDomains(_entity?: OwnedData): ProjectDomainConfig[] {
+    return this.domains;
+  }
+
+  private getUpdatedDomain(path: ShareDB.Path, entity: OwnedData): ProjectDomainConfig | undefined {
+    const domainConfigs: ProjectDomainConfig[] = this.getApplicableDomains(entity);
+    const index: number = this.getMatchingPathTemplate(
+      domainConfigs.map(dc => dc.pathTemplate),
+      path,
+      entity
     );
     if (index !== -1) {
-      return this.domains[index];
+      return domainConfigs[index];
     }
-
     return undefined;
   }
 
@@ -281,7 +291,7 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
   private async handleApply(context: ShareDB.middleware.SubmitContext): Promise<void> {
     const connectSession = context.agent.connectSession as ConnectSession;
     if (context.op.del != null) {
-      const domain = this.getUpdatedDomain([]);
+      const domain = this.getUpdatedDomain([], context.snapshot!.data);
       if (domain != null) {
         await this.onBeforeDelete(connectSession.userId, context.id, domain.projectDomain, context.snapshot!.data);
       }
@@ -291,18 +301,18 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
   private async handleAfterSubmit(context: ShareDB.middleware.SubmitContext): Promise<void> {
     const connectSession = context.agent.connectSession as ConnectSession;
     if (context.op.create != null) {
-      const domain = this.getUpdatedDomain([]);
+      const domain = this.getUpdatedDomain([], context.op.create.data);
       if (domain != null) {
         await this.onInsert(connectSession.userId, context.id, domain.projectDomain, context.op.create.data);
       }
     } else if (context.op.del != null) {
-      const domain = this.getUpdatedDomain([]);
+      const domain = this.getUpdatedDomain([], context.snapshot!.data);
       if (domain != null) {
         await this.onDelete(connectSession.userId, context.id, domain.projectDomain, context.snapshot!.data);
       }
     } else if (context.op.op != null) {
       for (const op of context.op.op) {
-        const domain = this.getUpdatedDomain(op.p);
+        const domain = this.getUpdatedDomain(op.p, context.snapshot!.data);
         if (domain == null) {
           return;
         }
--- a/src/RealtimeServer/scriptureforge/models/note-thread.ts
+++ b/src/RealtimeServer/scriptureforge/models/note-thread.ts
@@ -5,6 +5,7 @@ import { VerseRefData } from './verse-ref-data';
 
 export const NOTE_THREAD_COLLECTION = 'note_threads';
 export const NOTE_THREAD_INDEX_PATHS = PROJECT_DATA_INDEX_PATHS;
+export const SF_NOTE_THREAD_PREFIX = 'SFNOTETHREAD_';
 
 /**
  * Note status, mimicking PT CommentList.cs.
--- a/src/RealtimeServer/scriptureforge/models/note-thread.ts
+++ b/src/RealtimeServer/scriptureforge/models/note-thread.ts
@@ -5,7 +5,7 @@ import { VerseRefData } from './verse-ref-data';
 
 export const NOTE_THREAD_COLLECTION = 'note_threads';
 export const NOTE_THREAD_INDEX_PATHS = PROJECT_DATA_INDEX_PATHS;
-export const SF_NOTE_THREAD_PREFIX = 'SFNOTETHREAD_';
+export const SF_NOTE_THREAD_PREFIX = 'SF_NOTE_THREAD_';
 
 /**
  * Note status, mimicking PT CommentList.cs.
--- a/src/RealtimeServer/scriptureforge/models/note-thread.ts
+++ b/src/RealtimeServer/scriptureforge/models/note-thread.ts
@@ -5,7 +5,6 @@ import { VerseRefData } from './verse-ref-data';
 
 export const NOTE_THREAD_COLLECTION = 'note_threads';
 export const NOTE_THREAD_INDEX_PATHS = PROJECT_DATA_INDEX_PATHS;
-export const SF_NOTE_THREAD_PREFIX = 'SF_NOTE_THREAD_';
 
 /**
  * Note status, mimicking PT CommentList.cs.
@@ -64,5 +63,6 @@ export interface NoteThread extends ProjectData {
   position: TextAnchor;
   status: NoteStatus;
   tagIcon: string;
+  publishedToSF?: boolean;
   assignment?: string;
 }
--- a/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts
+++ b/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts
@@ -10,7 +10,8 @@ export enum SFProjectDomain {
   AnswerComments = 'answer_comments',
   AnswerStatus = 'answer_status',
   Likes = 'likes',
-  NoteThreads = 'note_threads',
+  PTNoteThreads = 'pt_note_threads',
+  SFNoteThreads = 'sf_note_threads',
   Notes = 'notes'
 }
 
@@ -34,14 +35,16 @@ export class SFProjectRights extends ProjectRights {
 
       { projectDomain: SFProjectDomain.Likes, operation: Operation.View },
 
-      { projectDomain: SFProjectDomain.NoteThreads, operation: Operation.View },
+      { projectDomain: SFProjectDomain.SFNoteThreads, operation: Operation.View },
 
       { projectDomain: SFProjectDomain.Notes, operation: Operation.View }
     ];
     this.addRights(SFProjectRole.Observer, observerRights);
 
     const ptObserverRights: ProjectRight[] = observerRights.concat([
-      { projectDomain: SFProjectDomain.Project, operation: Operation.View }
+      { projectDomain: SFProjectDomain.Project, operation: Operation.View },
+
+      { projectDomain: SFProjectDomain.PTNoteThreads, operation: Operation.View }
     ]);
     this.addRights(SFProjectRole.ParatextObserver, ptObserverRights);
 
@@ -57,9 +60,8 @@ export class SFProjectRights extends ProjectRights {
       { projectDomain: SFProjectDomain.Likes, operation: Operation.Create },
       { projectDomain: SFProjectDomain.Likes, operation: Operation.DeleteOwn },
 
-      { projectDomain: SFProjectDomain.NoteThreads, operation: Operation.Create },
-      { projectDomain: SFProjectDomain.NoteThreads, operation: Operation.Edit },
-      { projectDomain: SFProjectDomain.NoteThreads, operation: Operation.Delete },
+      { projectDomain: SFProjectDomain.SFNoteThreads, operation: Operation.Create },
+      { projectDomain: SFProjectDomain.SFNoteThreads, operation: Operation.DeleteOwn },
 
       { projectDomain: SFProjectDomain.Notes, operation: Operation.Create },
       { projectDomain: SFProjectDomain.Notes, operation: Operation.EditOwn },
@@ -69,7 +71,10 @@ export class SFProjectRights extends ProjectRights {
     this.addRights(SFProjectRole.CommunityChecker, reviewerRights);
 
     const ptReviewerRights: ProjectRight[] = reviewerRights.concat([
-      { projectDomain: SFProjectDomain.Project, operation: Operation.View }
+      { projectDomain: SFProjectDomain.Project, operation: Operation.View },
+
+      { projectDomain: SFProjectDomain.PTNoteThreads, operation: Operation.Create },
+      { projectDomain: SFProjectDomain.PTNoteThreads, operation: Operation.DeleteOwn }
     ]);
     this.addRights(SFProjectRole.ParatextConsultant, ptReviewerRights);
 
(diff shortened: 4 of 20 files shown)


nigel-wells (changes requested):
I love that the backend is flexible enough to permissions updates flexible. I'm hoping the same can be applied to queries.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-:-NGnq_dQ5DxnJ04AciL8:b-olhueh)* status: all files reviewed, 6 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/models/note-thread.ts` line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnj92o-c8W7qYEjGNm:-NGnj92o-c8W7qYEjGNn:b-h3xiiw) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/RealtimeServer/scriptureforge/models/note-thread.ts#L8)):*
> ```XML
> export const NOTE_THREAD_COLLECTION = 'note_threads';
> export const NOTE_THREAD_INDEX_PATHS = PROJECT_DATA_INDEX_PATHS;
> export const SF_NOTE_THREAD_PREFIX = 'SFNOTETHREAD_';
> ```

Could we add some more underscores in as it is quite hard to understand what it says i.e. `SF_NOTE_THREAD_`
___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.ts` line 74 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnl8p8-MYwPddNvUiK:-NGnl8p8-MYwPddNvUiL:byyw3iq) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/RealtimeServer/scriptureforge/services/note-thread-service.ts#L74)):*
> ```XML
>       if (this.server == null) {
>         throw new Error('The doc service has not been initialized.');
>       }
> ```

You won't need these checks as they have already done in `super.allowRead`



_Code quote:_
```XML
      if (session.isServer || Object.keys(doc).length === 0) return true;

      if (this.server == null) {
        throw new Error('The doc service has not been initialized.');
      }
```
___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.ts` line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnlZyM09k6HX_l1Eht:-NGnlZyM09k6HX_l1Ehu:b-jdd2u9) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/RealtimeServer/scriptureforge/services/note-thread-service.ts#L78)):*
> ```XML
>       const project: Project | undefined = await this.server.getProject(doc.projectRef);
>       if (project == null) return false;
>       const isReviewer: boolean = project.userRoles[session.userId] === SFProjectRole.Reviewer;
> ```

What happens if a user isn't in `userRoles`? Or would that have been checked somewhere else already?
___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts` line 125 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnmrzg8Y0kVhRWeC98:-NGnmrzh-eIxdVwnNv6s:bq1vjql) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts#L125)):*
> ```XML
>     const noteThreadDocId: string = getNoteThreadDocId('project01', 'noteThread01');
>     await expect(async () =>
>       fetchDoc(conn, NOTE_THREAD_COLLECTION, getNoteThreadDocId('project01', 'noteThread01'))
> ```

Can use `noteThreadDocId` defined above.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts` line 165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnofgp01Hhcg1bHYEL:-NGnofgp01Hhcg1bHYEM:b-vzaga3) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts#L165)):*
> ```XML
>     if (sfNotesOnly) {
>       queryParams[obj<NoteThread>().pathStr(t => t.dataId)] = { $regex: SF_NOTE_THREAD_PREFIX };
>     }
> ```

I was surprised to see that this needed to happen. I had assumed the `allowRead` in the Realtime Server would have meant we didn't need to make any specific requests on the frontend. Or, was the change to above only relating to requesting a specific document whereas this is a query of documents? If it's because of a query then can we do something similar on the backend to exclude documents from a query. It would be nice if the backend took care of it all otherwise we'll need to remember to pass extra parameters in different parts of the codebase as everything grows.



_Code quote:_
```XML
    if (sfNotesOnly) {
      queryParams[obj<NoteThread>().pathStr(t => t.dataId)] = { $regex: SF_NOTE_THREAD_PREFIX };
    }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1162 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnpVV-9odnNVjnr17g:-NGnpVV-9odnNVjnr17h:b-q9zg63) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1162)):*
> ```XML
>     this.noteThreadQuery?.dispose();
>     const sfNotesOnly: boolean =
>       this.projectDoc?.data?.userRoles[this.userService.currentUserId] === SFProjectRole.Reviewer;
> ```

Same as my other comments - hoping we don't need to do this.



_Code quote:_
```XML
    const sfNotesOnly: boolean =
      this.projectDoc?.data?.userRoles[this.userService.currentUserId] === SFProjectRole.Reviewer;
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1598)* status: all files reviewed, 6 unresolved discussions (waiting on @nigel-wells)

___
*[`src/RealtimeServer/scriptureforge/models/note-thread.ts` line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnj92o-c8W7qYEjGNm:-NGs07Qc83ie14MQHzz3:blm5pm) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/RealtimeServer/scriptureforge/models/note-thread.ts#L8)):*

Could we add some more underscores in as it is quite hard to understand what it says i.e. `SF_NOTE_THREAD_`
</blockquote></details>

Good thinking, Done.
___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.ts` line 74 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnl8p8-MYwPddNvUiK:-NGs0Tqx3IiVvM0TTwTk:b-9mcpxa) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/RealtimeServer/scriptureforge/services/note-thread-service.ts#L74)):*

You won't need these checks as they have already done in `super.allowRead`
</blockquote></details>

`super.allowRead` will tell us if we have the standard read permission, but it does not say if the session is server or not, so we still need to check if `session.isServer` and immediately return true. It is also possible that the base method will get updated and does not check that `this.server != null` so I think we should leave that check in also.
___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.ts` line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnlZyM09k6HX_l1Eht:-NGs1LrbACg9vrlLIE-u:b1zkgek) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/RealtimeServer/scriptureforge/services/note-thread-service.ts#L78)):*

What happens if a user isn't in `userRoles`? Or would that have been checked somewhere else already?
</blockquote></details>

I've updated this so that if the userRole is null, then we return false.
___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts` line 125 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnmrzg8Y0kVhRWeC98:-NGs1qxy-5UC0yGh2zoz:bi4x02x) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts#L125)):*

Can use `noteThreadDocId` defined above.
</blockquote></details>

Nice. I'm sure I meant to do that.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts` line 165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnofgp01Hhcg1bHYEL:-NGs1uCX6GA4chnzvRsd:b-tb9qo2) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts#L165)):*

I was surprised to see that this needed to happen. I had assumed the `allowRead` in the Realtime Server would have meant we didn't need to make any specific requests on the frontend. Or, was the change to above only relating to requesting a specific document whereas this is a query of documents? If it's because of a query then can we do something similar on the backend to exclude documents from a query. It would be nice if the backend took care of it all otherwise we'll need to remember to pass extra parameters in different parts of the codebase as everything grows.
</blockquote></details>

My understanding of the realtime server may be limited, but this is my current understanding. When we make a query, we tell the server what note thread docs we want. The realtime server will check that a user has the correct permission to read the docs that were queried for (hence the `allowRead` changes. The realtime server does not filter note threads that a user does not have permission to read (but it would be nice if it could handle that too. I'm not familiar with whether it can or cannot). If my understand is right, we need to specify both on the front end and the back end which note threads are accessible to the user.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1162 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnpVV-9odnNVjnr17g:-NGs3MH58hcsKQdL-5XY:b2nreh6) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1162)):*

Same as my other comments - hoping we don't need to do this.
</blockquote></details>

Same as above.

nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1598)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.ts` line 74 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnl8p8-MYwPddNvUiK:-NGsKztO4ZIwWMZx1C2i:bnsf7q) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/RealtimeServer/scriptureforge/services/note-thread-service.ts#L74)):*

`super.allowRead` will tell us if we have the standard read permission, but it does not say if the session is server or not, so we still need to check if `session.isServer` and immediately return true. It is also possible that the base method will get updated and does not check that `this.server != null` so I think we should leave that check in also.
</blockquote></details>

Good point that if the base method changed then we'd still want to ensure it is done here. Thanks.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts` line 165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnofgp01Hhcg1bHYEL:-NGseTM9E1NWiA5nwJ-D:b-t0goda) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts#L165)):*

My understanding of the realtime server may be limited, but this is my current understanding. When we make a query, we tell the server what note thread docs we want. The realtime server will check that a user has the correct permission to read the docs that were queried for (hence the `allowRead` changes. The realtime server does not filter note threads that a user does not have permission to read (but it would be nice if it could handle that too. I'm not familiar with whether it can or cannot). If my understand is right, we need to specify both on the front end and the back end which note threads are accessible to the user.
</blockquote></details>

That makes sense. Did you receive permission denied errors without the above filter in use? I wonder if it would be possible to adjust our `subscribe` method on the `SharedbRealtimeQueryAdapter` to strip out denied docs from the results before returning them? It feels like we should be able to do this - but I also only have limited understanding with the setup.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1162 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnpVV-9odnNVjnr17g:-NGsf0Uz6RUduEEWDM7f:b-qndak3) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1162)):*

Same as above.
</blockquote></details>

See how you get on with my other comment.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1598)* status: all files reviewed, 2 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts` line 165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnofgp01Hhcg1bHYEL:-NH0_4HwD-gIvLXz017H:b86i7x3) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts#L165)):*

That makes sense. Did you receive permission denied errors without the above filter in use? I wonder if it would be possible to adjust our `subscribe` method on the `SharedbRealtimeQueryAdapter` to strip out denied docs from the results before returning them? It feels like we should be able to do this - but I also only have limited understanding with the setup.
</blockquote></details>

Hmm, I just looked into it, and the docs do not mention what happens if a database query results in errors to read the documents. This might be because we are using a middleware called ShareDBAccess that is outside of ShareDB to control if a user has read/write access to a doc.
I think that the filtering parameters on the front end was actually the main point of this change because we do not want reviewers to gain access to translation notes they should not see, and the additional guards on the backend ensures that access to certain translation notes are denied if somehow they tried to read it.
I agree it would be nice if we could just make a query and get what we have access to, but I do not see an easy way to go about doing that.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1162 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnpVV-9odnNVjnr17g:-NH0acN4CuOyKNoyla1q:bkoq5iw) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1162)):*

See how you get on with my other comment.
</blockquote></details>

Commented above.

nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1598)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts` line 165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnofgp01Hhcg1bHYEL:-NH1EygJ7USNZb4h8Q15:b-d0s5ha) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts#L165)):*

Hmm, I just looked into it, and the docs do not mention what happens if a database query results in errors to read the documents. This might be because we are using a middleware called ShareDBAccess that is outside of ShareDB to control if a user has read/write access to a doc.
I think that the filtering parameters on the front end was actually the main point of this change because we do not want reviewers to gain access to translation notes they should not see, and the additional guards on the backend ensures that access to certain translation notes are denied if somehow they tried to read it.
I agree it would be nice if we could just make a query and get what we have access to, but I do not see an easy way to go about doing that.
</blockquote></details>

So does this mean if we request a doc directly we'll get permission denied but if we request the doc using a query then we will have permission? Sounds like a security risk but I understand what you mean that our middleware is only working on fetching documents.

What are the requirements for this task - is it to just filter them out and security isn't a concern or do we need to figure out a way to ensure they can never be accessed based on permission? Looping in @nateowami

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1598)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts` line 165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnofgp01Hhcg1bHYEL:-NH1LQZI-5ZWnhoEnWYi:b2h97va) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts#L165)):*

So does this mean if we request a doc directly we'll get permission denied but if we request the doc using a query then we will have permission? Sounds like a security risk but I understand what you mean that our middleware is only working on fetching documents.

What are the requirements for this task - is it to just filter them out and security isn't a concern or do we need to figure out a way to ensure they can never be accessed based on permission? Looping in @nateowami 
</blockquote></details>

The user will see permission denied in both cases, whether they request a doc directly, or through a query that they are not authorized to see all the resulting docs for. There should not be a security risk in that sense.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1598)* status: 4 of 10 files reviewed, 2 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts` line 165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnofgp01Hhcg1bHYEL:-NI-AWwd8bwqh5OTgv3j:bhbx25l) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts#L165)):*

I've spent some time testing this out. Initially I wasn't getting any permission denied issues but that was because the notes were already in IndexDB. After I cleared that out I was able to receive permission denied errors if I removed the additional `sfNotesOnly` parameter - which is why you have it there. My observation was the doc was the query would send the ID of the document back but I didn't notice any data of the document being supplied unless already stored locally.

I looked into ShareDB and it does support middleware as we know. I had a play around and was able to adjust the parameters in the Realtime server to inject the `$regex: 'SF_NOTE_THREAD_` to the query. This means we could look to enforce improved security for queries to stop documents from ever being sent to the frontend.

This is very rough but it should give you an idea of how you could go about doing a full solution:
```
backend.use('query', (context: QueryContext, next: (err?: any) => void) => {
  if (context.collection === 'note_threads') {
    context.query = { ...context.query, ...{ dataId: { $regex: 'SF_NOTE_THREAD_' } } };
  }
  next();
});
```
I added that to `web-socket-stream-listener.ts` as part of the `wss.on('connection')` listener. It could/should be somewhere else but it proved the concept was fine.

Using `context.agent.connectSession` you have access to the connected user to help figure out if they have permission or not.

Also, FYI, there appears to be some other useful middleware options that would allow us inspect submissions right before they are submitted to mongo. This would hopefully allow us to test and reject operations that would corrupt data.

We have a debugging option for the Realtime server as well which makes it a lot easier to see what is going on.


</blockquote></details>

Hey, that seemed to do the trick! It is nice to be able to query notes, and the backend handles what notes we have access to and returns them. Thanks for the tip.

nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1598)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/scripture-utils/utils.ts` line 4 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NI0NZho3ZJVxMRmxIVZ:-NI0NZho3ZJVxMRmxIV_:b-cokr1p) ([raw file](https://github.com/sillsdev/web-xforge/blob/9ce9430f7737540a79688b2cc7f7dc7273064522/src/RealtimeServer/scriptureforge/scripture-utils/utils.ts#L4)):*
> ```XML
> 
> export function canViewParatextNotes(role: string): boolean {
>   return role.startsWith('pt_') || role === SFProjectRole.Observer;
> ```

This was supposed to be the Reviewer wasn't it?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts` line 165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NGnofgp01Hhcg1bHYEL:-NI0Mmfj8KVrqt07KhhQ:bwboqnh) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ce39d2c1a85a52fb48e670b11738d0b916fd076/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts#L165)):*

Hey, that seemed to do the trick! It is nice to be able to query notes, and the backend handles what notes we have access to and returns them. Thanks for the tip.
</blockquote></details>

Awesome, glad it worked and you found a nice way to implement it.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1598)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/RealtimeServer/scriptureforge/scripture-utils/utils.ts` line 4 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NI0NZho3ZJVxMRmxIVZ:-NI8t1dZ7Af9EoWJ-Nuv:b-1ssnaz) ([raw file](https://github.com/sillsdev/web-xforge/blob/9ce9430f7737540a79688b2cc7f7dc7273064522/src/RealtimeServer/scriptureforge/scripture-utils/utils.ts#L4)):*

This was supposed to be the Reviewer wasn't it?
</blockquote></details>

Instead of `role.startsWith('pt_')`, how about `hasParatextRole(role)` from `src/RealtimeServer/scriptureforge/models/sf-project-role.ts`? (That function should probably be called `isParatextRole()`, but that's another topic).

Actually I would think this should use the rights service, though I haven't looked at the broader context.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1598)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/RealtimeServer/scriptureforge/scripture-utils/utils.ts` line 4 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NI0NZho3ZJVxMRmxIVZ:-NIYHRfu7cLSx4ZRLUF_:bxu47h6) ([raw file](https://github.com/sillsdev/web-xforge/blob/9ce9430f7737540a79688b2cc7f7dc7273064522/src/RealtimeServer/scriptureforge/scripture-utils/utils.ts#L4)):*

Instead of `role.startsWith('pt_')`, how about `hasParatextRole(role)` from `src/RealtimeServer/scriptureforge/models/sf-project-role.ts`? (That function should probably be called `isParatextRole()`, but that's another topic).

Actually I would think this should use the rights service, though I haven't looked at the broader context.
</blockquote></details>

I added a comment because this can be confusing. This checks whether a user should see note threads created in Paratext (i.e. paratext users, or observers). Reviewers should only see notes created in SF. @nateowami I did try to use the rights service, but this is a right within the right to view note threads, so the rights service doesn't cover that case, and I'm not sure that it should. To do that I would need to split out the Note thread domain, between Paratext notes and SF notes, that might be overkill.

nigel-wells (approved):
:lgtm:

> Reviewed 9 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-:-NIdXLjK8Huhsscx62-t:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/scripture-utils/utils.ts` line 4 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1598#-NI0NZho3ZJVxMRmxIVZ:-NIdWc0N4vppqrq2cxuE:bouvetj) ([raw file](https://github.com/sillsdev/web-xforge/blob/9ce9430f7737540a79688b2cc7f7dc7273064522/src/RealtimeServer/scriptureforge/scripture-utils/utils.ts#L4)):*

I added a comment because this can be confusing. This checks whether a user should see note threads created in Paratext (i.e. paratext users, or observers). Reviewers should only see notes created in SF. @nateowami I did try to use the rights service, but this is a right within the right to view note threads, so the rights service doesn't cover that case, and I'm not sure that it should. To do that I would need to split out the Note thread domain, between Paratext notes and SF notes, that might be overkill.
</blockquote></details>

Thanks for clarifying. Out of interest, why can an Observer - whose rights are less than a Reviewer - have permission to view PT notes when a Reviewer can't? I would have thought that neither could? I'm sure there is a reason so just wanting to understand.

Nateowami:
@RaymondLuong3 I'm going through PRs to see the status of each one. It looks like Nigel made some comments and you've updated this PR without commenting. Maybe that was intentional? Just want to make sure Nigel and you aren't both thinking this one is ready for the other.

RaymondLuong3:
Actually, this one is waiting for PR #1635 which uses the sf project rights to distinguish if a user can view and edit a note thread. That one is ready for review.

PR #1605, opened 2022-11-18
Improve efficiency and fix bug in question filtering

Nateowami:
Note that this is a PR against a branch with an open PR.

@nigel-wells I made the following changes:

- Replaced `EventEmitter` with `Subject<number>`, which I believe is the more correct option here. I see a couple other examples of using `EventEmitter` in this class. I think it works fine but isn't the most ideal choice. I didn't change the other instances.
- I fixed what I think was a bug in filtering for questions that have not yet been handled. It looked to me like any question that had an answer that had been handled (marked resolved or marked for export) would not match the filter. I think the number one use case for this feature is "filter for questions that have an answer that hasn't been handled yet", which would make this bug fall squarely within the most likely use case. I'm starting to wonder if you may have written what you intended to write and I just didn't communicate the intended behavior correctly. We're really filtering for _answers_, not questions.
- I optimized the filter so we only filter the answers one time.
- I optimized the filter so we don't run any filters if no filter is selected. Since most of the time I expect most users will not have a filter active, we can avoid iterating over the question list, and more importantly, avoid allocating a new array. Filtering a couple thousand questions should be instant, and allocating a new array of `QuestionDoc`s isn't going to clone the questions, just create an array of references (probably 64-bit pointers?), so it shouldn't be that bad. But it's possible for this to fire thousands of times in a single second, so they could potentially add up to something significant. And the optimization is trivial. And we've had major performance issues in the component before.

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts
@@ -63,8 +63,8 @@ export class CheckingQuestionsComponent extends SubscriptionDisposable {
   activeQuestionDoc?: QuestionDoc;
   activeQuestionDoc$ = new Subject<QuestionDoc>();
   @ViewChild(MdcList, { static: true }) mdcList?: MdcList;
-  @Output() totalVisibleQuestions = new EventEmitter<number>();
-  visibleQuestions: QuestionDoc[] = [];
+  @Output() totalVisibleQuestions = new Subject<number>();
+  visibleQuestions: Readonly<QuestionDoc[]> = [];
 
   private _filter: QuestionFilter = QuestionFilter.None;
   private project?: SFProjectProfile;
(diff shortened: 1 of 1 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1605)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 178 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1605#-NHC3TS137Srs2ykiWe1:-NHC3TS137Srs2ykiWe2:buhh5j5) ([raw file](https://github.com/sillsdev/web-xforge/blob/6e1f664918997e0be62357daef4b8849fa38776f/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L178)):*
> ```XML
>       }
>       this.questionDocs = docs;
>       this.filter = this._filter;
> ```

I changed this because I thought it wasn't the greatest style-wise. This line is assuming the setter has side effects and isn't idempotent, which happens to be the case, but doesn't seem like something that should be assumed by code outside a setter.

When reading this line you have to either look up the setter and see what it does, or assume it does more than than the name implies.

Nateowami:
I just added another commit to make the localization dynamic, rather than pre-computed. This has a number of advantages, including making it much easier to test translations to make sure they are working properly and don't break the layout. There may also be automated tests in the future that switch locales while the tests are running. And it makes for a nicer user experience.

Nateowami:
@nigel-wells Just want to make sure this PR is on your radar.

nigel-wells:
Yep - will on to it tomorrow

nigel-wells:
Thanks for your suggestions in this. As discussed during our meeting I had already taken a similar approach as well as a much larger refactor. I've pulled across from this PR the bits around filtering and translation. Feel free to close this PR and we'll continue on the other one.

PR #1607, opened 2022-11-21
SF-1764 C# API for Machine API

pmachapman:
Adds support for the new Machine API.

Key changes include:
 - Using IdentityModel to manage the access token and refresh for the new Machine API (https://identitymodel.readthedocs.io/en/latest/aspnetcore/worker.html)
 - Using HttpClientFactory to create HttpClient instances, so that the HTTP requests can be correctly unit tested (https://learn.microsoft.com/en-us/aspnet/core/fundamentals/http-requests)
 - Using Polly for HTTP client exponential backoff retry and circuit breaker policies for resilience (https://learn.microsoft.com/en-us/dotnet/architecture/microservices/implement-resilient-applications/implement-http-call-retries-exponential-backoff-polly)
 - Feature flag support to enabled/disable Machine API (https://learn.microsoft.com/en-us/azure/azure-app-configuration/use-feature-flags-dotnet-core)
 - Fixing a bug with `SFProjectService` and `ParatextService` where the `WritingSystem.Tag` was always null
 - Adding a C# API to access the new Machine API
 - `ParatextService` now uses `IMachineProjectService` instead of `IEngineService`, so that the in-memory Machine is used, and if the Feature flag is enabled, the Machine API is used.
 - A sync is only triggered for the new Machine API if the text corpus has changed
 - Objects that will need to be changed or removed when the Machine library is updated to 2.6 or higher are marked with TODO: flags. Note that updating Machine above 2.5.X will remove the in-memory Machine implementation.

To enable the Machine API on your local instance, configure the following user secrets (or environment variables):
 - Machine.ClientId = "(get this value from the document)"
 - Machine.ClientSecret = "(get this value from the document)"
 - Machine.ApiServer = "https://localhost:8444" (if using a local instance)
 - FeatureManagement:MachineApi = true

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)


Nateowami (commented):
> Reviewed 10 of 35 files at r1, 3 of 12 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 13 of 37 files reviewed, 3 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/appsettings.json` line 57 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NHXHojx6M1VF47CkkVg:-NHXHojx6M1VF47CkkVh:b4t1ikt) ([raw file](https://github.com/sillsdev/web-xforge/blob/405d3ba639f9e0dad5470d226d301de83536e97d/src/SIL.XForge.Scripture/appsettings.json#L57)):*
> ```JSON
>   },
>   "Machine": {
>     "ApiServer": "https://machine-api.org",
> ```

I think we discussed this briefly but I can't remember the conclusion. This domain isn't registered. Is it in a hosts file? It seems odd to me to deliberately use a top-level domain, though I think you were following what the NLP team was already doing?
___
*[`src/SIL.XForge.Scripture/Models/MachineApiCorpusFile.cs` line 7 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NI_Abk_12RAyiz2udF9:-NI_Abk_12RAyiz2udFA:brw47dl) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b60f70f9353b3de33208c2267dd97f56960f8a4/src/SIL.XForge.Scripture/Models/MachineApiCorpusFile.cs#L7)):*
> ```Smalltalk
>     /// </summary>
>     /// <remarks>
>     /// TODO: When Machine >= 2.5.12, change any code that uses this to use DataFileDto
> ```

Can you elaborate on the relationship between this class and `DataFileDto`, and the other similar instances of this (e.g. MachineApiTranslationEngine)? My current best guess is that there are certain properties the in-process instance of Machine needs, and there are certain properties the new network-based Machine instance needs, and that this class provides the information that will satisfy both (i.e. the union of the data each needs).

Though I see `DataFileDto` defines:
- `ResourceDto Corpus { get; set; }`
- `string LanguageTag { get; set; }`
- `string Name { get; set; }`
- `string TextId { get; set; }`

And that seems to contradict my conjecture, which leaves me with no guess as to how this is working.
___
*[`src/SIL.XForge.Scripture/Models/MachineData.cs` line 27 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NI_O7eP9uaXAqVK7oLg:-NI_O7eP9uaXAqVK7oLh:b-9d0qyz) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b60f70f9353b3de33208c2267dd97f56960f8a4/src/SIL.XForge.Scripture/Models/MachineData.cs#L27)):*
> ```Smalltalk
>         /// The machine corpora.
>         /// </value>
>         public Dictionary<string, MachineCorpus> Corpora { get; set; } = new Dictionary<string, MachineCorpus>();
> ```

What are the keys of this dictionary? Corpus ids?

Nateowami (commented):
> Reviewed 1 of 35 files at r1, 3 of 8 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 14 of 37 files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Models/MachineApiCorpusFile.cs` line 7 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NI_Abk_12RAyiz2udF9:-NId3stm-9oNbWhPzyZK:b-d0tkj2) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b60f70f9353b3de33208c2267dd97f56960f8a4/src/SIL.XForge.Scripture/Models/MachineApiCorpusFile.cs#L7)):*

It was supposed to be the bare minimum of properties SF needed from the `DataFileDto`, but I see the confusion now. I have updated this object to match `DataFileDto` exactly.
</blockquote></details>

If it matches exactly, then why wouldn't we be able to use it directly?

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 14 of 37 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/Models/MachineApiCorpusFile.cs` line 7 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NI_Abk_12RAyiz2udF9:-NId8Jbh1IIJ3XgpaPjG:b-du3hc6) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b60f70f9353b3de33208c2267dd97f56960f8a4/src/SIL.XForge.Scripture/Models/MachineApiCorpusFile.cs#L7)):*

If it matches exactly, then why wouldn't we be able to use it directly?
</blockquote></details>

It is not available yet in the published Machine NuGet package, unfortunately. That is why I have left the TODO: comments to remove these objects when Machine is updated to a version greater than 2.5.X. When Machine is updated to a version with these DTOs, we will lose the in-process Machine instance, so it will be something that will have to take place later.

Nateowami (commented):
> Reviewed 2 of 35 files at r1, 1 of 5 files at r2, 2 of 12 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 22 of 37 files reviewed, 2 unresolved discussions (waiting on @pmachapman)

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 22 of 37 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/Models/MachineApiCorpusFile.cs` line 7 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NI_Abk_12RAyiz2udF9:-NIdEJg13UtCxupPy3YU:b-y5j8hp) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b60f70f9353b3de33208c2267dd97f56960f8a4/src/SIL.XForge.Scripture/Models/MachineApiCorpusFile.cs#L7)):*

Ah, I see. It should probably say `Machine > 2.5.X` rather than `Machine >= 2.5.X`. Or perhaps "when machine is updated to a version that includes the DTOs".
</blockquote></details>

:facepalm: Fixed. Thank you :-)
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineBuildServiceTests.cs` line 15 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdARe4EizjmxHmSr3D:-NIdET_E6sxgO6NMwph7:bh95xrj) ([raw file](https://github.com/sillsdev/web-xforge/blob/c062582bb8eb065d6a95070446b1d5ea9174c463/test/SIL.XForge.Scripture.Tests/Services/MachineBuildServiceTests.cs#L15)):*

I wonder if it might make sense to also declare the base address (`"http://localhost"`) here as well? 
</blockquote></details>

I have rewritten the HttpClient creation as a static method in TestEnvironment. I think that will help remove the copy-pasta boilerplate code, and help make it more maintainable if other properties have to be set in future.

Nateowami (changes requested):
> Reviewed 1 of 9 files at r3, 1 of 8 files at r6.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 20 of 37 files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 65 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdHEXI4u81_-BDS3rn:-NIdHEXI4u81_-BDS3ro:b-tg3gf) ([raw file](https://github.com/sillsdev/web-xforge/blob/c062582bb8eb065d6a95070446b1d5ea9174c463/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L65)):*
> ```Smalltalk
>                     cancellationToken
>                 );
>                 return corpus?.corpus.id == corpusId;
> ```

It looks to me like we are returning a boolean value indicating whether the corpus was successfully added to the translation engine. I also note that we check this value in the tests, but not in `MachineProjectService`.

If returning false implies failure, it should probably throw an exception. If it indicates something else, that should probably be documented, since I haven't been able to figure it out.
___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 82 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdImjjC6SnHx7vuYRj:-NIdImjjC6SnHx7vuYRk:b69u1d6) ([raw file](https://github.com/sillsdev/web-xforge/blob/c891ec812de7584c4639ae26c42bef9a11403df2/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L82)):*
> ```Smalltalk
>                 {
>                     name,
>                     format = paratext ? "Paratext" : "Text",
> ```

I don't think you came up with these values, but having content posted with the format set to "Text" or "Paratext" seems odd. I would expect values like "plain text" and "USFM" (Paratext is not a format, USFM is, and has an actual spec and version numbers: https://ubsicap.github.io/usfm/).
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineBuildServiceTests.cs` line 15 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdARe4EizjmxHmSr3D:-NIdIFpF7UtBSSW-11pJ:b-vs1pr4) ([raw file](https://github.com/sillsdev/web-xforge/blob/c062582bb8eb065d6a95070446b1d5ea9174c463/test/SIL.XForge.Scripture.Tests/Services/MachineBuildServiceTests.cs#L15)):*

I have rewritten the HttpClient creation as a static method in TestEnvironment. I think that will help remove the copy-pasta boilerplate code, and help make it more maintainable if other properties have to be set in future.
</blockquote></details>

Looks good.

Nateowami (commented):
> Reviewed 2 of 35 files at r1, 2 of 8 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 24 of 37 files reviewed, 2 unresolved discussions (waiting on @pmachapman)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 24 of 37 files reviewed, 3 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 50 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdMZib2bXM5zk2hLgp:-NIdMZib2bXM5zk2hLgq:ba59p3p) ([raw file](https://github.com/sillsdev/web-xforge/blob/c891ec812de7584c4639ae26c42bef9a11403df2/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L50)):*
> ```Smalltalk
>             ValidateId(corpusId);
>             string requestUri = $"translation-engines/{translationEngineId}/corpora";
>             using var response = await MachineClient.PostAsJsonAsync(
> ```

Do we know how an engine behaves when it has multiple corpora? Is the pairing within a corpora, or between corpora?

In other words, in a simple case with a project based on another project, would the source and target both be within a single corpus, or are there two corpora and they are set to be paired with each other?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 24 of 37 files reviewed, 3 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 50 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdMZib2bXM5zk2hLgp:-NIdPIIdBTQYDvTNJ1Ch:bveweee) ([raw file](https://github.com/sillsdev/web-xforge/blob/c891ec812de7584c4639ae26c42bef9a11403df2/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L50)):*

Do we know how an engine behaves when it has multiple corpora? Is the pairing within a corpora, or between corpora?

In other words, in a simple case with a project based on another project, would the source and target both be within a single corpus, or are there two corpora and they are set to be paired with each other?
</blockquote></details>

Thinking about this more, I think it's the former (a source and target file goes within the same corpus), but I'm not sure how it's determined which files pairs witch which. Maybe if two files in the same corpus have the same `TextId` they are considered paired?

Nateowami (commented):
> Reviewed 2 of 35 files at r1, 2 of 12 files at r4, 4 of 8 files at r6.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 32 of 37 files reviewed, 5 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/IMachineBuildService.cs` line 15 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIrnEDp5--9RluR1xIC:-NIrnEDp5--9RluR1xID:b-kaqtvm) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/IMachineBuildService.cs#L15)):*
> ```Smalltalk
>             CancellationToken cancellationToken
>         );
>         Task<BuildDto> StartBuildAsync(string translationEngineId, CancellationToken cancellationToken);
> ```

Am I right in thinking "build" means "train engine"?
___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 65 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdHEXI4u81_-BDS3rn:-NIdgJhgByB9eQ20H6Xc:b-henx7k) ([raw file](https://github.com/sillsdev/web-xforge/blob/c062582bb8eb065d6a95070446b1d5ea9174c463/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L65)):*

Good spotting - that slipped through. Fixed.
</blockquote></details>

Is there a reason we wouldn't throw an exception? In general I think returning failure codes is to be avoided, since the caller needs to remember to check for it, which is very easy to forget.
___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 50 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdMZib2bXM5zk2hLgp:-NIs559IDbCgOc1putoH:b-ucfkn4) ([raw file](https://github.com/sillsdev/web-xforge/blob/c891ec812de7584c4639ae26c42bef9a11403df2/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L50)):*

Multiple corpora are flattened together with the current SmtTransferEngine (see https://github.com/sillsdev/machine/blob/master/src/SIL.Machine.WebApi/Services/SmtTransferEngineBuildJob.cs#L97).

Yes, the pairing is within a corpora. `TextId` is a unique generated ID from Mongo, so I understand that pairing is instead based on the verse reference in each row of the text (see https://github.com/sillsdev/machine/blob/master/src/SIL.Machine/Corpora/ParallelTextCorpus.cs#L94).
</blockquote></details>

Thanks.
___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 82 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdImjjC6SnHx7vuYRj:-NIs5ix7Ak-JHADsJsmX:b-p5xpww) ([raw file](https://github.com/sillsdev/web-xforge/blob/c891ec812de7584c4639ae26c42bef9a11403df2/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L82)):*

The Paratext format is a zip file of a Paratext directory. My initial implementation used this, but I since discovered that this was the incorrect method to upload a translation to the Machine API. I have only carried these across from the new Machine API for the sake of completion, and in case they are needed in the future. I can strip out this functionality if you think it is just confusing, and likely just adding bloat?
</blockquote></details>

Now that I have Swagger UI opened again I see these are documented values. What do you mean by it being "incorrect"? I initially thought you meant it wasn't supported by the new API, but I see it there in the docs.
___
*[`src/SIL.XForge.Scripture/Services/MachineTranslationService.cs` line 23 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIrxplOAP1vnlY1gBCE:-NIrxplOAP1vnlY1gBCF:b-z705h8) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineTranslationService.cs#L23)):*
> ```Smalltalk
> 
>         public async Task<string> CreateTranslationEngineAsync(
>             string name,
> ```

I found it quite odd that this variable is named `name`, rather than `id`. It looks like we pass in the project ID. So maybe this is a place where the client can provide a unique identifier (unique to the client), but the actual engine ID will be determined by the Machine server? Maybe the name is non-critical, but is a place where we can include the project ID as a bit of extra metadata? Otherwise I'm not sure what its purpose is. Am I right in guessing that it's not used by the Machine API, except to return it to us when we fetch an engine?
___
*[`src/SIL.XForge.Scripture/Services/MachineTranslationService.cs` line 38 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIrypGy7OGwPHJNV6St:-NIrypGy7OGwPHJNV6Su:b-crstyj) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineTranslationService.cs#L38)):*
> ```Smalltalk
>                     sourceLanguageTag,
>                     targetLanguageTag,
>                     type = "SmtTransfer"
> ```

Do you know why the possible values here are `SmtTransfer, Nmt`? Why is "transfer" in the name of one but not the other? Also, are we able to reference the enum directly, rather than having a magic string? Or maybe it's only available in an updated version of Machine? If that's the case, maybe we should leave a TODO to switch it to using the enum later.
___
*[`src/SIL.XForge.Scripture/Services/MachineTranslationService.cs` line 118 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIs04M5BV9JmYrHdgKc:-NIs04M5BV9JmYrHdgKd:bscpnqv) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineTranslationService.cs#L118)):*
> ```Smalltalk
>         public async Task TrainSegmentAsync(
>             string translationEngineId,
>             SegmentPairDto segmentPair,
> ```

Does this pairing come from the SF client, based on its own understanding of what segment is paired with what segment? If so, that is significant, because I know the client has a bug in how it pairs segments (it's a bit of an edge case), and I had assumed it wasn't affecting the engine, since it pairs segments itself using its own logic. But if the client is supplying the pairing for an incremental training, then that bug may be affecting translation suggestions as well (though only for incremental training).

Nateowami (commented):
> Reviewed 1 of 8 files at r6.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 33 of 37 files reviewed, 6 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 50 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdMZib2bXM5zk2hLgp:-NIsA9aUDF1Zoc_0NBgV:b-z7duio) ([raw file](https://github.com/sillsdev/web-xforge/blob/c891ec812de7584c4639ae26c42bef9a11403df2/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L50)):*

Thanks.
</blockquote></details>

OK, actually this raises a followup question. In `MachineCorporaServiceTests.cs` I see this example corpus text: `"1-1\tLine 1\r\n1-2\tLine 2\tsr"` I think the `1-1` and `1-2` mean chapter verses 1 and 2. Does that mean the book is inferred from the file name? If that's the case, I would be curious how key Biblical terms are handled, as they don't have an associated book. Do you know if this format is documented anywhere?
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 276 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIs6rMiAsbjt6y8FehV:-NIs6rMiAsbjt6y8FehW:b-dz7zu3) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L276)):*
> ```Smalltalk
>                         projectSecret.MachineData.TranslationEngineId,
>                         corpusId,
>                         pretranslate: false,
> ```

I'm interpreting "pretranslate" as "start training the engine" (without a separate POST to `/translation-engines/{id}/builds`). Is that how you're understanding it?

Nateowami (commented):
> Reviewed 1 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 34 of 37 files reviewed, 12 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Models/MachineCorpusFile.cs` line 8 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsL6XPDSDvqsIMsCP8:-NIsL6XPDSDvqsIMsCP9:bjhqcpr) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Models/MachineCorpusFile.cs#L8)):*
> ```Smalltalk
>     public class MachineCorpusFile
>     {
>         public string FileChecksum { get; set; } = string.Empty;
> ```

I think this could use a doc comment saying it's the MD5 hash of the file contents.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 163 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsE2ee6aLMqvDba-Mr:-NIsE2ee6aLMqvDba-Ms:bvz9v7c) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L163)):*
> ```Smalltalk
>             }
> 
>             // Remove the corpus files
> ```

Does the Machine APi not automatically delete files when a corpus is deleted? I can see why it wouldn't automatically delete corpora (since they can be used by multiple engines, if I understand correctly, though I don't think we're making use of that).
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 175 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsH75w2UruxBEVl5Fv:-NIsH75w2UruxBEVl5Fw:bxelbie) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L175)):*
> ```Smalltalk
>                     {
>                         // A 404 means that the file does not exist
>                         if (e.StatusCode == HttpStatusCode.NotFound)
> ```

Shouldn't there be an else clause here (and throughout this method, in similar situations), so if it _isn't_ a 404, it isn't silently caught and then not even logged? I could see why we might not want it to be fatal, since it's in a loop, but right now we could have failures other than 404 and they're hidden.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 195 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsEkSA0TUpe3b7fhmT:-NIsEkSA0TUpe3b7fhmU:b-4s284e) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L195)):*
> ```Smalltalk
>                 catch (HttpRequestException e)
>                 {
>                     // A 404 means that the translation engine does not exist
> ```

Would it not imply that the corpus you are trying to remove does not exist? I suppose it might be logical that attempting to remove a corpus that doesn't exist might silently succeed (since the desired state has already been obtained), and that a 404 would indicate that the parent object (the translation engine) couldn't be found.

That would be logical, but it would be inconsistent with the behavior higher up, where a file is removed and the comment says "A 404 means that the file does not exist".

Here's a visual representation of what I'm seeing as inconsistent:

Removing this | From this | 404 Means
---------------------|---------------|---------------
A file | a corpus | the file doesn't exit
A corpus | an engine | the engine doesn't exit

Either there's a mistake here, or the meaning of a 404 seems to be inconsistent/unpredictable across methods (which seems highly undesirable, and could lead to mistakes).
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 211 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsEqC041w0n8GA6eCz:-NIsEqC041w0n8GA6eD-:bu6k4u3) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L211)):*
> ```Smalltalk
>                 catch (HttpRequestException e)
>                 {
>                     // A 404 means that the file does not exist
> ```

I think this is supposed to say "the corpus does not exist"
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 234 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsCgrm6qcU-kfhfdbz:-NIsCgrm6qcU-kfhfdc-:b-sxy7tc) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L234)):*
> ```Smalltalk
>         )
>         {
>             // We will return whether the corpus was updated
> ```

I think this might be more useful as a doc comment, since it's not obvious what the return value indicates (same for SyncTextCorpusAsync)
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 331 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsMAIn9S1AfhAqQRr2:-NIsMAIn9S1AfhAqQRr3:bc3qb51) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L331)):*
> ```Smalltalk
>         }
> 
>         private async Task<bool> SyncTextCorpusAsync(
> ```

This can wait until after the PR is merged, but I think this method should be refactored, as it's pretty hard to read and reason about.

(In the interests of expediting this PR, I would even say it *shouldn't* be refactored until after it's merged (unless it needs major changes or something))

Nateowami (changes requested):
Getting closer. I have 1.5 files left to review.

> Reviewed 1 of 8 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-:-NItDbdF9u06ycpDBj9b:b-eujym3)* status: 35 of 37 files reviewed, 17 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 397 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsr2w2BWksgDZQOKO6:-NIsr2w2BWksgDZQOKO7:bv2xsfm) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L397)):*
According to the docs, this:
> Appends the default line terminator to the end of the current `StringBuilder` object.

Given that we are using checksums, it seems like the format of this data should not depend on the environment it's running on. Git, for example, usually normalizes the line endings when checking code in to prevent issues between operating systems. Since this data isn't being written to disk on the client, only sent to the server, it seems like it doesn't need OS-specific line endings.
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 225 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsvi2T-TVrEOLvb3LX:-NIsvi2T-TVrEOLvb3LY:bmlgnr9) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L225)):*
Could this be updated to:
> The machine service requires the project secrets, so call it before removing them

It took me a while to understand the meaning after I read through it the first time, since I hadn't looked at the context and seen that the project secretes were deleted shortly afterwards.
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 251 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIswTPrAwHaGmM5QMOQ:-NIswTPrAwHaGmM5QMOR:b-h4ehe3) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L251)):*
> ```Smalltalk
>                 bool unsetSourceProject = settings.SourceParatextId == ProjectSettingValueUnset;
>                 string writingSystemTag = projectDoc.Data.WritingSystem.Tag;
>                 if ((writingSystemTag == null || settings.SourceParatextId != null) && !unsetSourceProject)
> ```

I find this conditional and the logic that follows very difficult to comprehend. If I understand correctly,
- There are two separate reasons we might need to fetch the list of projects.
- One is to update the writing systems tag
- The other is to update the source

And they're sort of sharing the if statement, and then sharing the ptProjects that results? Is there a simpler way this could be written so it's clearer what's happening? Maybe something like:
```
projects = null
if need to get source:
  projects = fetchProjects()
  source = projects.find(...)
if need to update lang tag:
  if projects = null:
    projects = fetchProjects()
  langTag = projects.find(...).langTag
```
I don't like this solution either (I'd prefer something cleaner), but it would allow us to make sure projects are only ever fetched once, and still the logic is easier to follow.

Another option would be:
```
needToGetSource = ...
needToGetLangTag = ...
if needToGetSource or needToGetLangTag:
  projects = fetchProjects()
  if needToGetSource:
    ...
  if needToGetLangTag:
    ...
```
That's much closer to the current structure, but the use of variable names could make it a lot easier to follow the logic.
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineProjectServiceTests.cs` line 84 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsgcr3BpN4ALVwiZFp:-NIsgcr3BpN4ALVwiZFq:b-ssn03c) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/test/SIL.XForge.Scripture.Tests/Services/MachineProjectServiceTests.cs#L84)):*
> ```Smalltalk
>                 .Returns(
>                     Task.FromResult<ITextCorpus>(
>                         new MockTextCorpus
> ```

Given the number of times a mock corpus is created, and the difficulty with which it's ready, it seems like it might be useful to create a method to create this mock corpus.

I think there is value in being verbose in tests (they become more self-contained), but it's hard enough to read this structure that I think it would be easier to read it once and then recognize that the same value is being used everywhere.

I did a global search, and it looks like the same 28 line are used 6 times, with a 7th time in SyncProjectCorporaAsync_UpdatesSourceAndTargetTexts differing only in `TextCorpusType.Source` being replaced with `Arg.Any<TextCorpusType>()` (which was a necessary change for that test). maybe just the `MockTextCorpus` could be refactored out?
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineProjectServiceTests.cs` line 216 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsfPB15hV_Hnh4vZrG:-NIsfPB15hV_Hnh4vZrH:b9wqxo7) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/test/SIL.XForge.Scripture.Tests/Services/MachineProjectServiceTests.cs#L216)):*
> ```Smalltalk
>             await env.MachineCorporaService
>                 .DidNotReceiveWithAnyArgs()
>                 .DeleteCorpusFileAsync(Corpus01, File02, CancellationToken.None);
> ```

This assertion looks to me like it is a redundant check, given the use of `DidNotReceiveWithAnyArgs()`, meaning checking against both `File01` and `File01` is redundant. Is it done just for the sake of consistency, or am I missing some other reason?
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineProjectServiceTests.cs` line 536 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NItDOZI3UDEHiNg1EiW:-NItDOZI3UDEHiNg1EiX:bi8mtnf) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/test/SIL.XForge.Scripture.Tests/Services/MachineProjectServiceTests.cs#L536)):*
> ```Smalltalk
> 
>         [Test]
>         public async Task SyncProjectCorporaAsync_UpdatesCorpusIfTextExists()
> ```

I've leaving this note here as a marker of my progress through the tests so I can pick up more easily later.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 20 of 37 files reviewed, 17 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/Models/MachineCorpusFile.cs` line 8 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsL6XPDSDvqsIMsCP8:-NJ1rBxqEkdpR3Qz-MQl:b1ye208) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Models/MachineCorpusFile.cs#L8)):*

I think this could use a doc comment saying it's the MD5 hash of the file contents.
</blockquote></details>

Great idea - done.
___
*[`src/SIL.XForge.Scripture/Services/IMachineBuildService.cs` line 15 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIrnEDp5--9RluR1xIC:-NJ1skTP4OCWFK8n3OlU:b-tmgws1) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/IMachineBuildService.cs#L15)):*

Am I right in thinking "build" means "train engine"?
</blockquote></details>

Yes. The in-process Machine and new Machine API both use methods beginning with "StartBuild", so I think train engine is just an SF way of describing it?
___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 65 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdHEXI4u81_-BDS3rn:-NJ26kW90ZrPlbncY99T:b-mgfhz6) ([raw file](https://github.com/sillsdev/web-xforge/blob/c062582bb8eb065d6a95070446b1d5ea9174c463/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L65)):*

Is there a reason we wouldn't throw an exception? In general I think returning failure codes is to be avoided, since the caller needs to remember to check for it, which is very easy to forget.
</blockquote></details>

Yes - excellent idea. Fixed.
___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 50 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdMZib2bXM5zk2hLgp:-NJ22Zs05Fc0rRyEgxLy:b-2m38ai) ([raw file](https://github.com/sillsdev/web-xforge/blob/c891ec812de7584c4639ae26c42bef9a11403df2/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L50)):*

OK, actually this raises a followup question. In `MachineCorporaServiceTests.cs` I see this example corpus text: `"1-1\tLine 1\r\n1-2\tLine 2\tsr"` I think the `1-1` and `1-2` mean chapter verses 1 and 2. Does that mean the book is inferred from the file name? If that's the case, I would be curious how key Biblical terms are handled, as they don't have an associated book. Do you know if this format is documented anywhere?
</blockquote></details>

The biblical terms are in a file called `$"{projectId}_biblical_terms"` - see the constructor of `SFBiblicalTermsText`.

From what I can see from the Machine source code, the `TextFileText` corpora (see https://github.com/sillsdev/machine/blob/master/src/SIL.Machine/Corpora/TextFileText.cs) does not use the book name. It seems that the book name in the file name is just to keep the filename unique and clearly labelled. The TextFileText source file linked to above is the only format documentation I am aware of.
___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 82 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdImjjC6SnHx7vuYRj:-NJ1uE8jAvrzRzWUZByl:b-3gdd02) ([raw file](https://github.com/sillsdev/web-xforge/blob/c891ec812de7584c4639ae26c42bef9a11403df2/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L82)):*

Now that I have Swagger UI opened again I see these are documented values. What do you mean by it being "incorrect"? I initially thought you meant it wasn't supported by the new API, but I see it there in the docs.
</blockquote></details>

Incorrect, as in Damien told me not to use it. SF uses a different way of aligning text segments than the Paratext format, and so we do these in `SFScriptureText` and upload each chapter as a text file.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 163 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsE2ee6aLMqvDba-Mr:-NJ2BKf9-7n5bOXgqnET:bsmnchd) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L163)):*

Does the Machine APi not automatically delete files when a corpus is deleted? I can see why it wouldn't automatically delete corpora (since they can be used by multiple engines, if I understand correctly, though I don't think we're making use of that).
</blockquote></details>

No. Calling `DELETE /corpora/{id}` only deletes the entity from the database (see https://github.com/sillsdev/machine/blob/master/src/SIL.Machine.WebApi/Services/EntityServiceBase.cs#L26). I figured it would be good housekeeping to delete the files as well to stop the server clogging up from unused files (which would be difficult to track down later to see if they were still in use or not).
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 175 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsH75w2UruxBEVl5Fv:-NJ2CnVg0HYQsE2acrLb:b5n1cxh) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L175)):*

Shouldn't there be an else clause here (and throughout this method, in similar situations), so if it _isn't_ a 404, it isn't silently caught and then not even logged? I could see why we might not want it to be fatal, since it's in a loop, but right now we could have failures other than 404 and they're hidden.
</blockquote></details>

Yes, good spotting. Fixed here and elsewhere in this method, and took the time to try and keep the lines of code here less than the line length of 120.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 195 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsEkSA0TUpe3b7fhmT:-NJ2HtBHEj-I-_5r1189:bhp320x) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L195)):*

Would it not imply that the corpus you are trying to remove does not exist? I suppose it might be logical that attempting to remove a corpus that doesn't exist might silently succeed (since the desired state has already been obtained), and that a 404 would indicate that the parent object (the translation engine) couldn't be found.

That would be logical, but it would be inconsistent with the behavior higher up, where a file is removed and the comment says "A 404 means that the file does not exist".

Here's a visual representation of what I'm seeing as inconsistent:

Removing this | From this | 404 Means
---------------------|---------------|---------------
A file | a corpus | the file doesn't exit
A corpus | an engine | the engine doesn't exit

Either there's a mistake here, or the meaning of a 404 seems to be inconsistent/unpredictable across methods (which seems highly undesirable, and could lead to mistakes).
</blockquote></details>

The comment is correct - see https://github.com/sillsdev/machine/blob/bd11cf391afb0617c3947a59aba9c9b05f7386f8/src/SIL.Machine.WebApi/Services/TranslationEngineService.cs#L129. If the corpus does not exist, a Success code is still returned.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 211 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsEqC041w0n8GA6eCz:-NJ2HEDR6n0_iGvrXR08:brpp5ji) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L211)):*

I think this is supposed to say "the corpus does not exist"
</blockquote></details>

Yes - good spotting. Fixed.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 234 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsCgrm6qcU-kfhfdbz:-NJ2Ia1A-0D-5dD8pM3d:bn9m6jo) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L234)):*

I think this might be more useful as a doc comment, since it's not obvious what the return value indicates (same for SyncTextCorpusAsync)
</blockquote></details>

Sounds good. Fixed for both.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 276 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIs6rMiAsbjt6y8FehV:-NJ2MGCb-6VZ6GQDJs5-:b-lcg9wf) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L276)):*

I'm interpreting "pretranslate" as "start training the engine" (without a separate POST to `/translation-engines/{id}/builds`). Is that how you're understanding it?
</blockquote></details>

Pretranslations are a feature of ClearML, and so not used by the SmtTransfer engine that SF uses. I've kept it for API completeness, but can remove it if you think it is just clutter.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 331 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsMAIn9S1AfhAqQRr2:-NJ1vgpB2ZNAVgy7G-Kf:bq1reb4) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L331)):*

This can wait until after the PR is merged, but I think this method should be refactored, as it's pretty hard to read and reason about.

(In the interests of expediting this PR, I would even say it *shouldn't* be refactored until after it's merged (unless it needs major changes or something))
</blockquote></details>

I refactored one key part of this method, as it made sense to do so while I was implementing your request below on line terminations.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 397 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsr2w2BWksgDZQOKO6:-NJ1zmhM61DEQ1R_0hoo:b-meef1i) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L397)):*

According to the docs, this:
> Appends the default line terminator to the end of the current `StringBuilder` object.

Given that we are using checksums, it seems like the format of this data should not depend on the environment it's running on. Git, for example, usually normalizes the line endings when checking code in to prevent issues between operating systems. Since this data isn't being written to disk on the client, only sent to the server, it seems like it doesn't need OS-specific line endings.
</blockquote></details>

Yes, good idea. I have normalized these to Unix EOLs (`\n`).
___
*[`src/SIL.XForge.Scripture/Services/MachineTranslationService.cs` line 23 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIrxplOAP1vnlY1gBCE:-NJ2Q5FNDuLq5uCrAFin:b9gbcki) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineTranslationService.cs#L23)):*

I found it quite odd that this variable is named `name`, rather than `id`. It looks like we pass in the project ID. So maybe this is a place where the client can provide a unique identifier (unique to the client), but the actual engine ID will be determined by the Machine server? Maybe the name is non-critical, but is a place where we can include the project ID as a bit of extra metadata? Otherwise I'm not sure what its purpose is. Am I right in guessing that it's not used by the Machine API, except to return it to us when we fetch an engine?
</blockquote></details>

The name is just a description, returned by the API. We do not use it to reference the translation engine, but I figured that using the project ID might help act as a weak "unofficial" reference from the Translation Engine to the Project, if we were debugging an issue. I wasn't sure what else to set it to, as it is required. It does not need to be unique.
___
*[`src/SIL.XForge.Scripture/Services/MachineTranslationService.cs` line 38 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIrypGy7OGwPHJNV6St:-NJ2PYxpCYqt6zUxyYh3:blftbwh) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineTranslationService.cs#L38)):*

Do you know why the possible values here are `SmtTransfer, Nmt`? Why is "transfer" in the name of one but not the other? Also, are we able to reference the enum directly, rather than having a magic string? Or maybe it's only available in an updated version of Machine? If that's the case, maybe we should leave a TODO to switch it to using the enum later.
</blockquote></details>

Yes, it is in a later version. I have added a parameter to the method that determines the string, to make it less obscure. There is already a TODO comment above the block to update the object.
___
*[`src/SIL.XForge.Scripture/Services/MachineTranslationService.cs` line 118 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIs04M5BV9JmYrHdgKc:-NJ2SdJA5VtsM3FqFSZj:bp60ixt) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineTranslationService.cs#L118)):*

Does this pairing come from the SF client, based on its own understanding of what segment is paired with what segment? If so, that is significant, because I know the client has a bug in how it pairs segments (it's a bit of an edge case), and I had assumed it wasn't affecting the engine, since it pairs segments itself using its own logic. But if the client is supplying the pairing for an incremental training, then that bug may be affecting translation suggestions as well (though only for incremental training).
</blockquote></details>

Yes, this does come directly from the SF client, both in the current in-process implementation, and in the new Machine API implementation. If you send me details of the bug, I can look at a resolution to it when I write the client-side machine-api v2 implementation.
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 225 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsvi2T-TVrEOLvb3LX:-NJ2TKJjCa-MkCCgBAi1:b-caoz23) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L225)):*

Could this be updated to:
> The machine service requires the project secrets, so call it before removing them

It took me a while to understand the meaning after I read through it the first time, since I hadn't looked at the context and seen that the project secretes were deleted shortly afterwards.
</blockquote></details>

Totally - done.
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 251 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIswTPrAwHaGmM5QMOQ:-NJ2Vbzj0M4o0u8Io2S3:b-fnwvfc) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L251)):*

I find this conditional and the logic that follows very difficult to comprehend. If I understand correctly,
- There are two separate reasons we might need to fetch the list of projects.
- One is to update the writing systems tag
- The other is to update the source

And they're sort of sharing the if statement, and then sharing the ptProjects that results? Is there a simpler way this could be written so it's clearer what's happening? Maybe something like:
```
projects = null
if need to get source:
  projects = fetchProjects()
  source = projects.find(...)
if need to update lang tag:
  if projects = null:
    projects = fetchProjects()
  langTag = projects.find(...).langTag
```
I don't like this solution either (I'd prefer something cleaner), but it would allow us to make sure projects are only ever fetched once, and still the logic is easier to follow.

Another option would be:
```
needToGetSource = ...
needToGetLangTag = ...
if needToGetSource or needToGetLangTag:
  projects = fetchProjects()
  if needToGetSource:
    ...
  if needToGetLangTag:
    ...
```
That's much closer to the current structure, but the use of variable names could make it a lot easier to follow the logic.
</blockquote></details>

After some thought, I have removed this logic from here, and created a separate method in SFProjectService to set the writing system tag, if it is missing. This also means that if we decide to have a migrator program, we can wrap this in a simple public function (or make it internal...however we decide to do it). It is also now a lot clearer, and because it will be used at most once per project, I think the minor additional overhead is worth it.



_Code quote:_
```Smalltalk
if ((writingSystemTag == null || settings.SourceParatextId != null) && !u
```
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineProjectServiceTests.cs` line 84 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsgcr3BpN4ALVwiZFp:-NJ2agLZ2qNUz_CQoitf:baz2qh4) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/test/SIL.XForge.Scripture.Tests/Services/MachineProjectServiceTests.cs#L84)):*

Given the number of times a mock corpus is created, and the difficulty with which it's ready, it seems like it might be useful to create a method to create this mock corpus.

I think there is value in being verbose in tests (they become more self-contained), but it's hard enough to read this structure that I think it would be easier to read it once and then recognize that the same value is being used everywhere.

I did a global search, and it looks like the same 28 line are used 6 times, with a 7th time in SyncProjectCorporaAsync_UpdatesSourceAndTargetTexts differing only in `TextCorpusType.Source` being replaced with `Arg.Any<TextCorpusType>()` (which was a necessary change for that test). maybe just the `MockTextCorpus` could be refactored out?
</blockquote></details>

Thank you for spotting that - I have cleaned it up.
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineProjectServiceTests.cs` line 216 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsfPB15hV_Hnh4vZrG:-NJ2dpX1APN_t6UMz273:bx7ky2n) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/test/SIL.XForge.Scripture.Tests/Services/MachineProjectServiceTests.cs#L216)):*

This assertion looks to me like it is a redundant check, given the use of `DidNotReceiveWithAnyArgs()`, meaning checking against both `File01` and `File01` is redundant. Is it done just for the sake of consistency, or am I missing some other reason?
</blockquote></details>

Good spotting - fixed.



_Code quote:_
```Smalltalk
.DeleteCorpusFileAsync(Corpus01, File02, CancellationToken.None);
```
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineProjectServiceTests.cs` line 536 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NItDOZI3UDEHiNg1EiW:-NJ2dH7c8hwLCcmUyC8p:bgbfz0o) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/test/SIL.XForge.Scripture.Tests/Services/MachineProjectServiceTests.cs#L536)):*

I've leaving this note here as a marker of my progress through the tests so I can pick up more easily later.
</blockquote></details>

I'm leaving a comment here so it doesn't disappear...

Nateowami (commented):
> Reviewed 15 of 16 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 35 of 37 files reviewed, 8 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 65 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdHEXI4u81_-BDS3rn:-NJ65VOC4f_TSAEjize_:b-vs1pr4) ([raw file](https://github.com/sillsdev/web-xforge/blob/c062582bb8eb065d6a95070446b1d5ea9174c463/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L65)):*

Yes - excellent idea. Fixed.
</blockquote></details>

Looks good.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 163 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsE2ee6aLMqvDba-Mr:-NJ65u80-e3nWvvxvUxN:b4ob8z1) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L163)):*

No. Calling `DELETE /corpora/{id}` only deletes the entity from the database (see https://github.com/sillsdev/machine/blob/master/src/SIL.Machine.WebApi/Services/EntityServiceBase.cs#L26). I figured it would be good housekeeping to delete the files as well to stop the server clogging up from unused files (which would be difficult to track down later to see if they were still in use or not).
</blockquote></details>

OK. This seems like something we may want to bring up with the NLP team, since we're not planning on being the only client, and presumably they want good housekeeping on their end without relying on clients to clean things up. But having this here makes sense.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 276 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIs6rMiAsbjt6y8FehV:-NJ68pFQAsTllwznuza1:b-hmgeqr) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L276)):*

Pretranslations are a feature of ClearML, and so not used by the SmtTransfer engine that SF uses. I've kept it for API completeness, but can remove it if you think it is just clutter.
</blockquote></details>

OK. I wasn't aware ClearML was involved in this API. 
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 397 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIsr2w2BWksgDZQOKO6:-NJ67DM44iQPO8VXbBma:b-vs1pr4) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L397)):*

Yes, good idea. I have normalized these to Unix EOLs (`\n`).
</blockquote></details>

Looks good.
___
*[`src/SIL.XForge.Scripture/Services/MachineTranslationService.cs` line 23 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIrxplOAP1vnlY1gBCE:-NJ60lGR9W7a70PYUe3A:bt8pus2) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineTranslationService.cs#L23)):*

The name is just a description, returned by the API. We do not use it to reference the translation engine, but I figured that using the project ID might help act as a weak "unofficial" reference from the Translation Engine to the Project, if we were debugging an issue. I wasn't sure what else to set it to, as it is required. It does not need to be unique.
</blockquote></details>

Makes sense; I just wanted to make sure I wasn't missing anything.
___
*[`src/SIL.XForge.Scripture/Services/MachineTranslationService.cs` line 38 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIrypGy7OGwPHJNV6St:-NJ64IXDBm15rflQ8uo7:b1btrgg) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineTranslationService.cs#L38)):*

Yes, it is in a later version. I have added a parameter to the method that determines the string, to make it less obscure. There is already a TODO comment above the block to update the object.
</blockquote></details>

Ah, good point. I missed the block comment.
___
*[`src/SIL.XForge.Scripture/Services/MachineTranslationService.cs` line 118 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIs04M5BV9JmYrHdgKc:-NJ64OQ47y91dTEgYoi3:b-2mv25r) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/MachineTranslationService.cs#L118)):*

Yes, this does come directly from the SF client, both in the current in-process implementation, and in the new Machine API implementation. If you send me details of the bug, I can look at a resolution to it when I write the client-side machine-api v2 implementation.
</blockquote></details>

I thought we had an issue for this bug I can't find it. I think it happens when we have two similar markers, e.g. two headings that both use the `\s` marker. If in the target editor the user puts the cursor in the second heading, the source highlights the first heading (at least sometimes). I can't find this in Jira though.

I'll see if I can find out who saw it and why we don't have an issue (or maybe we do and I just can't find it).
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 1152 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NJ6BpWGAtoPb_-mX_ch:-NJ6BpWGAtoPb_-mX_ci:bym5mrk) ([raw file](https://github.com/sillsdev/web-xforge/blob/bef5c5922b95f5d8ff1fee42eb4f791dc3e71095/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L1152)):*
> ```Smalltalk
>         }
> 
>         private async Task EnsureWritingSystemTagIsSetAsync(
> ```

If I understand correctly, this is only necessary because of a we've been failing to write it like we should? Has that been fixed? If so, this method should is basically incrementally migrating projects, if I'm not mistaken? It might be worth documenting the purpose of this method. It's pretty clear what it's doing, but not why it needs to exist.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 35 of 37 files reviewed, 3 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/IMachineBuildService.cs` line 15 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIrnEDp5--9RluR1xIC:-NJ6DEmo2a1Wi9622peC:bv819lu) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4bc420f8dfcb1cd9d3ca1a609d898f17fffedcb/src/SIL.XForge.Scripture/Services/IMachineBuildService.cs#L15)):*

Yes. The in-process Machine and new Machine API both use methods beginning with "StartBuild", so I think train engine is just an SF way of describing it?
</blockquote></details>

Yes. I just wasn't 100% certain about the terminology.
___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 50 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdMZib2bXM5zk2hLgp:-NJ6Ebdk6JN5yyDGFCvG:b-eg41t9) ([raw file](https://github.com/sillsdev/web-xforge/blob/c891ec812de7584c4639ae26c42bef9a11403df2/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L50)):*

The biblical terms are in a file called `$"{projectId}_biblical_terms"` - see the constructor of `SFBiblicalTermsText`.

From what I can see from the Machine source code, the `TextFileText` corpora (see https://github.com/sillsdev/machine/blob/master/src/SIL.Machine/Corpora/TextFileText.cs) does not use the book name. It seems that the book name in the file name is just to keep the filename unique and clearly labelled. The TextFileText source file linked to above is the only format documentation I am aware of.
</blockquote></details>

Let's discuss the later outside the context of the PR. I'd like to understand how segments are uniquely identified in order to be paired.
___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 82 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdImjjC6SnHx7vuYRj:-NJ6DlE3A-tXOvzNvZaR:ban7ybl) ([raw file](https://github.com/sillsdev/web-xforge/blob/c891ec812de7584c4639ae26c42bef9a11403df2/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L82)):*

Incorrect, as in Damien told me not to use it. SF uses a different way of aligning text segments than the Paratext format, and so we do these in `SFScriptureText` and upload each chapter as a text file.
</blockquote></details>

Can you leave the Paratext upload method but leave a doc comment saying it's not used and why?

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 35 of 37 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 82 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdImjjC6SnHx7vuYRj:-NJ6cLV-91Sx9wvOAkYd:b-dbgf5e) ([raw file](https://github.com/sillsdev/web-xforge/blob/c891ec812de7584c4639ae26c42bef9a11403df2/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L82)):*

Can you leave the Paratext upload method but leave a doc comment saying it's not used and why?
</blockquote></details>

I have created doc comments throughout the MachineCorporaService, as the behaviour of the methods and parameters seemed to have the most questions surrounding them.
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 1152 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NJ6BpWGAtoPb_-mX_ch:-NJ6eYjb00Nbo5oss_L2:b-1fxrb5) ([raw file](https://github.com/sillsdev/web-xforge/blob/bef5c5922b95f5d8ff1fee42eb4f791dc3e71095/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L1152)):*

If I understand correctly, this is only necessary because of a we've been failing to write it like we should? Has that been fixed? If so, this method should is basically incrementally migrating projects, if I'm not mistaken? It might be worth documenting the purpose of this method. It's pretty clear what it's doing, but not why it needs to exist.
</blockquote></details>

Great idea. I have documented the fine details in the doc comment remarks.



_Code quote:_
```Smalltalk
EnsureWritingSystemTagIsSetAsync(
```

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: 35 of 37 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 1152 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NJ6BpWGAtoPb_-mX_ch:-NJ6kIRm85-5GuHiUBBN:b-biynut) ([raw file](https://github.com/sillsdev/web-xforge/blob/bef5c5922b95f5d8ff1fee42eb4f791dc3e71095/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L1152)):*

> This issue has since been fixed

Did you fix it in this PR? If so, where? I must have overlooked it. If not, then I don't think it's actually fixed.
</blockquote></details>

The fix was in ParatextService.cs, line 1713. See commit https://github.com/sillsdev/web-xforge/pull/1607/commits/947f9ea1aff717c6319d7aa711e1bd20e9101350

Nateowami (approved):
> Reviewed 2 of 2 files at r9.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/MachineCorporaService.cs` line 82 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NIdImjjC6SnHx7vuYRj:-NJ7-dXnCsfk6_Q1O8kH:b-ucfkn4) ([raw file](https://github.com/sillsdev/web-xforge/blob/c891ec812de7584c4639ae26c42bef9a11403df2/src/SIL.XForge.Scripture/Services/MachineCorporaService.cs#L82)):*

I have created doc comments throughout the MachineCorporaService, as the behaviour of the methods and parameters seemed to have the most questions surrounding them.
</blockquote></details>

Thanks.
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 1152 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-NJ6BpWGAtoPb_-mX_ch:-NJ7-HhB7CBR8ADQFf9K:b46f787) ([raw file](https://github.com/sillsdev/web-xforge/blob/bef5c5922b95f5d8ff1fee42eb4f791dc3e71095/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L1152)):*

The fix was in ParatextService.cs, line 1713. See commit https://github.com/sillsdev/web-xforge/pull/1607/commits/947f9ea1aff717c6319d7aa711e1bd20e9101350
</blockquote></details>

Ah, right. Thanks.

Nateowami (changes requested):
I'm seeing issues connecting new projects while on the branch. Errors are occurring because `await _featureManager.IsEnabledAsync(FeatureFlags.MachineApi)` is returning true at `src/SIL.XForge.Scripture/Services/MachineProjectService.cs:80`

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-:-NJ7BOZhBzNzv50qBwnP:b8hv256)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (commented):
The MachineApi feature flag is disabled by default, so it will likely be enabled via your user-secrets.

Run the following command in `src/SIL.XForge.Scripture/` to see the setting:

`dotnet user-secrets list`

You can disable it by running this command in `src/SIL.XForge.Scripture/`:

`dotnet user-secrets set "FeatureManagement:MachineApi" false`

Or, start the following Machine projects to use the new Machine API: ApiServer, EngineServer, and JobServer

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607#-:-NJ7DaLZArt14na3lRDC:b9i4196)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

Nateowami:
Is this ready to be rebased? I was about to test it, but given we've updated a number of dependencies, if this was rebased then my test would be more informative since it would test with the new dependencies (also, I wouldn't have to drop my database to go to a previous version when testing this).

Nateowami (dismissed):
> Reviewed 1 of 29 files at r10, 28 of 28 files at r11, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

Nateowami (approved):

Nateowami (approved):
> Reviewed 2 of 2 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1607)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #1610, opened 2022-11-22
Update jest-expect-message and remove @types/jest-expect-message

Nateowami:
- jest-expect-message ships its own types, and @types/jest-expect-message is deprecated ue to being unnecessary
- Updated tsconfig.json to import types from jest-expect-message

See https://github.com/mattphillips/jest-expect-message#configure-typescript for jest-expect-message's documentation for how to configure types.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1610)

### DIFF
--- a/src/RealtimeServer/common/utils/eq.spec.ts
+++ b/src/RealtimeServer/common/utils/eq.spec.ts
@@ -1,4 +1,5 @@
 import { eq } from './eq';
+import 'jest-expect-message';
 
 describe('eq', () => {
   it('compares two objects', () => {
--- a/src/RealtimeServer/package-lock.json
+++ b/src/RealtimeServer/package-lock.json
@@ -25,7 +25,6 @@
       },
       "devDependencies": {
         "@types/jest": "^27.5.1",
-        "@types/jest-expect-message": "^1.0.3",
         "@types/jsonwebtoken": "^8.5.8",
         "@types/node": "^16.11.36",
         "@types/ws": "^8.5.3",
@@ -38,7 +37,7 @@
         "eslint-plugin-jsdoc": "^36.1.1",
         "eslint-plugin-prettier": "^4.0.0",
         "jest": "^27.5.1",
-        "jest-expect-message": "^1.0.2",
+        "jest-expect-message": "^1.1.3",
         "jest-teamcity-reporter": "^0.9.0",
         "prettier": "^2.6.2",
         "sharedb-mingo-memory": "^1.2.0",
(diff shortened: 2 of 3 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1610)* status: 0 of 3 files reviewed, 1 unresolved discussion

___
*[`src/RealtimeServer/tsconfig.json` line 20 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1610#-NHVMTUjBJv1x8DibJZF:-NHVMTUjBJv1x8DibJZG:b-qp9xnm) ([raw file](https://github.com/sillsdev/web-xforge/blob/f0a92cc97ab60d6417e876e0ba8617ae98317676/src/RealtimeServer/tsconfig.json#L20)):*
> ```jsonc
>     }
>   },
>   "include": ["scriptureforge/**/*", "common/**/*", "typings/**/*", "node_modules/jest-expect-message/types/index.d.ts"]
> ```

See the documentation for this setting: https://www.typescriptlang.org/tsconfig#include

It says the default value is `[] if files is specified,,** otherwise.` We don't have `files` set anywhere, so this property is changing from `**` to listing which files should be included (it was broader before this change). I wasn't sure whether including `typings/**/*` was necessary, but since we were previously including everything, I figured including it was the safer option.

Usually you don't need to include a `.d.ts` file from a package like this; [you can just import it in the file where you need it](https://www.typescriptlang.org/docs/handbook/2/type-declarations.html#bundled-types):
``` ts
import X from 'x';
const x = new X();
```
The way this works is package `x` adds `"types": "./lib/main.d.ts"` to the `package.json`, and this is resolved by the compiler. But `jest-expect-message` doesn't have anything that you need to import to use, so unless you add it to `tsconfig.json`, it won't be found.

Another option would be to add `import 'jest-expect-message';` to a file where you want to have it available. This works, but it's unintuitive. A developer will see `expect(true, 'should be true').toBe(true)` working in one file, but not another. This is confusing, because there's nothing that obviously indicates that `jest-expect-message` is being used.

nigel-wells (approved):
Sometimes it would be nice if Reviewable could order the files to review by important things to know i.e. your very last comment which is way at the bottom. Also, sometimes you see tests written for this method you know nothing about yet as you haven't reached that part of the review.

> Reviewed 2 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1610#-:-NHflmAg9N8fmwJf-rs9:b-6bo2hw)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/RealtimeServer/tsconfig.json` line 20 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1610#-NHVMTUjBJv1x8DibJZF:-NHflQi44XXqw5Ltlk4X:b-ywlyjy) ([raw file](https://github.com/sillsdev/web-xforge/blob/f0a92cc97ab60d6417e876e0ba8617ae98317676/src/RealtimeServer/tsconfig.json#L20)):*

Well nevermind about all of this. This change worked for running the tests, but I forgot to run a build, and that's where it failed. One would think it would not be difficult to add a single file to be included.

The trouble comes from the fact that we have many tsconfig files that extend each other and can be used to do different builds. `src/RealtimeServer/common/tsconfig.build.json` extends this file, and adding these includes messes it up. I've tried a ton of combinations, and it's just not worth it to spend more time on it. I've just added `import 'jest-expect-message';` to the only file that uses it.
</blockquote></details>

I think that is an acceptable outcome. If we find we need it across other files as well in the future we can always revisit it.

PR #1614, opened 2022-11-23
SF-1825 Add pseudo localization

Nateowami:
Currently when the test team tests a new feature, it's not possible to tell whether it's been properly localized, because the strings haven't been translated yet, so it will always fall back to English.

This adds a locale behind a feature flag, with each string being generated by applying a [Caesar cipher](https://en.wikipedia.org/wiki/Caesar_cipher) with a right shift of 1 (a becomes b, b becomes c, z becomes a).

Here's an example from the settings page:

Our i18n service already has the concept of an "unpublished" localization, and already hides them behind a feature flag, so hiding this behind a feature flag is just a matter of registering the locale with `production: false`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1614)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts
@@ -18,9 +18,10 @@ import { BugsnagService } from './bugsnag.service';
 import { FeatureFlagService } from './feature-flags/feature-flag.service';
 import { LocationService } from './location.service';
 import { Locale } from './models/i18n-locale';
+import { PseudoLocalization } from './pseudo-localization';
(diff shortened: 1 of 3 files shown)


nigel-wells (approved):
This is a really great idea especially when you're writing a bunch of new strings. Something else that would be nice is if we could either change font or insert more characters to existing strings to help see how the interface responds to longer words that other languages can introduce

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1614#-:-NHfoR4c3ro7yg5fQkbD:b6tx9zc)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

nigel-wells (approved):
Sorry, that message was in draft!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1614#-:-NI9_2kOD-eRs4yrfiQ6:bezwcvy)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1615, opened 2022-11-23
SF-1822 Fix race condition for querying documents before share link has resolved

nigel-wells:
- Moved questions query in the app component to after a valid project has been selected

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1615)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
@@ -358,9 +358,6 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
         // the project deleted dialog should be closed by now, so we can reset its ref to null
         if (projectId == null) {
           this.projectDeletedDialogRef = null;
-        } else {
-          this.userService.setCurrentProjectId(this.currentUserDoc!, projectId);
-          this.refreshQuestionsQuery(projectId);
         }
       })
     );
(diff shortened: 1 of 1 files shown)


Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1615#-:-NHXiKKb8Mmixq4ClNhe:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 397 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1615#-NHXb18yCXRmDnsC2GRs:-NHXb18yCXRmDnsC2GRt:b-4ge7kz) ([raw file](https://github.com/sillsdev/web-xforge/blob/f56731c6eef1e33be62250d90353cb2b4004d1e3/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L397)):*
> ```XML
>       }
>       this.userService.setCurrentProjectId(this.currentUserDoc!, this.selectedProjectDoc.id);
>       this.refreshQuestionsQuery(this.selectedProjectDoc.id);
> ```

I think you identified the problem correctly. I was guessing this was what it was. My initial concern with this is that this line runs more often than the user changes projects. It runs whenever the user is added to or removed from a project. That's probably not very often, but this entire chunk of logic (everything in the subscribe) feels like it should be refactored to untangle some of this spaghetti (there are other race conditions in here, in part due to the complexity).

But that's probably something we should do in a future step, so I think it's better to get this approved and tested and shipped.

PR #1616, opened 2022-11-23
Add type safe I18nKey type

Nateowami:
Adds type checking of the localization keys we pass around, albeit not within the HTML templates (at least for now).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1616)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts
@@ -39,3 +39,50 @@ export interface CaretPosition {
   readonly offset: number;
   readonly offsetNode: Node;
 }
+
+// The following types up to ObjectPaths are fairly complicated. The earlier types are supporting types for the
+// ObjectPaths type definition
+// See https://stackoverflow.com/a/58436959 for an explanation of how this works.
+// Look up type manipulation in the TypeScript handbook for an explanation of the operators that are being used.
+
+// Define an array type that contains types that correspond to values ones less than the index they are at. This allows
+// subtracting one when evaluating the Leaves type below, thereby preventing excessive recursion (which would produce a
+// compiler error).
+// This is not a list of values. This is a list of types. For example, `5` is a type that contains only the number 5.
+type OneLessThan = [never, 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, ...10[]];
+
+/**
+ * Given two types K and P:
+ * - If `K` is a string or number, and `P` is a string or number, and P is not the empty string, constructs a string
+ * type consisting of the concatenation of `K` and `P` with a `.` between them.
+ * - If `K` is a string or number, and `P` is the empty string, constructs a string type consisting of `K` converted to
+ * a string.
+ * - Otherwise resolves to `never`.
+ */
+type JoinWithDot<K, P> = K extends string | number
+  ? P extends string | number
+    ? `${K}${'' extends P ? '' : '.'}${P}`
+    : never
+  : never;
+
+/**
+ * Given an object type `T`, where T has properties, and the keys to those properties are numbers or strings, and the
+ * values of those properties are strings or objects (which objects in turn satisfy the same conditions, recursively):
+ * Resolves to a type that is the union of all string paths to string values within the object structure, formed joining
+ * the key paths with the `.` character between each key.
+ *
+ * For example, `ObjectPaths<{ a: { b: 'c' }, d: 'e' }>` will resolve to `"d" | "a.b"`.
+ *
+ * @param T The type from which to derive paths.
+ * @param Depth The maximum recursive depth to search for string values. Default is 10. Maximum is 10, and setting it
+ * higher than 10 will have the same effect as setting it to 10. This should be considered an approximate value, rather
+ * than a precise value, since the meaning of searching to a given depth is not clearly defined, and the implementation
+ * has not been carefully crafted to ensure it is precisely followed with no off-by-one-errors (since the precise
+ * value is considered unimportant).
+ */
+// prettier-ignore
+export type ObjectPaths<T, Depth extends number = 10> = Depth extends never
+  ? never
+  : T extends object
+    ? { [Key in keyof T]-?: JoinWithDot<Key, ObjectPaths<T[Key], OneLessThan[Depth]>> }[keyof T]
+    : '';
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/dialog.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/dialog.service.ts
@@ -4,7 +4,7 @@ import { Injectable, TemplateRef } from '@angular/core';
 import { MatDialog, MatDialogConfig, MatDialogRef } from '@angular/material/dialog';
 import { Observable } from 'rxjs';
 import { GenericDialogComponent, GenericDialogOptions } from './generic-dialog/generic-dialog.component';
-import { I18nService } from './i18n.service';
+import { I18nKey, I18nService } from './i18n.service';
 
 @Injectable({
   providedIn: 'root'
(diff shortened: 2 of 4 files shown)


RaymondLuong3 (commented):
I like that we can see in the drop-down menu if a key exists or not. I have a couple questions.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1616#-:-NI-OFuc3-BoNSkm5HNh:b406if8)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts` line 29 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1616#-NI-BkHM7rQIjUp0qaf1:-NI-BkHM7rQIjUp0qaf2:bb7phvt) ([raw file](https://github.com/sillsdev/web-xforge/blob/c9b347dc556d746f06ffc515c70655081c25f903/src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts#L29)):*
> ```XML
> // Look up type manipulation in the TypeScript handbook for an explanation of the operators that are being used.
> 
> // Define an array type that contains types that correspond to values ones less the the index they are at. This allows
> ```

typo



_Code quote:_
```XML
to values ones less the the index they are at.
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts` line 36 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1616#-NI-D0osA3MZXtCSk_oK:-NI-D0osA3MZXtCSk_oL:bf2hcfd) ([raw file](https://github.com/sillsdev/web-xforge/blob/c9b347dc556d746f06ffc515c70655081c25f903/src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts#L36)):*
> ```XML
> 
> /**
>  * Given two types K and P:
> ```

Thanks for the explanation :D



_Code quote:_
```XML
/**
 * Given two types K and P:
 * - If `K` is a string or number, and `P` is a string or number, and P is not the empty string, constructs a string
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts` line 58 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1616#-NI-NgJN5zuimjl_cKKB:-NI-NgJN5zuimjl_cKKC:b-2vrsag) ([raw file](https://github.com/sillsdev/web-xforge/blob/c9b347dc556d746f06ffc515c70655081c25f903/src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts#L58)):*
> ```XML
>  *
>  * @param T The type from which to derive paths.
>  * @param Depth The maximum recursive depth to search for string values. Default is 10. Maximum is 10, and setting it
> ```

Is there a reason to have 10 as the depth? I see generally the depth is only 2 deep. Were you thinking that we would have paths that would be much longer?



_Code quote:_
```XML
 * @param Depth The maximum recursive depth to search for string values. Default is 10. Maximum is 10, and setting it
 * higher than 10 will have the same effect as setting it to 10. This should be considered an approximate value, rather
```

Nateowami (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1616#-:-NI0D7vJCGz7cR7kfmpd:bnfp4nl)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts` line 29 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1616#-NI-BkHM7rQIjUp0qaf1:-NI0C_AA4_qQIch1NJNU:b-k2rwtz) ([raw file](https://github.com/sillsdev/web-xforge/blob/c9b347dc556d746f06ffc515c70655081c25f903/src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts#L29)):*

typo
</blockquote></details>

Good catch. Fixed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts` line 58 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1616#-NI-NgJN5zuimjl_cKKB:-NI0Cd2Q1nSBWMQIJexb:bylik77) ([raw file](https://github.com/sillsdev/web-xforge/blob/c9b347dc556d746f06ffc515c70655081c25f903/src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts#L58)):*

Is there a reason to have 10 as the depth? I see generally the depth is only 2 deep. Were you thinking that we would have paths that would be much longer?
</blockquote></details>

Well, I wanted this to be a generic type definition, and not necessarily specific to localization files (not that I can think of anything else to apply it to). But unless a max recursion depth is set, it causes errors. So 10 seemed like a nice round number of max recursion depth that we would probably never run into even if this gets used for something I don't foresee (unless it's something really, really different, which probably shouldn't be trying to use it).

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1616)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts` line 68 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1616#-NI3SBdH-hI2KU7uWlqr:-NI3SBdI-2eZFJW_mHRD:bovf7w0) ([raw file](https://github.com/sillsdev/web-xforge/blob/85433a74d4c12a201c36194dd2c896684bc11a24/src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts#L68)):*
> ```XML
>   ? never
>   : T extends object
>     ? { [Key in keyof T]-?: JoinWithDot<Key, ObjectPaths<T[Key], OneLessThan[Depth]>> }[keyof T]
> ```

Not a big deal, but is there a reason that you had to use the `OneLessThan` array and you couldn't just do `Depth - 1`? I guess the `OneLessThan` array makes handling the max depth recursion work.



_Code quote:_
```XML
ey, ObjectPaths<T[Key], OneLessThan[Depth]>> }[keyof T]
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1616)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts` line 68 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1616#-NI3SBdH-hI2KU7uWlqr:-NI3UPfO0PtsUXj_atlj:b-itnw9k) ([raw file](https://github.com/sillsdev/web-xforge/blob/85433a74d4c12a201c36194dd2c896684bc11a24/src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts#L68)):*

Not a big deal, but is there a reason that you had to use the `OneLessThan` array and you couldn't just do `Depth - 1`? I guess the `OneLessThan` array makes handling the max depth recursion work.
</blockquote></details>

This is a type definition, so no JavaScript allowed. It's all evaluated by the compiler, so everything has to be done with type definitions, not mathematical expressions.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1616)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts` line 68 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1616#-NI3SBdH-hI2KU7uWlqr:-NI4N4CGAKBvuBZmerXf:b-otw2v6) ([raw file](https://github.com/sillsdev/web-xforge/blob/85433a74d4c12a201c36194dd2c896684bc11a24/src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts#L68)):*

This is a type definition, so no JavaScript allowed. It's all evaluated by the compiler, so everything has to be done with type definitions, not mathematical expressions.
</blockquote></details>

Right! That makes sense.

PR #1617, opened 2022-11-23
Fix catching of errors to be type safe

Nateowami:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1617)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts
@@ -189,7 +189,7 @@ export class ConnectProjectComponent extends DataLoadingComponent implements OnI
       let projectId: string = '';
       try {
         projectId = await this.projectService.onlineCreate(settings);
-      } catch (err: unknown) {
+      } catch (err) {
         if (!hasStringProp(err, 'message') || !err.message.includes(ConnectProjectComponent.errorAlreadyConnectedKey)) {
           throw err;
         }
--- a/src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts
@@ -1,4 +1,7 @@
-export function isObj(value: unknown): value is {} {
+/**
+ * Determines whether a value is a non-null object, and narrows the type to object if so.
+ */
+export function isObj(value: unknown): value is object {
   return typeof value === 'object' && value !== null;
 }
 
@@ -6,12 +9,28 @@ export function hasProp<X, Y extends PropertyKey>(value: X, property: Y): value 
   return isObj(value) && property in value;
 }
 
-export function hasStringProp<X, Y extends PropertyKey>(value: X, property: Y): value is X & Record<Y, String> {
+export function hasStringProp<X, Y extends PropertyKey>(value: X, property: Y): value is X & Record<Y, string> {
   return hasProp(value, property) && typeof value[property] === 'string';
 }
 
 export function hasFunctionProp<X, Y extends PropertyKey>(value: X, property: Y): value is X & Record<Y, Function> {
   return hasProp(value, property) && typeof value[property] === 'function';
+}
+
+export function hasObjectProp<X, Y extends PropertyKey>(value: X, property: Y): value is X & Record<Y, object> {
+  return hasProp(value, property) && isObj(value[property]);
+}
+
+export function hasNumberProp<X, Y extends PropertyKey>(value: X, property: Y): value is X & Record<Y, number> {
+  return hasProp(value, property) && typeof value[property] === 'number';
+}
+
+export function hasPropWithValue<X, Y extends PropertyKey>(
+  obj: X,
+  prop: Y,
+  val: unknown
+): obj is X & Record<Y, unknown> {
+  return hasProp(obj, prop) && obj[prop] === val;
 }
 
 // The CaretPosition type is no longer in the TypeScript DOM API type definitions, though it is still in Firefox,
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts
@@ -15,6 +15,7 @@ import { SystemRole } from 'realtime-server/lib/esm/common/models/system-role';
 import { of, Subscription, timer } from 'rxjs';
 import { filter, mergeMap } from 'rxjs/operators';
 import { PwaService } from 'xforge-common/pwa.service';
+import { hasPropWithValue } from '../type-utils';
 import { environment } from '../environments/environment';
 import { Auth0Service } from './auth0.service';
 import { BugsnagService } from './bugsnag.service';
@@ -274,7 +275,7 @@ export class AuthService {
       }
       await this.remoteStore.init(() => this.getAccessToken());
       return { loggedIn: true, newlyLoggedIn: false };
-    } catch (error: any) {
+    } catch (error) {
       await this.handleLoginError('tryLogIn', error);
       return { loggedIn: false, newlyLoggedIn: false };
     }
@@ -315,7 +316,7 @@ export class AuthService {
         }
       }
       return { loggedIn: true, newlyLoggedIn: false };
-    } catch (error: any) {
+    } catch (error) {
       await this.handleLoginError('tryOnlineLogIn', error);
       return { loggedIn: false, newlyLoggedIn: false };
     }
@@ -425,9 +426,9 @@ export class AuthService {
     });
   }
 
-  private async handleLoginError(method: string, error: object): Promise<void> {
+  private async handleLoginError(method: string, error: unknown): Promise<void> {
     console.error(error);
-    this.reportingService.silentError(`Error occurred in ${method}`, error);
+    this.reportingService.silentError(`Error occurred in ${method}`, ErrorReportingService.normalizeError(error));
     await this.dialogService.message('error_messages.error_occurred_login', 'error_messages.try_again');
   }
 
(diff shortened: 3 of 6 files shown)


RaymondLuong3 (commented):
I have a couple questions. Thanks for your work on this!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1617#-:-NI-TGrM44xQKaqM3G-C:byjptjs)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts` line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1617#-NI-RPEsA6hscq6E9ON3:-NI-RPEsA6hscq6E9ON4:bjs52sf) ([raw file](https://github.com/sillsdev/web-xforge/blob/33fa41bafc673d4aa264fc592b216c8a0f304ff0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L92)):*
> ```XML
>       let data: unknown;
>       if (error instanceof ErrorEvent) {
>         const errorEvent: ErrorEvent = error.error;
> ```

This is probably on purpose, but why is it important to assign errorEvent to `error.error`, and not just `error`?



_Code quote:_
```XML
    const errorEvent: ErrorEvent = error.error;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts` line 217 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1617#-NI-T1S0EczAqwjsIVKV:-NI-T1S1AsPAwr8jIgvS:b-lxsmby) ([raw file](https://github.com/sillsdev/web-xforge/blob/33fa41bafc673d4aa264fc592b216c8a0f304ff0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L217)):*
> ```XML
>     const messages = messageKeys
>       .map(key => (hasStringProp(error, key) ? error[key].split('\n')[0] : null))
>       .filter((s): s is string => s != null);
> ```

why is it important to specify the `s is string` type?



_Code quote:_
```XML
      .filter((s): s is string => s != null);
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1617)* status: 4 of 6 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts` line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1617#-NI-RPEsA6hscq6E9ON3:-NI0EWol3yBFPZXYtM1D:b-3fauyl) ([raw file](https://github.com/sillsdev/web-xforge/blob/33fa41bafc673d4aa264fc592b216c8a0f304ff0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L92)):*

This is probably on purpose, but why is it important to assign errorEvent to `error.error`, and not just `error`?
</blockquote></details>

We're making an HTTP request using Angular's `HttpClient`, which, when it fails, produces an `HttpErrorResponse`. Here's what `http.d.ts` says:

```
/**
 * A response that represents an error or failure, either from a
 * non-successful HTTP status, an error while executing the request,
 * or some other failure which occurred during the parsing of the response.
 *
 * Any error returned on the `Observable` response stream will be
 * wrapped in an `HttpErrorResponse` to provide additional context about
 * the state of the HTTP layer when the error occurred. The error property
 * will contain either a wrapped Error object or the error response returned
 * from the server.
 *
 * @publicApi
 */
export declare class HttpErrorResponse extends HttpResponseBase implements Error {
    readonly name = "HttpErrorResponse";
    readonly message: string;
    readonly error: any | null;
    /**
     * Errors are never okay, even when the status code is in the 2xx success range.
     */
    readonly ok = false;
    constructor(init: {
        error?: any;
        headers?: HttpHeaders;
        status?: number;
        statusText?: string;
        url?: string;
    });
}
```

So to summarize: When we catch an error here, we don't know where it came from. And if it came from `HttpClient`, it will be an object with information, and a property called `error` that has the original error.

All that said, I made a mistake, checking if `error instanceof ErrorEvent` rather than `error.error instanceof ErrorEvent`, which I've fixed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts` line 217 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1617#-NI-T1S0EczAqwjsIVKV:-NI0IFcqAjgYl23s1ndU:b-ddcc8j) ([raw file](https://github.com/sillsdev/web-xforge/blob/33fa41bafc673d4aa264fc592b216c8a0f304ff0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L217)):*

why is it important to specify the `s is string` type?
</blockquote></details>

This is a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates), which tells the compiler that when this function returns true, the value is a string, and when it returns false, the value is not a string.

`messageKeys.map(...)` returns an array of type `(string | null)[]`, and then we filter out the nullish values, leaving only string values. But TypeScript won't narrow the type to `string[]` unless we tell it to (it doesn't follow the logic and see that all null values have been filtered out, so all remaining values must be string values). But if we give it a type predicate it will understand and infer that `messages` has a type of `string[]`.

However, while writing this I realized that if the type predicate is removed everything still works; it just infers a type of `(string | null)[]`, which doesn't cause any compilation errors, because on the next line when we read `messages[0]` we use nullish coalescing to handle nullish values. Everything will work correctly, but `messages` may have null values. So I'm adding a specific type definition of `string[]`.

RaymondLuong3 (approved):
Thanks for this improvement.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1617#-:-NI3UE2R3HAau145d7ZT:bab8xg1)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts` line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1617#-NI-RPEsA6hscq6E9ON3:-NI3TaBK7B1q2YCbuGZw:blhl843) ([raw file](https://github.com/sillsdev/web-xforge/blob/33fa41bafc673d4aa264fc592b216c8a0f304ff0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L92)):*

We're making an HTTP request using Angular's `HttpClient`, which, when it fails, produces an `HttpErrorResponse`. Here's what `http.d.ts` says:

```
/**
 * A response that represents an error or failure, either from a
 * non-successful HTTP status, an error while executing the request,
 * or some other failure which occurred during the parsing of the response.
 *
 * Any error returned on the `Observable` response stream will be
 * wrapped in an `HttpErrorResponse` to provide additional context about
 * the state of the HTTP layer when the error occurred. The error property
 * will contain either a wrapped Error object or the error response returned
 * from the server.
 *
 * @publicApi
 */
export declare class HttpErrorResponse extends HttpResponseBase implements Error {
    readonly name = "HttpErrorResponse";
    readonly message: string;
    readonly error: any | null;
    /**
     * Errors are never okay, even when the status code is in the 2xx success range.
     */
    readonly ok = false;
    constructor(init: {
        error?: any;
        headers?: HttpHeaders;
        status?: number;
        statusText?: string;
        url?: string;
    });
}
```

So to summarize: When we catch an error here, we don't know where it came from. And if it came from `HttpClient`, it will be an object with information, and a property called `error` that has the original error.

All that said, I made a mistake, checking if `error instanceof ErrorEvent` rather than `error.error instanceof ErrorEvent`, which I've fixed.
</blockquote></details>

Nice, now it makes sense why we are using `error.error`, it was confusing that we check that `error` was a certain object, and then use its error property.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts` line 217 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1617#-NI-T1S0EczAqwjsIVKV:-NI3U1R5C9fO5y_PKoU8:b-gkvjzv) ([raw file](https://github.com/sillsdev/web-xforge/blob/33fa41bafc673d4aa264fc592b216c8a0f304ff0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L217)):*

This is a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates), which tells the compiler that when this function returns true, the value is a string, and when it returns false, the value is not a string.

`messageKeys.map(...)` returns an array of type `(string | null)[]`, and then we filter out the nullish values, leaving only string values. But TypeScript won't narrow the type to `string[]` unless we tell it to (it doesn't follow the logic and see that all null values have been filtered out, so all remaining values must be string values). But if we give it a type predicate it will understand and infer that `messages` has a type of `string[]`.

However, while writing this I realized that if the type predicate is removed everything still works; it just infers a type of `(string | null)[]`, which doesn't cause any compilation errors, because on the next line when we read `messages[0]` we use nullish coalescing to handle nullish values. Everything will work correctly, but `messages` may have null values. So I'm adding a specific type definition of `string[]`.
</blockquote></details>

I see, that is useful to know. In the past I would get squiggles in my editor because of things like this.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1617)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts` line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1617#-NI-RPEsA6hscq6E9ON3:-NI3W2UFE7Uoa9_-l_Fu:b-y6wypn) ([raw file](https://github.com/sillsdev/web-xforge/blob/33fa41bafc673d4aa264fc592b216c8a0f304ff0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L92)):*

Nice, now it makes sense why we are using `error.error`, it was confusing that we check that `error` was a certain object, and then use its error property.
</blockquote></details>

Good catch! I didn't even notice it until quite a ways into coming up with an explanation.

PR #1618, opened 2022-11-23
Replace deprecated substr with substring

Nateowami:
`substr()` is non-standard

The following are the descriptions MDN gives for these methods:

For [`substr()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr):
> The `substr()` method returns a portion of the string, starting at the specified index and extending for a given number of characters afterwards.

For [`substring()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substring):
> The `substring()` method returns the part of the `string` between the start and end indexes, or to the end of the string.

So when the first argument is 0, they should be equivalent.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1618)

### DIFF
--- a/src/RealtimeServer/common/models/user.ts
+++ b/src/RealtimeServer/common/models/user.ts
@@ -19,7 +19,7 @@ export function getAuthType(authId: string): AuthType {
     return AuthType.Unknown;
   }
 
-  const authIdType = authId.substr(0, authId.lastIndexOf('|'));
+  const authIdType = authId.substring(0, authId.lastIndexOf('|'));
   if (authIdType.includes('paratext')) {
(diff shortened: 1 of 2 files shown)


nigel-wells (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1618#-:-NHuzEUDCMUlJ2AkruXh:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1603, opened 2022-11-16
SF-1791 - Fix UpdateAvailableEvent deprecation in PWA service

nigel-wells:
- Removed deprecated event with version updates
- Added a timer to check for new updates rather than just on load

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1603)

PR #1622, opened 2022-11-29
SF-1791 Fix UpdateAvailableEvent deprecation in PWA service

Nateowami:
The original PR for this, #1603 was merged and then I reverted it because if caused errors on localhost in every browser I tried. 

The error message in both Chrome and Firefox is "Service workers are disabled or not supported by this browser".

I'm not sure how @nigel-wells wouldn't have run into this while working on this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1622)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -623,8 +623,8 @@ class TestEnvironment {
     when(mockedPwaService.isOnline).thenReturn(this.browserOnline$.getValue() && this.webSocketOnline$.getValue());
     when(mockedPwaService.isBrowserOnline).thenReturn(this.browserOnline$.getValue());
     when(mockedPwaService.online).thenReturn(comesOnline);
-    when(mockedPwaService.onlineStatus).thenReturn(this.webSocketOnline$);
-    when(mockedPwaService.onlineBrowserStatus).thenReturn(this.browserOnline$);
+    when(mockedPwaService.onlineStatus$).thenReturn(this.webSocketOnline$);
+    when(mockedPwaService.onlineBrowserStatus$).thenReturn(this.browserOnline$);
     if (initialConnectionStatus === 'offline') {
       this.goFullyOffline();
     } else {
@@ -632,7 +632,7 @@ class TestEnvironment {
       this.goFullyOnline();
     }
     when(mockedFileService.notifyUserIfStorageQuotaBelow(anything())).thenResolve();
-    when(mockedPwaService.hasUpdate).thenReturn(this.hasUpdate$);
+    when(mockedPwaService.hasUpdate$).thenReturn(this.hasUpdate$);
     when(mockedMdcDialog.open(ProjectDeletedDialogComponent, anything())).thenReturn(
       instance(this.mockedProjectDeletedDialogRef)
     );
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
@@ -127,7 +127,7 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
 
     // Check full online status changes
     this.isAppOnline = pwaService.isOnline;
-    this.subscribe(pwaService.onlineStatus, status => {
+    this.subscribe(pwaService.onlineStatus$, status => {
       if (status !== this.isAppOnline) {
         this.isAppOnline = status;
         this.checkDeviceStorage();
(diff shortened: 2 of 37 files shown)


RaymondLuong3 (commented):
I am still getting the "Service workers are disabled or not supported by this browser" error. I tried to log out and log back in, but it doesn't seem to fix the problem. I tested both on Firefox and Edge.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1622#-:-NI4jEBF89rY0QH8P2zd:b-bbbulc)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts` line 9 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1622#-NI4gWWQBwppvAAI7F1v:-NI4gWWQBwppvAAI7F1w:bygli9d) ([raw file](https://github.com/sillsdev/web-xforge/blob/5f45bafce1511eea07bb8c3c084bd1e2ba093374/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts#L9)):*
> ```XML
> import { LocationService } from './location.service';
> 
> export const PWA_CHECK_FOR_UPDATES = 30_000;
> ```

Neat, I didn't know you could define a number with this format.



_Code quote:_
```XML
export const PWA_CHECK_FOR_UPDATES = 30_000;
```

Nateowami:
@nigel-wells

nigel-wells (commented):
Sorry for missing this. All my testing was after it was compiled as a PWA where it works fine. I'll need to add something to handle localhost or any other browser that might not support service workers

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1622#-:-NI50uSo4AdvqtttVScQ:bp8glda)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

nigel-wells (approved):
I've updated the subscription to unsubscribe if it catches an error

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1622#-:-NI5AEmN2_9Kn3InD2s4:b-2jwngv)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

Nateowami (commented):
> Reviewed 36 of 37 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1622)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts` line 47 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1622#-NI5ezv376Npzvyw_55e:-NI5ezv376Npzvyw_55f:b-inm3qq) ([raw file](https://github.com/sillsdev/web-xforge/blob/046e292486d4add5283dbabbff4edf995d1537aa/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts#L47)):*
> ```XML
>     // Check for updates periodically if enabled and the browser supports it
>     if (this.updates.isEnabled) {
>       const checkForUpdatesInterval$ = interval(PWA_CHECK_FOR_UPDATES).subscribe(() =>
> ```

I notice we're no longer unsubscribing to this. Probably not an actual problem, but I really don't like things that look wrong even if they are actually OK, because then you end up getting used to seeing apparent problems that aren't real problems, and you have to investigate them to find out whether it's actually an OK exception to the rule.

How about `interval(PWA_CHECK_FOR_UPDATES).pipe(takeUntil(this.ngUnsubscribe)).subscribe(() => { })`?

Presumably you could then destroy the component and not need `discardPeriodicTasks()` in the tests.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1622)* status: 34 of 37 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts` line 47 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1622#-NI5ezv376Npzvyw_55e:-NI9Ma7W74-5daEW1cLF:b7eu5au) ([raw file](https://github.com/sillsdev/web-xforge/blob/046e292486d4add5283dbabbff4edf995d1537aa/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts#L47)):*

I notice we're no longer unsubscribing to this. Probably not an actual problem, but I really don't like things that look wrong even if they are actually OK, because then you end up getting used to seeing apparent problems that aren't real problems, and you have to investigate them to find out whether it's actually an OK exception to the rule.

How about `interval(PWA_CHECK_FOR_UPDATES).pipe(takeUntil(this.ngUnsubscribe)).subscribe(() => { })`?

Presumably you could then destroy the component and not need `discardPeriodicTasks()` in the tests.
</blockquote></details>

Good call. Done. I was also able to remove the discard call as well in the tests.

Nateowami (commented):
:lgtm:

> Reviewed 1 of 3 files at r6, 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1622#-:-NI9PgEW83pjW8No1g3z:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

PR #1625, opened 2022-11-29
SF-1827 Add username for community checkers

RaymondLuong3:
* determine if a user is a community checker
* add username as content on the answer and comment

Community Checker Name

UPDATED
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1625)

### DIFF
--- a/src/SIL.XForge.Scripture/Models/SFProjectRole.cs
+++ b/src/SIL.XForge.Scripture/Models/SFProjectRole.cs
@@ -11,5 +11,19 @@ namespace SIL.XForge.Scripture.Models
         public const string Reviewer = "sf_reviewer";
         public const string CommunityChecker = "sf_community_checker";
         public const string SFObserver = "sf_observer";
+
+        public static bool IsParatextRole(string role)
+        {
+            switch (role)
+            {
+                case SFProjectRole.Administrator:
+                case SFProjectRole.Translator:
+                case SFProjectRole.Consultant:
+                case SFProjectRole.PTObserver:
+                    return true;
+                default:
+                    return false;
+            }
+        }
     }
 }
--- a/src/SIL.XForge.Scripture/Services/IParatextNotesMapper.cs
+++ b/src/SIL.XForge.Scripture/Services/IParatextNotesMapper.cs
@@ -21,6 +21,7 @@ namespace SIL.XForge.Scripture.Services
             XElement oldNotesElem,
             IEnumerable<IDocument<Question>> questionsDocs,
             Dictionary<string, ParatextUserProfile> ptProjectUsers,
+            Dictionary<string, string> userRoles,
             string answerExportMethod
         );
     }
--- a/src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs
@@ -11,6 +11,7 @@ using SIL.XForge.DataAccess;
 using SIL.XForge.Models;
 using SIL.XForge.Realtime;
 using SIL.XForge.Realtime.Json0;
+using SIL.XForge.Services;
 using SIL.XForge.Scripture.Models;
 using SIL.XForge.Utils;
 
@@ -29,6 +30,7 @@ namespace SIL.XForge.Scripture.Services
     {
         private readonly IRepository<UserSecret> _userSecrets;
         private readonly IParatextService _paratextService;
+        private readonly IUserService _userService;
         private readonly IStringLocalizer<SharedResource> _localizer;
         private readonly IOptions<SiteOptions> _siteOptions;
         private UserSecret _currentUserSecret;
@@ -40,6 +42,7 @@ namespace SIL.XForge.Scripture.Services
         public ParatextNotesMapper(
             IRepository<UserSecret> userSecrets,
             IParatextService paratextService,
+            IUserService userService,
             IStringLocalizer<SharedResource> localizer,
             IOptions<SiteOptions> siteOptions,
             IGuidService guidService
@@ -47,6 +50,7 @@ namespace SIL.XForge.Scripture.Services
         {
             _userSecrets = userSecrets;
             _paratextService = paratextService;
+            _userService = userService;
             _localizer = localizer;
             _siteOptions = siteOptions;
             _guidService = guidService;
@@ -91,6 +95,8 @@ namespace SIL.XForge.Scripture.Services
             string answerExportMethod
         )
         {
+            // Usernames of SF community checker users. Paratext users are mapped to null.
+            Dictionary<string, string> checkerUsernames = new Dictionary<string, string>();
             var version = (string)oldNotesElem.Attribute("version");
             Dictionary<string, XElement> oldCommentElems = GetOldCommentElements(oldNotesElem, ptProjectUsers);
 
@@ -121,6 +127,7 @@ namespace SIL.XForge.Scripture.Services
                             )
                         );
                         var answerPrefixContents = new List<object>();
+
                         // Questions that have empty texts will show in Paratext notes that it is audio-only
                         string qText = string.IsNullOrEmpty(question.Text)
                             ? _localizer[SharedResource.Keys.AudioOnlyQuestion, _siteOptions.Value.Name]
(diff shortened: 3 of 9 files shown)


nigel-wells (commented):
This is looking good. Your written note on your screenshot does raise a good UX question. You needed to write that there because it wasn't obvious in the first place - and your name happens to appear 4 times in your example. What do you think about adding more context i.e. `Answered by Raymond Lu3` or `Answer by Raymond Lu3 on Scripture Forge`. 

Another question I had is if we should include the Paratext name as well as one might wonder why it is missing off some notes? If we went with the extra context above then I think that would help make it clear it was answered on Scripture Forge.

Is there flexibility to style it at all with HTML similar to how `Galatians 1 questions.` is? Perhaps italic, or underline, or a change in color if possible.

Another option, which might work better for comments, is you could prefix the comment with `[Name]` i.e. `[Raymond Lu3] Comment from Paratext admin`. I know Microsoft Word does this when replying to emails if you're commenting inline from a previous thread.

What do you think?

> Reviewed 2 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1625#-:-NIEvqOC5FYmUtfmwCMR:b-4fxfng)* status: 2 of 4 files reviewed, all discussions resolved

nigel-wells (commented):
Good point about the language. We could potentially use the project language but that is making an assumption. 
I've had a play around with it now to get a better idea of how it changes between a PT user and a community checker user. My comment around the "Paratext name" was it potentially looking "odd" if some comments have [NAME] while others don't have it. I thought it might be ok for consistency to show it even if the user was a Paratext user. I think its ok. I'd still be interested to know if adding "Scripture Forge" to the community checking name would work ok. If there were other PT users on the project who weren't looking at Scripture Forge it might just give some additional context around who or where this [NAME] came from. Given that we don't want to use other English words how about we separate it with a pipe character i.e. [Nigel Wells | Scripture Forge]
I think some color would be good as well. Our designer suggested to try it first with the same grey color as the name already appearing to keep a consistency for the user to know that "grey" means someones name. It may look odd (or not?) when it is an answer as the name would appear under the PT name in the same color but it should be different enough that it wouldn't matter.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1625#-:-NJ2Lpd00k2FoIaqhGUf:baso4st)* status: 2 of 4 files reviewed, all discussions resolved

RaymondLuong3 (commented):
I like the [Checker Name | Scripture Forge] idea. I made a change to add that, so it looks like this now. 

Initially I didn't think much about adding color to the text, but when I looked into it, I had a hard time trying to get it to work. The serialization to xml from c# wasn't working like I thought it would, and eventually I gave up. Maybe someone else can give it a try in the future.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1625#-:-NJCG9SH7bh82wcRDX8m:b-6xu3yi)* status: 2 of 4 files reviewed, all discussions resolved (waiting on @nigel-wells)

nigel-wells (changes requested):
Thanks for trying the color. Having the name is still an improvement so let's not hold that up over the color and we can look at that later.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1625#-:-NJCNDmu0yJTzewbHBf7:b-tk02it)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs` line 369 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1625#-NJCMUcx48p4QYw1AL3c:-NJCMUcx48p4QYw1AL3d:benu0ds) ([raw file](https://github.com/sillsdev/web-xforge/blob/d59ec98e9e29f8b3716dd5d5518d5b6f28a41afc/src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs#L369)):*
> ```Smalltalk
>                 return username;
> 
>             var userSecretAttempt = await _userSecrets.TryGetAsync(userId);
> ```

Just checking, do we have a method already to state if a user is has a PT role? Seems like we probably would have this somewhere already.
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextNotesMapperTests.cs` line 734 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1625#-NJCN2DQ9z9umJc-RCBW:-NJCN2DQ9z9umJc-RCBX:bc7p923) ([raw file](https://github.com/sillsdev/web-xforge/blob/d59ec98e9e29f8b3716dd5d5518d5b6f28a41afc/test/SIL.XForge.Scripture.Tests/Services/ParatextNotesMapperTests.cs#L734)):*
> ```Smalltalk
>                 UserService
>                     .GetUsernameFromUserId(Arg.Any<string>(), Arg.Is<string>(u => u == "user04"))
>                     .Returns("User 04");
> ```

Feels like you could probably have a `Dictionary` or list of some kind and do a lookup based on what was requested



_Code quote:_
```Smalltalk
                ParatextService.GetParatextUsername(Arg.Is<UserSecret>(u => u.Id == "user01")).Returns("PT User 1");
                ParatextService.GetParatextUsername(Arg.Is<UserSecret>(u => u.Id == "user03")).Returns("PT User 3");

                UserService = Substitute.For<IUserService>();
                UserService
                    .GetUsernameFromUserId(Arg.Any<string>(), Arg.Is<string>(u => u == "user01"))
                    .Returns("User 01");
                UserService
                    .GetUsernameFromUserId(Arg.Any<string>(), Arg.Is<string>(u => u == "user03"))
                    .Returns("User 03");
                UserService
                    .GetUsernameFromUserId(Arg.Any<string>(), Arg.Is<string>(u => u == "user02"))
                    .Returns("User 02");
                UserService
                    .GetUsernameFromUserId(Arg.Any<string>(), Arg.Is<string>(u => u == "user04"))
                    .Returns("User 04");
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1625)* status: 2 of 9 files reviewed, 2 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs` line 369 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1625#-NJCMUcx48p4QYw1AL3c:-NJH1j0M8bIf0VpukpdI:bw8xhbr) ([raw file](https://github.com/sillsdev/web-xforge/blob/d59ec98e9e29f8b3716dd5d5518d5b6f28a41afc/src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs#L369)):*

Just checking, do we have a method already to state if a user is has a PT role? Seems like we probably would have this somewhere already.
</blockquote></details>

Yes, but it was only a private method in SFProjectService. I've done some refactoring to put it into SFProjectRole class. I've also updated the logic. Since it is not guaranteed that a user with a project secret is a PT user on the current project, we need the userRoles from the project to determine if a user has a PT role on the project.
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextNotesMapperTests.cs` line 734 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1625#-NJCN2DQ9z9umJc-RCBW:-NJH1tVQAg7AIlY7YtjO:btyeq5o) ([raw file](https://github.com/sillsdev/web-xforge/blob/d59ec98e9e29f8b3716dd5d5518d5b6f28a41afc/test/SIL.XForge.Scripture.Tests/Services/ParatextNotesMapperTests.cs#L734)):*

Feels like you could probably have a `Dictionary` or list of some kind and do a lookup based on what was requested
</blockquote></details>

Good thinking. Done.

nigel-wells (approved):
Thanks for working through that refactor

:lgtm:

> Reviewed 7 of 7 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1625#-:-NJIQsjB2Y4ajvv4KZpp:b-vstav2)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

Nateowami (commented):
Two questions:

1. Am I correct in thinking that upon syncing all existing notes will be updated?
2. What do you think about using `-` as a separator rather than `|`? To me it seems more fitting. I don't think the vertical line is very commonly in use.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1625#-:-NJIo6Vx4At75mMBRBlG:b2hbg1q)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
Yes, that is correct. If an answer is marked for export, then that answer will be updated with the new content. Hmm, I think I also like the hyphen just slightly more. I'll get it updated and merged.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1625#-:-NJQeaej1vM4iOkAWxt5:bw1vsb9)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1626, opened 2022-11-30
SF-1823 Web API for Machine API

pmachapman:
Adds support for accessing the new Machine API via a Web API accessible from the Scripture Forge frontend.

Key changes include:
 * Moving the previous In Process Machine API implementation to /machine-api/v1/
 * Creating a fully compatible new Machine API implementation at /machine-api/v2/ as a drop in replacement for V1
 * Added a new `GetBuildById` endpoint to the C# API to enable better compatibility with the V1 API client (machine.js)
 * Fixed a bug where the Machine Build would not run on first sync (SF-1517)
 * Added a `MachineInMemory` feature flag to enable/disable the In Process Machine API (enabled by default)
 * Implemented the running of both the In Process and Machine API processes in tandem during the transition phase
 * Adding a Migration program to migrate projects to the new Machine API
 * Fixed a bug where the writing system tag was not set for Back Translations
 * Fixed a bug where verse segments (such as `\li`) were not included in a range, resulting in it being excluded from the training corpus

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626)

### DIFF
--- a/src/Migrations/MachineApiMigration/MachineApiMigration.csproj
+++ b/src/Migrations/MachineApiMigration/MachineApiMigration.csproj
@@ -1,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\SIL.XForge.Scripture\SIL.XForge.Scripture.csproj" />
+    <ProjectReference Include="..\..\SIL.XForge\SIL.XForge.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/src/Migrations/MachineApiMigration/MachineApiMigration.sln
+++ b/src/Migrations/MachineApiMigration/MachineApiMigration.sln
@@ -1,0 +1,42 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.4.33205.214
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "MachineApiMigration", "MachineApiMigration.csproj", "{FCE29858-6EF8-46CF-AB9A-390DA472F53E}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SIL.XForge", "..\..\SIL.XForge\SIL.XForge.csproj", "{665D56C2-6609-4477-91D0-16590AAEDDFD}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SIL.XForge.Scripture", "..\..\SIL.XForge.Scripture\SIL.XForge.Scripture.csproj", "{13D18568-D56E-4D4F-86C4-038DF06BBDDA}"
+EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{A3FFEBCE-071C-4FD7-BB5E-878802CA5302}"
+	ProjectSection(SolutionItems) = preProject
+		..\..\..\.editorconfig = ..\..\..\.editorconfig
+	EndProjectSection
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{FCE29858-6EF8-46CF-AB9A-390DA472F53E}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{FCE29858-6EF8-46CF-AB9A-390DA472F53E}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{FCE29858-6EF8-46CF-AB9A-390DA472F53E}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{FCE29858-6EF8-46CF-AB9A-390DA472F53E}.Release|Any CPU.Build.0 = Release|Any CPU
+		{665D56C2-6609-4477-91D0-16590AAEDDFD}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{665D56C2-6609-4477-91D0-16590AAEDDFD}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{665D56C2-6609-4477-91D0-16590AAEDDFD}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{665D56C2-6609-4477-91D0-16590AAEDDFD}.Release|Any CPU.Build.0 = Release|Any CPU
+		{13D18568-D56E-4D4F-86C4-038DF06BBDDA}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{13D18568-D56E-4D4F-86C4-038DF06BBDDA}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{13D18568-D56E-4D4F-86C4-038DF06BBDDA}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{13D18568-D56E-4D4F-86C4-038DF06BBDDA}.Release|Any CPU.Build.0 = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {EA40286F-FC4B-496F-B8D1-BC03011F6392}
+	EndGlobalSection
+EndGlobal
--- a/src/Migrations/MachineApiMigration/MachineApiMigrator.cs
+++ b/src/Migrations/MachineApiMigration/MachineApiMigrator.cs
@@ -1,0 +1,298 @@
+using System.IdentityModel.Tokens.Jwt;
+using System.Security.Claims;
+using Microsoft.FeatureManagement;
+using Org.BouncyCastle.Crypto;
+using SIL.XForge.DataAccess;
+using SIL.XForge.Models;
+using SIL.XForge.Realtime;
+using SIL.XForge.Scripture.Models;
+using SIL.XForge.Scripture.Services;
+
+namespace MachineApiMigration;
+
+/// <summary>
+/// Migrates all projects to the new Machine API, if translation is enabled.
+/// </summary>
+public class MachineApiMigrator
+{
+    /// <summary>
+    /// The feature manager.
+    /// </summary>
+    private readonly IFeatureManager _featureManager;
+
+    /// <summary>
+    /// The Machine project service.
+    /// </summary>
+    private readonly IMachineProjectService _machineProjectService;
+
+    /// <summary>
+    /// The Paratext service.
+    /// </summary>
+    private readonly IParatextService _paratextService;
+
+    /// <summary>
+    /// The project secrets repository.
+    /// </summary>
+    private readonly IRepository<SFProjectSecret> _projectSecrets;
+
+    /// <summary>
+    /// The project service.
+    /// </summary>
+    private readonly ISFProjectService _projectService;
+
+    /// <summary>
+    /// The realtime service.
+    /// </summary>
+    private readonly IRealtimeService _realtimeService;
+
+    /// <summary>
+    /// The user secrets repository.
+    /// </summary>
+    private readonly IRepository<UserSecret> _userSecrets;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="MachineApiMigrator" /> class.
+    /// </summary>
+    /// <param name="machineProjectService">The Machine project service.</param>
+    /// <param name="paratextService">The Paratext service.</param>
+    /// <param name="projectSecrets">The SF project secrets repository.</param>
+    /// <param name="projectService">The SF project service.</param>
+    /// <param name="realtimeService">The realtime service.</param>
+    /// <param name="userSecrets">The user secrets repository.</param>
+    public MachineApiMigrator(
+        IFeatureManager featureManager,
+        IMachineProjectService machineProjectService,
+        IParatextService paratextService,
+        IRepository<SFProjectSecret> projectSecrets,
+        ISFProjectService projectService,
+        IRealtimeService realtimeService,
+        IRepository<UserSecret> userSecrets
+    )
+    {
+        _featureManager = featureManager;
+        _machineProjectService = machineProjectService;
+        _paratextService = paratextService;
+        _projectSecrets = projectSecrets;
+        _projectService = projectService;
+        _realtimeService = realtimeService;
+        _userSecrets = userSecrets;
+    }
+
+    public async Task MigrateAllProjectsAsync(
+        bool doWrite,
+        ICollection<string> sfProjectIdsToMigrate,
+        IDictionary<string, string> sfAdminsToUse,
+        CancellationToken cancellationToken
+    )
+    {
+        // Ensure that the feature flags are set correctly
+        if (!await _featureManager.IsEnabledAsync(FeatureFlags.MachineApi))
+        {
+            Program.Log("ERROR: Cannot proceed. The Machine API feature flag is disabled.");
+            return;
+        }
+
+        if (await _featureManager.IsEnabledAsync(FeatureFlags.MachineInProcess))
+        {
+            Program.Log("ERROR: Cannot proceed. The In-Process Machine feature flag is enabled.");
+            return;
+        }
+
+        // Get all projects that have translation suggestions enabled, and are not configured for the Machine API
+        List<SFProject> projects = await _realtimeService
+            .QuerySnapshots<SFProject>()
+            .Where(p => p.TranslateConfig.TranslationSuggestionsEnabled)
+            .ToListAsync();
+
+        // If we are only to migrate some projects, restrict this list to them
+        if (sfProjectIdsToMigrate.Any())
+        {
+            projects.RemoveAll(p => !sfProjectIdsToMigrate.Contains(p.Id));
+            string ids = string.Join(' ', projects.Select(p => p.Id));
+            int count = projects.Count;
+            Program.Log($"Only working on the subset of projects (count {count}) with these SF project ids: {ids}");
+        }
+
+        // Get all of the project secrets that are not configured for the Machine API
+        List<SFProjectSecret> projectSecrets = await _projectSecrets
+            .Query()
+            .Where(p => p.MachineData == null || string.IsNullOrEmpty(p.MachineData!.TranslationEngineId))
+            .ToListAsync();
+
+        // Iterate over each project that is not configured for the Machine API
+        foreach (var project in projects.Where(p => projectSecrets.Select(ps => ps.Id).Contains(p.Id)))
+        {
+            Program.Log($"Migrating Project {project.Id}: {project.Name}");
+
+            // Get the admin users
+            List<string> projectSfAdminUserIds = project.UserRoles
+                .Where(ur => ur.Value == SFProjectRole.Administrator)
+                .Select(ur => ur.Key)
+                .ToList();
+            if (!projectSfAdminUserIds.Any())
+            {
+                List<string> projectSfUserIds = project.UserRoles.Select(ur => ur.Key).ToList();
+                string users = projectSfUserIds.Any() ? string.Join(", ", projectSfUserIds) : "None";
+                Program.Log($"Warning: No admin users. Non-admin users include: {users}");
+            }
+
+            // Report on all admins in a project
+            foreach (string sfUserId in projectSfAdminUserIds)
+            {
+                // See if we need to update the writing system tag
+                // We connect to the Paratext registry for this, so we need to make sure we can access it
+                if (string.IsNullOrWhiteSpace(project.WritingSystem.Tag))
+                {
+                    Program.Log("The writing system tag is not set, and will be updated.");
+
+                    // Ensure we have a user secret for this user
+                    UserSecret? userSecret = _userSecrets.Query().FirstOrDefault(us => us.Id == sfUserId);
+                    if (userSecret is null)
+                    {
+                        Program.Log($"Could not retrieve user secret for SF User: {sfUserId}");
+                        continue;
+                    }
+
+                    // Report on the Paratext username and id
+                    string ptUsername = string.Empty;
+                    string ptUserId;
+                    try
+                    {
+                        ptUsername = _paratextService.GetParatextUsername(userSecret) ?? string.Empty;
+                        ptUserId = GetParatextUserId(userSecret);
+                    }
+                    catch (Exception e)
+                    {
+                        Program.Log(
+                            "Failure getting SF user's PT username or PT user id. "
+                                + $"Skipping. SF user id was {sfUserId}. If known, PT username was {ptUsername}. "
+                                + $"Error with stack was {e}"
+                        );
+                        continue;
+                    }
+                    Program.Log(
+                        $"PT user '{ptUsername}', id {ptUserId}, using SF admin user id {sfUserId} on SF project."
+                    );
+
+                    // Display partial token details
+                    string rt = $"{userSecret.ParatextTokens.RefreshToken[..5]}..";
+                    string at = $"{userSecret.ParatextTokens.AccessToken[..5]}..";
+                    bool atv = userSecret.ParatextTokens.ValidateLifetime();
+                    Program.Log($"Paratext RefreshToken: {rt}, AccessToken: {at}, AccessToken initially valid: {atv}.");
+
+                    // Demonstrate access to PT Registry, and report Registry's statement of role.
+                    Program.Log("");
+                    IReadOnlyDictionary<string, string> ptProjectRoles;
+                    try
+                    {
+                        ptProjectRoles = await _paratextService.GetProjectRolesAsync(
+                            userSecret,
+                            project,
+                            cancellationToken
+                        );
+                    }
+                    catch (Exception e)
+                    {
+                        Program.Log($"Failure fetching user's PT project roles. Skipping. Error was {e.Message}");
+                        continue;
+                    }
+
+                    // Report that the user has a role on the project
+                    Program.Log(
+                        ptProjectRoles.TryGetValue(ptUserId, out string? ptRole)
+                            ? $"PT Registry report on role on PT project: {ptRole}"
+                            : "User role not found in PT Registry."
+                    );
+
+                    // Demonstrate access to PT Data Access.
+                    IReadOnlyList<ParatextProject> userPtProjects;
+                    try
+                    {
+                        userPtProjects = await _paratextService.GetProjectsAsync(userSecret);
+                    }
+                    catch (Exception e)
+                    {
+                        Program.Log($"Failure fetching user's PT projects. Skipping. Error was {e.Message}");
+                        continue;
+                    }
+
+                    // Report if the user can access a project
+                    List<string> ptProjectNamesList = userPtProjects
+                        .Where(ptProject => ptProject.ParatextId == project.ParatextId)
+                        .Select(ptProject => ptProject.ShortName)
+                        .ToList();
+                    if (ptProjectNamesList.Any())
+                    {
+                        string ptProjectNames = string.Join(',', ptProjectNamesList);
+                        Program.Log($"User can access PT Projects: {ptProjectNames}");
+                    }
+                    else
+                    {
+                        Program.Log(
+                            $"User is not on this project. PT reports they are on {userPtProjects.Count} PT projects"
+                        );
+                    }
+
+                    // If we are told to use specific admins, ensure they are used
+                    if (sfAdminsToUse.ContainsKey(project.Id))
+                    {
+                        sfAdminsToUse.TryGetValue(project.Id, out string? sfAdminIdToUse);
+                        bool isUserAtHand = sfUserId == sfAdminIdToUse;
+                        if (isUserAtHand)
+                        {
+                            Program.Log(
+                                $"For SF Project {project.Id}, we were asked to use this SF user {sfUserId} to migrate."
+                            );
+                        }
+                        else
+                        {
+                            Program.Log(
+                                $"For SF Project {project.Id}, we were asked to use SF user {sfAdminIdToUse}, "
+                                    + $"not {sfUserId}, to migrate. So skipping this user."
+                            );
+                            continue;
+                        }
+                    }
+
+                    if (doWrite)
+                    {
+                        await _projectService.EnsureWritingSystemTagIsSetAsync(sfUserId, project.Id);
+                    }
+                }
+                else
+                {
+                    Program.Log("The writing system tag does not need to be updated.");
+                }
+
+                // Add the project to the Machine API, and build it
+                if (doWrite)
+                {
+                    Program.Log("Adding project to Machine API...");
+                    await _machineProjectService.AddProjectAsync(sfUserId, project.Id, cancellationToken);
+                    Program.Log("Initiating first build...");
+                    await _machineProjectService.BuildProjectAsync(sfUserId, project.Id, cancellationToken);
+                }
+                else
+                {
+                    Program.Log("Project not migrated to Machine API, as test mode enabled.");
+                }
+
+                // We do not need to iterate any longer (the continue statements above ensure the correct user is used)
+                break;
+            }
+        }
+    }
+
+    /// <summary>
+    /// As claimed by tokens in userSecret. Looks like it corresponds to `userId` in PT Registry project members
+    /// query.
+    /// </summary>
+    private static string GetParatextUserId(UserSecret userSecret)
+    {
+        if (userSecret.ParatextTokens is null)
+            return string.Empty;
+        var accessToken = new JwtSecurityToken(userSecret.ParatextTokens.AccessToken);
+        Claim? claim = accessToken.Claims.FirstOrDefault(c => c.Type == "sub");
+        return claim?.Value ?? string.Empty;
+    }
+}
--- a/src/Migrations/MachineApiMigration/MachineApiMigrator.cs
+++ b/src/Migrations/MachineApiMigration/MachineApiMigrator.cs
@@ -1,9 +1,5 @@
-using System.IdentityModel.Tokens.Jwt;
-using System.Security.Claims;
 using Microsoft.FeatureManagement;
-using Org.BouncyCastle.Crypto;
 using SIL.XForge.DataAccess;
-using SIL.XForge.Models;
 using SIL.XForge.Realtime;
 using SIL.XForge.Scripture.Models;
 using SIL.XForge.Scripture.Services;
@@ -26,19 +22,9 @@ public class MachineApiMigrator
     private readonly IMachineProjectService _machineProjectService;
 
     /// <summary>
-    /// The Paratext service.
-    /// </summary>
-    private readonly IParatextService _paratextService;
-
-    /// <summary>
     /// The project secrets repository.
     /// </summary>
     private readonly IRepository<SFProjectSecret> _projectSecrets;
-
-    /// <summary>
-    /// The project service.
-    /// </summary>
-    private readonly ISFProjectService _projectService;
 
     /// <summary>
     /// The realtime service.
@@ -46,36 +32,23 @@ public class MachineApiMigrator
     private readonly IRealtimeService _realtimeService;
 
     /// <summary>
-    /// The user secrets repository.
-    /// </summary>
-    private readonly IRepository<UserSecret> _userSecrets;
-
-    /// <summary>
     /// Initializes a new instance of the <see cref="MachineApiMigrator" /> class.
     /// </summary>
+    /// <param name="featureManager">The feature manager.</param>
     /// <param name="machineProjectService">The Machine project service.</param>
-    /// <param name="paratextService">The Paratext service.</param>
     /// <param name="projectSecrets">The SF project secrets repository.</param>
-    /// <param name="projectService">The SF project service.</param>
     /// <param name="realtimeService">The realtime service.</param>
-    /// <param name="userSecrets">The user secrets repository.</param>
     public MachineApiMigrator(
         IFeatureManager featureManager,
         IMachineProjectService machineProjectService,
-        IParatextService paratextService,
         IRepository<SFProjectSecret> projectSecrets,
-        ISFProjectService projectService,
-        IRealtimeService realtimeService,
-        IRepository<UserSecret> userSecrets
+        IRealtimeService realtimeService
     )
     {
         _featureManager = featureManager;
         _machineProjectService = machineProjectService;
-        _paratextService = paratextService;
         _projectSecrets = projectSecrets;
-        _projectService = projectService;
         _realtimeService = realtimeService;
-        _userSecrets = userSecrets;
     }
 
     public async Task MigrateAllProjectsAsync(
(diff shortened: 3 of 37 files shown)


nigel-wells (changes requested):
I've done my first run through - only minor observations.
Are you able to provide some acceptance tests with this that I can run through.
It would be good to have some overall class comments to describe the purpose of each one
This PR would definitely benefit from someone who knows the Machine code really well - even if it was just to go over the main services.

> Reviewed 37 of 48 files at r1, 3 of 5 files at r2, 8 of 8 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-:-NInw_IFBI5BAEA9wcVq:b7l2s0t)* status: all files reviewed, 15 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/appsettings.json` line 59 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIeWixuAsJBdZyKh6fc:-NIeWixv14FKGwY2euvn:b-l0kwnl) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/appsettings.json#L59)):*
> ```JSON
>     "ApiServer": "https://machine-api.org",
>     "Audience": "https://machine.sil.org",
>     "TokenUrl": "https://sil-appbuilder.auth0.com/oauth/token"
> ```

Should this be `languagetechnology.auth0.com` and then have the other tenants defined in the other `appSettings.x.json` files?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/machine-http-client.ts` line 7 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIieymjCX7g2-AFSX8s:-NIieymjCX7g2-AFSX8t:b-lstsdh) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/ClientApp/src/app/core/machine-http-client.ts#L7)):*
> ```XML
> import { map } from 'rxjs/operators';
> 
> export const MACHINE_API_BASE_URL = 'machine-api/v2/';
> ```

Are we to default to `v1` for QA/Live and have `v2` on localhost? Or are you taking care of all that in the backend? This could have been an environment variable otherwise to specify which version to use. I'm not 100% sure on the rollout requirements though.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts` line 204 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIickS85ARJdTAp0jpH:-NIickS94MSUFcP7GTq7:b-mv9uxc) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L204)):*
> ```XML
>       error.url != null
>     ) {
>       // ignore 503 errors from ngsw-worker.js to machine-api,
> ```

Is this just handling when the Machine API is offline? How often would it get triggered as we'll potentially bombard Bugsnag with a heap of errors.
___
*[`src/SIL.XForge.Scripture/Controllers/MachineApiController.cs` line 38 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIipt6YEtLdumbcANsx:-NIipt6Z7IY-03c2G52E:b1a0qkw) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/Controllers/MachineApiController.cs#L38)):*
> ```Smalltalk
>         [HttpGet(MachineApi.GetBuild)]
>         public async Task<ActionResult<BuildDto?>> GetBuildAsync(
>             string projectId,
> ```

I'll mention it here once but it will apply everywhere. We had a recent discussion around variable names and particularly IDs. It isn't always obvious what ID it refers to without tracking it down. Is `projectId` a SF Project ID or a Paratext Project ID or something else? I believe we went for a prefix i.e. `sfProjectId`
___
*[`src/SIL.XForge.Scripture/Models/MachineApi.cs` line 19 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIijxtK7RaUQLM8hRo_:-NIijxtK7RaUQLM8hRoa:bqj7ape) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/Models/MachineApi.cs#L19)):*
> ```Smalltalk
>                 .Replace("{projectId}", projectId)
>                 .Replace("{buildId?}", buildId)
>                 .Replace("{reference}", "id");
> ```

Why not just have `id` as the string in `GetBuild` to start with?
___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 212 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NInZse0Bpz8RFDOoW4P:-NInZse0Bpz8RFDOoW4Q:b-39bmrj) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L212)):*
> ```Smalltalk
> 
>             // Execute the In Process Machine instance, if it is enabled
>             if (await _featureManager.IsEnabledAsync(FeatureFlags.MachineInProcess))
> ```

Will both feature flags ever be enabled at the same time as this one will overwrite the other and the former flag could be ignored if this one is enabled
___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 441 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIn_cliA3I_hlUmVcGn:-NIn_cliA3I_hlUmVcGo:b-e5xsmp) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L441)):*
> ```Smalltalk
> 
>             // Execute the In Process Machine instance, if it is enabled
>             if (await _featureManager.IsEnabledAsync(FeatureFlags.MachineInProcess))
> ```

Same as previous comment
___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 500 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIn_l8x6k7dEE26TIiT:-NIn_l8x6k7dEE26TIiU:b-e5xsmp) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L500)):*
> ```Smalltalk
> 
>             // Execute the In Process Machine instance, if it is enabled
>             if (await _featureManager.IsEnabledAsync(FeatureFlags.MachineInProcess))
> ```

Same as previous comment
___
*[`src/SIL.XForge.Scripture/Services/MachineBuildService.cs` line 27 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIoLg5C5hQ6MZBUbGeT:-NIoLg5D3QNZTwiq-WXT:b-5kscv3) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/Services/MachineBuildService.cs#L27)):*
> ```Smalltalk
>             ValidateId(translationEngineId);
> 
>             string requestUri = $"translation-engines/{translationEngineId}/current-build/cancel";
> ```

Could these (and corpora ones) be done in a similar way to `MachineApi.GetEngineHref`?
___
*[`src/SIL.XForge.Scripture/Services/MachineServiceBase.cs` line 39 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIjgl6pBVHkZhExoFpw:-NIjgl6pBVHkZhExoFpx:bhttw5f) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/Services/MachineServiceBase.cs#L39)):*
> ```Smalltalk
>         }
> 
>         protected void ValidateId(string id)
> ```

Can you add a comment around to its intent - similar to what we discussed
Also, are these Ids a set length as you could include that as well or perhaps within a range as that would help against bogus ids of `abc123`
___
*[`test/SIL.XForge.Scripture.Tests/Controllers/MachineApiControllerTests.cs` line 146 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NInun_01uvN5Ydl1m38:-NInun_1D91dhiwRQUNc:b-95fcfx) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/test/SIL.XForge.Scripture.Tests/Controllers/MachineApiControllerTests.cs#L146)):*
> ```Smalltalk
> 
>         [Test]
>         public async Task GetBuildAsync_NoBuildIdBuildEnded()
> ```

This test looks identical to `GetBuildAsync_BuildEnded`?

Also, should the name start with `GetCurrentBuildAsync`? Same with below tests
___
*[`test/SIL.XForge.Scripture.Tests/Controllers/MachineApiControllerTests.cs` line 151 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NInvdap0Mkgw86_GPH3:-NInvdap0Mkgw86_GPH4:b-uctaw0) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/test/SIL.XForge.Scripture.Tests/Controllers/MachineApiControllerTests.cs#L151)):*
> ```Smalltalk
>             env.MachineApiService
>                 .GetBuildAsync(User01, Project01, Build01, null, CancellationToken.None)
> ```

Should you be using `GetCurrentBuildAsync`?
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineBuildServiceTests.cs` line 68 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIoHS1-2X4P8nVMwLfy:-NIoHS1-2X4P8nVMwLfz:b-g3l3zw) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/test/SIL.XForge.Scripture.Tests/Services/MachineBuildServiceTests.cs#L68)):*
> ```Smalltalk
>                         ""href"": ""/translation-engines/{TranslationEngine01}""
>                     }},
>                     ""step"": 540,
> ```

Does this need to be asserted? Same with below tests.
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineBuildServiceTests.cs` line 73 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIoHcBw0h89UPZr6Hfo:-NIoHcBw0h89UPZr6Hfp:b-ypqyef) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/test/SIL.XForge.Scripture.Tests/Services/MachineBuildServiceTests.cs#L73)):*
> ```Smalltalk
>                     ""state"": ""{state}"",
>                     ""id"": ""{Build01}"",
>                     ""href"": ""/translation-engines/{TranslationEngine01}/builds/{Build01}""
> ```

Just noting this again if it needed to be asserted - same with below tests.
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineTranslationServiceTests.cs` line 109 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIo0yD61D38jUaAVoMA:-NIo0yD787FPo30JbXTw:blzyd75) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/test/SIL.XForge.Scripture.Tests/Services/MachineTranslationServiceTests.cs#L109)):*
> ```Smalltalk
>                 ""corpusSize"": {corpusSize},
>                 ""id"": ""{TranslationEngine01}"",
>                 ""href"": ""/translation-engines/{TranslationEngine01}""
> ```

Did you want to assert this as well? Initially I didn't comment on this but I see you are asserting it in other tests would pay to be consistent.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626)* status: all files reviewed, 15 unresolved discussions (waiting on @nigel-wells and @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts` line 204 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIickS85ARJdTAp0jpH:-NIsRkwv4QVBeUR1sTXa:b-ej3d8e) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L204)):*

Is this just handling when the Machine API is offline? How often would it get triggered as we'll potentially bombard Bugsnag with a heap of errors.
</blockquote></details>

This seems to me like the wrong place to handle an error like this. This is our application-wide catch-all global error handler. As much as possible, exceptions that we want to handle gracefully should be handled close to where they happen.

nigel-wells (changes requested):
Thanks for the updates - just a couple of comments. I'll start working through some testing now.

> Reviewed 23 of 23 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-:-NJ7ur75-8aybtgT51OM:bydt19n)* status: all files reviewed, 4 unresolved discussions (waiting on @Nateowami and @pmachapman)

___
*[`src/SIL.XForge.Scripture/appsettings.json` line 59 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIeWixuAsJBdZyKh6fc:-NJ7oUH82rH0dPvg5TiK:b-zgduo1) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/appsettings.json#L59)):*

Not that I am aware of? I was given these values by @nateowami from a Google doc I do not have access to.
</blockquote></details>

Thoughts here @nateowami ?
___
*[`src/SIL.XForge.Scripture/Models/MachineApi.cs` line 19 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIijxtK7RaUQLM8hRo_:-NJ7pODV8EizwRBC9WEU:bsr3809) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/Models/MachineApi.cs#L19)):*

Actually...I just remembered this needs to be here. The current version of machine.js specifies "engine" instead of "id", and so I need the {reference} parameter in the string.
</blockquote></details>

Does machine.js do its own string replacement on `{reference}`?
___
*[`src/SIL.XForge.Scripture/Services/MachineBuildService.cs` line 27 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIoLg5C5hQ6MZBUbGeT:-NJ7neWe954XHJzpVK1y:b-gxhrgj) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/Services/MachineBuildService.cs#L27)):*

They could. I didn't refactor these into a method, as each endpoint is slightly different, and I figured it would help when debugging to see the URL right there in a similar format to what you would use in Swagger to try and diagnose the issue. The methods in `MachineApi` only exists because they are called multiple times. If you think this is important, I can refactor them, but it might make the code a bit less easy to read?
</blockquote></details>

I'm ok with that as it is nice comparing to Swagger. My viewpoint was just around repeating the same 2/3 of the URL for every request
___
*[`src/SIL.XForge.Scripture/Services/MachineServiceBase.cs` line 39 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIjgl6pBVHkZhExoFpw:-NJ73qz63oivLwAWB1RN:bxxrssr) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/Services/MachineServiceBase.cs#L39)):*

I have added comments to help explain. I am not checking length, as the length is controlled by an external system (MongoDB), and may change implementation. I also do not see any vulnerability that might occur from a short length - the error passed from the Machine API should help a developer diagnose the issue if a short ID is being sent.
</blockquote></details>

Thanks for the comments added. Out of interest, does the Machine API have any protection in place from attacks? What would happen if  ID was bogus but passed this check - is our server at risk of being blocked for x amount of time by the Machine API server for repeated failures or any kind of rate limiting that we need to deal with?
___
*[`src/SIL.XForge.Scripture/Services/MachineTranslationService.cs` line 43 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJ7q-EOBdVaLzUTGN5Z:-NJ7q-EOBdVaLzUTGN5_:b-xcupa7) ([raw file](https://github.com/sillsdev/web-xforge/blob/805a4e6e4327bd5fd599c45ee51377b46cdf33c6/src/SIL.XForge.Scripture/Services/MachineTranslationService.cs#L43)):*
> ```Smalltalk
>                     sourceLanguageTag,
>                     targetLanguageTag,
>                     type = smtTransfer ? "SmtTransfer" : "Nmt",
> ```

What is this out of interest?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts` line 204 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIickS85ARJdTAp0jpH:-NJ7uJjT5m6YLvnA-NMG:b1nug96) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L204)):*

I have removed this. I am working on another PR that properly handles errors from machine-api on the frontend. When the machine API is offline, an error is flagged in the backend to bugsnag, which seems to be the correct place to do this?
</blockquote></details>

Yes, if the Machine API itself is offline then that can be handled initially in the backend. The frontend will still need to do something with the response it gets from the backend which sounds like what you're already working on.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626)* status: all files reviewed, 4 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/Models/MachineApi.cs` line 19 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIijxtK7RaUQLM8hRo_:-NJ7xj-2EoyLq85A37C3:b-enl55n) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/Models/MachineApi.cs#L19)):*

Does machine.js do its own string replacement on `{reference}`?
</blockquote></details>

It will sometimes substitute "engine", other times "id". See https://github.com/sillsdev/machine.js/blob/b46c0963cdaa1779cabc71c17fb254c1fd49f1c2/src/web-api/web-api-client.ts#L112 and https://github.com/sillsdev/machine/blob/2.5/src/SIL.Machine.WebApi/Controllers/BuildsController.cs#L141-L151 (although internally Machine and machine.js use the terminology locatorType).

Essentially, I take the URL parameter and ignore it as I do not use it. I named it reference, but if you can think of a clearer name, we can easily change it without issue. Maybe `ignore_this` or `kept_for_compatibility`?!
___
*[`src/SIL.XForge.Scripture/Services/MachineServiceBase.cs` line 39 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIjgl6pBVHkZhExoFpw:-NJ7zdyA7m9ckMjeR8jj:b86i7rn) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/Services/MachineServiceBase.cs#L39)):*

Thanks for the comments added. Out of interest, does the Machine API have any protection in place from attacks? What would happen if  ID was bogus but passed this check - is our server at risk of being blocked for x amount of time by the Machine API server for repeated failures or any kind of rate limiting that we need to deal with?
</blockquote></details>

`There is no rate limiting, as the Machine API is not publicly accessible. The ID is passed directly to Mongo - see https://github.com/sillsdev/machine/blob/7c9dc24522ad562088d37bb8ec9d9fade68282c1/src/SIL.Machine.WebApi/DataAccess/DataAccessExtensions.cs#L5-L15

Since SF uses Cloudflare, I think that will be appropriate to handle any rate abuse. Writing any custom rate limiting at our end might cause more issues than we fix.
___
*[`src/SIL.XForge.Scripture/Services/MachineTranslationService.cs` line 43 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJ7q-EOBdVaLzUTGN5Z:-NJ8-HWw1K_NTuh7OS94:bomfda2) ([raw file](https://github.com/sillsdev/web-xforge/blob/805a4e6e4327bd5fd599c45ee51377b46cdf33c6/src/SIL.XForge.Scripture/Services/MachineTranslationService.cs#L43)):*

What is this out of interest?
</blockquote></details>

SmtTransfer is the ML engine currently in use (Thot). Nmt will be used for the back translation machine learning, I believe (ClearML, I think?).

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/Models/MachineApi.cs` line 19 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIijxtK7RaUQLM8hRo_:-NJCV9P9BufkkTM3_FGg:b5yveq4) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/Models/MachineApi.cs#L19)):*

It will sometimes substitute "engine", other times "id". See https://github.com/sillsdev/machine.js/blob/b46c0963cdaa1779cabc71c17fb254c1fd49f1c2/src/web-api/web-api-client.ts#L112 and https://github.com/sillsdev/machine/blob/2.5/src/SIL.Machine.WebApi/Controllers/BuildsController.cs#L141-L151 (although internally Machine and machine.js use the terminology locatorType).

Essentially, I take the URL parameter and ignore it as I do not use it. I named it reference, but if you can think of a clearer name, we can easily change it without issue. Maybe `ignore_this` or `kept_for_compatibility`?!
</blockquote></details>

I guess you could call them `locatorType` as that is how they are referenced internally. Doesn't need to hold anything up though. Thanks.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/Models/MachineApi.cs` line 19 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIijxtK7RaUQLM8hRo_:-NJCWUCi74miL9GO6lJu:bynmu5y) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/Models/MachineApi.cs#L19)):*

I guess you could call them `locatorType` as that is how they are referenced internally. Doesn't need to hold anything up though. Thanks.
</blockquote></details>

Yes, good idea. Done.

nigel-wells (commented):
Initial testing observations:
* Changing user secrets did require the backend to be restarted
* If both `MachineApi` and `MachineInProcess` are set to `false` and you retrain, it gets stuck in and endless loop. This probably isn't really an issue as these wouldn't ever both be set to `false` but I'm mentioning in case there is something else that needs to be looked at as my expectation was that it wouldn't even try to train segments and the frontend wouldn't be stuck in a "Training" state that will never end.
* Getting a 404 not found error for one project when `MachineApi = true` and `MachineInProcess = false`. Project is, "Scc - SaudiArabic" and based on is resource, "aau - Abau New Testament". I'll add you as a PT admin.

I'll do some more testing this afternoon

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-:-NJCeInl9tHHjSTDR3mb:b-rrxdi4)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

nigel-wells (commented):
* With `MachineApi = true` and `MachineInProcess = false` I found as I typed and went to the overview screen it showed up saying "X trained segments". If I then clicked on "Retrain" it would reset this back to zero. If this doesn't happy for you then I'll add you to the project I was working on.
* On the same project as above, I set `MachineApi = true` and `MachineInProcess = true`, restarted backend, did not refresh the front end. Now clicking "Retain" results in a 404 as previously noted.
* On the same project as above, I set `MachineApi = false` and `MachineInProcess = true`, restarted backend, did not refresh the front end. Now clicking "Retain" still results in a 404.

I tried on another projected and also eventually ended up stuck in a 404 state. I'll stop testing for now until you've had time to look at that otherwise all my notes may end up with the same issue even if it isn't related to what I'm testing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-:-NJCzH63AJB2Kd7jY5gL:bl4e4ld)* status: 18 of 48 files reviewed, 4 unresolved discussions (waiting on @Nateowami and @pmachapman)

nigel-wells (changes requested):
Only other issue that came out of testing which feels like an issue but you did also say not to pay attention to the overview page at the moment.
When both `MachineApi = true` and `MachineInProcess = true`, and I retrain the engine, I get x number of trained segments:


If I then set `MachineInProcess = false`, and I retrain the engine, I get 0 trained segments. Is that to be expected at this point? It looks like training is happening though.


This appears to happen on any project.

Other than that, just minor feedback below.

> Reviewed 26 of 30 files at r7, 3 of 3 files at r8, 7 of 9 files at r9, 6 of 8 files at r10, 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-:-NJruB9b4CtaPZenFv5h:b7293ry)* status: all files reviewed, 11 unresolved discussions (waiting on @Nateowami and @pmachapman)

___
*[`src/Migrations/MachineApiMigration/MachineApiMigrator.cs` line 61 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJqmLEIEEnTqi6TxY6L:-NJqmLEJ9XsZrcXD0ef5:bnx12ym) ([raw file](https://github.com/sillsdev/web-xforge/blob/af2ba002cb0960e1497122002ce9c44ef84ed3e8/src/Migrations/MachineApiMigration/MachineApiMigrator.cs#L61)):*
> ```Smalltalk
>     )
>     {
>         // Ensure that the feature flags are set correctly
> ```

Is it possible to have a check to ensure the Machine app is running and accessible as well as node not in use ;)
___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 82 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJqpVqsE2kf3NFWDRQb:-NJqpVqsE2kf3NFWDRQc:bah6qlj) ([raw file](https://github.com/sillsdev/web-xforge/blob/805a4e6e4327bd5fd599c45ee51377b46cdf33c6/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L82)):*
> ```Smalltalk
>         )
>         {
>             BuildDto? buildDto = null;
> ```

I'm wondering why some of these you have removed the default `null` whereas on others you've set them to = `null`? In C#, what does it equal if no value is set - is it `undefined`? On the ones you are setting to `null` does that mean it no longer needs the "?" added as it is defined?  I'm not 100% sure in C# but that is how it would work in TS.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 463 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJrGJuE1OLkZn9eIp11:-NJrGJuE1OLkZn9eIp12:b-fwyk5v) ([raw file](https://github.com/sillsdev/web-xforge/blob/af2ba002cb0960e1497122002ce9c44ef84ed3e8/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L463)):*
> ```Smalltalk
> 
>                 // Strip characters from the key that will corrupt the line
>                 sb.Append(key.Replace('\n', '_').Replace('\t', '_'));
> ```

Does `\r` need to be considered as well? Or `\r\n`?
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 554 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJrGukz9w1lazlk2Fo_:-NJrGukz9w1lazlk2Foa:b-ikk319) ([raw file](https://github.com/sillsdev/web-xforge/blob/af2ba002cb0960e1497122002ce9c44ef84ed3e8/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L554)):*
> ```Smalltalk
>             // Get the language tag
>             string languageTag;
>             languageTag =
> ```

This could be made inline



_Code quote:_
```Smalltalk
            string languageTag;
            languageTag =
```
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 1191 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJqqjpWAYojgss-TCqC:-NJqqjpWAYojgss-TCqD:be1l3d4) ([raw file](https://github.com/sillsdev/web-xforge/blob/af2ba002cb0960e1497122002ce9c44ef84ed3e8/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L1191)):*
> ```Smalltalk
>         /// A issue was introduced in an early version of ScriptureForge where the writing system tag was not set when
>         /// the project was created. This issue has since been fixed. The Machine API requires the writing system tag
>         /// requires the writing system tag to be specified, and so this method should be called before a project is
> ```

nit: Double up in the wording here "requires the writing system tag"
___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 2525 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJrAvOSDfo4PelwZzaz:-NJrAvOT6p5ExyFDLjyS:b-p0of4l) ([raw file](https://github.com/sillsdev/web-xforge/blob/af2ba002cb0960e1497122002ce9c44ef84ed3e8/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L2525)):*
> ```Smalltalk
> 
>             // If the writing system tags are updated, the projects must be retrieved
>             await env.ParatextService.Received(1).GetProjectsAsync(Arg.Any<UserSecret>());
> ```

I don't think this is correct as the test is to assert that it isn't updated which you do below - perhaps the comment just needs to be amended.



_Code quote:_
```Smalltalk
            // If the writing system tags are updated, the projects must be retrieved
            await env.ParatextService.Received(1).GetProjectsAsync(Arg.Any<UserSecret>());
```

pmachapman (commented):
I have been unable to recreate this bug - the Migrator should start the first build, which will mean that this will have a value when you start web-xforge.

Were there any errors listed on the Machine Hangfire server at https://localhost:8666/hangfire/?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-:-NKizPv1E1Zirp4u7Is1:b-gbdmx7)* status: all files reviewed, 11 unresolved discussions (waiting on @github-code-scanning[bot], @Nateowami, and @nigel-wells)

___
*[`src/Migrations/MachineApiMigration/MachineApiMigrator.cs` line 61 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJqmLEIEEnTqi6TxY6L:-NKiiTy4-xGHU2Z2MPRY:b-6j5z3q) ([raw file](https://github.com/sillsdev/web-xforge/blob/2105312623324f64c076e2f906a0212ab7e2308e/src/Migrations/MachineApiMigration/MachineApiMigrator.cs#L61)):*

Is it possible to have a check to ensure the Machine app is running and accessible as well as node not in use ;)
</blockquote></details>

I have added a check to see if the Machine API is accessible.

Node already running throws an error which states "There was an error starting the program before getting to the migration part. Maybe the SF server is running and needs shut down?", so I don't think I need to check for that?
___
*[`src/SIL.XForge.Scripture/Controllers/MachineApiController.cs` line 67 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NI5FZ5gncBB95a6-0wl-r7-67:-NKiiiCx9iCEIedtM6IS:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/47d9e75841c6d1b2fce3a29e766ee5c2ece65901/src/SIL.XForge.Scripture/Controllers/MachineApiController.cs#L67)):*

## Missed ternary opportunity

Both branches of this 'if' statement write to the same variable - consider using '?' to express intent better.

[Show more details](https://github.com/sillsdev/web-xforge/security/code-scanning/317)
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/Controllers/MachineApiController.cs` line 65 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NI5FZ5gncBB95a6-0wl-r8-65:-NKiiiWP2DX_fg1EaBkr:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/22e5d8f30f97de6c0c927ee956e79aa3a066ac99/src/SIL.XForge.Scripture/Controllers/MachineApiController.cs#L65)):*

## Missed ternary opportunity

Both branches of this 'if' statement write to the same variable - consider using '?' to express intent better.

[Show more details](https://github.com/sillsdev/web-xforge/security/code-scanning/318)
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 82 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJqpVqsE2kf3NFWDRQb:-NKilW4_30jJjrv-ZNb7:b-vfv43u) ([raw file](https://github.com/sillsdev/web-xforge/blob/805a4e6e4327bd5fd599c45ee51377b46cdf33c6/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L82)):*

I'm wondering why some of these you have removed the default `null` whereas on others you've set them to = `null`? In C#, what does it equal if no value is set - is it `undefined`? On the ones you are setting to `null` does that mean it no longer needs the "?" added as it is defined?  I'm not 100% sure in C# but that is how it would work in TS.
</blockquote></details>

With the nullable annotations enabled, variable declarations should have the ? next to the type name to show that the object can be null. If that is not present, then the variable should not be able to be set to null. There is no undefined in C#.

My rule for whether some of the methods were nullable or not was based on whether the underlying Machine API endpoint was nullable or not (`null` corresponds to the HTTP response code 204).

Usually I want to avoid methods that return null, but if the endpoint demanded it, I thought it was logical to carry that across to the  method that calls it.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 463 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJrGJuE1OLkZn9eIp11:-NKimtkm8JmQhIcK-irJ:b2gvr8e) ([raw file](https://github.com/sillsdev/web-xforge/blob/2105312623324f64c076e2f906a0212ab7e2308e/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L463)):*

Does `\r` need to be considered as well? Or `\r\n`?
</blockquote></details>

No, it shouldn't as we send the text file with `\n` end of line delimiters. Also, if `\r` was found not next to a `\n`, it would not be interpreted as an end of line, if the Machine API was running on Windows.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 554 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJrGukz9w1lazlk2Fo_:-NKio8r96MZ6AwBoHXfp:bbuyr22) ([raw file](https://github.com/sillsdev/web-xforge/blob/2105312623324f64c076e2f906a0212ab7e2308e/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L554)):*

This could be made inline
</blockquote></details>

Good spotting. Fixed.
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 1191 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJqqjpWAYojgss-TCqC:-NKioqEMBe73NLs2WwwW:bqo5fir) ([raw file](https://github.com/sillsdev/web-xforge/blob/2105312623324f64c076e2f906a0212ab7e2308e/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L1191)):*

nit: Double up in the wording here "requires the writing system tag"
</blockquote></details>

Good spotting - thanks!
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineApiServiceTests.cs` line 69 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NI5FZ5gncBB95a6-0x4-r7-69:-NKiovzg10jpm1Kf8BZ8:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/47d9e75841c6d1b2fce3a29e766ee5c2ece65901/test/SIL.XForge.Scripture.Tests/Services/MachineApiServiceTests.cs#L69)):*

## Missing Dispose call on local IDisposable

Disposable 'Subscription<Build>' is created but not disposed.

[Show more details](https://github.com/sillsdev/web-xforge/security/code-scanning/315)
</blockquote></details>

Done.
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineApiServiceTests.cs` line 307 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NI5FZ5gncBB95a6-0x4-r7-307:-NKiowg04F8cXI9WSIss:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/47d9e75841c6d1b2fce3a29e766ee5c2ece65901/test/SIL.XForge.Scripture.Tests/Services/MachineApiServiceTests.cs#L307)):*

## Missing Dispose call on local IDisposable

Disposable 'Subscription<Build>' is created but not disposed.

[Show more details](https://github.com/sillsdev/web-xforge/security/code-scanning/316)
</blockquote></details>

Done.
___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 2525 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJrAvOSDfo4PelwZzaz:-NKirCtKA5vp9CcTRqKG:blnsygs) ([raw file](https://github.com/sillsdev/web-xforge/blob/2105312623324f64c076e2f906a0212ab7e2308e/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L2525)):*

I don't think this is correct as the test is to assert that it isn't updated which you do below - perhaps the comment just needs to be amended.
</blockquote></details>

Looking at this again after a week, the comments around the tests for `EnsureWritingSystemTagIsSetAsync` are unclear. I have tidied these up a bit.

nigel-wells (commented):
I don't see any errors. I can do a screenshare again if you think it might just be a configuration issue again at my end. Hit me up on slack when you're free.

> Reviewed 7 of 7 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-:-NLObv1BEElnAQoNvDbM:b-gxhocz)* status: all files reviewed, 7 unresolved discussions (waiting on @github-code-scanning[bot], @Nateowami, and @pmachapman)

___
*[`src/Migrations/MachineApiMigration/MachineApiMigrator.cs` line 61 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJqmLEIEEnTqi6TxY6L:-NLOEr_y2X2csiQunWAt:b2al0rw) ([raw file](https://github.com/sillsdev/web-xforge/blob/2105312623324f64c076e2f906a0212ab7e2308e/src/Migrations/MachineApiMigration/MachineApiMigrator.cs#L61)):*

I have added a check to see if the Machine API is accessible.

Node already running throws an error which states "There was an error starting the program before getting to the migration part. Maybe the SF server is running and needs shut down?", so I don't think I need to check for that?
</blockquote></details>

Thanks - happy with that. Just trying to cover bases for whomever decides to run this after us. Perhaps have a supporting document of the steps and requirements to run it. I'm not sure if our testers are going to go over this or not and, if they are, they will definitely need some steps to follow to get this up and running.
___
*[`src/Migrations/MachineApiMigration/MachineApiMigrator.cs` line 21 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NLOEYV453fMHP3teNOL:-NLOEYV453fMHP3teNOM:bv8nos1) ([raw file](https://github.com/sillsdev/web-xforge/blob/c90535ffc0bf0a7948ffdb8817a9d74fccf58d7b/src/Migrations/MachineApiMigration/MachineApiMigrator.cs#L21)):*
> ```Smalltalk
> 
>     /// <summary>
>     /// THe HTTP client to test if the Machine API is accessible.
> ```

nit: Change "THe" to "The"
___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 82 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJqpVqsE2kf3NFWDRQb:-NLOGPI9-Lj9yQrKr_0t:bge6tkm) ([raw file](https://github.com/sillsdev/web-xforge/blob/805a4e6e4327bd5fd599c45ee51377b46cdf33c6/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L82)):*

With the nullable annotations enabled, variable declarations should have the ? next to the type name to show that the object can be null. If that is not present, then the variable should not be able to be set to null. There is no undefined in C#.

My rule for whether some of the methods were nullable or not was based on whether the underlying Machine API endpoint was nullable or not (`null` corresponds to the HTTP response code 204).

Usually I want to avoid methods that return null, but if the endpoint demanded it, I thought it was logical to carry that across to the  method that calls it.
</blockquote></details>

Agreed - thanks for clarifying.
___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 2525 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NJrAvOSDfo4PelwZzaz:-NLOG29q9IekgCz1BPDZ:b-34myfe) ([raw file](https://github.com/sillsdev/web-xforge/blob/2105312623324f64c076e2f906a0212ab7e2308e/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L2525)):*

Looking at this again after a week, the comments around the tests for `EnsureWritingSystemTagIsSetAsync` are unclear. I have tidied these up a bit.
</blockquote></details>

Thanks - that's better.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626)* status: 57 of 58 files reviewed, 6 unresolved discussions (waiting on @github-code-scanning[bot], @Nateowami, and @nigel-wells)

___
*[`src/Migrations/MachineApiMigration/MachineApiMigrator.cs` line 21 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NLOEYV453fMHP3teNOL:-NLOgq_K3IShOtl3eYBm:bt2vr0y) ([raw file](https://github.com/sillsdev/web-xforge/blob/c90535ffc0bf0a7948ffdb8817a9d74fccf58d7b/src/Migrations/MachineApiMigration/MachineApiMigrator.cs#L21)):*

nit: Change "THe" to "The"
</blockquote></details>

Fixed!

pmachapman (commented):
I have managed to recreate the issue you were having and have fixed the bug. Please let me know if retraining after disabling the in process Machine is now working correctly for you.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-:-NLSyEr768_u1JE2Jah7:bnj5blz)* status: 55 of 58 files reviewed, 6 unresolved discussions (waiting on @github-code-scanning[bot], @Nateowami, and @nigel-wells)

nigel-wells (commented):
I'm happy other than that comment around the expected trained segment from V1 to V2. There is also an outstanding question for @nateowami and I've nudged him in Slack on this.

> Reviewed 1 of 1 files at r13, 2 of 2 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-:-NLTM2nb4JJ1uULHxbBR:b-b45itk)* status: all files reviewed, 5 unresolved discussions (waiting on @github-code-scanning[bot] and @Nateowami)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626)* status: 57 of 59 files reviewed, 5 unresolved discussions (waiting on @github-code-scanning[bot] and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/appsettings.json` line 59 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIeWixuAsJBdZyKh6fc:-NLX_Um90cHbrfUiYWEE:b-al7ivz) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/appsettings.json#L59)):*

Thoughts here @nateowami ?
</blockquote></details>

The way we've been doing settings is putting all the default or production values in `appsettings.json`, and overriding those values in files such as `appsettings.Staging.json` and `appsettings.Development.json`. So we shouldn't really have `sil-appbuilder` in this file, since it's currently only used in development mode. I think it should be in `appsettings.Development.json`.

pmachapman (commented):
Training segment has been sorted - there was a bug at out end. It is now only slightly different, which by my stepping through the code can determine is due to fixes in the new Machine API.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-:-NLXzc3a6wiYwXqPmkVc:b8eef5)* status: 56 of 62 files reviewed, 5 unresolved discussions (waiting on @github-code-scanning[bot] and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/appsettings.json` line 59 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-NIeWixuAsJBdZyKh6fc:-NLXz7R0DMNgDZ9jvKVk:b-yjgj0k) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ed99aa669e81e946a7ce6f508303273fe1fc5d4/src/SIL.XForge.Scripture/appsettings.json#L59)):*

The way we've been doing settings is putting all the default or production values in `appsettings.json`, and overriding those values in files such as `appsettings.Staging.json` and `appsettings.Development.json`. So we shouldn't really have `sil-appbuilder` in this file, since it's currently only used in development mode. I think it should be in `appsettings.Development.json`.
</blockquote></details>

Thank you - fixed!

nigel-wells (commented):
Thanks for working through this Peter - I think it is all working pretty well. Only odd thing, which may be me, but when I have `MachineApi = true` and the app running and on the translate overview page (I think it is on a book page as well) and I terminate dotnet (done when swapping between `true/false` settings, it produces a `Cannot access a disposed object` error
```
info: Microsoft.AspNetCore.Hosting.Diagnostics[2]
      Request finished HTTP/1.1 GET http://localhost:5000/machine-api/v2/translation/builds/engine:63bcdf1fe4328769581cde73?minRevision=0 application/json - - 499 82 application/problem+json;+charset=utf-8 12659.8872ms
Unhandled exception. System.ObjectDisposedException: Cannot access a disposed object.
   at System.Threading.TimerQueueTimer.Change(UInt32 dueTime, UInt32 period)
   at System.Threading.Timer.Change(Int32 dueTime, Int32 period)
   at SIL.Machine.WebApi.Utils.AsyncTimer.StopTimer()
   at SIL.Machine.WebApi.Utils.AsyncTimer.StopAsync()
   at SIL.Machine.WebApi.Utils.AsyncTimer.DisposeAsyncCore()
   at SIL.Machine.WebApi.Utils.AsyncDisposableBase.DisposeAsync()
   at SIL.Machine.WebApi.Services.EngineService.DisposeAsyncCore()
   at SIL.Machine.WebApi.Utils.AsyncDisposableBase.DisposeAsync()
   at Autofac.Core.Disposer.DisposeAsync(Boolean disposing)
   at Autofac.Core.Lifetime.LifetimeScope.DisposeAsync(Boolean disposing)
   at Autofac.Core.Container.DisposeAsync(Boolean disposing)
   at Autofac.Extensions.DependencyInjection.AutofacServiceProvider.DisposeAsync()
   at Microsoft.AspNetCore.Hosting.WebHost.DisposeServiceProviderAsync(IServiceProvider serviceProvider)
   at Microsoft.AspNetCore.Hosting.WebHost.DisposeAsync()
   at Microsoft.AspNetCore.Hosting.WebHostExtensions.RunAsync(IWebHost host, CancellationToken token, String startupMessage)
   at Microsoft.AspNetCore.Hosting.WebHostExtensions.RunAsync(IWebHost host, CancellationToken token)
   at Microsoft.AspNetCore.Hosting.WebHostExtensions.Run(IWebHost host)
   at SIL.XForge.Scripture.Program.Main(String[] args) in D:\Git Projects\web-xforge\src\SIL.XForge.Scripture\Program.cs:line 14
```
Is that a bug or something at my end?

> Reviewed 1 of 2 files at r15, 1 of 1 files at r16, 4 of 4 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-:-NLYt0oA5dCb707Rf_P_:bm79m53)* status: all files reviewed, 4 unresolved discussions (waiting on @github-code-scanning[bot])

pmachapman (commented):
From what I can see, that crash is from the in-process Machine, not the new Machine API or any of the SF code.

This bug has just appeared on bugsnag from QA, and I notice that the API endpoint for that bug report being called is not my new endpoint (`/machine-api/v2/`), but the old one (`/machine-api/`).

I am not sure what I can do to fix it, as it is in the disposal code of SIL.Machine? I think catching in Program.cs would be overkill, as this occurs only on process exit, and the goal is to remove the in-process Machine instance later this year.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-:-NLqd-gY4R1oFIRTVPnC:b-5820b0)* status: all files reviewed, 4 unresolved discussions (waiting on @github-code-scanning[bot])

nigel-wells (approved):
Ah ok. If it is already reported on QA then it isn't related to this branch. I'm happy to approve this one then :)
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1626#-:-NLrOHfFESKjL0EwEnqV:bcrnvlg)* status: all files reviewed, 4 unresolved discussions (waiting on @github-code-scanning[bot])

PR #1634, opened 2022-12-06
SF-1831 Specify min height on ql editor

RaymondLuong3:
Before
Select Text Dialog Offline Before

After
Select Text Dialog Offline After

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1634)


Nateowami (approved):
:lgtm:

PR #1635, opened 2022-12-07
Separate SF and PT note thread rights

RaymondLuong3:
This change separates the rights for PT and SF note threads. Note threads continue to use the same service in the realtime server, but permission to view note threads is based on the rights of the user to view pt notes and sf notes.
If desired, another step would be to further separate SF and PT notes into their own collections in mongo, but I do not see the need or benefit of that.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1635)

### DIFF
--- a/src/RealtimeServer/common/services/json-doc-service.ts
+++ b/src/RealtimeServer/common/services/json-doc-service.ts
@@ -1,5 +1,6 @@
 import ShareDB from 'sharedb';
 import { ObjProxyArg } from 'ts-object-path';
+import { OwnedData } from '../models/owned-data';
 import { obj, ObjPathTemplate } from '../utils/obj-path';
 import { DocService } from './doc-service';
 
@@ -16,20 +17,20 @@ export abstract class JsonDocService<T> extends DocService<T> {
     return obj<T>().pathTemplate(field, inherit);
   }
 
-  protected checkImmutableProps(ops: ShareDB.Op[] | ShareDB.Op): boolean {
+  protected checkImmutableProps(ops: ShareDB.Op[] | ShareDB.Op, entity?: OwnedData): boolean {
     if (ops instanceof Array) {
       for (const op of ops) {
-        if (this.getMatchingPathTemplate(this.immutableProps, op.p) !== -1) {
+        if (this.getMatchingPathTemplate(this.immutableProps, op.p, entity) !== -1) {
           return false;
         }
       }
       return true;
     }
 
-    return this.getMatchingPathTemplate(this.immutableProps, ops.p) === -1;
+    return this.getMatchingPathTemplate(this.immutableProps, ops.p, entity) === -1;
   }
 
-  protected getMatchingPathTemplate(pathTemplates: ObjPathTemplate[], path: ShareDB.Path): number {
+  protected getMatchingPathTemplate(pathTemplates: ObjPathTemplate[], path: ShareDB.Path, _entity?: OwnedData): number {
     for (let i = 0; i < pathTemplates.length; i++) {
       if (pathTemplates[i].matches(path)) {
         return i;
--- a/src/RealtimeServer/common/services/project-data-service.ts
+++ b/src/RealtimeServer/common/services/project-data-service.ts
@@ -81,7 +81,7 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
       throw new Error('The doc service has not been initialized.');
     }
     const project = await this.server.getProject(doc.projectRef);
-    const domain = this.getUpdatedDomain([]);
+    const domain = this.getUpdatedDomain([], doc);
     return project != null && domain != null && this.hasRight(project, domain, Operation.Create, session.userId, doc);
   }
 
@@ -94,7 +94,7 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
       throw new Error('The doc service has not been initialized.');
     }
     const project = await this.server.getProject(doc.projectRef);
-    const domain = this.getUpdatedDomain([]);
+    const domain = this.getUpdatedDomain([], doc);
     return project != null && domain != null && this.hasRight(project, domain, Operation.Delete, session.userId, doc);
   }
 
@@ -111,7 +111,7 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
       return false;
     }
 
-    for (const domain of this.domains) {
+    for (const domain of this.getApplicableDomains(doc)) {
       if (!this.hasRight(project, domain, Operation.View, session.userId, doc)) {
         return false;
       }
@@ -122,7 +122,7 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
   protected async allowUpdate(
     _docId: string,
     oldDoc: T,
-    _newDoc: T,
+    newDoc: T,
     ops: ShareDB.Op[],
     session: ConnectSession
   ): Promise<boolean> {
@@ -139,7 +139,7 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
     }
 
     for (const op of ops) {
-      const domain = this.getUpdatedDomain(op.p);
+      const domain: ProjectDomainConfig | undefined = this.getUpdatedDomain(op.p, newDoc);
       if (domain == null) {
         return false;
       }
@@ -245,15 +245,25 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
     return Promise.resolve();
   }
 
-  private getUpdatedDomain(path: ShareDB.Path): ProjectDomainConfig | undefined {
-    const index = this.getMatchingPathTemplate(
-      this.domains.map(dc => dc.pathTemplate),
-      path
+  /**
+   * Gets the applicable domains based on the properties in the entity.
+   * @param _entity A noteThread or note.
+   * @returns
+   */
+  protected getApplicableDomains(_entity?: OwnedData): ProjectDomainConfig[] {
+    return this.domains;
+  }
+
+  private getUpdatedDomain(path: ShareDB.Path, entity: OwnedData): ProjectDomainConfig | undefined {
+    const domainConfigs: ProjectDomainConfig[] = this.getApplicableDomains(entity);
+    const index: number = this.getMatchingPathTemplate(
+      domainConfigs.map(dc => dc.pathTemplate),
+      path,
+      entity
     );
     if (index !== -1) {
-      return this.domains[index];
+      return domainConfigs[index];
     }
-
     return undefined;
   }
 
@@ -281,7 +291,7 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
   private async handleApply(context: ShareDB.middleware.SubmitContext): Promise<void> {
     const connectSession = context.agent.connectSession as ConnectSession;
     if (context.op.del != null) {
-      const domain = this.getUpdatedDomain([]);
+      const domain = this.getUpdatedDomain([], context.snapshot!.data);
       if (domain != null) {
         await this.onBeforeDelete(connectSession.userId, context.id, domain.projectDomain, context.snapshot!.data);
       }
@@ -291,18 +301,18 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
   private async handleAfterSubmit(context: ShareDB.middleware.SubmitContext): Promise<void> {
     const connectSession = context.agent.connectSession as ConnectSession;
     if (context.op.create != null) {
-      const domain = this.getUpdatedDomain([]);
+      const domain = this.getUpdatedDomain([], context.op.create.data);
       if (domain != null) {
         await this.onInsert(connectSession.userId, context.id, domain.projectDomain, context.op.create.data);
       }
     } else if (context.op.del != null) {
-      const domain = this.getUpdatedDomain([]);
+      const domain = this.getUpdatedDomain([], context.snapshot!.data);
       if (domain != null) {
         await this.onDelete(connectSession.userId, context.id, domain.projectDomain, context.snapshot!.data);
       }
     } else if (context.op.op != null) {
       for (const op of context.op.op) {
-        const domain = this.getUpdatedDomain(op.p);
+        const domain = this.getUpdatedDomain(op.p, context.snapshot!.data);
         if (domain == null) {
           return;
         }
--- a/src/RealtimeServer/scriptureforge/models/note-thread.ts
+++ b/src/RealtimeServer/scriptureforge/models/note-thread.ts
@@ -5,7 +5,6 @@ import { VerseRefData } from './verse-ref-data';
 
 export const NOTE_THREAD_COLLECTION = 'note_threads';
 export const NOTE_THREAD_INDEX_PATHS = PROJECT_DATA_INDEX_PATHS;
-export const SF_NOTE_THREAD_PREFIX = 'SF_NOTE_THREAD_';
 
 /**
  * Note status, mimicking PT CommentList.cs.
@@ -64,5 +63,6 @@ export interface NoteThread extends ProjectData {
   position: TextAnchor;
   status: NoteStatus;
   tagIcon: string;
+  publishedToSF?: boolean;
   assignment?: string;
 }
--- a/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts
+++ b/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts
@@ -10,7 +10,8 @@ export enum SFProjectDomain {
   AnswerComments = 'answer_comments',
   AnswerStatus = 'answer_status',
   Likes = 'likes',
-  NoteThreads = 'note_threads',
+  PTNoteThreads = 'pt_note_threads',
+  SFNoteThreads = 'sf_note_threads',
   Notes = 'notes'
 }
 
@@ -34,14 +35,16 @@ export class SFProjectRights extends ProjectRights {
 
       { projectDomain: SFProjectDomain.Likes, operation: Operation.View },
 
-      { projectDomain: SFProjectDomain.NoteThreads, operation: Operation.View },
+      { projectDomain: SFProjectDomain.SFNoteThreads, operation: Operation.View },
 
       { projectDomain: SFProjectDomain.Notes, operation: Operation.View }
     ];
     this.addRights(SFProjectRole.Observer, observerRights);
 
     const ptObserverRights: ProjectRight[] = observerRights.concat([
-      { projectDomain: SFProjectDomain.Project, operation: Operation.View }
+      { projectDomain: SFProjectDomain.Project, operation: Operation.View },
+
+      { projectDomain: SFProjectDomain.PTNoteThreads, operation: Operation.View }
     ]);
     this.addRights(SFProjectRole.ParatextObserver, ptObserverRights);
 
@@ -57,9 +60,9 @@ export class SFProjectRights extends ProjectRights {
       { projectDomain: SFProjectDomain.Likes, operation: Operation.Create },
       { projectDomain: SFProjectDomain.Likes, operation: Operation.DeleteOwn },
 
-      { projectDomain: SFProjectDomain.NoteThreads, operation: Operation.Create },
-      { projectDomain: SFProjectDomain.NoteThreads, operation: Operation.Edit },
-      { projectDomain: SFProjectDomain.NoteThreads, operation: Operation.Delete },
+      { projectDomain: SFProjectDomain.SFNoteThreads, operation: Operation.Create },
+      { projectDomain: SFProjectDomain.SFNoteThreads, operation: Operation.Edit },
+      { projectDomain: SFProjectDomain.SFNoteThreads, operation: Operation.Delete },
 
       { projectDomain: SFProjectDomain.Notes, operation: Operation.Create },
       { projectDomain: SFProjectDomain.Notes, operation: Operation.EditOwn },
@@ -69,7 +72,11 @@ export class SFProjectRights extends ProjectRights {
     this.addRights(SFProjectRole.CommunityChecker, reviewerRights);
 
     const ptReviewerRights: ProjectRight[] = reviewerRights.concat([
-      { projectDomain: SFProjectDomain.Project, operation: Operation.View }
+      { projectDomain: SFProjectDomain.Project, operation: Operation.View },
+
+      { projectDomain: SFProjectDomain.PTNoteThreads, operation: Operation.Create },
+      { projectDomain: SFProjectDomain.PTNoteThreads, operation: Operation.Edit },
+      { projectDomain: SFProjectDomain.PTNoteThreads, operation: Operation.Delete }
     ]);
     this.addRights(SFProjectRole.ParatextConsultant, ptReviewerRights);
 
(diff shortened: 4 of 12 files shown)


Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1635)* status: all files reviewed, 3 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/realtime-server.ts` line 38 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NIikGlt4vA-UaPZVBLZ:-NIikGlt4vA-UaPZVBL_:b-uh6jtr) ([raw file](https://github.com/sillsdev/web-xforge/blob/180956bfae9e131345f461c31f1652fbf508d85c/src/RealtimeServer/scriptureforge/realtime-server.ts#L38)):*
> ```XML
>         }
>         this.getProject(context.query.projectRef).then(p => {
>           if (
> ```

I think this would be easier to read as:
``` ts
const userId = context.agent.connectSession.userId;
if (p != null && !SF_PROJECT_RIGHTS.hasRight(p, userId, SFProjectDomain.PTNoteThreads, Operation.View)) {
  context.query = { ...context.query, ...{ dataId: { $regex: SF_NOTE_THREAD_PREFIX } } };
}
```
With the user ID defined in a variable, the condition doesn't wrap, and the indention gets less wild.
___
*[`src/RealtimeServer/scriptureforge/models/sf-project-rights.ts` line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NIihClx4VK8kAcCThpS:-NIihClx4VK8kAcCThpT:b-i1a5qu) ([raw file](https://github.com/sillsdev/web-xforge/blob/180956bfae9e131345f461c31f1652fbf508d85c/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts#L78)):*
> ```XML
> 
>       { projectDomain: SFProjectDomain.PTNoteThreads, operation: Operation.Create },
>       { projectDomain: SFProjectDomain.PTNoteThreads, operation: Operation.Edit },
> ```

I think a PTReviewer can only view text and notes, and create, edit, and delete own notes. I'm not certain, but I think we should find out. Here's a screenshot from PT that leads me to think this:


(I could be wrong, so please don't just take my opinion and assume it's correct).
___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.ts` line 77 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NIipsmGAhot9X0kOh16:-NIipsmGAhot9X0kOh17:b-ms1lzi) ([raw file](https://github.com/sillsdev/web-xforge/blob/180956bfae9e131345f461c31f1652fbf508d85c/src/RealtimeServer/scriptureforge/services/note-thread-service.ts#L77)):*
> ```XML
>     const project: Project | undefined = await this.server.getProject(doc.projectRef);
>     if (project?.userRoles[session.userId] == null) return false;
>     if (
> ```

I think this conditional could be made a lot more readable:
``` ts
const userId = session.userId; // shorter variable name prevents wrapping on later lines
const canReadPTNotes = RIGHTS.hasRight(project, userId, SFProjectDomain.PTNoteThreads, Operation.View);
const canReadSFNotes = RIGHTS.hasRight(project, userId, SFProjectDomain.SFNoteThreads, Operation.View);
const isSFNote = doc.dataId.startsWith(SF_NOTE_THREAD_PREFIX);
// conditional is much easier to read because the variables tell you what each value represents
return canReadPTNotes || (canReadSFNotes && isSFNote);
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1635)* status: 2 of 10 files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[`src/RealtimeServer/scriptureforge/realtime-server.ts` line 38 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NIikGlt4vA-UaPZVBLZ:-NIizTEQ6SNcqinZ7Xez:b2mh6na) ([raw file](https://github.com/sillsdev/web-xforge/blob/180956bfae9e131345f461c31f1652fbf508d85c/src/RealtimeServer/scriptureforge/realtime-server.ts#L38)):*

I think this would be easier to read as:
``` ts
const userId = context.agent.connectSession.userId;
if (p != null && !SF_PROJECT_RIGHTS.hasRight(p, userId, SFProjectDomain.PTNoteThreads, Operation.View)) {
  context.query = { ...context.query, ...{ dataId: { $regex: SF_NOTE_THREAD_PREFIX } } };
}
```
With the user ID defined in a variable, the condition doesn't wrap, and the indention gets less wild.
</blockquote></details>

Good thinking. I've made some changes to the formatting.
___
*[`src/RealtimeServer/scriptureforge/models/sf-project-rights.ts` line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NIihClx4VK8kAcCThpS:-NIizaLF19yiPR8IzTMb:b46aslp) ([raw file](https://github.com/sillsdev/web-xforge/blob/180956bfae9e131345f461c31f1652fbf508d85c/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts#L78)):*

I think a PTReviewer can only view text and notes, and create, edit, and delete own notes. I'm not certain, but I think we should find out. Here's a screenshot from PT that leads me to think this:


(I could be wrong, so please don't just take my opinion and assume it's correct).
</blockquote></details>

These rights are on the note thread, so a reviewer should be able to create and edit note thread, even if the thread was not created by the reviewer. I did remove the delete right, I'm not sure we will delete note threads when they get resolved, instead we will likely just mark them as resolved. So the delete right may not be important for admins either. But I left it in for them.
___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.ts` line 77 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NIipsmGAhot9X0kOh16:-NIj-FHm3TVs_2Rg2GMZ:beej048) ([raw file](https://github.com/sillsdev/web-xforge/blob/180956bfae9e131345f461c31f1652fbf508d85c/src/RealtimeServer/scriptureforge/services/note-thread-service.ts#L77)):*

I think this conditional could be made a lot more readable:
``` ts
const userId = session.userId; // shorter variable name prevents wrapping on later lines
const canReadPTNotes = RIGHTS.hasRight(project, userId, SFProjectDomain.PTNoteThreads, Operation.View);
const canReadSFNotes = RIGHTS.hasRight(project, userId, SFProjectDomain.SFNoteThreads, Operation.View);
const isSFNote = doc.dataId.startsWith(SF_NOTE_THREAD_PREFIX);
// conditional is much easier to read because the variables tell you what each value represents
return canReadPTNotes || (canReadSFNotes && isSFNote);
```
</blockquote></details>

Nice suggestion. done.

Nateowami (changes requested):
Nice work. Just a couple comments.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-:-NIkFgT54x2u6MpqP52F:b-2m5sxg)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/realtime-server.ts` line 44 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NIkED7KCwDhzbcvJSb6:-NIkED7KCwDhzbcvJSb7:b-9drsmw) ([raw file](https://github.com/sillsdev/web-xforge/blob/b92e22ddc4513d0a3eb107977a21c376718345ec/src/RealtimeServer/scriptureforge/realtime-server.ts#L44)):*
> ```XML
>             SF_PROJECT_RIGHTS.hasRight(p, userId, SFProjectDomain.SFNoteThreads, Operation.View)
>           ) {
>             context.query = { ...context.query, ...{ publishedToSF: true } };
> ```

I think this can be simplified to:
``` ts
context.query = { ...context.query, publishedToSF: true }
```
___
*[`src/RealtimeServer/scriptureforge/models/sf-project-rights.ts` line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NIihClx4VK8kAcCThpS:-NIkEbAp-kyt_QMehnS6:baq367d) ([raw file](https://github.com/sillsdev/web-xforge/blob/180956bfae9e131345f461c31f1652fbf508d85c/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts#L78)):*

These rights are on the note thread, so a reviewer should be able to create and edit note thread, even if the thread was not created by the reviewer. I did remove the delete right, I'm not sure we will delete note threads when they get resolved, instead we will likely just mark them as resolved. So the delete right may not be important for admins either. But I left it in for them.
</blockquote></details>

I think the issue here is that the domains aren't fully set up. If you look at the permissions for the questions, you'll see that only admins have the `questions.edit` permission, despite the fact that community checkers can add answers (which edits a question).

Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1635)* status: all files reviewed, 2 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/models/sf-project-rights.ts` line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NIihClx4VK8kAcCThpS:-NIkG5_U7Cnyi2b0Bnrp:b-ew6vnd) ([raw file](https://github.com/sillsdev/web-xforge/blob/180956bfae9e131345f461c31f1652fbf508d85c/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts#L78)):*

I think the issue here is that the domains aren't fully set up. If you look at the permissions for the questions, you'll see that only admins have the `questions.edit` permission, despite the fact that community checkers can add answers (which edits a question).
</blockquote></details>

Just to elaborate on what I was saying a bit:
When a reviewer adds a note to a note thread, the note thread object is being edited. So if you just look at it from the perspective of "which models can the user modify", then yes, the user is editing the note thread. But our permissions system is more granular than the model level and specifies which parts of an object can be edited by a given user. I'm not very familiar with how it works, but @nigel-wells has done some work related to this recently and may be able to provide insight.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1635)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/models/sf-project-rights.ts` line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NIihClx4VK8kAcCThpS:-NJ25Gwm0H4a5aWGncue:bdp3fcd) ([raw file](https://github.com/sillsdev/web-xforge/blob/180956bfae9e131345f461c31f1652fbf508d85c/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts#L78)):*

Just to elaborate on what I was saying a bit:
When a reviewer adds a note to a note thread, the note thread object is being edited. So if you just look at it from the perspective of "which models can the user modify", then yes, the user is editing the note thread. But our permissions system is more granular than the model level and specifies which parts of an object can be edited by a given user. I'm not very familiar with how it works, but @nigel-wells has done some work related to this recently and may be able to provide insight.
</blockquote></details>

Each collection can have a corresponding service which is used to define an array of object path templates to help the permissions system understand the different parts of the collection. The project domain is setup here to define which permissions are relevant to the collection. A request is then made to read, update, delete, etc. and the rights are checked against what has been defined in the object path templates.

All of this can be overwritten (and are in some cases) per collection as these are just protected methods.

Looking at `setupDomains()` in `note-thread-services.ts` it looks like a permission is needed for `PTNoteThreads` to create the actual note thread and then anyone with access to `Notes` can do actions on the `note_threads.notes` part of the object. This means the `SFNoteThreads` are redundant in the sense of enforcing permissions in the realtime server but can obviously be checked from the frontend. If you created an additional `SFProjectDomain.SFNotes` then you could use that to enforce the permissions in the realtime server. The existing `SFProjectDomain.Notes` can be used for PT permissions. This is essentially giving two sets of permissions to do exactly the same thing. You can overwrite the `allowRead` (which it looks like you've done) to check the user wanting to access/change the object has the business logic permission to that specific object which I think you did by changing the ID of the note thread? The middleware you wrote the other day can be updated to return the relevant note threads based on permission as well.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1635)* status: 7 of 12 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/RealtimeServer/scriptureforge/realtime-server.ts` line 44 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NIkED7KCwDhzbcvJSb6:-NKy-iNuDKXbeIe2iMaJ:bo5nztk) ([raw file](https://github.com/sillsdev/web-xforge/blob/b92e22ddc4513d0a3eb107977a21c376718345ec/src/RealtimeServer/scriptureforge/realtime-server.ts#L44)):*

I think this can be simplified to:
``` ts
context.query = { ...context.query, publishedToSF: true }
```
</blockquote></details>

Nice. Done.
___
*[`src/RealtimeServer/scriptureforge/models/sf-project-rights.ts` line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NIihClx4VK8kAcCThpS:-NKxylCe55KaRDGfGz3I:bsntdih) ([raw file](https://github.com/sillsdev/web-xforge/blob/180956bfae9e131345f461c31f1652fbf508d85c/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts#L78)):*

Each collection can have a corresponding service which is used to define an array of object path templates to help the permissions system understand the different parts of the collection. The project domain is setup here to define which permissions are relevant to the collection. A request is then made to read, update, delete, etc. and the rights are checked against what has been defined in the object path templates.

All of this can be overwritten (and are in some cases) per collection as these are just protected methods.

Looking at `setupDomains()` in `note-thread-services.ts` it looks like a permission is needed for `PTNoteThreads` to create the actual note thread and then anyone with access to `Notes` can do actions on the `note_threads.notes` part of the object. This means the `SFNoteThreads` are redundant in the sense of enforcing permissions in the realtime server but can obviously be checked from the frontend. If you created an additional `SFProjectDomain.SFNotes` then you could use that to enforce the permissions in the realtime server. The existing `SFProjectDomain.Notes` can be used for PT permissions. This is essentially giving two sets of permissions to do exactly the same thing. You can overwrite the `allowRead` (which it looks like you've done) to check the user wanting to access/change the object has the business logic permission to that specific object which I think you did by changing the ID of the note thread? The middleware you wrote the other day can be updated to return the relevant note threads based on permission as well.
</blockquote></details>

Thanks for the insight! I was happily surprised to learn that we, in fact, do not need to be concerned about SFNotes and PTNotes. When we specify that a note thread is a PT note thread, a reviewer will not be able to read or insert notes to that note thread because ShareDBAccess requires that the user has read permission on the PT note thread. That's a nice trick :)

I have added some logic to find the applicable domains for a NoteThreadDoc. For SF note threads, the `SFProjectDomain.PTNoteThread` will be removed from the applicable domains and `SFProjectDomain.SFNoteThread` will be used as the domain for the thread. This has the advantage of not needing to overwrite any of the access methods like `allowRead`.  This should be tested since it does affect other services that implement the ProjectDataService class.

Let me know what you think.

nigel-wells (changes requested):
Looking pretty good now - feels like we're landing in the right spot with all of this which is great. Just a few comments below for clarification.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-:-NLO9lx8-RIAzrYHEHR_:b5w1ho4)* status: all files reviewed, 6 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/models/note-thread.ts` line 66 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NLO93IqAktq29xmIxoU:-NLO93IqAktq29xmIxoV:b-6zflx4) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba78f556afc60138cb3455eeac95e65e3e006392/src/RealtimeServer/scriptureforge/models/note-thread.ts#L66)):*
> ```XML
>   status: NoteStatus;
>   tagIcon: string;
>   publishedToSF?: boolean;
> ```

I agree that it will make it easier to run checks against a specific property. I'm not 100% on calling it "published". I would have thought "create" but I see you had then and changed your mind? Published to me says this note thread has been published on SF and others (PT) aren't which seems confusing. "Created On" seems right - what made you change it?
___
*[`src/RealtimeServer/scriptureforge/models/sf-project-rights.ts` line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NIihClx4VK8kAcCThpS:-NLO9bISEg6tTbgJO0G3:b-1lmv8t) ([raw file](https://github.com/sillsdev/web-xforge/blob/180956bfae9e131345f461c31f1652fbf508d85c/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts#L78)):*

Thanks for the insight! I was happily surprised to learn that we, in fact, do not need to be concerned about SFNotes and PTNotes. When we specify that a note thread is a PT note thread, a reviewer will not be able to read or insert notes to that note thread because ShareDBAccess requires that the user has read permission on the PT note thread. That's a nice trick :)

I have added some logic to find the applicable domains for a NoteThreadDoc. For SF note threads, the `SFProjectDomain.PTNoteThread` will be removed from the applicable domains and `SFProjectDomain.SFNoteThread` will be used as the domain for the thread. This has the advantage of not needing to overwrite any of the access methods like `allowRead`.  This should be tested since it does affect other services that implement the ProjectDataService class.

Let me know what you think.
</blockquote></details>

That's a nice way to do it, make the entity data available so you can check it and make a decision off that - I like it.
___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts` line 135 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NLO5bMP7CrZZIrPRIzZ:-NLO5bMP7CrZZIrPRIz_:bygxixo) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba78f556afc60138cb3455eeac95e65e3e006392/src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts#L135)):*
> ```XML
>     const env = new TestEnvironment();
>     await env.createData();
>     const conn: Connection = clientConnect(env.server, 'reviewer');
> ```

I was confused initially around "reviewer". When I first started reviewing I thought this was related to the reviewer role and I wasn't sure which role it was. After reviewing I see "reviewer", in this instance, is just the user ID. If roles aren't relevant here then perhaps we should call this a different name? I can understand why you'd want to call it a reviewer as it links back to how we happen to have implemented that in the project rights.



_Code quote:_
```XML
  it('allows a reviewer to create a note thread', async () => {
    const env = new TestEnvironment();
    await env.createData();
    const conn: Connection = clientConnect(env.server, 'reviewer');
```
___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts` line 181 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NLO58ql1bgZ_6Eswi39:-NLO58ql1bgZ_6Eswi3A:bcqdqqz) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba78f556afc60138cb3455eeac95e65e3e006392/src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts#L181)):*
> ```XML
>   });
> 
>   it('allows reviewer to update own note', async () => {
> ```

state `update and delete` own note
___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts` line 533 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NLO32h3B9JfeYep681O:-NLO32h46Wba26ednSNC:bn25das) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba78f556afc60138cb3455eeac95e65e3e006392/src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts#L533)):*
> ```XML
>       dateModified: '',
>       ownerRef,
>       content: 'this note should not be inserted',
> ```

As this is a generic method I think `content` should be more generic as well otherwise it may confuse future tests if it is used when the content should be inserted.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1635)* status: all files reviewed, 6 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/RealtimeServer/scriptureforge/models/note-thread.ts` line 66 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NLO93IqAktq29xmIxoU:-NLSr40gCik6N9HmdVRl:b-caj048) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba78f556afc60138cb3455eeac95e65e3e006392/src/RealtimeServer/scriptureforge/models/note-thread.ts#L66)):*

I agree that it will make it easier to run checks against a specific property. I'm not 100% on calling it "published". I would have thought "create" but I see you had then and changed your mind? Published to me says this note thread has been published on SF and others (PT) aren't which seems confusing. "Created On" seems right - what made you change it?
</blockquote></details>

This was something that Nathaniel and I had talked about. You are right that currently this describes notes that are created in SF, but we thought in the future there may be a reason to publish translator notes created in PT so they are visible to reviewer users. That's why we named it this way.
___
*[`src/RealtimeServer/scriptureforge/models/sf-project-rights.ts` line 78 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NIihClx4VK8kAcCThpS:-NLSrTMS81RWtFiKsYg7:b-qczc61) ([raw file](https://github.com/sillsdev/web-xforge/blob/180956bfae9e131345f461c31f1652fbf508d85c/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts#L78)):*

That's a nice way to do it, make the entity data available so you can check it and make a decision off that - I like it.
</blockquote></details>

Thanks. It saves us a lot of code when we do not have to overwrite the allowRead and the other permission checks.
___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts` line 135 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NLO5bMP7CrZZIrPRIzZ:-NLSu-ms-3LOhpQHdTd_:b-qobao9) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba78f556afc60138cb3455eeac95e65e3e006392/src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts#L135)):*

I was confused initially around "reviewer". When I first started reviewing I thought this was related to the reviewer role and I wasn't sure which role it was. After reviewing I see "reviewer", in this instance, is just the user ID. If roles aren't relevant here then perhaps we should call this a different name? I can understand why you'd want to call it a reviewer as it links back to how we happen to have implemented that in the project rights.
</blockquote></details>

I was thinking the same way that you were, that the `reviewer` describes the userId only. But the project is defined with 3 users with the IDs `[projectAdmin, checker, reviewer]` and that does come into play. I'll leave a comment so that it is clear. I'll also replace the instances of the userIds with a variable.
___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts` line 181 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NLO58ql1bgZ_6Eswi39:-NLSsyO0Cdq9fNtHH8Gk:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba78f556afc60138cb3455eeac95e65e3e006392/src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts#L181)):*

state `update and delete` own note
</blockquote></details>

Done.
___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts` line 533 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NLO32h3B9JfeYep681O:-NLSsPud8_DMwC_WvC5r:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba78f556afc60138cb3455eeac95e65e3e006392/src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts#L533)):*

As this is a generic method I think `content` should be more generic as well otherwise it may confuse future tests if it is used when the content should be inserted.
</blockquote></details>

Done.

nigel-wells (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-:-NLTB_J01wte-6HXEol9:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/models/note-thread.ts` line 66 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NLO93IqAktq29xmIxoU:-NLTBS9v6VSILwoazK6u:b-blwvn1) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba78f556afc60138cb3455eeac95e65e3e006392/src/RealtimeServer/scriptureforge/models/note-thread.ts#L66)):*

This was something that Nathaniel and I had talked about. You are right that currently this describes notes that are created in SF, but we thought in the future there may be a reason to publish translator notes created in PT so they are visible to reviewer users. That's why we named it this way.
</blockquote></details>

Right ok - good to know.
___
*[`src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts` line 135 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1635#-NLO5bMP7CrZZIrPRIzZ:-NLT9v8tBZob88QTe82p:b-zcjvte) ([raw file](https://github.com/sillsdev/web-xforge/blob/ba78f556afc60138cb3455eeac95e65e3e006392/src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts#L135)):*

I was thinking the same way that you were, that the `reviewer` describes the userId only. But the project is defined with 3 users with the IDs `[projectAdmin, checker, reviewer]` and that does come into play. I'll leave a comment so that it is clear. I'll also replace the instances of the userIds with a variable.
</blockquote></details>

Thanks

PR #1637, opened 2022-12-08
Fix Angular warnings regarding CommonJS and AMD modules

Nateowami:
Angular prints warnings like this when building:
```
Warning: /home/nate/src/web-xforge/src/SIL.XForge.Scripture/ClientApp/node_modules/@sillsdev/machine/lib/tokenization/latin-word-tokenizer.js depends on 'xregexp'. CommonJS or AMD dependencies can cause optimization bailouts.
For more info see: https://angular.io/guide/build#configuring-commonjs-dependencies
```
CommonJS modules are defined in a such a way that prevent effective tree-shaking.

These warnings are "fixed" only in the sense that they are suppressed, by adding these packages to a list to ignore. As far as I am aware, there really isn't much we can do about these modules except stop using these packages (or hope they eventually change how they package output). 

I've deliberately not added xregexp to the list here, so that we will still be warned about it. It's significantly larger than most of the other packages, and we're in the process of removing it entirely.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1637)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/angular.json
+++ b/src/SIL.XForge.Scripture/ClientApp/angular.json
@@ -41,7 +41,22 @@
             "stylePreprocessorOptions": {
               "includePaths": ["node_modules"]
             },
-            "allowedCommonJsDependencies": ["@bugsnag/js"],
+            "allowedCommonJsDependencies": [
+              "@bugsnag/browser",
+              "@bugsnag/js",
+              "arraydiff",
+              "bowser",
+              "bson-objectid",
+              "crc-32",
+              "flat",
+              "gensequence",
+              "mingo",
+              "papaparse",
+              "quill",
+              "recordrtc",
+              "tinycolor2",
+              "ts-md5"
+            ],
             "scripts": [],
             "aot": false,
             "vendorChunk": true,


RaymondLuong3 (approved):
Looks good. I see two more warming that are not related to the commonJS modules. They are both related to a unused import progress-indicator. I did a search on the repo, and I don't see us importing it directly. Just FYI.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1637#-:-NJ6OC0L4WF6Al22wFgt:b-9bcbys)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami (commented):
@raymondluong3 Yes, I noticed those other warnings as well. Since they're imports from Material, and we aren't importing them anywhere, one guess I have is that Angular MDC may import them (Angular MDC imports from Material in a number of places), and we import something from MDC that imports them, and that's how they're getting included. But that's just a guess.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1637#-:-NJ6SENBDy-9ae2uymuR:b5jmck)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1639, opened 2022-12-08
SF-1618 SF-1619 Make question import more flexible

Nateowami:
Changes to how questions are imported
- Allow omitting book in references if it is specified in the file name
- Allow extraneous columns (data is now found by looking at headings)

Instructions before | Instructions after
--------------------------|----------------------
 | 

Note that this is not backwards compatible; whereas previously files were not supposed to have headings, now they require them.

There are two main goals with this project:
1. It should be possible to import files from unfoldingWord with no changes to them.
2. Any "reasonable" spreadsheet should be able to be easily modified to produce a format that can be imported. For example, before this change the reference and question column had to be the first and second columns respectively. While that's not a difficult change to make for someone familiar with working with spreadsheets, it's a lot more straightforward to give a column a heading. It's also easier to communicate to users.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1639)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html
@@ -53,7 +53,11 @@
                 <li>
                   {{ csvInstructions[0] }}
                   <table>
-                    <!-- TODO internationalize the verse references, but only once we accept localized inputs -->
+                    <!-- TODO internationalize the references and headings, but only once we accept localized inputs -->
+                    <tr>
+                      <th>Reference</th>
+                      <th>Question</th>
+                    </tr>
                     <tr *ngFor="let verseNumber of [1, 2]">
                       <td>1JN 1:{{ verseNumber }}</td>
                       <td>{{ t("question_for_verse", { number: verseNumber }) }}</td>
@@ -159,6 +163,11 @@
 
       <div *ngIf="status === 'update_transcelerator'">{{ t("update_transcelerator") }}</div>
 
+      <div
+        *ngIf="status === 'missing_header_row'"
+        [innerHTML]="i18n.translateAndInsertTags('import_questions_dialog.missing_header_row')"
+      ></div>
+
       <div *ngIf="status === 'progress'">
         <mat-progress-bar mode="determinate" [value]="(importedCount / toImportCount) * 100"></mat-progress-bar>
       </div>
@@ -186,7 +195,7 @@
       {{ t("cancel") }}
     </button>
     <button
-      *ngIf="status === 'no_questions' || status === 'update_transcelerator'"
+      *ngIf="status === 'no_questions' || status === 'update_transcelerator' || status === 'missing_header_row'"
       mat-flat-button
       mat-dialog-close
       color="primary"
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss
@@ -59,7 +59,8 @@
   table {
     font-size: small;
     border-collapse: collapse;
-    td {
+    td,
+    th {
       border: 1px solid lightgray;
       padding: 0 0.5em;
     }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts
@@ -314,7 +314,7 @@ describe('ImportQuestionsDialogComponent', () => {
     const genQuestions = [['Genesis 1:1', 'Question for Genesis 1:1']];
     const matQuestions = Array.from(Array(100), (_, i) => [`MAT 1:${i + 1}`, `Question for Matthew 1:${i + 1}`]);
 
-    env.selectFileWithContents(genQuestions.concat(matQuestions));
+    env.selectFileWithContents([['Reference', 'Questions'], ...genQuestions, ...matQuestions]);
 
     expect(env.tableRows.length).toBe(1);
     expect(env.getColumnTwoText(env.tableRows[0])).toEqual('Genesis 1:1');
@@ -326,7 +326,10 @@ describe('ImportQuestionsDialogComponent', () => {
     const question = { data: { text: 'Matthew 1:1 question', verseRef: { bookNum: 40, chapterNum: 1, verseNum: 1 } } };
     const env = new TestEnvironment({ existingQuestions: [question as QuestionDoc] });
 
-    env.selectFileWithContents([['MAT 1:1', 'Matthew 1:1 question']]);
+    env.selectFileWithContents([
+      ['Reference', 'Questions'],
+      ['MAT 1:1', 'Matthew 1:1 question']
+    ]);
 
     expect(env.tableRows.length).toBe(1);
     expect(env.getColumnTwoText(env.tableRows[0])).toEqual('Matthew 1:1 question');
@@ -340,7 +343,12 @@ describe('ImportQuestionsDialogComponent', () => {
   it('it informs the user about invalid rows in the CSV file and skips them', fakeAsync(() => {
     const env = new TestEnvironment();
 
-    env.selectFileWithContents([['Lorem ipsum'], ['MAT 1:1', ' '], ['MAT 1:2', 'Question for MAT 1:2']]);
+    env.selectFileWithContents([
+      ['Reference', 'Questions'],
+      ['Lorem ipsum'],
+      ['MAT 1:1', ' '],
+      ['MAT 1:2', 'Question for MAT 1:2']
+    ]);
 
     expect(env.headerText).toBe('These rows in the CSV file were invalid and will be skipped.');
     const invalidRows = env.tableRows;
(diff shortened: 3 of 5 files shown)


RaymondLuong3 (changes requested):
Nice work. I saw that if a row was invalid, and the document did not happen to be in the expected format (i.e. columns were out of order), then the dialog doesn't show the error row correctly. 


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1639#-:-NJ6drAr1MoR6iA_95NS:b95lkq5)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html` line 197 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1639#-NJ6SHEz-zmnpRJd43cD:-NJ6SHEz-zmnpRJd43cE:bmxy376) ([raw file](https://github.com/sillsdev/web-xforge/blob/39dd5c2634287f3ce28f4e486ada52c43a16f259/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L197)):*
> ```HTML
>       {{ t("cancel") }}
>     </button>
>     <button
> ```

I wonder if this button is needed at all, since the cancel button would be sufficient. But it's a minor thing.



_Code quote:_
```HTML
    <button
      *ngIf="status === 'no_questions' || status === 'update_transcelerator' || status === 'missing_header_row'"
      mat-flat-button
      mat-dialog-close
      color="primary"
    >
      {{ t("close") }}
    </button>
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts` line 424 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1639#-NJ6YThc39ZPgG6kxcwx:-NJ6YThc39ZPgG6kxcwy:bwqdcxv) ([raw file](https://github.com/sillsdev/web-xforge/blob/39dd5c2634287f3ce28f4e486ada52c43a16f259/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L424)):*
> ```XML
>         // if the reference doesn't start with a book id, and the file name includes the book id, prepend the book id to
>         // the reference
>         const refStartsWithBook = Canon.allBookIds.includes(reference.slice(0, 3)) && reference[3] === ' ';
> ```

Add the types here. I'm not sure we need to check this for every row. I see that if we check each row, we can attempt to find a reference for every row so it works if a user has a file with a mix of reference like `GEN 1:1` and `1:1`. But I wonder if we are trying to be too smart. I can't really think of a better alternative though. 



_Code quote:_
```XML
        const refStartsWithBook = Canon.allBookIds.includes(reference.slice(0, 3)) && reference[3] === ' ';
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts` line 425 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1639#-NJ6ZpvPBR_Vn6QZDw73:-NJ6ZpvPBR_Vn6QZDw74:bzgf19t) ([raw file](https://github.com/sillsdev/web-xforge/blob/39dd5c2634287f3ce28f4e486ada52c43a16f259/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L425)):*
> ```XML
>         // the reference
>         const refStartsWithBook = Canon.allBookIds.includes(reference.slice(0, 3)) && reference[3] === ' ';
>         const fullReference = refStartsWithBook || defaultBookId == null ? reference : defaultBookId + ' ' + reference;
> ```

If `refStartsWithBook` is false and `defaultBookId` is true, then the reference should be invalid, and we should not try to import that row. Isn't that the case?



_Code quote:_
```XML
        const fullReference = refStartsWithBook || defaultBookId == null ? reference : defaultBookId + ' ' + reference;
        questions.push({
          verseRef: VerseRef.parse(fullReference),
          text: questionText
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1639)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts` line 425 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1639#-NJ6ZpvPBR_Vn6QZDw73:-NJ6eKRI0u8EdAa1oN3g:bacu2gb) ([raw file](https://github.com/sillsdev/web-xforge/blob/39dd5c2634287f3ce28f4e486ada52c43a16f259/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L425)):*

If `refStartsWithBook` is false and `defaultBookId` is true, then the reference should be invalid, and we should not try to import that row. Isn't that the case?
</blockquote></details>

Oh, I see now. If there is an error importing the row, then it gets added to the invalid row. Nice.

Nateowami (commented):
Good catch. Fixed.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1639#-:-NJCs55P5OoXCn3w-CaZ:b-k2rwtz)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html` line 197 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1639#-NJ6SHEz-zmnpRJd43cD:-NJCtdDd0KrpST9yGx4x:b-uywgqu) ([raw file](https://github.com/sillsdev/web-xforge/blob/39dd5c2634287f3ce28f4e486ada52c43a16f259/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L197)):*

I wonder if this button is needed at all, since the cancel button would be sufficient. But it's a minor thing.
</blockquote></details>

The cancel button is there for when things are normal and the user wants to cancel. The close button us there for when we tell the user questions cannot be imported and why. Perhaps instead of hiding most elements and showing a message, we should close the dialog and open a message dialog. The message dialog didn't exist when this component was written though.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts` line 424 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1639#-NJ6YThc39ZPgG6kxcwx:-NJCs9oZ42odNcWkE_Bf:b-4tnnvq) ([raw file](https://github.com/sillsdev/web-xforge/blob/39dd5c2634287f3ce28f4e486ada52c43a16f259/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L424)):*

Add the types here. I'm not sure we need to check this for every row. I see that if we check each row, we can attempt to find a reference for every row so it works if a user has a file with a mix of reference like `GEN 1:1` and `1:1`. But I wonder if we are trying to be too smart. I can't really think of a better alternative though. 
</blockquote></details>

Yeah, I thought about that too. A couple things to consider:
- Just because we found a book in the file name doesn't necessarily mean the references are written without a book name.
- It would be unfortunate if having a mistake on line 2 (the first line after the headers) caused every line afterwards to be interpreted differently and seen as invalid.

So by checking every row we can make sure that we don't tell the user 500 lines are invalid when really it's just the first non-header line.

RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1639)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html` line 197 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1639#-NJ6SHEz-zmnpRJd43cD:-NJpXqv40Whu5qZNE66P:bo7vsx6) ([raw file](https://github.com/sillsdev/web-xforge/blob/39dd5c2634287f3ce28f4e486ada52c43a16f259/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.html#L197)):*

The cancel button is there for when things are normal and the user wants to cancel. The close button us there for when we tell the user questions cannot be imported and why. Perhaps instead of hiding most elements and showing a message, we should close the dialog and open a message dialog. The message dialog didn't exist when this component was written though.
</blockquote></details>

Hmm, I'm not a fan of dialogs opening on top of dialogs. Ok, the close button has the purpose to inform the user that they cannot proceed. That seems alright.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts` line 424 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1639#-NJ6YThc39ZPgG6kxcwx:-NJpWxDe0KX7Icktfqq7:bslyhw1) ([raw file](https://github.com/sillsdev/web-xforge/blob/39dd5c2634287f3ce28f4e486ada52c43a16f259/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L424)):*

Yeah, I thought about that too. A couple things to consider:
- Just because we found a book in the file name doesn't necessarily mean the references are written without a book name.
- It would be unfortunate if having a mistake on line 2 (the first line after the headers) caused every line afterwards to be interpreted differently and seen as invalid.

So by checking every row we can make sure that we don't tell the user 500 lines are invalid when really it's just the first non-header line.
</blockquote></details>

Yea, so in this case, being smart is better than trying to force a rigid schema that the user is responsible to follow. I think this will serve us.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts` line 418 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1639#-NJpXXwy4KauENGTK-zb:-NJpXXwy4KauENGTK-zc:b-hegu7x) ([raw file](https://github.com/sillsdev/web-xforge/blob/9066c41a20b40d443455fc44ce377060dcfdbef7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L418)):*
> ```XML
>       const questionText: string | undefined = row[questionColumn]?.trim();
>       if (row.length < 2 || reference === '' || questionText === '') {
>         invalidRows.push([rowNumber].concat(row));
> ```

Would this need to be updated also?



_Code quote:_
```XML
    invalidRows.push([rowNumber].concat(row));
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1639)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts` line 418 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1639#-NJpXXwy4KauENGTK-zb:-NJpZd5w0X5XvB_pAztY:b-h7gokb) ([raw file](https://github.com/sillsdev/web-xforge/blob/9066c41a20b40d443455fc44ce377060dcfdbef7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L418)):*

Would this need to be updated also?
</blockquote></details>

You mean to get a correct error message? I thought I tested this; now I see it looks wrong to me too. I'll test it again and see if I missed something.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1639)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts` line 418 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1639#-NJpXXwy4KauENGTK-zb:-NJq73HI480SOsvTZXeQ:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/9066c41a20b40d443455fc44ce377060dcfdbef7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L418)):*

You mean to get a correct error message? I thought I tested this; now I see it looks wrong to me too. I'll test it again and see if I missed something.
</blockquote></details>

Done.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1639)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1640, opened 2022-12-12
Update .editorconfig

pmachapman:
This update to .editorconfig is to help Visual Studio's suggestions match Scripture Forge's coding style, and to enable use of the new EditorConfig UI without the .editorconfig file being modified substantially. This update:

 * Adds the default rules with default values from Microsoft Visual Studio 2022
 * Updates rules to match Scripture Forge coding style:
   - **Preferred 'using' directive placement:** outside namespace
   - **Prefer implicit object creation when type is apparent:** No
   - **Place 'System' directives first when sorting usings:** Yes

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1640)


marksvc (approved):
I did not carefully examine all of these lines. 
> without the .editorconfig file being modified substantially.

I guess that would have been really large then :)

> Prefer implicit object creation when type is apparent: No

I don't know if this is related, but:
Damien has wanted us to write `var foo = new Baz();` rather than `Baz foo = new Baz();`. Though it's a bit muddier now with nullable types. I've been using `Baz foo = new()` (or maybe sometimes `Baz? foo = new()`) recently. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1640#-:-NJ76ezV3znfuU38TkKn:b-7evu8j)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

marksvc:
Sounds good.

> every time you open the .editorconfig in Visual Studio

Oh okay, that makes sense now :)

PR #1641, opened 2022-12-12
doc: add section Cleaning

marksvc:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1641)

PR #1643, opened 2022-12-12
doc: add section on invalidating refresh tokens

marksvc:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1643)

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ The rest of this document discusses the development of the underlying software.
   - [PWA Testing](#pwa-testing)
   - [Physical Device Testing](#physical-device-testing)
   - [Offline testing](#offline-testing)
+  - [Tokens](#tokens)
 - [Backend Development](#backend-development)
   - [Model Changes](#model-changes)
 - [Debugging](#debugging)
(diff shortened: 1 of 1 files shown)


marksvc:
@Nateowami Hmm, this was composed before our discussion of manual wrapping in the .md file. Are you wanting individual paragraphs to be on the same line, rather than broken apart with newlines as in this PR?

Nateowami:
I'm concerned our README is getting unweildy. Originally it provided basic documentation, but as we start to add lengthy sections on increasingly obscure topics, I think it may be time to create a developer documentation folder and have individual files there.

pmachapman:
> I'm concerned our README is getting unwieldy.

Do we want to keep developer information like this in a [GitHub wiki](https://docs.github.com/en/communities/documenting-your-project-with-wikis/about-wikis)?

Other projects I have forked have used a `/Documentation/` folder, with multiple md files, but I have found that approach to be difficult to find information in, unless that information automatically is published to a static website.

Nateowami:
That is a good point. Given that there are multiple options with pros and cons, I don't want this to hold up this PR (or #1641). Let's get this into the README for now, and decide how we want to handle documentation going forward.

Regardless of what we choose, as long as it's in a standard format like Markdown, I think we can easily change our minds later. I like the GitHub wiki idea.

pmachapman (approved):
Looks good! As per the PR discussion, the README is the best place for this info until there is a wiki or other system.

PR #1644, opened 2022-12-13
Add option to run tests in parallel

Nateowami:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1644)


nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1644)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`.github/workflows/build-and-test.yml` line 43 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1644#-NJ8ElwtAfPcZra0JL2z:-NJ8ElwtAfPcZra0JL3-:bfkh6w5) ([raw file](https://github.com/sillsdev/web-xforge/blob/e0dd9679fefbaa4c4efb1d3ee6f5d16b38136d1d/.github/workflows/build-and-test.yml#L43)):*
> ```YAML
>           set -xueo pipefail
>           lsb_release -a
>           node -e "console.log(os.cpus())"
> ```

Probably not needed?
___
*[`src/SIL.XForge.Scripture/ClientApp/package.json` line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1644#-NJ8Eq4Y1zr6qrsf_MXx:-NJ8Eq4ZBiy5laNFvzlX:bk9ckfa) ([raw file](https://github.com/sillsdev/web-xforge/blob/e0dd9679fefbaa4c4efb1d3ee6f5d16b38136d1d/src/SIL.XForge.Scripture/ClientApp/package.json#L13)):*
> ```JSON
>     "build": "ng build",
>     "test": "ng test --source-map",
>     "test:parallel": "KARMA_PARALLEL=true npm test",
> ```

On Windows this needs to be `set KARMA_PARALLEL=true&& npm test` - does that still work on Linux?

nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1644)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/package.json` line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1644#-NJ8Eq4Y1zr6qrsf_MXx:-NJCJoja6qN7pIPL0KAX:b-6tfver) ([raw file](https://github.com/sillsdev/web-xforge/blob/e0dd9679fefbaa4c4efb1d3ee6f5d16b38136d1d/src/SIL.XForge.Scripture/ClientApp/package.json#L13)):*

No, that doesn't work.

I tried modifying `karma.conf.js` to take command line arguments, but `ng test` won't accept custom arguments. See https://github.com/angular/angular-cli/issues/12305

So then I thought we could smuggle them in in the list of reporters, like so:
``` bash
ng test --reporters=mocha,coverage-istanbul,parallel
```
And then read the arguments from `process.argv` and add reporters, and if "parallel" is in the list, treat it as a config option and not a reporter. But it turns out Karma reads the list of reports specified on the command line, and uses them, and complains "parallel" is not a valid reporter. So that hack didn't work.
</blockquote></details>

What about:
```
export KARMA_PARALLEL=true npm test || set KARMA_PARALLEL=true&& npm test
```
I should note that I'm not overly fussed as I used my IDE to run the tests which calls Karma direct and I can set the environment variables in there.
Alternatively, we could have two different scripts to suit Windows or Linux.

Nateowami:
@pmachapman Do you know what the best way is to craft a command that runs `npm test` with the environment variable `KARMA_PARALLEL=true` on both Linux and Windows? On Linux we would just do `KARMA_PARALLEL=true npm test`.

pmachapman:
> @pmachapman Do you know what the best way is to craft a command that runs `npm test` with the environment variable `KARMA_PARALLEL=true` on both Linux and Windows? On Linux we would just do `KARMA_PARALLEL=true npm test`.

You can run: `set KARMA_PARALLEL=true && npm test`, with the only difference to that it will set that environment variable for the rest of your command line session (i.e. until closing the terminal window).

Nateowami:
@pmachapman That does not actually work in my testing, and I don't think that's a valid way to set an environment variable in Bash (or similar shells):
```
$ type set
set is a shell builtin
$ help set
set: set [-abefhkmnptuvxBCHP] [-o option-name] [--] [arg ...]
    Set or unset values of shell options and positional parameters.
    
    Change the value of shell attributes and positional parameters, or
    display the names and values of shell variables.
```
Also:
```
$ set KARMA_PARALLEL=true && echo $KARMA_PARALLEL

```
(that outputs an empty line).
Whereas we get the expect output if we run this:
```
$ KARMA_PARALLEL=true && echo $KARMA_PARALLEL
true
```

pmachapman:
>  That does not actually work in my testing, and I don't think that's a valid way to set an environment variable in Bash (or similar shells):

Sorry I misunderstood your question. Yes, using `set` only works in Windows for environment variables.

For a cross platform solution, we could modify the test script in package.json to use cross-env? https://www.npmjs.com/package/cross-env

marksvc:
This is an interesting problem. In PR #1966 I ended up suggesting

```json
"test:headless": "echo;export KARMA_REPORTERS=mocha && set KARMA_REPORTERS=mocha && ng test --browsers xForgeChromiumHeadless --source-map",
```

On Linux it will echo a blank line, export KARMA_REPORTERS to the environment, run `set` meaninglessly, and run ng.
On Windows, it will echoes `export KARMA_REPORTERS=mocha` (but at least does not print any error messages), runs set, and runs ng.

I think there must remain some clever solution. For example, I was hopeful of prefixing Linux-only commands with `:;`, but found that `:` ignores the rest of the line on cmd and it can't be ended with even `&&`.

marksvc:
@Nateowami , @nigel-wells : What about this?

BTW, to not have a bunch of test windows open up, you can use 
```bash
KARMA_PARALLEL=true npm run test:headless
```

nigel-wells:
Yeah, that works for me. I found headless is best otherwise I get errors for audio checks when the correct browser tab isn't in focus

Nateowami:
@marksvc If you want to review it, I think we can merge. I just tested it and it seems to work on Linux.

pmachapman:
I have tested this on Windows, and it runs. Some tests fail because the browser window does not have focus (this happens when running `npm test` if the browser window that is created does not have the caret), but that is a separate issue to this PR (and something that should be addressed in another PR).

Nateowami:
It seems a lot of people want this, but no one has reviewed it...

pmachapman (changes requested):
@Nateowami The `package-lock.json` file has revert to an older version (version 2 format) - are you able to make this so it uses the version 3 format (the diff for this file is massive when it should be small).

Also, do all the tests run for you? On Windows, chapter audio tests and question audio tests fail because the window running the tests does not have focus (a known bug). I'm OK with this change if these only errors happen on Windows, as we can then fix these separately.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1644#-:-NwkWHN-Dl0VnQdGvt6n:bvzhs8z)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

Nateowami:
@pmachapman Fixed, I think.

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1644#-:-Nwl3U-U3DQ7XWZ21_2Q:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

pmachapman (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1644)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/package.json` line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1644#-NJ8Eq4Y1zr6qrsf_MXx:-Nwl3g-0B3YskvEJAtZ_:bjm27f0) ([raw file](https://github.com/sillsdev/web-xforge/blob/e0dd9679fefbaa4c4efb1d3ee6f5d16b38136d1d/src/SIL.XForge.Scripture/ClientApp/package.json#L13)):*

What about:
```
export KARMA_PARALLEL=true npm test || set KARMA_PARALLEL=true&& npm test
```
I should note that I'm not overly fussed as I used my IDE to run the tests which calls Karma direct and I can set the environment variables in there.
Alternatively, we could have two different scripts to suit Windows or Linux.
</blockquote></details>

Done. The method used works on Windows

pmachapman (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1644)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1645, opened 2022-12-13
CodeQL fixes in Unit Tests

pmachapman:
This PR fixes many straightforward CodeQL warnings in the unit tests. This should reduce the CodeQL alert count, and mean that we can then start looking at the CodeQL warnings in the main project.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1645)

### DIFF
--- a/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs
+++ b/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs
@@ -1011,30 +1011,23 @@ namespace SIL.XForge.Scripture.Services
         [Test]
         public async Task RunAsync_NoRecordOfSyncedToRepositoryVersion_DoesFullSync()
         {
-            foreach (bool isChanged in new bool[] { true, false })
-            {
-                var env = new TestEnvironment();
-                string projectSFId = "project03";
-                string userId = "user01";
+            var env = new TestEnvironment();
+            string projectSFId = "project03";
+            string userId = "user01";
 
-                SFProject project = env.SetupProjectWithExpectedImportedRev(projectSFId, null);
-                Assert.That(
-                    project.Sync.DataInSync,
-                    Is.Null,
-                    "setup. Should be testing what happens when this is null."
-                );
-                string projectPTId = project.ParatextId;
-                env.ParatextService.GetLatestSharedVersion(Arg.Any<UserSecret>(), projectPTId).Returns("1", "2");
-                // The expected after-sync repository version can be past the original, before-sync project version,
-                // even if there were no local changes, since there may be incoming changes from the PT Archives server.
-                string expectedRepositoryVersion = "2";
-                // SUT
-                await env.RunAndAssertContinuesAsync(projectSFId, userId, expectedRepositoryVersion);
-                // Shouldn't be setting repo rev.
-                env.ParatextService
-                    .DidNotReceive()
-                    .SetRepoToRevision(Arg.Any<UserSecret>(), Arg.Any<string>(), Arg.Any<string>());
-            }
+            SFProject project = env.SetupProjectWithExpectedImportedRev(projectSFId, null);
+            Assert.That(project.Sync.DataInSync, Is.Null, "setup. Should be testing what happens when this is null.");
+            string projectPTId = project.ParatextId;
+            env.ParatextService.GetLatestSharedVersion(Arg.Any<UserSecret>(), projectPTId).Returns("1", "2");
+            // The expected after-sync repository version can be past the original, before-sync project version,
+            // even if there were no local changes, since there may be incoming changes from the PT Archives server.
+            string expectedRepositoryVersion = "2";
+            // SUT
+            await env.RunAndAssertContinuesAsync(projectSFId, userId, expectedRepositoryVersion);
+            // Shouldn't be setting repo rev.
+            env.ParatextService
+                .DidNotReceive()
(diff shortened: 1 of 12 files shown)


test/SIL.XForge.Tests/Realtime/ConnectionTests.cs
```diff
@@ -408,7 +408,7 @@ public async Task SubmitOpAsync_UsesUnderlyingRealtimeServerOutsideOfATransactio
             List<Json0Op> op = builder.Op;
 
             // SUT
-            var result = await env.Service.SubmitOpAsync(collection, id, op, snapshot.Data, snapshot.Version);
+            _ = await env.Service.SubmitOpAsync(collection, id, op, snapshot.Data, snapshot.Version);
```
Nateowami, line 411: Is it required to assign the value when awaiting? I think in TS we frequently just don't assign the returned value.

test/SIL.XForge.Tests/Services/UserServiceTests.cs
```diff
@@ -295,9 +295,9 @@ private class TestEnvironment
             public UserService Service { get; }
             public MemoryRepository<UserSecret> UserSecrets { get; }
             public MemoryRealtimeService RealtimeService { get; }
-            public IAuthService AuthService = Substitute.For<IAuthService>();
+            public readonly IAuthService AuthService = Substitute.For<IAuthService>();
```
Nateowami, line 298: How are you deciding which members to mark `readonly`? At first I thought it was the ones that didn't have getters, but `IssuedAt` doesn't follow that pattern.

test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs
```diff
@@ -1011,7 +1011,7 @@ public async Task SyncAsync_ParatextMissingAllChapters()
         [Test]
         public async Task RunAsync_NoRecordOfSyncedToRepositoryVersion_DoesFullSync()
         {
-            foreach (bool isChanged in new bool[] { true, false })
+            foreach (bool _ in new bool[] { true, false })
```
Nateowami, line 1014: I'm confused what's going on here. It looks like this is looping to run the test twice, presumably to test what happens the second time through. But a new TestEnvironment is created, so I would think the second run through would have lost the context of the first, and therefore be testing nothing?

test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs
```diff
@@ -2882,15 +2882,6 @@ public void SetupNoteChanges(string threadId, string verseRef = "MAT 1:1", bool
                         )
                         .Returns(userIdsToUsernames);
                 }
-                else
```
Nateowami, line 2885: I'm missing the context for why this was removed.

Nateowami:
Still trying out reviewing in VS Code. I much prefer that context over Reviewable, but I find the GitHub extension to be a bit buggy.

test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs
pmachapman: Nothing was returned from the code. Variables were set, but they were not returned.

test/SIL.XForge.Tests/Realtime/ConnectionTests.cs
pmachapman: You are right, it is not necessary. Sometimes I will use this pattern to make sure it is clear in the test that the return value is ignored, but it is not done elsewhere. Removed.

test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs
Nateowami: Ah, right. There were no side effects.

test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs
pmachapman: Hmmm you are right. I originally thought it needed to run twice, but it looks like the environment around the test changed, but the test did not keep up. I have removed the foreach loop.

test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs
Nateowami: Ah, right. There were no side effects.

test/SIL.XForge.Tests/Services/UserServiceTests.cs
pmachapman: `readonly` means that the variable can only be set in the constructor. It stops other code inadvertently changing the variable to refer to another object and causing unintended side effects.

Nateowami: So is there a reason `IssuedAt` wouldn't be marked `readonly` as well? As far as I can tell it is only ever read.

pmachapman: Good spotting. Because it had an `=>` assignment, not an `=`. Fixed - thank you! I couldn't find any others like this in the test classes with a quick search.

Nateowami: Ah, OK, I didn't even notice it was a method.

Nateowami (approved):

PR #1646, opened 2022-12-15
Change default settings for exporting answers

nigel-wells:
- New projects added will default to "Answers marked for export"
- Existing projects with no method set will default to "All answers" to mimic the pre-existing workflow

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts
@@ -354,8 +354,7 @@ export class SettingsComponent extends DataLoadingComponent implements OnInit {
       usersSeeEachOthersResponses: this.projectDoc.data.checkingConfig.usersSeeEachOthersResponses,
       checkingShareEnabled: this.projectDoc.data.checkingConfig.shareEnabled,
       checkingShareLevel: this.projectDoc.data.checkingConfig.shareLevel,
-      checkingAnswerExport:
-        this.projectDoc.data.checkingConfig.answerExportMethod ?? CheckingAnswerExport.MarkedForExport
+      checkingAnswerExport: this.projectDoc.data.checkingConfig.answerExportMethod ?? CheckingAnswerExport.All
     };
     this.form.reset(this.previousFormValues);
     this.setIndividualControlDisabledStates();
--- a/src/SIL.XForge.Scripture/Models/CheckingConfig.cs
+++ b/src/SIL.XForge.Scripture/Models/CheckingConfig.cs
@@ -6,6 +6,6 @@ namespace SIL.XForge.Scripture.Models
         public bool UsersSeeEachOthersResponses { get; set; } = true;
         public bool ShareEnabled { get; set; } = false;
         public string ShareLevel { get; set; } = CheckingShareLevel.Specific;
-        public string AnswerExportMethod { get; set; } = CheckingAnswerExport.MarkedForExport;
+        public string AnswerExportMethod { get; set; }
     }
 }
--- a/src/SIL.XForge.Scripture/Models/SFProjectCreateSettings.cs
+++ b/src/SIL.XForge.Scripture/Models/SFProjectCreateSettings.cs
@@ -6,5 +6,6 @@ namespace SIL.XForge.Scripture.Models
         public bool TranslationSuggestionsEnabled { get; set; }
         public string SourceParatextId { get; set; }
         public bool CheckingEnabled { get; set; }
+        public string AnswerExportMethod { get; set; } = CheckingAnswerExport.MarkedForExport;
     }
 }
--- a/src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs
@@ -105,7 +105,11 @@ namespace SIL.XForge.Scripture.Services
                     for (int j = 0; j < question.Answers.Count; j++)
                     {
                         Answer answer = question.Answers[j];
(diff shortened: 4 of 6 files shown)


Nateowami (commented):
I'm a little bit confused about this change. Is there a reason we can't just make All the default, but when a project is created, set it to MarkedForExport? I think there's a reason for some of these changes that I'm not seeing.

> Reviewed 2 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1646#-:-NJIg_Qs2dSPY9JwrtEW:bqsuqk4)* status: 2 of 6 files reviewed, all discussions resolved

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1646)* status: 2 of 6 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/Models/CheckingConfig.cs` line 9 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1646#-NJIiUCi3oxichDmhgOi:-NJIiUCi3oxichDmhgOj:br9ensz) ([raw file](https://github.com/sillsdev/web-xforge/blob/2a343eb9370dc159b566cf001505f308285f5239/src/SIL.XForge.Scripture/Models/CheckingConfig.cs#L9)):*
> ```Smalltalk
>         public bool ShareEnabled { get; set; } = false;
>         public string ShareLevel { get; set; } = CheckingShareLevel.Specific;
>         public string AnswerExportMethod { get; set; } = CheckingAnswerExport.MarkedForExport;
> ```

The strategy you're describing is what I expected, but I'm not following how it corresponds to what I'm seeing. Take for example this line. Isn't the default value here going to be overwritten by whatever we pull from the DB? Or maybe I'm misunderstanding how the CheckingConfig works and what the default values here do.

(If we can't make this relatively simple without a migration, I'd rather just do a migrating and put the problem behind us. I thought this would be fairly simple without a migration).

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1646)* status: 2 of 6 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/Models/CheckingConfig.cs` line 9 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1646#-NJIiUCi3oxichDmhgOi:-NJIn7VqD5MyuVQU4YqB:bqicgb4) ([raw file](https://github.com/sillsdev/web-xforge/blob/2a343eb9370dc159b566cf001505f308285f5239/src/SIL.XForge.Scripture/Models/CheckingConfig.cs#L9)):*

The strategy you're describing is what I expected, but I'm not following how it corresponds to what I'm seeing. Take for example this line. Isn't the default value here going to be overwritten by whatever we pull from the DB? Or maybe I'm misunderstanding how the CheckingConfig works and what the default values here do.

(If we can't make this relatively simple without a migration, I'd rather just do a migrating and put the problem behind us. I thought this would be fairly simple without a migration).
</blockquote></details>

Correct. In the case of existing projects, `null`, will be the value for this as nothing is in the database. In that instance it will be treated as "All" and answers will be exported. If something is in the database then that value will be used and that will only be true for newly created projects (where this value is set on creation) or anyone who has changed the setting.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1646)* status: 2 of 6 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/Models/CheckingConfig.cs` line 9 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1646#-NJIiUCi3oxichDmhgOi:-NJJ74J1ALy4ij7JtRZD:bj2bws7) ([raw file](https://github.com/sillsdev/web-xforge/blob/2a343eb9370dc159b566cf001505f308285f5239/src/SIL.XForge.Scripture/Models/CheckingConfig.cs#L9)):*

Correct. In the case of existing projects, `null`, will be the value for this as nothing is in the database. In that instance it will be treated as "All" and answers will be exported. If something is in the database then that value will be used and that will only be true for newly created projects (where this value is set on creation) or anyone who has changed the setting.
</blockquote></details>

What I'm missing is why this line wouldn't be:
``` cs
public string AnswerExportMethod { get; set; } = CheckingAnswerExport.All;
```
Then I would think we wouldn't need null checks elsewhere, and the logic would be contained here (rather than spread around elsewhere).

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1646)* status: 2 of 6 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/Models/CheckingConfig.cs` line 9 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1646#-NJIiUCi3oxichDmhgOi:-NJMot00C5gMvfYZDftI:bwovd1f) ([raw file](https://github.com/sillsdev/web-xforge/blob/2a343eb9370dc159b566cf001505f308285f5239/src/SIL.XForge.Scripture/Models/CheckingConfig.cs#L9)):*

What I'm missing is why this line wouldn't be:
``` cs
public string AnswerExportMethod { get; set; } = CheckingAnswerExport.All;
```
Then I would think we wouldn't need null checks elsewhere, and the logic would be contained here (rather than spread around elsewhere).
</blockquote></details>

Thanks for persevering on this. I had tried that and it wasn't working as expected across existing and new projects. I suspect my issue is related to what I mentioned in Slack around `dotnet clean` to have these model changes take affect. 

I've tried again now and it is working as hoped with this being set.

Take another look now.

Nateowami (approved):
:lgtm:

PR #1652, opened 2022-12-21
Update the RealtimeServer Connection to be IAsyncDisposable

pmachapman:
The `Dispose()` method of the `Connection` class calls the `Disconnect()` method in the RealtimeServer in a synchronous manner. As this code is supposed to be asynchronous (as an `HttpClient` is utilized by Jering to talk to the RealtimeServer), it will block the thread until the call is completed. If there are issues, this could result in thread pool starvation or worst case, a possible deadlock.

C# 8.0 added `IAsyncDisposable`, which means that classes that implement it can be disposed asynchronously by changing:

```C#
using IConnection conn = await _realtimeService.ConnectAsync();
```

to:

```C#
await using IConnection conn = await _realtimeService.ConnectAsync();
```

This Pull Request adds support for the `IConnection` interface to be `IAsyncDisposable` and updates `using` statements that connect to the RealtimeService to be `await using`.

In addition, I have updated 4 other places which open Streams that can be disposed asynchronously, to help improve concurrency.

Finally, I have not updated the RealtimeService to implement `IAsyncDisposable` because that class is a singleton.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1652)

### DIFF
--- a/src/SIL.XForge.Scripture/Controllers/SFProjectsUploadController.cs
+++ b/src/SIL.XForge.Scripture/Controllers/SFProjectsUploadController.cs
@@ -43,17 +43,15 @@ namespace SIL.XForge.Scripture.Controllers
         {
             try
             {
-                using (Stream stream = file.OpenReadStream())
-                {
-                    Uri uri = await _projectService.SaveAudioAsync(
-                        _userAccessor.UserId,
-                        projectId,
-                        dataId,
-                        Path.GetExtension(file.FileName),
-                        stream
-                    );
-                    return Created(uri.PathAndQuery, Path.GetFileName(uri.AbsolutePath));
-                }
+                await using Stream stream = file.OpenReadStream();
+                Uri uri = await _projectService.SaveAudioAsync(
+                    _userAccessor.UserId,
+                    projectId,
+                    dataId,
+                    Path.GetExtension(file.FileName),
+                    stream
+                );
+                return Created(uri.PathAndQuery, Path.GetFileName(uri.AbsolutePath));
             }
             catch (ForbiddenException)
             {
--- a/src/SIL.XForge.Scripture/Services/SFBuildHandler.cs
+++ b/src/SIL.XForge.Scripture/Services/SFBuildHandler.cs
@@ -25,7 +25,7 @@ namespace SIL.XForge.Scripture.Services
 
         public override async Task OnCompleted(BuildContext context)
         {
-            using (IConnection conn = await _realtimeService.ConnectAsync())
+            await using (IConnection conn = await _realtimeService.ConnectAsync())
             {
                 IDocument<SFProject> project = await conn.FetchAsync<SFProject>(context.Engine.Projects.First());
                 if (!project.IsLoaded)
--- a/src/SIL.XForge.Scripture/Services/SFProjectService.cs
+++ b/src/SIL.XForge.Scripture/Services/SFProjectService.cs
@@ -106,7 +106,7 @@ namespace SIL.XForge.Scripture.Services
                 throw new ForbiddenException();
 
             string projectId = ObjectId.GenerateNewId().ToString();
-            using (IConnection conn = await RealtimeService.ConnectAsync(curUserId))
+            await using (IConnection conn = await RealtimeService.ConnectAsync(curUserId))
             {
                 if (
                     this.RealtimeService
@@ -196,7 +196,7 @@ namespace SIL.XForge.Scripture.Services
             await _syncService.CancelSyncAsync(curUserId, projectId);
 
             string ptProjectId;
-            using (IConnection conn = await RealtimeService.ConnectAsync(curUserId))
+            await using (IConnection conn = await RealtimeService.ConnectAsync(curUserId))
             {
                 IDocument<SFProject> projectDoc = await conn.FetchAsync<SFProject>(projectId);
                 if (!projectDoc.IsLoaded)
@@ -242,7 +242,7 @@ namespace SIL.XForge.Scripture.Services
 
         public async Task UpdateSettingsAsync(string curUserId, string projectId, SFProjectSettings settings)
         {
-            using (IConnection conn = await RealtimeService.ConnectAsync(curUserId))
+            await using (IConnection conn = await RealtimeService.ConnectAsync(curUserId))
             {
                 IDocument<SFProject> projectDoc = await conn.FetchAsync<SFProject>(projectId);
                 if (!projectDoc.IsLoaded)
@@ -604,7 +604,7 @@ namespace SIL.XForge.Scripture.Services
         /// <summary> Check that a share link is valid for a project and add the user to the project. </summary>
         public async Task CheckLinkSharingAsync(string curUserId, string projectId, string shareKey)
         {
-            using (IConnection conn = await RealtimeService.ConnectAsync(curUserId))
+            await using (IConnection conn = await RealtimeService.ConnectAsync(curUserId))
             {
                 IDocument<SFProject> projectDoc = await GetProjectDocAsync(projectId, conn);
                 SFProject project = projectDoc.Data;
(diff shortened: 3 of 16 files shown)


RaymondLuong3 (changes requested):
Is there a reason why the instances of `IConnection` in `ProjectService.cs` do not get awaited? It looks like it is used in the same way.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1652#-:-NKti-sI8emONUdBTmDs:b-p63qrr)* status: all files reviewed, 3 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge/Realtime/Connection.cs` line 392 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1652#-NKtg-GH4SvamYOSp9oz:-NKtg-GH4SvamYOSp9p-:bkbeace) ([raw file](https://github.com/sillsdev/web-xforge/blob/87fdc26d2925546c87b3e1667840b93d6ca52b02/src/SIL.XForge/Realtime/Connection.cs#L392)):*
> ```Smalltalk
>         }
> 
>         public async ValueTask DisposeAsync()
> ```

Public methods should come above other methods.



_Code quote:_
```Smalltalk
        public async ValueTask DisposeAsync()
        {
            await _realtimeService.Server.DisconnectAsync(_handle);
            GC.SuppressFinalize(this);
        }
```
___
*[`src/SIL.XForge/Realtime/Connection.cs` line 395 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1652#-NKtg9TWCBceJwp1Zpgd:-NKtg9TWCBceJwp1Zpge:bizm1eb) ([raw file](https://github.com/sillsdev/web-xforge/blob/87fdc26d2925546c87b3e1667840b93d6ca52b02/src/SIL.XForge/Realtime/Connection.cs#L395)):*
> ```Smalltalk
>         {
>             await _realtimeService.Server.DisconnectAsync(_handle);
>             GC.SuppressFinalize(this);
> ```

Why do we need to suppress the garbage collector? Does it have to do with the asynchronous nature of this task?



_Code quote:_
```Smalltalk
        GC.SuppressFinalize(this);
```
___
*[`src/SIL.XForge/Realtime/IConnection.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1652#-NKtgU2r4NfS_wy4x2Nc:-NKtgU2r4NfS_wy4x2Nd:b13g6vi) ([raw file](https://github.com/sillsdev/web-xforge/blob/87fdc26d2925546c87b3e1667840b93d6ca52b02/src/SIL.XForge/Realtime/IConnection.cs#L8)):*
> ```Smalltalk
> namespace SIL.XForge.Realtime
> {
>     public interface IConnection : IDisposable, IAsyncDisposable
> ```

Does this class still need to implement IDisposable?



_Code quote:_
```Smalltalk
 IDisposable,
```

pmachapman (commented):
They are awaited, unless I have missed some? Please let me know if I have accidentally skipped any.

`await using (IConnection conn = await RealtimeService.ConnectAsync(curUserId))`

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1652#-:-NKtpTpt3CN0Usrr5xP-:b-k25t4o)* status: all files reviewed, 3 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge/Realtime/Connection.cs` line 392 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1652#-NKtg-GH4SvamYOSp9oz:-NKtpnIe9S-JWFbhpvll:b-lssfsj) ([raw file](https://github.com/sillsdev/web-xforge/blob/87fdc26d2925546c87b3e1667840b93d6ca52b02/src/SIL.XForge/Realtime/Connection.cs#L392)):*

Public methods should come above other methods.
</blockquote></details>

Thank you! Fixed.
___
*[`src/SIL.XForge/Realtime/Connection.cs` line 395 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1652#-NKtg9TWCBceJwp1Zpgd:-NKtq-fR67Aj7KqchlwC:btbzptz) ([raw file](https://github.com/sillsdev/web-xforge/blob/87fdc26d2925546c87b3e1667840b93d6ca52b02/src/SIL.XForge/Realtime/Connection.cs#L395)):*

Why do we need to suppress the garbage collector? Does it have to do with the asynchronous nature of this task?
</blockquote></details>

This statement notifies the GC that it does not need to dispose this object again, and so run this method again.

See Microsoft's suggested `DisposeAsync()` pattern at https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync for details.
___
*[`src/SIL.XForge/Realtime/IConnection.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1652#-NKtgU2r4NfS_wy4x2Nc:-NKtqXA1BEMOiZ5rK05v:baybe04) ([raw file](https://github.com/sillsdev/web-xforge/blob/87fdc26d2925546c87b3e1667840b93d6ca52b02/src/SIL.XForge/Realtime/IConnection.cs#L8)):*
It is not a requirement, but it is suggested by Microsoft to stop resource leaks:

> It is typical when implementing the IAsyncDisposable interface that classes will also implement the IDisposable interface. A good implementation pattern of the IAsyncDisposable interface is to be prepared for either synchronous or asynchronous disposal, however, this isn't a requirement.

> If you implement the IAsyncDisposable interface but not the IDisposable interface, your app can potentially leak resources. If a class implements IAsyncDisposable, but not IDisposable, and a consumer only calls Dispose, your implementation would never call DisposeAsync. This would result in a resource leak.

Source: https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync

RaymondLuong3 (approved):
You're right. I'm not sure how I missed that. Ok, looks like a nice improvement.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1652#-:-NKtsFk-42MV0YMbsVc7:b-nf6ih)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge/Realtime/Connection.cs` line 395 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1652#-NKtg9TWCBceJwp1Zpgd:-NKts7p15SVwttXafQgn:b-114t9v) ([raw file](https://github.com/sillsdev/web-xforge/blob/87fdc26d2925546c87b3e1667840b93d6ca52b02/src/SIL.XForge/Realtime/Connection.cs#L395)):*

This statement notifies the GC that it does not need to dispose this object again, and so run this method again.

See Microsoft's suggested `DisposeAsync()` pattern at https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync for details.
</blockquote></details>

Thanks for the link. Good.
___
*[`src/SIL.XForge/Realtime/IConnection.cs` line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1652#-NKtgU2r4NfS_wy4x2Nc:-NKtsD8d8Nx_vEVPU1qo:bb70w76) ([raw file](https://github.com/sillsdev/web-xforge/blob/87fdc26d2925546c87b3e1667840b93d6ca52b02/src/SIL.XForge/Realtime/IConnection.cs#L8)):*

It is not a requirement, but it is suggested by Microsoft to stop resource leaks:

> It is typical when implementing the IAsyncDisposable interface that classes will also implement the IDisposable interface. A good implementation pattern of the IAsyncDisposable interface is to be prepared for either synchronous or asynchronous disposal, however, this isn't a requirement.

> If you implement the IAsyncDisposable interface but not the IDisposable interface, your app can potentially leak resources. If a class implements IAsyncDisposable, but not IDisposable, and a consumer only calls Dispose, your implementation would never call DisposeAsync. This would result in a resource leak.

Source: https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync
</blockquote></details>

Nice

PR #1656, opened 2023-01-04
SF-1421 The correct first chapter is now selected if chapter 1 is missing from a book

pmachapman:
When chapter one is removed from a book (see SF-1415), the editor will not select the new first chapter of the book. Instead, the chapter dropdown will be blank, and the user needs to manually select the first chapter.

This Pull Request updates the editor so that it will always select the first chapter of a book when that book is navigated to.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1656)


Nateowami (commented):
Looks good. This has been annoying me for a long time.

Nateowami (approved):

PR #1671, opened 2023-01-20
Switched to AsyncKeyedLock

MarkCiliaVincenti:
This will reduce memory allocations and will clean up from the dictionary when it's done.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1671)

### DIFF
--- a/src/SIL.XForge.Scripture/SIL.XForge.Scripture.csproj
+++ b/src/SIL.XForge.Scripture/SIL.XForge.Scripture.csproj
@@ -18,6 +18,7 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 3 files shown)


MarkCiliaVincenti:
@pmachapman

Nateowami:
Thanks for your contribution. However, I prefer to avoid adding external dependencies when it's possible to avoid doing so. I realize the current implementation creates a dictionary that only ever grows over time, but I'm fairly confident if that memory usage every becomes a problem, we would be long past the point where it's necessary to run a pool of servers, which would necessitate using an out of process system, such as Redis, to synchronize this and a whole host of other things.

PR #1673, opened 2023-01-24
SF-1844a Add Authorization to the SignalR hub

pmachapman:
During implementation of the SignalR hub, the requirement to authorize connections was accidentally omitted. Due to SignalR's slightly different method of configuration than controllers, this was not picked up prior to merging.

This Pull Request:

 * Enables authorization for the Project Notifications SignalR hub
 * Modifies the Project Notification Service to send the access_token
 * Updates the JwtBearer event handler to read the access_token from the query string if the request is for the Project Notifications SignalR hub (this is how SignalR sends the token)
   * Code is adapted from https://learn.microsoft.com/en-us/aspnet/core/signalr/authn-and-authz?view=aspnetcore-7.0
   * See https://github.com/dotnet/aspnetcore/blob/main/src/SignalR/clients/ts/signalr/src/WebSocketTransport.ts#L75

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1673)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/project-notification.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/project-notification.service.ts
@@ -1,14 +1,18 @@
 import { Injectable } from '@angular/core';
-import { HubConnection, HubConnectionBuilder } from '@microsoft/signalr';
+import { HubConnection, HubConnectionBuilder, IHttpConnectionOptions } from '@microsoft/signalr';
+import { AuthService } from 'xforge-common/auth.service';
 
 @Injectable({
   providedIn: 'root'
 })
 export class ProjectNotificationService {
   private connection: HubConnection;
+  private options: IHttpConnectionOptions = {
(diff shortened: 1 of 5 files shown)


Nateowami:
How feasible would it be to add a test for the authentication?

Nateowami:
I clicked the wrong button; I mean to click "close" not "close with comment".

pmachapman:
> How feasible would it be to add a test for the authentication?

I can't think of a straightforward way to do it, without spinning up the backend server. I think end-to-end testing would be the best way to achieve this, when we SF running in containers.

Nateowami (approved):
Ah, I understand now. The authentication is exactly the same as for other endpoints, the only change is how the token is sent. Perhaps I should have reviewed things more closely before thinking about tests.

> Reviewed 2 of 3 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1673#-:-NMfYmD33R8REc3fvvBc:b10znm9)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #1675, opened 2023-01-24
SF-1549 log when ignoring incoming older token

marksvc:
A hypothesis has been that this IssuedAt check may be preventing us
from storing a refresh token. This could be changed to always accept
the incoming token, but for now just log a note if this situation
occurs.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1675)

### DIFF
--- a/src/SIL.XForge/Services/UserService.cs
+++ b/src/SIL.XForge/Services/UserService.cs
@@ -2,6 +2,7 @@ using System;
 using System.Linq;
 using System.Text.RegularExpressions;
 using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Newtonsoft.Json.Linq;
 using SIL.XForge.Configuration;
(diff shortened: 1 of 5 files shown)


src/SIL.XForge/Services/UserService.cs
```diff
@@ -109,6 +113,20 @@ await using (IConnection conn = await _realtimeService.ConnectAsync(curUserId))
...
+                    string incomingIAt = newPTTokens.IssuedAt.ToString(
+                        "o",
+                        System.Globalization.CultureInfo.InvariantCulture
+                    );
+                    string currentIAt = userSecret.ParatextTokens.IssuedAt.ToString(
+                        "o",
+                        System.Globalization.CultureInfo.InvariantCulture
+                    );
+                    _logger.LogWarning(
+                        $"When updating user from auth0 profile, ignoring incoming tokens which were issued at {incomingIAt}, which is earlier than the current tokens {currentIAt}."
```
pmachapman, line 127: Would we want to log the user Id so we can begin diagnosis?

pmachapman (approved):
Approved - just one (non-blocking) question.

src/SIL.XForge/Services/UserService.cs
marksvc: Yes :). Thank you.

PR #1677, opened 2023-01-26
Add return type warning to eslint

nigel-wells:
Show a lint warning if a method doesn't have a return type set.

We decided against making this an error as there are 300 methods that need to be fixed. We decided that we can fix them as we come across them and then change the lint rule to an error in the future.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1677)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/.eslintrc.json
+++ b/src/SIL.XForge.Scripture/ClientApp/.eslintrc.json
@@ -35,6 +35,7 @@
         ],
         "@typescript-eslint/consistent-type-definitions": "error",
         "@typescript-eslint/dot-notation": "off",
+        "@typescript-eslint/explicit-function-return-type": "warn",
         "@typescript-eslint/explicit-member-accessibility": [
           "off",
           {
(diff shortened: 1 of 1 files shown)


src/SIL.XForge.Scripture/ClientApp/.eslintrc.json
```diff
@@ -35,6 +35,7 @@
         ],
         "@typescript-eslint/consistent-type-definitions": "error",
         "@typescript-eslint/dot-notation": "off",
+        "@typescript-eslint/explicit-function-return-type": "warn",
```
Nateowami, line 38: Some of the warnings seem not that helpful:


What about turning on `allowExpressions`? That drops the lint warnings from 299 to 245.

``` suggestion
        "@typescript-eslint/explicit-function-return-type": [
          "warn",
          {
            "allowExpressions": true
          }
        ],
```

Nateowami (changes requested):

src/SIL.XForge.Scripture/ClientApp/.eslintrc.json
nigel-wells: Agreed

Nateowami (approved):

PR #1678, opened 2023-01-27
Save sync metrics logs to DB as they are created

Nateowami:
We've had a number of sync failures recently with very little insight into what went wrong because Jering.Javascript doesn't give a complete stack trace (nowhere in the stack trace does our code appear). The failures usually end with a log of "RunAsync: ParatextData SendReceive finished without throwing." followed by a Jering.Javascript exception.

I'm not certain that this is the right approach, but we clearly need to have more detailed logging of where things are going wrong.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1678)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
@@ -1668,6 +1668,16 @@ namespace SIL.XForge.Scripture.Services
             LogMetric(message);
         }
 
-        private void LogMetric(string message) => _syncMetrics.Log.Add($"{DateTime.UtcNow} {message}");
+        private void LogMetric(string message)
+        {
+            _syncMetrics.Log.Add($"{DateTime.UtcNow} {message}");
+            _syncMetricsRepository
+                .ReplaceAsync(_syncMetrics, true)
+                .ContinueWith(task =>
+                {
+                    if (!task.Result)
+                        _logger.LogInformation("The sync metrics could not be updated in MongoDB");
+                });
+        }
     }
 }


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1678)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @github-code-scanning[bot] and @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 1678 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1678#-NMorisEsnwKPiv-DA-Y-r1-1678:-NMozXRc1DxFo1J6RbLH:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/0b8339a82b61820b7e627eeaa21d4d654e08ba6c/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L1678)):*

## Unnecessarily complex Boolean expression

The expression 'A == false' can be simplified to '!A'.

[Show more details](https://github.com/sillsdev/web-xforge/security/code-scanning/461)
</blockquote></details>

Done.

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1678)* status: all files reviewed, 1 unresolved discussion (waiting on @github-code-scanning[bot] and @pmachapman)

Nateowami:
@pmachapman This might still be worth reviewing, despite being merged already, since it's a candidate to be a hot fix for live.

src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
```diff
@@ -1668,6 +1668,16 @@ private void Log(string message, string? projectSFId = null, string? userId = nu
             LogMetric(message);
         }
 
-        private void LogMetric(string message) => _syncMetrics.Log.Add($"{DateTime.UtcNow} {message}");
+        private void LogMetric(string message)
+        {
+            _syncMetrics.Log.Add($"{DateTime.UtcNow} {message}");
+            _syncMetricsRepository
+                .ReplaceAsync(_syncMetrics, true)
```
pmachapman, line 1675: As this is not called with an `await`, there is a risk that this may be executed out of order, or that the async state might get messed up. Depending on how the MongoDB library uses the async state, it may cause bigger issues than just out of order executions. It probably won't...but I can't tell without heavy testing.

Is there a reason you didn't make this method an `async Task`?

PR #1682, opened 2023-01-30
Deploy production as not self-contained

marksvc:
Building for production to use the host dotnet, rather than building a
self-contained set of files, may broaden the diagnostics we can gather
regarding memory usage.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1682)

### DIFF
--- a/scripts/build-production
+++ b/scripts/build-production
@@ -30,4 +30,4 @@ EOF
 rm -rf "${BUILD_OUTPUT}"
 dotnet publish "src/${PROJECT}/${PROJECT}.csproj" -c "${CONFIGURATION}" -r "${DEPLOY_RUNTIME}" \
   -o "${BUILD_OUTPUT}/app" /p:Version="${BUILD_NUMBER}" /p:AngularConfig="${ANGULAR_CONFIG}" \
-  --self-contained
+  --self-contained false


marksvc:
The output of building this way omits Microsoft.*.dll, System.*.dll, and other files that were included with the self-contained build. This commit was not accompanied by testing as to what will actually happen on the server, so QA will be helpful.

scripts/build-production
```diff
@@ -30,4 +30,4 @@ EOF
 rm -rf "${BUILD_OUTPUT}"
 dotnet publish "src/${PROJECT}/${PROJECT}.csproj" -c "${CONFIGURATION}" -r "${DEPLOY_RUNTIME}" \
   -o "${BUILD_OUTPUT}/app" /p:Version="${BUILD_NUMBER}" /p:AngularConfig="${ANGULAR_CONFIG}" \
-  --self-contained
+  --self-contained true
```
pmachapman, line 33: Doesn't this enable self-contained? `--no-self-contained` disables self-contained.

Nateowami:
Is our current deploy process going to properly handle multiple files being output, rather than deploying just a single file? See `scripts/build-and-ship`

scripts/build-production
marksvc: Thank you. Looks like I forgot to _save_ my file again after some testing.

marksvc:
> Is our current deploy process going to properly handle multiple files being output, rather than deploying just a single file? See `scripts/build-and-ship`

Good question. With self-contained=true, we produce dozens of output files, including .dll files. With self-contained=false, we produce those same output files plus dozens more Microsoft.*.dll and System.*.dll files. I think the rsync deploy should still work. Though I will be more comfortable when I see it all go through and run successfully on QA.

This does however make me wonder if this setting is doing the right thing, since we discussed the usefulness of turning off single file output. It looks like single file output is another setting, which we are not using. I just tried single file output and it omits dozens of .dll files in favour of a larger main executable. 
@pmachapman , maybe we are already doing what we need since we aren't using single file output already?

marksvc:
(And FWIW, I observe that self-contained=true, PublishSingleFile=true, produces the same amount of files, but the main executable is much bigger.)

pmachapman:
> maybe we are already doing what we need since we aren't using single file output already?

Hmmmm. It might be something else causing `dotnet dump` to fail. I'll see what I can find.

marksvc:
I think I'm going to close this PR. We can come back to this if we think it's something we want to pursue further.

PR #1683, opened 2023-01-31
Stop Hangfire jobs deadlocking when a preceding job will not start

pmachapman:
When Scripture Forge restarts, Hangfire assigns a new server GUID. This replaces the previous server instance, and so any jobs that were scheduled for execution before reboot will be stuck in an aborted state, with the following status on the Job Details screen:

image

If another job is waiting for that job to finish (it will not start in this state), that job will check every 15 seconds to see if the preceding job has finished (it will not finish), until the maximum document size is reached and Hangfire crashes.

To resolve this, an `InvisibilityTimeout` setting was added to Hangfire, however this is not configured by default for the Mongo storage engine we use. See [Hangfire.Mongo's README.md](https://github.com/Hangfire-Mongo/Hangfire.Mongo#migration):

> NOTE: By default the parameter `InvisibilityTimeout` of the `MongoStorageOptions` is configured with the value null, making the job to stay in status 'processing' in case of an error in the application. To solve this issue, set the value to 30 minutes like in the [SqlServerStorageOptions](https://docs.hangfire.io/en/latest/configuration/using-sql-server.html).

This pull request resolves this deadlock by setting the` InvisibilityTimeout` to 30 minutes, as suggested above.

I was able to successfully replicate this issue by restoring the most recent backup of `sf_jobs` from production to my workstation, and running Scripture Forge first without this fix, then with this fix to confirm that this is indeed the issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1683)

### DIFF
--- a/src/SIL.XForge/DataAccess/DataAccessServiceCollectionExtensions.cs
+++ b/src/SIL.XForge/DataAccess/DataAccessServiceCollectionExtensions.cs
@@ -27,7 +27,7 @@ namespace Microsoft.Extensions.DependencyInjection
                         new MongoStorageOptions
                         {
                             CheckQueuedJobsStrategy = CheckQueuedJobsStrategy.TailNotificationsCollection,
-                            InvisibilityTimeout = TimeSpan.FromMinutes(30),
+                            InvisibilityTimeout = TimeSpan.FromMinutes(120),
                             MigrationOptions = new MongoMigrationOptions
                             {
                                 MigrationStrategy = new MigrateMongoMigrationStrategy(),
--- a/src/SIL.XForge/DataAccess/DataAccessServiceCollectionExtensions.cs
+++ b/src/SIL.XForge/DataAccess/DataAccessServiceCollectionExtensions.cs
@@ -27,6 +27,7 @@ namespace Microsoft.Extensions.DependencyInjection
                         new MongoStorageOptions
                         {
                             CheckQueuedJobsStrategy = CheckQueuedJobsStrategy.TailNotificationsCollection,
+                            InvisibilityTimeout = TimeSpan.FromMinutes(30),
                             MigrationOptions = new MongoMigrationOptions
                             {
                                 MigrationStrategy = new MigrateMongoMigrationStrategy(),


Nateowami:
> that job will check every 15 seconds to see if the preceding job has finished (it will not finish), until the maximum document size is reached and Hangfire crashes.

Do you know why this doesn't happen more often? It's happens a couple times, but not afterwards.

pmachapman:
> Do you know why this doesn't happen more often? It's happens a couple times, but not afterwards.

It seems to have a very specific set of circumstances. The reboot/crash must occur during syncing of the source, and the crash must be in such a way that Hangfire's process does not mark the job as failed. I was quite lucky that due to the crash on live, and the copying of the Hangfire database that a suitable job was stuck - there were several other stuck jobs, but they were not blocking any target translation syncs.

We may want to clear the aborted jobs on live, as one will be stuck there right now, I believe?

Nateowami (commented):
I think this looks good. My understanding of the invisibility timeout is that if a worker picked up a job and then the worker hasn't been heard from after this interval, it will be handed off to another available worker?

It's hard to find documentation for InvisibilityTimeout, I think due to using an older version of Hangfire (I can't find docs for our version).

I'm reading https://docs.hangfire.io/en/latest/configuration/using-redis.html?highlight=invisibilitytimeout:
> Obsolete since 2.4.0 Time interval, within which background job is considered to be still successfully processed by a worker. When a timeout is elapsed, another worker will be able to pick the same background job.

marksvc:
So, right now the InvisibilityTimeout is sort of 'infinite', and so it never gets passed to another worker. But setting it to 30 minutes means that after 30 minutes, Hangfire will start processing the job again, but perhaps using another worker instead of the original. 

So there is a 30-minute window during which the every-15-second issue could still happen, but we shorten the window from infinite minutes down to 30 minutes. Is that right?

pmachapman:
> My understanding of the invisibility timeout is that if a worker picked up a job and then the worker hasn't been heard from after this interval, it will be handed off to another available worker?

@Nateowami Sort of. It is more like if a job is still supposed running after 30 minutes, it will be passed to another worker. Would we want to increase this timeout value as we are really only combatting jobs which get stuck long enough to exceed a hard limit in the MongoDB doc size?  We could increase to a day maybe?

> It's hard to find documentation for InvisibilityTimeout, I think due to using an older version of Hangfire (I can't find docs for our version).
@Nateowami The best documentation is the implementation: [Hangfire.Mongo/MongoJobFetcher.cs](https://github.com/Hangfire-Mongo/Hangfire.Mongo/blob/master/src/Hangfire.Mongo/MongoJobFetcher.cs#L130-L139)

> Obsolete since 2.4.0 Time interval, within which background job is considered to be still successfully processed by a worker. When a timeout is elapsed, another worker will be able to pick the same background job.

@Nateowami It is obsolete for Hangfire's Redis and SQL Server storage engines, but not for the MongoDB storage engine (see the various issues on the Hangfire.Mongo Github repo https://github.com/Hangfire-Mongo/Hangfire.Mongo/search?q=InvisibilityTimeout&type=issues).

> So, right now the InvisibilityTimeout is sort of 'infinite', and so it never gets passed to another worker. But setting it to 30 minutes means that after 30 minutes, Hangfire will start processing the job again, but perhaps using another worker instead of the original.

@marksvc Correct.

> So there is a 30-minute window during which the every-15-second issue could still happen, but we shorten the window from infinite minutes down to 30 minutes. Is that right?

@marksvc Yes. We could set it to longer, if you think 30 minutes is too short for a sync or training job execution.

Nateowami (commented):
I think training a New Testament often takes around 20 minutes, so 30 minutes is far from out of the realm of possibility. Maybe set it to an hour or two?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1683#-:-NNo2wzd3oDT6x_eyCOS:b-g4jgk3)* status: 0 of 1 files reviewed, all discussions resolved

pmachapman (commented):
Change to 2 hours.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1683#-:-NNo6v-ZEZtH_ZUbaSEm:b-gfcjbt)* status: 0 of 1 files reviewed, all discussions resolved

Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1683#-:-NNoAxJ53-jzAg2Y8bl4:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #1684, opened 2023-01-31
Revert "Save sync metrics logs to DB as they are created"

Nateowami:
Reverts sillsdev/web-xforge#1678

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1684)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
@@ -1668,16 +1668,6 @@ namespace SIL.XForge.Scripture.Services
             LogMetric(message);
         }
 
-        private void LogMetric(string message)
-        {
-            _syncMetrics.Log.Add($"{DateTime.UtcNow} {message}");
-            _syncMetricsRepository
-                .ReplaceAsync(_syncMetrics, true)
-                .ContinueWith(task =>
-                {
-                    if (!task.Result)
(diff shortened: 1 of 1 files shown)


src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
```diff
@@ -1668,16 +1668,6 @@ private void Log(string message, string? projectSFId = null, string? userId = nu
...
-            _syncMetrics.Log.Add($"{DateTime.UtcNow} {message}");
-            _syncMetricsRepository
-                .ReplaceAsync(_syncMetrics, true)
-                .ContinueWith(task =>
-                {
-                    if (!task.Result)
-                        _logger.LogInformation("The sync metrics could not be updated in MongoDB");
-                });
-        }
+        private void LogMetric(string message) => _syncMetrics.Log.Add($"{DateTime.UtcNow} {message}");
```
pmachapman, line 1671: You didn't want to log to stdout?

pmachapman (approved):

src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
Nateowami: Logging to stdout was done as a hotfix to live, and I also made this change on master. So I was thinking we would revert the change on master completely, and then if we decided we wanted to log to stdout we could chery-pick that change from live.

PR #1676, opened 2023-01-25
SF-1854 Fix note thread permissions

Nateowami:
- I've reformatted the permissions to make them more maintainable
- All Paratext roles have been given view permission on the note threads (thus fixing the actual bug).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1676)

PR #1687, opened 2023-02-01
SF-1854b Fix permissions for community checker

Nateowami:
The original change in #1676 fixed SF-1854, but caused BVT failures because the community checker could no longer view answer status.

- Community checkers need permission to view the answer status, otherwise they won't be able to view questions at all (since we don't have a way to split up the question doc)
- Re-ordered permissions to be more consistent

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1687)


marksvc (approved):

Nateowami:
Closing because this is merged, but GitHub doesn't appear to recognize that.

PR #1689, opened 2023-02-01
Logging hotfix

marksvc:
These cherry-picks are to add additional logging to Live around the area where the Out of memory errors occurred. 

However, Live seems to have recovered from Out of memory troubles from a recent Jering HTTP version change. So I'm not in a rush to send this to Live right away. ?

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1689)


pmachapman (approved):

marksvc:
I'm going to close this and it can gradually make it to Live unless we find again a more urgent reason.

PR #1690, opened 2023-02-06
Update http-cache-semantics to 4.1.1 to fix vulnerability

Nateowami:
This was done with `npm audit fix`

As with most updates like this, I highly doubt we're actually impacted, but I would rather keep packages up to date than try do determine which packages with known vulnerabilities we need to care about and which we should ignore.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1690)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/package-lock.json
+++ b/src/SIL.XForge.Scripture/ClientApp/package-lock.json
@@ -15876,9 +15876,10 @@
       "license": "MIT"
     },
     "node_modules/http-cache-semantics": {
-      "version": "4.1.0",
-      "dev": true,
-      "license": "BSD-2-Clause"
+      "version": "4.1.1",
+      "resolved": "https://registry.npmjs.org/http-cache-semantics/-/http-cache-semantics-4.1.1.tgz",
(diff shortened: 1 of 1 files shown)


pmachapman (approved):
Does this PR remove the audit warning for you? I still get the warning after running `npm ci`, and running `npm audit fix` displays the following message:

> npm WARN audit fix http-cache-semantics@4.1.0 node_modules/npm/node_modules/http-cache-semantics
> npm WARN audit fix http-cache-semantics@4.1.0 is a bundled dependency of
> npm WARN audit fix http-cache-semantics@4.1.0 npm@8.19.2 at node_modules/npm
> npm WARN audit fix http-cache-semantics@4.1.0 It cannot be fixed automatically.
> npm WARN audit fix http-cache-semantics@4.1.0 Check for updates to the npm package.

If this clears a warning you are seeing, though, I don't see any problem merging this, as this is a dependency of npm and not used in the ClientApp.

Nateowami:
npm shows a fix even after this is applied (and keeps offering a fix via `npm audit fix`). Closing until that changes.

PR #1693, opened 2023-02-07
Optimize Comment Manager Loading

pmachapman:
This Pull Request fixes an issue when `CommentManager` reloads each comment XML file from disk every time that the Note Thread Docs are updated for a book.

This bug occurs because `CommentManager.LoadIfChanged()` reloads the comment XML files not only comment XML they have changed on disk, but also if each XML file has not been updated via `CommentManager.SaveUser()`.

This affects not only speed but memory usage as each note in each comment XML file is serialized into `Comment` classes, for every book.

To resolve this issue, the `CommentManager` loads the XML files from disk just once, immediately after the Paratext Sync has completed.

In my testing, the sync time for a project with 10,000 notes and 70 books reduced from ~2:00 minutes to ~1:30 minutes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1693)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -228,11 +228,18 @@ namespace SIL.XForge.Scripture.Services
             if (ptProject is not ParatextResource)
             {
                 StartProgressReporting(progress);
+
+                string username = GetParatextUsername(userSecret);
+                using ScrText scrText = ScrTextCollection.FindById(username, paratextId);
+                if (scrText == null)
+                    throw new Exception(
+                        $"Failed to fetch ScrText for PT project id {paratextId} using PT username {username}"
+                    );
+
                 SharedProject sharedProj = CreateSharedProject(
-                    userSecret,
                     paratextId,
                     ptProject.ShortName,
-                    source.AsInternetSharedRepositorySource(),
+                    scrText,
                     sendReceiveRepository
                 );
                 List<SharedProject> sharedPtProjectsToSr = new List<SharedProject> { sharedProj };
@@ -284,6 +291,10 @@ namespace SIL.XForge.Scripture.Services
                         $"Failed: Errors occurred while performing the sync with the Paratext Server. More information: noErrors: {noErrors}. success: {success}. null results: {results == null}. results: {resultsInfo}"
                     );
                 }
+
+                // Update the cached comment manager
+                CommentManager manager = CommentManager.Get(scrText);
+                manager.Load();
             }
 
             return ptProject;
@@ -1177,13 +1188,7 @@ namespace SIL.XForge.Scripture.Services
                 return null;
 
             CommentManager manager = CommentManager.Get(scrText);
-            var threads = manager.FindThreads(
-                (commentThread) =>
-                {
-                    return commentThread.VerseRef.BookNum == bookNum;
-                },
-                true
-            );
+            var threads = manager.FindThreads(commentThread => commentThread.VerseRef.BookNum == bookNum, true);
             return NotesFormatter.FormatNotes(threads);
         }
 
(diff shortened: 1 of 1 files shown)


src/SIL.XForge.Scripture/Services/ParatextService.cs
```diff
@@ -228,11 +228,19 @@ public void Init()
             if (ptProject is not ParatextResource)
             {
                 StartProgressReporting(progress);
+
+                string username = GetParatextUsername(userSecret);
+                // Specifically set the ScrText property of the SharedProject to indicate the project is available locally
```
Nateowami, line 233: This comment feels like it is now out of place. I can't make sense of what it's trying to communicat here, though it looks like it may have made some sense in its previous context.

src/SIL.XForge.Scripture/Services/ParatextService.cs
```diff
@@ -1177,13 +1189,7 @@ public string GetNotes(UserSecret userSecret, string projectId, int bookNum)
...
 
             CommentManager manager = CommentManager.Get(scrText);
-            var threads = manager.FindThreads(
-                (commentThread) =>
-                {
-                    return commentThread.VerseRef.BookNum == bookNum;
-                },
-                true
-            );
+            var threads = manager.FindThreads(commentThread => commentThread.VerseRef.BookNum == bookNum, true);
```
Nateowami, line 1192: How inefficient do you think this is? It looks like we filter over all threads for every book, so we have number_of_books * number_of_threads operations. If we assume 77 books (not uncommon, I think) and 100,000 threads (probably uncommon), we get 7,700,000 operations. That's inefficient, but these comparisons should be blazing fast, so possibly there is no real issue.

However, a) I'm not super confident it doesn't affect the speed, and b) I'm noticing that we're calling `CommentManager.FindThreads`. Given the one-to-many relationship between threads and comments, I'm wondering if this is an expensive operation to be repeating. Also, we're reading properties like `VerseRef.BookNum`. Since notes can be moved after they're created, I'm guessing it would take some computation to determine what the current verse ref is (even if they're rarely moved, it would still mean you have to read and apply the operations that have happened to the note, not just read a static property). There are probably other properties that could be similarly slow to compute. Of course, I don't know this for sure; it's possible all these operations happen once in the `CommentManager` and the only slowdown here is in repeated filtering, which may be negligible.

src/SIL.XForge.Scripture/Services/ParatextService.cs
pmachapman: Good observation - I've moved it to a more sensible place.

src/SIL.XForge.Scripture/Services/ParatextService.cs
pmachapman: My observations of the code in ParatextData:

 * The lambda function is executed in a `Where()` clause in `CommentManager.FindThreads`. The code in `CommentManager.FindThreads` could be optimized - an in-memory collection is populated from the in-memory `CommentThread` cache then the clause is executed, which is not the most efficient way to do it.
 * The `VerseRef` is retrieved from the first comment in the thread, then the collection of `Comment` objects in the thread is checked from last to first for a reattached location (if there is no reattached location, every `Comment` will be checked).

This is not the most efficient way, but I think we are dealing in tens of milliseconds of inefficiency, not hundreds of milliseconds. I think the best approach would be to create the test project with 100,000 notes, and see how things work out. If the inefficiency does prove detrimental, our only option I can think of would be to implement our own `CommentManager`.

The biggest slowdowns for code are usually IO related, which is what this PR has addressed.

marksvc:
:eyes:

marksvc:
ParatextSyncRunner.RunAsync() calls UpdateParatextNotesAsync() (and soon UpdateParatextCommentsAsync()). 
  This calls ParatextService.GetNotes(), which looks to be fetching the PT comments from the PT project.
Then there is SendReceiveAsync(). Now the comments in the PT project may be different.
Then later UpdateDocsAsync() is called. 
  This goes to UpdateNoteThreadDocsAsync(). 
  This goes to GetNoteThreadChanges().
  This goes to GetCommentThreads().
  And there it queries threads.
If GetCommentThreads() no longer does LoadIfChanged(), might that mean that changes to threads, from doing SendReceive, may not be reflected in GetCommentThreads() output?

I suppose I could prove or disprove this by trying it. And I may not be understanding how this works adequately.

pmachapman:
> If GetCommentThreads() no longer does LoadIfChanged(), might that mean that changes to threads, from doing SendReceive, may not be reflected in GetCommentThreads() output?

@marksvc Correct. To resolve this, I added the following lines to [`ParatextService.SendReceiveAsync()`](https://github.com/sillsdev/web-xforge/blob/0a77c6f85c341794c8a112a27acab44061e4b3d3/src/SIL.XForge.Scripture/Services/ParatextService.cs#L295-L297):

```csharp
// Update the cached comment manager
CommentManager manager = CommentManager.Get(scrText);
manager.Load();
```

This should ensure that the comment manager is updated after SendReceive, unless you see a different path that I have missed?

Nateowami (approved):

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1693)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #1694, opened 2023-02-07
SF-1868 Show mobile keyboard when adding a note

nigel-wells:
- Enable autofocus for the note dialog
- Reduce height of the note dialog for smaller devices
- Only force focus on the target if there are no dialogs open

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1694)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
@@ -893,7 +893,7 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
     const dialogRef = this.dialogService.openMatDialog<NoteDialogComponent, NoteDialogData, boolean>(
       NoteDialogComponent,
       {
-        autoFocus: false,
+        autoFocus: true,
         width: '600px',
         disableClose: true,
         data: noteDialogData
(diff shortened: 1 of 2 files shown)


RaymondLuong3 (changes requested):
This might need more work. I noticed that when I open the note dialog, the three dot menu next to the scripture text gets autofocused. I think there is a way to exclude that as an autofocus element.
BTW, thanks for the instructions for physical device testing in the ReadMe! It worked like a charm.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1694#-:-NOB-9H_Ch0U0z5gUw51:buespj2)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

nigel-wells (commented):
I've tested that scenario out for existing notes and had implemented a fix by setting the menu to not be focusable and have the textarea focusable. I found the user experience wasn't great and probably isn't what is desired. Tapping on a note icon means you will want to read the note first and then decide if you want to leave a comment. If we focus immediately on the textarea, and have the keyboard appear, there is very little screen left to read the note above it. Then image that happening every single time you clicked a note icon and you're likely to start getting frustrated. I'm ok leaving the focus on the menu as it means the keyboard is not displayed unless you want to click into the textarea to leave a note.

Thougths?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1694#-:-NOCUQH83CRDn0UcAJED:bkovpkg)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (approved):
Hmm, I didn't think that it would be annoying that the text area would be in focus each time you open a note. That is probably try. Ok, that seems like the right thing to do.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1694#-:-NOFcVO9ABeguLiGzTGq:b-bnssid)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

PR #1696, opened 2023-02-08
SF-1783 Allow admins to answer questions

nigel-wells:
- Updated project rights for admin to create answers
- Adjusted logic to still allow admin to view answers without being required to answer a question first
- eslint return type fixes

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1696)

### DIFF
--- a/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts
+++ b/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts
@@ -83,7 +83,7 @@ const rightsByRole: Record<SFProjectRole, { [domain in `${SFProjectDomain}`]?: `
     project: ['view'],
     texts: ['view', 'edit'],
     questions: ['view', 'create', 'edit', 'delete'],
-    answers: ['view', 'delete', 'edit_own'],
+    answers: ['view', 'create', 'delete', 'edit_own'],
     answer_comments: ['view', 'create', 'edit_own', 'delete'],
     answer_status: ['view', 'edit'],
     likes: ['view', 'create', 'delete_own'],
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts
@@ -34,6 +34,7 @@ import { CheckingAudioCombinedComponent } from '../checking-audio-combined/check
 import { AudioAttachment } from '../checking-audio-recorder/checking-audio-recorder.component';
 import { CheckingTextComponent } from '../checking-text/checking-text.component';
 import { SFProjectService } from '../../../core/sf-project.service';
+import { SFProjectProfileDoc } from '../../../core/models/sf-project-profile-doc';
 import { CommentAction } from './checking-comments/checking-comments.component';
 
 export interface AnswerAction {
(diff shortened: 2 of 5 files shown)


RaymondLuong3 (changes requested):
I have just one concern which is an edge case.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1696#-:-NNmqqMN2a9DZ7geMVhR:b3uc45p)* status: all files reviewed, 2 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts` line 142 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1696#-NNmpPEU085BWkYEvWVX:-NNmpPEU085BWkYEvWVY:b-sqh31j) ([raw file](https://github.com/sillsdev/web-xforge/blob/e27ba969209553d6b069bfc536cda02ff96d38b4/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L142)):*
> ```XML
>     }
>     this._projectId = projectId;
>     this.projectService.isProjectAdmin(projectId, this.userService.currentUserId).then(isProjectAdmin => {
> ```

I was thinking if it might be better to use a getter called isProjectAdmin so that we can always get the up to date info for the project. The isProjectAdmin property could get stale if a sync happens, and the user's role changed. I think it is unlikely, but still worth mentioning.



_Code quote:_
```XML
    this.projectService.isProjectAdmin(projectId, this.userService.currentUserId).then(isProjectAdmin => {
      this.isProjectAdmin = isProjectAdmin;
    });
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 186 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1696#-NNmq_OdBHv3l63R8AuK:-NNmq_OdBHv3l63R8AuL:b-vv74cc) ([raw file](https://github.com/sillsdev/web-xforge/blob/e27ba969209553d6b069bfc536cda02ff96d38b4/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L186)):*
> ```XML
>   getUnreadAnswers(questionDoc: QuestionDoc): number {
>     if (
>       (this.canAddAnswer && !this.isProjectAdmin) ||
> ```

Nice work finding this change.



_Code quote:_
```XML
    if (
      (this.canAddAnswer && !this.isProjectAdmin) ||
```

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1696)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts` line 142 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1696#-NNmpPEU085BWkYEvWVX:-NNo2Bs5ACrcXIDnhJWT:b-s6pf3z) ([raw file](https://github.com/sillsdev/web-xforge/blob/e27ba969209553d6b069bfc536cda02ff96d38b4/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L142)):*

I was thinking if it might be better to use a getter called isProjectAdmin so that we can always get the up to date info for the project. The isProjectAdmin property could get stale if a sync happens, and the user's role changed. I think it is unlikely, but still worth mentioning.
</blockquote></details>

Good point. I've refactored this a bit so we can subscribe to the doc. I wanted to avoid getting it on demand as it would require everything to be async.

PR #1698, opened 2023-02-08
SF-1870 Parse comment content idempotently

RaymondLuong3:
The version number on community checking notes would increment with each synchronize operation. This was caused by a difference in newLine whitespace around the <p> tags around <bold>community checking question</bold> xml. I could not determine what was removing the newLine whitespace around Paragraph tags.
This change ensures that regardless of those newLine whitespaces so that child nodes in paragraph nodes retain style features.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1698)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/NotesFormatter.cs
+++ b/src/SIL.XForge.Scripture/Services/NotesFormatter.cs
@@ -98,7 +98,7 @@ public static class NotesFormatter
     private static XElement FormatParagraph(XmlNode paraNode)
     {
         XElement paraElem = new XElement("p");
-        if (paraNode.ChildNodes.Count == 1)
+        if (paraNode.ChildNodes.Count == 1 && paraNode.FirstChild.NodeType == XmlNodeType.Text)
             paraElem.Value = paraNode.InnerText;
         else
         {
--- a/test/SIL.XForge.Scripture.Tests/Services/NotesFormatterTests.cs
+++ b/test/SIL.XForge.Scripture.Tests/Services/NotesFormatterTests.cs
@@ -207,6 +207,63 @@ public class NotesFormatterTests
     }
 
     [Test]
+    public void FormatNotes_SingleNodeContentParagraph()
+    {
+        // Setup test environment
+        var env = new TestEnvironment();
+        var associatedPtUser = new SFParatextUser(env.Username01);
+        env.SetupProject(env.Project01, associatedPtUser);
+
+        // Setup the test data
+        const string thread = "Answer_dataId0123";
+        const string verseRefStr = "RUT 1:1";
+        DateTimeOffset commentDate = DateTimeOffset.Now;
+
+        // We user StringBuilder so we can have the environment specific new line
+        StringBuilder sb = new StringBuilder();
+        sb.AppendLine("<notes version=\"1.1\">");
+        sb.AppendLine($"  <thread id=\"{thread}\">");
+        sb.AppendLine($"    <selection verseRef=\"{verseRefStr}\" startPos=\"0\" selectedText=\"\" />");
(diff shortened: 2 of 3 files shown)


marksvc (approved):

RaymondLuong3 (commented):
@marksvc I added a test when I saw that the test coverage dropped dramatically with the changes I made. At first I wasn't sure how to write the test, but then it turned out to be quite simple.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1698#-:-NOGBPDN49N07vKketHQ:b-gckeyo)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @marksvc)

marksvc (commented):
Great!

BTW there is now a NotesFormatterTests.cs.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1698#-:-NOGh4xZ683gt6Wjovkf:bwoplax)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs` line 660 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1698#-NOGh6m9--ohxzNfwFv-:-NOGh6m9--ohxzNfwFv0:br28p2r) ([raw file](https://github.com/sillsdev/web-xforge/blob/ce5d12eb9f5d5f5fdebedfa3ba6e2b37205f3796/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L660)):*
> ```Smalltalk
>             Assert.True(notes.StartsWith(expected));
>             // The paragraph style gets preserved
>             Assert.True(notes.Contains("<span style=\"bold\">Question text</span>"));
> ```

Oh. The PR description seems to imply that the difference being accounted for is whitespace. Trying out the code in debugger with this test I can see that the change you made is resulting in preserving bold styling instead of losing it. (Which seems good.) Am I missing something?

RaymondLuong3 (commented):
Wow, that is what I was looking for but didn't find the first go around.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1698#-:-NOLCKAX826w7fn2_EpR:b-vc54qr)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs` line 660 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1698#-NOGh6m9--ohxzNfwFv-:-NOLCQc11LWVPJnFmSBh:bv143c1) ([raw file](https://github.com/sillsdev/web-xforge/blob/ce5d12eb9f5d5f5fdebedfa3ba6e2b37205f3796/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L660)):*

Oh. The PR description seems to imply that the difference being accounted for is whitespace. Trying out the code in debugger with this test I can see that the change you made is resulting in preserving bold styling instead of losing it. (Which seems good.) Am I missing something?
</blockquote></details>

The reason that some projects didn't run into this problem was because paragraph nodes typically had newLine whitespace that enclosed the child node. But sometimes those newLines are not there, so the paragraph node only has the single child node, and the styles get dropped.  I updated the description. The symptom was that style features get dropped off of the nodes. The fix is to ensure that the paragraph child nodes get processed the same way regardless of whether the whitespace does or does not exist.

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1698)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1701, opened 2023-02-13
SF-1865 Add default tag for community checking answers

RaymondLuong3:
* add the NoteTagId property on the CheckingConfig class
* create community checking tag if project has answers
* add tagsAdded element when getting note thread changelist

When community checking answers get synchronized with Paratext, any new and existing community checking answers will be exported and giving the default checking tag ID. If the default did not previously exist, then it creates a new one. If the 07conversation2 icon is already in use, that tag ID will be given to the question answers.

Community Checking Tag

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1701)

### DIFF
--- a/src/SIL.XForge.Scripture/Models/CheckingConfig.cs
+++ b/src/SIL.XForge.Scripture/Models/CheckingConfig.cs
@@ -7,4 +7,5 @@ public class CheckingConfig
     public bool ShareEnabled { get; set; } = false;
     public string ShareLevel { get; set; } = CheckingShareLevel.Specific;
     public string AnswerExportMethod { get; set; } = CheckingAnswerExport.All;
+    public int? NoteTagId { get; set; }
 }
--- a/src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs
@@ -157,7 +157,8 @@ public class ParatextNotesMapper : IParatextNotesMapper
                         answer,
                         ptProjectUsers,
                         answerPrefixContents,
-                        checkingNoteTagId
+                        checkingNoteTagId,
+                        true
                     );
                     if (answer.SyncUserRef == null)
                         answerSyncUserIds.Add((j, answerSyncUserId));
@@ -237,7 +238,8 @@ public class ParatextNotesMapper : IParatextNotesMapper
         Comment comment,
         Dictionary<string, ParatextUserProfile> ptProjectUsers,
         IReadOnlyList<object> prefixContent = null,
-        int tagId = NoteTag.notSetId
+        int tagId = NoteTag.notSetId,
+        bool setCheckingTag = false
     )
     {
         (string syncUserId, string user, bool canWritePTNoteOnProject) = await GetSyncUserAsync(
@@ -276,7 +278,7 @@ public class ParatextNotesMapper : IParatextNotesMapper
 
         var threadId = (string)threadElem.Attribute("id");
         string key = GetCommentKey(threadId, commentElem, ptProjectUsers);
-        if (IsCommentNewOrChanged(oldCommentElems, key, commentElem))
+        if (IsCommentNewOrChanged(oldCommentElems, key, commentElem, setCheckingTag))
             threadElem.Add(commentElem);
         oldCommentElems.Remove(key);
         return syncUserId;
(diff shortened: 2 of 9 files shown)


src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
```diff
@@ -483,13 +483,18 @@ CancellationToken token
                 if (!_paratextService.IsResource(paratextId))
                 {
                     LogMetric("Updating Paratext notes");
+                    if (questionDocs.Count > 0)
```
pmachapman, line 486: This block gets run, even if no questions are being exported. This means that the note tag is created even if no notes will use it. Is that behavior OK?

src/SIL.XForge.Scripture/Models/CheckingConfig.cs
```diff
@@ -7,5 +7,6 @@ public class CheckingConfig
         public bool ShareEnabled { get; set; } = false;
         public string ShareLevel { get; set; } = CheckingShareLevel.Specific;
         public string AnswerExportMethod { get; set; } = CheckingAnswerExport.All;
+        public int? NoteTagId { get; set; }
```
pmachapman, line 10: I am assuming a later change may allow customization of this value?

src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs
```diff
@@ -265,6 +268,11 @@ string answerExportMethod
                 contentElem.Add(new XElement("p", responseText));
             }
             commentElem.Add(contentElem);
+            if (tagId != NoteTag.notSetId)
```
pmachapman, line 271: I notice that existing answers to not have this tag added (they still have the red checkbox for me), although new answers made after running this change have the new icon. Is that something we want to do?

pmachapman (changes requested):
Looks good - I like the new icon, and it worked well for my test projects. Just a few questions for feedback.

Also, if you want me to do the rebase, and deal with the merge conflicts (seeing as I caused them!) I am happy to.

src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
Nateowami: @pmachapman See the "Paratext note tag integration" document; it has requirements for things like this. One of the things that is specifically called out is to not create a note tag unless it's actually needed to be used. So good catch here (just pointing out the doc so you're aware of it and can see the specs we came up with).

src/SIL.XForge.Scripture/Models/CheckingConfig.cs
Nateowami: @pmachapman The note tag ID cannot be changed, but its properties can, in Paratext. So the associated icon, name, and more can be changed. We don't plan to add a UI for that. See the spec document for more details around this.

src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs
Nateowami: It looks like this is something we forgot to write in the spec.

My initial thought is I think we should update all existing answers to use the proper note tag, since that seems more helpful. Otherwise existing projects wouldn't get their existing answers brought along to the new note tag, and now their answers are spread across two tags with two icons.

However, that could cause a problem if users want to be able to change the note tag on an answer. If we overwrite it on every sync, and a user had deliberately changed what tag was assigned to a note, that could be annoying. (For example, perhaps a note tag called "Answers that we should act upon" would be created, and some answers would be changed to use that tag, so they could be filtered for in Paratext).

We could solve both of these by checking if the note ever had the expected tag, and if never did, update it to the correct tag. But if it used to have the correct tag, but doesn't anymore, then assume a user changed it in Paratext and leave it alone. But that's more work, and depending on how complicated it is to check the history of the note, it may or may not be worth it.

src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
RaymondLuong3: Good catch. I've updated the logic. Now it will only create a note tag if there are exportable answers, meaning that there is an answer that is marked exportable, or there are answers and the project setting for exporting all answers is set.

src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs
RaymondLuong3: Actually, as I was investigating this I found a simple way to retroactively set the note tag. We will always set the answer tag to the current community checking tag, but if a user in Paratext decides to change the tag, a new comment is created to represent the change. Because of how that works, if we retroactively always set the answer tag to be the current checking tag, subsequent tag changes are unaffected.

There may be an argument that we do not want to always change existing answers to have the current tag, but I would argue that we DO want this and that when a user decides that the purple convo icon isn't what they want, they should expect that when they synchronize with SF, all existing answers will get updated to the new tag.

src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
```diff
@@ -1223,6 +1229,22 @@ private async Task UpdateTranslateNoteTag(string targetParatextId)
             }
         }
 
+        private async Task<int> UpdateCheckingNoteTag(string targetParatextId)
```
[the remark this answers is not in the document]
RaymondLuong3: Done.

pmachapman (approved):

PR #1702, opened 2023-02-13
Paratext Comment XML Generator

pmachapman:
I have created a program that can generate a Paratext Comments XML file, in a new directory called "Tools".

If you would rather this goes in another repository, and we just link to it from the SF wiki (when that is done), then that is not problem as it does not actually require any of the Scripture Forge components.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1702)

### DIFF
--- a/tools/CommentGenerator/CommentGenerator.csproj
+++ b/tools/CommentGenerator/CommentGenerator.csproj
@@ -1,0 +1,15 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net6.0</TargetFramework>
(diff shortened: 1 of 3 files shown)


Nateowami:
I think going in this repo is the right choice. I'm not sure the best place to put it. Right now we have several types of tools already, and the directory structure we're using doesn't make much sense. I've broached the subject in Slack

RaymondLuong3 (approved):
Wow, this program allowed me to generate a file with 404k lines in a couple seconds! Nice.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1702#-:-NOLHZuP9o6She0nWPCH:bhgsnrg)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #1703, opened 2023-02-14
tool: allow email arg to db-token-info

marksvc:
commit-id:3d92d270

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1703)

### DIFF
--- a/scripts/db_tools/db-token-info.ts
+++ b/scripts/db_tools/db-token-info.ts
@@ -1,6 +1,8 @@
-#!./node_modules/.bin/ts-node
+#!/usr/bin/env -S bash -c '"$(dirname "$0")"/node_modules/.bin/ts-node "$(dirname "$0")/$(basename "$0")" "$@"'
+// The above causes the local ts-node to be used even if run from another directory. Setup: npm ci
 
 import { MongoClient } from 'mongodb';
+import { exit } from 'node:process';
 import { showJwt } from './show-jwt';
 import { devConfig } from './utils';
 
@@ -17,8 +19,15 @@ type UserSecret = {
   };
 };
 
-// Set to the email of the user to look up
-const email = '';
+// Set to the email of the user to look up (here or via command line argument)
+let email = '';
+const args = process.argv.slice(2);
+if (args.length > 0 && email === '') email = args[0];
+if (email === '') {
+  console.error('Specify user email.');
+  exit(1);
+}
+
 // Set to the connection to fetch the token from
 const connectionConfig = devConfig;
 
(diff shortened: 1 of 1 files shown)


Nateowami:
I have avoided adding command line arguments to scripts because it runs counter to how I find I use the scripts we write. Generally when I need to run a script I find I need to read the code itself in order to understand what I'm really doing (make sure it's the right script, and it's doing precisely what I intend), and I find very often I need to change the script a bit to do what I want.

For example, see `auth0-token-info.ts`, which is essentially the sister of this script (I've used them side-by-side to see the tokens in Auth0 and in Mongo simultaneously, running them with `watch` so I can see the output every couple of seconds). I've needed to modify that script to search for users in a different way than it's set up to do. Because it's a short simple script with very little complexity, it's easy to see what is going on and make changes.

Adding command line arguments makes sense if the requirements for the script are not going to change often, and the parameters are going to change often. But when the requirements change almost as often as the parameters, it seems to me that keeping the script more short and sweet is advantageous, and hard-coding the values is trivially simple if you're editing the script already. I see the scripts more like building blocks to work with than as finished tooling.

There's also the matter of consistency. This script takes only one argument, but `auth0-token-info.ts` has three parameters to set, which would be much more complicated to parse, so it would be even harder to add there (unless we pull in yargs). (Also maybe you start getting into cross-platform issues there, with Linux and Windows handling command line arguments differently? Not sure).

marksvc:
In my experience, I find it frustrating that I need to edit the code of the scripts to simply use them :)
We have lots of nice scripts, and if there's a `text-doc-history.js` or `db-token-info.ts` or `import-project.ts`, I want to be able to run `./text-doc-history.js some-doc-id` or `./db-token-info.ts some-user-identifier` or `./import-projects.ts some-paratext-id some-sf-id`.

I do have `manipulate-sharedb.ts` that I use as a temporary scaffolding to let me use some one-off code to manipulate the sharedb or mongo. But usually with the tools, I want to do something that the script is already designed to do. 

One of the things I find somewhat annoying about the tools that need edited to use is that they show up as modified in git status, they get saved with `git stash`, and/or they need reverted as part of moving from one branch to another.

I have a nice tool that I haven't proposed yet that lets me unlink auth0 accounts from one another. In testing various scenarios, I've run this script many times, and with different arguments depending on whether I'm unlinking PT from an account, on unlinking Google from an account, and for what account I'm unlinking it from. I can easily repeat the right operation by going back to the right command in Terminal history. 

For db-token-info here, I know you feel more comfortable editing the script and so I had made it so that it still allows for use via editing, even tho it now gives me the ability to specify the email by commandline. 
But we don't need to make db-token-info.ts listen for commandline arguments if you prefer.

Nateowami (approved):

PR #1704, opened 2023-02-14
doc: test and refresh tokens

marksvc:
commit-id:9fd57166

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1704)

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,8 +456,14 @@ pkill --parent $(pgrep SIL.XForge.Scri) --full 'node -e module.exports' --signal
 
 ### Tokens
 
+You can fetch your PT tokens from the SF DB with the following command, or manually looking in mongo.
+
+```bash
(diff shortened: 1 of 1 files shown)


marksvc:
(I know we discussed moving docs to the wiki, but it's here right now :)

Nateowami:
@marksvc The wiki is now live. Perhaps create a new page with your suggested changes applied, then change this PR to delete this section?

Also, could you add context to the phrase "your PT tokens". What kind of tokens are these? Are they JWTs? Issued through OAuth? Are they documented somewhere (as in, on a Paratext doc site)? I think there's an access token and a refresh token. I think adding just a little bit of context may be helpful.

marksvc:
@Nateowami I added context and links to its new location at https://github.com/sillsdev/web-xforge/wiki/Testing#tokens
Another PR removes this section from README.md.
Thanks.

PR #1705, opened 2023-02-14
log: report issuedAt time when trouble with token

marksvc:
commit-id:da23732e

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1705)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -2486,7 +2486,7 @@ public class ParatextService : DisposableBase, IParatextService
                 catch
                 {
                     _logger.LogWarning(
-                        $"ParatextService.GetParatextAccessLock for sfUserId {sfUserId} is throwing from call RefreshAccessTokenAsync()."
+                        $"ParatextService.GetParatextAccessLock for sfUserId {sfUserId} is throwing from call RefreshAccessTokenAsync(). The current access token has issuedAt time of {userSecret.ParatextTokens.IssuedAt:o}."
                     );
                     throw;
                 }


marksvc:
Example:
>       ParatextService.GetParatextAccessLock for sfUserId 63c5c8557107a01100f54321 is throwing from call RefreshAccessTokenAsync(). The current access token has issuedAt time of 2023-02-13T21:27:35.0000000Z.

src/SIL.XForge.Scripture/Services/ParatextService.cs
```diff
@@ -2486,7 +2486,7 @@ private async Task<ParatextAccessLock> GetParatextAccessLock(string sfUserId, Ca
                 catch
                 {
                     _logger.LogWarning(
-                        $"ParatextService.GetParatextAccessLock for sfUserId {sfUserId} is throwing from call RefreshAccessTokenAsync()."
+                        $"ParatextService.GetParatextAccessLock for sfUserId {sfUserId} is throwing from call RefreshAccessTokenAsync(). The current access token has issuedAt time of {userSecret.ParatextTokens.IssuedAt:o}."
```
Nateowami, line 2489: Looks good. I was pretty sure I knew what this was doing (ISO 8601), but it takes a couple minutes to find the right terminology and documentation even so. In general I think it would be helpful to introduce language features in code review by writing a review comment something like "The `o` is a [format string](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated#structure-of-an-interpolated-string) for [DateTime](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#table-of-format-specifiers) that converts to string [using ISO 8601](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#the-round-trip-o-o-format-specifier)."

I had some idea what it was already, but if I hadn't, I'm not sure I would have known what to search for.

Alternatively, maybe this is such a fundamental part of C# that it shouldn't really need explaining and I should already know about it. I see we've used it a couple places in our codebase already, but not many.

Nateowami (approved):
Looks good.

src/SIL.XForge.Scripture/Services/ParatextService.cs
marksvc: Haha, actually I didn't write this. I used `.ToString("o")`. It was presumably changed by `dotnet format` at commit time :). 

Sorry for the trouble. I agree that making a note on the PR can be helpful. On the other hand it can be hard to know what deserves a preemptive explanation. 
In this case, I'm not accustomed to this format myself, tho even `.ToString("o")` could have benefited from a note.

BTW, `o`, `u` and `s` are all [date formats](https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings) that may be preferable in different situations.

PR #1707, opened 2023-02-15
Fix project migration tests to prevent new migrations from being run

Nateowami:
Every migration test was running all migrations that were available, rather than just one. I updated the tests so you could specify both the version the document is currently at, and the migration version to stop at.

Probably we will want to do this for the other data migration tests as well, but I want to put this out first and decide on the best solution before changing them all.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1707)

### DIFF
--- a/src/RealtimeServer/scriptureforge/realtime-server.ts
+++ b/src/RealtimeServer/scriptureforge/realtime-server.ts
@@ -12,10 +12,11 @@ import { QuestionService } from './services/question-service';
 import { SFProjectService } from './services/sf-project-service';
 import { SFProjectUserConfigService } from './services/sf-project-user-config-service';
 import { TextService } from './services/text-service';
+import { SF_PROJECT_MIGRATIONS } from './services/sf-project-migrations';
 
 const SF_DOC_SERVICES: DocService[] = [
   new UserService(),
-  new SFProjectService(),
+  new SFProjectService(SF_PROJECT_MIGRATIONS),
   new SFProjectUserConfigService(),
   new TextService(),
   new QuestionService(),
--- a/src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts
+++ b/src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts
@@ -8,6 +8,7 @@ import { createDoc, fetchDoc } from '../../common/utils/test-utils';
 import { SF_PROJECTS_COLLECTION } from '../models/sf-project';
 import { SFProjectRole } from '../models/sf-project-role';
 import { TextInfoPermission } from '../models/text-info-permission';
+import { SF_PROJECT_MIGRATIONS } from './sf-project-migrations';
 import { SFProjectService } from './sf-project-service';
 
 describe('SFProjectMigrations', () => {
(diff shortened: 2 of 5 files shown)


nigel-wells (changes requested):
Just a couple of comments

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1707#-:-NOLxoc08pjnpeg42fYW:b-rap791)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts` line 196 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1707#-NOLxL2O6azwlF5Ndt3y:-NOLxL2O6azwlF5Ndt3z:bh5c9vk) ([raw file](https://github.com/sillsdev/web-xforge/blob/e8944917e3b24e699ba8a082b3d817ca658ab6e9/src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts#L196)):*
> ```XML
>   readonly server: RealtimeServer;
> 
>   constructor(startVersion: number, endVersion: number) {
> ```

As the `endVersion` will 99% of the time be the `startVersion + 1`  how about we make it optional and then set it to `+1` if not defined
___
*[`src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts` line 205 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1707#-NOLxWVhENdhEcCc7vQ5:-NOLxWVhENdhEcCc7vQ6:bxl2rye) ([raw file](https://github.com/sillsdev/web-xforge/blob/e8944917e3b24e699ba8a082b3d817ca658ab6e9/src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts#L205)):*
> ```XML
>       'TEST',
>       false,
>       [new SFProjectService(SF_PROJECT_MIGRATIONS.slice(0, endVersion))],
> ```

Wouldn't it make sense to just pass in the tests between `startVersion` and `endVersion` otherwise the more migrations we create still has to pass all the previous migrations which isn't relevant as the test as already indicated which test to start on.

nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1707)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts` line 205 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1707#-NOLxWVhENdhEcCc7vQ5:-NOLysYtC9GrLwb4YCrp:b885ydq) ([raw file](https://github.com/sillsdev/web-xforge/blob/e8944917e3b24e699ba8a082b3d817ca658ab6e9/src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts#L205)):*

Wouldn't it make sense to just pass in the tests between `startVersion` and `endVersion` otherwise the more migrations we create still has to pass all the previous migrations which isn't relevant as the test as already indicated which test to start on.
</blockquote></details>

I wish there was an edit comment button in here. I meant to say "migrations" not "tests".

nigel-wells (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1707)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1708, opened 2023-02-15
SF-1877 Set the ext user attribute on comments when needed

RaymondLuong3:
* remove extUserId from data model
* set ext user when the note is created by a non-paratext user

The external user field on a Paratext comment tracks the ID of a user that authored a comment but is not a PT user. PT users leaving notes in SF should not have their external user field set on their comments. Additionally, the extUserId in the SF model is useless since this never changes once a comment is created. So I've removed the property from the data model.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1708)


marksvc (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1708)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/services/note-thread-migrations.ts` line 32 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1708#-NOQTZ228HYzGDndSgoo:-NOQTZ23Bo57l9nCyW8S:b-fikcrw) ([raw file](https://github.com/sillsdev/web-xforge/blob/9fc58e6f4fc5b0c3a4d072b9fb2d70e087eff5cf/src/RealtimeServer/scriptureforge/services/note-thread-migrations.ts#L32)):*
> ```XML
>     for (let i = 0; i < doc.data.notes.length; i++) {
>       if (doc.data.notes[i].extUserId != null) {
>         ops.push({ p: ['notes', i, 'extUserId'], od: true });
> ```

So here you are specifying an arbitrary and dummy value to `od`, right? Because `true` isn't what the value will be, but you just needed to specify something to `od`, and so you chose `true`; is that right?
I think the [delete op](https://github.com/ottypes/json0#inserting-deleting-and-replacing-in-an-object) is supposed to have the current value passed to `od`, so here it might be something like `od: '6cabc12341234231432'`. That may help with going back and forth in OT history. I'm not super familiar with how important that is. I know the actual value doesn't need to be specified when I am changing data on my workstation for testing. 

Hmm, I see that in the migration above, `od: true` was used. But I think the right way to do it is to pass the value. Do you know more about this?



_Code quote:_
```XML
 od: true 
```
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs` line 1916 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1708#-NOv27C97QNO-3TfLWox:-NOv27C97QNO-3TfLWoy:b-r3ikb5) ([raw file](https://github.com/sillsdev/web-xforge/blob/9fc58e6f4fc5b0c3a4d072b9fb2d70e087eff5cf/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L1916)):*
> ```Smalltalk
>         Assert.That(thread.Comments.Count, Is.EqualTo(1));
>         comment = thread.Comments.First();
>         // expect the non-paratext ext user to be user04
> ```

Thank you for helpful commentary :)



_Code quote:_
```Smalltalk
// expect the non-paratext ext user to be user04
```
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs` line 1927 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1708#-NOv2hBQ8m-6G9o4BA2j:-NOv2hBQ8m-6G9o4BA2k:b-hvffnp) ([raw file](https://github.com/sillsdev/web-xforge/blob/9fc58e6f4fc5b0c3a4d072b9fb2d70e087eff5cf/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L1927)):*
> ```Smalltalk
>         Assert.That(ptProjectUsers.Keys, Is.EquivalentTo(new[] { env.Username01, env.Username02 }));
>         IDocument<NoteThread> noteThreadDoc = noteThreadDocs.First(d => d.Data.DataId == thread1Id);
>         Assert.That(noteThreadDoc.Data.Notes[0].SyncUserRef, Is.EqualTo("syncuser02"));
> ```

And I'm thinking the SyncUserRef for thread2 note 1 is going to be something like "syncuser01",  would that be right?



_Code quote:_
```Smalltalk
 Assert.That(noteThreadDoc.Data.Notes[0].SyncUserRef, Is.EqualTo("syncuser02"));
```

marksvc:
In some discussion, it was pointed out that the Note OwnerRef and ExtUserId are redundant, and that our Note OwnerRef contains the information.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1708)* status: 10 of 13 files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[`src/RealtimeServer/scriptureforge/services/note-thread-migrations.ts` line 32 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1708#-NOQTZ228HYzGDndSgoo:-NOv8qOU5AuiyKw4FYCT:b8zubi5) ([raw file](https://github.com/sillsdev/web-xforge/blob/9fc58e6f4fc5b0c3a4d072b9fb2d70e087eff5cf/src/RealtimeServer/scriptureforge/services/note-thread-migrations.ts#L32)):*

So here you are specifying an arbitrary and dummy value to `od`, right? Because `true` isn't what the value will be, but you just needed to specify something to `od`, and so you chose `true`; is that right?
I think the [delete op](https://github.com/ottypes/json0#inserting-deleting-and-replacing-in-an-object) is supposed to have the current value passed to `od`, so here it might be something like `od: '6cabc12341234231432'`. That may help with going back and forth in OT history. I'm not super familiar with how important that is. I know the actual value doesn't need to be specified when I am changing data on my workstation for testing. 

Hmm, I see that in the migration above, `od: true` was used. But I think the right way to do it is to pass the value. Do you know more about this?
</blockquote></details>

Thanks for the suggestion. I wasn't sure if the value for `od` was important, so I was just using the `true` as a working placeholder. I've updated it to include the actual value of the `extUserId`
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs` line 1927 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1708#-NOv2hBQ8m-6G9o4BA2j:-NOvHKPvCk9VlLbcetO-:b37qu2b) ([raw file](https://github.com/sillsdev/web-xforge/blob/9fc58e6f4fc5b0c3a4d072b9fb2d70e087eff5cf/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L1927)):*

And I'm thinking the SyncUserRef for thread2 note 1 is going to be something like "syncuser01",  would that be right?
</blockquote></details>

I added a check to show this, but I needed to modify the input `ptProjectUsers`. Since we mock the `IGuidService` to start at 2, the value would actually be `syncuser03`.

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1708)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1710, opened 2023-02-15
SF-1849 Remove secondary linked account paratextId and tokens

marksvc:
When a password- or google-authenticated user Connects a project, they
first need to log in with Paratext. This creates a new auth0 account,
that is linked/merged into the original auth0 account. The original SF
account then has a paratextId and Paratext tokens set.

In certain instances, another SF account is created for the new auth0
account as well. This new SF account also gains the paratextId and
Paratext tokens. This SF account can not be logged into because the
auth0 id on the account is for the new auth0 account that was
linked/merged into the original auth0 account. But the new SF account
now causes a problem because two SF accounts have the same paratextId.

It looks as though there may have been some unfortunate timing for the
problem instance, where auth0 calls pushAuthUserProfile and SF creates
an account for the new auth0, before auth.service handleOnlineAuth
finishes a call to linkParatextAccount.

This secondary SF user can probably be safely deleted in
LinkParatextAccountAsync(). We know the secondary account has only
been logged into once, and so would not have other associated data.
However, because of the difficulty in clearly identifying the problem
at hand, I am removing paratextId and the Paratext tokens, and leaving
the user doc and user_secret doc. If we find that the solution at hand
is in fact a fix for the problem, we can change the code so it deletes
the secondary SF user and user_secret docs.

commit-id:84793324

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1710)


src/SIL.XForge/Services/UserService.cs
```diff
@@ -151,6 +155,23 @@ public async Task LinkParatextAccountAsync(string primaryAuthId, string paratext
...
+            );
+            await secondaryUserDoc.SubmitJson0OpAsync(op => op.Set(u => u.ParatextId, null));
+        }
+
+        if (await _userSecrets.GetAsync(secondarySFUserId) != null)
+        {
+            await _userSecrets.UpdateAsync(
+                secondarySFUserId,
+                update => update.Set(us => us.ParatextTokens, null),
+                true
```
pmachapman, line 172: You have retrieved the object in line 167 to check its existence, but here you are upserting in case it does not exist? I would have thought that if the user secret record did not exist (i.e. was somehow deleted by another action on the database at the same time), you would not want to insert the null value?

pmachapman (changes requested):

src/SIL.XForge/Services/UserService.cs
marksvc: Thank you! By leaving upset set to true, I was creating my own need to first check the existence. Leaving upsert as false is much better.

pmachapman (approved):
Looks good - approved!

PR #1711, opened 2023-02-15
tool: allow running manipulate-sharedb.ts from another dir

marksvc:
commit-id:952d2bce

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1711)


pmachapman (approved):
That is the coolest hashbang I have seen in a while. It also works great for me.

PR #1713, opened 2023-02-16
tool: add auth0-manage.ts

marksvc:
commit-id:44faa3d5

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1713)

### DIFF
--- a/scripts/auth0-manage.ts
+++ b/scripts/auth0-manage.ts
@@ -1,0 +1,89 @@
+#!/usr/bin/env -S deno run --allow-net --allow-env=AUTH0_TOKEN
+
+const usage: string = `Usage:
+At present, this script only interacts with the development Auth0 tenant.
+
+First: export AUTH0_TOKEN="eyAbCd..."
+Your Auth0 authorization bearer token can be found in the Auth0 Dashboard at
+Applications - APIs - Auth0 Management API - API Explorer - Token.
+
+Unlink auth0 accounts:
+  ./auth0-manage.ts unlink --unlinkSecondaryUserId 'paratext|ABCABC' --fromPrimaryUserId 'auth0|12341234' --provider 'oauth2'
+
+Delete auth0 account:
+  ./auth0-manage.ts delete --userId 'oauth2|paratext|ABCABC'
+`;
+
+import { parser } from 'https://deno.land/x/args_command_parser@v1.2.4/mod.js';
+
+class Program {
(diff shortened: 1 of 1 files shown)


scripts/auth0-manage.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+
+  async main() {
+    if (this.args.longSwitches['help'] != null) {
+      this.usage();
+    }
+
+    this.authorizationBearerToken =
+      Deno.env.get('AUTH0_TOKEN') ??
+      this.fail("specify auth0 authorization bearer token as --auth, such as 'eyAbCd...'");
```
pmachapman, line 71: This message isn't correct - should it refer to the environment variable?

scripts/auth0-manage.ts
```diff
@@ -0,0 +1,84 @@
...
+    );
+    console.log(response);
+  }
+
+  async delete() {
+    const userId =
+      this.args.longSwitches['userId'] ?? this.fail("specify --userId to delete, such as 'oauth2|paratext|ABCABC'");
+
+    // API https://auth0.com/docs/api/management/v2#!/Users/delete_users_by_id
+    const response = await fetch(`https://sil-appbuilder.auth0.com/api/v2/users/${userId}`, {
```
pmachapman, line 57: Is it OK that this only works on our non-production Auth0?

scripts/auth0-manage.ts
pmachapman, line 71: This message is incorrect - it should reference the environment variable.

Do you also want to include the usage after the failure message?

pmachapman (changes requested):

scripts/auth0-manage.ts
marksvc: Maybe. There may come a time when we find it would be helpful to have a tool like this that can unlink on the production Auth0 tenant. For now, I only needed to modify accounts on the dev tenant, and so I also haven't done any testing of this script against the production tenant. So I think it's okay for the script to just use the dev tenant for now, and we can enhance it in the future if the desire arises. 

I added a note to the Usage message about it only interacting with the dev tenant.

scripts/auth0-manage.ts
marksvc: Thank you :)
Good idea.

pmachapman (approved):

PR #1714, opened 2023-02-17
Fix date format tests for Chromium 110

Nateowami:
Chromium now uses a NARROW NO-BREAK SPACE between the minutes and the am/pm designation

It would probably behoove us to merge this sooner rather than later, because it's likely that our builds are going to start failing at any moment when GitHub Actions starts using Chromium 110 for the builds (which may be now already).

(For what it's worth, I don't actually know that it's Chromium itself; it could be a localization package that Chromium depends on that got upgraded on the system, but I think Chromium tends to ship such functionality internally rather than depend on system libraries)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1714)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.spec.ts
@@ -92,10 +92,12 @@ describe('I18nService', () => {
   it('should localize dates', () => {
     const date = new Date('November 25, 1991 17:28');
     const service = getI18nService();
-    expect(service.formatDate(date)).toEqual('Nov 25, 1991, 5:28 PM');
+    // As of Chromium 110 the space between the minutes and AM/PM has been changed to U+202F (NARROW NO-BREAK SPACE)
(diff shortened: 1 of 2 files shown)


pmachapman (commented):
I notice that Chromium have reverted this change because it broke similar code for too many people. I just updated my Chrome to the latest 110, and the date string is as it used to be.

See https://bugs.chromium.org/p/chromium/issues/detail?id=1414292 for details

If you still want to go ahead with the fix, I can approve it, but I don't think there is a need to?

pmachapman (approved):
My PR builds have started failing from this issue, so I think it is a good idea to merge it, at least until the Chromium update rolls out more widely.

Nateowami:
Interesting. I didn't look into the change to Chromium. We'd already had changes to the zh-CN format cause failures in the past, so I wasn't terribly surprised when we ran into this.

Also, I amended the commit because I found a typo. I had written:
> the space between the minutes and AM/PM is **not** a NARROW NO-BREAK SPACE

Which should have said:
> the space between the minutes and AM/PM is **now** a NARROW NO-BREAK SPACE

PR #1715, opened 2023-02-17
SF-1880 Prompt before deleting comment

Nateowami:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1715)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html
@@ -24,7 +24,7 @@
           *ngIf="canDeleteComment(comment)"
           mdc-button
           type="button"
-          (click)="deleteComment(comment)"
+          (click)="deleteCommentClicked(comment)"
           class="comment-delete"
         >
           {{ t("delete") }}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.ts
@@ -8,6 +8,7 @@ import { Comment } from 'realtime-server/lib/esm/scriptureforge/models/comment';
 import { SFProjectProfile } from 'realtime-server/lib/esm/scriptureforge/models/sf-project';
 import { SFProjectDomain, SF_PROJECT_RIGHTS } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-rights';
 import { debounceTime } from 'rxjs/operators';
+import { DialogService } from 'xforge-common/dialog.service';
 import { SubscriptionDisposable } from 'xforge-common/subscription-disposable';
 import { UserService } from 'xforge-common/user.service';
 import { QuestionDoc } from '../../../../core/models/question-doc';
@@ -38,7 +39,7 @@ export class CheckingCommentsComponent extends SubscriptionDisposable implements
   showAllComments: boolean = false;
   private initUserCommentRefsRead: string[] = [];
 
-  constructor(private userService: UserService) {
+  constructor(private readonly dialogService: DialogService, private userService: UserService) {
     super();
   }
 
@@ -79,17 +80,19 @@ export class CheckingCommentsComponent extends SubscriptionDisposable implements
     return this.answer != null ? sortBy(this.answer.comments, c => c.dateCreated) : [];
   }
 
-  editComment(comment: Comment) {
+  editComment(comment: Comment): void {
     this.activeComment = cloneDeep(comment);
     this.showCommentForm();
   }
 
-  deleteComment(comment: Comment) {
-    this.action.emit({
-      action: 'delete',
-      answer: this.answer,
-      comment: comment
-    });
+  async deleteCommentClicked(comment: Comment): Promise<void> {
+    if (await this.dialogService.confirm('checking_comments.confirm_delete', 'checking_comments.delete')) {
+      this.action.emit({
+        action: 'delete',
+        answer: this.answer,
+        comment: comment
+      });
+    }
   }
 
   canEditComment(comment: Comment): boolean {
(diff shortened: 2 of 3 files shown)


nigel-wells (approved):
Just one point that I'm not overly fussed about - your call.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1715#-:-NOfb2oi682H4eeQBFPO:b-5ssqdi)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1715#-NOfadR7-wawZ33sB7Qa:-NOfadR7-wawZ33sB7Qb:b-as2ycl) ([raw file](https://github.com/sillsdev/web-xforge/blob/5d4d8aa2ca71c4bdc0f5d24cd0828149105662cc/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L27)):*
> ```HTML
>           mdc-button
>           type="button"
>           (click)="deleteCommentClicked(comment)"
> ```

Do you think we need to rename this to include `Clicked`? I think it is fine without and then it leave the method available to be called outside of `Clicked` event without feeling weird about it.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1715)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1715#-NOfadR7-wawZ33sB7Qa:-NOjgVMh1QA_57FyLnm1:b-3cnf2v) ([raw file](https://github.com/sillsdev/web-xforge/blob/5d4d8aa2ca71c4bdc0f5d24cd0828149105662cc/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L27)):*

Do you think we need to rename this to include `Clicked`? I think it is fine without and then it leave the method available to be called outside of `Clicked` event without feeling weird about it.
</blockquote></details>

I don't like to have a method named `deleteComment()` that doesn't actually delete a comment (it prompts the user to ask if the comment should be deleted). If we wanted to call it elsewhere, not due to the user clicking the button, then I would question whether we want to show the prompt or not in that situation. We might want to refactor to have a method called `deleteComment()` and a method called `deleteCommentClicked()`, where `deleteCommentClicked()` is essentially:
``` ts
if (await this.dialogService.confirm(...)) this.deleteComment();
```
We have the same issue in a lot of other components (pretty much any time we want to confirm anything), so if there's a better convention for naming things I'm all for it.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1715)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1715#-NOfadR7-wawZ33sB7Qa:-NOkdEae3IWFtTEYyZ_6:b-mcu4wh) ([raw file](https://github.com/sillsdev/web-xforge/blob/5d4d8aa2ca71c4bdc0f5d24cd0828149105662cc/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L27)):*

I don't like to have a method named `deleteComment()` that doesn't actually delete a comment (it prompts the user to ask if the comment should be deleted). If we wanted to call it elsewhere, not due to the user clicking the button, then I would question whether we want to show the prompt or not in that situation. We might want to refactor to have a method called `deleteComment()` and a method called `deleteCommentClicked()`, where `deleteCommentClicked()` is essentially:
``` ts
if (await this.dialogService.confirm(...)) this.deleteComment();
```
We have the same issue in a lot of other components (pretty much any time we want to confirm anything), so if there's a better convention for naming things I'm all for it.
</blockquote></details>

I personally see `deleteComment` being fine and it just happens to contain a safe guard to the action of deleting. If anything else wanted to delete a comment it should also show any safe guards before deleting. I do see your point though. Perhaps the real `deleteComment` - the one that actually deletes the comment - should be a private method. We can then have one called `confirmDeleteComment`. This way templates can only call the public method requiring confirmation but we have the flexibility within the component to also directly delete the comment without confirmation.

I realize this is overkill for what you're doing and adding code for something that may or may not happen. Good discussion though :)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1715)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1715#-NOfadR7-wawZ33sB7Qa:-NOvhehKAKDtKGJuYeqE:b-xl59na) ([raw file](https://github.com/sillsdev/web-xforge/blob/5d4d8aa2ca71c4bdc0f5d24cd0828149105662cc/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L27)):*

I personally see `deleteComment` being fine and it just happens to contain a safe guard to the action of deleting. If anything else wanted to delete a comment it should also show any safe guards before deleting. I do see your point though. Perhaps the real `deleteComment` - the one that actually deletes the comment - should be a private method. We can then have one called `confirmDeleteComment`. This way templates can only call the public method requiring confirmation but we have the flexibility within the component to also directly delete the comment without confirmation.

I realize this is overkill for what you're doing and adding code for something that may or may not happen. Good discussion though :)
</blockquote></details>

I see what you mean. An example of why I think this could be confusing is something similar to the checking overview component. In that component I originally started writing something like:
``` ts
archiveQuestionsInBook(book: number) {
  if (userConfirms) {
    for (const chapter of book) {
      archiveQuestionsInChapter() // but this will open a confirmation for each chapter
    }
  }
}
```
However, I was also confirming if the user wanted to archive all the questions in the chapter, so that didn't work. I ended up just nesting a loop in a loop in the method for archiving all questions in a book.

I also ended up naming the methods not very well, because they had to be able to both archive and republish, but that's kind of beside the point.

PR #1717, opened 2023-02-22
Added support for batch migrating ShareDB documents

pmachapman:
This pull request:
 * Performs migrations on batches of 10,000 documents
 * Sends a log message when each batch is processed
 * Suppresses the CodeQL warning "_Unexpected any. Specify a different type_" in realtime-server.ts
 * Disables the Jering invocation timeout when a debugging parameter (`--inspect` or `--inspect-brk`) is specified for NodeJS
    - The primary purpose of this is to stop debugging pauses or huge collections from causing an invocation timeout
    - This can also be used to give the very long execution time required for some migrations, by specifying `--inspect` in the `node-options` environment variable, which will ensure that the migration process is not interrupted by Jering timing out the request to `Server.Start()`.

**Problem Summary**

When a collection has hundreds of thousands of documents, and a migration is run on the collection, the following crash can occur in the Realtimeserver on startup:

>   FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory

To fix this error, migration of documents now takes place in batches of 10,000 with a mechanism (by specifying `--inspect`) to ensure that the execution of `Server.Start()` takes as long as it needs to running the migrations.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1717)

### DIFF
--- a/src/SIL.XForge/Realtime/RealtimeServiceCollectionExtensions.cs
+++ b/src/SIL.XForge/Realtime/RealtimeServiceCollectionExtensions.cs
@@ -28,6 +28,13 @@ public static class RealtimeServiceCollectionExtensions
                 options.NodeAndV8Options = nodeOptions;
             }
         });
+
+        // Disable timeout so the debugger can be paused
+        if (nodeOptions?.Contains("--inspect", StringComparison.OrdinalIgnoreCase) == true)
+        {
+            services.Configure<OutOfProcessNodeJSServiceOptions>(options => options.TimeoutMS = -1);
+        }
+
         services.AddSingleton<IJsonService, RealtimeJsonService>();
 
         services.Configure(configureOptions);
--- a/src/RealtimeServer/common/realtime-server.ts
+++ b/src/RealtimeServer/common/realtime-server.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import { Db } from 'mongodb';
 import ShareDB from 'sharedb';
 import shareDBAccess from 'sharedb-access';
@@ -234,15 +235,32 @@ export class RealtimeServer extends ShareDB {
       if (curVersion === version) {
         continue;
       }
-      const query = await createFetchQuery(this.defaultConnection!, docService.collection, {});
-      while (curVersion < version) {
-        curVersion++;
-        const promises: Promise<void>[] = [];
-        const migration = docService.getMigration(curVersion);
-        for (const doc of query.results) {
-          promises.push(migration.migrateDoc(doc));
+      const limit = 10000;
+      let skip = 0;
+      let query = await createFetchQuery(this.defaultConnection!, docService.collection, {
+        $sort: { _id: 1 },
+        $skip: skip,
+        $limit: limit
+      });
+      while (query.results.length > 0) {
+        console.log(`Migrating ${docService.collection}: ${skip + 1} to ${skip + query.results.length}`);
+        let docVersion = curVersion;
+        while (docVersion < version) {
+          docVersion++;
+          const promises: Promise<void>[] = [];
+          const migration = docService.getMigration(docVersion);
+          for (const doc of query.results) {
+            promises.push(migration.migrateDoc(doc));
+          }
+          await Promise.all(promises);
         }
-        await Promise.all(promises);
+
+        skip += limit;
+        query = await createFetchQuery(this.defaultConnection!, docService.collection, {
+          $sort: { _id: 1 },
+          $skip: skip,
+          $limit: limit
+        });
       }
 
       await this.schemaVersions.set(docService.collection, version);


src/SIL.XForge/Realtime/RealtimeServiceCollectionExtensions.cs
```diff
@@ -28,6 +28,13 @@ public static class RealtimeServiceCollectionExtensions
                 options.NodeAndV8Options = nodeOptions;
             }
         });
+
+        // The debugger requires the timeout to be disabled
```
marksvc, line 32: It may obscure what is being done to say the debugger _requires_ the timeout to be disabled (eg maybe `--inspect` is being confused for `--inspect-brk`). If my other comment about this is accurate, what do you think about saying something like
> // Disable timeout so the debugger can be paused

marksvc:
I think the migration change looks fine. I am not particularly experienced in this area, but I don't see anything wrong with the change. 
I notice that this migrates groups of documents all the way to the latest version before moving on to the next group. Rather than getting all the documents (in groups) migrated one version at a time. But I'm not aware that there would be a problem with that.

I was glad to read and learn about differences between StringComparison.Ordinal and InvariantCulture.

> Disables the Jering invocation timeout when a debugging parameter (--inspect or --inspect-brk) is specified for NodeJS
>
>    The primary purpose of this is to stop debugging pauses [...]

So my understanding is that although `--inspect-brk` will pause the node startup until a debugger is attached, `--inspect` will not, and so it's not necessary to disable the timeout because dotnet will not be waiting for node to start up (while node is waiting for the developer to connect their debugger). 

But it sounds like a potential problem still is that even with just `--inspect`, if a developer pauses the node process using their debugger, then _while NodeJS is paused_, the timeout can be reached and there can be a problem. And so by turning off the timeout when `--inspect` is specified, NodeJS can stay paused by the developer's debugger as long as needed, and Jering will not have a problem with it. Is this so?

> or heap collections from causing an invocation timeout

Are heap collections that cause a timeout something we will expect to happen sometimes from NodeJS running. Because we might we be better off by increasing the timeout to a value that we think will handle our Realtime Server's needs. But from other following notes in the PR, maybe you meaning this would be an exceptional and rare situations, where we will just want to manually pass `--inspect` anyway?

marksvc:
@Nateowami I'm still learning about migrations, but FWIW I approve, other than a quibble over a comment wording.

RaymondLuong3 (approved):
I was able to do some testing on the note thread collection with a database of 2259 documents. It seems to happily migrate my database before the realtime server is started. I don't have anything else to add other than Mark's comments.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1717#-:-NOu9aPF6wwJ2NKVMAHM:b-m3ok8z)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/RealtimeServer/common/realtime-server.ts` line 245 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1717#-NOu9Bg47yUNmknc1Tsa:-NOu9Bg47yUNmknc1Tsb:b-elm7rb) ([raw file](https://github.com/sillsdev/web-xforge/blob/2410fc1ce3467e2d3dac00d39a3fe2203c3858c3/src/RealtimeServer/common/realtime-server.ts#L245)):*
> ```XML
>         $limit: limit
>       });
>       while (query.results.length > 0) {
> ```

I was able to see the results running when I reduced the limit to 500. It was about to process my 2259 documents in about 2 seconds. Nice.



_Code quote:_
```XML
   while (query.results.length > 0) {
```

src/SIL.XForge/Realtime/RealtimeServiceCollectionExtensions.cs
pmachapman: That is much better wording - thank you!

pmachapman:
> And so by turning off the timeout when --inspect is specified, NodeJS can stay paused by the developer's debugger as long as needed, and Jering will not have a problem with it. Is this so?

@marksvc Yes, that is my understanding. I can make it `--inspect-brk` only - my main rational for making it broader (i.e. both `--inspect` and `--inspect-brk`) is that it is an argument that should never be a part of a standard production deploy, so will not be triggered accidentally as a part of a normal config. And the times you do want it, you would really want the timeout to be disabled.

> Are heap collections that cause a timeout something we will expect to happen sometimes from NodeJS running.

@marksvc Sorry for the confusion, that is a typo - I meant **huge** collections. And it will happen very, very, rarely. Apart from the migrations, I have witnessed it once in bugsnag, and there were other contributing factors, so I believe it is acceptable to have very rare occurrences of a timeout. The default timeout is 60 seconds, which appears to be adequate based on the exception reports in bugsnag.

PR #1719, opened 2023-02-28
SF-1878b Highlight note thread only if unread notes exist

RaymondLuong3:
When a note gets inserted, the scripture editor responds to the new note too fast sometimes, and the note appears in the editor before our system has had the chance to mark it read. To prevent this race-condition from happening, we have to explicitly wait for the note ID to be recorded in the projectUserConfigDoc before saving the note.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1719)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -3573,9 +3573,10 @@ class TestEnvironment {
     const note = this.targetTextEditor.querySelector(`usx-segment display-note[data-thread-id=${threadId}]`);
     expect(note).withContext('note thread highlight').not.toBeNull();
     const thread: HTMLElement | null = this.targetTextEditor.querySelector(
-      `usx-segment display-note[data-thread-id=${threadId}].note-thread-highlight`
+      `usx-segment display-note[data-thread-id=${threadId}]`
     );
-    return thread != null;
+    expect(thread).withContext('note thread highlight').not.toBeNull();
+    return thread!.classList.contains('note-thread-highlight');
   }
 
   openNoteDialogAndEdit(segmentRef: string, thread: NoteThread): void {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -1535,8 +1535,9 @@ describe('EditorComponent', () => {
       env.dispose();
     }));
 
-    it('shows highlights note icons when new content is unread', fakeAsync(() => {
+    it('highlights note icons when new content is unread', fakeAsync(() => {
       const env = new TestEnvironment();
+      env.setCurrentUser('user02');
       env.setProjectUserConfig({ noteRefsRead: ['thread01_note0', 'thread02_note0'] });
       env.wait();
 
@@ -1553,7 +1554,7 @@ describe('EditorComponent', () => {
       let iconElement: HTMLElement = env.getNoteThreadIconElement('verse_1_1', 'thread01')!;
       iconElement.click();
       env.wait();
-      puc = env.getProjectUserConfigDoc('user01');
+      puc = env.getProjectUserConfigDoc('user02');
       expect(puc.data!.noteRefsRead).toContain('thread01_note1');
       expect(puc.data!.noteRefsRead).toContain('thread01_note2');
       expect(env.isNoteIconHighlighted('thread01')).toBe(false);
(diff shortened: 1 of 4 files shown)


Nateowami:
@RaymondLuong3 Isn't the editor listening to changes to projectUserConfigDoc, so it will only briefly show it as unread (presumably for a moment that is briefer than is perceptible)? Alternatively, might we change the logic for determining whether a note is read to consider who the author is?

RaymondLuong3:
That is a great idea! I'm not sure why I didn't think of it. Hmm, I'll turn this into a draft.

RaymondLuong3:
@Nateowami This is ready for re-review when you have the chance.

src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
```diff
@@ -3469,8 +3495,10 @@ class TestEnvironment {
   }
 
   isNoteIconHighlighted(threadId: string): boolean {
+    const note = this.targetTextEditor.querySelector(`usx-segment display-note[data-thread-id=${threadId}]`);
+    expect(note).withContext('note thread highlight').not.toBeNull();
     const thread: HTMLElement | null = this.targetTextEditor.querySelector(
-      `usx-segment display-note[data-thread-id="${threadId}"].note-thread-highlight`
+      `usx-segment display-note[data-thread-id=${threadId}].note-thread-highlight`
     );
     return thread != null;
```
Nateowami, line 3503: I think this would be a much better test if you selected the thread and then checked whether the thread has the `note-thread-highlight` class. As it exists now, if your selector was wrong you would get a negative result, and getting selectors wrong is really easy (especially if someone comes and edits the tests later.)

Maybe something like this:
``` suggestion
    const thread: HTMLElement | null = this.targetTextEditor.querySelector(
      `usx-segment display-note[data-thread-id=${threadId}]`
    );
    expect(thread).withContext('note thread highlight').not.toBeNull();
    return thread!.classList.contains('note-thread-highlight');
```

src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
```diff
@@ -1637,11 +1637,13 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy,
...
-      return false;
-    }
-    return !this.projectUserConfigDoc.data.noteRefsRead.includes(noteId);
+    if (thread.data == null || this.projectUserConfigDoc?.data == null) return false;
+    // look for any note that has not been read and was authored by another user
+    const noteRefsRead: string[] = this.projectUserConfigDoc.data.noteRefsRead;
+    const unreadNoteIndex: number = thread.data.notes.findIndex(
+      n => n.ownerRef !== this.userService.currentUserId && !noteRefsRead.includes(n.dataId)
+    );
+    return unreadNoteIndex > -1;
```
Nateowami, line 1646: How about using `some()`:
``` suggestion
    return thread.data.notes.some(
      n => n.ownerRef !== this.userService.currentUserId && !noteRefsRead.includes(n.dataId)
    );
```

Nateowami (changes requested):
I'm sorry I just discovered I forgot about this. I can see that the highlight is now no longer briefly shown.

I have a couple suggested improvements.

src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
RaymondLuong3: Great thinking. I've made these changes.

src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
RaymondLuong3: I like it. I had thought I tried the some() method, but I must not have.

Nateowami (approved):
> Reviewed 2 of 4 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1719)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3503 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1719#-NPOxt4YbAvRff3XLmak-r1-3503:-NRdmvUf-ZkuMm9G3wtm:b-bna9xi) ([raw file](https://github.com/sillsdev/web-xforge/blob/e29c2938167aaa357421a31515c1f04c571ca0e7/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3503)):*

Great thinking. I've made these changes.
</blockquote></details>

I'm not sure why my suggestion included `.withContext('note thread highlight')`; it doesn't seem to be particularly helpful.

PR #1728, opened 2023-03-01
SF-1892 Remove displayName from ProjectRoleInfo

Nateowami:
The display name for a role is now stored in our localization files, and that is the only place we should be getting it from. Having it elsewhere is just a source of potential trouble.

Unfortunately the system administration page was still using the old displayName, so it needed some changes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1728)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/sf-project-role-info.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/sf-project-role-info.ts
@@ -4,17 +4,13 @@ import { isTranslateRole, SFProjectRole } from 'realtime-server/lib/esm/scriptur
 import { ProjectRoleInfo } from 'xforge-common/models/project-role-info';
 
 export const SF_PROJECT_ROLES: ProjectRoleInfo[] = [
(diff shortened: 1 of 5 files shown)


pmachapman (changes requested):
I ran this change, and noticed that the translation string for `pt_observer` and `sf_observer` is both "Observer":
image

Do we want to display the `role_descriptions` string in the dropdown below the role name, like in `share-control.component.html`?

image

Another option, since this is an area used by technical users, we could include the name of the role (i.e. sf_observer), if the role description doesn't work.

Nateowami (commented):
That's a good point; we need to figure out how to differentiate them. Though this problem is pre-existing; on the users page the ambiguity is already present.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1728#-:-NPyvjzM4CUg3B2lnbiA:brs2wnv)* status: 0 of 5 files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (approved):
Am happy to approve this, as we are aware of the pre-existing bug which affects other areas, and so should be fixed separately.

PR #1731, opened 2023-03-07
SF-1881 Biblical Terms Support

pmachapman:
This Pull Request:
 * Bundles the Biblical Terms XML Files
 * Adds, updates, and deletes the Biblical Terms in SF from Paratext on Sync
 * Displays Biblical Terms on the frontend
 * Adds, updates, and deletes the Biblical Term's Notes in SF from Paratext on Sync
 * Displays Biblical Term Notes

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1731)


pmachapman (commented):
> Dismissed @github-code-scanning[bot] from 8 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1731)* status: 0 of 127 files reviewed, all discussions resolved (waiting on @github-code-scanning[bot])

josephmyers:
As part of my review, I just finished running through the Acceptance Tests for this. All look good, except #12, which doesn't appear to be working. I can't get any notes to be saved or synced. See below screenshots. As a side note, after syncing, other things with notes are broken until you refresh the page (e.g. the note icon doesn't update when adding notes, and the scrolling on the right side is jittering all over the place).
image
image

josephmyers:
I've got another, though slightly less important issue. On mobile, biblical terms gets collapsed on page reload. See video:
https://github.com/sillsdev/web-xforge/assets/43733878/61b21f9e-51cf-42f9-b2ab-d150221a2e9f

josephmyers:
The two new settings in Translator Settings are cut off on mobile (Pixel 5 is what I always use)
image

pmachapman (commented):
I have fixed the note writing to Paratext, but you will need to enable a Feature Flag for this to work:
```sh
cd src/SIL.XForge.Scripture
dotnet user-secrets set "FeatureManagement:WriteNotesToParatext" true
```
Then restart Scripture Forge.

Regarding the Biblical Terms collapsing, this is because Biblical Terms are collapsed by default on mobile (due to space constraints).

For the terms being cut off, I am not sure what we can do given that Translation suggestions is cut off too?

Finally,  I have tried to reduce the scrollbar appearing/jittering with the add note fab. Please let me know if it is still an issue?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-:-NbzvkpY0mGULc7-frdq:bkuoocb)* status: 0 of 128 files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (changes requested):
Thanks for the fixes so far. This review was a beast. If possible in the future, let's try to split things out into more bitesize pieces! Your reviewers will appreciate it :D

I get that we're constrained when on mobile, but I don't think that resizing the pane to 0 is a good option. In my opinion, I don't think a refresh justifies resetting the user's layout. Now, I think another option could be to use a docked, toggleable pane, a la your favorite IDE.

For terms being cut off, I believe it's as simple as CSS overflow-wrap/word-wrap

> Reviewed 18 of 33 files at r1, 3 of 14 files at r3, 1 of 53 files at r4, 11 of 81 files at r5, 4 of 21 files at r6, 4 of 41 files at r7, 2 of 18 files at r8, 9 of 53 files at r9, 2 of 12 files at r10, 50 of 62 files at r11, 21 of 22 files at r12, 10 of 10 files at r13, 3 of 3 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-:-Nc0ocAPDYeRGxhFP0y5:b-c3z1iw)* status: all files reviewed, 22 unresolved discussions (waiting on @pmachapman)

___
*[`src/RealtimeServer/scriptureforge/models/biblical-term.ts` line 19 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxpBg8A_yht2A6oru5:-NbxpBg8A_yht2A6oru6:b-susqhf) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/RealtimeServer/scriptureforge/models/biblical-term.ts#L19)):*
> ```XML
>   language: string;
>   links: string[];
>   references: number[];
> ```

Can we not store this as a VerseRef?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.html` line 92 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxJITN1cnYG5VrIOMu:-NbxJITN1cnYG5VrIOMv:blx40ff) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.html#L92)):*
> ```HTML
>             </div>
>             <mat-error *ngIf="biblicalTermsMessage">
>               {{ biblicalTermsMessage }}
> ```

You don't have to do this, but I think this would be better served and more mobile-conscious as a tooltip (app-info)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.html` line 1 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxML_XCLiO5lKfL1i4:-NbxML_YAIwouNFQXneh:b-bsa0y7) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.html#L1)):*
> ```HTML
> <ng-container *transloco="let t; read: 'biblical_term_dialog'">
> ```

This dialog is really wide on desktop. Mobile looks good. On other dialogs we've used maxWidth: 560px
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.html` line 16 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxMrDeB3LbFZ_xtTDh:-NbxMrDeB3LbFZ_xtTDi:b-k2wn) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.html#L16)):*
> ```HTML
>       ></mdc-textarea>
>     </mdc-form-field>
>     <mdc-form-field [formGroup]="form">
> ```

We need some sort of guidance for this field that entries should be "separated by a new line."
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.ts` line 61 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxPtuzBiOf3fEnIh7j:-NbxPtuzBiOf3fEnIh7k:b-m0074s) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.ts#L61)):*
> ```XML
> 
>   submit(): void {
>     const renderings = this.renderings.value.split(/\r?\n/).map((rendering: string) => rendering.trim());
> ```

Can we get a strong type on this sucker?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 223 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nbxrk-M7DJvRGgj5Her:-Nbxrk-M7DJvRGgj5Hes:b-mvt670) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L223)):*
> ```XML
>     } else {
>       let verseRef = new VerseRef(
>         Canon.bookNumberToId(this._bookNum ?? 0),
> ```

Do we need to check these again? It looks like we check them a few lines above
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 331 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxpV5i5q0NkfW80lxE:-NbxpV5i5q0NkfW80lxF:bz9d24f) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L331)):*
> ```XML
>   }
> 
>   // TODO: Implement this functionality in @sillsdev/scripture
> ```

Yes, can we please! :D
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 379 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxsF6T5sDWGa-AmCZv:-NbxsF6T5sDWGa-AmCZw:b-m5pneg) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L379)):*
> ```XML
>       }
> 
>       let noteThreadDoc: NoteThreadDoc | undefined;
> ```

Is it a concern that we always populate this, when it appears to only be used if we are going to display the term? Could we "continue" above this if we're not? Or move it into the conditional?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 411 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nbxt3GH-hmVS8z0b7Hz:-Nbxt3GH-hmVS8z0b7I-:br131n9) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L411)):*
> ```XML
>   }
> 
>   private getVerses(verseRef: VerseRef): number[] {
> ```

We should also consider moving this to VerseRef. Seems handy!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html` line 51 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxTbY42fCIUCUoRKef:-NbxTbY42fCIUCUoRKeg:bb87po0) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html#L51)):*
> ```HTML
>     </div>
>     <mat-form-field *ngIf="canInsertNote" class="full-width" appearance="outline">
>       <mat-label>{{ t("your_comment") }}</mat-label>
> ```

I think it's a touch odd that we don't say "note" here. Was there a specific reason for not doing so?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts` line 50 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc-wsur8L3mRVi6OO57:-Nc-wsur8L3mRVi6OO58:b-4msx61) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L50)):*
> ```XML
>   showSegmentText: boolean = false;
>   currentNoteContent: string = '';
>   private biblicalTermDoc?: BiblicalTermDoc;
> ```

Did you consider subclassing? It seems bad practice to toss in extra weight here. Or the Decorator pattern.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 9 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxRTseBtjHBd5OtDSA:-NbxRTsfDN4aXnyYnLSE:b-ajbm65) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L9)):*
> ```JSON
>     "users": "Users"
>   },
>   "biblical_term_dialog": {
> ```

At one point I was told we weren't checking in the non-English versions of this file. Is that not the case?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 286 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxRnOO0hxb4Pv3pBmo:-NbxRnOO0hxb4Pv3pBmp:b14zlmt) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L286)):*
> ```JSON
>   "suggestions_settings_dialog": {
>     "better": "Better",
>     "biblical_terms": "Show/Hide Biblical Terms",
> ```

I would just say "Show Biblical Terms" here. I think that'd be sufficient with the slider. Thoughts?
___
*[`src/SIL.XForge.Scripture/Models/NoteThread.cs` line 31 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0UJIr8ethGmzZEyqm:-Nc0UJIr8ethGmzZEyqn:brgtvb6) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Models/NoteThread.cs#L31)):*
> ```Smalltalk
>     /// The Biblical Term this note is for. Null if not for a Biblical Term.
>     /// </summary>
>     public string? BiblicalTermId { get; set; }
> ```

As before, these properties are prime candidates for subclassing or decoration. Thoughts?
___
*[`src/SIL.XForge.Scripture/Services/IParatextService.cs` line 75 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0VP7M16KS5bR2zhgr:-Nc0VP7M16KS5bR2zhgs:b-3xdzge) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/IParatextService.cs#L75)):*
> ```Smalltalk
>     );
>     int UpdateCommentTag(UserSecret userSecret, string paratextId, NoteTag noteTag);
>     Task<BiblicalTermsChanges> GetBiblicalTermsAsync(UserSecret userSecret, string paratextId, IEnumerable<int> books);
> ```

Gotta love a 96-line interface file :|
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1446 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0WOvP7x4W_Q5VOvlj:-Nc0WOvP7x4W_Q5VOvlk:byhwcwx) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1446)):*
> ```Smalltalk
>     }
> 
>     public async Task<BiblicalTermsChanges> GetBiblicalTermsAsync(
> ```

Method documentation would be useful here. Books as int enumerable is initially unclear to me (that's picky)
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1580 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0_DyR1HnXyc4RYHXt:-Nc0_DyR1HnXyc4RYHXu:b-1ug16a) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1580)):*
> ```Smalltalk
>         {
>             // Dispose the Biblical Terms ScrText from Settings, if it was set
>             biblicalTermsScrText?.Dispose();
> ```

Do we also need to dispose of scrText? I see that done in some other methods
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1584 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0aSly0OdZojLGPZ63:-Nc0aSly0OdZojLGPZ64:b-gaqqz1) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1584)):*
> ```Smalltalk
>     }
> 
>     public void UpdateBiblicalTerms(UserSecret userSecret, string paratextId, IReadOnlyList<BiblicalTerm> biblicalTerms)
> ```

Method comment here, too, please
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1930 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0bTan7ex9F9Vm0v4W:-Nc0bTan7ex9F9Vm0v4X:b-u0w94u) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1930)):*
> ```Smalltalk
> 
>     /// <summary>
>     /// Converts a Paratext language code into a language-country codes for the frontend.
> ```

Grammar typo
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 608 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0f95k1qNtFEDWSkT-:-Nc0f95k1qNtFEDWSkT0:bh5259c) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L608)):*
> ```Smalltalk
>         SyncPhase syncPhase,
>         string paratextId,
>         List<IDocument<BiblicalTerm>> biblicalTermDocs
> ```

Is it normal to accept a list as a parameter instead of incorporating it into the return type? Wouldn't it be clearer to just return it in the Task?
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 646 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0goqn8GLeaqCh1Njw:-Nc0goqn8GLeaqCh1Njx:by4cf19) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L646)):*
> ```Smalltalk
>         LogMetric("Updating Paratext biblical terms");
>         await NotifySyncProgress(syncPhase, 50);
>         double i = 0.0;
> ```

I think you can just do int here
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 720 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0iADfC4uGhcyXMUqb:-Nc0iADfC4uGhcyXMUqc:b-8qeich) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L720)):*
> ```Smalltalk
>         // Add and Update existing terms
>         LogMetric("Updating biblical terms");
>         double i = 0.0;
> ```

Use int here
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs` line 2136 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0lQR5DlxZmtSEcUji:-Nc0lQR5DlxZmtSEcUjj:b2s430p) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2136)):*
> ```Smalltalk
>         Assert.That(change.NotesAdded.Single().DataId, Is.EqualTo(newDataId));
>     }
> ```

We should also be testing the two new public methods, right?

josephmyers:
Adding the first note doesn't update the notes icon
https://github.com/sillsdev/web-xforge/assets/43733878/89b4540e-dd46-4588-8f0e-763ec5986e4a

pmachapman (commented):
Thanks for your review. This was originally meant to be a much smaller PR, but the interconnected nature of the components made it grow too much unfortunately. Thank you for your patience!

> In my opinion, I don't think a refresh justifies resetting the user's layout.

Hmmm I agree in principle, however this is the behavior everywhere else angular split is used. Biblical Terms was originally not supposed to be visible on mobile, so I am loathe to give it too much real estate by default (i.e. default to a number > 0 on xs devices). I did some experiments with storing the value, but things like rotating the screen threw this off for me, and the code started getting spaghetti-like for just this use case...

Perhaps as a separate PR we could create in implementation of the angular split component that stores the split area size values in the session, and handles things like window resizes and screen rotation, not causing the important pane to shrink to 0 on these events (I think a "visibility priority" or "important pane" property might be need for this).

I think that the fact that this will be little used on XS devices, and because the user would have had to drag the slider up to initially use the feature makes me think this might be OK to leave as-is for now?

> Now, I think another option could be to use a docked, toggleable pane, a la your favorite IDE.

I am not aware of any open source components that do this - my initial thought is we would need to apply this consistently, so perhaps as part of the UI revamp investigation?

> For terms being cut off, I believe it's as simple as CSS overflow-wrap/word-wrap

I looked into this, and it is the default styling from angular material. I tried tweaking it, but the slide toggle alignments were messed up with the overflow. I think we should leave this at the default for consistency?

> Adding the first note doesn't update the notes icon

When I added via SF the icon updated immediately, and adding via PT then syncing caused it to update immediately. Are there any other errors in your browser console?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-:-NcK8kUX1PVeJPj_5Gwa:bl7cdtj)* status: all files reviewed, 22 unresolved discussions (waiting on @josephmyers)

___
*[`src/RealtimeServer/scriptureforge/models/biblical-term.ts` line 19 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxpBg8A_yht2A6oru5:-NcKQ4X00neI52YyfbWk:b-qo4lup) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/RealtimeServer/scriptureforge/models/biblical-term.ts#L19)):*

Can we not store this as a VerseRef?
</blockquote></details>

Good observation. We thought of doing this originally, but chose the reference number as it is sortable, meaning we can use it in Mongo queries easily. We also don't need the additional feature that VerseRef provides of partial verses. It also will mean less data is stored on the client, etc.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.html` line 92 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxJITN1cnYG5VrIOMu:-NcKfDSS9XhCfkoB-gCw:b-qy0g3i) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.html#L92)):*

You don't have to do this, but I think this would be better served and more mobile-conscious as a tooltip (app-info)
</blockquote></details>

That is a great idea! Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.html` line 1 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxML_XCLiO5lKfL1i4:-NcKj-GY9kfQNBAtmK7u:bw87i2k) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.html#L1)):*

This dialog is really wide on desktop. Mobile looks good. On other dialogs we've used maxWidth: 560px
</blockquote></details>

I have changed it to 560px max.

Also, I have updated the dialog to use mat- components not mdc- (so it looks a littler different now)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.html` line 16 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxMrDeB3LbFZ_xtTDh:-NcKseMr83Xc9R-zsJP2:bc4w83q) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.html#L16)):*

We need some sort of guidance for this field that entries should be "separated by a new line."
</blockquote></details>

Good idea - I have added a hint:

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.ts` line 61 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxPtuzBiOf3fEnIh7j:-NcL3__6-w47wZgjzTt7:binzrnk) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.ts#L61)):*

Can we get a strong type on this sucker?
</blockquote></details>

Done. Have added types elsewhere I forgot, too.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 223 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nbxrk-M7DJvRGgj5Her:-NcL5ucO-csWwAx8fEeg:beb1f58) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L223)):*

Do we need to check these again? It looks like we check them a few lines above
</blockquote></details>

No - fixed - thanks!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 331 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxpV5i5q0NkfW80lxE:-NcPZGdmCdzVE_8s-mDr:b-52f39i) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L331)):*

Yes, can we please! :D
</blockquote></details>

Fixed - thanks for spotting it. I have added it to sillsdev/scripture, I just forgot about this TODO when rebasing...
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 379 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxsF6T5sDWGa-AmCZv:-NcP_Cj04jRCgUOOjOoU:b-hurtcv) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L379)):*

Is it a concern that we always populate this, when it appears to only be used if we are going to display the term? Could we "continue" above this if we're not? Or move it into the conditional?
</blockquote></details>

I think we should move it. Fixed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 411 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nbxt3GH-hmVS8z0b7Hz:-NcPloEnDbg2hi9pMYwv:bted126) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L411)):*

We should also consider moving this to VerseRef. Seems handy!
</blockquote></details>

I have moved it to utils.ts where other VerseRef utilities are.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html` line 51 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxTbY42fCIUCUoRKef:-NcPqJr17_EFYkzogBDe:b-wpgdn4) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html#L51)):*

I think it's a touch odd that we don't say "note" here. Was there a specific reason for not doing so?
</blockquote></details>

Not sure - this was done in #1754.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts` line 50 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc-wsur8L3mRVi6OO57:-NcPrh7z4oN9zHH53eQA:b-km5kdj) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L50)):*

Did you consider subclassing? It seems bad practice to toss in extra weight here. Or the Decorator pattern.
</blockquote></details>

I based this off how PT's note dialog does things - the only part of this I really don't like is the change to `getNoteContextText`. My main motivation to keep it here was to try and reduce maintenance by keeping note dialog code all in one place, and being aware that this dialog is still being updated and modified heavily in other PRs (I didn't want to accidentally fork the notes dialog).

Maybe when the notes feature goes live, we can look at a refactor and probably subclassing (although the Decorator pattern might be easier)?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 9 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxRTseBtjHBd5OtDSA:-NcPt7bD4tbxPmDP6xKC:b-b4xqih) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L9)):*

At one point I was told we weren't checking in the non-English versions of this file. Is that not the case?
</blockquote></details>

I have added non-English translations as many of the strings are borrowed from Paratext (which has the translations already). But yes, I did this a long time ago before I was made aware that I should only do the English strings.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 286 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxRnOO0hxb4Pv3pBmo:-NcPvopx5dKatzuRKG1u:bb6un1j) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L286)):*

I would just say "Show Biblical Terms" here. I think that'd be sufficient with the slider. Thoughts?
</blockquote></details>

Done
___
*[`src/SIL.XForge.Scripture/Models/NoteThread.cs` line 31 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0UJIr8ethGmzZEyqm:-NcPw-4y4R8dAD8a5YGi:b-le1sl2) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Models/NoteThread.cs#L31)):*

As before, these properties are prime candidates for subclassing or decoration. Thoughts?
</blockquote></details>

This is the class that corresponds to the Mongo collection `note_threads`, so I am not sure of the value of subclassing in this case?
___
*[`src/SIL.XForge.Scripture/Services/IParatextService.cs` line 75 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0VP7M16KS5bR2zhgr:-NcPwE5uEHOVPZ0x8gVn:br5g5q6) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/IParatextService.cs#L75)):*

Gotta love a 96-line interface file :|
</blockquote></details>

I know. I do try and avoid adding more, but given the singleton nature of the class that implements this interface, and how ParatextData keeps a lot of caches in static variables, we are sadly trapped (for now).
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1446 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0WOvP7x4W_Q5VOvlj:-NcPyL5W4G0BXsOJa3vy:bb6un1j) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1446)):*

Method documentation would be useful here. Books as int enumerable is initially unclear to me (that's picky)
</blockquote></details>

Done
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1580 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0_DyR1HnXyc4RYHXt:-NcQ32iiBFC7qxFtmelC:brypfim) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1580)):*

Do we also need to dispose of scrText? I see that done in some other methods
</blockquote></details>

The variable `scrText` is created in a using statement, so is automatically disposed when the function returns.
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1584 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0aSly0OdZojLGPZ63:-NcQ2kvg2w820y_d3zDu:bb6un1j) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1584)):*

Method comment here, too, please
</blockquote></details>

Done
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1930 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0bTan7ex9F9Vm0v4W:-NcQ2qku-qFeMPcLOLUV:bt2vr1b) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1930)):*

Grammar typo
</blockquote></details>

Fixed.
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 608 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0f95k1qNtFEDWSkT-:-NcQFvCy02yTTCKICXCL:b-ikyoi8) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L608)):*

Is it normal to accept a list as a parameter instead of incorporating it into the return type? Wouldn't it be clearer to just return it in the Task?
</blockquote></details>

You are right it is good practice to return a list if you are modifying it (as the parameter is passed by reference, it isn't necessary, but it does make it clearer). Done.
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 646 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0goqn8GLeaqCh1Njw:-NcQFE58DLA93BWuL6r5:b-t0s59n) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L646)):*

I think you can just do int here
</blockquote></details>

I use a double so that floating point division is used rather than integer division in the call to `NotifySyncProgress` (which accepts a double).
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 720 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0iADfC4uGhcyXMUqb:-NcQFN3TAH5RY7U9c5rR:b-otilcs) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L720)):*

Use int here
</blockquote></details>

See comment above.
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs` line 2136 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0lQR5DlxZmtSEcUji:-NcUKwq9DztnhT5IPD1s:b2whxcu) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2136)):*

We should also be testing the two new public methods, right?
</blockquote></details>

Unfortunately I have not been able to successfully create unit tests for these two methods due to the heavy reliance on static methods in ParatextData and the file system.

josephmyers (changes requested):
> this is the behavior everywhere else angular split is used.

You doubtless know the software better than I. However, the only other place I know that we use a split is the question/answer component in Community Checking, and it is actually a fantastic example of what we're looking for. It autosizes dynamically, and it handles screen rotation. Can we match that behavior? If you're really concerned about real estate, make the default height small. We just want it obvious that toggling the switch changed something

> it is the default styling from angular material.

Take a look at how the checkboxes work in Settings. We do resort to a couple lines of ng-deep; however, it's better than the users not being able to read the fields. Another option would be to use a header, e.g. Biblical Terms, and pull out the term from each field. That may actually be cleaner

> When I added via SF the icon updated immediately

Yes, I'm seeing it fixed now, too!

> Reviewed 27 of 27 files at r15, all commit messages.
> Dismissed @josephmyers from 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-:-NcWC2fa0trZz_elUee4:b-cbj95v)* status: all files reviewed, 3 unresolved discussions (waiting on @pmachapman)

___
*[`src/RealtimeServer/scriptureforge/models/biblical-term.ts` line 19 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxpBg8A_yht2A6oru5:-Nc_8KQS47wiUCRS6ziZ:brxp9sd) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/RealtimeServer/scriptureforge/models/biblical-term.ts#L19)):*

Good observation. We thought of doing this originally, but chose the reference number as it is sortable, meaning we can use it in Mongo queries easily. We also don't need the additional feature that VerseRef provides of partial verses. It also will mean less data is stored on the client, etc.
</blockquote></details>

Sounds good!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.html` line 16 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxMrDeB3LbFZ_xtTDh:-Nc_AdVP2W7Sf3aSLFOD:b-5gwryt) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.html#L16)):*

Good idea - I have added a hint:

</blockquote></details>

Fantastic!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html` line 51 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxTbY42fCIUCUoRKef:-Nc_COQ5Cri48g_pBOas:bqlzutm) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html#L51)):*

Not sure - this was done in #1754.
</blockquote></details>

It's probably fine, but we should be careful about our terminology. Especially since each means a specific thing in at least one other place in the app
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts` line 50 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc-wsur8L3mRVi6OO57:-Nc_D0ha8tlMZPcchqEm:b-uqi1wn) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L50)):*

I based this off how PT's note dialog does things - the only part of this I really don't like is the change to `getNoteContextText`. My main motivation to keep it here was to try and reduce maintenance by keeping note dialog code all in one place, and being aware that this dialog is still being updated and modified heavily in other PRs (I didn't want to accidentally fork the notes dialog).

Maybe when the notes feature goes live, we can look at a refactor and probably subclassing (although the Decorator pattern might be easier)?
</blockquote></details>

We can postpone it if you want. Just add a comment/todo to circle back!
___
*[`src/SIL.XForge.Scripture/Models/NoteThread.cs` line 31 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0UJIr8ethGmzZEyqm:-Nc_EcV68GyWZeLm-bO8:b-5fgf1y) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Models/NoteThread.cs#L31)):*

This is the class that corresponds to the Mongo collection `note_threads`, so I am not sure of the value of subclassing in this case?
</blockquote></details>

I'm not sure how difficult it would be. It's just weird that we're tacking on biblical terms to all notes, when only some of them should have that data. We're trying to reuse classes, which is great, but we really want a modified version of it. Conceptually, pre-existing note threads should be able to be unaffected. I do get that it reflects the Mongo collection. Maybe that makes it okay?
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1580 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0_DyR1HnXyc4RYHXt:-Nc_Bgay7c-0kuKXpJ9t:b-ij9bwm) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1580)):*

The variable `scrText` is created in a using statement, so is automatically disposed when the function returns.
</blockquote></details>

Duh! Thanks
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs` line 2136 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0lQR5DlxZmtSEcUji:-Nc_VUs1D_aFLKaDcGes:b9p4xed) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/test/SIL.XForge.Scripture.Tests/Services/ParatextServiceTests.cs#L2136)):*

Unfortunately I have not been able to successfully create unit tests for these two methods due to the heavy reliance on static methods in ParatextData and the file system.
</blockquote></details>

Definitely unfortunate! But no worries

pmachapman (commented):
> However, the only other place I know that we use a split is the question/answer component in Community Checking, and it is actually a fantastic example of what we're looking for. It autosizes dynamically, and it handles screen rotation. Can we match that behavior? If you're really concerned about real estate, make the default height small. We just want it obvious that toggling the switch changed something

You raise a good point - the hidden state is bad UX. After some tinkering, I have made it the same 75/25 split as for larger screen sizes. I will create a ticket for us to extend the splitter to save split positions for the user - I think that will help make it less obtrusive when the user shrinks or expands it, and moves to other chapters, etc.

> Take a look at how the checkboxes work in Settings. We do resort to a couple lines of ng-deep; however, it's better than the users not being able to read the fields. Another option would be to use a header, e.g. Biblical Terms, and pull out the term from each field. That may actually be cleaner

I have investigated this, and it is a bug with Angular Material v14, which is fixed in v15 (see the screenshots below). Can we wait for a resolution to this until we upgrade @angular/material?

https://v14.material.angular.io/components/slide-toggle/examples:


https://v15.material.angular.io/components/slide-toggle/examples:


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-:-Nct6YOnBgXrkEvfsBSv:bkn0szy)* status: all files reviewed, 3 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html` line 51 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-NbxTbY42fCIUCUoRKef:-Ncsoey1EwJrUDB6bQhr:b-2vb508) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d995b7129f54e579a9d1c4a018f4946b598243a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html#L51)):*

It's probably fine, but we should be careful about our terminology. Especially since each means a specific thing in at least one other place in the app
</blockquote></details>

Agreed. Will keep an eye on it.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts` line 50 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc-wsur8L3mRVi6OO57:-NcstF2f7RQBLxXRa56o:b-ut8jqq) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L50)):*

We can postpone it if you want. Just add a comment/todo to circle back!
</blockquote></details>

Yes - I think it would be a good idea to refactor it (I don't think it will be too hard, although the tests will need some refactoring to stop copy-pasta). Added a TODO note.
___
*[`src/SIL.XForge.Scripture/Models/NoteThread.cs` line 31 at r14](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-Nc0UJIr8ethGmzZEyqm:-NcsuYTiD5tLAR5aQNn2:b4tfqc8) ([raw file](https://github.com/sillsdev/web-xforge/blob/600829e71f108df0ffb0da858df0410945e2bcc8/src/SIL.XForge.Scripture/Models/NoteThread.cs#L31)):*

I'm not sure how difficult it would be. It's just weird that we're tacking on biblical terms to all notes, when only some of them should have that data. We're trying to reuse classes, which is great, but we really want a modified version of it. Conceptually, pre-existing note threads should be able to be unaffected. I do get that it reflects the Mongo collection. Maybe that makes it okay?
</blockquote></details>

I think so. Also Paratext uses the same structure, so we more or less are following their way of doing this for this feature.

josephmyers:
Small issue I'm just now seeing. Is this expected?
https://github.com/sillsdev/web-xforge/assets/43733878/6b1349db-e0a0-4b18-836f-f312076aaad9

josephmyers (approved):
Ah, yes, let's wait til v15. Thanks for figuring that out! And the new sizing looks great!

I saw the "note icon not refreshing when you add the first note" issue again, but it seems that refreshing the page prevents the bug from happening again. I also am seeing some strangeness when sorting the entries by Renderings (video above). Maybe it's expected? I'm going to go ahead and approve, assuming this is not an issue. Thanks for all your work!

> Reviewed 18 of 18 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-:-NctiQEp1T1O-HJk39a1:b-m4m3xr)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (commented):
> I saw the "note icon not refreshing when you add the first note" issue again, but it seems that refreshing the page prevents the bug from happening again.

Very strange. I am going to keep an eye on this one.

> I also am seeing some strangeness when sorting the entries by Renderings (video above)

That looks to be expected, as the sort is a simple alpha sort. Do you think it more usable to have blank values below words when sorting? (I can create a ticket to address this post-merge?)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1731#-:-Ncu3A598yLMqAb6NFvt:bvrntg9)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #1732, opened 2023-03-07
SF-1889 Attribute new SF notes to reviewer user on sync

RaymondLuong3:
* store note thread doc values in variable at sync time
* throw if a note cannot be found when it should exist
* check the sync user ref on a note before creating a new comment
* some clean up

The primary problem for this issue was that we fetched the note thread docs from the realtime server twice. Once when we export the note thread doc changes to PT, and then a second time when we pull in the up-to-date note threads from PT. When we did this, the SyncUserRef property on the note threads were not up to date on the second set of note thread docs, which caused the sync to incorrectly assume that it needed to delete the existing docs and recreate them with the sync user as the owner. This fixes that problem as well as introduces an exception that is thrown if we have a sync user on a note, but cannot locate the PT equivalent comment.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1732)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -1225,11 +1225,7 @@ public class ParatextService : DisposableBase, IParatextService
                 if (matchedComment != null)
                 {
                     matchedCommentIds.Add(matchedComment.Id);
-                    Paratext.Data.ProjectComments.CommentTag commentIconTag = GetCommentTag(
-                        existingThread,
-                        matchedComment,
-                        commentTags
-                    );
+                    CommentTag commentIconTag = GetCommentTag(existingThread, matchedComment, commentTags);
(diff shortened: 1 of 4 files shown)


pmachapman (changes requested):
The change looks good - thank you for cleaning up other bits of code, too.

Just one query before I approve - I performed the test, and had the following difference display for the note in SF (the notes were correctly attributed) and PT (the notes were both attributed to the sync user who created the second note) - is this expected behavior?
image

image

RaymondLuong3 (commented):
That's a good question. Yes, this is expected. The user who does the sync gets attributed as the author for any notes left by non-pt users.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1732#-:-NQQqxneD-EroK_GbVIJ:b-6ph61j)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (approved):

PR #1733, opened 2023-03-07
SF-1891 Do not select a verse when clicking note thread icon

RaymondLuong3:
Previously if you opened and close a note dialog, the verse would be selected because clicking on the icon actually also selected the verse. However, the UI would not look like the verse was selected.
To fix this issue, we avoid triggering selecting the verse if a user only clicked the note thread icon on the verse.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1733)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -2643,6 +2643,36 @@ describe('EditorComponent', () => {
       env.dispose();
     }));
 
+    it('does not select verse when opening a note thread', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig();
+      env.setReviewerUser();
+      env.addParatextNoteThread(
+        6,
+        'MAT 1:1',
+        '',
+        { start: 0, length: 0 },
+        ['user01'],
+        NoteStatus.Todo,
+        undefined,
+        true
+      );
+      env.wait();
+
+      const elem: HTMLElement = env.getNoteThreadIconElement('verse_1_1', 'thread06')!;
+      elem.click();
+      env.mockNoteDialogRef.close();
+      env.wait();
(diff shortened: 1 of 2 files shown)


Nateowami (commented):
Did you consider using [`Event.stopPropagation()`](https://developer.mozilla.org/en-US/docs/Web/API/Event/stopPropagation)? I think if you used that on the note icon when it gets clicked, it will prevent the parent (the verse) from being told it was clicked. This is how we've usually solved the "a thing is getting clicked inside a thing that is clickable" problem.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1733#-:-NPysHGk8p4GtLxhA6F_:b-49nfg6)* status: 0 of 2 files reviewed, all discussions resolved

RaymondLuong3 (commented):
What an excellent idea!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1733#-:-NQ2-cNH0YbqDoxOpZAU:b-zedkcc)* status: 0 of 2 files reviewed, all discussions resolved

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1733)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1213 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1733#-NQ27DYA0gZZQWTlTDV_:-NQ27DYA0gZZQWTlTDVa:b8w3bcs) ([raw file](https://github.com/sillsdev/web-xforge/blob/84f31b8a778ac3d646e6ca417fa581b53ccc996f/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1213)):*
> ```XML
>               this.updateReadNotes(threadId);
>             }
>             event.stopPropagation();
> ```

Would you mind adding a comment here explaining why? In general this only has to be done because someone discovered that something was broken. The comment would allow someone in the future to know why it was added and then be able to reason about whether it could be removed, without having to check git history.

PR #1736, opened 2023-03-08
SF-1900 Delete comments in a deleted note thread

RaymondLuong3:
* delete comment threads when corresponding note thread is deleted
* ignore new comment threads where all notes are deleted
* update logic for when to delete a note thread

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1736)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -2745,6 +2745,21 @@ describe('EditorComponent', () => {
       expect(noteThreadEmbedCount).toEqual(4);
       env.dispose();
     }));
+
+    it('should remove note thread icon from editor when thread is deleted', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig();
+      env.wait();
+
+      const threadId = 'thread02';
+      const segmentRef = 'verse_1_3';
+      let thread2Elem: HTMLElement | null = env.getNoteThreadIconElement(segmentRef, threadId);
+      expect(thread2Elem).not.toBeNull();
+      env.deleteNoteThread('project01', segmentRef, threadId);
+      thread2Elem = env.getNoteThreadIconElement(segmentRef, threadId);
+      expect(thread2Elem).toBeNull();
+      env.dispose();
+    }));
   });
 
   describe('Translation Suggestions disabled', () => {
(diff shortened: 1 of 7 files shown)


pmachapman (changes requested):
The new logic works well for the case documented in the ticket, although while reviewing I found two bugs I think are related to the change in logic:.

**First Bug:** When I delete the last note in a thread, it reappears when I sync.

My process was:
 1. Uncomment and modify the following line:
    ```cs
    // TODO: Sync Note changes back to Paratext, and record sync metric info
    // await _paratextService.UpdateParatextCommentsAsync(_userSecret, paratextId, text.BookNum,
    //     noteThreadDocs, _currentPtSyncUsers);
    ```
    To: 
    ```cs
    // TODO: Sync Note changes back to Paratext, and record sync metric info
    await _paratextService.UpdateParatextCommentsAsync(_userSecret, paratextId, text.BookNum,
        noteThreadDocs, _currentPtSyncUsers, _projectDoc.Data.TranslateConfig.DefaultNoteTagId!.Value);
    ```
 2. Run SF, add a new note, then sync:
     image
 3. In PT, Sync, then add a new note to the thread, then send/receive project:
 image
 4. In SF, Sync, delete the note, then sync again:
    image
   To:
   image
 5. When opting the note again, the Second note in thread reappears:
    image

**Second Bug:** When I deleted both notes in the thread, the note remains stuck in the editor, and clicking on it showed this dialog (although it did delete correctly on Sync):
image

Finally after my testing, empty notes like this were now present in my Mongo, but I cannot trace an exact origin to them (I am flagging in case you see these too in your testing):
image

RaymondLuong3 (commented):
Thanks for those steps! I was able to reproduce the bugs you were seeing and fix them. It was more involved because I needed to think through how we use the `Deleted` property on Notes, but I think I have the answer. `Deleted` would not be set in SF, so we can ignore it, but if a note is altogether deleted, then we have to handle that.
As for your empty note thread doc, I think that is ok. As long as the data property is empty in mongo, it will appear as if it doesn't exist to our current system (unless you are seeing otherwise)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1736#-:-NQRcJop6a2koTL4bgxo:b-2487w)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1736#-:-NQSYzhB37hFOoOueDaZ:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1743, opened 2023-03-14
SF-1712 - Transparent authentication

nigel-wells:
- Expanded the join component to allow non-authenticated users to create a user and join the project
- Created an anonymous backend service
- Updated to the latest version of auth0-spa-js
- Generate a cookie for transparent authenticated users to store credentials
- Added a login state to the auth service allowing the app to respond to changes in the state of user becoming logged in after initial load

image

### Acceptance tests

- Can be used for recipient only links
- Can be used for anyone links
- Share link will automatically login any user who has previously used the share link and joined the project for the same user on the same device
- Logged in (normal authentication) users who use the share link will continue to join the project without any prompts providing they are on the project or the share link is valid for them to use
- Admin can not use the share link after it has been used to gain access to the other user accounts
- Transparent login is locked to the device and browser it is first opened on
- Logging out of a shared account should prompt a warning about losing access to the account
- Normal accounts should not be prompted with a dialog when logging out

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1743)


Nateowami (commented):
Just tried this out. It's incredible. 

One thing I noticed is that it still asks you to confirm your name when adding an answer. This should not be necessary because the user has already provided their name.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-:-NRaGYUDEIHlpT4SKNcZ:bfakssy)* status: 0 of 41 files reviewed, all discussions resolved

src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json
```diff
@@ -146,6 +146,16 @@
     "transcelerator_some_questions_already_imported": "Note: Some selected questions have already been imported. Questions that have been edited in Transcelerator will be updated in Scripture Forge, and exact duplicates will be skipped.",
     "update_transcelerator": "The version of Transcelerator used in this project is not supported. Please update to at least Transcelerator version 1.5.3."
   },
+  "join": {
+    "error_occurred_login": "An error occurred trying to join the project",
+    "invited_to_join": "You've been invited to join {{ boldStart }}{{ projectName }}{{ boldEnd }} project as a {{ boldStart }}{{ role }}{{ boldEnd }}",
```
Nateowami, line 151: The more I think about this the more I think it's not important to state the role. While it might be helpful to tell the user specifically what they're expected to do (e.g. comment or answer questions), this is probably not the place for that. It would be either in the invitation, or after logging in. Removing it would also solve the localization issue.

src/SIL.XForge/Controllers/UrlConstants.cs
```diff
@@ -2,6 +2,7 @@ namespace SIL.XForge;
 
 public static class UrlConstants
 {
+    public const string Anonymous = "anonymous";
```
Nateowami, line 5: I wonder if "unauthenticated" might make more sense here than "anonymous"?

src/SIL.XForge/Services/AuthService.cs
```diff
@@ -40,6 +40,19 @@ public bool ValidateWebhookCredentials(string username, string password)
 
     public Task<string> GetUserAsync(string authId) => CallApiAsync(HttpMethod.Get, $"users/{authId}");
 
+    public Task<string> GenerateTransparentUser(string name, string username, string password, string language)
+    {
+        var content = new JObject(
+            new JProperty("name", name),
+            new JProperty("username", username),
+            new JProperty("email", $"{username}@users.noreply.scriptureforge.org"),
```
Nateowami, line 48: Have you checked various places where we might send a user an email (e.g. to reset their password) to see how we will handle this? We don't want to be sending emails into the void.

src/SIL.XForge/Services/AuthService.cs
```diff
@@ -40,6 +40,19 @@ public bool ValidateWebhookCredentials(string username, string password)
 
     public Task<string> GetUserAsync(string authId) => CallApiAsync(HttpMethod.Get, $"users/{authId}");
 
+    public Task<string> GenerateTransparentUser(string name, string username, string password, string language)
```
Nateowami, line 43: How about anonymous user? The authentication is transparent, the user is anonymous.

src/SIL.XForge.Scripture/Services/SFProjectService.cs
```diff
@@ -361,6 +361,8 @@ public async Task CancelSyncAsync(string curUserId, string projectId)
         await _syncService.CancelSyncAsync(curUserId, projectId);
     }
 
+    public new async Task<SFProject> GetProjectAsync(string projectId) => await base.GetProjectAsync(projectId);
```
Nateowami, line 364: Maybe just mark the method as public in the base class? Or is there more going on here that I'm missing?

src/SIL.XForge.Scripture/Services/SFProjectService.cs
```diff
@@ -614,8 +617,6 @@ public async Task<string> CheckLinkSharingAsync(string curUserId, string shareKe
 
         string projectId = projectSecret.Id;
         ShareKey projectSecretShareKey = projectSecret.ShareKeys.FirstOrDefault(sk => sk.Key == shareKey);
-        if (projectSecretShareKey == null)
```
Nateowami, line 617: Was this removed because it will throw an exception anyway if the share key is null?

src/SIL.XForge.Scripture/Services/AnonymousService.cs
```diff
@@ -0,0 +1,65 @@
...
+        await CheckSharingKey(shareKey);
+        // Generate random username and password for the account
+        var credentials = new TransparentAuthenticationCredentials
+        {
+            Username = _securityService.GenerateKey(),
+            Password = _securityService.GenerateKey(16)
+        };
+        _ = await _authService.GenerateTransparentUser(
+            displayName,
+            credentials.Username,
```
Nateowami, line 59: If you pass the credentials to `GenerateTransparentUser()` then there are fewer arguments to get right, and you can't mix them up. As it is right now I'd have to go check the auth service to see whether you have the four string arguments in the right order.

Of course that doesn't fully solve the problem, there will still be two string arguments and one `TransparentAuthenticationCredentials` argument, but it's slightly more type safe. It just makes the code a little more maintainable in the future.

Nateowami (commented):
I've only reviewed a few files so far. Leaving my comments so far.

You'll have to forgive the typos that I'm sure are in here, because the VS Code GitHub Pull Request extension doesn't have spellcheck.

src/SIL.XForge.Scripture/Services/SFProjectService.cs
```diff
@@ -653,40 +691,35 @@ public async Task<string> CheckLinkSharingAsync(string curUserId, string shareKe
...
-                projectSecretShareKey.Key
-            );
-            return projectId;
+            return false;
         }
-        throw new ForbiddenException();
+        return true;
+    }
+
+    public SFProjectSecret GetProjectSecret(string shareKey)
```
pmachapman, line 715: I think it would make it clearer to rename this method to `GetProjectSecretByShareKey`, as usually when there is a single string parameter it is assumed to be the id.

test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs
```diff
@@ -340,6 +340,28 @@ public async Task ReserveLinkSharingKeyAsync_GenerateNewKeyIfReserved()
         );
     }
 
+    [Test]
+    public void GetProjectSecret_ReturnsSecret()
+    {
+        var env = new TestEnvironment();
+        var shareKey = "abcd";
```
pmachapman, line 347: This should be a constant in `SFProjectServiceTests`, like:
```csharp
private const string ShareKey = "abcd";
```
(and any other copy-pasted constant string you want to clean up in this class?)

pmachapman (changes requested):
Looks good, Nigel - thank you for all the work on this change!

Just a few comments on C# specific stuff - if I am being too picky, please push back.

Nateowami (commented):
I started a review a while ago and just discovered I didn't post it. Here are my comments now. It's possible some of them are outdated.

Nateowami (commented):
> Reviewed 4 of 27 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1743)* status: 4 of 44 files reviewed, 25 unresolved discussions (waiting on @nigel-wells and @pmachapman)

___
*[`src/SIL.XForge/Services/SecurityService.cs` line 10 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NSDHT1n05cctP4F9ckk:-NSDHT1n05cctP4F9ckl:b-2kbx0) ([raw file](https://github.com/sillsdev/web-xforge/blob/d3b4774624d7a766f8ad9edca0a1f4267bd155f2/src/SIL.XForge/Services/SecurityService.cs#L10)):*
> ```Smalltalk
> {
>     /// <summary>Return a random 16-character base-64 string that is safe to use in URLs.</summary>
>     /// <param name="length">Length of data to generate the key from</param>
> ```

How about specifying that this is in bytes? Under "remarks" it says "For example, 12 bytes of data will become 16 bytes of base-64 text", but now that this variable is no longer named "bytes" it isn't as clear that this isn't the final length.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/anonymous.service.ts` line 24 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZQK-r1-24:-NSDI0Uu3ieVfINIXF1v:b-vb4f7k) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/anonymous.service.ts#L24)):*

Hmmm, I was following what we did in other parts of the front end that weren't directly using the JSON-RPC controllers i.e. file.service.ts. I'm not sure if there was rational when that code was first created - I had assumed it was because we weren't dealing with the RPC controller so we decided to use form data instead.
</blockquote></details>

@nigel-wells Files can't be posted as JSON; I assume that's why we used form data rather than JSON RPC there.

pmachapman (changes requested):
> Reviewed 2 of 41 files at r1, 5 of 27 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1743)* status: 11 of 44 files reviewed, 18 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge/Services/ISecurityService.cs` line 5 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NSDUfyp-_4xev5lFhkm:-NSDUfyp-_4xev5lFhkn:b2xrdud) ([raw file](https://github.com/sillsdev/web-xforge/blob/d3b4774624d7a766f8ad9edca0a1f4267bd155f2/src/SIL.XForge/Services/ISecurityService.cs#L5)):*
> ```Smalltalk
> public interface ISecurityService
> {
>     string GenerateKey(int bytes = 12);
> ```

NIT: Should probably change length here too for consistency.
___
*[`src/SIL.XForge/Services/UserService.cs` line 56 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NSDX_UH0-FjyNRFy8gW:-NSDX_UH0-FjyNRFy8gX:bgep2kf) ([raw file](https://github.com/sillsdev/web-xforge/blob/d3b4774624d7a766f8ad9edca0a1f4267bd155f2/src/SIL.XForge/Services/UserService.cs#L56)):*
> ```Smalltalk
>         bool isAnonymousUser =
>             identities.OfType<JObject>().FirstOrDefault(i => (string)i["connection"] == "Transparent-Authentication")
>             != null;
> ```

NIT: For readability, you could rewrite this as:
```
        bool isAnonymousUser = identities
            .OfType<JObject>()
            .Any(i => (string)i["connection"] == "Transparent-Authentication");
```



_Code quote:_
```Smalltalk
        bool isAnonymousUser =
            identities.OfType<JObject>().FirstOrDefault(i => (string)i["connection"] == "Transparent-Authentication")
            != null;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/anonymous.service.ts` line 15 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZQK-r1-15:-NSD_8ZQ0PBmtk1oNP4a:b-slvh8b) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/anonymous.service.ts#L15)):*

I could, I was trying to be consistent (see my other note) with how data is posted to the backend outside of the RPC controller. I guess this is a good time as any to discuss the best approach.
</blockquote></details>

See my comment in the other note.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/anonymous.service.ts` line 24 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZQK-r1-24:-NSDYKx-8baBCt6LA4uj:b-jo5c4x) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/anonymous.service.ts#L24)):*

@nigel-wells Files can't be posted as JSON; I assume that's why we used form data rather than JSON RPC there.
</blockquote></details>

@Nateowami That is my understanding too.
@nigel-wells The machine-api does not use JSON-RPC - I think JSON is a helpful format for all APIs where practical.
___
*[`src/SIL.XForge.Scripture/Controllers/AnonymousController.cs` line 26 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZQV-r1-26:-NSD_NpW5bF7fc7SWYSv:bsjet85) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/src/SIL.XForge.Scripture/Controllers/AnonymousController.cs#L26)):*

I was in two minds about this. Our other calls to the backend are actually first letter lowercase which we then convert to upper case first letter. I get it looks odd in the backend like this but I was trying to be consistent with the front end lol. I'm on the fence and happy to lean either way.
</blockquote></details>

Case will not matter for the endpoint resolution - Calling "checkSharingKey" will execute "CheckSharingKey()" - my query is about "Check**Sharing**Key" vs "Check**Share**Key".

I raised this as I tried to see where it was called in the frontend by searching for "CheckSharingKey", and didn't find it. Then I realized a different endpoint name (checkShareKey) was defined in the HttpPost attribute, and searched for that (and found it). I don't mind which name you use ("CheckShareKey" or "CheckSharingKey"), so long as it is consistent (the case of the first letter excluded) when running a find all across the project.
___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 347 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZQs-r1-347:-NSDZgqaCWETBg0CQTZh:b6fcfbg) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L347)):*

But it is only relevant to the one test. I change it again in `GetProjectSecret_ThrowsOnNotFound`. I could just inline them if that is you're preference?
</blockquote></details>

My query is more because this key is defined in the TestEnvironment ProjectSecret repository:
```
                            new ShareKey
                            {
                                Key = "abcd",
                                ProjectRole = SFProjectRole.CommunityChecker,
                                ShareLinkType = ShareLinkType.Recipient
                            }
```
I don't see it as a big issue, so if you are happy, I am happy.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1743)* status: 9 of 46 files reviewed, 16 unresolved discussions (waiting on @Nateowami and @pmachapman)

___
*[`src/SIL.XForge/Services/SecurityService.cs` line 10 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NSDHT1n05cctP4F9ckk:-NSi-IGm9X2I_xVZz4Kk:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/d3b4774624d7a766f8ad9edca0a1f4267bd155f2/src/SIL.XForge/Services/SecurityService.cs#L10)):*

How about specifying that this is in bytes? Under "remarks" it says "For example, 12 bytes of data will become 16 bytes of base-64 text", but now that this variable is no longer named "bytes" it isn't as clear that this isn't the final length.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 302 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L5ZVrJVAzJUhgq-r1-302:-NSi4ncD2QpNdpBPsKB-:b5k45zh) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L302)):*

So... we call `loadingStarted()` only *after* the user has been successfully logged in? This seems very odd to me. Also, all the code that follows the first few lines requires the user to be logged in, so rather than wrapping it in a method, which makes it harder to follow the structure, what about something like (I have no idea if this is correct):
``` ts
this.loadingStarted();
const loggedIn = await this.authService.loggedInState$.pipe(filter(state => state.loggedIn), take(1)).toPromise();
```
Increasing the depth of brackets tends to make the logic harder to follow.
</blockquote></details>

Yes, that works well. My initial thoughts were to be able to handle login states in the future but, as there is no need at the moment, there is no point having it setup how I did.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.scss` line 19 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZPj-r1-19:-NSiP_7ZAfM-l9dFN8a7:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.scss#L19)):*

If you set `column-gap: 20px;` above, rather than `gap: 20px;`, then I don't think you would need to set `gap: 0;` for small screens. The documentation for gap says:

> The gap CSS property is a shorthand property for row-gap and column-gap specifying the gutters between grid rows and columns.
> Syntax: <'row-gap'> <'column-gap'>?
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.scss` line 28 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZPj-r1-28:-NSi-bat3KfQipAUtFsE:b-otwdjv) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.scss#L28)):*

This seems oddly specific. Can we not use pixels? Or a round-ish number?
</blockquote></details>

This is a case of not having good theme support. The specific number is generated by the theme we're using rather than something we have specific access to set. I could set a pixel equivalent but wanted to play safe in case different translation font would cause the em value to be greater than a set pixel amount.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts` line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZPl-r1-33:-NSiTtIk3PvqQD7_hgMc:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts#L33)):*

Was it not possible to use a typed form control? The reason we use `UntypedFormControl` is because Angular added strict typing to the form controls, but in order to not break existing applications, migrated us to `UntypedFormControl` so we could transition to strict typing slowly.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts` line 34 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZPl-r1-34:-NSi0nYyBsRPvYOY9jy3:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts#L34)):*

This looks to me like a status variable. Maybe consider `status: 'input' | 'joining' | 'unavailable'`? I think enumerating all possible values and making it required is preferable over having it be optional. This also more closely matches what we do in some other components.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts` line 67 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZPl-r1-67:-NSi2bdB1s6Obc4bMLya:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts#L67)):*

Maybe the method should be renamed `updateOfflineMessage()` or something? It looks like it's being called any time the online status changes, not just when the user is found to be offline.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts` line 155 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZPl-r1-155:-NSi19v933vs3B6zjI-D:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts#L155)):*

I think this should be in a `finally` block.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts` line 177 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZPl-r1-177:-NSi1lJC6tmu-XxXEo5U:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts#L177)):*

This should be given a desciptive name that says what it does without looking at the method. Maybe `informInvalidShareLinkAndRedirect()`? That's pretty wordy, but it's more descriptive.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/anonymous.service.ts` line 15 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZQK-r1-15:-NSmapwnDREOw7DobPJT:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/anonymous.service.ts#L15)):*

See my comment in the other note.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/anonymous.service.ts` line 24 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZQK-r1-24:-NSmaqLa18JapusnJfkA:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/anonymous.service.ts#L24)):*

@Nateowami That is my understanding too.
@nigel-wells The machine-api does not use JSON-RPC - I think JSON is a helpful format for all APIs where practical.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/Controllers/AnonymousController.cs` line 26 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZQV-r1-26:-NSi3phNA_fX0kB1NzyU:bosr0ea) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/src/SIL.XForge.Scripture/Controllers/AnonymousController.cs#L26)):*

Case will not matter for the endpoint resolution - Calling "checkSharingKey" will execute "CheckSharingKey()" - my query is about "Check**Sharing**Key" vs "Check**Share**Key".

I raised this as I tried to see where it was called in the frontend by searching for "CheckSharingKey", and didn't find it. Then I realized a different endpoint name (checkShareKey) was defined in the HttpPost attribute, and searched for that (and found it). I don't mind which name you use ("CheckShareKey" or "CheckSharingKey"), so long as it is consistent (the case of the first letter excluded) when running a find all across the project.
</blockquote></details>

Ah, yes, of course! Changed.
___
*[`test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs` line 347 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NQSg2L6ziPqMOfxqZQs-r1-347:-NShzDgI87YU1FMmHf19:bisr3th) ([raw file](https://github.com/sillsdev/web-xforge/blob/21b6fa798ec50dddb68da1b4dec04d14e3597528/test/SIL.XForge.Scripture.Tests/Services/SFProjectServiceTests.cs#L347)):*

My query is more because this key is defined in the TestEnvironment ProjectSecret repository:
```
                            new ShareKey
                            {
                                Key = "abcd",
                                ProjectRole = SFProjectRole.CommunityChecker,
                                ShareLinkType = ShareLinkType.Recipient
                            }
```
I don't see it as a big issue, so if you are happy, I am happy.
</blockquote></details>

I'm happy in the sense it is typical of the front end tests and in other places of backend tests I've seen. I think creating constants would get very crowded in large test files.

pmachapman (commented):
Looks good. Just one nit - for your classes `CheckShareKeyRequest` and `GenerateAccountRequest`, either instantiate the string properties with `string.Empty`, or change the declarations to `public string? `. Nullable reference types can be annoying to set up, but make things way easier in the long run. (**Done** - thank you!)

> Reviewed 1 of 41 files at r1, 3 of 15 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-:-NSmpq9cDFu8YMQBS3zp:b-ix0jyy)* status: 13 of 46 files reviewed, 12 unresolved discussions (waiting on @Nateowami)

src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts
```diff
@@ -45,40 +65,121 @@ export class JoinComponent extends DataLoadingComponent {
...
+        this.name.enable();
+      }
+    } catch (e) {
+      await this.informInvalidShareLinkAndRedirect();
+    }
+    this.status = 'input';
+  }
+
+  logIn(): void {
+    if (this.status != null) {
```
pmachapman, line 121: When I test the code, status is not null, so clicking the log in link does not work for me?

src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.html
```diff
@@ -1 +1,18 @@
...
+    <form>
+      <mat-form-field appearance="fill">
+        <mat-label>{{ t("your_name") }}</mat-label>
+        <input matInput [formControl]="name" />
+      </mat-form-field>
+      <button mat-flat-button color="primary" type="submit" (click)="joinProject()" [disabled]="!isFormEnabled">
+        {{ t("join") }}
+      </button>
+    </form>
+    <a (click)="logIn()" [innerHtml]="i18nService.translateAndInsertTags('join.login_existing_account')"></a>
```
pmachapman, line 15: Is it possible to make "Log in" a proper hyperlink instead of underlined text? I found the lack of normal hyperlink hover effects confusing, and made me think it was not a real link.

pmachapman: Hmmm the log in link still doesn't feel right to me. This isn't my area of expertise (i.e. this is not a blocker), but would something like this work:

image

If it doesn't feel better to you, just tell me no :-)

pmachapman:
I found an unusual bug while testing this on my machine:

1. Generate a link as admin to share a translation as a community checker
2. Log in anonymously using that link in an other browser
3. Generate another link as admin to share a translation as an observer
4. Try to open that link in the same browser window as you logged in anonymously.

The following error occurs:
image

nigel-wells (commented):
Doh! Done.

Sorry Peter, I see I've had your feedback better sitting here in draft and not published! Regarding your error, currently SF doesn't support a single user having multiple roles. The error is not helpful, I agree. There is another Jira task to provide more relevant error messages rather than using generic ones. I've done the first targeted error message in this PR which will then be the base for improving the other ones.

I've pushed through a new update which includes the new limiter based on how many users can be generated per share key. I've need written any tests yet as I wanted your feedback before writing them - any change you could do that tomorrow @Nateowami ?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-:-NTCCQpyBbrCfX1IhnmM:bt7dye8)* status: 7 of 49 files reviewed, 16 unresolved discussions (waiting on @Nateowami and @pmachapman)

___
*[`src/SIL.XForge/Services/UserService.cs` line 56 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1743#gh-1170730394:-NU3wWQEAONeWp4_0aIf:bbsbrbv) ([raw file](https://github.com/sillsdev/web-xforge/blob/e2b11ffcfd2ce9e2fb30ae594abd7463c3ae5bb0/src/SIL.XForge/Services/UserService.cs#L56)):*

Shouldn't it be `.Every`?
</blockquote></details>

`.Every` is not an option
```
'IEnumerable<JObject>' does not contain a definition for 'Every' and no accessible extension method 'Every' accepting a first argument of type 'IEnumerable<JObject>' could be found
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 292 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1743#gh-1170428268:-NU3x8Nn5nlEbe9XhZaD:bkx5dti) ([raw file](https://github.com/sillsdev/web-xforge/blob/e2b11ffcfd2ce9e2fb30ae594abd7463c3ae5bb0/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L292)):*

This feels like something the auth service should provide. I think a number of components are going to want to know when the user is logged in.
</blockquote></details>

I don't follow - the auth service is providing the state? There is a separate one specifically for `isLoggedIn` but, in this instance, I need it to work for both logged in and logged out.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.html` line 15 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1743#gh-1164644982:-NTCPJwU6Uz4nLMis5Zj:budo7uh) ([raw file](https://github.com/sillsdev/web-xforge/blob/be7a431fd0d740f68daa634ffa1a367ade47aaae/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.html#L15)):*

Hmmm the log in link still doesn't feel right to me. This isn't my area of expertise (i.e. this is not a blocker), but would something like this work:

image

If it doesn't feel better to you, just tell me no :-)
</blockquote></details>

I've added a hover state to login. The intention for this UX decision is the focus here is for making it clear for other users to join a project while offering a subtle way for others who happened to be logged out to still gain access to their account. We did explore other UI options to make it really obvious for either party but sided with  keeping it really simple for one non-technically party who we believe will be the majority of people hitting this form.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts` line 121 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1743#gh-1164639885:-NTCClFN0ixeJjqWyNwg:b-w1u4ai) ([raw file](https://github.com/sillsdev/web-xforge/blob/be7a431fd0d740f68daa634ffa1a367ade47aaae/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts#L121)):*

When I test the code, status is not null, so clicking the log in link does not work for me?
</blockquote></details>

Ah, I missed this one after a refactor. Have updated.

pmachapman (commented):
> Reviewed 1 of 27 files at r2, 3 of 15 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1743)* status: 9 of 49 files reviewed, 14 unresolved discussions (waiting on @Nateowami)

Nateowami (dismissed):
> Reviewed 1 of 41 files at r1, 2 of 15 files at r4, 1 of 16 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1743)* status: 13 of 49 files reviewed, 6 unresolved discussions (waiting on @nigel-wells and @pmachapman)

___
*[`src/SIL.XForge/Services/UserService.cs` line 56 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1743#gh-1170730394:-NU7lCpu0xw0nNVTM484:b-q4rdhb) ([raw file](https://github.com/sillsdev/web-xforge/blob/e2b11ffcfd2ce9e2fb30ae594abd7463c3ae5bb0/src/SIL.XForge/Services/UserService.cs#L56)):*

`.Every` is not an option
```
'IEnumerable<JObject>' does not contain a definition for 'Every' and no accessible extension method 'Every' accepting a first argument of type 'IEnumerable<JObject>' could be found
```
</blockquote></details>

(I wrote this a while back and forgot to post it)

I guess the method I was looking for is [`IEnumerable.All`](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.all)`.

The point I was trying to make is that if the user has multiple identities the user isn't really anonymous anymore, so the variable name and the logic used to compute it don't agree.

I think it would be more correct though to write something like:
``` csharp
bool displayNameSetAtSignup = identities
    .OfType<JObject>()
    .Any(i => (string)i["connection"] == "Transparent-Authentication");
```

And then later on do:
``` csharp
IsDisplayNameConfirmed = displayNameSetAtSignup,
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 292 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1743#gh-1170428268:-NU7nel06v8SLt6FVOW2:bv7j2ee) ([raw file](https://github.com/sillsdev/web-xforge/blob/e2b11ffcfd2ce9e2fb30ae594abd7463c3ae5bb0/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L292)):*

I don't follow - the auth service is providing the state? There is a separate one specifically for `isLoggedIn` but, in this instance, I need it to work for both logged in and logged out.
</blockquote></details>

(I wrote this a while back and forgot to post it)

I mean instead of writing:
``` ts
await this.authService.loggedInState$
  .pipe(
    filter(state => state.loggedIn),
    take(1)
  )
  .toPromise();
```
I think you should write:
``` ts
await this.authService.loggedIn();
```
___
*[`test/SIL.XForge.Scripture.Tests/Services/AnonymousServiceTests.cs` line 17 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NUZNRhwCt7-b9bBEtUv:-NUZNRhwCt7-b9bBEtUw:b-kpj3de) ([raw file](https://github.com/sillsdev/web-xforge/blob/764dc45458ea1fd36ab6a2b62561edd33d610cb5/test/SIL.XForge.Scripture.Tests/Services/AnonymousServiceTests.cs#L17)):*
> ```Smalltalk
> public class AnonymousServiceTests
> {
>     private const int MaxUsers = 50;
> ```

Peter and I talked about this and think it should be bumped to 250. A little abuse of the system is preferable to real users not being able to join.

pmachapman (approved):
Looks good, and the rate limiter functions as expected. :lgtm: 

> Reviewed 8 of 41 files at r1, 3 of 27 files at r2, 4 of 15 files at r3, 8 of 15 files at r4, 12 of 16 files at r5, 5 of 5 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-:-NUZ_rs84-VX_Xi4OSQG:bb6x2hr)* status: all files reviewed, 7 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.spec.ts` line 20 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NUZYC004pYDtfhIUYks:-NUZYC004pYDtfhIUYkt:b-v8u2jz) ([raw file](https://github.com/sillsdev/web-xforge/blob/764dc45458ea1fd36ab6a2b62561edd33d610cb5/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.spec.ts#L20)):*
> ```XML
> import { By } from '@angular/platform-browser';
> import { NoopAnimationsModule } from '@angular/platform-browser/animations';
> import { DebugElement } from '@angular/core';
> ```

Nit I am surprised that lint doesn't pick up how these are out of order?

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1743)* status: 41 of 49 files reviewed, 7 unresolved discussions (waiting on @Nateowami and @pmachapman)

___
*[`src/SIL.XForge/Services/UserService.cs` line 56 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1743#gh-1170730394:-NUsUSbS4Hh3p1AMZ9eY:b-fjvpx3) ([raw file](https://github.com/sillsdev/web-xforge/blob/e2b11ffcfd2ce9e2fb30ae594abd7463c3ae5bb0/src/SIL.XForge/Services/UserService.cs#L56)):*

(I wrote this a while back and forgot to post it)

I guess the method I was looking for is [`IEnumerable.All`](https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.all)`.

The point I was trying to make is that if the user has multiple identities the user isn't really anonymous anymore, so the variable name and the logic used to compute it don't agree.

I think it would be more correct though to write something like:
``` csharp
bool displayNameSetAtSignup = identities
    .OfType<JObject>()
    .Any(i => (string)i["connection"] == "Transparent-Authentication");
```

And then later on do:
``` csharp
IsDisplayNameConfirmed = displayNameSetAtSignup,
```
</blockquote></details>

I see you're point that `isAnonymousUser` could be confusing under other contexts. I've changed to your suggestion. It can be revisited if we found a need to do more with anonymous users.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 292 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1743#gh-1170428268:-NUsSEwH38qvc9VpVJ2r:b4iynif) ([raw file](https://github.com/sillsdev/web-xforge/blob/e2b11ffcfd2ce9e2fb30ae594abd7463c3ae5bb0/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L292)):*

(I wrote this a while back and forgot to post it)

I mean instead of writing:
``` ts
await this.authService.loggedInState$
  .pipe(
    filter(state => state.loggedIn),
    take(1)
  )
  .toPromise();
```
I think you should write:
``` ts
await this.authService.loggedIn();
```
</blockquote></details>

I've moved it into AuthService.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.spec.ts` line 20 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NUZYC004pYDtfhIUYks:-NUsyINi1UCZ59X69fP0:bi5ox0u) ([raw file](https://github.com/sillsdev/web-xforge/blob/764dc45458ea1fd36ab6a2b62561edd33d610cb5/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.spec.ts#L20)):*

Nit I am surprised that lint doesn't pick up how these are out of order?
</blockquote></details>

You're right - looks like we'll need to add some extra settings (and maybe another package) to aid in ordering these automatically. I've manually moved them to the top for now.
___
*[`test/SIL.XForge.Scripture.Tests/Services/AnonymousServiceTests.cs` line 17 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1743#-NUZNRhwCt7-b9bBEtUv:-NUsSKll-u6EaGtPXCbp:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/764dc45458ea1fd36ab6a2b62561edd33d610cb5/test/SIL.XForge.Scripture.Tests/Services/AnonymousServiceTests.cs#L17)):*

Peter and I talked about this and think it should be bumped to 250. A little abuse of the system is preferable to real users not being able to join.
</blockquote></details>

Done.

pmachapman (approved):
> Reviewed 8 of 8 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1743)* status: all files reviewed, 6 unresolved discussions (waiting on @Nateowami)

Nateowami (commented):
> Reviewed 2 of 41 files at r1, 4 of 27 files at r2, 2 of 15 files at r3, 2 of 15 files at r4, 2 of 16 files at r5, 2 of 5 files at r6, 1 of 8 files at r7, 1 of 1 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1743)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

PR #1747, opened 2023-03-15
SF-1908 Fix order of comments that were added in Paratext

nigel-wells:
- Set type for sortBy to use date rather than string

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1747)


nigel-wells (commented):
I've changed to `toJSON()` for consistency - I took the date format set from PT rather than looking which specific method we used when creating a date format. As you've stated, `toJSON()` uses `toISOString()` anyway so the output for testing is the same.

The main purpose of the test is to ensure dates saved from a different time zone still get ordered correctly. The format I came across is the time zone set, saved, and synced when done via PT. Converting the date strings to a date object o lodash `sortBy` allows it to sort correctly as a real date.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1747#-:-NQb-L4B-Mnq7_B-dnvN:b-hfq5co)* status: 0 of 2 files reviewed, all discussions resolved

Nateowami (approved):

PR #1755, opened 2023-03-17
SF-1914 - Remove new note from toolbar and show FAB to admin

nigel-wells:
- Removed button from toolbar
- Allow anyone with edit text rights to see the FAB

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1755)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
@@ -27,16 +27,6 @@
             <mat-icon>settings</mat-icon>
           </button>
         </ng-container>
-        <button
-          *ngIf="canInsertNote"
-          mat-icon-button
-          type="button"
-          id="create-note-btn"
-          (click)="insertNote()"
-          title="{{ t('insert_note') }}"
-        >
-          <mat-icon>add_comment</mat-icon>
-        </button>
         <ng-container *ngIf="canShare">
           <div class="toolbar-separator">&nbsp;</div>
           <app-share-button [defaultRole]="defaultShareRole"></app-share-button>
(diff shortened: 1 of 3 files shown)


RaymondLuong3 (commented):
What is the purpose of allowing translators to see the insert note FAB? I wonder if it will be distracting since it appears anytime a translator or admin begins editing text. I notice in Google docs that the add comment button doesn't appear unless a user selects some text in the document.

My preference is that we enable translators and admins to primarily do scripture drafting, and then have the add note button in the toolbar rather than a FAB that may become annoying for translators not interested in adding a comment. This is more in line with how it is used in Paratext where a user has to select the Insert note option from a menu or by right-clicking on the verse. What are your thoughts?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1755#-:-NR9Wkf80OUegeAg7cQO:b-jb0rh1)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

nigel-wells (commented):
Sorry, I should have added some more commentary around this to support the reason for the change as well as addressing the valid points you've made.

When I started looking at the tool as a whole I found it odd that, as an admin, I couldn't do what commenters could do without discovering the add comment button in the toolbar. This brought up recent discussions we've had in community checking where we locked admin's off from doing tasks which them steps them from both fully experiencing what SF is capable of doing as well as potentially causing confusion around admin training commenters where admin can't do the same thing as a commenter - I do understand that they can both technically do the same thing (add a comment) but in different ways. 

After talking with Natheniel he came around to agree with the sentiment around having a consistent experience for users in terms of where we are placing buttons, what they are called, and when they appear.

This then raised the valid points around the FAB appearing when it might come across as annoying to translators. I created a new Jira card to help address these concerns - SF-1922. I've currently not included it under the Epic as I was on the fence about if we want to address those as part of our initial release or wait until after - I've not heard from Natheniel as to what his opinion is and fully welcome if you think we should do it now or later as well as if you have other ideas to improve.

Nathaniel had also mention that when we first starting introducing the notes feature the focus was on translators and now the focus has shifted to be more on commenters which has been the main driver for some of these recent UI changes.

Does that help explain things? I'm happy to chat through anything with you.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1755#-:-NRADvra-NL0iBJe81KU:b-fy9uis)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1755)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

PR #1648, opened 2022-12-19
SF-1810 Allow inviting users without sending invitation email

nigel-wells:
- Created a dedicated sharing dialog separate from the share control
- Removed link sharing from the users component
- Moved Email sharing on the users component to the bottom
- Added a link to the new share dialog on the users component
- Removed project setting for sharing level i.e. anyone or email only
- Admin can now share even when sharing is disabled
- Enabling sharing enables the share with 'anyone' option

**DATA MIGRATION**
* A migration is setup to update realtime docs
* Another migration is required to be run manually on the `sf_project_secrets` collection. There is a file stored in`mongodb/Migrations/AddRecipientShareLinkType.mongodb` which will need to be run after being deployed to QA and again to live. The app will continue to run without errors but it will treat any `Email` share keys as being available for anyone to use rather than restricted to a single recipient.

**USERS PAGE**
image

**SHARE DIALOG**
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1648)

PR #1763, opened 2023-03-23
SF-1930 Replace share icon with add user icon

Nateowami:
### Rationale
- I think the "add user" icon is a lot easier to understand than the "share" icon that is supposed to evoke the idea of a network. Some users may recognize it but I think it's less helpful in most cases.
- It probably can't be used in every situation that the share icon would be used, since it's possible to share something without adding a user to it.

### Before
share_button_with_text_before
share_icon_button_before
### After
share_button_with_text_after
share_icon_button_after
In RTL mode:
share_button_with_text_after_rtl
I don't think flipping the icon is that important, but it's trying to convey "add person", so I think having the "+" come first makes the most sense.

I removed some CSS that wasn't actually being used because it wasn't imported into the component. It got added very recently in #1648.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1763)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-button.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-button.component.html
@@ -1,8 +1,8 @@
 <ng-container *transloco="let t; read: 'share'">
   <button *ngIf="iconOnlyButton" (click)="openDialog()" mat-icon-button title="{{ t('share') }}" id="share-btn">
-    <mat-icon>share</mat-icon>
+    <mat-icon class="mirror-rtl">person_add</mat-icon>
   </button>
-  <button *ngIf="!iconOnlyButton" mat-flat-button (click)="openDialog()" color="primary">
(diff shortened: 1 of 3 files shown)


RaymondLuong3 (approved):
I think you are right. I have only recently started recognizing the share icon, but for the longest time I've just ignored it. The add user icon is more intuitive, the user will know that it is for adding a user to the project.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1763#-:-NRZY0Aw6Ll7PBgWenoW:boig3sd)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

RaymondLuong3 (approved):
I'm not sure if it is worth testing, but since there is already an issue, and that RTL has an effect on this, let's give this one to testers

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1763#-:-NRZYUWn7yNMgsLNIkuG:b1yp1jd)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1767, opened 2023-03-27
Add Storybook

Nateowami:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1767)


marksvc:
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 1 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRjOvNq-PCV8_m04MHb:-NRjOvNq-PCV8_m04MHc:bxmrf6n) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L1)):*
> ```XML
> import { HttpClientModule } from '@angular/common/http';
> ```

Reload to .storybook directory?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1767)* status: 0 of 28 files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 1 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRjOvNq-PCV8_m04MHb:-NRkJ2G64yShP7FenP9S:b-s0iqxq) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L1)):*

Reload to .storybook directory?
</blockquote></details>

I tried this, unsuccessfully. I can't get the TSConfig to work with it in the `.storybook` directory.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 56 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRkIhwhAp8eem0678va:-NRkIhwhAp8eem0678vb:bpe70mm) ([raw file](https://github.com/sillsdev/web-xforge/blob/65c53586afdc6f5bee327501b6c41ffe135565b3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L56)):*
> ```XML
>     { provide: TRANSLOCO_CONFIG, useValue: translocoConfig },
>     { provide: TRANSLOCO_LOADER, useClass: TranslationLoader },
>     { provide: IGNORE_COOKIE_LOCALE, useValue: true }
> ```

I attempted to swap this out for a mock of the CookieService, but not successfully. I think it was still ending up injecting the real service, since that's provided in the AppModule.

There's probably a way to do this, but it's probably more complicated and I think it makes sense to merge as-is for now.

marksvc (changes requested):
Thanks for your work integrating and getting this to work in our codebase. I have just a few files left that I haven't reviewed, but here are comments thus far.

> Reviewed 22 of 27 files at r1, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-:-NRok6zC9XQ5hR7M3fYF:bk2bbbj)* status: 22 of 28 files reviewed, 21 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/.gitignore` line 36 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRnWqDAAhWK7V_THIRD:-NRnWqDB8BwuJ6IEVfGH:b-4ekt1f) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/.gitignore#L36)):*
> ```ignore
> npm-debug.log
> testem.log
> /documentation.json
> ```

Wow, that file is huge. I see that .storybook/preview.js references it. Do you think it would be possible to put it more out of the way into the .storybook/ directory ?
___
*[`src/SIL.XForge.Scripture/ClientApp/.storybook/preview.js` line 30 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRnceNzBRnDmxAUP2j3:-NRnceNzBRnDmxAUP2j4:b9ton79) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/.storybook/preview.js#L30)):*
> ```JavaScript
>         title: locale.englishName,
>         value: locale.canonicalTag
>       }))
> ```

I was a bit puzzled by 
```ts
map(locale => ({
  title: .., value: .. 
}))
```
. In playing with it, I determined that if we just wrote
```ts
map(locale => {
  title: .., value: ..
})
```
it thinks that the lambda's body is the block delineated by the `{}` characters. But what we want is to make an _object_ using the `{}` characters. And putting the `{}` inside parentheses, as `({})` allows us to specify the lambda as giving the object in parentheses. 

Is this correct?



_Code quote:_
```JavaScript
      items: getI18nLocales().map(locale => ({
        title: locale.englishName,
        value: locale.canonicalTag
      }))
```
___
*[`src/SIL.XForge.Scripture/ClientApp/.storybook/tsconfig.json` line 11 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRnfTYm2Knd9qKIBewl:-NRnfTYm2Knd9qKIBewm:b-lb3flb) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/.storybook/tsconfig.json#L11)):*
> ```jsonc
>     "../src/**/*.spec.ts",
>     "../projects/**/*.spec.ts",
>     "../src/xforge-common/test-utils.ts",
> ```

Ok. Tho all of these are already captured in ../src/tsconfig.app.json (with the exception of mock-console.ts). And if we list them here, then when we edit this list some day we may be confused why it doesn't behave like we expect (since the list is redundant with the parent/base). Maybe omit?



_Code quote:_
```jsonc
    "../src/test.ts",
    "../src/**/*.spec.ts",
    "../projects/**/*.spec.ts",
    "../src/xforge-common/test-utils.ts",
```
___
*[`src/SIL.XForge.Scripture/ClientApp/.storybook/tsconfig.json` line 14 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRngDtv5QcKR8ZULfNI:-NRngDtv5QcKR8ZULfNJ:b-1tat97) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/.storybook/tsconfig.json#L14)):*
> ```jsonc
>     "../src/xforge-common/mock-console.ts"
>   ],
>   "include": ["../src/**/*", "../projects/**/*"],
> ```

So I don't have a `projects` directory. Am I missing out on something?



_Code quote:_
```jsonc
 "../projects/**/*"]
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comment-form/checking-comment-form.stories.ts` line 34 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRo3G6bCwTDJ8uAng28:-NRo3G6bCwTDJ8uAng29:bfukr9v) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comment-form/checking-comment-form.stories.ts#L34)):*
For these three `await`s, VSCode tells me, 
> 'await' has no effect on the type of this expression.ts(80007)

Are these needed?



_Code quote:_
```XML
  const saveButton = await canvas.getByRole('button', { name: /Save/i });
  await userEvent.click(saveButton);
  const error = await canvas.getByText(/You need to enter your comment before saving/i);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.html` line 8 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRo4Xe12OEdqLFYlE8C:-NRo4Xe12OEdqLFYlE8D:brzn409) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.html#L8)):*
> ```HTML
>     
> ```

So what's the purpose of having this here again? If this is a work-in-progress comment, let's omit it.

And same for the commented code for chapter below.



_Code quote:_
```HTML
    <!-- <select matNativeControl [(ngModel)]="book" (valueChange)="bookChanged()" panelClass="book-select-menu">
      <option *ngFor="let b of books" [value]="b">{{ bookName(b) }}</option>
    </select> -->
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.html` line 34 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRo5UOeDYD7PIlBTkyf:-NRo5UOeDYD7PIlBTkyg:be9xjfi) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.html#L34)):*
> ```HTML
>     fxHide.xs
>   >
>     <mat-icon>keyboard_arrow_{{ i18n.forwardDirectionWord }}</mat-icon>
> ```

It seems to me that the _effect_ of the  buttons should be changed in the RTL situation, rather than their icon direction. So for example, rather than changing 
```
Ruth 1 < >
```
to
```
          > < 1 htuR
```
it could instead be
```
          < > 1 htuR
```
where the `<` button now goes forward rather than backward.

But I also haven't done research to know what's expected. 



_Code quote:_
```HTML
{{ i18n.forwardDirectionWord }}</
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts` line 9 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoLPwnDDpC5xuRnGt3:-NRoKo8jE9KohMhxPsvE:b-e4hfik) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts#L9)):*
> ```XML
>   styleUrls: ['./book-chapter-chooser.component.scss']
> })
> export class BookChapterChooserComponent {
> ```

I know this class name is pretty descriptive for the amount of UI this component has, but I think we should have a comment on all classes in our code. So for example, a class comment here could be
```
/** Component allowing book and chapter selection and navigation. */
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts` line 21 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoMITtAYHiwaYf6Wrx:-NRoMITtAYHiwaYf6Wry:b-swqw9l) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts#L21)):*
> ```XML
> 
>   bookName(b: number): string {
>     return b == null ? '' : this.i18n.localizeBook(b);
> ```

Can't `b` not be null here because the type is `number`, not `number|null`?



_Code quote:_
```XML
 b == null ? '' 
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts` line 17 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoPXmC5-68djHih7e-:-NRoPXmC5-68djHih7e0:b2nmpwr) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts#L17)):*
> ```XML
>   argTypes: {
>     book: {
>       options: arrayOfIntsFromOne(Canon.allBookIds.length),
> ```

The `@Input() book?` allows undefined. That would ideally be something to include in this list. I went to the list to try to pick an empty selection (before getting to this file in review) but it didn't seem to have such an option.



_Code quote:_
```XML
 options: arrayOfIntsFromOne
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts` line 20 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoPB6A2H7fj-wfSFz0:-NRoPB6A2H7fj-wfSFz1:b-e7hjrj) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts#L20)):*
> ```XML
>       control: 'select'
>     },
>     chapter: { control: 'select', options: arrayOfIntsFromOne(10) },
> ```

150 would let us get to the highest valid number. 151 would let us see what the control does when fed the wrong thing. I suggest specifying 151 instead of 10. What do you think?

Actually, because the input is `@Input() chapter?`, undefined would be something that could be useful to include in the list here as well.
I guess we could write a separate test story down below for situations where book or chapter is undefined or out of an assumed range.



_Code quote:_
```XML
10
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts` line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoQWKq1ajvBV3r5QNO:-NRoQWKq1ajvBV3r5QNP:b-3x9b8z) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts#L33)):*
> ```XML
> 
> const defaultArgs = {
>   books: arrayOfIntsFromOne(66),
> ```

You went to the trouble to use Canon.allBookIds.length for the book options above. I was going to suggest to use a similar thing to fetch the number of books in the Bible but it looks like we don't have such. 



_Code quote:_
```XML
66
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts` line 59 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoRRqiB69b_9LFg_bM:-NRoRRqiB69b_9LFg_bN:b-vn3wcc) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts#L59)):*
> ```XML
>   const nextButton = await canvas.findByTitle('Next chapter');
>   const prevButton = await canvas.findByTitle('Previous chapter');
>   // Jest types aren't working
> ```

This would benefit from a bit of clarity about what you are meaning about the below. I suggest 
```ts
// Using `as any` since types aren't working in Jest.
```

Also, I am still learning about these things, but it looks like the expected types may be for "@testing-library", so maybe better might be the following?

```ts
// Using `as any` since testing-library types aren't working.
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 31 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoZF-28a9VcEdG7TDk:-NRoZF-28a9VcEdG7TDl:bf6xkim) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L31)):*
> ```XML
>       .pipe(
>         distinctUntilChanged(),
>         delay(100), // hideous but necessary
> ```

Can you elaborate in your comment about why it is necessary?



_Code quote:_
```XML
// hideous but necessary
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 33 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoZe0zCrPcrFUrqu4q:-NRoZe0zCrPcrFUrqu4r:b-j2kuwh) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L33)):*
> ```XML
>         delay(100), // hideous but necessary
>         tap(() => {
>           document.body.setAttribute('dir', getLocaleDir(transloco.getActiveLang()));
> ```

It looks like when localizationInit() is called, if the local is for a rtl language, we will `setActiveLang()` but not end up doing `document.body.setAttribute('dir'` without the language again changing (and triggering `.langChange$`). Will that be okay? It seem like maybe we should set up our subscription to `.langChange$` first and then call `setActiveLang()`. 



_Code quote:_
```XML
document.body.setAttribute('dir', getLocaleDir(transloco.getActiveLang()));
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 41 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRo_Kfj6okPHsbyGDn4:-NRo_Kfj6okPHsbyGDn5:b-6ns5dw) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L41)):*
> ```XML
> }
> 
> export const I18nStoryDecorator: any = (Story: any, context: any) => {
> ```

That's a lot of `any` usage :). Important?



_Code quote:_
```XML
 any = (Story: any, context: any
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 43 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRo_iImASF_qhFg6DHh:-NRo_iImASF_qhFg6DHi:bjaen4c) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L43)):*
> ```XML
> export const I18nStoryDecorator: any = (Story: any, context: any) => {
>   // In some cases the locale has been known to be the empty string, so make sure not to set it to that.
>   const locale = context.globals.locale || I18nService.defaultLocale.canonicalTag;
> ```

I think when this has come up in the past I haven't been very persuasive with my concern :), but I'll mention it again. I think using `||` here falls into a JavaScript gotcha, because it's going to fall back to `canonicalTag` if `context.globals.locale` is **falsy**.
That happens (imo unfortunately) for the case of the empty string, and also happens for things like 0. So I think that avoiding this aspect of JavaScript and writing something like the following is better here:
```ts
const locale = context.globals.locale !== '' ? context.globals.locale : I18nService.defaultLocale.canonicalTag;
```
or maybe
```ts
let locale = context.globals.locale;
if (locale === '') locale = I18nService.defaultLocale.canonicalTag;
```

Writing `const locale = context.globals.locale || I18nService.defaultLocale.canonicalTag;` is like writing
```ts
const locale == context.globals.locale != false ? context.globals.locale : I18nService.defaultLocale.canonicalTag;
```
(which is comparing a string to a boolean, and so a mismatch of types when comparing.)

Your code is _cleaner_, and my suggested change to the code makes it _uglier_. But because the use of `||` uses JavaScript "falsy", I think it should be avoided and maybe should flag a lint error :-)

I think the last time I went to write on this I may have found that, Oh I do this too. But I can't remember where that would be right now, so I'm saying something against it :-)



_Code quote:_
```XML
 context.globals.locale ||
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 43 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRofQp4EDvvR-v1WMmx:-NRofQp4EDvvR-v1WMmy:b-9lanl9) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L43)):*
> ```XML
> export const I18nStoryDecorator: any = (Story: any, context: any) => {
>   // In some cases the locale has been known to be the empty string, so make sure not to set it to that.
>   const locale = context.globals.locale || I18nService.defaultLocale.canonicalTag;
> ```

No explicit type?



_Code quote:_
```XML
 const locale =
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 45 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRogbkF8vBwIHjMDH_f:-NRogbkF8vBwIHjMDH_g:b-opo5ww) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L45)):*
> ```XML
>   const locale = context.globals.locale || I18nService.defaultLocale.canonicalTag;
>   selectedLocale = locale;
>   if (translocoService != null) translocoService.setActiveLang(locale);
> ```

Having both I18nStoryDecorator and localizationInit do this seems like there is some redundancy. But I suppose whatever flow of what is happening relies on these both being able to `setActiveLang()`?

I haven't attached the debugger to Storybook yet to watch when each of these methods gets run thru :)



_Code quote:_
```XML
.setActiveLang(
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 56 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoiTKl92y5ZEH94VEL:-NRoiTKl92y5ZEH94VEM:bnapr3k) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L56)):*
> ```XML
>     { provide: TRANSLOCO_CONFIG, useValue: translocoConfig },
>     { provide: TRANSLOCO_LOADER, useClass: TranslationLoader },
>     { provide: IGNORE_COOKIE_LOCALE, useValue: true }
> ```

This is interesting, so here you are inventing a new variable that then gets injected with this value to the `I18nService` constructor. Is that so?



_Code quote:_
```XML
{ provide: IGNORE_COOKIE_LOCALE, useValue: true }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/test-utils.ts` line 155 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRnoDSP3iMSI75xTgQf:-NRnoDSP3iMSI75xTgQg:b-xq5a1r) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/test-utils.ts#L155)):*
> ```XML
> }
> 
> export function arrayOfIntsFromZero(size: number): number[] {
> ```

I don't think of `int` as a TypeScript thing. But I won't suggest here that you make the function name even longer to say `Numbers` :)



_Code quote:_
```XML
Ints
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRnnJBf352n2Aqy_QPD:-NRnnJBf352n2Aqy_QPE:b7k3ewn) ([raw file](https://github.com/sillsdev/web-xforge/blob/90021f27bae60125d5b3def0ef9a1646dbc0caae/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html#L3)):*
> ```HTML
> <div [ngClass]="{ outline }" [class]="type">
>   <mat-icon *ngIf="icon">{{ icon }}</mat-icon>
>   <span>{{ notice }}</span>
> ```

(So I'm waiting on this to be adjusted by rebasing onto an improved mechanism.)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1767)* status: 21 of 28 files reviewed, 20 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/.gitignore` line 36 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRnWqDAAhWK7V_THIRD:-NRsB5o87IuQeTorTzkJ:b-drhoie) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/.gitignore#L36)):*

Wow, that file is huge. I see that .storybook/preview.js references it. Do you think it would be possible to put it more out of the way into the .storybook/ directory ?
</blockquote></details>

The `.storybook` directory is intended for configuration, not build artifacts, so it seems like an odd choice. It's possible compodoc has a setting that allows specifying a different output directory, but I don't see a lot of value in doing that. I think it's possible we'll end up using the documentation feature at some point, though I've yet to see much value in it.
___
*[`src/SIL.XForge.Scripture/ClientApp/.storybook/preview.js` line 30 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRnceNzBRnDmxAUP2j3:-NRs9InQ6XWvxngA63K9:bvqgv2l) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/.storybook/preview.js#L30)):*

I was a bit puzzled by 
```ts
map(locale => ({
  title: .., value: .. 
}))
```
. In playing with it, I determined that if we just wrote
```ts
map(locale => {
  title: .., value: ..
})
```
it thinks that the lambda's body is the block delineated by the `{}` characters. But what we want is to make an _object_ using the `{}` characters. And putting the `{}` inside parentheses, as `({})` allows us to specify the lambda as giving the object in parentheses. 

Is this correct?


</blockquote></details>

Yes. And this is a common pattern. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#function_body
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.html` line 8 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRo4Xe12OEdqLFYlE8C:-NRs9SIo6msvHcRZPOAA:b-jgunlz) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.html#L8)):*

So what's the purpose of having this here again? If this is a work-in-progress comment, let's omit it.

And same for the commented code for chapter below.
</blockquote></details>

Removed. I was waiting for another change to be merged in so I could update my branch without conflicts.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.html` line 34 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRo5UOeDYD7PIlBTkyf:-NRs9YUA98X3SqWx3_7i:bfckzae) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.html#L34)):*

It seems to me that the _effect_ of the  buttons should be changed in the RTL situation, rather than their icon direction. So for example, rather than changing 
```
Ruth 1 < >
```
to
```
          > < 1 htuR
```
it could instead be
```
          < > 1 htuR
```
where the `<` button now goes forward rather than backward.

But I also haven't done research to know what's expected. 

</blockquote></details>

When the document is RTL, the first arrow specified will end up being on the right, so if we didn't swap the directions here we would end up with `> <`.

Instead of thinking of pointing to the right or pointing to the left, think in terms of forward or backward and it all makes sense. We always want the first arrow to point backward, regardless of direction. And we always want the second arrow to point forward, regardless of what direction that is.

This logic is copied from the currently-in-use ChapterNav component.

~You can also test in Storybook and see that the behavior is correct.~ Welp, nope, actually the story is not configured well enough to get the correct behavior. The I18n service isn't aware of the current locale being Arabic and so the arrows point the wrong direction. I think that should be fixed at a later time though. It's not a bug in the component, just a test that isn't set up completely correctly.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 31 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoZF-28a9VcEdG7TDk:-NRsBhbT83mbdmnmBumq:biyjvyr) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L31)):*

Can you elaborate in your comment about why it is necessary?
</blockquote></details>

Because "it doesn't work" without it.

My conjecture is that Transloco emits the language change event when a new language has been selected, not when the translations for it have been successfully loaded. This would produce a race condition. I tried to make it pre-load languages, but even then I couldn't get rid of a race condition.

Transloco is supposed to trigger a re-render automatically when the language changes, but it the context of a story. I'm not entirely sure why. So we're waiting a brief amount of time and then telling Angular to detect changes globally.

I now see that Transloco has an `events$` property that can emit a `translationLoadSuccess` event. That might be worth investigating.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 56 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoiTKl92y5ZEH94VEL:-NRsHyfP9UDJ3fELBM8f:blg5gcr) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L56)):*

This is interesting, so here you are inventing a new variable that then gets injected with this value to the `I18nService` constructor. Is that so?
</blockquote></details>

Yes. And it's optional, so if no value is provided the I18nService assumes the default of not ignoring the cookie.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/test-utils.ts` line 155 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRnoDSP3iMSI75xTgQf:-NRs741j3ej9AZcSOB8M:bn6zfaa) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/test-utils.ts#L155)):*

I don't think of `int` as a TypeScript thing. But I won't suggest here that you make the function name even longer to say `Numbers` :)
</blockquote></details>

Integers aren't a JavaScript concept, but that doesn't mean we can't have integers. A `number[]` can contain non-integers, but we want an array of integers, so that's what we specify. We could have a function called `arrayOfFiveWords()` that returns `['the', 'quick', 'brown', 'fox', 'jumped']`, and the fact that neither JavaScript nor TypeScript have a "word" type is not particularly relevant.

Side note: Actually JavaScript *does* have typed arrays that only contain integers, mainly so you can write highly performant code if needed (I think that's why at least). But that's also completely beside the point.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRnnJBf352n2Aqy_QPD:-NRs71zu5vFnyRkd4o_4:bxhm2lg) ([raw file](https://github.com/sillsdev/web-xforge/blob/90021f27bae60125d5b3def0ef9a1646dbc0caae/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html#L3)):*

(So I'm waiting on this to be adjusted by rebasing onto an improved mechanism.)
</blockquote></details>

This has been updated.

marksvc (commented):
> Reviewed 1 of 27 files at r1, 5 of 5 files at r2, 1 of 1 files at r4, 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1767)* status: all files reviewed, 11 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.html` line 34 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRo5UOeDYD7PIlBTkyf:-NS6uvaK3lTPP1uFe-Kl:b-cmw3tn) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.html#L34)):*

When the document is RTL, the first arrow specified will end up being on the right, so if we didn't swap the directions here we would end up with `> <`.

Instead of thinking of pointing to the right or pointing to the left, think in terms of forward or backward and it all makes sense. We always want the first arrow to point backward, regardless of direction. And we always want the second arrow to point forward, regardless of what direction that is.

This logic is copied from the currently-in-use ChapterNav component.

~You can also test in Storybook and see that the behavior is correct.~ Welp, nope, actually the story is not configured well enough to get the correct behavior. The I18n service isn't aware of the current locale being Arabic and so the arrows point the wrong direction. I think that should be fixed at a later time though. It's not a bug in the component, just a test that isn't set up completely correctly.
</blockquote></details>

Ahh. Gotcha. I see that in the Scripture text view area, with Arabic as the localization, we get the equivalent of
```
        < > 1 htuR
```
Good.
Ironic that trying it out in Storybook to see how it behaved, showed the wrong thing :)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.stories.ts` line 32 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NS7-bNy3v5YyluHkFB-:-NS7-bNy3v5YyluHkFB0:btmf6qd) ([raw file](https://github.com/sillsdev/web-xforge/blob/cd539f34ba2ad78b14aea115554a40b2ce39464a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.stories.ts#L32)):*
> ```XML
> WarningWithIcon.args = { ...Warning.args, icon: 'warning' };
> 
> export const Error = Template.bind({});
> ```

Hmm, it's kind of too bad that these tests don't show up in the same order in the Storybook UI, that they are written in here. But alphabetical is helpful, too.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 118 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NS7VZDK4fEFMBhuyDAY:-NS7VZDK4fEFMBhuyDAZ:b-v7wjao) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L118)):*
> ```XML
>     if (urlLocale != null) {
>       this.trySetLocale(urlLocale);
>     } else if (!ignoreCookieLocale) {
> ```

BTW an alternative would be to put this check in `trySetLocale` itself. Tho it looks like `trySetLocale` isn't called from all over the place at present. I see a call from app.component.ts that probably doesn't matter for the stories.



_Code quote:_
```XML
if (!ignoreCookieLocale)
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/test-utils.ts` line 155 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRnoDSP3iMSI75xTgQf:-NS7WPcqD1AVCGl3lHdu:by998u6) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/test-utils.ts#L155)):*
> but that doesn't mean we can't have integers

True.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1767)* status: all files reviewed, 11 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 43 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRo_iImASF_qhFg6DHh:-NS7bK6u1oD5-0aqWWyE:bgcv4k1) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L43)):*
> and maybe should flag a lint error :-)

Oh, it looks like `"@typescript-eslint/strict-boolean-expressions": "error",` flags this and other situations where non-booleans are being compared like booleans.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1767)* status: all files reviewed, 11 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 43 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRo_iImASF_qhFg6DHh:-NS7c5Px89VNzHk4ISA-:b-dhazj4) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L43)):*

> and maybe should flag a lint error :-)

Oh, it looks like `"@typescript-eslint/strict-boolean-expressions": "error",` flags this and other situations where non-booleans are being compared like booleans.
</blockquote></details>

(Oops, reference: https://typescript-eslint.io/rules/strict-boolean-expressions/ )

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1767)* status: 23 of 29 files reviewed, 10 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/.storybook/tsconfig.json` line 11 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRnfTYm2Knd9qKIBewl:-NS8nnzw-vz-Z-q93adY:b-3yga5d) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/.storybook/tsconfig.json#L11)):*

Ok. Tho all of these are already captured in ../src/tsconfig.app.json (with the exception of mock-console.ts). And if we list them here, then when we edit this list some day we may be confused why it doesn't behave like we expect (since the list is redundant with the parent/base). Maybe omit?
</blockquote></details>

It's not redundant. I tested it. I'm guessing the lists aren't merged. If the test files are included in the build, it will cause failures.

I added a comment for why we're ignoring to test util files.
___
*[`src/SIL.XForge.Scripture/ClientApp/.storybook/tsconfig.json` line 14 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRngDtv5QcKR8ZULfNI:-NS8nXsU3ZKqxYtLxtyM:b-b6bill) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/.storybook/tsconfig.json#L14)):*

So I don't have a `projects` directory. Am I missing out on something?
</blockquote></details>

Good point. I wonder whether some Angular projects have it, and so Storybook just adds it when generating the file. Removed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comment-form/checking-comment-form.stories.ts` line 34 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRo3G6bCwTDJ8uAng28:-NS8MulO-sLt3bIqy4iK:btw1j0m) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comment-form/checking-comment-form.stories.ts#L34)):*

For these three `await`s, VSCode tells me, 
> 'await' has no effect on the type of this expression.ts(80007)

Are these needed?
</blockquote></details>

Good catch. Interesting that `canvas.getByRole()`returns `HTMLElement`, but `canvas.findByTitle()` returns `Promise<HTMLElement>`. Changed a bunch of these.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts` line 9 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoLPwnDDpC5xuRnGt3:-NS8RYwSDI0x5RgawtcM:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts#L9)):*

I know this class name is pretty descriptive for the amount of UI this component has, but I think we should have a comment on all classes in our code. So for example, a class comment here could be
```
/** Component allowing book and chapter selection and navigation. */
```

</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts` line 21 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoMITtAYHiwaYf6Wrx:-NS8RZlzDk_IRAkGyEVL:b-m8vgpr) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts#L21)):*

Can't `b` not be null here because the type is `number`, not `number|null`?
</blockquote></details>

Good catch. The check came from the ChapterNav component, where it actually was needed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts` line 17 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoPXmC5-68djHih7e-:-NS8RhoY9yMUnaOOu3JM:bsqir8j) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts#L17)):*

The `@Input() book?` allows undefined. That would ideally be something to include in this list. I went to the list to try to pick an empty selection (before getting to this file in review) but it didn't seem to have such an option.
</blockquote></details>

`undefined` is not a valid input. We just can't guarantee with type safety that it will be set. And it won't get set at the time the component is constructed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts` line 20 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoPB6A2H7fj-wfSFz0:-NS8SDwG3L7R-Zm9oO13:b-k6g2eh) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts#L20)):*

150 would let us get to the highest valid number. 151 would let us see what the control does when fed the wrong thing. I suggest specifying 151 instead of 10. What do you think?

Actually, because the input is `@Input() chapter?`, undefined would be something that could be useful to include in the list here as well.
I guess we could write a separate test story down below for situations where book or chapter is undefined or out of an assumed range.
</blockquote></details>

It doesn't seem like that useful of a test to me. This component knows nothing about the canon, or number of chapters in each book.

Actually, looking at this more, I don't think these controls need to be defined this way. I was experimenting with it at one point, but leaving it with the defaults is probably preferable.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts` line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoQWKq1ajvBV3r5QNO:-NS8TZtE3hVot43Ul3mp:b-mmbp8d) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts#L33)):*

You went to the trouble to use Canon.allBookIds.length for the book options above. I was going to suggest to use a similar thing to fetch the number of books in the Bible but it looks like we don't have such. 
</blockquote></details>

We only localize the 66 canonical books, so I'm updating the test to not use the full canon length from Canon.ts.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.stories.ts` line 32 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NS7-bNy3v5YyluHkFB-:-NS8Tk5xBeazYYKO5lnc:bm89xz1) ([raw file](https://github.com/sillsdev/web-xforge/blob/cd539f34ba2ad78b14aea115554a40b2ce39464a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.stories.ts#L32)):*

Hmm, it's kind of too bad that these tests don't show up in the same order in the Storybook UI, that they are written in here. But alphabetical is helpful, too.
</blockquote></details>

Yes. They're supposed to show up in the order they're defined. I think webpack might be messing that up. I'm still hoping we can fix this at some point.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 33 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoZe0zCrPcrFUrqu4q:-NS8WKDf7fytJIcuxDnh:b-7rj4cd) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L33)):*

It looks like when localizationInit() is called, if the local is for a rtl language, we will `setActiveLang()` but not end up doing `document.body.setAttribute('dir'` without the language again changing (and triggering `.langChange$`). Will that be okay? It seem like maybe we should set up our subscription to `.langChange$` first and then call `setActiveLang()`. 
</blockquote></details>

Good catch. It seemed to work as written, but is probably more correct reversing the order.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 41 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRo_Kfj6okPHsbyGDn4:-NS8hDwxAmnZ3YSbabOF:bbzursm) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L41)):*

That's a lot of `any` usage :). Important?
</blockquote></details>

I don't like to declare types unless I actually have some way of guaranteeing that they're correct. After some digging I discovered I can `import { DecoratorFunction } from '@storybook/csf'`.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 43 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRo_iImASF_qhFg6DHh:-NS8iPBM8Mnn6R_iVS13:bdlipky) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L43)):*
My comment says "In some cases the locale has been known to be the empty string". What I meant was "in addition to checking for null/undefined, we also need to consider the empty string, hence the use of || instead of ??"

While I generally agree that we should make our code as specific as possible, I'm *intentionally* including more values in this check.

> Writing const locale = context.globals.locale || I18nService.defaultLocale.canonicalTag; is like writing
`const locale == context.globals.locale != false ? context.globals.locale : I18nService.defaultLocale.canonicalTag;`

That wouldn't handle null. I'm trying to handle null, undefined, and the empty string. Usually we only want to handle null and undefined, so nullish coalescing makes sense (`a ?? b`). But in this case I found I needed to include the empty string as well.

An alternative would be:
``` ts
const locale = context.globals.locale == null || context.globals.locale === '' ? I18nService.defaultLocale.canonicalTag : context.globals.locale;
```

But there's no value in being that specific. We're expecting a non-empty string, so it would make more sense to write:
``` ts
const locale = typeof context.globals.locale === 'string' && context.globals.locale !== '' ? context.globals.locale : I18nService.defaultLocale.canonicalTag;
```

This would be the actual defensive code. But that seemed more complicated than needed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 43 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRofQp4EDvvR-v1WMmx:-NS8km-K6SHJZTjab2Je:b-h0vkyn) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L43)):*

No explicit type?
</blockquote></details>

Well, since I didn't do an actual check of the type we ended up with, I can't be 100% sure it's a string as I expect. Adding a type when the input was `any` is basically just telling the type system "trust me, this is a string". Whereas if I don't, then it just leaves the type as any, and it's more clear that we haven't proved what the type is.

Declaring it as string would in effect obfuscate what we do and don't know.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 45 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRogbkF8vBwIHjMDH_f:-NS8lrHX6SCXQfvYqEfr:b-geyzo2) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L45)):*

Having both I18nStoryDecorator and localizationInit do this seems like there is some redundancy. But I suppose whatever flow of what is happening relies on these both being able to `setActiveLang()`?

I haven't attached the debugger to Storybook yet to watch when each of these methods gets run thru :)
</blockquote></details>

Yes, it feels redundant. Hopefully we can simplify this somewhat. I wrote it this way to avoid any race conditions, and I know I was having trouble getting it working before doing this.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 118 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NS7VZDK4fEFMBhuyDAY:-NS8mAttDsVSvGaNzDJo:bcpttih) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L118)):*

BTW an alternative would be to put this check in `trySetLocale` itself. Tho it looks like `trySetLocale` isn't called from all over the place at present. I see a call from app.component.ts that probably doesn't matter for the stories.
</blockquote></details>

Tthat's a good point. We may be able to make it more robust. However, if I can figure out how to do it, I'd rather just mock the cookie service. I wasn't able to figure that out when I tried.

marksvc (approved):
> Reviewed 6 of 6 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1767)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/.storybook/tsconfig.json` line 11 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRnfTYm2Knd9qKIBewl:-NShKA-zBLOBEHLTXZ9h:bbmzeg0) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/.storybook/tsconfig.json#L11)):*

It's not redundant. I tested it. I'm guessing the lists aren't merged. If the test files are included in the build, it will cause failures.

I added a comment for why we're ignoring to test util files.
</blockquote></details>

Ahh, interesting. Okay.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comment-form/checking-comment-form.stories.ts` line 34 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRo3G6bCwTDJ8uAng28:-NShKLLMCAkw_hp2Wmqo:b-fr0ia6) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comment-form/checking-comment-form.stories.ts#L34)):*

Good catch. Interesting that `canvas.getByRole()`returns `HTMLElement`, but `canvas.findByTitle()` returns `Promise<HTMLElement>`. Changed a bunch of these.
</blockquote></details>

Yeah. That's an unexpected difference.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts` line 17 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoPXmC5-68djHih7e-:-NSlxe498hc_1RjjTHHc:b-dgrdpg) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts#L17)):*

`undefined` is not a valid input. We just can't guarantee with type safety that it will be set. And it won't get set at the time the component is constructed.
</blockquote></details>

But isn't the reason for having the set of values to select from, so that we can see what a widget looks like with different sets of data that might be tricky to re-create in practice? When I define the list like `options: [1, 2, undefined],`, I can select undefined in the Storybook "Controls" tab and see what the widget looks like when the book is set to undefined. 

Tho I suppose, just because it's a possible thing that could be set, doesn't mean it needs to be one of the options in the Storybook selector.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts` line 20 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRoPB6A2H7fj-wfSFz0:-NSm-t3q991alPduCQI4:b-6cqxb3) ([raw file](https://github.com/sillsdev/web-xforge/blob/a46d12206deeb4f40f098a9181cca88f6232bca7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts#L20)):*
> It doesn't seem like that useful of a test to me. This component knows nothing about the canon, or number of chapters in each book.

But it does have behaviour based on the inputs. If the `chapters` input is `[1,2,3,4]`, and the `chapter` input is `5`, will it crash? Will it display 5? Will it display 0? Will the component crash when it thinks about whether the arrows should be enabled or disabled? Because the "chapter" input now lets me type in my own number, I can go to the Ruth story, enter 5, and see what it does and shows.

[I'm done, but I don't want to set my disposition to Satisfied in case it doesn't show you my last message. I forget how Reviewable handles that by default.]
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.stories.ts` line 32 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NS7-bNy3v5YyluHkFB-:-NSm49aY98huB4rvdC4b:bwm298q) ([raw file](https://github.com/sillsdev/web-xforge/blob/cd539f34ba2ad78b14aea115554a40b2ce39464a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.stories.ts#L32)):*

Yes. They're supposed to show up in the order they're defined. I think webpack might be messing that up. I'm still hoping we can fix this at some point.
</blockquote></details>

Ahh, okay.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 43 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRo_iImASF_qhFg6DHh:-NSlE2xw1QO8bMD3VtBg:bngr3bz) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L43)):*

My comment says "In some cases the locale has been known to be the empty string". What I meant was "in addition to checking for null/undefined, we also need to consider the empty string, hence the use of || instead of ??"

While I generally agree that we should make our code as specific as possible, I'm *intentionally* including more values in this check.

> Writing const locale = context.globals.locale || I18nService.defaultLocale.canonicalTag; is like writing
`const locale == context.globals.locale != false ? context.globals.locale : I18nService.defaultLocale.canonicalTag;`

That wouldn't handle null. I'm trying to handle null, undefined, and the empty string. Usually we only want to handle null and undefined, so nullish coalescing makes sense (`a ?? b`). But in this case I found I needed to include the empty string as well.

An alternative would be:
``` ts
const locale = context.globals.locale == null || context.globals.locale === '' ? I18nService.defaultLocale.canonicalTag : context.globals.locale;
```

But there's no value in being that specific. We're expecting a non-empty string, so it would make more sense to write:
``` ts
const locale = typeof context.globals.locale === 'string' && context.globals.locale !== '' ? context.globals.locale : I18nService.defaultLocale.canonicalTag;
```

This would be the actual defensive code. But that seemed more complicated than needed.
</blockquote></details>

Ahh. I did not understand until your further explanation that you were meaning to check for null/undefined. I only understood that you were checking for the empty string. It looks like (with the help of the code comment) we are expecting `context.globals.locale` to normally be a locale string, or sometimes to be `''`, and if it's `''` then we will fall back to `...canonicalTag`. 

I think your explanation re-enforces the idea that `||` is not the right operator for this situation :-). 
The reason is that not only is `||` being used to check for `''`, but also for null/undefined. 

I think that whether we think this code is OK probably depends on how we view the `||` operator, or perhaps our views on Boolean coercion. 
If we view the `||` operator as an operator that compares true and false values, we may conclude that `null || 'abc'` is unexpected and incorrect use of the operator, even if [valid JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators#logical_operators). (And that this would be a JavaScript feature to avoid.) 
If we view the `||` operator as an operator that compares truthy and  falsy values, we may conclude that `null || 'abc'` is correct use of the operator.

Probably we differ on our views about whether we think "truthy and falsy" is a feature of JavaScript to avoid or embrace.

Looking at the eslint [Fixes and Suggestions](https://typescript-eslint.io/rules/strict-boolean-expressions/#fixes-and-suggestions) section, it looks like some other ways to render the code could be:

```ts
const locale = Boolean(context.globals.locale) ? context.globals.locale : I18nService.defaultLocale.canonicalTag;
```

```ts
const locale = (context.globals.locale ?? '') !== '' ? context.globals.locale : I18nService.defaultLocale.canonicalTag;
```

(And `Boolean(foo)` does use Boolean coerceon, but it's explicit.)

Hey, I see that maybe what I want is an analog of strict equality for the [logical operators](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators#logical_operators). So `|||` and `&&&` could be like `||` and `&&` but without type conversion.

Anyway. I won't plan to push more for changing the introduction of this line.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 43 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1767#-NRofQp4EDvvR-v1WMmx:-NSlZ721-cdsk9Sh8G-9:bsbd8t3) ([raw file](https://github.com/sillsdev/web-xforge/blob/49de884163123b092f8ce809d8809e6911696038/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L43)):*
Okay, interesting. I might tend to think that `const locale =` doesn't have a type because the programmer thought it was obvious what the type was (or forgot to specify a type), rather than that they didn't know. We need a type called `idontknow` :). 
I tried setting it to `unknown`, but the next line complains, saying that 
> Type 'unknown' is not assignable to type 'string | undefined'.ts(2322)

It's interesting that `unknown` can't be assigned there, but `any` can.

PR #1769, opened 2023-03-28
Add files for Trusted Web Activity

laineyhm:
@megahirt is interested in helping develop an Android app for Scripture Forge. With the asset links file included in this PR, and the manifest file, we can use Bubblewrap CLI to generate a Trusted Web Activity that will display the web app inside an Android activity, with some UI changed to make it look like an app rather than a browsing window.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1769)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/.well-known/assetlinks.json
+++ b/src/SIL.XForge.Scripture/ClientApp/src/.well-known/assetlinks.json
@@ -1,0 +1,12 @@
+[
+  {
+    "relation": ["delegate_permission/common.handle_all_urls"],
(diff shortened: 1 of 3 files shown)


laineyhm:
Got the checks for formatting passing. This PR is ready for review whenever you have a chance to take a look.

Nateowami:
Thanks for your work on this @laineyhm. I have some concerns about supporting an Android app, which I've mentioned to Chris. It's possible we'll decide to create a Bubblewrap version of the app at some point in the future, but for now, I'm going to close this PR.

PR #1770, opened 2023-03-28
SF-1464 Fix wrapping in user list

Nateowami:
Previously the role and permissions checkbox label would not wrap, leaving all the wrapping to the name, which caused it to wrap a ton if it had to wrap.

Here's how it looks now (in French, where the issue was originally reported):

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1770)


marksvc (approved):
I also tried it out a bit locally.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1770#-:-NRiz1GN9bGH8jx4NbhK:bqwl8u4)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1772, opened 2023-03-28
SF-1932 - Store updated avatar URL in Auth0 user metadata

nigel-wells:
- Save any update to the Avatar URL (picture) in the user metadata in Auth0
- On receiving the user profile from Auth0, use the picture value from metadata if available
- Update presence information when the user doc changes
- Added return types
- Added a user doc to various tests that use the text component

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1772)

### DIFF
--- a/src/SIL.XForge/Services/UserService.cs
+++ b/src/SIL.XForge/Services/UserService.cs
@@ -69,9 +69,12 @@ public class UserService : IUserService
             );
             await userDoc.SubmitJson0OpAsync(op =>
             {
+                string picture =
+                    userProfile["user_metadata"] == null ? null : (string)userProfile["user_metadata"]["picture"];
+                string avatarUrl = string.IsNullOrWhiteSpace(picture) ? (string)userProfile["picture"] : picture;
                 op.Set(u => u.Name, name);
                 op.Set(u => u.Email, (string)userProfile["email"]);
-                op.Set(u => u.AvatarUrl, (string)userProfile["picture"]);
+                op.Set(u => u.AvatarUrl, avatarUrl);
                 op.Set(u => u.Role, (string)userProfile["app_metadata"]["xf_role"]);
                 if (ptIdentity != null)
                 {
--- a/src/SIL.XForge/Services/UserService.cs
+++ b/src/SIL.XForge/Services/UserService.cs
@@ -69,8 +69,7 @@ public class UserService : IUserService
             );
             await userDoc.SubmitJson0OpAsync(op =>
             {
-                string picture =
-                    userProfile["user_metadata"] == null ? null : (string)userProfile["user_metadata"]["picture"];
+                string picture = (string)userProfile["user_metadata"]?["picture"];
                 string avatarUrl = string.IsNullOrWhiteSpace(picture) ? (string)userProfile["picture"] : picture;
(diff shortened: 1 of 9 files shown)


src/SIL.XForge/Services/UserService.cs
```diff
@@ -69,9 +69,12 @@ await using (IConnection conn = await _realtimeService.ConnectAsync(curUserId))
             );
             await userDoc.SubmitJson0OpAsync(op =>
             {
+                string picture =
```
pmachapman, line 72: I found this block a little confusing on first glance (I had to double take to figure out what it was doing). One way you could make this more readable is to simplify this line as follows:
```csharp
string picture = (string)userProfile["user_metadata"]?["picture"];
```

pmachapman (changes requested):
Looks good. Thank you for cleaning up the TypeScript warnings and removing deadwood as you were updating. Just one comment on a bit I found confusing.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1772)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge/Services/UserService.cs` line 72 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1772#-NReTw0JdiGwA23GEj-D-r1-72:-NS3C3H2A1Ylf0Yy8YBw:b69u4km) ([raw file](https://github.com/sillsdev/web-xforge/blob/d64a6981f95dc4083ddb64a7d232c3bf1e9f2828/src/SIL.XForge/Services/UserService.cs#L72)):*

I found this block a little confusing on first glance (I had to double take to figure out what it was doing). One way you could make this more readable is to simplify this line as follows:
```csharp
string picture = (string)userProfile["user_metadata"]?["picture"];
```
</blockquote></details>

Ah neat, didn't realise we could do null checks like that. Does this same approach work for anything in C#? I assume it would work with more dimensions i.e. `(string)userProfile["user_metadata"]?["picture"]?["url"];` ?

pmachapman (approved):
> Reviewed 8 of 9 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1772)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge/Services/UserService.cs` line 72 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1772#-NReTw0JdiGwA23GEj-D-r1-72:-NS3CjJG3ISfisz42vgE:b-p4x4wl) ([raw file](https://github.com/sillsdev/web-xforge/blob/d64a6981f95dc4083ddb64a7d232c3bf1e9f2828/src/SIL.XForge/Services/UserService.cs#L72)):*

Ah neat, didn't realise we could do null checks like that. Does this same approach work for anything in C#? I assume it would work with more dimensions i.e. `(string)userProfile["user_metadata"]?["picture"]?["url"];` ?
</blockquote></details>

It works for anything that is nullable in C# (strings, objects, etc...). And yes, it works multiple levels deep.

PR #1774, opened 2023-03-29
SF-1917 Add reviewer label on note content

RaymondLuong3:
This change adds the user label to the notes left by non-paratext users. Initially I tried to do this on the back-end, but there were some round-tripping issues that were becoming more complex. Instead, I decided to add the label on the front-end, and make the label invisible in the note dialog.

SF User Label

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1774)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts
@@ -22,13 +22,16 @@ import { UserService } from 'xforge-common/user.service';
 import { objectId } from 'xforge-common/utils';
 import { FeatureFlagService } from 'xforge-common/feature-flags/feature-flag.service';
 import { isParatextRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-role';
+import { UserDoc } from 'xforge-common/models/user-doc';
 import { NoteThreadDoc, defaultNoteThreadIcon } from '../../../core/models/note-thread-doc';
 import { SFProjectDoc } from '../../../core/models/sf-project-doc';
 import { SFProjectProfileDoc } from '../../../core/models/sf-project-profile-doc';
 import { SFProjectService } from '../../../core/sf-project.service';
-import { SFProjectUserConfigDoc } from '../../../core/models/sf-project-user-config-doc';
 import { TextDoc, TextDocId } from '../../../core/models/text-doc';
 import { canInsertNote, formatFontSizeToRems } from '../../../shared/utils';
+import { environment } from '../../../../environments/environment';
+
+const COMMENTER_LABEL_REGEX = /<p>\[.+\s-\s.+\]<\/p>\s*<p>(.+)<\/p>/s;
 
 export interface NoteDialogData {
   threadId?: string;
@@ -53,7 +56,6 @@ export class NoteDialogComponent implements OnInit {
   private textDoc?: TextDoc;
   private paratextProjectUsers?: ParatextUserProfile[];
   private noteBeingEdited?: Note;
-  private projectUserConfigDoc?: SFProjectUserConfigDoc;
   private userRole?: string;
 
   constructor(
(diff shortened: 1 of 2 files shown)


src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts
```diff
@@ -21,13 +21,17 @@ import { I18nService } from 'xforge-common/i18n.service';
...
 import { NoteThreadDoc, defaultNoteThreadIcon } from '../../../core/models/note-thread-doc';
 import { SFProjectDoc } from '../../../core/models/sf-project-doc';
 import { SFProjectProfileDoc } from '../../../core/models/sf-project-profile-doc';
 import { SFProjectService } from '../../../core/sf-project.service';
-import { SFProjectUserConfigDoc } from '../../../core/models/sf-project-user-config-doc';
 import { TextDoc, TextDocId } from '../../../core/models/text-doc';
 import { canInsertNote, formatFontSizeToRems } from '../../../shared/utils';
+import { environment } from '../../../../environments/environment';
+
+const COMMENTER_LABEL_REGEX = /<p>\[.+\s-\s.+\]<\/p>\s*<p>(.+)<\/p>/s;
```
pmachapman, line 34: This regular expression is greedy. For example, it will match:
```
<p>[Peter] I wrote this</p>
<p>[Test] This is a comment - read</p>
<p>[Peter]</p>
<p>Something</p>
```
You could make it non-greedy by replacing the wildcard with a negation, and anchoring it to the start, for example:
```
/^<p>\[[^\]]+\s-\s[^\]]+\]<\/p>\s*<p>(.+)<\/p>/s
```

pmachapman (changes requested):
Thank you for this change - it matches the behavior for answers well. My only comment is around the regular expression being greedy.

Also, thanks for the clean up of the left over conflict code (viz. SF-1852).

As a note for future (possible) thought, I noticed during my review & testing that Paratext accepts any username for a note when it is created (it just creates a matching `Notes_User Name.xml` file). If we wanted to, we could create notes as the SF user, with (at least in my quick tests) no apparent issues in PT. If we were going to go down that path, we should update the logic that writes answers too.

RaymondLuong3 (commented):
Hmm, that is a good thought. One of the benefits of that would be that the user's name would integrate into the already existing UI in PT. I wonder what a Paratext project admin would think. It might confuse the team if they begin seeing notes authored by people who are not team members, but I don't know. We can come back to this after our initial release.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1774#-:-NS74iHdD2YwlItkHs-p:bnrp375)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts` line 34 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1774#-NRjAykmsQlJbibXxvrx-r1-34:-NS75Zr-35d4JIdeVTNZ:bvhyfie) ([raw file](https://github.com/sillsdev/web-xforge/blob/54a85675948ee03601aed85f4750981a413c80d1/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts#L34)):*

This regular expression is greedy. For example, it will match:
```
<p>[Peter] I wrote this</p>
<p>[Test] This is a comment - read</p>
<p>[Peter]</p>
<p>Something</p>
```
You could make it non-greedy by replacing the wildcards with negations, and anchoring to the start, for example:
```
/^<p>\[[^\n]+\s-\s[^\]\n]+\]<\/p>\s*<p>(.+)<\/p>/s
```
</blockquote></details>

Good thought! The greedy-ness of regex expressions have tripped me up before.

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1774#-:-NS7UvaU9FInob9lnhSS:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1775, opened 2023-03-31
SF-1935 - Close note bottom sheet when not in use

nigel-wells:
- Dismiss an open bottom sheet in the editor on disposal
- Change location for material overlays to be child elements of <app-root>
 - Update z-index for nav drawer and overlay to be higher than other material overlays

**Before**
image

**After**
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1775)


pmachapman (approved):
Looks good - thanks Nigel!

PR #1776, opened 2023-04-03
SF-1924 PT comments on SF community checking answers lost after sync

pmachapman:
**Overview**

Comments made in Paratext to Answers exported to Paratext from Scripture Forge were being lost on sync if a newer comment was made in Scripture Forge. This occurred because there was no accurate tracking of comments being deleted. To address that, this PR adds a deleted flag to Comments and Answers that is set when these are deleted, as opposed to the previous behavior which was the removal of the comment or answer from the question document.

**Changes in this Pull Request**

 - Rename `tagsAdded` XML tag to `tagAdded` to conform to Paratext Comments XML tag name
 - Fixed bug where the comment was incorrectly seen as modified when  no change had occurred
 - Add a `deleted` flag to answers and comments
 - Fixed a bug where comments created in Paratext are deleted when a newer comment is added in Scripture Forge
 - When an answer is deleted in SF that has Paratext comments, those comments will remain as their own notes in Paratext

**Known Issues**

 - Updating or deleting a comment or answer in Paratext will be reverted by Scripture Forge on sync
 - Resolving the note corresponding to the question in Paratext will be reverted by Scripture Forge on sync
 - Answers and comments that have been synced then deleted before this PR is merged will not be deleted from Paratext

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1776)

### DIFF
--- a/src/RealtimeServer/common/services/project-data-service.spec.ts
+++ b/src/RealtimeServer/common/services/project-data-service.spec.ts
@@ -253,23 +253,43 @@ describe('ProjectDataService', () => {
     const userConn = clientConnect(env.server, 'user');
     await expect(
       submitJson0Op<TestData>(userConn, TEST_DATA_COLLECTION, 'test02', ops =>
-        ops
-          .add(d => d.children, { id: 'sub04', ownerRef: 'user', children: [] })
-          .insert(d => d.children[1].children, 0, { id: 'sub05', ownerRef: 'user' })
+        ops.insert(d => d.children[0].children, 0, { id: 'sub05', ownerRef: 'user' })
       )
     ).resolves.not.toThrow();
 
     const adminConn = clientConnect(env.server, 'admin');
     await expect(
       submitJson0Op<TestData>(adminConn, TEST_DATA_COLLECTION, 'test02', ops =>
-        ops.set<number>(d => d.children[1].children[0].num!, 1)
+        ops.set<number>(d => d.children[0].children[0].num!, 1)
       )
     ).rejects.toThrow();
     await expect(
       submitJson0Op<TestData>(adminConn, TEST_DATA_COLLECTION, 'test02', ops =>
-        ops.set<boolean>(d => d.children[1].children[0].deleted!, true)
+        ops.set<boolean>(d => d.children[0].children[0].deleted!, true)
       )
     ).resolves.not.toThrow();
+  });
+
+  it('denies updates to the deleted property if delete operations are not permitted', async () => {
+    const env = new TestEnvironment();
+    await env.createData();
+
+    const userOwnConn = clientConnect(env.server, 'userOwn');
+    await expect(
+      submitJson0Op<TestData>(userOwnConn, TEST_DATA_COLLECTION, 'test03', ops =>
+        ops.insert(d => d.children[0].children, 0, { id: 'sub05', ownerRef: 'userOwn' })
+      )
+    ).resolves.not.toThrow();
+    await expect(
+      submitJson0Op<TestData>(userOwnConn, TEST_DATA_COLLECTION, 'test03', ops =>
+        ops.set<number>(d => d.children[0].children[0].num!, 1)
+      )
+    ).resolves.not.toThrow();
+    await expect(
+      submitJson0Op<TestData>(userOwnConn, TEST_DATA_COLLECTION, 'test03', ops =>
+        ops.set<boolean>(d => d.children[0].children[0].deleted!, true)
+      )
+    ).rejects.toThrow();
   });
 });
 
@@ -349,8 +369,7 @@ class TestDataService extends ProjectDataService<TestData> {
 
       [TestProjectDomain.SubSubData, Operation.ViewOwn],
       [TestProjectDomain.SubSubData, Operation.Create],
-      [TestProjectDomain.SubSubData, Operation.EditOwn],
-      [TestProjectDomain.SubSubData, Operation.DeleteOwn]
+      [TestProjectDomain.SubSubData, Operation.EditOwn]
     ],
     observer: [
       [TestProjectDomain.TestData, Operation.View],
--- a/src/RealtimeServer/common/services/project-data-service.spec.ts
+++ b/src/RealtimeServer/common/services/project-data-service.spec.ts
@@ -245,6 +245,32 @@ describe('ProjectDataService', () => {
       )
     ).resolves.not.toThrow();
   });
+
+  it('handles updates to the deleted property as delete operations', async () => {
+    const env = new TestEnvironment();
+    await env.createData();
+
+    const userConn = clientConnect(env.server, 'user');
+    await expect(
+      submitJson0Op<TestData>(userConn, TEST_DATA_COLLECTION, 'test02', ops =>
+        ops
+          .add(d => d.children, { id: 'sub04', ownerRef: 'user', children: [] })
+          .insert(d => d.children[1].children, 0, { id: 'sub05', ownerRef: 'user' })
+      )
+    ).resolves.not.toThrow();
+
+    const adminConn = clientConnect(env.server, 'admin');
+    await expect(
+      submitJson0Op<TestData>(adminConn, TEST_DATA_COLLECTION, 'test02', ops =>
+        ops.set<number>(d => d.children[1].children[0].num!, 1)
+      )
+    ).rejects.toThrow();
+    await expect(
+      submitJson0Op<TestData>(adminConn, TEST_DATA_COLLECTION, 'test02', ops =>
+        ops.set<boolean>(d => d.children[1].children[0].deleted!, true)
+      )
+    ).resolves.not.toThrow();
+  });
 });
 
 enum TestProjectDomain {
@@ -256,6 +282,7 @@ enum TestProjectDomain {
 interface TestSubSubData extends OwnedData {
   id: string;
   num?: number;
+  deleted?: boolean;
 }
 
 interface TestSubData extends OwnedData {
@@ -290,7 +317,7 @@ class TestDataService extends ProjectDataService<TestData> {
 
       [TestProjectDomain.SubSubData, Operation.View],
       [TestProjectDomain.SubSubData, Operation.Create],
-      [TestProjectDomain.SubSubData, Operation.Edit],
+      [TestProjectDomain.SubSubData, Operation.EditOwn],
       [TestProjectDomain.SubSubData, Operation.Delete]
     ],
     user: [
--- a/src/RealtimeServer/common/services/project-data-service.ts
+++ b/src/RealtimeServer/common/services/project-data-service.ts
@@ -149,9 +149,16 @@ export abstract class ProjectDataService<T extends ProjectData> extends JsonDocS
         // property update
         const entityPath = op.p.slice(0, domain.pathTemplate.template.length);
         const oldEntity = this.deepGet(entityPath, oldDoc);
+
+        // Changing the deleted property should be treated as a delete operation
+        let operation: Operation = Operation.Edit;
+        if (op.p[op.p.length - 1] === 'deleted') {
+          operation = Operation.Delete;
+        }
+
         // if the entity doesn't exist in the old doc, then it must be inserted by a previous op that the user has a
         // right to perform, so we don't need to check this edit right
-        if (oldEntity != null && !this.hasRight(project, domain, Operation.Edit, session.userId, oldEntity)) {
+        if (oldEntity != null && !this.hasRight(project, domain, operation, session.userId, oldEntity)) {
           return false;
         }
       } else {
--- a/src/RealtimeServer/common/services/project-data-service.ts
+++ b/src/RealtimeServer/common/services/project-data-service.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
 import ShareDB from 'sharedb';
 import { ConnectSession } from '../connect-session';
 import { MigrationConstructor } from '../migration';
--- a/src/RealtimeServer/scriptureforge/models/comment.ts
+++ b/src/RealtimeServer/scriptureforge/models/comment.ts
@@ -2,6 +2,7 @@ import { OwnedData } from '../../common/models/owned-data';
 
 export interface Comment extends OwnedData {
   dataId: string;
+  deleted: boolean;
   syncUserRef?: string;
   text?: string;
   dateModified: string;
--- a/src/RealtimeServer/scriptureforge/models/note.ts
+++ b/src/RealtimeServer/scriptureforge/models/note.ts
@@ -7,7 +7,6 @@ export interface Note extends Comment {
   threadId: string;
   type: string;
   conflictType: string;
-  deleted: boolean;
   status: NoteStatus;
   tagId?: number;
   reattached?: string;
--- a/src/RealtimeServer/scriptureforge/services/question-migrations.spec.ts
+++ b/src/RealtimeServer/scriptureforge/services/question-migrations.spec.ts
@@ -1,0 +1,59 @@
+import ShareDB from 'sharedb';
+import ShareDBMingo from 'sharedb-mingo-memory';
+import { instance, mock, when } from 'ts-mockito';
+import { MetadataDB } from '../../common/metadata-db';
+import { RealtimeServer } from '../../common/realtime-server';
+import { SchemaVersionRepository } from '../../common/schema-version-repository';
+import { createDoc, fetchDoc } from '../../common/utils/test-utils';
+import { QUESTIONS_COLLECTION } from '../models/question';
+import { QuestionService } from './question-service';
+
+describe('QuestionMigrations', () => {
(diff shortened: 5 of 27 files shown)


RaymondLuong3 (changes requested):
Nice work. This must have been something that was overlooked. Maybe we did not expect PT users to comment on community checking answers in Paratext. But it makes more sense to support it if we can, rather than make the assumption that PT users will not want to add their comments to community checking answers in PT. I checked this branch out and it appears to work as expected.

> Reviewed 22 of 22 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-:-NSHqTU3AKsNBMilKvgh:btlaccx)* status: all files reviewed, 9 unresolved discussions (waiting on @pmachapman)

___
*[`src/RealtimeServer/scriptureforge/services/question-migrations.ts` line 11 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSHc_NLA246twxWfeEv:-NSHc_NM0L1nnO-Ucp1P:bgonhcz) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e043b3cb36cda4a5ce15089899c5d25596be9c8/src/RealtimeServer/scriptureforge/services/question-migrations.ts#L11)):*
> ```XML
>     const ops: Op[] = [];
>     for (let i = 0; i < doc.data.answers.length; i++) {
>       if (doc.data.answers[i].deleted === undefined) {
> ```

We typically have been using `deleted == null` to cover both the case of deleted being undefined or null.



_Code quote:_
```XML
doc.data.answers[i].deleted === undefined) 
```
___
*[`src/RealtimeServer/scriptureforge/services/question-migrations.spec.ts` line 25 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSHdLbu1OppN7mV2KPM:-NSHdLbu1OppN7mV2KPN:b-3i2we5) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e043b3cb36cda4a5ce15089899c5d25596be9c8/src/RealtimeServer/scriptureforge/services/question-migrations.spec.ts#L25)):*
> ```XML
> 
>       let questionDoc = await fetchDoc(conn, QUESTIONS_COLLECTION, 'question01');
>       expect(questionDoc.data.answers[0].comments[0].deleted).toBe(false);
> ```

We should also check that both answers should have `deleted: false` set.



_Code quote:_
```XML
      expect(questionDoc.data.answers[0].comments[0].deleted).toBe(false);
      expect(questionDoc.data.answers[0].comments[1].deleted).toBe(true);
      expect(questionDoc.data.answers[1].comments[0].deleted).toBe(false);
      expect(questionDoc.data.answers[1].comments[1].deleted).toBe(false);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 896 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSHhvdL3_dTHKNHl-2q:-NSHhvdL3_dTHKNHl-2r:b-xz82q8) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e043b3cb36cda4a5ce15089899c5d25596be9c8/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L896)):*
> ```XML
>     const answerIndex = this.getAnswerIndex(answer);
>     if (answerIndex >= 0) {
>       const commentLength = activeQuestionDoc.data!.answers[answerIndex].comments.length;
> ```

Nit: Add a type here.



_Code quote:_
```XML
 const commentLength = activeQuestionDoc.data!.answers[answerIndex].comments.length;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts` line 228 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSHiIaT9fqX9g93YG5E:-NSHiIaT9fqX9g93YG5F:bwqrr9x) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e043b3cb36cda4a5ce15089899c5d25596be9c8/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.ts#L228)):*
> ```XML
>       return [];
>     }
>     return this._questionDoc.data.answers.filter(answer => !answer.deleted);
> ```

I see this is a fairly common change. What do you think about creating a helper function. Or better yet, you could include this helper function in the `QuestionDoc` class so you could simply call `questionDoc.activeAnswers`.



_Code quote (i):_
```XML
   return this._questionDoc.data.answers.filter(answer => !answer.deleted);
```



_Code snippet (ii):_
```XML
get activeAnswers(): Answer[] {
  return this.data?.answers.filter(answer => !answer.deleted) ?? [];
}
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.ts` line 147 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSHhBR3ChXwdE3O9JUg:-NSHhBR3ChXwdE3O9JUh:b-k1ggz4) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e043b3cb36cda4a5ce15089899c5d25596be9c8/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.ts#L147)):*
> ```XML
>           return;
>         }
>         const commentsLength = this.answer.comments.filter(comment => !comment.deleted).length;
> ```

Nit: add a type here.



_Code quote:_
```XML
 const commentsLength = this.answer.comments.filter(comment => !comment.deleted).length;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 257 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSHhpQt1y3mgQwslLdR:-NSHhpQt1y3mgQwslLdS:b-k1ggz4) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e043b3cb36cda4a5ce15089899c5d25596be9c8/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L257)):*
> ```XML
>               op.add(puc => puc.answerRefsRead, answer.dataId);
>             }
>             const comments = sortBy(
> ```

Nit: add a type here.



_Code quote:_
```XML
            const comments = sortBy(
              answer.comments.filter(c => !c.deleted),
              c => c.dateCreated
            );
```
___
*[`src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs` line 288 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSHn0QJ9veRW-tYO3bN:-NSHn0QJ9veRW-tYO3bO:bl3mfew) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e043b3cb36cda4a5ce15089899c5d25596be9c8/src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs#L288)):*
> ```Smalltalk
>         switch (comment.Deleted)
>         {
>             case true when oldCommentElems.TryGetValue(key, out XElement oldCommentElem):
> ```

I'm curious why you used a switch here instead of just an if/else if. Is there an advantage?



_Code quote:_
```Smalltalk
            case true when oldCommentElems.TryGetValue(key, out XElement oldCommentElem):
                commentsToDelete.Add(oldCommentElem);
                break;
```
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextNotesMapperTests.cs` line 682 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSHpBxR6A-HtMJMCnot:-NSHpBxR6A-HtMJMCnou:b-fy6wov) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e043b3cb36cda4a5ce15089899c5d25596be9c8/test/SIL.XForge.Scripture.Tests/Services/ParatextNotesMapperTests.cs#L682)):*
> ```Smalltalk
> 
>     [Test]
>     public async Task GetNotesChangelistAsync_DoesDeleteParatextNotes()
> ```

Is this supposed to say "DoesNotDeleteParatextNotes"



_Code quote:_
```Smalltalk
    [Test]
    public async Task GetNotesChangelistAsync_DoesDeleteParatextNotes()
```
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextNotesMapperTests.cs` line 754 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSHoVBH2vhChwBEvdIE:-NSHoVBH2vhChwBEvdIF:b-l0tftt) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e043b3cb36cda4a5ce15089899c5d25596be9c8/test/SIL.XForge.Scripture.Tests/Services/ParatextNotesMapperTests.cs#L754)):*
> ```Smalltalk
>                         <thread id=""ANSWER_answer06"">
>                             <selection verseRef=""MAT 1:1"" startPos=""0"" selectedText="""" />
>                             <comment user=""PT User 1"" extUser=""user04"" date=""2019-01-05T08:00:00.0000000+00:00"">
> ```

You should be able to remove this since comments in PT would not have an extUser set.



_Code quote:_
```Smalltalk
extUser=""user04""
```

RaymondLuong3 (changes requested):
> Reviewed 12 of 14 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1776)* status: 24 of 26 files reviewed, 4 unresolved discussions (waiting on @pmachapman)

___
*[`src/RealtimeServer/common/services/project-data-service.ts` line 153 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSO_mbhDCVqD8cSAcNs:-NSO_mbhDCVqD8cSAcNt:b-49wxt1) ([raw file](https://github.com/sillsdev/web-xforge/blob/e65337db2479090c5c1bb248a27e34272f1e80d0/src/RealtimeServer/common/services/project-data-service.ts#L153)):*
> ```XML
>         const oldEntity = this.deepGet(entityPath, oldDoc);
> 
>         // Changing the deleted property should be treated as a delete operation
> ```

Hmm, this is an interesting case. Theoretically the operation is more of an update, but to keep things consistent, we should be treating this as a delete operation. I think this is a fair way to do this.



_Code quote:_
```XML
        // Changing the deleted property should be treated as a delete operation
```
___
*[`src/RealtimeServer/common/services/project-data-service.spec.ts` line 249 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSObXkeA8D1h_rWUh03:-NSObXkf7pe9u71AKH6F:b-fngm3d) ([raw file](https://github.com/sillsdev/web-xforge/blob/e65337db2479090c5c1bb248a27e34272f1e80d0/src/RealtimeServer/common/services/project-data-service.spec.ts#L249)):*
> ```XML
>   });
> 
>   it('handles updates to the deleted property as delete operations', async () => {
> ```

This shows that it works to delete the data, but doesn't show that when delete is not allowed, that a delete operation will cause a permission denied error. Would it be hard to add that test case?



_Code quote:_
```XML
 it('handles updates to the deleted property as delete operations', async () => {
```
___
*[`src/RealtimeServer/common/services/project-data-service.spec.ts` line 258 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSObBu7Dasn8vJoSwNK:-NSObBu7Dasn8vJoSwNL:bu3du6c) ([raw file](https://github.com/sillsdev/web-xforge/blob/e65337db2479090c5c1bb248a27e34272f1e80d0/src/RealtimeServer/common/services/project-data-service.spec.ts#L258)):*
> ```XML
>         ops
>           .add(d => d.children, { id: 'sub04', ownerRef: 'user', children: [] })
>           .insert(d => d.children[1].children, 0, { id: 'sub05', ownerRef: 'user' })
> ```

What is the purpose of setting the properties on `d.children[1]` Would it not make more sense to set it on `d.children[0]`? Or maybe I'm not quite understanding the structure of the test data.



_Code quote:_
```XML
.insert(d => d.children[1].children
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts` line 461 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSOZtnh2kC6hpt75ktr:-NSOZtnh2kC6hpt75kts:b-167koc) ([raw file](https://github.com/sillsdev/web-xforge/blob/e65337db2479090c5c1bb248a27e34272f1e80d0/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L461)):*
> ```XML
>         change.sfVersionOfQuestion!.data!.text,
>       after: this.referenceForDisplay(change.question) + ' ' + change.question.text,
>       answerCount: change.sfVersionOfQuestion?.getAnswers?.().length || 0,
> ```

The `getAnswers()` call shouldn't need the null check.



_Code quote:_
```XML
change.sfVersionOfQuestion?.getAnswers?.().length 
```
___
*[`src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs` line 288 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSHn0QJ9veRW-tYO3bN:-NSOc0FF-475aKYGlxSB:bm5uk03) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e043b3cb36cda4a5ce15089899c5d25596be9c8/src/SIL.XForge.Scripture/Services/ParatextNotesMapper.cs#L288)):*

No real benefit - my IDE suggested it, and I thought it looked clearer than the `if` block I wrote:

```csharp
        if (comment.Deleted && oldCommentElems.TryGetValue(key, out XElement oldCommentElem))
        {
            commentsToDelete.Add(oldCommentElem);
        }
        else if (!comment.Deleted && IsCommentNewOrChanged(oldCommentElems, key, commentElem, setCheckingTag))
        {
            threadElem.Add(commentElem);
        }
```

I think using a switch is much more defendable if I had multiple blocks in the `if` statement, and all of them dependent on the value of one variable.

If you think the `if` block is clearer, I can happily revert it to that.
</blockquote></details>

Oh neat, ok. I'll keep that in mind that this is also a valid way to write it. If your editor suggested it then someone thinks it is a better way to write it.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1776)* status: 24 of 26 files reviewed, 3 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/RealtimeServer/common/services/project-data-service.spec.ts` line 249 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSObXkeA8D1h_rWUh03:-NSggF3f2BCB6PHYkkN5:b-7ok6fb) ([raw file](https://github.com/sillsdev/web-xforge/blob/e65337db2479090c5c1bb248a27e34272f1e80d0/src/RealtimeServer/common/services/project-data-service.spec.ts#L249)):*

This shows that it works to delete the data, but doesn't show that when delete is not allowed, that a delete operation will cause a permission denied error. Would it be hard to add that test case?
</blockquote></details>

Good idea - I have added an additional test case.
___
*[`src/RealtimeServer/common/services/project-data-service.spec.ts` line 258 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSObBu7Dasn8vJoSwNK:-NSgb3lnAfCDZSXQ53dP:b-g58m8s) ([raw file](https://github.com/sillsdev/web-xforge/blob/e65337db2479090c5c1bb248a27e34272f1e80d0/src/RealtimeServer/common/services/project-data-service.spec.ts#L258)):*

What is the purpose of setting the properties on `d.children[1]` Would it not make more sense to set it on `d.children[0]`? Or maybe I'm not quite understanding the structure of the test data.
</blockquote></details>

I copied a previous test, but you are right - it doesn't make much sense. I have changed it to use `d.children[0]`.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts` line 461 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-NSOZtnh2kC6hpt75ktr:-NSgjRm9-mWfCbEUQEv8:b-c6bknt) ([raw file](https://github.com/sillsdev/web-xforge/blob/e65337db2479090c5c1bb248a27e34272f1e80d0/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts#L461)):*

The `getAnswers()` call shouldn't need the null check.
</blockquote></details>

I did it to fix some tests that were breaking, but have fixed the data for those tests instead.

RaymondLuong3 (approved):
:lgtm: 

> Reviewed 1 of 14 files at r2, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1776#-:-NSl9PAMAdkhR5R2KGWZ:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #1785, opened 2023-04-12
SF-1950 Fixed machine-api 403 errors when no edit permission

pmachapman:
This PR fixes a bug when logged in as a user that does not have edit permissions (such as Consultant, Observer, Reviewer, and Community checker), where the console is will contain 403 error messages from the machine-api.

To fix that, `TrainingProgressComponent.setupTranslationEngine()` is updated to check whether the user has permission to edit texts, and so matching the logic in `TranslateOverviewComponent.setupTranslationEngine()` and `EditorComponent.setupTranslationEngine()`.


src/SIL.XForge.Scripture/ClientApp/src/app/translate/training-progress/training-progress.component.ts
```diff
@@ -87,6 +91,14 @@ export class TrainingProgressComponent extends DataLoadingComponent implements O
     this.trainingProgressClosed = true;
   }
 
+  private get canEditTexts(): boolean {
+    const project = this.projectDoc?.data;
+    return (
+      project != null &&
+      SF_PROJECT_RIGHTS.hasRight(project, this.userService.currentUserId, SFProjectDomain.Texts, Operation.Edit)
```
Nateowami, line 98: What if the user has edit permission, but not for this text? Permission to edit can be at the chapter level, and the rights service will return true if the user has a general right to edit. See the editor component.

Nateowami (commented):
Thanks for working on this. Just one concern.

(I haven't had a chance to look at the code yet).

src/SIL.XForge.Scripture/ClientApp/src/app/translate/training-progress/training-progress.component.ts
pmachapman: Good question. As this component only shows training status, not suggestions, it is appropriate to check for the general edit permission (the same as is checked on the Translation Overview screen). If someone was editing a chapter they had permission to view, navigated to the Translation Overview screen, then navigated to a chapter they could not edit, the Training Status should still display while training is in progress.

For clarity, I have renamed this method to `userHasGeneralEditRight`, matching `editor.component.ts`.

Nateowami: Ah, right, somehow I misunderstood the context we were dealing with.

src/SIL.XForge.Scripture/ClientApp/src/app/translate/training-progress/training-progress.component.ts
```diff
@@ -87,6 +91,17 @@ export class TrainingProgressComponent extends DataLoadingComponent implements O
     this.trainingProgressClosed = true;
   }
 
+  /**
+   * Determines whether the user has the right to edit texts generally, without considering permissions on this chapter.
```
Nateowami, line 95: I'm not sure this comment makes sense in the context of this component, which isn't concerned with any particular chapter.

Nateowami (approved):
I have one change to suggest, but I'm going to mark this as approved and ready to test.

src/SIL.XForge.Scripture/ClientApp/src/app/translate/training-progress/training-progress.component.ts
pmachapman: You are right - fixed.

Nateowami (approved):

PR #1788, opened 2023-04-18
SF-1951 Fix position of offline indicator

Nateowami:
Before | After
-------|------
 | 

I also changed a CSS rule from:
``` scss
::ng-deep .avatar-content { ... }
```
To:
``` scss
:host ::ng-deep .avatar-content { ... }
```
Using `::ng-deep` should generally be avoided because it breaks out of Angular's view encapsulation so the styles apply globally. In this case we want that in order to style a subcomponent. Adding the `:host` selects the current component and applies the rule only to itself and its children, getting rid of any global impact.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1788)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/avatar/avatar.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/avatar/avatar.component.html
@@ -7,6 +7,4 @@
   [borderColor]="borderColor"
   [bgColor]="borderColor"
 ></ngx-avatar>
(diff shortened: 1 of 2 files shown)


RaymondLuong3 (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1788#-:-NTPUYDE1aiFUw_EsuHq:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1792, opened 2023-04-19
SF-1958 Show the commenter selected verse on bottom sheet

RaymondLuong3:
* use the commenter selected verse when saving a note thread

The currentOrDefault verse for a text isn't the same as the commenter selected verse which will always be set when a user opens a note dialog or bottom sheet to add a new comment. This change abandons the use of currentOrDefault verse.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1792)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -2618,7 +2618,7 @@ describe('EditorComponent', () => {
     it('shows current selected verse on bottom sheet', fakeAsync(() => {
       const env = new TestEnvironment();
       env.setProjectUserConfig();
-      env.setReviewerUser();
+      env.setCommenterUser();
       env.updateParams({ projectId: 'project01', bookId: 'LUK' });
       env.wait();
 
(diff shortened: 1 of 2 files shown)


pmachapman (changes requested):
Paratext allows us to create notes at "verse 0" (in the heading/intro). This change doesn't break the ability to read notes at that reference, but I just thought I would flag it, in case we want to be able to create notes at that position in SF (its OK if we are not supposed to - this is just an FYI)

I noticed some unusual behavior when testing out this PR - when I selected a verse, then changed to a book where a verse had the cursor, and the verse text in the bottom sheet was stuck at the verse position for the previous book:



> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1792#-:-NTjqT2d44R_6oy4NDMU:bmxaxln)* status: 1 of 2 files reviewed, all discussions resolved

RaymondLuong3 (commented):
Yes, eventually we will think about the use case of adding notes to headings, but right now we don't really support that.
I tried to follow your steps for the strange behaviour that you see. Can you still reproduce it? When I navigate to a different book, the bottom sheet gets dismissed. Are you able to give more specific reproduction steps for this behaviour?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1792#-:-NTykFYNBLh5_n00_-m3:bc8k6mb)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (approved):
It longer occurs for me - I did properly clean out my project directory, so it must have been some other form of confusion or corruption. :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1792#-:-NUPIypx0yMc2rDNPO21:b-r8haj1)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1796, opened 2023-04-25
SF-1888 Change role names for SF reviewers and observers

RaymondLuong3:
This change updates the role names from reviewer to commenter, and observer to viewer. This change will clarify the role so that there is less confusion between checking reviewers and translation reviewers. The string in the backend will remain "sf_reviewer" and "sf_observer" since there is no real benefit to changing those strings which would require a migration.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1796)

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -12,6 +12,7 @@
     "caniuse",
     "cdnjs",
     "combobox",
+    "commenters",
     "compodoc",
     "dropdown",
     "ethnologue",
--- a/src/RealtimeServer/scriptureforge/models/sf-project-role.ts
+++ b/src/RealtimeServer/scriptureforge/models/sf-project-role.ts
@@ -15,8 +15,8 @@ export function isTranslateRole(role: string | undefined): boolean {
     return true;
   }
   switch (role) {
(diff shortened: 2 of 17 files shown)


Nateowami (commented):
> The string in the backend will remain "sf_reviewer" and "sf_observer" since there is no real benefit to changing those strings which would require a migration.

The sf_reviewer role is not in use on live, only QA. Since that's only used for testing, I think we can just remove the reviewers from their projects (no need to delete the users themselves) and rename the role. I think that's likely to lead to less confusion down the road.

> Reviewed 17 of 17 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1796#-:-NTuTy6oEGznJhlqxCn-:b-xbud68)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @RaymondLuong3)

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1796)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1798, opened 2023-04-26
SF-1973 Flag all chapters of extra-Biblical books as invalid

pmachapman:
Extra-biblical books (like Introduction, Indexes, Extra Material) should not be editable in Scripture Forge. However, if one of these books has more than one chapter, and those subsequent chapters (2, 3, etc) are valid, they will be marked as valid and editable in Scripture Forge.

This Pull Request fixes this, by ensuring that if a book is defined as invalid (by not being in the valid books list in `usx-sf.xsd`), all subsequent chapters will be invalid too.

In addition, this PR enables the test `ToDelta_LineBreakWithinVerse`, which was noticed have accidentally had its Test attribute omitted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1798)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/DeltaUsxMapper.cs
+++ b/src/SIL.XForge.Scripture/Services/DeltaUsxMapper.cs
@@ -144,6 +144,7 @@ public class DeltaUsxMapper : IDeltaUsxMapper
         var chapterDelta = new Delta();
         var nextIds = new Dictionary<string, int>();
         var state = new ParseState();
+        bool bookIsValid = true;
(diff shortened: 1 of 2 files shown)


Nateowami:
@pmachapman Not a big deal, but in the future I'd prefer for the final commit message to contain the issue number (if there is one), like this:
```
SF-1234 Fix some bug (#123)
```
The commit message for this PR ended up being:
```
Flag all chapters of extra-Biblical books as invalid (#1798)
```
Which makes it a bit more difficult for me to correlate commits to issues when checking that what we're going to ship is fully tested and ready to go.

PR #1801, opened 2023-04-26
Add Mongo query to count answers by month

Nateowami:
I've been using this query successfully for a good while now to count answers by the month they were added.

I wanted it to print out a total for every month regardless of whether any answers were added in that month, but unfortunately that requires a fairly verbose script at the end. It does make it easy to copy the output though and put it into a spreadsheet for plotting, which is what I've been doing with the results.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1801)

### DIFF
--- a/mongodb/Questions/CountAnswersByMonth.mongodb
+++ b/mongodb/Questions/CountAnswersByMonth.mongodb
@@ -1,0 +1,39 @@
+use('xforge');
+
+// Get the number of answers per month across all projects
+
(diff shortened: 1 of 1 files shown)


RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1801)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1805, opened 2023-04-26
Add MDC dialog stories

Nateowami:
This adds stories for all the MDC dialogs except the system admin delete user dialog, which is more complicated.

You can see the stories here: https://643eb1587a4d88f5ee921c56-yjkeljooff.chromatic.com/?path=/story/mdc-dialogs--question-answered

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1805)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/.storybook/main.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/.storybook/main.ts
@@ -6,7 +6,7 @@ module.exports = {
     options: {}
   },
   docs: {
-    autodocs: true
+    autodocs: 'tag'
   },
   staticDirs: [{ from: '../src/assets', to: '/assets' }]
 };
--- a/src/SIL.XForge.Scripture/ClientApp/.storybook/preview.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/.storybook/preview.ts
@@ -15,8 +15,7 @@ export const parameters = {
       color: /(background|color)$/i,
       date: /Date$/
     }
-  },
-  docs: { inlineStories: true }
+  }
 };
 
 export const globalTypes = {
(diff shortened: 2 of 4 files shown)


src/SIL.XForge.Scripture/ClientApp/.storybook/main.ts
```diff
@@ -6,7 +6,7 @@ module.exports = {
     options: {}
   },
   docs: {
-    autodocs: true
+    autodocs: 'tag'
```
Nateowami, line 9: The autogenerated docs page shows each story on the same page. This works OK unless your stories all open dialogs, in which case they all open at once on top of each other, and it just makes a mess. Hence turning it off, except when specifically turned on by specifying the autodocs tag.

src/SIL.XForge.Scripture/ClientApp/.storybook/preview.ts
```diff
@@ -15,8 +15,7 @@ export const parameters = {
       color: /(background|color)$/i,
       date: /Date$/
     }
-  },
-  docs: { inlineStories: true }
```
Nateowami, line 19: This line was added when migrating to Storybook 7 in order to keep old behavior. It's not needed as I'm turning off autodocs anyway.

RaymondLuong3 (approved):
I probably wouldn't be too much work to update those dialogs to material now. The import question dialog would probably be one of the harder ones, but the others would be easier. Since you've already spent the time to do this, let's get it in anyways.

> Reviewed 3 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1805#-:-NU8yaRO7Ns5IzB-Cm-2:bx5r16e)* status: all files reviewed, 2 unresolved discussions

Nateowami:
@RaymondLuong3 Actually, the conversion to Material is the precise reason I made these stories, so it will be easy to see a before and after and test how they look in each localization and in each scenario.

PR #1816, opened 2023-05-03
Fixed Kayah Li script on Windows

pmachapman:
This Pull Request fixes Kayah Li script on Windows 10 and 11 by including the "Noto Sans Kayah Li" font.

Support for other scripts is not included, as testing a broad range of DBL resources in non-latin scripts all displayed correctly on Windows 10 and 11.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1816)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
@@ -9,6 +9,7 @@ quill-editor {
   > .ql-container {
     @include typography.typography(body1);
     > .ql-editor {
+      font-family: Roboto, 'Noto Sans Kayah Li', sans-serif;
       line-height: 1.6;
       width: 100%;
       word-break: break-word;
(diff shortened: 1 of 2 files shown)


Nateowami:
Interesting. Do you know why this particular script doesn't fall back on Windows?

pmachapman:
> Interesting. Do you know why this particular script doesn't fall back on Windows?

Windows 10 it is straightforward - the character set is not in the fallback fonts, so unless we support Kayah Li characters, they will not display.

For Windows 11, it is complicated:

 * On localhost, bold didn't work for Kayah Li with the default fallback character set (Sans Serif Collection). An odd bug!
 * On QA, Kayah Li characters would not display at all. I initially blamed Angular's optimization, but I am not convinced it is at fault - it is either a bug with Chromium or Windows.

Interestingly, a similar bug with the Sans Serif Collection font was recently fixed in Windows 11 (see https://blogs.windows.com/windows-insider/2023/02/01/announcing-windows-11-insider-preview-build-25290/), so it would not surprise me if the bug lay with Windows.

Nateowami (approved):

Nateowami:
@pmachapman I've tested this locally and seen it load the fonts correctly. I think I consider that tested well enough. Shall we merge?

PR #1824, opened 2023-05-04
tool: enhance time output from text-doc-history

marksvc:
commit-id:30dc7691

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1824)

### DIFF
--- a/scripts/db_tools/text-doc-history.js
+++ b/scripts/db_tools/text-doc-history.js
@@ -68,8 +68,8 @@ async function run() {
 function logEdit(snapshot, user, editCount, startTime, endTime) {
   const time =
     editCount === 1
-      ? `at ${new Date(startTime).toUTCString()}`
-      : `from ${new Date(startTime).toUTCString()} to ${new Date(endTime).toUTCString()}`;
+      ? `at ${new Date(startTime).toISOString()} (${startTime})`
+      : `from ${new Date(startTime).toISOString()} (${startTime}) to ${new Date(endTime).toISOString()} (${endTime})`;
 
   console.log(`Modified by ${user} in ${editCount} edits ${time}`);
   if (snapshot.data == null) {


scripts/db_tools/text-doc-history.js
```diff
@@ -68,8 +68,8 @@ async function run() {
 function logEdit(snapshot, user, editCount, startTime, endTime) {
   const time =
     editCount === 1
-      ? `at ${new Date(startTime).toUTCString()}`
-      : `from ${new Date(startTime).toUTCString()} to ${new Date(endTime).toUTCString()}`;
+      ? `at ${new Date(startTime).toISOString()} (${startTime})`
+      : `from ${new Date(startTime).toISOString()} (${startTime}) to ${new Date(endTime).toISOString()} (${endTime})`;
```
Nateowami, line 72: Do I understand correctly that we're outputting the Unix timestamp in addition to the ISO string? I'm sure you have a reason, but I'm not seeing it.

marksvc:
*[`scripts/db_tools/text-doc-history.js` line 72 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1824#gh-1185686630:-NWnE77M8b1ZgUVawc--:b5iqgry) ([raw file](https://github.com/sillsdev/web-xforge/blob/b3e982d766fa15cc7c7a9120c4730b37de6a3e03/scripts/db_tools/text-doc-history.js#L72)):*

Do I understand correctly that we're outputting the Unix timestamp in addition to the ISO string? I'm sure you have a reason, but I'm not seeing it.
</blockquote></details>

Yes! The reason is because I need to work with the Unix timestamp sometimes when investigating ops. I don't remember exactly what I was doing at the time, but for a contrived example, maybe I want to query o_texts documents that have an m.ts greater than a certain value, and so I could compare it to a specific Unix timestamp returned by text-doc-history. (And there may be a fancy date-conversion way to do that one, but other tools or APIs in the area may not be as clever). 
But I don't _just_ want the Unix timestamp, so I have it shown together with the other date format.

And I changed from `toUTCString` (Wed, 31 May 2023 19:32:16 GMT) to `toISOString` (2023-05-31T19:32:16.485Z), because when I need to pay attention to the date, and compare it with others, it helps to have the date written out in ISO 8601. For me, at least. There are clearly different preferences here, so I don't need to get my way on using this format here :)

Nateowami (approved):

PR #1829, opened 2023-05-05
SF-2000 Update audio recorder component

Nateowami:
- Convert to Material
- Make "Stop Recording" button color primary. Some time ago while watching a user use SF, I noticed that the browser's permission prompt for recording audio had taken his focus away from the dialog, so by the time he'd figured out he needed to give permission and then recorded the audio, he was kind of lost and had trouble finding the "stop recording" button. Making the button color primary isn't going to totally fix that, but when a user is recording, stopping recording is always the expected next action, so it makes sense to make it as prominent as possible.
- I changed the file name of the recorded audio. Previously it was using the users's display name. I can't think of any benefit of doing this (it's not guaranteed to be unique, and it's not shown to users), and I can think of several downsides (there might be characters that cause problems, and there's a slight security concern of having the user's name in the file name). The back end does not use the file name, except to look at the extension and see whether it's specified as MP3, and if not, does a conversion.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1829)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-recorder/checking-audio-recorder.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-recorder/checking-audio-recorder.component.html
@@ -1,27 +1,22 @@
 <ng-container *transloco="let t; read: 'checking_audio_recorder'">
   <ng-container *ngIf="!hasAudioAttachment">
-    <button *ngIf="!isRecording" mdc-button [unelevated]="true" type="button" (click)="startRecording()" class="record">
-      <mdc-icon>mic</mdc-icon>
+    <button *ngIf="!isRecording" mat-flat-button color="primary" (click)="startRecording()" class="record">
+      <mat-icon>mic</mat-icon>
       {{ t("record") }}
     </button>
-    <button *ngIf="isRecording" mdc-button type="button" (click)="stopRecording()" class="stop-recording">
-      <mdc-icon>stop</mdc-icon>
+    <button *ngIf="isRecording" mat-flat-button color="primary" (click)="stopRecording()" class="stop-recording">
+      <mat-icon>stop</mat-icon>
       {{ t("stop_recording") }}
     </button>
   </ng-container>
   <div *ngIf="hasAudioAttachment" class="has-attachment">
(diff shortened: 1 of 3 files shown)


RaymondLuong3 (approved):
It's funny how many different ways users can get confused in our app.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1829#-:-NUhvfa232ibvKS0E_Iv:bb7fc19)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-recorder/checking-audio-recorder.component.ts` line 67 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1829#-NUhvCb67ooT2sGnklHg:-NUhvCb67ooT2sGnklHh:b-oy1trm) ([raw file](https://github.com/sillsdev/web-xforge/blob/6c834475e74f7a39358039e3118aa8e9c1b1d65d/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-recorder/checking-audio-recorder.component.ts#L67)):*
> ```XML
>       status: 'processed',
>       blob: this.recordRTC.getBlob(),
>       fileName: objectId() + '.webm'
> ```

I like this too, I'm not sure why we thought it would be good to include the user's display name in the file name.

Nateowami (commented):
@RaymondLuong3 I've updated this to remove the UserService from the spec for this component. I first noticed that it was using the display name when writing a story for this component and setting up the mocks and was surprised it needed the UserService.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1829#-:-NUhymzN0RT6X82qrMRl:b-c7prft)* status: 2 of 3 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (approved):
Good thinking. Sorry I missed the user service changes in the spec

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1829#-:-NUi1O3m-hJEUz4ydCAi:bwmhx0i)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1830, opened 2023-05-05
SF-1997 Delete paratext comments only when the corresponding note is marked deleted

RaymondLuong3:
Instead of delete paratext notes when a corresponding note in a note thread does not exist, I have updated the logic to only delete paratext notes if the corresponding note in a note thread is marked deleted. This will hopefully significantly reduce the chance of losing Paratext notes.

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts
@@ -101,7 +101,7 @@ export class NoteThreadDoc extends ProjectDataDoc<NoteThread> {
       return '';
     }
 
-    const iconDefinedNotes = this.notesInOrderClone(this.data.notes).filter(n => n.tagId != null);
+    const iconDefinedNotes = this.notesInOrderClone(this.data.notes).filter(n => n.tagId != null && !n.deleted);
     let tagId: number | undefined =
       iconDefinedNotes.length === 0 ? undefined : iconDefinedNotes[iconDefinedNotes.length - 1].tagId;
     if (tagId == null) {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -2975,7 +2975,7 @@ describe('EditorComponent', () => {
       const segmentRef = 'verse_1_3';
       let thread2Elem: HTMLElement | null = env.getNoteThreadIconElement(segmentRef, threadId);
       expect(thread2Elem).not.toBeNull();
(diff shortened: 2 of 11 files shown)


src/SIL.XForge.Scripture/ClientApp/src/app/core/models/note-thread-doc.ts
```diff
@@ -101,7 +101,7 @@ export class NoteThreadDoc extends ProjectDataDoc<NoteThread> {
       return '';
     }
 
-    const iconDefinedNotes = this.notesInOrderClone(this.data.notes).filter(n => n.tagId != null);
+    const iconDefinedNotes = this.notesInOrderClone(this.data.notes).filter(n => n.tagId != null && !n.deleted);
```
Nateowami, line 104: Rather than filtering on the front end, we should be filtering on the back end. We do this when fetching questions, where we query for only those that are not archived. I think we could do the same thing here.

RaymondLuong3: It is possible that this can be done, but it wouldn't be as simple. The questions collection can be filtered for not archived questions because that is a property on the Question model, but there isn't a defined way to query for note threads and then then filter for a specific subset of notes on a thread. So the simplest way to do it is just to filter it on the front end.

pmachapman (changes requested):
I had trouble having the delete operation occur for the Paratext note when testing. My process was:

1. Create a note in Paratext and sync
2. Delete the note in SF and sync
3. Sync in Paratext, and no changes were incoming, and the note was not deleted

Also, creating a brand new note in SF (after enabling the feature flag) did not result in it replicating to Paratext.

Do these issues occur for you, or has something gone wrong on my end?

> Reviewed 7 of 11 files at r1, 5 of 5 files at r2, all commit messages.
> Dismissed @github-code-scanning[bot] from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1830#-:-NW-1aDx5exbOrzp--1_:b-o0b91r)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

RaymondLuong3 (commented):
Yes, you are right. I made a test branch to re-enabled pushing notes to PT, but then I deleted that branch without updating this one. It should be working now.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1830#-:-NW9h0-0920ppaoHIDpT:b-yi8xkv)* status: 5 of 11 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @pmachapman)

pmachapman (approved):
:lgtm: The PR looks good, and worked perfectly for me.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1830#-:-NW9uBGTD8a7t9z5yVDC:b-7l450x)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

PR #1831, opened 2023-05-08
SF-1990 Prevent scrolling to the selection when keyboard opens

RaymondLuong3:
* also fixes SF-1868

On mobile viewports, opening the keyboard causes the editor to scroll to the selected verse. This works nicely until a user clicks to add a comment. For Admins and translators with edit permissions, opening the bottom sheet to create a note causes the editor to loss focus, which causes the keyboard to disappear. Then the bottom sheet opens, and the keyboard reappears with the scroll jumping to the top of the chapter. Users without edit permissions see their keyboards open and close abruptly.

This change disables scrolling to the selection when the keyboard opens for the bottom sheet so the editor stays at the same scroll position. This does not affect when a user first selects a verse and the editor scrolls up for the keyboard to appear, which is most likely a feature on mobile devices which keeps the cursor in view. Note that the bottom sheet to add a comment may cover over the verse that is selected. In other apps, this appears to be the same behaviour; opening the bottom sheet should not scroll the editor.

To test this locally on a physical device, follow the instructions in the wiki under testing.

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
@@ -492,7 +492,8 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
 
   ngAfterViewInit(): void {
     this.subscribe(fromEvent(window, 'resize'), () => {
-      this.setTextHeight();
+      // only scroll to selection when window is resized on larger devices
+      this.setTextHeight(this.mediaObserver.isActive('gt-md'));
       // Note: this does not appear to get triggered when the window changes by opening dev tools
       this.resetInsertNoteFab();
     });
@@ -576,7 +577,7 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
             if (this.translationEngine == null || !this.translationSuggestionsProjectEnabled || !this.hasEditRight) {
               this.setupTranslationEngine();
             }
-            setTimeout(() => this.setTextHeight());
+            setTimeout(() => this.setTextHeight(true));
           });
 
           if (this.metricsSession != null) {
@@ -597,7 +598,7 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
       }
     );
 
-    setTimeout(() => this.setTextHeight());
+    setTimeout(() => this.setTextHeight(true));
   }
 
   ngOnDestroy(): void {
@@ -1115,7 +1116,7 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
       .subscribe();
   }
 
-  private setTextHeight(): void {
+  private setTextHeight(scrollToSelection: boolean): void {
     if (this.target == null || this.targetContainer == null) {
       return;
     }
@@ -1125,12 +1126,14 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
     const bounds = elem.getBoundingClientRect();
     // add bottom padding
     const top = bounds.top + (this.mediaObserver.isActive('xs') ? 0 : 14);
-    if (this.target.editor != null && this.targetFocused) {
+    // give the option to disable scrolling to the selection which causes issues
+    // when the keyboard opens on smaller devices and the scrolling is not helpful
+    if (this.target.editor != null && this.targetFocused && scrollToSelection) {
       // reset scroll position
       this.target.editor.scrollingContainer.scrollTop = 0;
     }
     this.textHeight = `calc(100vh - ${top}px)`;
-    if (this.targetFocused && this.dialogService.openDialogCount < 1) {
+    if (this.targetFocused && scrollToSelection && this.dialogService.openDialogCount < 1) {
       setTimeout(() => {
         // reset focus, which causes Quill to scroll to the selection
         this.target!.focus();
@@ -1188,7 +1191,7 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
         this.router.navigateByUrl('/projects/' + this.projectDoc!.id + '/translate', { replaceUrl: true });
       });
     });
-    setTimeout(() => this.setTextHeight());
+    setTimeout(() => this.setTextHeight(true));
   }
 
   private onStartTranslating(): void {


nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1831)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 496 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NV6jh6x1nGwEiGZwVb-:-NV6jh6x1nGwEiGZwVb0:b-ppuuz3) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab5952d0f5e2d08c1f507ece4f41eef5d751c529/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L496)):*
> ```XML
>       // only scroll to selection when window is resized on larger devices
>       this.setTextHeight(this.mediaObserver.isActive('gt-md'));
> ```

I get what you're wanting to do here but the check in use is based on the viewport width rather than the height. I guess, for a phone or tablet, the width is not likely to suddenly change so you could assume on smaller viewports that it is likely a phone/tablet therefor if the `resize` event is triggered then it is probably the keyboard. What happens on devices that are rotating from portrait to landscape? This might not be too much of a concern for a phone but it could easily take place on a tablet. Also, does this event trigger on scroll as some browsers will hide the header/footer controls to show more of the web page - does this trigger the resize event as well? If it did, and we didn't change the height, there will likely be a gap somewhere near the bottom as a result.

Detecting keyboards are hard from all the research I've done. There are APIs coming but not yet widely supported across all browsers. They also won't go back across all devices we currently support either. It's a shame as it feels like this should be a straightforward problem to solve but iOS have done one thing and Android did something else...

I'm not sure of the correct solution. Perhaps getting "close enough" is ok for now so your proposal might be ok? If you can test out the scenarios above - mainly for tablet users - then we can decide which path to take for now.

nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1831)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 498 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NV6umBiEEUJvYj3Omi7:-NV6umBiEEUJvYj3Omi8:b4d242z) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab5952d0f5e2d08c1f507ece4f41eef5d751c529/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L498)):*
> ```XML
>       this.setTextHeight(this.mediaObserver.isActive('gt-md'));
>       // Note: this does not appear to get triggered when the window changes by opening dev tools
>       this.resetInsertNoteFab();
> ```

I've just found that this line is also causing an issue related to what you're trying to address. As the `resize` event is triggered by the keyboard it then resets the FAB when the keyboard first shows requiring the user to click a segment a second time. Whatever we decide above to best detect the keyboard should apply here as well. Perhaps we need to have a helper method or service so that, whatever we decide, can be used throughout the app.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1831)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 496 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NV6jh6x1nGwEiGZwVb-:-NVGYJ0YCHz5Cxaaa6Uf:b-jgyain) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab5952d0f5e2d08c1f507ece4f41eef5d751c529/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L496)):*

I get what you're wanting to do here but the check in use is based on the viewport width rather than the height. I guess, for a phone or tablet, the width is not likely to suddenly change so you could assume on smaller viewports that it is likely a phone/tablet therefor if the `resize` event is triggered then it is probably the keyboard. What happens on devices that are rotating from portrait to landscape? This might not be too much of a concern for a phone but it could easily take place on a tablet. Also, does this event trigger on scroll as some browsers will hide the header/footer controls to show more of the web page - does this trigger the resize event as well? If it did, and we didn't change the height, there will likely be a gap somewhere near the bottom as a result.

Detecting keyboards are hard from all the research I've done. There are APIs coming but not yet widely supported across all browsers. They also won't go back across all devices we currently support either. It's a shame as it feels like this should be a straightforward problem to solve but iOS have done one thing and Android did something else...

I'm not sure of the correct solution. Perhaps getting "close enough" is ok for now so your proposal might be ok? If you can test out the scenarios above - mainly for tablet users - then we can decide which path to take for now.
</blockquote></details>

I double checked to see if rotating my phone causes the text height to not be adequate, and from my observation there isn't any blank space. The logic to calculate the text height hasn't changed since this only affects whether we scroll to the selection after the size of the window has changed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 498 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NV6umBiEEUJvYj3Omi7:-NVG_XTR3I-vnw6AjAO5:b-plteaj) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab5952d0f5e2d08c1f507ece4f41eef5d751c529/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L498)):*

I've just found that this line is also causing an issue related to what you're trying to address. As the `resize` event is triggered by the keyboard it then resets the FAB when the keyboard first shows requiring the user to click a segment a second time. Whatever we decide above to best detect the keyboard should apply here as well. Perhaps we need to have a helper method or service so that, whatever we decide, can be used throughout the app.
</blockquote></details>

Yes, that makes sense that the resizing of the screen when the keyboard appears could cause the selection to get reset. I've removed that logic now whenever the window gets resized. The strategy we will use is to ignore scrolling to selection when resizing the window on smaller devices and to keep selected verses selected when the window is resized. The text component and suggestions component also listen to the resize event, it may be helpful to have an app wide keyboard detection service. Something to keep in mind.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1442 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NVGZkk4CCJY6iaSGNb4:-NVGZkk4CCJY6iaSGNb5:b-ci7pfi) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab5952d0f5e2d08c1f507ece4f41eef5d751c529/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1442)):*
> ```XML
>       return;
>     }
>     this.resetCommenterVerseSelection();
> ```

I decided that it would be best to remove the logic that resets the commenter selection anytime we reset the insert note FAB. So when a user resizes their window, or when the keyboard appears, the selection will be unchanged. This looks mostly good in my testing, except if a user is rotating their screen. In that situation, the FAB and selected verse potentially goes off-screen. But I think that may not be a common scenario.

nigel-wells (changes requested):
Seems to be working well when I test on my phone. I've added some comments more around a discussion which we could probably do at tomorrows planning meeting.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-:-NVRXCs825CcVnq-xwoi:b48egwy)* status: all files reviewed, 5 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 496 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NV6jh6x1nGwEiGZwVb-:-NVRX1v5EXyLMBKQwUI6:b-xtyn46) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab5952d0f5e2d08c1f507ece4f41eef5d751c529/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L496)):*

I double checked to see if rotating my phone causes the text height to not be adequate, and from my observation there isn't any blank space. The logic to calculate the text height hasn't changed since this only affects whether we scroll to the selection after the size of the window has changed.
</blockquote></details>

Ah yes, good point.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1442 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NVGZkk4CCJY6iaSGNb4:-NVRQ8D92uI_XAKnilvy:b-g1x2w5) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab5952d0f5e2d08c1f507ece4f41eef5d751c529/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1442)):*

I decided that it would be best to remove the logic that resets the commenter selection anytime we reset the insert note FAB. So when a user resizes their window, or when the keyboard appears, the selection will be unchanged. This looks mostly good in my testing, except if a user is rotating their screen. In that situation, the FAB and selected verse potentially goes off-screen. But I think that may not be a common scenario.
</blockquote></details>

I'm ok with that, changing orientations would typically move you out of place in most apps.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1120 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NVRWsBcDsBlehEwjJOn:-NVRWsBcDsBlehEwjJOo:b-fvmhpz) ([raw file](https://github.com/sillsdev/web-xforge/blob/fcac6b860466329cdf301e936c315588cac19191/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1120)):*
> ```XML
>   }
> 
>   private setTextHeight(scrollToSelection: boolean): void {
> ```

Could have this default to `true` so that you don't have to pass a value in everywhere. May not be relevant based on other comments.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1134 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NVRWPdFDK9MVuytEIO0:-NVRWPdFDK9MVuytEIO1:bms3a87) ([raw file](https://github.com/sillsdev/web-xforge/blob/fcac6b860466329cdf301e936c315588cac19191/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1134)):*
> ```XML
>     if (this.target.editor != null && this.targetFocused && scrollToSelection) {
>       // reset scroll position
>       this.target.editor.scrollingContainer.scrollTop = 0;
> ```

Do you know why we did this in the first place? Do we actually need it? Perhaps we could discuss at our planning meeting unless you know why?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1140 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NVRWXfGBxra8ijjtdd9:-NVRWXfGBxra8ijjtddA:bne8zwv) ([raw file](https://github.com/sillsdev/web-xforge/blob/fcac6b860466329cdf301e936c315588cac19191/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1140)):*
> ```XML
>       setTimeout(() => {
>         // reset focus, which causes Quill to scroll to the selection
>         this.target!.focus();
> ```

Same with this one? I know Nathaniel has an open PR that would actually remove the entire `setTextHeight` method as it seems letting the CSS `grid` doesn't cause the same performance issues that we saw initially which is what prompted us to need to do this in the first place.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1831)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1140 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NVRWXfGBxra8ijjtdd9:-NVVvM728LBaiBSvs2xf:b941alv) ([raw file](https://github.com/sillsdev/web-xforge/blob/fcac6b860466329cdf301e936c315588cac19191/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1140)):*

Which PR is that? I would like to take a look.
</blockquote></details>

@RaymondLuong3 It's not a PR; it's the branch `fix/editor-height`. I did it quickly late one night, so it's far from polished or complete.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1831)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1134 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NVRWPdFDK9MVuytEIO0:-NVev51o-KxE20K8eMvc:b-67o1rw) ([raw file](https://github.com/sillsdev/web-xforge/blob/fcac6b860466329cdf301e936c315588cac19191/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1134)):*

Yea, we can bring this up in planning meeting.
</blockquote></details>

I can see what it is doing now. If resizing the window causes the selection to be off screen, this will reset the scroll which informs quill to scroll to the selection (I assume). When I comment this out, that behaviour is lost. That behaviour is useful on desktop, but it is not helpful on mobile.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1140 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NVRWXfGBxra8ijjtdd9:-NVevb7p6HG3YSY7U7JT:b-xpby6u) ([raw file](https://github.com/sillsdev/web-xforge/blob/fcac6b860466329cdf301e936c315588cac19191/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1140)):*

@RaymondLuong3 It's not a PR; it's the branch `fix/editor-height`. I did it quickly late one night, so it's far from polished or complete.
</blockquote></details>

Ok, when that is finished, we can adapt that behaviour, but for the time being I think skipping this on smaller devices is our best option to prevent jumpy scrolling when the keyboard opens.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1831)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1134 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NVRWPdFDK9MVuytEIO0:-NVfUxng85VUjK8QQSN3:bpoi0fr) ([raw file](https://github.com/sillsdev/web-xforge/blob/fcac6b860466329cdf301e936c315588cac19191/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1134)):*

I can see what it is doing now. If resizing the window causes the selection to be off screen, this will reset the scroll which informs quill to scroll to the selection (I assume). When I comment this out, that behaviour is lost. That behaviour is useful on desktop, but it is not helpful on mobile.
</blockquote></details>

How often do users resize the window? I would think not that often. Sure, it would be nice if it auto-scrolled to the correct location after a resize, but that doesn't feel critical. If it's causing a problem, and that's really the only benefit, and there isn't an obvious simple solution, just get rid of it.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1831)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @nigel-wells and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1134 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NVRWPdFDK9MVuytEIO0:-NVfWSxe3lTTGy9m4oSl:bnqmwmz) ([raw file](https://github.com/sillsdev/web-xforge/blob/fcac6b860466329cdf301e936c315588cac19191/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1134)):*

How often do users resize the window? I would think not that often. Sure, it would be nice if it auto-scrolled to the correct location after a resize, but that doesn't feel critical. If it's causing a problem, and that's really the only benefit, and there isn't an obvious simple solution, just get rid of it.
</blockquote></details>

Also, can't we easily scroll to an element without using Quill to set the focus? There are certainly times when we do want to scroll the selection into view, and it seems like we could solve it for all those situations without focusing the element.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1831)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1134 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NVRWPdFDK9MVuytEIO0:-NVkOITw-RD4JQyf0ciu:bnssycl) ([raw file](https://github.com/sillsdev/web-xforge/blob/fcac6b860466329cdf301e936c315588cac19191/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1134)):*

Also, can't we easily scroll to an element without using Quill to set the focus? There are certainly times when we do want to scroll the selection into view, and it seems like we could solve it for all those situations without focusing the element.
</blockquote></details>

I've removed the changes that scroll to the selection. The behaviour still works, quill will automatically scroll to the selection when navigating to the text because target editor is focused when the text changes. This should only affect scroll to the selection when the window gets resized, which does not happen anymore.

nigel-wells (approved):
:lgtm:

Happy to see all those lines of code be removed :) The testers can have a good play around - I'd be interested to see if iPhone has any issues as I've not tested on that device.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-:-NVlPOxr2jcImFuCf1pP:b-wx2l44)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1134 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1831#-NVRWPdFDK9MVuytEIO0:-NVlP96u9EsZkboP3HPW:b63zpr) ([raw file](https://github.com/sillsdev/web-xforge/blob/fcac6b860466329cdf301e936c315588cac19191/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1134)):*

I've removed the changes that scroll to the selection. The behaviour still works, quill will automatically scroll to the selection when navigating to the text because target editor is focused when the text changes. This should only affect scroll to the selection when the window gets resized, which does not happen anymore.
</blockquote></details>

I've done some testing and I think it works well even on mobile. The adjustment to height that Nathanial has looked at will also help a few other things around the height of the editor on mobile. Happy with this.

PR #1817, opened 2023-05-04
Fix height of send button in share control

Nateowami:
A | B
--|--
Before | 
After | 

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1817)

PR #1832, opened 2023-05-09
Fix share control appearance on mobile

Nateowami:
I broke the layout of the share control on mobile in #1817 (the inputs are not full width like they should be):

To be fair, it was never great to begin with; the button was full width which looks very odd:

Here's how it now looks with this change:

And on desktop (where it's presumably the most likely to be used, it continues to look like this:

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1832)


RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1832)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1834, opened 2023-05-09
SF-2004 Make appAutofocus directive work with non-MDC components

Nateowami:
While experimenting with what it would take to convert the `CheckingCommentFormComponent` to Material recently, I ran into the problem that the `appAutofocus` directive did not work with the Material inputs. For MDC we would use the directive in the template by writing something like:
``` html
<mdc-text-field appAutofocus></mdc-text-field>
```
And then the MdcTextField component would be injected into `appAutofocus`. However, Material does inputs with a directive on an element, so as far as I'm aware we can't inject a component, only an element:
``` html
<input matInput type="text">
```
Hence, I've updated the directive to request an element rather than a component, and it continues to work with MDC, and should work with Material as well. I think it should generally work with any other component library, or any input we make that doesn't use a component library.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1834)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/autofocus.directive.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/autofocus.directive.ts
@@ -1,22 +1,24 @@
-import { MdcTextarea, MdcTextField } from '@angular-mdc/web/textfield';
-import { AfterViewInit, Directive, Host, Optional, Self } from '@angular/core';
+import { AfterViewInit, Directive, ElementRef } from '@angular/core';
 
 /**
- * Auto focuses MdcTextField and MdcTextarea. HTML autofocus attribute does not work for dynamically generated content.
+ * Auto focuses text inputs and textarea. HTML autofocus attribute does not work for dynamically generated content.
  */
 @Directive({
   selector: '[appAutofocus]'
 })
 export class AutofocusDirective implements AfterViewInit {
-  private component: { focus: () => void };
+  constructor(private readonly elementRef: ElementRef<HTMLElement>) {}
 
-  // Angular allows injecting the component or element but doesn't have a good way to handle components of variable type
-  // See https://github.com/angular/angular/issues/8277#issuecomment-323678013 for this workaround
-  constructor(@Host() @Self() @Optional() textField: MdcTextField, @Host() @Self() @Optional() textArea: MdcTextarea) {
-    this.component = (textField || textArea)!;
-  }
+  ngAfterViewInit(): void {
+    const hostElement: HTMLElement = this.elementRef.nativeElement;
 
-  ngAfterViewInit() {
-    setTimeout(() => this.component.focus(), 0);
+    const input =
+      hostElement instanceof HTMLInputElement || hostElement instanceof HTMLTextAreaElement
+        ? hostElement
+        : hostElement.querySelector('input, textarea');
+
+    if (input instanceof HTMLElement) {
+      setTimeout(() => input.focus());
+    }
   }
 }


RaymondLuong3 (changes requested):
What is a good way to test this fix? Can you come up with reproduction steps that I can follow?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1834#-:-NV5YZo_BX8aabnhro5N:b-cyq2lu)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/autofocus.directive.ts` line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1834#-NV5Y1Vo3iDJ1OZQyHt-:-NV5Y1Vo3iDJ1OZQyHt0:boe9nj0) ([raw file](https://github.com/sillsdev/web-xforge/blob/5430451257f0e1ddc825cbecb7d6427bbab4a02b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/autofocus.directive.ts#L13)):*
> ```XML
> 
>   ngAfterViewInit(): void {
>     const hostElement = this.elementRef.nativeElement;
> ```

Can you add the types here and below?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/autofocus.directive.ts` line 16 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1834#-NV5YFwv1dfztmByCi2Q:-NV5YFwv1dfztmByCi2R:bd0ulne) ([raw file](https://github.com/sillsdev/web-xforge/blob/5430451257f0e1ddc825cbecb7d6427bbab4a02b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/autofocus.directive.ts#L16)):*
> ```XML
>     const input = hostElement instanceof HTMLInputElement ? hostElement : hostElement.querySelector('input, textarea');
> 
>     if (input) {
> ```

In this situation, I would do `!!input` to assert it as a boolean, but I know there is nothing wrong  with this either.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1834)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/autofocus.directive.ts` line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1834#-NV5Y1Vo3iDJ1OZQyHt-:-NV5ommO5vCujAbijaTf:b1ad9k5) ([raw file](https://github.com/sillsdev/web-xforge/blob/5430451257f0e1ddc825cbecb7d6427bbab4a02b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/autofocus.directive.ts#L13)):*

Can you add the types here and below?
</blockquote></details>

Good catch. `ElementRef.nativeElement` is `any` by default. I changed the type to `ElementRef<HTMLElement>` so now it will be `HTMLElement`.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/autofocus.directive.ts` line 16 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1834#-NV5YFwv1dfztmByCi2Q:-NV5pka46BsXkkHvn9Ai:bbjuez8) ([raw file](https://github.com/sillsdev/web-xforge/blob/5430451257f0e1ddc825cbecb7d6427bbab4a02b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/autofocus.directive.ts#L16)):*

In this situation, I would do `!!input` to assert it as a boolean, but I know there is nothing wrong  with this either.
</blockquote></details>

Typically we do a null check here, which is what I meant to do. However, in this case, in addition to possibly being null, the return type of `HTMLElement.querySelector()` is `Element`, which lacks the `focus()`method. So I've checked if it's an instance of `HTMLElement` instead of checking if it's not null.

Nateowami:
@RaymondLuong3 Acceptance tests on on the issue.

RaymondLuong3 (approved):
I see that now. Thanks.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1834#-:-NVVPJRS61FJJLdoqvuF:bky699p)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1834)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1823, opened 2023-05-04
SF-1998 Fix help links

Nateowami:
1. Right now the help links in the import questions dialog do not go to the right place anymore. I've updated them to link to the new help site.
2. I added a link to the manual
3. I fixed the locale paths (see comments)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1823)

PR #1840, opened 2023-05-10
SF-1989 Update users table to material

RaymondLuong3:
* add icons to indicate additional user info
* add a more menu to contain additional actions for users
* update component to use angular material components

Before
image

After (Updated)
collaborators updated desktop

Mobile (Updated)
collaborators updated mobile

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840)


RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840)* status: 0 of 6 files reviewed, 1 unresolved discussion

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html` line 62 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NV6kk0uEtTTStdW17K2:-NV6kk0uEtTTStdW17K3:bikppru) ([raw file](https://github.com/sillsdev/web-xforge/blob/6c482bcb9a96d174be3157e7dd2f268076c9bf08/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html#L62)):*
> ```HTML
>             <td mat-cell *matCellDef="let userRow">
>               <div title="{{ t('allow_add_edit_questions') }}" *ngIf="userRow.allowCreatingQuestions">
>                 <mat-icon>library_add</mat-icon>
> ```

I can't seem to find the right icon to use here. Let me know if you have suggestions.

Nateowami:
Feedback from our conversation:
- [ ] Consider using the add question icon from the add question button
- [ ] Consider making the notice outlined so it's not so visually distracting
- [ ] Consider showing the role under the user name so it can be shown on mobile
- [ ] Consider adding a mat-tooltip to the icons indicating user is on Paratext and/or has question management permission
- [ ] Consider adding two more tabs: "Project guests" and "Invited"
- [ ] Consider renaming renaming "Paratext" tab to "Paratext member" or something (wording needs to be worked on; this is a half-baked suggestion)
- [ ] Consider consolidating the two notices about roles into one, shortening it, and ending it with a learn more link.
- [ ] Test whether the pagination works, and remove it if it doesn't.

RaymondLuong3 (commented):
Thanks for the suggestions, and the useful check boxes. I've made the requested changes except for two.

For the tabs, I added the Project Guests tab, but instead of added a Fourth tab, I just included invited guests into the Project Guests tab. I think that is sufficient for now. 3 tabs seem clean, and displays well in mobile viewports, the fourth tab begins to clutter the tab selector, and I'm not sure how beneficial it is.

For the user permission description, I followed the mock up from the original issue. I think the information may be helpful and having it on screen is nice. If I remember correctly, we have already trimmed the instructions down once already and settled on the current version.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-:-NVHBy0D8Z5dOCQrpD3T:bl701t6)* status: 0 of 9 files reviewed, 1 unresolved discussion

src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html
```diff
@@ -1,4 +1,7 @@
-<div [ngClass]="{ outline }" [class]="type">
-  <mat-icon *ngIf="icon">{{ icon }}</mat-icon>
-  <span><ng-content></ng-content></span>
-</div>
+<ng-container *transloco="let t; read: 'notice'">
+  <div [ngClass]="{ outline }" [class]="type">
+    <mat-icon *ngIf="icon">{{ icon }}</mat-icon>
+    <span class="notice-content"><ng-content></ng-content></span>
+    <a mat-flat-button *ngIf="helpLink" [href]="helpLink" target="_blank">{{ t("learn_more") }}</a>
```
Nateowami, line 5: How about adding this to the ExternalUrlService:
``` ts
get rolesHelpPage(): string {
  return this.helps + '/Concepts/Roles.htm';
}
```
And then referencing it here.

Keep in mind #1823 though; it's making needed changes to the way we link to helps, which may impact this.

Then we can get rid of `@Input() helpLink?: string;`. I don't think the parent component should be responsible for figuring out where the roles help page is at.

Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840)* status: 1 of 9 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html` line 5 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1840#gh-1194294851:-NVjSfco7EQP5CsplKE8:b-hje38l) ([raw file](https://github.com/sillsdev/web-xforge/blob/6554ea3ced2dc18565cd18be3c09d11633a3b87f/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html#L5)):*

How about adding this to the ExternalUrlService:
``` ts
get rolesHelpPage(): string {
  return this.helps + '/Concepts/Roles.htm';
}
```
And then referencing it here.

Keep in mind #1823 though; it's making needed changes to the way we link to helps, which may impact this.

Then we can get rid of `@Input() helpLink?: string;`. I don't think the parent component should be responsible for figuring out where the roles help page is at.
</blockquote></details>

I mis-read what was happening here when I wrote this. I still think it would be good to put in the the external URL service, but more importantly I think it should be removed from the notice service.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840)* status: 1 of 9 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html` line 5 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1840#gh-1194294851:-NVk8JATAmtbzhLH98j7:b-h1j0gx) ([raw file](https://github.com/sillsdev/web-xforge/blob/6554ea3ced2dc18565cd18be3c09d11633a3b87f/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html#L5)):*

I mis-read what was happening here when I wrote this. I still think it would be good to put in the the external URL service, but more importantly I think it should be removed from the notice service.
</blockquote></details>

Ok, I understand that the best thing is to keep the learn more button separate from the notice component. Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html` line 5 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVjO8KT-0ttTEs5Jh0v:-NVk8S5a-Ps3wNsPMO71:b-bzm9j1) ([raw file](https://github.com/sillsdev/web-xforge/blob/6554ea3ced2dc18565cd18be3c09d11633a3b87f/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html#L5)):*

This doesn't seem like something that belongs in the notice component.

Can we not just interpolate the button?
``` html
<app-notice icon="help">
  Some message <button>Learn more</button>
</app-notice>

If we want to make something like this reusable, it might make sense to make a help component that in turn uses the notice component, but I don't think we want to turn the notice component into something that does everything we might need.
```

</blockquote></details>

Good thinking. It seems to work with interpolation. Thanks.

Nateowami (changes requested):
> Reviewed 1 of 8 files at r2, 4 of 6 files at r3.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840)* status: 6 of 10 files reviewed, 7 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html` line 1 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVkbXd20PvWPtnzSOxM:-NVkbXd20PvWPtnzSOxN:b-ranqf9) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f0927c4d181c0a6ac3fb4b81c6ce6d01c286084/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html#L1)):*
> ```HTML
> <ng-container *transloco="let t; read: 'notice'">
> ```

Pretty sure the `<ng-container>` can be removed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.ts` line 12 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVkbgQ66aOzv9kLRxsf:-NVkbgQ66aOzv9kLRxsg:b91ugwr) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f0927c4d181c0a6ac3fb4b81c6ce6d01c286084/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.ts#L12)):*
> ```XML
>   @Input() type: 'normal' | 'warning' | 'error' = 'normal';
>   @Input() outline: boolean = false;
>   @Input() helpLink?: string;
> ```

This should be removed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html` line 13 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVkblYHEQfRh49XYgRO:-NVkblYHEQfRh49XYgRP:b-cbpvbn) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f0927c4d181c0a6ac3fb4b81c6ce6d01c286084/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html#L13)):*
> ```HTML
>     }}</span>
>     <app-notice icon="live_help" [outline]="true"
>       >{{ t("uses_roles_to_access_project")
> ```

Is it a problem to have a space between `>` and `{`? There are situations where it matters, and for that reason Prettier will not add a space for you. But if you do add a space, it prints a lot nicer. And I don't think it would cause a problem in this situation.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html` line 33 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVkeF_-C4Jb8jz38g05:-NVkeF_-C4Jb8jz38g06:bccdsax) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f0927c4d181c0a6ac3fb4b81c6ce6d01c286084/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html#L33)):*
> ```HTML
>     </div>
>     <div *ngIf="!isLoading">
>       <div *ngIf="totalUsers > 0">
> ```

If it's loading or the number of users for the tab is 0, there should be some kind of message indicating that. Often if I go to a tab it's empty, or it starts empty and after a second or two the data loads.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.scss` line 23 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVkcFweBiVFuvEbrCme:-NVkcFweBiVFuvEbrCmf:br9zgkm) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f0927c4d181c0a6ac3fb4b81c6ce6d01c286084/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.scss#L23)):*
> ```SCSS
>   width: 24px;
>   height: 24px;
>   margin-right: 8px;
> ```

How does this look in RTL? I can't really tell that this is needed at all; the spacing seems fine without it. But if we're going to add spacing it should work with RTL.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts` line 56 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVkeaw1057e_kjkdHzM:-NVkeaw1057e_kjkdHzN:b-v4tqu4) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f0927c4d181c0a6ac3fb4b81c6ce6d01c286084/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts#L56)):*
> ```XML
>   filterForm: UntypedFormGroup = new UntypedFormGroup({ filter: new UntypedFormControl('') });
>   isAppOnline = true;
>   showParatextOnly = false;
> ```

This isn't being used.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/external-url.service.ts` line 21 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVkes3KChJnhdUfZMd6:-NVkes3KChJnhdUfZMd7:b-iy11fu) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f0927c4d181c0a6ac3fb4b81c6ce6d01c286084/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/external-url.service.ts#L21)):*
> ```XML
> 
>   get rolesHelpPage(): string {
>     return this.getUrlToManual('/#t=Concepts%2FRoles.htm');
> ```

How about `/Concepts/Roles.htm`? That's how the help files link it. The hash part only happens after a redirect. The examples lower down in the file aren't great, but they're being replaced in #1823.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840)* status: 6 of 10 files reviewed, 7 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html` line 1 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVkbXd20PvWPtnzSOxM:-NVom30lDLGtDDLr8BdM:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f0927c4d181c0a6ac3fb4b81c6ce6d01c286084/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html#L1)):*

Pretty sure the `<ng-container>` can be removed.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.ts` line 12 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVkbgQ66aOzv9kLRxsf:-NVom6HWALHbW26FK8Pc:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f0927c4d181c0a6ac3fb4b81c6ce6d01c286084/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.ts#L12)):*

This should be removed.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html` line 13 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVkblYHEQfRh49XYgRO:-NVomc2pENbbFGJgRE3Z:bq1myxq) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f0927c4d181c0a6ac3fb4b81c6ce6d01c286084/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html#L13)):*

Is it a problem to have a space between `>` and `{`? There are situations where it matters, and for that reason Prettier will not add a space for you. But if you do add a space, it prints a lot nicer. And I don't think it would cause a problem in this situation.
</blockquote></details>

No, I don't think it has an effect in this situation. I agree, I think it looks neater when tags aren't broken over multiple lines if they don't need to be.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html` line 33 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVkeF_-C4Jb8jz38g05:-NVorh92EpRm3F6w1AgQ:bidhqhm) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f0927c4d181c0a6ac3fb4b81c6ce6d01c286084/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html#L33)):*

If it's loading or the number of users for the tab is 0, there should be some kind of message indicating that. Often if I go to a tab it's empty, or it starts empty and after a second or two the data loads.
</blockquote></details>

Ok, I've updated the page so that the no users label appears when there are no users when a user filters or tabs and no users match the criteria. For the loading thing, there is a test that specify that if the page is loading, it does not show the no users label. I think this is ok.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.scss` line 23 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVkcFweBiVFuvEbrCme:-NVooJuc79EycCCqD5mF:bw26a3b) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f0927c4d181c0a6ac3fb4b81c6ce6d01c286084/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.scss#L23)):*

How does this look in RTL? I can't really tell that this is needed at all; the spacing seems fine without it. But if we're going to add spacing it should work with RTL.
</blockquote></details>

You're right. Thanks for spotting that.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts` line 56 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVkeaw1057e_kjkdHzM:-NVooVfAC2MAaCB_Ip05:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f0927c4d181c0a6ac3fb4b81c6ce6d01c286084/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts#L56)):*

This isn't being used.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/external-url.service.ts` line 21 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVkes3KChJnhdUfZMd6:-NVoolrR5ruiGf1GVUlv:b-rzwzq9) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f0927c4d181c0a6ac3fb4b81c6ce6d01c286084/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/external-url.service.ts#L21)):*

How about `/Concepts/Roles.htm`? That's how the help files link it. The hash part only happens after a redirect. The examples lower down in the file aren't great, but they're being replaced in #1823.
</blockquote></details>

Unfortunately if we just use `concepts/roles.htm` I see that the help loses the navigation bar in a way where you cannot retrieve it. I'm not sure the reasoning for that.

Nateowami (changes requested):
> Reviewed 2 of 7 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840)* status: 4 of 10 files reviewed, 3 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss` line 30 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVpQfvv601Jdr8EhKFK:-NVpQfvv601Jdr8EhKFL:bw7frij) ([raw file](https://github.com/sillsdev/web-xforge/blob/fa852d17f24d7a4d682705f6e75207e01bf190cf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss#L30)):*
> ```SCSS
>     align-self: flex-start;
>   }
>   .notice-content {
> ```

I think the notice component should be kept as general as possible and not need to concern itself with how its internals are laid out.

I made a branch with a couple of changes that move this layout logic to the host component. See `task/sf-1989-users-page-suggested-changes`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html` line 128 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVpRL5207-eKmv6f2nI:-NVpRL5207-eKmv6f2nJ:b3f9lfu) ([raw file](https://github.com/sillsdev/web-xforge/blob/fa852d17f24d7a4d682705f6e75207e01bf190cf/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html#L128)):*
> ```HTML
>         </mat-paginator>
>       </div>
>       <div *ngIf="filteredLength === 0" class="no-users-label">{{ t("no_users_found") }}</div>
> ```

Can you give this a mat-hint class? The color feels out of place.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/external-url.service.ts` line 21 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVkes3KChJnhdUfZMd6:-NVp53vj6rCx_q_r27Yv:blltxk3) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f0927c4d181c0a6ac3fb4b81c6ce6d01c286084/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/external-url.service.ts#L21)):*

Unfortunately if we just use `concepts/roles.htm` I see that the help loses the navigation bar in a way where you cannot retrieve it. I'm not sure the reasoning for that.
</blockquote></details>

Good point. Somehow I missed that. Seems like a bad limitation of Robohelp.

Nateowami (commented):
I've noticed that every time I click a tab it jumps up a couple pixels and then back down. Any idea why?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-:-NVpVD4X0J2Xg1BZXnYO:b-hzuwpq)* status: 4 of 10 files reviewed, 3 unresolved discussions (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
Wow, you're right, it does. I took a look, but it happens so fast I'm not really able to see what's going on. I notice it happens on mobile to, I thought it might have to do with the filter box being on the same line, but that's not the case.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-:-NVpgPzLASuybBBBAVnM:brxoey6)* status: 4 of 10 files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss` line 30 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVpQfvv601Jdr8EhKFK:-NVpgersAMlArKE828Rl:bx662fa) ([raw file](https://github.com/sillsdev/web-xforge/blob/fa852d17f24d7a4d682705f6e75207e01bf190cf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss#L30)):*

I think the notice component should be kept as general as possible and not need to concern itself with how its internals are laid out.

I made a branch with a couple of changes that move this layout logic to the host component. See `task/sf-1989-users-page-suggested-changes`
</blockquote></details>

Nice, thanks for doing that.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html` line 128 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVpRL5207-eKmv6f2nI:-NVpgj6g2u794eKqdcVM:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/fa852d17f24d7a4d682705f6e75207e01bf190cf/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html#L128)):*

Can you give this a mat-hint class? The color feels out of place.
</blockquote></details>

Done.

Nateowami (commented):
I'm pretty sure the jumping motion is because the way you're using mat-tab is not how it's intended, if you look at the docs. I think you did this in order to place the filter to the right of the tabs?

Assuming that's correct, our options are:
1. Leave it as-is, now that we know the source of the behavior (assuming my assessment is correct)
2. Remove the filter altogether (the five projects with the most users have 23, 22, 15, 15, 13 respectively)
3. Move it to the bottom like this:



I think option 1 is fine, but we should be aware when we're using components in a way other than they were designed to be used. And it should probably have a comment so it doesn't become an example that gets copied.

> Reviewed 1 of 7 files at r4, 1 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-:-NW3Bckn8pEcFE9gHVHx:b-de1x8m)* status: 5 of 10 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html` line 128 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVpRL5207-eKmv6f2nI:-NW3B_978MQEg15xRMmg:b-tv3o43) ([raw file](https://github.com/sillsdev/web-xforge/blob/fa852d17f24d7a4d682705f6e75207e01bf190cf/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html#L128)):*

Done.
</blockquote></details>

This class is not actually getting set. I haven't figured out why.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts` line 135 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVpVSNwEjh98pfLtP3j:-NVpVSNwEjh98pfLtP3k:buf2zye) ([raw file](https://github.com/sillsdev/web-xforge/blob/fa852d17f24d7a4d682705f6e75207e01bf190cf/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts#L135)):*
> ```XML
> 
>   get tableColumns(): string[] {
>     const columns = ['avatar', 'name', 'info', 'questions_permission', 'role', 'more']; //, 'questions_permission', 'remove'];
> ```

Is this comment still supposed to be here?

RaymondLuong3 (commented):
I commented the filter box from the page, and the pixel jumping behaviour moves to the user data table. It is a strange one. I agree, the pixel jumping is likely not going to be noticeable for most users, so let's stick with it as is. I left a comment so that other developers are aware of the non-standard use of the tab component.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-:-NW8lvaw4muB89fvHQIX:bt2yyfd)* status: 5 of 10 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html` line 128 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVpRL5207-eKmv6f2nI:-NW8lXvV9MO2tLYrztYl:b-7w5z04) ([raw file](https://github.com/sillsdev/web-xforge/blob/fa852d17f24d7a4d682705f6e75207e01bf190cf/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html#L128)):*

This class is not actually getting set. I haven't figured out why.
</blockquote></details>

I've changed this to use the mat-hint component. It formats the color and looks better.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts` line 135 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NVpVSNwEjh98pfLtP3j:-NW8jYlT9WdyOgc6u9kD:bobclp9) ([raw file](https://github.com/sillsdev/web-xforge/blob/fa852d17f24d7a4d682705f6e75207e01bf190cf/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts#L135)):*

Is this comment still supposed to be here?
</blockquote></details>

Good catch. Removed.

Nateowami (commented):
How about adding this to get rid of the jumpyness?
``` scss
:host ::ng-deep .mat-tab-body-wrapper {
  display: none;
}
```

> Reviewed 1 of 4 files at r7, 1 of 4 files at r11, 1 of 3 files at r12.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-:-NWiIKQu2Ua_mt6CfCEV:bicsgdg)* status: 6 of 10 files reviewed, all discussions resolved

RaymondLuong3 (commented):
Excellent, thanks for figuring this out!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-:-NWiTcUk2A6TrUJ4farM:b-4cgvk1)* status: 5 of 10 files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami (changes requested):
I'm sorry I forgot to submit the rest of my comments.

> Reviewed 1 of 4 files at r11, 1 of 3 files at r14, 2 of 3 files at r15.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-:-NWiWXYD7_Val8nBCsqj:baobjia)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html` line 51 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NWiLmTZ3COD5tWSk1TN:-NWiLmTZ3COD5tWSk1TO:bfqrxw9) ([raw file](https://github.com/sillsdev/web-xforge/blob/01ac4d1db5c3db1c9eae663b9f6f9338b913c2aa/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html#L51)):*
> ```HTML
>                 <b *ngIf="isCurrentUser(userRow)" class="current-user-label">&nbsp;{{ t("me") }}</b>
>               </div>
>               <div *ngIf="!userRow.inviteeStatus" fxHide.gt-xs>
> ```

Shouldn't the role be shown even for invitees?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts` line 116 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NWiJWy20L_S8_44jn34:-NWiJWy20L_S8_44jn35:b-dfc0vd) ([raw file](https://github.com/sillsdev/web-xforge/blob/01ac4d1db5c3db1c9eae663b9f6f9338b913c2aa/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts#L116)):*
> ```XML
>   get userRows(): Row[] {
>     const term = this.term && this.term.trim().toLowerCase();
>     let rows: Row[] = term ? this.filteredRows : this.userRowsOnTab;
> ```

This getter is kind of kind of odd. You're converting the term to lowercase to then do nothing with it. How about something like:
``` ts
    let rows: Row[] = term == null ? this.userRowsOnTab : this.filiter(term, this.userRowsOnTab)
```

Also, there are three getters with the names:
- userRowsOnTab
- userRows
- filteredRows

It would seem "userRows" is not filtered, and "filteredRows" is, but actually both are filtered. I think a better naming scheme would be something like:
- userRowsForSelectedTab
- rowsToDisplay
- a helper method called something like `filiterRowsBySearchTerm()` ("by search term" is clearer than just "filtered", since selecting a tab also filters the rows)

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840)* status: 8 of 10 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html` line 51 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NWiLmTZ3COD5tWSk1TN:-NWibcFi7T_yZTsoEd17:b-mvbzp) ([raw file](https://github.com/sillsdev/web-xforge/blob/01ac4d1db5c3db1c9eae663b9f6f9338b913c2aa/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.html#L51)):*

Shouldn't the role be shown even for invitees?
</blockquote></details>

Hmm, it is a bit cluttered, but should be okay.

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts` line 116 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/1840#-NWiJWy20L_S8_44jn34:-NWidVIW8WyefETVRlbW:b-o5ztsu) ([raw file](https://github.com/sillsdev/web-xforge/blob/01ac4d1db5c3db1c9eae663b9f6f9338b913c2aa/src/SIL.XForge.Scripture/ClientApp/src/app/users/collaborators/collaborators.component.ts#L116)):*

This getter is kind of kind of odd. You're converting the term to lowercase to then do nothing with it. How about something like:
``` ts
    let rows: Row[] = term == null ? this.userRowsOnTab : this.filiter(term, this.userRowsOnTab)
```

Also, there are three getters with the names:
- userRowsOnTab
- userRows
- filteredRows

It would seem "userRows" is not filtered, and "filteredRows" is, but actually both are filtered. I think a better naming scheme would be something like:
- userRowsForSelectedTab
- rowsToDisplay
- a helper method called something like `filiterRowsBySearchTerm()` ("by search term" is clearer than just "filtered", since selecting a tab also filters the rows)
</blockquote></details>

Ok, I've updated the names. What was previously filteredRows is actually the rows filtered by search term and by tab.  So I've updated the name to be filteredRowsBySearchTermAndTab.

Nateowami (approved):
> Reviewed 3 of 3 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

Nateowami (approved):
> Reviewed 1 of 6 files at r17, 1 of 3 files at r18, 4 of 4 files at r19, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1840)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1842, opened 2023-05-15
SF-1459 Fix checking overview question list not always updating

Nateowami:
~This may also fix [SF-1756](https://jira.sil.org/browse/SF-1756).~

It appears the problem is that `throttleTime()` by default does not emit the last value by default.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1842)


josephmyers:
Screenshot 2023-05-16 153308

Can repro on master. Verified the fix on this branch.

josephmyers (approved):

josephmyers:
I also confirmed that SF-1756 still remains. The verse is updated, but the chapter number is not (this behavior was present on master, as well)

image

PR #1844, opened 2023-05-16
SF-2014 Convert checking comment form to Material

Nateowami:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1844)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
@@ -1846,7 +1846,7 @@ class TestEnvironment {
   }
 
   get commentFormTextFields(): DebugElement[] {
-    return this.fixture.debugElement.queryAll(By.css('mdc-text-field[formControlName="commentText"]'));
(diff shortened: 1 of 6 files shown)


src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
```diff
@@ -2179,7 +2179,11 @@ class TestEnvironment {
   }
 
   setTextFieldValue(textField: DebugElement, value: string): void {
-    const input = textField.query(By.css('input, textarea'));
+    // The text field may be either an input/textarea, or a component with an input/textarea within it.
```
josephmyers, line 2182: Can we remove MDC altogether? Seems like that'd be cleaner. I only see one remaining usage of it

josephmyers: Plus I don't see a JIRA task that captures that last usage

josephmyers: Talked in person. We'll put this work off until another subtask

josephmyers (approved):

PR #1852, opened 2023-05-22
SF-1960 Set display name when a user signs up with a phone number

nigel-wells:
- Update name, nickname, avatar on Auth0 profile if set to phone_number

*Notes*
* Decided to go with Anonymous rather than trying to generate a random name which may just cause other confusion
* Decided not to use any kind of gravatar type image which either needed an email address or storage for a randomly generated image in order for it to work in both the Auth0 profile and be accessible by our Angular Avatar component.
* As soon as a user updates their display name it will also update their avatar to something relevant
* When signing up for the first time on localhost auth0 tenant (sil-appbuilder) you get a oauth prompt which displays the phone number. At this point the name hasn't been updated. This prompt only appears on this tenant and is not displayed on QA/Live
* The implemented approach will resolve any user currently in this state the next time the log in to SF

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1852)

### DIFF
--- a/src/SIL.XForge/Services/AuthService.cs
+++ b/src/SIL.XForge/Services/AuthService.cs
@@ -79,6 +79,20 @@ public class AuthService : DisposableBase, IAuthService
         return CallApiAsync(new HttpMethod("PATCH"), $"users/{authId}", content);
     }
 
+    /// <summary>
+    /// Used for SMS authentication as the user has their name and nickname set to their phone number which may get
+    /// exposed in future user requests. This method sets both values to "Anonymous" and is intended to be run once.
+    /// </summary>
+    public async Task<string> UpdateUserToAnonymous(string authId)
+    {
+        var content = new JObject(
+            new JProperty("name", "Anonymous"),
+            new JProperty("nickname", "Anonymous"),
+            new JProperty("picture", "https://cdn.auth0.com/avatars/a.png")
+        );
+        return await CallApiAsync(new HttpMethod("PATCH"), $"users/{authId}", content);
+    }
+
     private async Task<string> CallApiAsync(HttpMethod method, string url, JToken content = null)
     {
         bool refreshed = false;
--- a/src/SIL.XForge/Services/IAuthService.cs
+++ b/src/SIL.XForge/Services/IAuthService.cs
@@ -11,4 +11,5 @@ public interface IAuthService
     Task LinkAccounts(string primaryAuthId, string secondaryAuthId);
     Task UpdateAvatar(string authId, string url);
     Task UpdateInterfaceLanguage(string authId, string language);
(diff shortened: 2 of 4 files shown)


pmachapman (approved):
:lgtm: Great work - the test case clearly explained the change, and the method comments are succinct.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1852#-:-NW0Je1QAosmTLZNKJYa:b-akirxr)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

Nateowami (dismissed):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1852)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge/Services/AuthService.cs` line 71 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1852#-NW2nT-aACKHTE2txT13:-NW2nT-b6weTzyWlgmap:b-iq9dgk) ([raw file](https://github.com/sillsdev/web-xforge/blob/21bb950ba341ebf4a3485d033ffe660d3764e932/src/SIL.XForge/Services/AuthService.cs#L71)):*
> ```Smalltalk
>     {
>         var content = new JObject(
>             new JProperty("name", "Anonymous"),
> ```

Can it not be set to undefined, null, or the empty string? There's a significant difference between a user not having their display name set and having it set to "Anonymous". It would result in this confirmation dialog which is kind of odd. Also, it won't be localized.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1852)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge/Services/AuthService.cs` line 71 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1852#-NW2nT-aACKHTE2txT13:-NW4BXMw04MaAUydbx5m:b-1yjf2b) ([raw file](https://github.com/sillsdev/web-xforge/blob/21bb950ba341ebf4a3485d033ffe660d3764e932/src/SIL.XForge/Services/AuthService.cs#L71)):*

Can it not be set to undefined, null, or the empty string? There's a significant difference between a user not having their display name set and having it set to "Anonymous". It would result in this confirmation dialog which is kind of odd. Also, it won't be localized.


</blockquote></details>

We'd need to change a number of screens to deal with a display name not being set i.e. Users menu, users page (for admin), presence, etc. and seems a bit much to change all of them rather than simply ensuring the display name is set to something in the first place which is what our app currently expects.

The confirm name dialog probably doesn't need to have the display name already set - it could easily be made blank regardless of it being set to "Anonymous".

PR #1860, opened 2023-05-24
SF-2031 Fix handling of unexpected errors in join component

nigel-wells:
- Catch exception when Auth0 returns a bad response code and show a generic message on the front end

Testing of this either requires intercepting local network requests OR removing a permission for `create:users` in Auth0.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1860)


Nateowami:
What if it's a different error (e.g. a gateway timeout)? I don't think we can assume that we can predict all possible errors and that it will be one of the localization keys.

What about changing this:
``` ts
this.dialogService.message(of(this.transloco.translate(`join.${error}`)))
```
To something like:
``` ts
const key = ['error_occurred_login', 'max_users_reached', 'please_connect_to_use_link', 'project_link_is_invalid'].includes(error) ? error : 'unknown_error';
// notice that transloco.translate() and of() are not needed
this.dialogService.message(`join.${key}`);
```

To be clear, I'm not saying the change in this PR shouldn't be done, just that I think it's not really cutting off the problem at the root.

nigel-wells (commented):
I've added in two more exceptions to catch - one for `SecurityException` (401) and `taskCanceledException` (timeout). I also spoke with Peter to ensure the errors are still reported to bugsbang. I also changed the returned response to no content which also triggers the generic error message. I don't think the user needs to know specifically that Auth0 exists and that it had a specific timeout, 401, realm issue, etc.

I have SF-1939 setup to improve errors in the join component so I'll take your comments above and include it in there.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1860#-:-NWEyftL5CehiH7KVVK4:beipws8)* status: 0 of 2 files reviewed, all discussions resolved

Nateowami:
@marksvc Regarding type safety of the error messages, see #1879

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1860)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Controllers/AnonymousController.cs` line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1860#-NWn9FeDA6DM7QfHKOyj:-NXI2XnECulVR6x8oD25:brud0wn) ([raw file](https://github.com/sillsdev/web-xforge/blob/5d38a5f631c66b002f7b60b4d1cfa223fb70421c/src/SIL.XForge.Scripture/Controllers/AnonymousController.cs#L92)):*

I'm not super familiar with the API, but wouldn't we want to do this RecordEndpointInfoForException before the above ReportException calls? So, perhaps RecordEndpointInfoForException should be performed at the beginning of the method, and then whenever we `ReportException` or `throw`, we will have the endpoint information. 
</blockquote></details>

That is an interesting point. @pmachapman do you have any objections or thoughts around why we should or shouldn't simply record the info at the start of the method regardless of an exception needing to be handled? It looks like it just stored in bugsnag but only submitted when we notify bugsnag. I guess if another class had an exception we may end up cross contaminating that exception with data from this class?

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1860)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/Controllers/AnonymousController.cs` line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1860#-NWn9FeDA6DM7QfHKOyj:-NXIxhDBAECPSQlkB_M9:bcq7owf) ([raw file](https://github.com/sillsdev/web-xforge/blob/39fd8180faca4f6b6243302bf13984aef8175f4c/src/SIL.XForge.Scripture/Controllers/AnonymousController.cs#L92)):*

That is an interesting point. @pmachapman do you have any objections or thoughts around why we should or shouldn't simply record the info at the start of the method regardless of an exception needing to be handled? It looks like it just stored in bugsnag but only submitted when we notify bugsnag. I guess if another class had an exception we may end up cross contaminating that exception with data from this class?
</blockquote></details>

That should be OK. The information that is recorded is the endpoint info, and the `IExceptionHandler` is scoped, so I don't think there will not be any contamination.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1860)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells and @pmachapman)

___
*[`src/SIL.XForge.Scripture/Controllers/AnonymousController.cs` line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1860#-NWn9FeDA6DM7QfHKOyj:-NXlm7gTCj0-Kf1qG-QO:b-bv0h7s) ([raw file](https://github.com/sillsdev/web-xforge/blob/39fd8180faca4f6b6243302bf13984aef8175f4c/src/SIL.XForge.Scripture/Controllers/AnonymousController.cs#L92)):*

That should be OK. The information that is recorded is the endpoint info, and the `IExceptionHandler` is scoped, so I don't think there will not be any contamination.
</blockquote></details>

It sounds like putting `RecordEndpointInfoForException` at the beginning of the method is going to be okay and will help to provide more information for when the above `_exceptionHandler.ReportException(e);` lines run. I would advocate for doing that unless there's something I'm not understanding.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1860)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @marksvc and @pmachapman)

___
*[`src/SIL.XForge.Scripture/Controllers/AnonymousController.cs` line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1860#-NWn9FeDA6DM7QfHKOyj:-NXn6ZnL9k_Gvnz4mjnL:b-yd3vm4) ([raw file](https://github.com/sillsdev/web-xforge/blob/39fd8180faca4f6b6243302bf13984aef8175f4c/src/SIL.XForge.Scripture/Controllers/AnonymousController.cs#L92)):*

It sounds like putting `RecordEndpointInfoForException` at the beginning of the method is going to be okay and will help to provide more information for when the above `_exceptionHandler.ReportException(e);` lines run. I would advocate for doing that unless there's something I'm not understanding. 
</blockquote></details>

I've moved it to the top as suggested.

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1860)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

PR #1864, opened 2023-05-24
SF-1934 Fix RTL language marker not recognized in combined verses

RaymondLuong3:
The right-to-left marker exists in the usfm of right-to-left projects on combined verses. This was causing problems because when we looked up text for a specific verse, if the right-to-left marker existed on the op, we would not get a match. This change helps to generate a verse ref that correctly matches what exists on the segment attribute of text ops so we can correctly look up the text strings in a text doc so they can be displayed in the note dialog.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1864)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/text-doc.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/text-doc.ts
@@ -7,6 +7,7 @@ import {
   TEXT_INDEX_PATHS
 } from 'realtime-server/lib/esm/scriptureforge/models/text-data';
 import { RealtimeDoc } from 'xforge-common/models/realtime-doc';
+import { RIGHT_TO_LEFT_MARK, VERSE_FROM_SEGMENT_REF_REGEX } from '../../shared/utils';
 
 export const Delta: new (ops?: DeltaOperation[] | { ops: DeltaOperation[] }) => DeltaStatic = Quill.import('delta');
 
@@ -82,7 +83,7 @@ export class TextDoc extends RealtimeDoc<TextData, TextData, RangeStatic> {
     return text;
   }
 
-  getSegmentTextIncludingRelated(ref: string): string {
+  getSegmentTextIncludingRelated(verseStr: string): string {
     if (this.data == null || this.data.ops == null) {
       return '';
     }
@@ -101,10 +102,9 @@ export class TextDoc extends RealtimeDoc<TextData, TextData, RangeStatic> {
         continue;
       }
       // Locate range of ops that match the verse segments
-      if (
-        op.attributes?.segment != null &&
-        (op.attributes.segment === ref || op.attributes.segment.indexOf(ref + '/') === 0)
-      ) {
+      const opSegmentRef: string = op.attributes?.segment ?? '';
+      const match: RegExpMatchArray | null = VERSE_FROM_SEGMENT_REF_REGEX.exec(opSegmentRef);
+      if (match != null && match[1].replace(RIGHT_TO_LEFT_MARK, '') === verseStr) {
         text += textBetweenRelatedSegments + op.insert;
         // Reset text so no double-ups
         textBetweenRelatedSegments = '';
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/test-utils.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/test-utils.ts
@@ -4,7 +4,7 @@ import { isParatextRole, SFProjectRole } from 'realtime-server/lib/esm/scripture
 import { TextData } from 'realtime-server/lib/esm/scriptureforge/models/text-data';
 import { CheckingAnswerExport } from 'realtime-server/lib/esm/scriptureforge/models/checking-config';
 import { Delta, TextDocId } from '../core/models/text-doc';
-import { rightLeftMarker } from './utils';
+import { RIGHT_TO_LEFT_MARK } from './utils';
 
 export function getTextDoc(id: TextDocId): TextData {
   const delta = new Delta();
@@ -38,7 +38,7 @@ export function getTextDoc(id: TextDocId): TextData {
 }
 
 export function getCombinedVerseTextDoc(id: TextDocId, rtl: boolean = false): TextData {
-  const verseStr: string = rtl ? `2${rightLeftMarker}-3` : '2-3';
+  const verseStr: string = rtl ? `2${RIGHT_TO_LEFT_MARK}-3` : '2-3';
   const delta = new Delta();
   delta.insert(`Title for chapter ${id.chapterNum}`, { segment: 's_1' });
   delta.insert('\n', { para: { style: 's' } });
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/test-utils.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/test-utils.ts
@@ -4,6 +4,7 @@ import { isParatextRole, SFProjectRole } from 'realtime-server/lib/esm/scripture
 import { TextData } from 'realtime-server/lib/esm/scriptureforge/models/text-data';
 import { CheckingAnswerExport } from 'realtime-server/lib/esm/scriptureforge/models/checking-config';
 import { Delta, TextDocId } from '../core/models/text-doc';
+import { rightLeftMarker } from './utils';
 
 export function getTextDoc(id: TextDocId): TextData {
   const delta = new Delta();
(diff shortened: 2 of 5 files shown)


Nateowami (commented):
I don't think this will actually solve the problem in every instance, since it assumes the RTL mark will be there or not be there only based on whether the project is RTL. A project I've been working with is a LTR back translation of a RTL project, and has RTL marks in at least some of the verse ranges (presumably due to being a daughter project). It shouldn't be that way, but I'm guessing it's not that uncommon. And it's really hard for users to detect the presence of the marks.

It looks to me like the verse slug is used in only one place, and that's to pass it to `TextDoc.getSegmentTextIncludingRelated()`. Maybe that method could just be updated to accept a VerseRef and have more relaxed matching logic?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1864#-:-NWI3kFPCt0__f_5F0SL:bvnsyn2)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
I see what you mean. With the projects that have the erroneous rlm, it is likely that the previous code would not work since the segment ref with the rlm might exist regardless of what the project isRightToLeft setting is set to. I've updated it so that there is some more wiggle room to account for rlm. Let me know what you think about this.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1864#-:-NWJQlHS2AjPQ6WyjLlV:b-mzn1kp)* status: 1 of 5 files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts` line 12 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1864#-NWFOR0g6kPTzvmXVbnp:-NWJSYcD9V5LUxxJKYMm:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/eb15032d844b9a16f804624303553e05dfa97ce7/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L12)):*

This should be in caps like most other constants are, i.e. `RIGHT_TO_LEFT_MARK`.

Also, naming it `RIGHT_TO_LEFT_MARK` (as opposed to right left marker) will mean that if someone searches it in Google, they will get the Wikipedia page and find out all about it.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts` line 10 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1864#-NWJSf0h5eC-DQweQsFH:-NWJSf0h5eC-DQweQsFI:b-uqmrdk) ([raw file](https://github.com/sillsdev/web-xforge/blob/615e3d55178d943ec2f4a19d6af68751fbf4d173/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L10)):*
> ```XML
> // Regular expression for getting the verse from a segment ref
> // Some projects will have the right to left marker in the segment attribute which we need to account for
> export const VERSE_FROM_SEGMENT_REF_REGEX = /verse_\d+_(\d+[‏]?-?\d*[^\/]?)/;
> ```

You cannot see it in reviewable, but inside the empty square bracket is the [U+200F] which shows up in my VS code.

Nateowami (changes requested):
There's another very similar situation I don't think this addresses. See https://docs.usfm.bible/usfm-usx-docs/latest/bidirectional.html

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1864#-:-NWJUOtFE5Jgj82Pv0hG:b17oste)* status: 1 of 5 files reviewed, 2 unresolved discussions (waiting on @pmachapman and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts` line 10 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1864#-NWJSf0h5eC-DQweQsFH:-NWJU__x3b2sBR2bXACt:bxakyhv) ([raw file](https://github.com/sillsdev/web-xforge/blob/615e3d55178d943ec2f4a19d6af68751fbf4d173/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L10)):*

You cannot see it in reviewable, but inside the empty square bracket is the [U+200F] which shows up in my VS code.
</blockquote></details>

Let's avoid putting invisible control characters in the source code and reference by its number instead.
``` ts
verse_\d+_(\d+\u200F?-?\d*[^\/]?)
```

RaymondLuong3 (commented):
Hmm, I don't think we will see commas in verse segments. So I'm not sure what you were directing me to. Can you explain further?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1864#-:-NWJlVup0FIoV4GKXmD9:b-5j33f7)* status: 1 of 5 files reviewed, 2 unresolved discussions (waiting on @Nateowami and @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts` line 10 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1864#-NWJSf0h5eC-DQweQsFH:-NWJlk_B0DAvuHiKqZQF:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/615e3d55178d943ec2f4a19d6af68751fbf4d173/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L10)):*

Let's avoid putting invisible control characters in the source code and reference by its number instead.
``` ts
verse_\d+_(\d+\u200F?-?\d*[^\/]?)
```
</blockquote></details>

Done.

Nateowami (commented):
It's a perfectly valid way of specifying verses in USFM. We have 119 instances of it on live. I wasn't able to find any instances that contained a RTL mark, though grepping for RTL marks is really tricky. It may be that no one is using it that way on live at the moment. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1864#-:-NWJszupAvaVPH_7TQpa:blqmn5z)* status: 1 of 5 files reviewed, 2 unresolved discussions (waiting on @pmachapman)

RaymondLuong3 (commented):
Ok, you were right. I was not aware that we could combine verses with a comma. It looks like our system already understands that, so I just had to add a little to the regex so that it understands that a comma is valid in a verse

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1864#-:-NWK4sqa1wOEzVEwN0Zo:btq3o7n)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @pmachapman)

pmachapman (commented):
> Reviewed 1 of 4 files at r2, 1 of 2 files at r3, 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1864)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts` line 10 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1864#-NWYmFFn8IwxI4o76k3d:-NWYmFFn8IwxI4o76k3e:bamslsw) ([raw file](https://github.com/sillsdev/web-xforge/blob/7198669d24c159f69384ecfb78eb8a8fde571b39/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L10)):*
NIT: I think the RegEx should be:

> /verse_\d+_(\d+[\u200f]?[,\-]?\d*[^\/]?)/

The current regex allows both a dash and a comma at the same between verse numbers.



_Code quote:_
```XML
-?,?
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1864)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts` line 10 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/1864#-NWYmFFn8IwxI4o76k3d:-NWc9PaI2m22zj0jqr_R:bi3aiej) ([raw file](https://github.com/sillsdev/web-xforge/blob/7198669d24c159f69384ecfb78eb8a8fde571b39/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L10)):*

NIT: I think the RegEx should be:

> /verse_\d+_(\d+[\u200f]?[,\-]?\d*[^\/]?)/

The current regex allows both a dash and a comma at the same between verse numbers.
</blockquote></details>

Thanks for always looking at my regex's so thoroughly!

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1864#-:-NWcuWax2PPn0hhQ_UYu:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1864)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1869, opened 2023-05-25
SF-2040 Autofocus join component

Nateowami:
Before | After
-------|------
 | 

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1869)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.html
@@ -4,9 +4,9 @@
     <p [innerHTML]="inviteText"></p>
     <app-notice *ngIf="!isOnline" type="error" icon="cloud_off">{{ t("please_connect_to_use_link") }}</app-notice>
     <form>
-      <mat-form-field appearance="fill">
+      <mat-form-field appearance="outline">
         <mat-label>{{ t("your_name") }}</mat-label>
-        <input matInput [formControl]="name" />
+        <input matInput [formControl]="name" appAutofocus />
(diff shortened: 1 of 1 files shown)


Nateowami:
Now that I've opened the PR I see that the placeholder "Enter your name" ends up being much smaller due to the input being focused. So the user loses the opportunity to see the text first. What do you think @nigel-wells?

nigel-wells:
Have you checked on mobile? Main concern would be the keyboard opening which may remove too much of the viewport for the user to read the text above which gives context as to why they needed to enter their name in the first place.

Nateowami:
@nigel-wells No, I don't have a good setup to test localhost on mobile. When I've tried in the past I've not been successful, and I haven't tried recently.

Nateowami:
Just tested on QA. On my device the "or log in to an existing account" part gets covered by the keyboard.

nigel-wells:
Browserstack works ok for Android testing if you wanted to give that a go - I could try it out in a bit on my device as well.

Nateowami:
I think the layout could be made to work so that the keyboard didn't cover anything up, but I'm starting to think there may be value to showing a mostly blank page, since it makes it more obvious that there is one and only one thing the user is being asked to provide. There may be a psychological aspect to it too -- a form looks shorter and smaller if surrounded by a lot of white space.

nigel-wells:
I'd lean more towards not focusing on the input and showing the keyboard. To this point the user has given an affirmative action to view a website only (by clicking the link) and not to start filling out a form - it may come across a little pushy/aggressive.

PR #1870, opened 2023-05-26
SF-1981 Fix editor highlights incorrect source segment

siltomato:


RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1870)* status: all files reviewed, 1 unresolved discussion (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1296 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1870#-NWP724E5xT8WG9H8vDG:-NWP724E5xT8WG9H8vDH:b8ugnl3) ([raw file](https://github.com/sillsdev/web-xforge/blob/4a64fd1fee6f97d62d2ebe3d9d960850193fcf9b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1296)):*
> ```XML
> 
>       // If verse segment ref has no exact match, check for segments that fall within a verse reference
>       if (segmentRef.startsWith('verse')) {
> ```

Nice work figuring this out. Let's use the VERSE_REGEX, so something like if (VERSE_REGEX.text(segmentRef)), then we can split it into its components and create a verseRef for it.

I see we do not currently have a test in place to test whether we highlight the corresponding segment in a source test. It's not necessary, but feel free to explore writing a test to show that your fix works. It would require text docs that do not all have corresponding segments. You can see in the editor.spec.ts TestEnvironment constructor that we have source texts, but they use the exact same mock TextData.

If it were not so complicated, I would ask you to include a test.

siltomato (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1870)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1296 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1870#-NWP724E5xT8WG9H8vDG:-NWkR0o_-xQu36eVN7Zt:b-3ri40k) ([raw file](https://github.com/sillsdev/web-xforge/blob/4a64fd1fee6f97d62d2ebe3d9d960850193fcf9b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1296)):*

Nice work figuring this out. Let's use the VERSE_REGEX, so something like if (VERSE_REGEX.text(segmentRef)), then we can split it into its components and create a verseRef for it.

I see we do not currently have a test in place to test whether we highlight the corresponding segment in a source test. It's not necessary, but feel free to explore writing a test to show that your fix works. It would require text docs that do not all have corresponding segments. You can see in the editor.spec.ts TestEnvironment constructor that we have source texts, but they use the exact same mock TextData.

If it were not so complicated, I would ask you to include a test.
</blockquote></details>

Modified to use `VERSE_REGEX`.  Yes, good point about the test case. I do want to explore testing the editor, and I'll dig in if you think it is necessary now.  If not, I may need to move on at the moment.

PR #1871, opened 2023-05-26
SF-2046 Maintain the user's current segment after closing the note di…

RaymondLuong3:
…alog

When a user opens and closes the note dialog, the cursor sometimes jumps to the first segment in the editor. I couldn't figure out why the cursor was jumping, but my best guess is that the first selectable element was being selected when the editor regained focus after the editor is closed.
For this change, I updated the logic so that the current segment is recorded before the dialog opens. When the dialog closes, it will set the current segment to the segment that the users had selected before opening the note dialog.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1871)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -38,32 +38,10 @@ import { Segment } from './segment';
 import { EditorRange, TextViewModel } from './text-view-model';
 import { NoteDialogData, TextNoteDialogComponent } from './text-note-dialog/text-note-dialog.component';
 
-const EDITORS = new Set<Quill>();
 // When a user is active in the editor a timer starts to mark them as inactive for remote presences
 export const PRESENCE_EDITOR_ACTIVE_TIMEOUT = 3500;
 
-function onNativeSelectionChanged(): void {
-  // workaround for bug where Quill allows a selection inside of an embed
-  const sel = window.document.getSelection();
-  if (sel == null || sel.rangeCount === 0 || !sel.isCollapsed) {
-    return;
-  }
-  const text = sel.getRangeAt(0).commonAncestorContainer.textContent;
-  if (text === '\ufeff') {
-    for (const editor of EDITORS) {
-      if (editor.hasFocus()) {
-        const editorSel = editor.getSelection();
-        if (editorSel != null) {
-          editor.setSelection(editorSel, 'silent');
-        }
-        break;
-      }
-    }
-  }
-}
-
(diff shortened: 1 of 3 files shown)


pmachapman:
I am not sure that the right solution to the bug is by changing the editor component whenever the note dialog is triggered - I also triggered the bug by clicking the "share button", then changing the "Only the recipient can" dropdown to another value (which removed the focus from the editor).

I also noticed with this fix, when I was testing it locally, that the page jumped around a bit, and did not go back to exactly the same spot.

In my experimenting, I commented out the following lines in text.component.ts:

```cs
L76:  // window.document.addEventListener('selectionchange', onNativeSelectionChanged);
L835: // this.focused.emit(focus);
```

And the bug stopped occurring. Those those lines are fixing other bugs, but I think the issue is related to code around these two events?

My hunch is the bug is in `onNativeSelectionChanged`, but we have written a hook for the focused event that works around a bug in `onNativeSelectionChanged`, which also adjusts the where the cursor goes on focus.

pmachapman (commented):
Awesome - thank you!

Is it possible to have it record the last scroll bar position? I really like how the "jumping around" bug is fixed, but I still get a bit confused when I close the dialog, and window scrolls to a position where the verse which I had selected (which is often in the middle of the screen) is now at the bottom of the screen. I get a bit lost trying to find where I am when this happens, and wonder if our users will struggle with the same?

To recreate what I am talking about:
1. Scroll half way down the page
2. Select a verse in the middle of the page
3. Click the Add Note button
4. Close the dialog

If this isn't fixable, no worries - I will approve the PR, as the rest looks good and functions well.

> Reviewed 1 of 2 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1871#-:-NWnnGw6Edc7eB3GmZ71:bv43jby)* status: all files reviewed, 1 unresolved discussion

RaymondLuong3 (commented):
Yes, I was just getting to that. It appears that I can also record the scroll position and reset it to that same scroll position. I don't see a flicker, which is a good sign.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1871#-:-NWnqsV_3RgvdV5uvkL8:bkti9dy)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @pmachapman)

pmachapman (approved):
:lgtm: Awesome - it works gloriously!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1871#-:-NWns14r7JskT424PwCe:bgtmj8)* status: all files reviewed, 1 unresolved discussion

PR #1877, opened 2023-05-30
log: resource needs updating or not

marksvc:
commit-id:a0aacad7

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1877)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
@@ -372,6 +372,8 @@ public class ParatextSyncRunner : IParatextSyncRunner
             bool resourceNeedsUpdating =
                 paratextProject is ParatextResource paratextResource
                 && _paratextService.ResourceDocsNeedUpdating(_projectDoc.Data, paratextResource);
+            if (paratextProject is ParatextResource)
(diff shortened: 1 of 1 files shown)


Nateowami (commented):
Can you explain why this is helpful? If it's just generally useful to understand how often we're actually updating resources, it would be great to have it in the sync metrics as well.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1877#-:-NWiP0q0-BsSh46O5DKj:bb5eniu)* status: 0 of 1 files reviewed, all discussions resolved

marksvc:
I added this in because when I was paying attention to whether resources were being updated or not, I didn't see a very clear and definitive way to tell whether we thought they did or did not need updating. 

This additional log item will have only occasional benefit. I do think it may be beneficial after setting all resources as needing a sync, as part of SF-1444 (whitespace fix, migration). Particularly if the system does not behave as expected :).

> If it's just generally useful to understand how often we're actually updating resources, it would be great to have it in the sync metrics as well.

I wanted to know what was going on in specific syncs, rather than gain a statistical understanding. (But I can look into adding that in if you desire.)

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1877)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1879, opened 2023-05-31
Improve type safety of join component error messages

Nateowami:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1879)


Nateowami:
I'm going to close this since Nigel did a better job than I realized fixing the types (I think I may have missed that PR altogether and confused it with a related one). Let's get Nigel's fixes in, and then consider adding this later.

PR #1886, opened 2023-06-06
SF-1601 Errors from addTranslateMetrics and getWordGraph are now silent

josephmyers:


josephmyers:
I marked this as Testing Not Required, but _only_ because I don't know how to reproduce these two error cases. If someone has steps to reproduce either of these, let me know

RaymondLuong3 (changes requested):
Nice work. I think I agree, these errors don't appear to be consistently reproducible. I have ja few comments.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1886#-:-NYJQADg1mTy1YJGMx-f:bjismzv)* status: all files reviewed, 3 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/machine-api/remote-translation-engine.ts` line 80 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1886#-NYJPVR44Uv7OWxxK2pj:-NYJPVR44Uv7OWxxK2pk:blmf7xl) ([raw file](https://github.com/sillsdev/web-xforge/blob/e70a069b6c99165f9bb0e6e8644aada45421bf18/src/SIL.XForge.Scripture/ClientApp/src/app/machine-api/remote-translation-engine.ts#L80)):*
> ```XML
>         .toPromise();
>       return this.createWordGraph(response.data as WordGraphDto);
>     } catch (err) {
> ```

Should this show the error only of the user is not online? It's possible that the remote translation engine is created while the user is online, but when they go offline, we probably want to not annoy the user while they type. I might be better to not try not try the post request at all unless the user is online.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/machine-api/remote-translation-engine.spec.ts` line 434 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1886#-NYJNe4_7gYxstm68zy3:-NYJNe4_7gYxstm68zy4:b-prnn4h) ([raw file](https://github.com/sillsdev/web-xforge/blob/e70a069b6c99165f9bb0e6e8644aada45421bf18/src/SIL.XForge.Scripture/ClientApp/src/app/machine-api/remote-translation-engine.spec.ts#L434)):*
> ```XML
>     ).thenThrow(new Error());
> 
>     const result = await env.client.getWordGraph(sourceSegment);
> ```

Add a type to result.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/translate-metrics-session.ts` line 243 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1886#-NYJOBklDZ5Z1la9dhc5:-NYJOBklDZ5Z1la9dhc6:b3p3uct) ([raw file](https://github.com/sillsdev/web-xforge/blob/e70a069b6c99165f9bb0e6e8644aada45421bf18/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/translate-metrics-session.ts#L243)):*
> ```XML
>           ErrorReportingService.normalizeError(err)
>         );
>         return;
> ```

I don't think this return is necessary.

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1886)* status: 7 of 8 files reviewed, 3 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/machine-api/remote-translation-engine.ts` line 80 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1886#-NYJPVR44Uv7OWxxK2pj:-NYLrlVq9gpwrPNZVtuT:b-3rdq32) ([raw file](https://github.com/sillsdev/web-xforge/blob/e70a069b6c99165f9bb0e6e8644aada45421bf18/src/SIL.XForge.Scripture/ClientApp/src/app/machine-api/remote-translation-engine.ts#L80)):*

Should this show the error only of the user is not online? It's possible that the remote translation engine is created while the user is online, but when they go offline, we probably want to not annoy the user while they type. I might be better to not try not try the post request at all unless the user is online.
</blockquote></details>

Ok, so you're suggesting we only make this call if we're online? This was already being done, by my testing. It probably isn't clear from the code changes -- this method isn't called directly. It's called by InteractiveTranslatorFactory.create in editor.component -- the translateSegment() method. And the first thing this method does is return if offline. I just re-confirmed the post isn't being called offline, as expected.

Now, maybe what you're getting at is that you want this check to be in the call itself. We luck out that it's only called from the one place above. Obviously it'd be safer to make the check in getWordGraph() itself. I opted against this simply since the code above was already in place and working, so it felt like overkill to now pass the pwaService to the translation engine in order to make a duplicate check.

Hopefully I understood your comment correctly! Let me know if I'm off base -- no problem making any changes
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/machine-api/remote-translation-engine.spec.ts` line 434 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1886#-NYJNe4_7gYxstm68zy3:-NYLvhFzDUb4IMvZQVVR:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/e70a069b6c99165f9bb0e6e8644aada45421bf18/src/SIL.XForge.Scripture/ClientApp/src/app/machine-api/remote-translation-engine.spec.ts#L434)):*

Add a type to result.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/translate-metrics-session.ts` line 243 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1886#-NYJOBklDZ5Z1la9dhc5:-NYLvxvpD5kxC51wBBm6:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/e70a069b6c99165f9bb0e6e8644aada45421bf18/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/translate-metrics-session.ts#L243)):*

I don't think this return is necessary.
</blockquote></details>

Done.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1886)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/machine-api/remote-translation-engine.ts` line 80 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1886#-NYJPVR44Uv7OWxxK2pj:-NYOTG_5B7CZNltdtNrV:bwx82bg) ([raw file](https://github.com/sillsdev/web-xforge/blob/e70a069b6c99165f9bb0e6e8644aada45421bf18/src/SIL.XForge.Scripture/ClientApp/src/app/machine-api/remote-translation-engine.ts#L80)):*

Ok, so you're suggesting we only make this call if we're online? This was already being done, by my testing. It probably isn't clear from the code changes -- this method isn't called directly. It's called by InteractiveTranslatorFactory.create in editor.component -- the translateSegment() method. And the first thing this method does is return if offline. I just re-confirmed the post isn't being called offline, as expected.

Now, maybe what you're getting at is that you want this check to be in the call itself. We luck out that it's only called from the one place above. Obviously it'd be safer to make the check in getWordGraph() itself. I opted against this simply since the code above was already in place and working, so it felt like overkill to now pass the pwaService to the translation engine in order to make a duplicate check.

Hopefully I understood your comment correctly! Let me know if I'm off base -- no problem making any changes
</blockquote></details>

Excellent, I was thinking that it was probably the case that this was not being called while offline, but just wanted to check to make sure that was the case. Without any reliable way to reproduce this, I think this is the best we will get.

PR #1887, opened 2023-06-06
SF-2053 Backend Support for NMT Pre-Translation

pmachapman:
This PR adds:
 * A pre-translate setting the translate config
 * Pre-Translation corpus syncing and build
 * Retrieval of the Draft Pre-Translation
 * Get Pre-Translation Build endpoint
 * An additional build status of "Queued" for builds that are being uploaded to Serval

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1887)

### DIFF
--- a/src/RealtimeServer/scriptureforge/models/translate-config.ts
+++ b/src/RealtimeServer/scriptureforge/models/translate-config.ts
@@ -10,6 +10,7 @@ export interface TranslateConfig {
   source?: TranslateSource;
   shareEnabled: boolean;
   defaultNoteTagId?: number;
+  preTranslate: boolean;
 }
 
 export interface TranslateSource {
--- a/src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts
+++ b/src/RealtimeServer/scriptureforge/services/note-thread-service.spec.ts
@@ -424,7 +424,8 @@ class TestEnvironment {
       writingSystem: { tag: 'qaa' },
       translateConfig: {
         translationSuggestionsEnabled: false,
-        shareEnabled: true
+        shareEnabled: true,
+        preTranslate: false
       },
       checkingConfig: {
         checkingEnabled: false,
--- a/src/RealtimeServer/scriptureforge/services/question-service.spec.ts
+++ b/src/RealtimeServer/scriptureforge/services/question-service.spec.ts
@@ -155,7 +155,8 @@ class TestEnvironment {
       writingSystem: { tag: 'qaa' },
       translateConfig: {
         translationSuggestionsEnabled: false,
-        shareEnabled: true
+        shareEnabled: true,
+        preTranslate: false
       },
       checkingConfig: {
         checkingEnabled: false,
--- a/src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts
+++ b/src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts
@@ -203,6 +203,23 @@ describe('SFProjectMigrations', () => {
       expect(projectDoc.data.checkingConfig.shareLevel).not.toBeDefined();
     });
   });
+
+  describe('version 10', () => {
+    it('adds preTranslate to translate config', async () => {
+      const env = new TestEnvironment(9);
+      const conn = env.server.connect();
+      await createDoc(conn, SF_PROJECTS_COLLECTION, 'project01', {
+        translateConfig: { translationSuggestionsEnabled: false }
+      });
+      let projectDoc = await fetchDoc(conn, SF_PROJECTS_COLLECTION, 'project01');
+      expect(projectDoc.data.translateConfig.preTranslate).not.toBeDefined();
+
+      await env.server.migrateIfNecessary();
+
+      projectDoc = await fetchDoc(conn, SF_PROJECTS_COLLECTION, 'project01');
+      expect(projectDoc.data.translateConfig.preTranslate).toBe(false);
+    });
+  });
 });
 
 class TestEnvironment {
--- a/src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts
+++ b/src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts
@@ -207,6 +207,22 @@ class SFProjectMigration9 implements Migration {
   }
 }
 
+class SFProjectMigration10 implements Migration {
+  static readonly VERSION = 10;
+
+  async migrateDoc(doc: Doc): Promise<void> {
+    const ops: Op[] = [];
+    if (doc.data.translateConfig != null) {
+      ops.push({ p: ['translateConfig', 'preTranslate'], oi: false });
+    }
+    await submitMigrationOp(SFProjectMigration10.VERSION, doc, ops);
+  }
+
+  migrateOp(_op: RawOp): void {
+    // do nothing
+  }
+}
+
 export const SF_PROJECT_MIGRATIONS: MigrationConstructor[] = [
   SFProjectMigration1,
   SFProjectMigration2,
@@ -216,5 +232,6 @@ export const SF_PROJECT_MIGRATIONS: MigrationConstructor[] = [
   SFProjectMigration6,
   SFProjectMigration7,
   SFProjectMigration8,
-  SFProjectMigration9
+  SFProjectMigration9,
+  SFProjectMigration10
 ];
--- a/src/RealtimeServer/scriptureforge/services/sf-project-service.spec.ts
+++ b/src/RealtimeServer/scriptureforge/services/sf-project-service.spec.ts
@@ -106,7 +106,8 @@ class TestEnvironment {
       },
       translateConfig: {
         translationSuggestionsEnabled: false,
-        shareEnabled: false
+        shareEnabled: false,
+        preTranslate: false
       },
       writingSystem: {
         tag: 'en'
--- a/src/RealtimeServer/scriptureforge/services/text-audio-service.spec.ts
+++ b/src/RealtimeServer/scriptureforge/services/text-audio-service.spec.ts
@@ -127,7 +127,8 @@ class TestEnvironment {
       writingSystem: { tag: 'qaa' },
       translateConfig: {
         translationSuggestionsEnabled: false,
-        shareEnabled: true
+        shareEnabled: true,
+        preTranslate: false
       },
       checkingConfig: {
         checkingEnabled: false,
--- a/src/RealtimeServer/scriptureforge/services/text-service.spec.ts
+++ b/src/RealtimeServer/scriptureforge/services/text-service.spec.ts
@@ -116,7 +116,8 @@ class TestEnvironment {
       writingSystem: { tag: 'qaa' },
       translateConfig: {
         translationSuggestionsEnabled: false,
-        shareEnabled: true
+        shareEnabled: true,
+        preTranslate: false
       },
       checkingConfig: {
         checkingEnabled: false,
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -964,7 +964,8 @@ class TestEnvironment {
         },
         translateConfig: {
           translationSuggestionsEnabled: false,
-          shareEnabled: false
+          shareEnabled: false,
+          preTranslate: false
         },
         checkingConfig: {
           checkingEnabled: true,
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts
@@ -591,7 +591,8 @@ class TestEnvironment {
     },
     translateConfig: {
       translationSuggestionsEnabled: false,
-      shareEnabled: false
+      shareEnabled: false,
+      preTranslate: false
     },
     sync: { queuedCount: 0 },
     editable: true,
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.stories.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.stories.ts
@@ -41,7 +41,8 @@ const defaultArgs = {
     },
     translateConfig: {
       translationSuggestionsEnabled: false,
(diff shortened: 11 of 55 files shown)


RaymondLuong3 (changes requested):
That is a lot of work. Good job. I know that QA.serval is live. Is there a way to test pre-translations, or is that not yet an option? Is there a reason that a build would have faulted if it has been more than 6 hours since it was queued? I wonder if we should have a hook that automatically cancels the build if more than 6 hours has passed since the build was queued.

> Reviewed 22 of 45 files at r1, 8 of 20 files at r2, 1 of 7 files at r3, 9 of 15 files at r4, 12 of 15 files at r5, 4 of 4 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-:-NZ32ZoP5PXOJriIrLoI:b-tm6dt4)* status: all files reviewed, 10 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Models/ServalData.cs` line 41 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2pUZt8wXTtgTZk5uW:-NZ2pUZt8wXTtgTZk5uX:b-3j5d9c) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Models/ServalData.cs#L41)):*
> ```Smalltalk
>     /// If this is longer than 6 hours ago (UTC), there will have been a crash, so an error should be reported.
>     /// </remarks>
>     public DateTime? PreTranslationQueued { get; set; }
> ```

What do you say about PreTranslationQueuedAt, then it would make sense that a DateTime is returned
___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 321 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2qoXt5wuDer7zZ6fN:-NZ2qoXt5wuDer7zZ6fO:b66prkj) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L321)):*
> ```Smalltalk
>         await EnsurePermissionAsync(curUserId, sfProjectId);
> 
>         // We only support Serval for canceling the current build
> ```

Looks like copy and paste got you here!
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 439 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2t6Xk4Ux15y2fSzap:-NZ2t6Xk4Ux15y2fSzaq:bxgpxbe) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L439)):*
> ```Smalltalk
>         string? corpusId = projectSecret.ServalData.Corpora
>             .FirstOrDefault(c => c.Value.PreTranslate == preTranslate)
>             .Key;
> ```

This would throw an error if no Corpora in ServalData matched the preTranslate value. You could use optional chaining.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 689 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2uOvU9BbtRfqnx6FC:-NZ2uOvU9BbtRfqnx6FD:b-v790np) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L689)):*
> ```Smalltalk
>     /// <param name="translationEngineId">The translation engine id</param>
>     /// <returns><c>true</c> if the pre-translation build is running; otherwise, <c>false</c>.</returns>
>     private async Task<bool> PreTranslationBuildQueuedAsync(string translationEngineId)
> ```

IsPreTranslationBuildQueuedAsync would be a better name for this.
___
*[`src/SIL.XForge.Scripture/Services/PreTranslationService.cs` line 50 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2xR2Z1hPybq9mYGHz:-NZ2xR2Z1hPybq9mYGI-:bwwvl0p) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Services/PreTranslationService.cs#L50)):*
> ```Smalltalk
>         // Ensure we have the parameters to retrieve the pre-translation
>         string translationEngineId = projectSecret.ServalData?.PreTranslationEngineId;
>         string corpusId = projectSecret.ServalData?.Corpora.FirstOrDefault(c => c.Value.PreTranslate).Key;
> ```

Do we ever expect the Corpora to not contain a pre-translate engine? Maybe optional chaining would be best.
___
*[`src/SIL.XForge.Scripture/Services/PreTranslationService.cs` line 123 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2zGon0u3skOcg_wwN:-NZ2zGon0u3skOcg_wwO:bvw8hu9) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Services/PreTranslationService.cs#L123)):*
> ```Smalltalk
> 
>                 // Add a preceding space if this is not the first word or punctuation
>                 if (i > 0 && (word.Length != 1 || !char.IsPunctuation(word.First())))
> ```

What would happen in a language like Thai where there are now spaces between words?

pmachapman (commented):
Yes, you can test pre-translations via swagger until the UI is completed.

A fault in the 6 hour window will happen before the build is started (i.e. during file upload), so there will be no build to cancel.

> Dismissed @github-code-scanning[bot] from 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-:-NZ4Cabs7cotaCthlwRu:b-m43md9)* status: all files reviewed, 6 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/Models/ServalData.cs` line 41 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2pUZt8wXTtgTZk5uW:-NZ3yAxC4RYp-3El-McM:bf8quu8) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Models/ServalData.cs#L41)):*

What do you say about PreTranslationQueuedAt, then it would make sense that a DateTime is returned
</blockquote></details>

Sounds good. Changed.
___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 321 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2qoXt5wuDer7zZ6fN:-NZ3yPTu43UwP9glm7em:b-sxsadz) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L321)):*

Looks like copy and paste got you here!
</blockquote></details>

Thank you for finding the copy-pasta! Fixed.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 439 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2t6Xk4Ux15y2fSzap:-NZ3zaeN3psmOC3EL5hH:bhjgwyi) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L439)):*

This would throw an error if no Corpora in ServalData matched the preTranslate value. You could use optional chaining.
</blockquote></details>

It won't as a KeyValuePair is a struct, so its default value is an empty struct, not null, so in this case `new KeyValuePair<string, string>(null, null)`;
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 689 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2uOvU9BbtRfqnx6FC:-NZ4BReT5QME4BKZUKUz:b-nn3giw) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L689)):*

IsPreTranslationBuildQueuedAsync would be a better name for this.
</blockquote></details>

Good idea - changed.
___
*[`src/SIL.XForge.Scripture/Services/PreTranslationService.cs` line 50 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2xR2Z1hPybq9mYGHz:-NZ4BUpv6enHaNZOoxSL:ba0g4u6) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Services/PreTranslationService.cs#L50)):*

Do we ever expect the Corpora to not contain a pre-translate engine? Maybe optional chaining would be best.
</blockquote></details>

See reason given above.
___
*[`src/SIL.XForge.Scripture/Services/PreTranslationService.cs` line 123 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2zGon0u3skOcg_wwN:-NZ4BqYa3TIyUUDAt4Kp:bwuiz47) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Services/PreTranslationService.cs#L123)):*

What would happen in a language like Thai where there are now spaces between words?
</blockquote></details>

I am not sure - in my tests, the text we get back from the Serval pre-translation looks like:
```
"Abraham begot Isaac . And Isaac begot Jacob . And Jacob begot Judas and his brethren ."
```

What this code is trying to do is make the punctuation appear more normal.

RaymondLuong3 (changes requested):
Just one last minor change, then it will be good to push to QA. When we do that, are there any consequences we should think about? I assume that the pre-translations engine would not get built because the front end has no way to connect to those endpoints yet. Is that correct?

> Reviewed 15 of 15 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-:-NZXMzeKDCtVY1w8dJ85:bgxazbc)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 439 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2t6Xk4Ux15y2fSzap:-NZXKZO26H2_a1T8xLPu:bpqhk46) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L439)):*

It won't as a KeyValuePair is a struct, so its default value is an empty struct, not null, so in this case `new KeyValuePair<string, string>("", "")`;
</blockquote></details>

That's great. So the Key would be an empty string. So then does the corpusId need to be nullable?
___
*[`src/SIL.XForge.Scripture/Services/PreTranslationService.cs` line 123 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2zGon0u3skOcg_wwN:-NZXMC7f0Iptwr4J5rFh:bbvb42v) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Services/PreTranslationService.cs#L123)):*

I am not sure - in my tests, the text we get back from the Serval pre-translation looks like:
```
"Abraham begot Isaac . And Isaac begot Jacob . And Jacob begot Judas and his brethren ."
```

What this code is trying to do is make the punctuation appear more normal.
</blockquote></details>

Ok, this might be one of those instances where we might have to wait for real world trials to confirm if it is an issue. In Chinese, the punctuation does not need a space after it. But I'm not sure if there is a way to know that.

pmachapman (commented):
Yes, that is correct. I am not aware of any changes to current behavior with this PR, as it only contains additional functionality that is not yet called (as Serval SMT suggestions have not yet been enabled on QA or live).

A smoke test on QA will be sufficient, due to the migration for `TranslateConfig`.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-:-NZXoxjA8YDU3vgsG8Nf:b-lj1nc8)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 439 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2t6Xk4Ux15y2fSzap:-NZXvJLcEbt8dSIFKD59:bl8lja6) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L439)):*

That's great. So the Key would be an empty string. So then does the corpusId need to be nullable?
</blockquote></details>

Good point. Fixed.
___
*[`src/SIL.XForge.Scripture/Services/PreTranslationService.cs` line 123 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/1887#-NZ2zGon0u3skOcg_wwN:-NZXvmmo1nBJYqs44mf2:baszr0l) ([raw file](https://github.com/sillsdev/web-xforge/blob/b72cc51bfc83e0cd6c3b5b5abe0092c21af40793/src/SIL.XForge.Scripture/Services/PreTranslationService.cs#L123)):*

Ok, this might be one of those instances where we might have to wait for real world trials to confirm if it is an issue. In Chinese, the punctuation does not need a space after it. But I'm not sure if there is a way to know that.
</blockquote></details>

Yes, we won't know until some real world data comes in. If we have access to any existing back translations, I can setup test corpora and run using real world data in QA or on my machine. I don't have a grasp of any languages that use non-European punctuation, so am not able to create test data myself and know it is accurate.

RaymondLuong3 (approved):
> Reviewed 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1887)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #1888, opened 2023-06-07
SF-1814 User Invites Settings

josephmyers:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1888)


josephmyers:
image

I touched up the text from Nathaniel's and Nigel's discussion last year. This option matches well the existing pattern of listing all relevant details on the page.

josephmyers:
image

Here's another option. It wraps up the detailed information into a tooltip. This introduces a new pattern to the page, so if we like this, we can wrap the other "hints" up, as well, as much as needed. This option reduces page clutter, without preventing access to the full details, which the admin's will only need the first few times.

The only thing I haven't tested is how tooltip's in practice work on mobile devices. Per [the docs](https://material.angular.io/components/tooltip/overview#showing-and-hiding), a longpress activates the tooltip for touch devices. We'd need to test this for usability, for sure.

nigel-wells (changes requested):
Thanks for working through this. I'm ok with the idea of using tooltips to reduce clutter especially around information that is only needed the first few times. I think you may have cut the initial text down too much that the text against each toggle isn't clear unless you use the tooltip. Perhaps adding the word, "Allow" at the front would help? 

In saying that, we'll end up duplicating most of the text in each tooltip so perhaps it should just remain in the initial paragraph? You're revised wording in the first screenshot is only across 2 lines which makes it easy to read.

Do you have any thoughts on the, "Sharing" heading? We used to call it sharing - I think the share dialog also says, "Share" along with a "Share" button. Maybe we change it to, "User invites", or "Allow user invites". The label on the toggles could be, "Enable for Community Checking" or something affirmative like that.

> Reviewed 2 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-:-NXgZkW6AXeMtrlONozR:bjd8ydk)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @josephmyers and @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss` line 47 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-NXgbcSzBIM0-1SMB3In:-NXgbcSzBIM0-1SMB3Io:biemzgv) ([raw file](https://github.com/sillsdev/web-xforge/blob/7d7f2dd28dcdc8a9c148734d28f9fc25d3a88b88/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss#L47)):*
> ```SCSS
>     > app-write-status {
>       @include media-breakpoint-up(sm) {
>         margin-left: 22px;
> ```

How does this look on RTL? Might need to change to `margin-inline-start`.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss` line 52 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-NXgboCa0xxqa5R4AgBa:-NXgboCa0xxqa5R4AgBb:b-xincuh) ([raw file](https://github.com/sillsdev/web-xforge/blob/7d7f2dd28dcdc8a9c148734d28f9fc25d3a88b88/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss#L52)):*
> ```SCSS
> 
>     .mat-icon {
>       transform: scale(0.9);
> ```

In keeping with the Material guidelines I'd either level this at the default 24px or reduce down specifically to 18px which they also say is ok. This will ensure it always displays at the icon at its optimal rendering. I would caution though as we'd need to remember to do this across all tooltips in order to be consistent. We could look to have a tooltip component that we can then ensure the styling is always consistant.

josephmyers (commented):
I am good changing "Sharing" to "User Invites," under the assumption that we are intending to separate the two. Leaving the heading as "Sharing" is okay in my mind, since we have the hint text that links to the admin inviting, and I like the grouping. Further, and maybe more importantly, "Sharing" is a familiar term, where "User Invites" or similar will put some cognitive load on the reader, until they actually understand what the term means. It's subtle, and I only want to be picky when it's important. So, one is very precise, and the other is recognizable.

The full details option is 3-4 lines on mobile, which I didn't really like. I think it's a lot of information to digest, at minimum to scroll past.

Of course you're aware that I removed the "Allow/Enable" word from the toggle because I thought it was redundant. I do think the user will understand that a checked checkbox equates to "Allow/Enable," but this point really isn't a big deal to me. Plus, your "Enable for..." suggestion looks solid to me. If we change the heading to "User Invites" we don't need to repeat it for the toggles. Of course, if we ever have any other "Sharing"-ish settings, they will get their own section.

Hopefully my changes capture your intent. Let me know if I missed something!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-:-NXi0tdWCVi5qBziH5D5:bh1htlk)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss` line 47 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-NXgbcSzBIM0-1SMB3In:-NXi7dQBBGYPojMXJMpT:b-mueuls) ([raw file](https://github.com/sillsdev/web-xforge/blob/7d7f2dd28dcdc8a9c148734d28f9fc25d3a88b88/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss#L47)):*

How does this look on RTL? Might need to change to `margin-inline-start`.
</blockquote></details>

Great catch. May've been broken before
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss` line 52 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-NXgboCa0xxqa5R4AgBa:-NXi8IvJDZcUZaM8eNBs:b-sfxaow) ([raw file](https://github.com/sillsdev/web-xforge/blob/7d7f2dd28dcdc8a9c148734d28f9fc25d3a88b88/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss#L52)):*

In keeping with the Material guidelines I'd either level this at the default 24px or reduce down specifically to 18px which they also say is ok. This will ensure it always displays at the icon at its optimal rendering. I would caution though as we'd need to remember to do this across all tooltips in order to be consistent. We could look to have a tooltip component that we can then ensure the styling is always consistant.
</blockquote></details>

Sounds good. I thought it looked goofy at default, but I'll remove it to be consistent. If you don't like the way it looks at default, I can change it back.

nigel-wells (commented):
I think we're getting close to the right result. I do find it odd that both tooltips say exactly the same thing leaving me to wonder what the difference is (from an admins POV) as to what is happening - perhaps I've been looking at it all for too long!

Trying to look at this with fresh administrator eyes, I wonder the following when I see it:
* I can invite people outside of Paratext via the users page - great!
* There are 2 checkboxes to enable something - do I need to enable these before I can invite people? 
* If I enable those options does that mean other users now get access to the same "Users" page I get access to?

Also, should the tooltip be "Allow users..." rather than, "Allow**s** users..." with the extra "s" on allows?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-:-NXn1WSI31SPM4zQePaz:b-pqlj5s)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

josephmyers (commented):
Yeah, I agree that the new wording is a little confusing in terms of what the checkboxes do. "User Invites" with fresh eyes could mean "users inviting" or "invite users" -- it's not clear at first glance, and it moreso now depends on its explanation. Even the previous wording, "Allow user invites...", kind of worked, despite the duplication, but distancing the term from the tooltip increases confusion, I believe. It doesn't help that the hint for a different feature is in between them. So I reworded the checkboxes; they now indirectly explain the heading. I also moved the hint text below and reworded it to distinguish it from what's above.

The extra "s" was intentional. The tooltip describes the checkbox: "[This] allows..."; however, I have no problem changing it. Also, I looked into moving to a single tooltip by the heading, but to me, the further abstraction is more confusing.

I feel like the changes I made address the issues you point out while still providing an overall clear experience for first-time readers.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-:-NXnFalG3MBOi2jkK3B7:b8b2qye)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @Nateowami and @nigel-wells)

josephmyers:
Here's an up-to-date screenshot:

image

nigel-wells (approved):
I think that is looking pretty good - I like the admin comment moved under the checkboxes. Perhaps the only additional tweak I would make is to say "Administrations can `always` invite..." just to make it clear that the settings have no baring on their ability to invite.
The tooltip icons show funny on my screen with the top/bottom cut off:

This will be related to the font size associated with it reducing it from the Material optimal size. It may just be a thing with my screen so just pointing it out as something we need to keep in mind.

A storybook would be good for the new info component when you get a chance.

I'm happy to approve this so as not to hold it up any more - you can decide if you want to make the tweak mentioned as my English gramma was barely passable at high school lol

> Reviewed 7 of 8 files at r5, 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-:-NYGiQxt5rmhdWyeBC6c:b-3lad8c)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

josephmyers (commented):
Cool, thanks. I can make the "always" change -- I agree that it adds good value!

What are your screen specs, etc.? That would help me repro the cutoff issue and know when I've fixed it. I believe you said at one point that either 24px or 18px were acceptable. The current styling (from the font size) has them at 18.2px on my screen; perhaps the 0.2 is causing them to act strange? I also wonder if the pre-existing icon (in the checking audio component) was behaving the same for you. I figured it would be safe, since the precedent had been set. I still think the default 24px looks goofy, but I have no problem just putting a hardcode 18px in there, instead of font size, as before -- IF I can verify it fixes the issue you're seeing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-:-NYH7bzX6-R_9A6C9pTA:b-xcod0p)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

josephmyers (commented):
Now, I do see cutoff when switching to tablet devices, e.g. iPad Air and Surface Pro 7. However, this appears due simply to the zoom level, which for those big guys is ~50% in my browser. Using 100% zoom resolves it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-:-NYHD9D51CDOhboErTcD:bcym24f)* status: 8 of 11 files reviewed, all discussions resolved (waiting on @Nateowami and @nigel-wells)

nigel-wells (changes requested):
My screen is at 2560x1440 and 100% scaling. Well done finding a fix for that. I wonder how that works on Mac devices - I imagine it is fine as mac and iOS devices typically have high density screens that everything renders really well.

Just a couple of comments on the storybook config.

> Reviewed 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-:-NYKPJP_1O_mZ7650_Kt:b4ukl96)* status: all files reviewed, 4 unresolved discussions (waiting on @josephmyers and @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss` line 59 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-NYKNLfYCntN88OpJ4sY:-NYKNLfYCntN88OpJ4sZ:b-9jygvv) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ba2784076283ffc45721fcc007a2ebd910b2da9/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss#L59)):*
> ```SCSS
> mat-icon {
>   color: $greyLight;
>   overflow: visible;
> ```

Was this still needed after the rotation hack? There is no visible difference on my screen with this change. I won't block it in case it was addressing something else.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/info/info.component.scss` line 16 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-NYKN2EP1E8WlKeEpMGH:-NYKN2EP1E8WlKeEpMGI:b-6laehp) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ba2784076283ffc45721fcc007a2ebd910b2da9/src/SIL.XForge.Scripture/ClientApp/src/app/shared/info/info.component.scss#L16)):*
> ```SCSS
>     font-size: 18px;
>     // https://github.com/google/material-design-icons/issues/648
>     transform: rotate(0.03deg);
> ```

What a crazy hack! It totally worked though :)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/info/info.stories.ts` line 17 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-NYKOPcm-hc12pLcCRCm:-NYKOPcm-hc12pLcCRCn:b-5f4y6) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ba2784076283ffc45721fcc007a2ebd910b2da9/src/SIL.XForge.Scripture/ClientApp/src/app/shared/info/info.stories.ts#L17)):*
> ```XML
> export default meta;
> 
> const Template: StoryFn = args => ({
> ```

Switch over to Component Story Format 3 (CSF 3) - see `join.stories.ts as an example`. Essentially you're changing from `StoryFn` to `StoryObj`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/info/info.stories.ts` line 18 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-NYKOLuJ1y7tD4KFQQZ-:-NYKOLuK7ciTZ1zWvmCm:b1kv7v5) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ba2784076283ffc45721fcc007a2ebd910b2da9/src/SIL.XForge.Scripture/ClientApp/src/app/shared/info/info.stories.ts#L18)):*
> ```XML
> 
> const Template: StoryFn = args => ({
>   moduleMetadata: { imports: [UICommonModule, CommonModule] },
> ```

Move into `Meta`

josephmyers (commented):
Awesome, thanks

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-:-NYM-5HZEmH-lvoZIUI8:b4qy5ln)* status: 11 of 12 files reviewed, 3 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss` line 59 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-NYKNLfYCntN88OpJ4sY:-NYLyjAi6gIxtABZyTEe:b-xmwkhv) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ba2784076283ffc45721fcc007a2ebd910b2da9/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss#L59)):*

Was this still needed after the rotation hack? There is no visible difference on my screen with this change. I won't block it in case it was addressing something else.
</blockquote></details>

Yeah, the admin icon gets cut off horizontally at narrow dimensions
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/info/info.stories.ts` line 17 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-NYKOPcm-hc12pLcCRCm:-NYLzuva6Kobh2Jtn16i:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ba2784076283ffc45721fcc007a2ebd910b2da9/src/SIL.XForge.Scripture/ClientApp/src/app/shared/info/info.stories.ts#L17)):*

Switch over to Component Story Format 3 (CSF 3) - see `join.stories.ts as an example`. Essentially you're changing from `StoryFn` to `StoryObj`
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/info/info.stories.ts` line 18 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-NYKOLuJ1y7tD4KFQQZ-:-NYLzu_J3Hpswlfoc1jH:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/7ba2784076283ffc45721fcc007a2ebd910b2da9/src/SIL.XForge.Scripture/ClientApp/src/app/shared/info/info.stories.ts#L18)):*

Move into `Meta`
</blockquote></details>

Done.

nigel-wells (approved):
:lgtm:

We made it :)

Looks like it has already passed through tests? You can merge it if you're happy.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1888#-:-NYM9ilQ7DgWlMFBoU5S:b-9zmcob)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1892, opened 2023-06-12
SF-2066 Keep insert note fab hidden for commenters on smaller devices

RaymondLuong3:
Users with commenting permission will see the add note FAB only on larger devices. On smaller devices, commenters will see the bottom sheet appear when selecting a verse. This change ensures that the add note FAB is never visible for commenters on smaller devices.

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -2878,6 +2878,34 @@ describe('EditorComponent', () => {
       env.dispose();
     }));
 
+    it('keeps insert note fab hidden for commenters on mobile devices', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig();
+      when(mockedMediaObserver.isActive(anything())).thenReturn(true);
+      env.addParatextNoteThread(
+        6,
+        'MAT 1:1',
+        '',
+        { start: 0, length: 0 },
+        ['user01'],
+        NoteStatus.Todo,
+        undefined,
+        true
+      );
+      env.setCommenterUser();
+      env.wait();
+
+      env.clickSegmentRef('verse_1_3');
+      expect(window.getComputedStyle(env.insertNoteFab.nativeElement)['visibility']).toBe('hidden');
+      const noteElem: HTMLElement = env.getNoteThreadIconElement('verse_1_1', 'thread06')!;
+      noteElem.click();
+      env.wait();
+      env.mockNoteDialogRef.close();
+      env.wait();
+      expect(window.getComputedStyle(env.insertNoteFab.nativeElement)['visibility']).toBe('hidden');
+      env.dispose();
+    }));
+
     it('shows the correct combined verse ref for a new note', fakeAsync(() => {
       const env = new TestEnvironment();
       env.setProjectUserConfig();
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
@@ -438,16 +438,12 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
     return this.pwaService.isOnline && this.multiCursorViewers.length > 0;
   }
 
-  get isInsertNoteFabEnabled(): boolean {
-    return this.isAddNotesEnabled && this.canShowInsertNoteFab;
-  }
-
   set showInsertNoteFab(value: boolean) {
     if (this.insertNoteFab == null || this.TemplateBottomSheet == null) return;
     this.addingMobileNote = false;
     // Mobile users without editing rights will see a bottom sheet instead of a FAB
-    if (this.mediaObserver.isActive('lt-lg') && !this.hasEditRight) {
-      this.insertNoteFab.nativeElement.style.visibility = 'hidden';
+    if (this.isCommenterOnMobileDevice) {
+      this.setNoteFabVisibility('hidden');
       if (value) {
         if (this.bottomSheetRef?.containerInstance == null) {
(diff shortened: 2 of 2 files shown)


Nateowami (changes requested):
It looks like this is consistently failing with `Test did not clean up its overlay container content`

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1892#-:-NXpRt0pC8GvYeukbVYI:b-3shqny)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 441 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1892#-NXpSCRf3T6uQiuRGsAe:-NXpSCRf3T6uQiuRGsAf:b-hgz7f2) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e6526f89345177c27c2839e7bfd30821688b34f/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L441)):*
> ```XML
>   }
> 
>   set showInsertNoteFab(value: boolean) {
> ```

I think the name of this setter is causing confusion. It doesn't make much sense that `showInsertNoteFab = true` would call `this.setNoteFabVisibility('hidden')`. There's also a getter called `isInsertNoteFabEnabled`.

FAB stands for "floating action button". In my opinion it should only be used for floating buttons, but however we do it we should be consistent. Here this setter enables a floating action button, or a bottomsheet. It could be called something like `showAddCommentButton`.

RaymondLuong3 (commented):
Yes, I figured out how to clean it up. Thanks

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1892#-:-NXq2aSH5htNzGphW4WF:bslwoda)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 441 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1892#-NXpSCRf3T6uQiuRGsAe:-NXq2eDH-0QgZyIN15-R:b-vfqc5d) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e6526f89345177c27c2839e7bfd30821688b34f/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L441)):*

I think the name of this setter is causing confusion. It doesn't make much sense that `showInsertNoteFab = true` would call `this.setNoteFabVisibility('hidden')`. There's also a getter called `isInsertNoteFabEnabled`.

FAB stands for "floating action button". In my opinion it should only be used for floating buttons, but however we do it we should be consistent. Here this setter enables a floating action button, or a bottomsheet. It could be called something like `showAddCommentButton`.
</blockquote></details>

That is a good point. This method really is more about showing the button in any form, whether that is the FAB or the bottom sheet. I've renamed it to be `showAddCommentButton`. I added a description to the setting to help clarify the confusion.

Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1892)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1097 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1892#-NXqG8cJ1x7LMwDg9U2M:-NXqG8cJ1x7LMwDg9U2N:b-vkkqpl) ([raw file](https://github.com/sillsdev/web-xforge/blob/048e4bef576bf9d02b6ca4e6d4a76fb1abe1fe29/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1097)):*
> ```XML
>         return;
>       }
>       this.insertNoteFab.nativeElement.style.visibility = 'hidden';
> ```

This setter no longer does with it says it does. If you call `setNoteFabVisibility('visible')` it may decide to hide it instead. If it were named `updateNoteFabVisibility` and didn't take an argument, then I would expect that behavior. If some piece of code is calling `updateNoteFabVisibility('visible')` when `this.isInsertNoteFabEnabled` is false, then that code should be fixed.

Or this could be renamed to something else.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1892)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1097 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1892#-NXqG8cJ1x7LMwDg9U2M:-NXqSAHj0yyy39wQpoby:bydemkf) ([raw file](https://github.com/sillsdev/web-xforge/blob/048e4bef576bf9d02b6ca4e6d4a76fb1abe1fe29/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1097)):*

This setter no longer does with it says it does. If you call `setNoteFabVisibility('visible')` it may decide to hide it instead. If it were named `updateNoteFabVisibility` and didn't take an argument, then I would expect that behavior. If some piece of code is calling `updateNoteFabVisibility('visible')` when `this.isInsertNoteFabEnabled` is false, then that code should be fixed.

Or this could be renamed to something else.
</blockquote></details>

I couldn't decide whether I wanted to rename the method to something like `updateNoteFabVisibilityIfAllowed`, but in the end only one place needed a check which is when the note dialog is closed. So, I decided to do a check if the note fab is enabled before setting it to visible there.

Nateowami (approved):
:lgtm:

> Reviewed 1 of 1 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1892#-:-NXqTPHJ3GngIu8dMmmA:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1896, opened 2023-06-13
Update I18nStoryModule to work with I18nService

Nateowami:
Previously we only set the locale via `TranslocoService`, so `I18nService` was unaware of what locale was set.

Now we set it via `I18nService`, so everything works about as it does in the application.

This actually simplifies things signficantly.

And now we can write stories that rely on `I18nService` and they'll work correctly. I've added an example for `BookChapterChooserComponent` that previously wouldn't have worked correctly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1896)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.stories.ts
@@ -92,3 +92,7 @@ SelectChapter.play = async ({ canvasElement }) => {
   const chapter2 = await within(menu).findByText('2');
   userEvent.click(chapter2);
 };
+
+export const Arabic = Template.bind({});
+Arabic.args = { ...defaultArgs };
+Arabic.parameters = { locale: 'ar' };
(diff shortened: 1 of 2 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1896)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts` line 32 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1896#-NXtL1SyCu4ia1fRiqAc:-NXuG2rmCBWfQ_vFmBzN:b-k2rwtz) ([raw file](https://github.com/sillsdev/web-xforge/blob/30bda0182ead6e4add5cf7792e66a6957c97ccab/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n-story.module.ts#L32)):*

Is ApplicationRef unused now?
</blockquote></details>

Good catch. Fixed.

josephmyers (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1896)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1901, opened 2023-06-14
SF-2065 Update the selected verse when user opens a note dialog

RaymondLuong3:
When the editor is editable, there is different logic that determines when a user has selected a verse. This is because users can use the keyboard to jump to the next verse, so we cannot rely on click events. This was causing problems because the selection would change when a user clicks on a note icon to open the dialog, but we introduced logic to prevent the current commenter selection from being updated when the note dialog is open.
This change removes that faulty logic so the commenter selection is always up to date. This also adds extra logic to position the note FAB at text load and when the note dialog closes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1901)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -2830,7 +2830,7 @@ describe('EditorComponent', () => {
       env.dispose();
     }));
 
-    it('verse keeps selection when opening a note dialog', fakeAsync(() => {
+    it('updates verse selection when opening a note dialog', fakeAsync(() => {
       const env = new TestEnvironment();
       env.setProjectUserConfig();
       env.wait();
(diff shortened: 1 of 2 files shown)


josephmyers (changes requested):
Preliminary testing shows a lot of good progress; however, I am seeing the verse number highlight disappearing when adding a comment to a note. I'm assuming this is not intended! Looks like good changes overall. This class is really scary, and I wish we could get in here and reduce some of the badness.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1901#-:-NXxDHf9Dc6lQQFO6fln:b-80zsl9)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
Yes, I see what you mean. The issue was that the segment momentarily gets reset when the dialog closes after adding a note, which also un-selects the current selected verse. We have logic to reset the selection to the active verse, but it didn't trigger update to re-select the current verse, which I have fixed now.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1901#-:-NY-0VX6AdtlSpG7S2kN:b-4wdex)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1901)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1902, opened 2023-06-14
dev: trust mongodb repo

marksvc:
The mongo 4.0 key[1] seems to have expired without an updated key being available. So set `trusted=yes` to trust the mongo 4.0 repository anyway. This is not the greatest solution for security reasons.

The ansible page for apt_key shows[2] how to move away from deprecated apt-key. This patch also changes to use get_url instead of apt_key, following this example. This was also applied to the dotnet key. This change may fix something before it becomes a problem in a future version of Ubuntu.

[1] https://www.mongodb.com/docs/v4.0/tutorial/install-mongodb-on-ubuntu/ 
[2] https://docs.ansible.com/ansible/latest/collections/ansible/builtin/apt_key_module.html

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1902)

### DIFF
--- a/deploy/dependencies.yml
+++ b/deploy/dependencies.yml
@@ -15,23 +15,28 @@
     repo_path: "{{playbook_dir}}/.."
   pre_tasks:
     - name: "Add .NET Core | add key"
-      apt_key:
-        id: BE1229CF
+      get_url:
+        # key id BC528686B50D79E339D3721CEB3E94ADBE1229CF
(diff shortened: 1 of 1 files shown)


marksvc:
Marked as draft. Probably going to go with another solution.

Nateowami:
Closing. You can re-open if you change your mind.

PR #1905, opened 2023-06-15
SF-2061 - Address race condition with question query changing after route params change

nigel-wells:
- Subscribe to questions query onready to ensure the book and text is set
- Unsubscribe from various queries immediately if the route params have changed

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
@@ -31,7 +31,7 @@ import { VerseRef } from 'realtime-server/lib/esm/scriptureforge/scripture-utils
 import * as RichText from 'rich-text';
 import { BehaviorSubject, of, Subject } from 'rxjs';
 import { first } from 'rxjs/operators';
-import { anyString, anything, deepEqual, instance, mock, reset, resetCalls, spy, verify, when } from 'ts-mockito';
+import { anyString, anything, instance, mock, reset, resetCalls, spy, verify, when } from 'ts-mockito';
 import { AuthService } from 'xforge-common/auth.service';
 import { AvatarTestingModule } from 'xforge-common/avatar/avatar-testing.module';
 import { BugsnagService } from 'xforge-common/bugsnag.service';
@@ -50,6 +50,7 @@ import { configureTestingModule, getAudioBlob, TestTranslocoModule } from 'xforg
 import { UICommonModule } from 'xforge-common/ui-common.module';
 import { UserService } from 'xforge-common/user.service';
 import { objectId } from 'xforge-common/utils';
+import { QueryParameters } from 'xforge-common/query-parameters';
 import { QuestionDoc } from '../../core/models/question-doc';
 import { SFProjectDoc } from '../../core/models/sf-project-doc';
 import { SFProjectUserConfigDoc } from '../../core/models/sf-project-user-config-doc';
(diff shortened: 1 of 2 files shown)


nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1905)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 433 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1905#-NXzq7_Q1kYvdPxCFBp4:-NY0VE-uEF1Pm4U1Dt7n:b-h4pu8r) ([raw file](https://github.com/sillsdev/web-xforge/blob/328974a3b4036d34b15203efbac92ec00da382a0/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L433)):*

I've been thinking about this. I see what you are trying to do is wait for the question query to be ready before setting the this.book setter which then sets the questions to the question component.
I wonder if it would make more sense for this logic to be in the setter. Then regardless of where `this.book` is set from, it knows to subscribe to `this.questionsQuery.ready$` before running through the logic.

Let me know what you think.
</blockquote></details>

Thanks for questioning this. I've had a re-think. Initially I was trying to avoid refactoring too much but decided to do that on the `book` getter. There really was no need to have one off logic for ALL books set in there - which was causing some of the underlaying issues - as well as setting the chapter. I've also improved the subscription to better handle both ready conditions. Let me know what you think.

RaymondLuong3 (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1905#-:-NY4Ab7u8Q52mOeYcoUi:bg7igkv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 433 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1905#-NXzq7_Q1kYvdPxCFBp4:-NY49_Tn-vHSCN---wBW:bh9rho5) ([raw file](https://github.com/sillsdev/web-xforge/blob/328974a3b4036d34b15203efbac92ec00da382a0/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L433)):*

Thanks for questioning this. I've had a re-think. Initially I was trying to avoid refactoring too much but decided to do that on the `book` getter. There really was no need to have one off logic for ALL books set in there - which was causing some of the underlaying issues - as well as setting the chapter. I've also improved the subscription to better handle both ready conditions. Let me know what you think.
</blockquote></details>

I was wondering if there might be a nice way to merge the questionsQuery.ready and ready$ observables, and it looks like you found a way! I also like that now the book setter is a cleaner setter, thanks for the refactor.

RaymondLuong3 (approved):
Is there a way that testers will be able to test and confirm this fixes the issue. Either way, this will need acceptance tests so testers can confirm it is working as expected.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1905#-:-NY4B0ej-66zvslekEmh:by6b9ew)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

nigel-wells (commented):
I've added in some acceptance tests. The testers can't actually test it until it reaches QA unfortunately. How did you want to handle this @Nateowami ?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1905#-:-NYGg94DDs6sD5kr4vEv:bez1ooi)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

PR #1907, opened 2023-06-15
SF-2073 Replaced Project Deleted dialog with built-in

josephmyers:
After much research and testing, I determined that the old dialogRef was unnecessary. It was used once, in ngOnInit, potentially to prevent subscriptions later on in the function. However, the code both with and without my changes operates essentially the same and follows almost the same code path with a minor exception. ngOnInit executes multiple (three) times on project delete, and the code before and after these changes unsubscribes and then subscribes to the new project. Instead of returning at the top of the function, it returns when it finds that the selectedProjectDoc is null. The only difference is the project delete unsubscribe, which happens now in the first call instead of the second.

If there is an actual difference (maybe a use case I'm not aware of), it wasn't discernible to me. Nor do we have unit test coverage to track it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1907)


josephmyers:
I added some logs to demonstrate my findings:

(Before)
without_changes

(After)
with_changes

Nateowami (approved):
Looks good. I tracked down when the original check for the dialog ref was added, and the logic that follows (that it was avoiding running) is significantly different from what we have today. Just leaving this link here in case it ends up being useful.

https://github.com/sillsdev/web-xforge/commit/df269e302e114b37acbaea7be19b3237e9eadcd9#diff-e98dc4f713ae6a22d1bd689ad7dca2538d5b7eae7f6f2eee6e610aea9864d05e

Let's send this to the test team and see if they can break it.

> Reviewed 6 of 7 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1907#-:-NXzV9YT3Rda-3Smr7JE:bn0g094)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

PR #1909, opened 2023-06-16
SF-2059 Provide unique data ids to new note threads

RaymondLuong3:
* note threads created in SF get a new thread id
* new Paratext note threads get a unique data id
* add migrator for existing note threads to have the thread id set to the data id

This change introduces a new threadId property on the NoteThread model. On existing threads, this corresponds to the dataId property. For all new note threads, the dataId will note match the threadId. The unique data id ensures that regardless of whether a note new or recreated, we have a guarantee that no corresponding document exists in Mongo to prevent 'document already exists' errors that can occur if a note was deleted in Paratext, and recreated with the same thread ID.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1909)


pmachapman (changes requested):
The change works well. Just a couple of questions in the code, and the following question:

Is it worth migrating old notes to the new generated dataIds, or at least the notes for Biblical Terms? (querying for "648f93bcb738d6a549a9d883:BT_ἀγαθός" by id is hard, "648f93bcb738d6a549a9d883:BT_מַשָּׂא-1" is harder!) This would probably be done by deleting the existing BT notes and recreating them with the new dataIds. I don't think we need to, but I would like to know your thoughts on this.

> Reviewed 15 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1909#-:-NYFrmzB50hUCNtWM7rE:b1f3bhs)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/Models/NoteThreadChange.cs` line 19 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1909#-NYFnpYkDAiBlkDDnEkn:-NYFnpYkDAiBlkDDnEko:b7yf9v1) ([raw file](https://github.com/sillsdev/web-xforge/blob/4ea67bdb1c624cfa43e3e3e9a71949b6425139ba/src/SIL.XForge.Scripture/Models/NoteThreadChange.cs#L19)):*
> ```Smalltalk
> {
>     public const string noneId = "none";
>     public string ThreadDataId { get; set; } = noneId;
> ```

Is there a reason you use `noneId` instead of making the field nullable and just assigning null? From what I can see the ThreadDataId is not checked after it is set to `noneId` when a `NoteThreadChange` is created for a new note.
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1217 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1909#-NYFrUfPESADvoBNWBmc:-NYFrUfPESADvoBNWBmd:bxpx004) ([raw file](https://github.com/sillsdev/web-xforge/blob/4ea67bdb1c624cfa43e3e3e9a71949b6425139ba/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1217)):*
> ```Smalltalk
>         List<string> matchedThreadIds = new List<string>();
>         List<NoteThreadChange> changes = new List<NoteThreadChange>();
>         IEnumerable<IDocument<NoteThread>> activeNoteThreadDocs = noteThreadDocs.Where(
> ```

Why the change in behavior here? Does this fix another bug which we need to document in this PR?

RaymondLuong3 (commented):
That is a good question. If we do migrate the existing notes, by deleting and recreating them, it would work. Let's bring this question up at our planning meeting, someone might have an opinion of what is the better option.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1909#-:-NYJQnkU3mQjJgHh1TF2:b-ouy48z)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Models/NoteThreadChange.cs` line 19 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1909#-NYFnpYkDAiBlkDDnEkn:-NYJU1Rk5CzUv9ruh2_x:b9fs2fz) ([raw file](https://github.com/sillsdev/web-xforge/blob/4ea67bdb1c624cfa43e3e3e9a71949b6425139ba/src/SIL.XForge.Scripture/Models/NoteThreadChange.cs#L19)):*

Is there a reason you use `noneId` instead of making the field nullable and just assigning null? From what I can see the ThreadDataId is not checked after it is set to `noneId` when a `NoteThreadChange` is created for a new note.
</blockquote></details>

Yes, that sounds reasonable to me. I added a check to see if the ThreadDataId is null and if it is, we automatically create a new thread doc.
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 1217 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1909#-NYFrUfPESADvoBNWBmc:-NYJVIAG91AAXM7oSSA_:b-sd17x3) ([raw file](https://github.com/sillsdev/web-xforge/blob/4ea67bdb1c624cfa43e3e3e9a71949b6425139ba/src/SIL.XForge.Scripture/Services/ParatextService.cs#L1217)):*

Why the change in behavior here? Does this fix another bug which we need to document in this PR?
</blockquote></details>

In order for the new test `GetNoteThreadChanges_DeletedThreadRestored` to pass, I created an existing note thread doc with all the notes marked deleted. This logic was introduced because instead of matching deleted notes, and just turning them back on, we want an entirely new thread to be created. It looks like you opened SF-2086, so this might change after I look into that.

RaymondLuong3 (commented):
I added the run sheet for dropping the collection for note threads and o note threads from our database. The file is called drop note thread collection procedure. If everything here is looking good, we can send it to testing and make a plan to run the procedure on QA

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1909#-:-NYdsSdjE_xznrHhHHwy:b103rqe)* status: 8 of 15 files reviewed, 2 unresolved discussions (waiting on @pmachapman)

pmachapman (changes requested):
I have reviewed the run sheet as well (it and you changes look good), and run it on my PC (MongoDB 6.0.6).

The notes were not recreated, which is due to my PR that implements optimizations to the sync to only sync areas of Paratext that have changed (SF-1972). As no commits were incoming that changed notes, the notes were not recreated on sync.

One option to work around this is to add the following block to ParatextSyncRunner.UpdateDocsAsync():


```cs
// If we have no notes, we should retrieve them all
if (!syncResults.NotesChanged)
{
    syncResults.NotesChanged = !noteThreadDocsByBook.SelectMany(nt => nt.Value).Any();
}
```

Or we can rollback SF-1972?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1909#-:-NYoQFEPDNi5me4EtpUx:b-8fu7ks)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
Ah yes, thanks for catching that! Let's bring this up at our planning meeting.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1909#-:-NYsBHzv2KvlbZae3Kgb:b-idnkx5)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1909#-:-NYupm5-5stDUkUoZ0xu:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1920, opened 2023-06-26
SF-1991 Record Scripture Forge Version on Sync

pmachapman:
This Pull Request:
 * Adds a new `ProductVersion` property to the `SyncMetrics` class, which is set when the sync starts running
 * Centralizes retrieval of the product version in C# to a new `Product.Version` property

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1920)

### DIFF
--- a/src/SIL.XForge.Scripture/Migrator.cs
+++ b/src/SIL.XForge.Scripture/Migrator.cs
@@ -9,7 +9,7 @@ public static class Migrator
 {
     public static void RunMigrations(string environment)
     {
-        string version = FileVersionInfo.GetVersionInfo(Assembly.GetEntryAssembly().Location).ProductVersion;
+        string version = Product.Version;
         string projectPath = Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location);
 
         string migratorPath = Path.Combine(
--- a/src/SIL.XForge.Scripture/Models/SyncMetrics.cs
+++ b/src/SIL.XForge.Scripture/Models/SyncMetrics.cs
@@ -16,6 +16,7 @@ public class SyncMetrics : IIdentifiable
     public string Id { get; set; }
     public string ErrorDetails { get; set; }
     public string RequiresId { get; set; }
+    public string ProductVersion { get; set; }
     public string ProjectRef { get; set; }
     public SyncStatus Status { get; set; }
     public string UserRef { get; set; }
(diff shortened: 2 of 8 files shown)


RaymondLuong3 (commented):
I ran the sync on my machine, and I could see that mongo recorded a product version of 1.0.0 in the sync metrics collection. I assume that is what is expected when running locally. However, for localhost the version is always set to 9.9.9. Is that something we want to fix?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1920#-:-NYs_b039qc-Py8v3Wtn:b-4abal1)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge/Utils/Product.cs` line 18 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1920#-NYsZ_krDdKdRv0UGmW4:-NYsZ_krDdKdRv0UGmW5:b-3nabwk) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b9fd030ff19d91019fb998695c6cf200efb7c9b/src/SIL.XForge/Utils/Product.cs#L18)):*
> ```Smalltalk
>         }
> 
>         Version = version ?? "1.0";
> ```

Should we default to 'n/a' instead? Not a big deal since we really don't expect not to get a version.

pmachapman (commented):
I noticed that difference too - the 1.0 is coming from .NET's default version in the Assembly, the 9.9.9 from `version.json`. This discrepancy will only affect dev, as the `build-production` script sets the version in both of those places.

I can't figure out which way to go - the 9.9.9 is helpful for the frontend to show that something has gone wrong in the production build script, while 1.0.0 is the "not set default" for C#...the joys of programming in two different worlds...

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1920#-:-NYtGq7e-azHNjozKEJa:b-2dpv4h)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge/Utils/Product.cs` line 18 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1920#-NYsZ_krDdKdRv0UGmW4:-NYtJLqq2HHGPC9JpJSk:b-cd1j32) ([raw file](https://github.com/sillsdev/web-xforge/blob/8b9fd030ff19d91019fb998695c6cf200efb7c9b/src/SIL.XForge/Utils/Product.cs#L18)):*

Should we default to 'n/a' instead? Not a big deal since we really don't expect not to get a version.
</blockquote></details>

I think having a parseable version number is the best option, as this version is sent to other clients via HTTP client, so we don't know how it will be used.

PR #1922, opened 2023-06-27
SF-2111 Add Mongo Milestone Adapter

pmachapman:
To speed up retrieval of snapshots for project history, a [Milestone adapter](https://share.github.io/sharedb/adapters/milestone) is required.

This Pull Request adds the ShareDB MongoDB Milestone adapter, and enables it for just the `texts` collection with milestones created for a document every 1000 ops. When this is triggered, a `m_texts` collection is created if it does not exist, and a snapshot record is created:
image

Although we could enable this for all ShareDB collections, but as Milestones only speeds up `connection.fetchSnapshot()` and `connection.fetchSnapshotByTimestamp()`, and the default milestone interval is 1000 ops, it should only be enabled for collections that meet these requirements.

Finally, retrieval of milestones is handled transparently by the adapter when the `fetchSnapshot()` routines are called, so there are no additional changes required except the configuration of the adapter.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1922)


RaymondLuong3 (changes requested):
I was able to see the `m_text` collection appear in mongo as I was typing in a text, so it looks like it is working as it should. I just had one question about the version number.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1922#-:-NZ1yHnxEYSk6ftDrJf9:b-7jpfdm)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/RealtimeServer/package-lock.json` line 21 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1922#-NZ1pxFE-fnqaRDSqR6k:-NZ1pxFE-fnqaRDSqR6l:bbjgmee) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4206d7fbf20d19fd746ab31b720f57d456173dc/src/RealtimeServer/package-lock.json#L21)):*
> ```JSON
>         "sharedb": "^1.9.2",
>         "sharedb-access": "^5.0.0",
>         "sharedb-milestone-mongo": "^0.5.0",
> ```

I s there a reason not to use the latest verse 1.0.0?
___
*[`src/RealtimeServer/typings/sharedb/index.d.ts` line 169 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1922#-NZ1s7AF2lBCEY0DLotP:-NZ1s7AF2lBCEY0DLotQ:b-7qnkv1) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4206d7fbf20d19fd746ab31b720f57d456173dc/src/RealtimeServer/typings/sharedb/index.d.ts#L169)):*
> ```XML
>       timestamp: number,
>       callback: (err: Error | null) => void
>     ): void;
> ```

I would not expect these functions to return void. Is this because when we actually implement a SFMilestoneDB we know which methods to override, and we can change the return type then?

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1922)* status: 5 of 8 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/RealtimeServer/package-lock.json` line 21 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1922#-NZ1pxFE-fnqaRDSqR6k:-NZ2MGP88vbj0SEBvTdr:b-qs98kt) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4206d7fbf20d19fd746ab31b720f57d456173dc/src/RealtimeServer/package-lock.json#L21)):*

I s there a reason not to use the latest verse 1.0.0?
</blockquote></details>

I originally chose this version as it was the analog to our version of ShareDB, but have just reviewed the code for sharedb-mongo-milestone, I see that it properly supports our version of the MongoDB adapter and ShareDB, so I have updated to 1.0. Thanks for the nudge!
___
*[`src/RealtimeServer/typings/sharedb/index.d.ts` line 169 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1922#-NZ1s7AF2lBCEY0DLotP:-NZ2OTzM7GMkiVCvJSXG:bm3ahzu) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4206d7fbf20d19fd746ab31b720f57d456173dc/src/RealtimeServer/typings/sharedb/index.d.ts#L169)):*

I would not expect these functions to return void. Is this because when we actually implement a SFMilestoneDB we know which methods to override, and we can change the return type then?
</blockquote></details>

I based this definition off the definition at https://github.com/DefinitelyTyped/DefinitelyTyped/blob/047819443546cf8dfd7c46490309554eaf7eef77/types/sharedb/index.d.ts#L156-L163.

I notice the `abstract class DB` definition in `index.d.ts` returns void for `getOps`, `getSnapshot`, etc, so it must be because the abstract classes are overridden and the type changed there?

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1922)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/RealtimeServer/package-lock.json` line 21 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1922#-NZ1pxFE-fnqaRDSqR6k:-NZ2WE7GDrv_ESI_F5ag:b-muvaos) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4206d7fbf20d19fd746ab31b720f57d456173dc/src/RealtimeServer/package-lock.json#L21)):*

I originally chose this version as it was the analog to our version of ShareDB, but have just reviewed the code for sharedb-mongo-milestone, I see that it properly supports our version of the MongoDB adapter and ShareDB, so I have updated to 1.0. Thanks for the nudge!
</blockquote></details>

Thanks for checking on that. Glad to use the most current version when we can.
___
*[`src/RealtimeServer/typings/sharedb/index.d.ts` line 169 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1922#-NZ1s7AF2lBCEY0DLotP:-NZ2WR_G36InYqE06GkS:bnkuq0j) ([raw file](https://github.com/sillsdev/web-xforge/blob/d4206d7fbf20d19fd746ab31b720f57d456173dc/src/RealtimeServer/typings/sharedb/index.d.ts#L169)):*

I based this definition off the definition at https://github.com/DefinitelyTyped/DefinitelyTyped/blob/047819443546cf8dfd7c46490309554eaf7eef77/types/sharedb/index.d.ts#L156-L163.

I notice the `abstract class DB` definition in `index.d.ts` returns void for `getOps`, `getSnapshot`, etc, so it must be because the abstract classes are overridden and the type changed there?
</blockquote></details>

Ah, interesting. That must be it, the class with have this methods overridden.
___
*[`src/SIL.XForge.Scripture/ClientApp/package-lock.json` line 143 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1922#-NZ2VdYDCvqnunj3b1YX:-NZ2VdYEAbMH6Xp6p-LI:b-l083qn) ([raw file](https://github.com/sillsdev/web-xforge/blob/1c4cda1a00409b3fb7cce5e2e7ac17bcbd366888/src/SIL.XForge.Scripture/ClientApp/package-lock.json#L143)):*
> ```JSON
>         "sharedb": "^1.9.2",
>         "sharedb-access": "^5.0.0",
>         "sharedb-milestone-mongo": "^1.0.0",
> ```

Interesting, I wasn't expecting the change to add sharedb-milestone-mongo to change this lock file, but it looks like the other sharedb dependencies like sharedb-access is here too. So I guess it should be here.

PR #1924, opened 2023-06-27
SF-2112 Create dummy audio data

Nateowami:
This says "create dummy audio data", but it seemed easier to me to just partially implement an API method we're going to need anyway and then create a button to call it.

To try this out:

1. Add questions to a project
2. Go to the checking page
3. Click "Add timing data" by the "Add question" button
4. Fake timing data is created, and you now have a play button that doesn't play audio


RaymondLuong3 (approved):
Thanks! I did some testing and it looks like it does the trick. We can always improve on this.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1924#-:-NYx_xqx6Uzsw76WxLAG:b-hbvq6)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1929, opened 2023-06-28
SF-2120 Add ability to disable the document cache

pmachapman:
This PR adds the ability to disable the connection's document cache. This is of importance for Migrators, which will load every document in a collection just once, and not need to retrieve them again.

To enable this feature, add the following line to the Migrator's `Startup` class constructor:
```cs
Configuration["Realtime:DocumentCacheDisabled"] = "true";
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1929)

### DIFF
--- a/src/SIL.XForge/Configuration/RealtimeOptions.cs
+++ b/src/SIL.XForge/Configuration/RealtimeOptions.cs
@@ -10,7 +10,8 @@ public class RealtimeOptions
 {
     public string AppModuleName { get; set; }
     public int Port { get; set; }
-    public bool MigrationsDisabled = false;
+    public bool MigrationsDisabled { get; set; }
+    public bool DocumentCacheDisabled { get; set; }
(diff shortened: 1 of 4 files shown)


marksvc (approved):
This is great. Thanks for your work on this!

I did some testing using the migrator on a small set of projects on my workstation (10-15). I observe that watching the memory usage of the migrator with the cache disabled was tending to use 800M to 1.6 GB of residential memory. Using the migrator with the cache enabled as normal was using up to 2.7 GB.

I observed that with the cache disabled, the memory usage would fall back down again after rising. So hopefully when used on large project sets, it will continue to fall back down and not continue to grow.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1929#-:-NZ2RuHqD2M9PFrW5MlL:bruw41j)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #1930, opened 2023-06-28
SF-1502 Make email validation more flexible

spark005:
Validation for non-english emails.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1930)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/xfvalidators.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/xfvalidators.ts
@@ -1,6 +1,6 @@
 import { AbstractControl, ValidationErrors, Validators } from '@angular/forms';
 
-const EMAIL_REGEXP = /^[a-zA-Z0-9.+_-]{1,}@[a-zA-Z0-9.-]{1,}[.]{1}[a-zA-Z]{2,}$/;
+const EMAIL_REGEXP = /^.+@.+\..{2,}$/;
 const DATE_REGEXP = /^\d{4}-\d{2}-\d{2}$/;
 
 export class XFValidators {
@@ -9,13 +9,7 @@ export class XFValidators {
       return null;
     }
 
-    const emailPattern = /^.+@.+\..{2,}$/;
-    const result = Validators.pattern(emailPattern)(control);
-
-    if (result != null) {
-      return result;
-    }
-    return null;
+    return Validators.pattern(EMAIL_REGEXP)(control);
   }
 
   /** Pass if control value contains any non-whitespace. Error otherwise. */
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/xfvalidators.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/xfvalidators.ts
@@ -9,12 +9,13 @@ export class XFValidators {
       return null;
     }
 
-    const result = Validators.email(control);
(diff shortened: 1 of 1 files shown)


Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1930)* status: all files reviewed, 3 unresolved discussions (waiting on @spark005)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/xfvalidators.ts` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1930#-NZ-CUfd0caOAVIn1aab:-NZ-CUfd0caOAVIn1aac:bjknd1q) ([raw file](https://github.com/sillsdev/web-xforge/blob/c98cd5a59d6f3d88c9507a17bf2b0443b4476cc5/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/xfvalidators.ts#L3)):*
> ```XML
> import { AbstractControl, ValidationErrors, Validators } from '@angular/forms';
> 
> const EMAIL_REGEXP = /^[a-zA-Z0-9.+_-]{1,}@[a-zA-Z0-9.-]{1,}[.]{1}[a-zA-Z]{2,}$/;
> ```

EMAIL_REGEXP is no longer used and can be removed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/xfvalidators.ts` line 12 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1930#-NZ-Dq4_Eo227wx4qjGj:-NZ-Dq4_Eo227wx4qjGk:b5corh) ([raw file](https://github.com/sillsdev/web-xforge/blob/c98cd5a59d6f3d88c9507a17bf2b0443b4476cc5/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/xfvalidators.ts#L12)):*
> ```XML
>     }
> 
>     const emailPattern = /^.+@.+\..{2,}$/;
> ```

Can you move this to the top of a file as a constant (maybe name it `EMAIL_REGEXP` like the old one?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/xfvalidators.ts` line 18 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1930#-NZ-Cd-H2c6nVisEHv6y:-NZ-Cd-H2c6nVisEHv6z:bz1ecys) ([raw file](https://github.com/sillsdev/web-xforge/blob/c98cd5a59d6f3d88c9507a17bf2b0443b4476cc5/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/xfvalidators.ts#L18)):*
> ```XML
>       return result;
>     }
>     return null;
> ```

This last section:
``` ts
    if (result != null) {
      return result;
    }

    return null;
```
Is basically saying if the result is not null return it, and if it is null, return null (which in that case would be the same as returning it). (Technically it's not *exactly* the same, because if the result was `undefined`, `result != null` would be false, and it would return `null` instead of `undefined`, but the validator's documentation says it returns an error map or null, so it shouldn't ever return undefined)

It would make a lot more sense to just do:

``` ts
return Validators.pattern(emailPattern)(control);
```

Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1930#-:-NZ1sQ7hD2vfQDB3C_GQ:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @spark005)

PR #1946, opened 2023-07-05
SF-2129 Add Health Endpoint

pmachapman:
This Pull Request adds a `/health-check/` endpoint, accessible only from localhost, that can be used with the NodePing Agent. This will enable use to improve monitoring of Scripture Forge.

To test this, use http://localhost:5000/swagger/index.html. No authentication is necessary due to the localhost restriction.

Documentation is available in the document "Health endpoint" in the shared drive.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1946)

### DIFF
--- a/src/SIL.XForge.Scripture/Controllers/HealthController.cs
+++ b/src/SIL.XForge.Scripture/Controllers/HealthController.cs
@@ -1,0 +1,165 @@
+using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Mvc;
+using Serval.Client;
+using SIL.XForge.DataAccess;
+using SIL.XForge.Realtime;
+using SIL.XForge.Scripture.Models;
+
+namespace SIL.XForge.Scripture.Controllers;
+
+/// <summary>
+/// Provides methods for checking on instance health.
+/// </summary>
+/// <remarks>
+/// This can only be accessed from localhost, and is designed for NodePing.
+/// </remarks>
+[Route("health-check")]
+[ApiController]
+public class HealthController : ControllerBase
+{
+    public const int Status531MongoDown = 531;
+    public const int Status532RealtimeServerDown = 532;
+    public const int Status533ServalDown = 533;
+
+    private readonly IExceptionHandler _exceptionHandler;
+    private readonly IRealtimeService _realtimeService;
+    private readonly ITranslationEnginesClient _translationEnginesClient;
+
+    public HealthController(
+        IExceptionHandler exceptionHandler,
+        IRealtimeService realtimeService,
+        ITranslationEnginesClient translationEnginesClient
+    )
+    {
+        _exceptionHandler = exceptionHandler;
+        _realtimeService = realtimeService;
+        _translationEnginesClient = translationEnginesClient;
+    }
+
+    /// <summary>
+    /// Executes the health check.
+    /// </summary>
+    /// <returns>A JSON object containing values corresponding to the health of various systems.</returns>
+    /// <response code="200">All systems are healthy.</response>
+    /// <response code="403">You do not have permission to view the health check.</response>
+    /// <response code="531">Mongo is down.</response>
+    /// <response code="532">The Realtime Server is down.</response>
+    /// <response code="533">Serval is down.</response>
+    [HttpGet]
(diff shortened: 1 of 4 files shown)


RaymondLuong3 (changes requested):
Nice work. I tried this out on Swagger, but I couldn't get it to return success. It showed that my real-time server was down. What should I do to get the health check working? I started the app with `dotnet run`


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1946#-:-NZlCwYN-DHyBE1qxMHw:b-5t24bl)* status: all files reviewed, 4 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Controllers/HealthController.cs` line 139 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1946#-NZlCIwJ2iwRxiuZdVoE:-NZlCIwJ2iwRxiuZdVoF:b32ogsx) ([raw file](https://github.com/sillsdev/web-xforge/blob/e1ce2f8727c8908ed8e9c6d239c3530d0da77317/src/SIL.XForge.Scripture/Controllers/HealthController.cs#L139)):*
> ```Smalltalk
>         else
>         {
>             return Ok(response);
> ```

The final return would never get hit from what I can tell since the OK response will always return.

pmachapman (commented):
I am not sure why that is happening - what does the first project in your MongoDB `sf_projects` collection look like?

> Dismissed @github-code-scanning[bot] from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1946#-:-NZwAJzG9DUzdonjrFDh:brcribo)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/Controllers/HealthController.cs` line 139 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1946#-NZlCIwJ2iwRxiuZdVoE:-NZw9uhT8yih76SsLmHG:b6h29w3) ([raw file](https://github.com/sillsdev/web-xforge/blob/e1ce2f8727c8908ed8e9c6d239c3530d0da77317/src/SIL.XForge.Scripture/Controllers/HealthController.cs#L139)):*

The final return would never get hit from what I can tell since the OK response will always return.
</blockquote></details>

The final return returns the status codes in the `if else` blocks above this `else`. I could just change this line to `statusCode = 200` if you think that's clear - I chose the `Ok(response)` as it made my tests look a little clearer when checking for success.

RaymondLuong3 (approved):
Ah, you were on to it. The first entry in the `sf_projects` collection was a deleted project that was left over somehow. So the data was null. I removed it from my mongo, and the health endpoint works now. I also see that it is working. I will point out that the message from the serval health endpoint says `the circuit is now open and is not allowing calls` doesn't make a whole lot of sense to me. Do we have control over that message?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1946#-:-N_00gAT9PNQQErOlYs5:b-yodn16)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Controllers/HealthController.cs` line 139 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1946#-NZlCIwJ2iwRxiuZdVoE:-N_-zdin0W3bfB63vRrE:bsgzc03) ([raw file](https://github.com/sillsdev/web-xforge/blob/2ce4c6e2a73b90f3eb7579ba6c5c5a6cd7573b23/src/SIL.XForge.Scripture/Controllers/HealthController.cs#L139)):*

The final return returns the status codes in the `if else` blocks above this `else`. I could just change this line to `statusCode = 200` if you think that's clear - I chose the `Ok(response)` as it made my tests look a little clearer when checking for success.
</blockquote></details>

Ah yes, you are right. I got hung up on the fact that the else block returned the OK response.

pmachapman (commented):
That error message is generated by Polly. I originally was going to have an entire stack trace, but then realized that will be on Bugsnag, so all I need to present is the error message, and the developer who is looking into it can view the complete details there.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1946#-:-N_0CW-m0QBBPmkGpbuD:b-wotspk)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #1948, opened 2023-07-05
SF-2132 Hotfix: Correctly handle note content with mixed styles

RaymondLuong3:
If a Paratext comment has content where the bold or italics style is used, the content does not show correctly in SF. The reason is because we used contentNode.Elements() to get all elements in the content node. However, unformatted text nodes are not elements, so they were excluded. The result is that when a note is created in SF, the content would be missing the unformatted text if the content had mixed styles like bold or italics.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1948)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -15,6 +15,7 @@ using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Xml;
 using System.Xml.Linq;
 using System.Xml.XPath;
 using IdentityModel;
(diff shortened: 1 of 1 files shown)


marksvc (approved):
This would be a good case to write a test over the problem.
I have a couple other suggestions for improvement, but none that I want to hold up the hotfix for. 

I did not locally test the code.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1948#-:-NZcC5eK3R7c2ER933lE:b-kj2j5h)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2410 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1948#-NZcAdog-_2v2QUr6O_j:-NZcAdog-_2v2QUr6O_k:b471k4w) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ecc0e62385d4c18a064492071c962413edb7b36/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2410)):*
> ```Smalltalk
>         XNode[] nodes = contentNode.Nodes().ToArray();
>         if (!nodes.Any())
>             return contentNode.Value;
> ```

This presumably used to return the single text node in the contentNode. Now it will be a bit different since a single text node would result in not getting here. I suspect we will only get here if contentNode is empty, like <Foo></Foo>. It might make more sense to `return string.Empty;` to be clear about what is going on. 
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2425 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1948#-NZcBPcmAdVrrJqk5W73:-NZcBPcmAdVrrJqk5W74:bexkren) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ecc0e62385d4c18a064492071c962413edb7b36/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2425)):*
> (XElement)node

Hmm. It could conceivably be something else, like a comment. It may be safer to do something like
```cs
if node is not an element, then:
    return
XElement element = (XElement)node;
```
oh or maybe
```cs
XElement element = node as XElement 
if (element is null)
    return;
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1948)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2410 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1948#-NZcAdog-_2v2QUr6O_j:-NZcFRnr0v0URCj76HSM:b9dd993) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ecc0e62385d4c18a064492071c962413edb7b36/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2410)):*

This presumably used to return the single text node in the contentNode. Now it will be a bit different since a single text node would result in not getting here. I suspect we will only get here if contentNode is empty, like <Foo></Foo>. It might make more sense to `return string.Empty;` to be clear about what is going on. 
</blockquote></details>

Hmm, I think you are right. We could return string.Empty here. When I put this change on master I'll use that.
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2425 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1948#-NZcBPcmAdVrrJqk5W73:-NZcFiA5Ev4P0aUwhOBW:bquvmmz) ([raw file](https://github.com/sillsdev/web-xforge/blob/6ecc0e62385d4c18a064492071c962413edb7b36/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2425)):*

> (XElement)node

Hmm. It could conceivably be something else, like a comment. It may be safer to do something like
```cs
if node is not an element, then:
    return
XElement element = (XElement)node;
```
oh or maybe
```cs
XElement element = node as XElement 
if (element is null)
    return;
```
</blockquote></details>

Good thinking. When I put this change on master I'll do that.

PR #1953, opened 2023-07-07
SF-2132 Correctly handle note content with mixed styles

RaymondLuong3:
This change introduces more reliable parsing of Paratext note contents. The problem was that in content with mixed formatting such as bold and italics, the regularly formatted text would be a text node that is not included when calling contentNode.Elements(). To prevent dropping the text, we use contentNode.Nodes() instead which retrieves all nodes in the Paratext note contents. More rigorous testing of the parsing results was also added.

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -15,6 +15,7 @@ using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Xml;
 using System.Xml.Linq;
 using System.Xml.XPath;
 using IdentityModel;
@@ -2409,26 +2410,34 @@ public class ParatextService : DisposableBase, IParatextService
             return content;
         XDocument doc = XDocument.Parse(content);
         XElement contentNode = (XElement)doc.FirstNode;
-        XElement[] elements = contentNode.Elements().ToArray();
-        if (!elements.Any())
+        XNode[] nodes = contentNode.Nodes().ToArray();
+        if (!nodes.Any())
             return contentNode.Value;
 
         int paragraphNodeCount = ((XElement)doc.FirstNode).Elements("p").Count();
         StringBuilder sb = new StringBuilder();
         bool isReviewer = false;
-        for (int i = 0; i < elements.Length; i++)
+        for (int i = 0; i < nodes.Length; i++)
         {
-            XElement elem = elements[i];
+            XNode node = nodes[i];
+            if (node.NodeType == XmlNodeType.Text)
+            {
+                // append text to the content string
+                sb.Append(node);
+                continue;
+            }
+
+            XElement element = (XElement)node;
             // check if the paragraph element contains the user label class
-            if (elem.Attribute("sf-user-label")?.Value == "true")
+            if (element.Attribute("sf-user-label")?.Value == "true")
                 isReviewer = true;
             if (i == 0 && isReviewer)
                 continue;
             // If there is only one paragraph node other than the SF user label then omit the paragraph tags
             if (isReviewer && paragraphNodeCount <= 2)
-                sb.Append(elem.Value);
+                sb.Append(element.Value);
             else
-                sb.Append(elem);
+                sb.Append(node);
         }
         return sb.ToString();
     }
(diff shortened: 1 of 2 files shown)


pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1953)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @marksvc and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2429 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1953#-NZwioQH-JjF-gp6gm29:-NZwioQH-JjF-gp6gm2A:b-qjrwhp) ([raw file](https://github.com/sillsdev/web-xforge/blob/b31ec73808c69bb9a9fdd9cfa00621e6cfd9dadc/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2429)):*
> ```Smalltalk
>                 continue;
>             }
>             else if (node.NodeType != XmlNodeType.Element)
> ```

Do we want to handle an `XmlNodeType.Whitespace` node when its occurs between two XmlNodeType.Element nodes?

For example `<p><bold>Bold</bold> <italic>Italic</italic></p>` will render as: **Bold***Italic*, not **Bold** *Italic*.

I notice that Paratext does not handle the whitespace nodes correctly, so your implementation has bug for bug compatibility (i.e. I don't think we need to change your implementation).

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1953)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @marksvc and @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2429 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1953#-NZwioQH-JjF-gp6gm29:-N_-xUSFASoGWwnfXMaf:blwlkcj) ([raw file](https://github.com/sillsdev/web-xforge/blob/b31ec73808c69bb9a9fdd9cfa00621e6cfd9dadc/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2429)):*

Do we want to handle an `XmlNodeType.Whitespace` node when its occurs between two XmlNodeType.Element nodes?

For example `<p><bold>Bold</bold> <italic>Italic</italic></p>` will render as: **Bold***Italic*, not **Bold** *Italic*.

I notice that Paratext does not handle the whitespace nodes correctly, so your implementation has bug for bug compatibility (i.e. I don't think we need to change your implementation).

</blockquote></details>

It would be good if we could handle the whitespace between nodes, and actually, when it is within a paragraph like your example, it gets handled since the whole paragraph node gets appended to the string builder. The time it does not handle it correctly is with whitespace between paragraph nodes as my test pointed out. And the newLine character doesn't register as a whitespace node and gets skipped. But the round-tripping still works.
But let me take a look to see if we might be able to infer a line break between paragraph nodes.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1953)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @marksvc and @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2429 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1953#-NZwioQH-JjF-gp6gm29:-N_0RO3kALDIBabh0zfq:b-ciwx6m) ([raw file](https://github.com/sillsdev/web-xforge/blob/b31ec73808c69bb9a9fdd9cfa00621e6cfd9dadc/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2429)):*

It would be good if we could handle the whitespace between nodes, and actually, when it is within a paragraph like your example, it gets handled since the whole paragraph node gets appended to the string builder. The time it does not handle it correctly is with whitespace between paragraph nodes as my test pointed out. And the newLine character doesn't register as a whitespace node and gets skipped. But the round-tripping still works.
But let me take a look to see if we might be able to infer a line break between paragraph nodes.
</blockquote></details>

You're right, it looks like we strip out the white space between xml nodes when we compare PT comments to notes. It works for paragraph nodes, but is not helpful for other types of nodes. I've made the change so that it only strips white space from between paragraph nodes.
So now we should be handling the case with white space between bold and italics nodes correctly.

pmachapman (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1953)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1954, opened 2023-07-08
Losslessly compress PNG images using zopflipng

Nateowami:
This was done by running:
``` bash
git ls-files | grep '\.png$' | xargs -n 1 -P 16 -I {} zopflipng --iterations=1000 -y {} {}
```

- The `-n 1` to xargs tells it to only give one file each time it calls zopflipng, rather than passing a whole bunch of files to a single invocation
- The `-P 16` tells xargs it can run 16 processes in parallel. This speeds things up dramatically.
- The `-I {}` tells xargs to replace `{}` with the args it's passing. This allows us to pass the file name to zopflipng as both an input and output file name.
- 1000 iterations is pretty ridiculous but was done for good measure

Zopfli is a compression algorithm from Google that trades speed for compression: https://github.com/google/zopfli

It's a compression-only algorithm that implements DEFLATE, zlib, and gzip (I think those are all different things?). Files can be decompressed in the exact same way as before.

You wouldn't use it for compression on the fly; it's intended for assets you compress once and then serve to users repeatedly.

Most of the files that were dramatically compressed were fairly small to begin with.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1954)


pmachapman (approved):
Thank you for this change! The command you documented will work for just about any git repo, too.

> Reviewed 122 of 122 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1954#-:-NZwMpIzE92QRv5Dfty4:b-xekp1u)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #1958, opened 2023-07-11
SF-2137 Add notes on threads with the correct thread Id

RaymondLuong3:
The problem was that the thread Id for new notes on a thread was mistakenly being set to the thread data Id. The result was the thread Id for new notes was different than the first note in the thread, so when the note was synchronized, the note ended up on a new thread.
This change ensures that the correct thread Id is used for subsequent notes on a thread.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1958)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts
@@ -96,8 +96,8 @@ export class SFProjectService extends ProjectService<SFProject, SFProjectDoc> {
     return this.realtimeService.subscribe(TextDoc.COLLECTION, textId instanceof TextDocId ? textId.toString() : textId);
   }
 
-  getNoteThread(threadId: string): Promise<NoteThreadDoc> {
-    return this.realtimeService.subscribe(NoteThreadDoc.COLLECTION, threadId);
+  getNoteThread(threadDataId: string): Promise<NoteThreadDoc> {
+    return this.realtimeService.subscribe(NoteThreadDoc.COLLECTION, threadDataId);
   }
 
   queryQuestions(
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -2734,23 +2734,27 @@ describe('EditorComponent', () => {
     it('allows adding a note to an existing thread', fakeAsync(() => {
       const projectId: string = 'project01';
       const threadDataId: string = 'dataid04';
+      const threadId: string = 'thread04';
       const segmentRef: string = 'verse_1_3';
       const env = new TestEnvironment();
       const content: string = 'content in the thread';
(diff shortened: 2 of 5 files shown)


pmachapman (changes requested):
Looks good - just one comment on a line that initially puzzled me.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1958#-:-N_e_yFvCcXUQS-sXkED:bohtb62)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 960 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1958#-N_eYJYL9nw-TW5K0ihQ:-N_eYJYL9nw-TW5K0ihR:bi6vghz) ([raw file](https://github.com/sillsdev/web-xforge/blob/77059714816296f5545aa933b84240de30053317/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L960)):*
> ```XML
>     }
>     const currentDate: string = new Date().toJSON();
>     const threadId: string = params.threadDataId != null ? '' : objectId();
> ```

I think I understand why you are setting this to a blank value based on the logic below (i.e. if the threadDataId is not null, you will set the threadId further down), but it still looks a little unconventional. Are you able to write a brief comment for this line?

I think you don't even have to include this modification, as you correct the note.threadId before the submoitJson0Op anyway?

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1958)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 960 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1958#-N_eYJYL9nw-TW5K0ihQ:-N_iyd2D31fQQZVVVQU6:b-lnlxeg) ([raw file](https://github.com/sillsdev/web-xforge/blob/77059714816296f5545aa933b84240de30053317/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L960)):*

I think I understand why you are setting this to a blank value based on the logic below (i.e. if the threadDataId is not null, you will set the threadId further down), but it still looks a little unconventional. Are you able to write a brief comment for this line?

I think you don't even have to include this modification, as you correct the note.threadId before the submoitJson0Op anyway?
</blockquote></details>

Yes, I can see why it is confusing. The threadId isn't the best name, I changed it to newThreadId. Since we only need it if we do have a comment on a new thread, I chose to set it to the empty string in the event the thread already exists. The only benefit is that we do not call `objectId()` and then discard the value. I added a comment and hope it helps.

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1958#-:-N_jUoyUAZ0fSSOFpRhZ:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1960, opened 2023-07-12
SF-2146 Write Comment XML Files with the BOM

pmachapman:
This Pull Request changes the method to write Comment XML files to use a custom XML Serializer rather than ParatextData's `CommentManager SaveUser()` method which uses `Momento.SetXml()`.

A benefit of this method is the fact we can now force the BOM to be written (matching the behavior of ParatextData running on the .NET Framework), and we can also unit test comment writing, to flag in future if either Paratext changes its XML schema for the Comment class or if future versions of .NET change the XML Serialization (as happened with the upgrade to .NET 6.0).

This implementation does not create the .bak file when the new comment XML is written - if this is a blocker we may want to look at our own method of storing this (i.e. versioned .bak files or in a different directory).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1960)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -2048,7 +2048,7 @@ public class ParatextService : DisposableBase, IParatextService
         try
         {
             foreach (string user in users)
-                manager.SaveUser(user, false);
+                WriteCommentXml(manager, user);
             _paratextDataHelper.CommitVersionedText(
                 scrText,
                 $"{syncMetricInfo.Added} notes added and "
@@ -2068,6 +2068,17 @@ public class ParatextService : DisposableBase, IParatextService
         }
 
         return syncMetricInfo;
+    }
+
+    private void WriteCommentXml(CommentManager commentManager, string username)
+    {
+        CommentList userComments = new CommentList(
+            commentManager.AllComments.Where(comment => comment.User == username)
+        );
+        string fileName = commentManager.GetUserFileName(username);
+        string path = Path.Combine(commentManager.ScrText.Directory, fileName);
+        using Stream stream = _fileSystemService.CreateFile(path);
+        _fileSystemService.WriteXmlFile(stream, userComments);
     }
 
     private CommentTags? GetCommentTags(UserSecret userSecret, string paratextId)
--- a/src/SIL.XForge/Services/FileSystemService.cs
+++ b/src/SIL.XForge/Services/FileSystemService.cs
@@ -1,5 +1,8 @@
 using System.Collections.Generic;
 using System.IO;
+using System.Text;
+using System.Xml;
+using System.Xml.Serialization;
 
 namespace SIL.XForge.Services;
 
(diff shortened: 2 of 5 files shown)


RaymondLuong3 (changes requested):
This all looks good, but I was surprised to see that with this branch checked out, I did a send/receive on my windows machine and it still had the issue of removing the BOM from between status and type tags in my notes xml file. Are you seeing something different? Maybe there was an additional configuration I had not set.



> Reviewed 3 of 5 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1960#-:-N__0sr7CfFo9VRlx647:b-pp7xg8)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge/Services/FileSystemService.cs` line 47 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1960#-N__0cKG7asFz_DoiPaM:-N__0cKHClE7EB2oTOaz:bw7wccq) ([raw file](https://github.com/sillsdev/web-xforge/blob/ad77f45484ad0e457a2953dac72859c5fa2b62bc/src/SIL.XForge/Services/FileSystemService.cs#L47)):*
> ```Smalltalk
>             Encoding = new UTF8Encoding(true),
>             Indent = true,
>             NewLineChars = "\r\n",
> ```

Is this because xml files for Paratext by default have the \r\n line ending? That is good to know.

pmachapman (commented):
The diff you have screenshotted looks correct to me? From my understanding it is:

 * Adds the missing BOM
 * Changes the newlines in the `<Status>` and `<Type>` elements to an empty string.

When I create notes in Paratext on Windows, the XML file has the BOM and no newline for the `<Status>` and `<Type>` elements, so your XML has now had the SF changes undone.

Do notes created in Paratext for you include the BOM and exclude the newlines between the Status and Type elements?

If you add me to the project, I can see what else is going on.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1960#-:-N__hDKECb8HKz8yoEaH:bm54vsa)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge/Services/FileSystemService.cs` line 47 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1960#-N__0cKG7asFz_DoiPaM:-N__h0T8282TDhHiajzd:bx3yxz2) ([raw file](https://github.com/sillsdev/web-xforge/blob/ad77f45484ad0e457a2953dac72859c5fa2b62bc/src/SIL.XForge/Services/FileSystemService.cs#L47)):*

Is this because xml files for Paratext by default have the \r\n line ending? That is good to know.
</blockquote></details>

Yes, the `\r\n` is from Paratext's default `XmlWriterSettings`.

RaymondLuong3 (commented):
Yes, you are correct, I was not looking at it the correct way. It looks like now both Paratext and SF are creating notes with the newline removed and the BOM in place. I'll do more testing tomorrow and then give it my approval. I did add you to the project in case you wanted to see the history.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1960#-:-N_a0oYq7vPfhhgxCGl1:b-8klisn)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

RaymondLuong3 (approved):
Thanks, it is working as I expected it, or more precisely, I have learned how it should work and it works that way! I'm not too concerned about the missing `.BAK` files since we already make a backup of a repo before we sync, so I'm happy with it as-is.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1960#-:-N_j1EZGB85iO2Jvjyse:b-jph1e2)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #1966, opened 2023-07-13
dev: add test:headless

marksvc:
commit-id:b0c24161

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/package.json
+++ b/src/SIL.XForge.Scripture/ClientApp/package.json
@@ -12,6 +12,7 @@
     "test": "ng test --source-map",
     "test:tc": "ng test --browsers xForgeChromiumHeadless --watch false --code-coverage --source-map",
     "test:gha": "KARMA_REPORTERS=mocha,coverage-istanbul ng test --browsers xForgeChromiumHeadless --watch false --code-coverage --source-map",
+    "test:headless": "KARMA_REPORTERS=mocha ng test --browsers xForgeChromiumHeadless --source-map",
     "test-storybook": "test-storybook",
     "lint": "ng lint",
     "prettier": "prettier --write \"**/*.{ts,js,json,css,scss,less,html,md,yml}\"",
--- a/src/SIL.XForge.Scripture/ClientApp/src/karma.conf.js
+++ b/src/SIL.XForge.Scripture/ClientApp/src/karma.conf.js
@@ -55,6 +55,10 @@ module.exports = function (config) {
 
     port: 9876,
     colors: true,
+    mochaReporter: {
+      // https://github.com/litixsoft/karma-mocha-reporter
+      output: 'autowatch'
+    },
     logLevel: config.LOG_INFO,
     autoWatch: true,
     captureTimeout: 120000, // compile needs to finished otherwise first capture fails


marksvc:
This is a nice little way to run tests headlessly, and with watching on. 

image

It is compatible with "Attach to frontend tests" ([ref](https://github.com/sillsdev/web-xforge/wiki/Debugging#attach-debugger)), and can be used with `--include` (eg 
```bash
npm run test:headless -- --include src/app/shared/text/text.component.spec.ts --include src/app/translate/editor/editor.component.spec.ts
```

Nateowami:
Pretty sure this is not going to work on Windows for the same reason #1644 is stalled. You can start a command with `KARMA_REPORTERS=mocha` on Linux but not Windows.

pmachapman (changes requested):
We can get it working on Windows and Linux using a pre script that sets the environment variable in this way:
```
    "pretest:headless": "(KARMA_REPORTERS=mocha) || (set KARMA_REPORTERS=mocha)",
    "test:headless": "ng test --browsers xForgeChromiumHeadless --source-map",
```

What the pre script tries to do is set the variable the Unix way, then if that fails (which it will on Windows), the Windows way.

Does this work in your environment?

If it does we could use the same method for `test:gha` and #1644.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1966#-:-N__U5FEAtQxGSm4vU8m:bdsdy9c)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

pmachapman (changes requested):
It turns out environment variables are not carried across from the pre script to the main script.

I had another look into this, and the following works for me - the only downside is Windows users get a `'KARMA_REPORTERS' is not recognized as an internal or external command, operable program or batch file.` error in the console, but the code still runs (and the environment variable is set):

```
"test:headless": "(KARMA\_REPORTERS=mocha || set KARMA\_REPORTERS=mocha) && ng test --browsers xForgeChromiumHeadless --source-map",
```

The other alternative would be to use the `cross-env` npm library?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1966#-:-N___LLC66RXzTez237Q:b-catzce)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

marksvc:
Note that the parentheses cause KARMA_REPORTERS to be set in a subshell, which won't make the variable available to the parent. For example, consider
```bash
$ (A=b ; echo 1 $A) ; echo 2 $A
1 b
2
```
`b` is not printed after `2`.

Also, we need to export KARMA_REPORTERS so it's available to `ng test`.

I tried using
```json
    "pretest:headless": "export KARMA_REPORTERS=mocha",
    "test:headless": "ng test --browsers xForgeChromiumHeadless --source-map",
```
But I didn't see any indication that it had run the pre command. :/

It's possible to use `set` in bash if we change how it behaves. For example, 
```bash
$ alias set=
$ set A=b
$ echo $A
b
```

The stack answer at https://stackoverflow.com/a/68482269 shows writing a bash command prefixed by `echo ; ` and suffixed with `#>NUL`, to make it be tolerated on cmd and not print anything. Unfortunately, `#` ends the command on bash and so I can't follow up with `&& npm test ...`. 

I would have thought the following would work on Linux, but it doesn't.
```bash
sh -c 'alias set=export  && set KARMA_REPORTERS=mocha && echo KR is $KARMA_REPORTERS'
```
(Note that `sh` is being called by npm rather than `bash`).
It seems that `sh` needs you to press enter after `alias set=export` rather than use `;` or `&&`.
And `set` seems to be a reserved name, so I can't write a function with that name, as in the stack answer.

A working solution seems to be
```json
"test:headless": "echo;export KARMA_REPORTERS=mocha && set KARMA_REPORTERS=mocha && ng test --browsers xForgeChromiumHeadless --source-map",
```
But then it will print `export KARMA_REPORTERS=mocha` on Windows. Though I suppose that's not terribly bad.

pmachapman (approved):
That is awesome! :lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1966#-:-NbMcFaZ-VTsRUZ-8d0a:b-r2zz6c)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

pmachapman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1966)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/package.json` line 16 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1966#-NbMf1oc87ixbmop4aSd:-NbMf1oc87ixbmop4aSe:b4ex61w) ([raw file](https://github.com/sillsdev/web-xforge/blob/dd955063a37012e3721d55b002a2b8808959b2ae/src/SIL.XForge.Scripture/ClientApp/package.json#L16)):*
> ```JSON
>     "test:gha": "KARMA_REPORTERS=mocha,coverage-istanbul ng test --browsers xForgeChromiumHeadless --watch false --code-coverage --source-map",
>     "//": "The environment variable is set in a way compatible with both sh and cmd.",
>     "test:headless": "echo;export KARMA_REPORTERS=mocha && set KARMA_REPORTERS=mocha && ng test --browsers xForgeChromiumHeadless --source-map",
> ```

I was too excited by your solution to let my test on Windows finish...

Just one minor fix to get this running on Windows.
```
"test:headless": "echo;export KARMA_REPORTERS=mocha && set KARMA_REPORTERS=mocha&& ng test --browsers xForgeChromiumHeadless --source-map",
```
(removal of the space before `&& ng test`)

This fixes the error: `09 08 2023 12:20:51.662:ERROR [reporter]: Can not load reporter "mocha ", it is not registered!`

marksvc:
> (removal of the space before && ng test)
> 
> This fixes the error: 09 08 2023 12:20:51.662:ERROR [reporter]: Can not load reporter "mocha ", it is not registered!

Interesting. Why do you think that is? And you're sure it isn't something like your karma.config.js file wasn't saved at the time or something?

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1966)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/package.json` line 16 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1966#-NbMf1oc87ixbmop4aSd:-NbuJ-dh9Zs7ENTa2Kuh:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/dd955063a37012e3721d55b002a2b8808959b2ae/src/SIL.XForge.Scripture/ClientApp/package.json#L16)):*

I was too excited by your solution to let my test on Windows finish...

Just one minor fix to get this running on Windows.
```
"test:headless": "echo;export KARMA_REPORTERS=mocha && set KARMA_REPORTERS=mocha&& ng test --browsers xForgeChromiumHeadless --source-map",
```
(removal of the space before `&& ng test`)

This fixes the error: `09 08 2023 12:20:51.662:ERROR [reporter]: Can not load reporter "mocha ", it is not registered!`
</blockquote></details>

Done.

pmachapman (approved):
It failed because Windows does not trim the space at between the `mocha` and `&&` (behavior I was unaware of!), so set the reporter to `"mocha "`.

Your fix works perfectly - thank you!

:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1966#-:-NbuaT1d9NvlhKytEs5E:b-nyz5ac)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #1967, opened 2023-07-13
SF-2028 Hide community checking text when requested

marksvc:
Note that this PR introduces a regression: "Audio unavailable" is not shown when the user navigates to a chapter that does not have audio. This can be addressed in followup.

----

Note that this PR is composed in different commits. There are two refactor commits at the beginning of the PR that are separate from other changes. Hopefully that will make the review easier. **It is okay if the whole PR is squashed together for submitting. Although I would like for all the commit messages to be retained.** The individual commits were not individually tested.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1967)

### DIFF
--- a/src/RealtimeServer/scriptureforge/models/checking-config.ts
+++ b/src/RealtimeServer/scriptureforge/models/checking-config.ts
@@ -10,4 +10,5 @@ export interface CheckingConfig {
   shareEnabled: boolean;
   answerExportMethod: CheckingAnswerExport;
   noteTagId?: number;
+  hideCommunityCheckingText?: boolean;
 }
--- a/src/RealtimeServer/scriptureforge/models/sf-project-test-data.ts
+++ b/src/RealtimeServer/scriptureforge/models/sf-project-test-data.ts
@@ -23,7 +23,8 @@ function testProjectProfile(ordinal: number): SFProjectProfile {
       checkingEnabled: true,
       usersSeeEachOthersResponses: true,
       shareEnabled: false,
-      answerExportMethod: CheckingAnswerExport.MarkedForExport
+      answerExportMethod: CheckingAnswerExport.MarkedForExport,
+      hideCommunityCheckingText: false
     },
     texts: [],
     sync: {
--- a/src/RealtimeServer/scriptureforge/models/text-audio-test-data.ts
+++ b/src/RealtimeServer/scriptureforge/models/text-audio-test-data.ts
@@ -1,0 +1,18 @@
+import merge from 'lodash/merge';
+import { RecursivePartial } from '../../common/utils/type-utils';
+import { TextAudio } from './text-audio';
+
+function testTextAudio(ordinal: number): TextAudio {
+  return {
+    ownerRef: '',
+    projectRef: '',
+    dataId: '',
+    timings: [],
+    mimeType: 'audio/opus',
+    audioUrl: `https://example.com/file${ordinal}.opus`
+  };
+}
+
+export function createTestTextAudio(overrides?: RecursivePartial<TextAudio>, ordinal = 1): TextAudio {
+  return merge(testTextAudio(ordinal), overrides);
+}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.html
@@ -10,7 +10,7 @@
       <mat-icon class="mirror-rtl">skip_next</mat-icon>
     </button>
   </div>
-  <button mat-icon-button class="close-button" (click)="close()">
+  <button *ngIf="canClose" mat-icon-button class="close-button" (click)="close()">
     <mat-icon>close</mat-icon>
   </button>
   <app-audio-player #audioPlayer [source]="audioSource"></app-audio-player>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts
@@ -16,36 +16,26 @@ const audioFile = 'test-audio-player.webm';
 const textDocId: TextDocId = new TextDocId('project01', 1, 1);
 
 describe('ScriptureAudioComponent', () => {
-  let env: TestEnvironment;
-  beforeEach(fakeAsync(() => {
-    const template = `<app-checking-scripture-audio-player source="${audioFile}"></app-checking-scripture-audio-player>`;
-    env = new TestEnvironment(template);
-    env.fixture.detectChanges();
-
-    env.component.audioPlayer.timing = getAudioTimings();
-    env.component.audioPlayer.textDocId = textDocId;
-    env.wait();
-  }));
-
   it('can play and pause audio', fakeAsync(() => {
-    const play = spyOn(env.audioPlayer.audio, 'play').and.callThrough();
-    const pause = spyOn(env.audioPlayer.audio, 'pause').and.callThrough();
+    const env = new TestEnvironment();
 
     env.playButton.nativeElement.click();
     env.wait();
 
     expect(env.isPlaying).toBe(true);
-    expect(play).toHaveBeenCalledTimes(1);
-    expect(pause).toHaveBeenCalledTimes(0);
+    expect(env.audioPlaySpy).toHaveBeenCalledTimes(1);
+    expect(env.audioPauseSpy).toHaveBeenCalledTimes(0);
 
     env.playButton.nativeElement.click();
     env.wait();
 
-    expect(play).toHaveBeenCalledTimes(1);
-    expect(pause).toHaveBeenCalledTimes(1);
+    expect(env.audioPlaySpy).toHaveBeenCalledTimes(1);
+    expect(env.audioPauseSpy).toHaveBeenCalledTimes(1);
   }));
 
   it('can skip to next and previous verse', fakeAsync(() => {
+    const env = new TestEnvironment();
+
     expect(env.verseLabel.nativeElement.textContent).toEqual('Genesis 1:1');
 
     env.clickNextRef();
@@ -63,7 +53,7 @@ describe('ScriptureAudioComponent', () => {
   }));
 
   it('can skip forward and back through section headings', fakeAsync(() => {
-    env.component.audioPlayer.timing = getAudioTimingWithHeadings();
+    const env = new TestEnvironment({ timings: getAudioTimingWithHeadings() });
 
     env.clickNextRef();
     env.wait();
@@ -84,73 +74,64 @@ describe('ScriptureAudioComponent', () => {
   }));
 
   it('emits verse changed event', fakeAsync(() => {
-    const verseChangedSpy = jasmine.createSpy('verseChanged');
-    env.component.audioPlayer.currentVerseChanged.subscribe(verseChangedSpy);
-    expect(verseChangedSpy).toHaveBeenCalledTimes(0);
+    const env = new TestEnvironment();
+
+    expect(env.verseChangedSpy).withContext('it already announced where we started').toHaveBeenCalledTimes(1);
 
     env.audioPlayer.audio.currentTime = 1.5;
     env.audioPlayer.audio.timeUpdated$.next();
-    expect(verseChangedSpy).toHaveBeenCalledTimes(1);
-    expect(verseChangedSpy).toHaveBeenCalledWith('verse_1_2');
+    expect(env.verseChangedSpy).toHaveBeenCalledTimes(2);
+    expect(env.verseChangedSpy).toHaveBeenCalledWith('verse_1_2');
   }));
 
   it('emits verse changed event for section headings', fakeAsync(() => {
     const timings: AudioTiming[] = getAudioTimingWithHeadings();
-    env.component.audioPlayer.timing = timings;
-
-    const verseChangedSpy = jasmine.createSpy('verseChanged');
-    env.component.audioPlayer.currentVerseChanged.subscribe(verseChangedSpy);
+    const env = new TestEnvironment({ timings });
 
     env.audioPlayer.audio.currentTime = 1;
     env.audioPlayer.audio.timeUpdated$.next();
     env.wait();
     expect(env.currentTime).toBeGreaterThan(timings[1].from);
     expect(env.verseLabel.nativeElement.textContent).toEqual('Genesis 1:1');
-    expect(verseChangedSpy).toHaveBeenCalledWith('s_1');
+    expect(env.verseChangedSpy).toHaveBeenCalledWith('s_1');
 
     env.audioPlayer.audio.currentTime = 2.5;
     env.audioPlayer.audio.timeUpdated$.next();
     env.wait();
     expect(env.currentTime).toBeGreaterThan(timings[3].from);
     expect(env.verseLabel.nativeElement.textContent).toEqual('Genesis 1:2');
-    expect(verseChangedSpy).toHaveBeenCalledWith('s_2');
+    expect(env.verseChangedSpy).toHaveBeenCalledWith('s_2');
   }));
 
   it('emits section heading when timing starts at greater than 0 seconds', fakeAsync(() => {
-    const timings: AudioTiming[] = [
-      { textRef: 's', from: 1.0, to: 2.0 },
-      { textRef: '1', from: 2.0, to: 3.0 }
-    ];
-    env.component.audioPlayer.timing = timings;
-
-    const verseChangedSpy = jasmine.createSpy('verseChanged');
-    env.component.audioPlayer.currentVerseChanged.subscribe(verseChangedSpy);
-
-    env.playButton.nativeElement.click();
-    env.audioPlayer.audio.timeUpdated$.next();
-    env.wait();
-
-    expect(verseChangedSpy).toHaveBeenCalledWith('s_1');
+    const env = new TestEnvironment({
+      timings: [
+        { textRef: 's', from: 1.0, to: 2.0 },
+        { textRef: '1', from: 2.0, to: 3.0 }
+      ]
+    });
+    expect(env.verseChangedSpy).toHaveBeenCalledWith('s_1');
   }));
 
   it('pauses and emits on close', fakeAsync(() => {
-    const pauseSpy = spyOn(env.component.audioPlayer, 'pause').and.callThrough();
-    let count = 0;
-    env.component.audioPlayer.closed.subscribe(() => count++);
-    expect(pauseSpy).not.toHaveBeenCalled();
+    const env = new TestEnvironment();
+
+    expect(env.audioPauseSpy).not.toHaveBeenCalled();
 
     env.component.audioPlayer.close();
     env.wait();
-    expect(pauseSpy).toHaveBeenCalled();
-    expect(count).toEqual(1);
+    expect(env.audioPauseSpy).toHaveBeenCalled();
+    expect(env.closedEventCount).toEqual(1);
   }));
 
   it('skipping to previous verse remains on the current verse if within grace period', fakeAsync(() => {
-    env.component.audioPlayer.timing = [
-      { textRef: '1', from: 0.0, to: 1.0 },
-      { textRef: '2', from: 1.0, to: 4.5 },
-      { textRef: '3', from: 4.5, to: 5.0 }
-    ];
+    const env = new TestEnvironment({
+      timings: [
+        { textRef: '1', from: 0.0, to: 1.0 },
+        { textRef: '2', from: 1.0, to: 4.5 },
+        { textRef: '3', from: 4.5, to: 5.0 }
+      ]
+    });
 
     env.currentTime = 4.1;
     expect(env.currentTime).toBeGreaterThan(4);
@@ -161,10 +142,12 @@ describe('ScriptureAudioComponent', () => {
   }));
 
   it('skipping to the next verse will skip to the start of the current timing data if it has not started yet', fakeAsync(() => {
-    env.component.audioPlayer.timing = [
-      { textRef: '1', from: 3.0, to: 4.0 },
-      { textRef: '2', from: 4.0, to: 5.0 }
-    ];
+    const env = new TestEnvironment({
+      timings: [
+        { textRef: '1', from: 3.0, to: 4.0 },
+        { textRef: '2', from: 4.0, to: 5.0 }
+      ]
+    });
 
     expect(env.currentTime).toEqual(0);
     env.clickNextRef();
@@ -172,17 +155,15 @@ describe('ScriptureAudioComponent', () => {
   }));
 
   it('emits once when chapter audio finishes', fakeAsync(() => {
-    const pause = spyOn(env.component.audioPlayer, 'pause').and.callThrough();
-    let eventCount = 0;
-    env.component.audioPlayer.closed.subscribe(() => eventCount++);
+    const env = new TestEnvironment();
 
     env.component.audioPlayer.audioPlayer?.isAudioAvailable$.next(false);
     env.component.audioPlayer.audioPlayer?.isAudioAvailable$.next(true);
     env.component.audioPlayer.audioPlayer?.audio?.finishedPlaying$.emit();
     env.component.audioPlayer.audioPlayer?.audio?.finishedPlaying$.emit();
 
-    expect(eventCount).toBe(1);
-    expect(pause).toHaveBeenCalledTimes(1);
+    expect(env.closedEventCount).toBe(1);
+    expect(env.audioPauseSpy).toHaveBeenCalledTimes(1);
   }));
 });
 
@@ -205,7 +186,9 @@ class AudioPlayerStubComponent {
     this.audio = new AudioPlayerStub(audioFile, instance(AudioPlayerStubComponent.onlineStatusService));
   }
 
-  @Input() set source(source: string | undefined) {}
+  @Input() set source(source: string | undefined) {
+    if (source != null) this.isAudioAvailable$.next(true);
+  }
 }
 
 class TestEnvironment {
@@ -213,8 +196,14 @@ class TestEnvironment {
   fixture: ComponentFixture<HostComponent>;
   component: HostComponent;
   ngZone: NgZone;
+  verseChangedSpy: jasmine.Spy<jasmine.Func> = jasmine.createSpy('verseChanged');
+  audioPlaySpy?: jasmine.Spy<jasmine.Func>;
+  audioPauseSpy?: jasmine.Spy<jasmine.Func>;
+  closedEventCount: number = 0;
 
-  constructor(template: string) {
+  constructor({ timings }: Partial<{ timings: AudioTiming[] }> = {}) {
+    const template = `<app-checking-scripture-audio-player source="${audioFile}"></app-checking-scripture-audio-player>`;
+
     TestBed.configureTestingModule({
       declarations: [HostComponent, CheckingScriptureAudioPlayerComponent, AudioPlayerStubComponent, AudioTimePipe],
       providers: [{ provide: OnlineStatusService, useFactory: () => instance(this.mockOnlineStatusService) }],
@@ -225,6 +214,17 @@ class TestEnvironment {
     this.ngZone = TestBed.inject(NgZone);
     this.fixture = TestBed.createComponent(HostComponent);
     this.component = this.fixture.componentInstance;
+
+    this.fixture.detectChanges();
+
+    this.component.audioPlayer.timing = timings ?? getAudioTimings();
+    this.component.audioPlayer.textDocId = textDocId;
+    this.component.audioPlayer.currentVerseChanged.subscribe(this.verseChangedSpy);
+    this.audioPlaySpy = spyOn(this.audioPlayer.audio, 'play').and.callThrough();
+    this.audioPauseSpy = spyOn(this.audioPlayer.audio, 'pause').and.callThrough();
+    this.component.audioPlayer.closed.subscribe(() => this.closedEventCount++);
+
+    this.wait();
   }
 
   get audioPlayer(): AudioPlayerStubComponent {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.ts
@@ -16,9 +16,13 @@ import { AudioHeadingRef, AudioTextRef, CheckingUtils } from '../../checking.uti
   styleUrls: ['./checking-scripture-audio-player.component.scss']
 })
 export class CheckingScriptureAudioPlayerComponent extends SubscriptionDisposable implements AfterViewInit {
+  @Input() canClose: boolean = true;
   @Output() currentVerseChanged = new EventEmitter<string>();
   @Output() closed: EventEmitter<void> = new EventEmitter<void>();
   @ViewChild('audioPlayer') audioPlayer?: AudioPlayerComponent;
+  /** Having some text in the verse label (rather than empty string or a space) helps containing components predict the
+   * likely height. */
+  private readonly emptyVerseLabel: string = ':';
 
   @Input() set textDocId(value: TextDocId | undefined) {
     this._textDocId = value;
@@ -35,7 +39,7 @@ export class CheckingScriptureAudioPlayerComponent extends SubscriptionDisposabl
     this.doAudioSubscriptions();
   }
 
-  verseLabel: string = '';
+  verseLabel: string = this.emptyVerseLabel;
   audioSource?: string;
 
   private _timing: AudioTiming[] = [];
@@ -57,7 +61,7 @@ export class CheckingScriptureAudioPlayerComponent extends SubscriptionDisposabl
   }
 
   private get currentVerseLabel(): string {
-    if (this._textDocId == null) return '';
+    if (this._textDocId == null) return this.emptyVerseLabel;
     const currentTime: number = this.audioPlayer?.audio?.currentTime ?? 0;
     const currentVerseStr: string = this.getCurrentVerseStr(currentTime);
     const verseRef = new VerseRef(
@@ -182,7 +186,7 @@ export class CheckingScriptureAudioPlayerComponent extends SubscriptionDisposabl
   private subscribeToAudioFinished(audio: AudioPlayer): void {
     this.finishedSubscription?.unsubscribe();
     this.finishedSubscription = this.subscribe(audio.finishedPlaying$.pipe(first()), () => {
-      this.close();
+      if (this.canClose) this.close();
     });
   }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.ts
@@ -32,11 +32,7 @@ export class CheckingScriptureAudioPlayerComponent extends SubscriptionDisposabl
 
   @Input() set source(value: string | undefined) {
     this.audioSource = value;
-    this.finishedSubscription?.unsubscribe();
-    this.finishedSubscription = undefined;
-    if (value != null) {
-      this.subscribeToAudioFinished();
-    }
+    this.doAudioSubscriptions();
   }
 
   verseLabel: string = '';
@@ -45,14 +41,15 @@ export class CheckingScriptureAudioPlayerComponent extends SubscriptionDisposabl
   private _timing: AudioTiming[] = [];
   private _textDocId?: TextDocId;
   private finishedSubscription?: Subscription;
+  private verseChangeSubscription?: Subscription;
+  private audioSubscription?: Subscription;
 
   constructor(readonly i18n: I18nService) {
     super();
   }
 
   ngAfterViewInit(): void {
-    this.subscribeToAudioFinished();
-    this.subscribePlayerVerseChange(this.audioPlayer!.audio!);
+    this.doAudioSubscriptions();
   }
 
   get isPlaying(): boolean {
@@ -131,19 +128,40 @@ export class CheckingScriptureAudioPlayerComponent extends SubscriptionDisposabl
     return this._timing.findIndex(t => t.to > currentTime);
   }
 
-  private subscribePlayerVerseChange(audio: AudioPlayer): void {
-    this.subscribe(
+  private doAudioSubscriptions(): void {
+    if (this.audioPlayer == null) return;
+    // wait until the next microtask cycle to get the audio player with the updated source
+    Promise.resolve(this.audioPlayer).then(audioPlayer => {
+      this.audioSubscription?.unsubscribe();
+      this.audioSubscription = this.subscribe(
+        audioPlayer.isAudioAvailable$.pipe(
+          filter(a => a),
+          first()
+        ),
+        () => {
+          if (audioPlayer.audio == null) {
+            console.log(`warning: audio player unexpectedly null.`);
+            return;
+          }
+          const audio: AudioPlayer = audioPlayer.audio;
+          this.subscribeToAudioFinished(audio);
+          this.subscribeToVerseChange(audio);
+        }
+      );
+    });
+  }
+
+  private subscribeToVerseChange(audio: AudioPlayer): void {
+    this.verseChangeSubscription?.unsubscribe();
+    this.verseChangeSubscription = this.subscribe(
       audio.timeUpdated$.pipe(
         map(() => this.getRefIndexInTimings(audio.currentTime)),
         distinctUntilChanged()
       ),
       () => {
-        if (this._textDocId == null || this.audioPlayer?.audio == null) return;
+        if (this._textDocId == null) return;
         this.verseLabel = this.currentVerseLabel;
-        const audioTextRef: AudioTextRef | undefined = CheckingUtils.parseAudioRef(
-          this._timing,
-          this.audioPlayer.audio.currentTime
-        );
+        const audioTextRef: AudioTextRef | undefined = CheckingUtils.parseAudioRef(this._timing, audio.currentTime);
         if (audioTextRef?.verseStr == null) {
           // emit the current ref that is a section heading
           const audioHeadingRef: AudioHeadingRef | undefined = CheckingUtils.parseAudioHeadingRef(
@@ -161,21 +179,10 @@ export class CheckingScriptureAudioPlayerComponent extends SubscriptionDisposabl
     );
   }
 
-  private subscribeToAudioFinished(): void {
-    if (this.audioPlayer == null || this.finishedSubscription != null) return;
-    // wait until the next microtask cycle to get the audio player with the updated source
-    Promise.resolve(this.audioPlayer).then(audioPlayer => {
-      this.subscribe(
-        audioPlayer.isAudioAvailable$.pipe(
-          filter(a => a),
-          first()
-        ),
-        () => {
-          this.finishedSubscription = this.subscribe(audioPlayer.audio!.finishedPlaying$.pipe(first()), () => {
-            this.close();
-          });
-        }
-      );
+  private subscribeToAudioFinished(audio: AudioPlayer): void {
+    this.finishedSubscription?.unsubscribe();
+    this.finishedSubscription = this.subscribe(audio.finishedPlaying$.pipe(first()), () => {
+      this.close();
     });
   }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html
@@ -124,6 +124,7 @@
                 (dragEnd)="checkSliderPosition($event)"
                 [useTransition]="true"
               >
+                <!-- Splitter area sizes are controlled programmatically. -->
                 <as-split-area [size]="100">
                   <div class="panel-content" #scripturePanelContainer>
                     <!-- TODO (scripture audio) Instead of covering scripture panel, move the audio player below it -->
@@ -132,6 +133,7 @@
                         [source]="chapterAudioSource"
                         [timing]="chapterTextAudioTiming"
                         [textDocId]="textDocId"
+                        [canClose]="!hideChapterText"
                         (currentVerseChanged)="handleAudioTextRefChanged($event)"
                         (closed)="hideChapterAudio()"
                       ></app-checking-scripture-audio-player>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html
@@ -144,6 +144,7 @@
                       [questionVerses]="questionVerseRefs"
                       (questionVerseSelected)="verseRefClicked($event)"
                       [isRightToLeft]="isRightToLeft"
+                      [class.hidden]="hideChapterText"
                     ></app-checking-text>
                   </div>
                 </as-split-area>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss
@@ -267,3 +267,7 @@ $questionsPanelBreakpoint: 600px;
   z-index: 1;
   padding: 0.25rem;
 }
+
+.hidden {
+  display: none;
+}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
@@ -176,7 +176,7 @@ describe('CheckingComponent', () => {
 
   describe('Interface', () => {
     it('can navigate using next button', fakeAsync(() => {
-      const env = new TestEnvironment(ADMIN_USER);
+      const env = new TestEnvironment({ user: ADMIN_USER });
       env.selectQuestion(1);
       env.clickButton(env.nextButton);
       tick(env.questionReadTimer);
@@ -185,7 +185,7 @@ describe('CheckingComponent', () => {
     }));
 
     it('can navigate using previous button', fakeAsync(() => {
-      const env = new TestEnvironment(ADMIN_USER);
+      const env = new TestEnvironment({ user: ADMIN_USER });
       env.selectQuestion(2);
       env.clickButton(env.previousButton);
       tick(env.questionReadTimer);
@@ -194,7 +194,7 @@ describe('CheckingComponent', () => {
     }));
 
     it('check navigate buttons disable at the end of the question list', fakeAsync(() => {
-      const env = new TestEnvironment(ADMIN_USER);
+      const env = new TestEnvironment({ user: ADMIN_USER });
       env.selectQuestion(1);
       const prev = env.previousButton;
       const next = env.nextButton;
@@ -207,19 +207,19 @@ describe('CheckingComponent', () => {
     }));
 
     it('should open question dialog', fakeAsync(() => {
-      const env = new TestEnvironment(ADMIN_USER);
+      const env = new TestEnvironment({ user: ADMIN_USER });
       env.clickButton(env.addQuestionButton);
       verify(mockedQuestionDialogService.questionDialog(anything())).once();
       expect().nothing();
     }));
 
     it('hides add question button for community checker', fakeAsync(() => {
-      const env = new TestEnvironment(CHECKER_USER);
+      const env = new TestEnvironment({ user: CHECKER_USER });
       expect(env.addQuestionButton).toBeNull();
     }));
 
     it('responds to remote removed from project', fakeAsync(() => {
-      const env = new TestEnvironment(CHECKER_USER);
+      const env = new TestEnvironment({ user: CHECKER_USER });
       env.selectQuestion(1);
       expect(env.component.questionDocs.length).toEqual(15);
       env.component.projectDoc!.submitJson0Op(op => op.unset<string>(p => p.userRoles[CHECKER_USER.id]), false);
@@ -230,7 +230,7 @@ describe('CheckingComponent', () => {
     }));
 
     it('responds to remote community checking disabled when checker', fakeAsync(() => {
-      const env = new TestEnvironment(CHECKER_USER);
+      const env = new TestEnvironment({ user: CHECKER_USER });
       env.selectQuestion(1);
       const projectUserConfig = env.component.projectUserConfigDoc!.data!;
       expect(projectUserConfig.selectedTask).toEqual('checking');
@@ -244,7 +244,7 @@ describe('CheckingComponent', () => {
 
     it('responds to remote community checking disabled when observer', fakeAsync(() => {
       // User with access to translate app should get redirected there
-      const env = new TestEnvironment(OBSERVER_USER, 'ALL');
+      const env = new TestEnvironment({ user: OBSERVER_USER, projectBookRoute: 'ALL' });
       env.selectQuestion(1);
       env.setCheckingEnabled(false);
       expect(env.location.path()).toEqual('/projects/project01/translate/JHN');
@@ -256,7 +256,7 @@ describe('CheckingComponent', () => {
 
   describe('Questions', () => {
     it('questions are displaying and audio is cached', fakeAsync(() => {
-      const env = new TestEnvironment(CHECKER_USER);
+      const env = new TestEnvironment({ user: CHECKER_USER });
       verify(mockedFileService.findOrUpdateCache(FileType.Audio, QuestionDoc.COLLECTION, anything(), anything())).times(
         31
       );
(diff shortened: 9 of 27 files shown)


marksvc:
This is a simple change to add a checkbox to the Settings page to toggle whether Scripture text should be shown or not. It also introduces a field to the data model. It does not actually influence whether Scripture is shown or not in the Community Checking area. 

I wanted to make the checkbox only appear in Settings for projects that have Scripture audio. To do that, I ended up with a BehaviorSubject, and doing some RxJS in ngOnInit. Is this the right approach, or over-complicating things?

In an earlier version of the code I had a Settings component `projectHasAudio()` method. I since removed that, but left in the part of its test that checks if the checkbox is present or absent. This isn't working, but I need to look at whether I am constructing the right situation for ngOnInit to detect Scripture audio.

One thing that I don't have working are the Storybook tests. There are Stories for the permutations of whether the Scripture audio feature flag is on or off, and whether the project has Scripture audio. But it isn't working at present.

marksvc:
Video showing the checkbox on Settings page of a project with Scripture audio:

[hide-scr-checkbox.webm](https://github.com/sillsdev/web-xforge/assets/7265309/330b3522-7699-4ffe-98c0-8b1ba51bb77e)

marksvc:
I continued having trouble with Storybook and removed it from the PR.

It occurred to me that if the project administrator adds Scripture Audio, checks the box to hide Scripture text, and then removes their Scripture Audio, the "hide Scripture text" selection will still be set. I started to write some code to make the checkbox be shown again so it could be deselected, but stopped when I realized that a broader and better solution is to show Scripture text in instances where we do not _have_ Scripture audio for a given chapter.

josephmyers:
This may not be an issue when it's not behind the feature flag, but I'm seeing the green checkmark when enabling Scripture Audio. Further, it doesn't go away on page refresh. I'll keep testing
Uploading Recording 2023-08-22 084731.mp4…

josephmyers:
Based on the team meeting, I'm gonna pause my review on this while we finish up the rest of the work on this

marksvc:
That sounds good, thank you.

marksvc:
Videos showing behaviour on phone and wider:

[phone-audio-sizing-3.webm](https://github.com/sillsdev/web-xforge/assets/7265309/2c981665-c212-4006-be85-2d1631817680)

[desktop-audio-sizing-3.webm](https://github.com/sillsdev/web-xforge/assets/7265309/6dd10389-39ca-48ed-8db2-789a2a035aa5)

marksvc:
This PR is using `$any('*')` as a workaround until this PR is merged: https://github.com/angular-split/angular-split/pull/289
After that, just `'*'` can be used.

josephmyers:
@marksvc, I'm still seeing the issue with the green checkmark! I'll keep reviewing though

josephmyers (changes requested):
I really like this feature! It looks cool to take away the text and see only the chapter audio with the questions and answers. And our GUI adapts to the absence pretty well. It will be really interesting to see this once all the open PR's for scripture audio get merged, particularly the more usability-focused ones. In fact, I think we should wait til some of them are before merging this, so that we can make sure the workflow really clicks. For example, how does it work when we swap the Delete button for Close? How does this work when finishing the chapter hides the audio player?

I also like that you can minimize the player while listening to get more screen space. Good thinking!

I found a minor issue where if you enable the setting and delete all audio files, it breaks things. Plus, you can't uncheck the setting until you had audio back.


I'm also wondering if we need a place to indicate to admin's when audio projects have chapters that don't have audio yet. What happens if checkers log into these projects and there's no way to listen to or read the verses? Aren't they screwed? What do you think? Maybe that's a separate feature.

> Reviewed 17 of 18 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-:-Nd4EGza2yKZ9KwnQngM:b-hcllbl)* status: all files reviewed, 5 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 138 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4IaYW0B08jrC2Et5v:-Nd4IaYW0B08jrC2Et5w:bbd1oie) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L138)):*
> ```HTML
>                 unit="pixel"
>               >
>                 <as-split-area [size]="hideChapterText ? 109 : $any('*')" [maxSize]="hideChapterText ? 109 : null">
> ```

Is there any way we can not use magic numbers here? It makes this part brittle. For example, when the audio is unavailable/missing, it takes up too much space, since the player is shorter.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 125 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4JNlGB4dkg3Fv27hJ:-Nd4JNlGB4dkg3Fv27hK:b-c39gj5) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L125)):*
> ```SCSS
>     #answer-panel {
>       height: 100%;
>       padding-top: 12px;
> ```

I think this guy looks funky now. We probably still need it, just elsewhere?


___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 1178 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4Kj5S3fTxYdTzYZLD:-Nd4Kj5S3fTxYdTzYZLE:b-agnip) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L1178)):*
> ```XML
>   }
> 
>   private showOrHideScriptureText(): void {
> ```

This is misnamed (or missing functionality). It only shows the player, never hides it.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.html` line 130 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4M35fEtP_BMy4reVh:-Nd4M35fEtP_BMy4reVi:b-l5xrcs) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.html#L130)):*
> ```HTML
>                 </div>
>               </div>
>               <div class="tool-setting" *ngIf="featureFlags.scriptureAudio.enabled && hasTextAudioDocs">
> ```

I'm a little skeptical about the extra hasTextAudioDocs check here. Let's say they want to check this setting and then add their audio files. They will come here and not find anything. Thoughts?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 268 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4C8Cu5oHqFbV-pttP:-Nd4C8Cu5oHqFbV-pttQ:briz1sl) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L268)):*
> ```JSON
>     "translations_will_be_suggested": "Translations for words and phrases will be suggested as you edit Scripture based on the project or resource selected above.",
>     "users_can_share_the_project": "Allow users to invite people outside Paratext to the project. They will receive the same role as the user who invited them.",
>     "hide_community_checking_text": "Do not show Scripture text in Community Checking area. Checkers will need to listen to the Scripture audio.",
> ```

In my opinion, this is a little wordy. We could save some space by saying "Hide Scripture text" and moving the second sentence to a tooltip with app-info. I don't think we need to specify Community Checking again, since we're in the Community Checking section. What do you think?

marksvc (commented):
I don't know why there is trouble with your green checkbox. Here is a video showing how it's working on my machine. I wonder if it would be helpful for you to enable the Scripture audio feature before going to the Settings page; maybe that might influence the behaviour? We can play with it together on a video call and see if we can figure it out.

[See chat for video scr-audio-checkbox.webm]

> In fact, I think we should wait til some of them are before merging this, so that we can make sure the workflow really clicks. For example, how does it work when we swap the Delete button for Close? How does this work when finishing the chapter hides the audio player?

Okay. I think my preference would be to get it merged in without waiting. Then as we make changes like to the Delete button or hiding the audio player, it can be in light of this positioning and also the reality that text will be hidden sometimes.

> Plus, you can't uncheck the setting until you had audio back.

Right! That's something we'll have to decide how to handle. My thinking is, if a project does not have any audio files, but has hide-text enabled, then we should still show the hide-text checkbox so they can uncheck it. (Okay, as discussed later, I made the checkbox visible regardless.)

> I found a minor issue where if you enable the setting and delete all audio files, it breaks things.

Yeah. We'll need to decide what to do in situations where a project admin wants hide-text, but also doesn't have audio for a chapter. Probably there should be a graphic or something to indicate that audio is missing for that chapter.

I'll start on this now. It could go into a followup PR if we finish the review before I'm done.

> I'm also wondering if we need a place to indicate to admin's when audio projects have chapters that don't have audio yet. What happens if checkers log into these projects and there's no way to listen to or read the verses?

So here's some useful framing. My thinking before was that maybe we could show text for chapters that don't have audio (like even if hide-text was enabled). But Nathaniel's picture of the hide-text situation is that some language projects don't want to show the text at all. So I'm seeing the hide-text setting as an all-or-nothing setting. You either hide all the text, or you don't hide all the text. 

So, yes, I do think it would be useful to project admins to be able to see that the have audio uploaded for Mark 1-8 and 10-16, but are missing audio for Mark 9. But because of the above framing, if there is no audio for Mark 9, then in a hide-text situation there will be no Scripture for the checkers to work with until it's uploaded. (Rather than us trying to compensate by showing the text.)

As we play with the audio feature we may find that it's easy to upload all your chapters, or obvious to an admin if a chapter is missing. We can see.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-:-Nd6nkIvB-9Wrj4T6UW-:bjm2xwt)* status: all files reviewed, 5 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 138 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4IaYW0B08jrC2Et5v:-Nd6sHaC9T-uTeVyr-6p:b2fg2tq) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L138)):*

Is there any way we can not use magic numbers here? It makes this part brittle. For example, when the audio is unavailable/missing, it takes up too much space, since the player is shorter.
</blockquote></details>

Right. I would prefer to make it some how be set to the size of the height of the audio player, or perhaps the height of the visible contents. I wasn't successful at finding a way to do that, and so manually determined the required px needed. Do you have any suggestions?

Regarding when the audio is missing, I don't think we need to worry about that for this situation. We can design a no-audio-available thing to show. It can still be 109 px tall, and if there is no audio available, and hide-text is enabled, then it's not a chapter for checkers to work in. 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 125 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4JNlGB4dkg3Fv27hJ:-Nd6u7C03KWELgwxEdmh:bbtxgeg) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L125)):*
Regarding the white padding above "1 Answers": If I turn off hide-text and show the Scripture, I do see that the Scripture text does _not_ have a corresponding bottom padding. It might be that the question area padding at the top should be eliminated. That might be something to throw out to the team to get UI opinions on, and make a jira issue for.

> looks funky now

I don't think this is something new with this PR, is it? I see the behaviour when hide-text is off.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 1178 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4Kj5S3fTxYdTzYZLD:-Nd6vgDwAySDxK0jLavW:b-tjhvuy) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L1178)):*

This is misnamed (or missing functionality). It only shows the player, never hides it.
</blockquote></details>

Ahh. But consider! In `ngOnInit()`, we do an initial showOrHideScriptureText(). Then we listen to `this.projectDoc.remoteChanges$`, upon which we again call showOrHideScriptureText(). 

That way the display in the community checking area will be dynamically updated in response to project settings changes regarding hide-text. For example, if you open one window and go to a chapter with audio. And you open another window and go to Settings. Then you select hide-text in the Settings window, the Checking window will instantly update by hiding the text and showing the audio player. If in Settings you clear hide-text, then in the Checking window, the text is again revealed (but it doesn't close the open audio player).

See the below video for a demonstration.

So I would say the method does cause the text to be hidden or shown. And there is a bonus effect of showing the audio player if hide-text is enabled.

But maybe there's something I can improve so it's more evident?

[see slack for hide-dynamic.webm]
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.html` line 130 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4M35fEtP_BMy4reVh:-Nd6xMr108d2D-0co6N3:bprmy9g) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.html#L130)):*

I'm a little skeptical about the extra hasTextAudioDocs check here. Let's say they want to check this setting and then add their audio files. They will come here and not find anything. Thoughts?
</blockquote></details>

Hmm. If a user has a text project and no Audio files, my thinking was that I don't want them to go turn off showing their Scripture text. And talking about hiding Scripture text and showing audio might not even make sense if they are all about thinking of showing the text and people engaging with the text. 

That said, (1) making the hide-text checkbox sometimes show and sometimes not show might be less desirable for support and help documentation. And (2) if they admin goes and turns off Scripture text, and the text disappears, that's what they told it to do and they can come turn it back on again. 

So maybe we should just have the checkbox be visible all the time. I'll go ahead and do that.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 268 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4C8Cu5oHqFbV-pttP:-Nd7G2mA4ujy2Ngq0lX6:b-nqk4nt) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L268)):*

In my opinion, this is a little wordy. We could save some space by saying "Hide Scripture text" and moving the second sentence to a tooltip with app-info. I don't think we need to specify Community Checking again, since we're in the Community Checking section. What do you think?
</blockquote></details>

Good point. We could fit in both thoughts if we use "Hide Scripture text. Checkers will listen to audio." What do you think about that?

marksvc (commented):
Okay, check it out, if there is no audio for a chapter, then part of the player is shown, but also "Audio unavailable". The cod eto do this was already there, but there was a particular situation that we were hitting where we only saw the back,play,forward,trash buttons, and not the "Audio unavailable" message. I adjusted the code so that it more consistently shows this message when there is no audio. The player is not 109 px high. I'm not sure that's a big deal. And if a user gets this player showing on their screen in a situation where they _do_ want to be looking at Scripture text, it may be good for it not to be taller. Do you think it's significant for the "Audio unavailable" player to be matching the full height of the text+audio split area?

See video of behaviour:
[audio-unavail-after-delete.webm](https://files.reviewableusercontent.io/images/reviewable/7265309/w1z1xyIX1JzJE4vp4QHJhaol/audio-unavail-after-delete.webm)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-:-Nd7_GfY4WyRlgItCJtQ:b-1zwh1w)* status: 13 of 19 files reviewed, 9 unresolved discussions (waiting on @josephmyers)

marksvc:
Fixed some code scanning alerts.

marksvc:
@josephmyers Chromatic shows that stories "Checking Audio / No Source" and "Audio Player Base / No Source" now show "Audio unavailable." instead of not showing that, with the changes in this PR. It looks like you authored audio-player.stories.ts, so you will be a good person to comment on whether that is good or not. (I think it seems good :), but I may not have understood some context or intent.) Thanks!

josephmyers (changes requested):
I'm fine giving a pass on the green checkmark. I'm not seeing it today, and I think it's just a first-time issue.

> Probably there should be a graphic or something to indicate that audio is missing for that chapter.

Yes, and I think we could justify splitting it out to its own task. I think to do it well it will warrant more thought and maybe more than a graphic somewhere. I followed up with Nathaniel on this.

> Reviewed 4 of 5 files at r3, 1 of 1 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-:-Nd86lHu7akbxaQfNAJw:baso1ny)* status: all files reviewed, 9 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 138 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4IaYW0B08jrC2Et5v:-NdSv1DAApGKniXBJe-0:b-woy2do) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L138)):*

Right. I would prefer to make it some how be set to the size of the height of the audio player, or perhaps the height of the visible contents. I wasn't successful at finding a way to do that, and so manually determined the required px needed. Do you have any suggestions?

Regarding when the audio is missing, I don't think we need to worry about that for this situation. We can design a no-audio-available thing to show. It can still be 109 px tall, and if there is no audio available, and hide-text is enabled, then it's not a chapter for checkers to work in.
</blockquote></details>

It should be possible, since we're already autosizing it for questions and answers. See calculateScriptureSliderPosition() in the .ts for examples of us doing this. We just need to get the height of the player (after it's loaded) and then set the sizes on the splitComponent. But you're probably more familiar with that code than I am!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 125 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4JNlGB4dkg3Fv27hJ:-Nd8Z2nV1gU8bt3LdxoQ:b-le13rt) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L125)):*

Regarding the white padding above "1 Answers": If I turn off hide-text and show the Scripture, I do see that the Scripture text does _not_ have a corresponding bottom padding. It might be that the question area padding at the top should be eliminated. That might be something to throw out to the team to get UI opinions on, and make a jira issue for.

> looks funky now

I don't think this is something new with this PR, is it? I see the behaviour when hide-text is off.
</blockquote></details>

That's a good point that you didn't introduce this; however, I do think it's a good improvement, and we're changing this area of the code already. We want to keep the padding, but not force it to be onscreen all the time. This is especially important for mobile. All we need to do is remove the "height: 100%" above.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 1178 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4Kj5S3fTxYdTzYZLD:-Nd8OfQjC0vyIct9slQt:b-pe210b) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L1178)):*

Ahh. But consider! In `ngOnInit()`, we do an initial showOrHideScriptureText(). Then we listen to `this.projectDoc.remoteChanges$`, upon which we again call showOrHideScriptureText(). 

That way the display in the community checking area will be dynamically updated in response to project settings changes regarding hide-text. For example, if you open one window and go to a chapter with audio. And you open another window and go to Settings. Then you select hide-text in the Settings window, the Checking window will instantly update by hiding the text and showing the audio player. If in Settings you clear hide-text, then in the Checking window, the text is again revealed (but it doesn't close the open audio player).

See the below video for a demonstration.

So I would say the method does cause the text to be hidden or shown. And there is a bonus effect of showing the audio player if hide-text is enabled.

But maybe there's something I can improve so it's more evident?

[see slack for hide-dynamic.webm]
</blockquote></details>

I understand that the overall behavior is there. I'm just saying that the showOrHideScriptureText() method isn't responsible for showing and hiding. It's responsible for conditionally showing -- the hiding comes from the fact that the player is hidden by default. This method is more of a "maybeShowScriptureText()" :smile: 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts` line 55 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd8aG_hDI_Ty-o632kX:-Nd8aG_hDI_Ty-o632kY:bc058fn) ([raw file](https://github.com/sillsdev/web-xforge/blob/6bdcab95db428b4ac6b0e8ad1f506f02687b25ff/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts#L55)):*
> ```XML
>       this._audio = undefined;
>       // We don't have audio to play, but we are finish initializing and the component can now say audio is unavailable.
>       this._isAudioInitComplete = true;
> ```

This is a fine addition to this class. It looks like they just missed this use case. Can you add a unit test for it? Maybe in the scripture audio player? Thanks
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 268 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4C8Cu5oHqFbV-pttP:-Nd8auasBynaZFKdok2m:b-p7bli8) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L268)):*

Good point. We could fit in both thoughts if we use "Hide Scripture text. Checkers will listen to audio." What do you think about that?
</blockquote></details>

I'm fine with it, but I liked your original second sentence better. And using the app-info for the second sentence really works, because it sticks to the UX paradigm of always displaying the label while providing supplemental information when desired. There, we are free to put as much explanation as we need, without worrying about space. You could say something like, "Checkers will be required to listen to Scripture audio." But I'm good with whatever you want to do!

marksvc (commented):
Check out how it works now. There was already an "Audio unavailable" that could be shown, it just needed some coaxing to consistently appear in certain cases. 

Video showing behaviour: 
[player-height-adjusts-unavailable.webm](https://files.reviewableusercontent.io/images/reviewable/7265309/lRqmv18bf_pP_V2VmFMcP6NX/player-height-adjusts-unavailable.webm)

BTW,
You'll notice that although moving between questions in chapter 1 does not have a flicker, there is a flicker when moving between questions in different chapters. I think the flicker is new, and it didn't used to do it before, and the reason was because there was essentially a logic error in audio-player-base.component.ts where once the audio player was initialized for one chapter, it was considered initialized forever even if it was fed different or even bad data subsequently. I have modified audio-player-base.component.ts so whenever we feed it new data, it starts off by resetting its is-initialized status, and doesn't specify that it's initialized for a bit until it loads up its data (or fails to load up its data). 

Although this logic seems a bit more correct I don't like the flicker that happens in the audio player when moving between chapters :-). Two thoughts on this are:
(1) Users will usually be moving between questions in the _same_ chapter, and flicker when moving to another chapter may not be very significant.
(2) When the audio player is still loading up its data, it could show one of those temporary grey rectangles ([skeleton screen](https://uxdesign.cc/what-you-should-know-about-skeleton-screens-a820c45a571a)). Then the audio player would be the same height as when fully loaded, and so the flicker would be greatly reduced. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-:-NdlQEhv37W0lhUdOi26:b3znb6)* status: all files reviewed, 9 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 138 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4IaYW0B08jrC2Et5v:-NdlEF-G7j6wTanH67W-:b8jg6sn) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L138)):*

It should be possible, since we're already autosizing it for questions and answers. See calculateScriptureSliderPosition() in the .ts for examples of us doing this. We just need to get the height of the player (after it's loaded) and then set the sizes on the splitComponent. But you're probably more familiar with that code than I am!
</blockquote></details>

Thank you. I was able to use a similar approach to get the audio player height, and set the splitter area height. I tried to find ways to fetch and apply heights after certain activity had taken place, rather than use setTimeout() or delay(), but in the end I used delay() to wait for the audio player to finish populating its various pieces and get its final height set.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 125 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4JNlGB4dkg3Fv27hJ:-NdlKLWmBl6L-qdoLwEh:bie7ggz) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L125)):*

That's a good point that you didn't introduce this; however, I do think it's a good improvement, and we're changing this area of the code already. We want to keep the padding, but not force it to be onscreen all the time. This is especially important for mobile. All we need to do is remove the "height: 100%" above.
</blockquote></details>

Done! I made it in a separate PR at #2047.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 1178 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4Kj5S3fTxYdTzYZLD:-NdlKvxJ6W_GTR6EerPb:by1roup) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L1178)):*
We might need to chat on the phone for this one :)

> It only shows the player, never hides it.

> isn't responsible for showing and hiding. It's responsible for conditionally showing -- the hiding comes from the fact that the player is hidden by default.

It sounds here like the method is being thought of as being for showing the _audio player_. I agree that this method only _shows_ the audio player, it never _hides_ the audio player. 

But the method does at times show the Scripture text, and at other times hides the Scripture text, which I captured in its name.

The method's primary function is to hide the Scripture text.

Oh, it might be useful to point out that showing the audio player isn't _how_ we hide the Scripture text. The Scripture text is hidden, as can be seen by dragging the slider down, or clicking the upper-right toolbar Play button (as shown in video [text-is-hidden.webm](https://files.reviewableusercontent.io/images/reviewable/7265309/w3IrDm4N66YpDpUnDo_hFenh/text-is-hidden.webm) ).

I think we may need to connect on the phone. Maybe after standup meeting?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts` line 55 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd8aG_hDI_Ty-o632kX:-NdlUeeB-KA22aeOPKLB:buyipg) ([raw file](https://github.com/sillsdev/web-xforge/blob/6bdcab95db428b4ac6b0e8ad1f506f02687b25ff/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts#L55)):*

This is a fine addition to this class. It looks like they just missed this use case. Can you add a unit test for it? Maybe in the scripture audio player? Thanks
</blockquote></details>

(Haven't gotten to this one yet. Pushing what I have for now tho.)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 268 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4C8Cu5oHqFbV-pttP:-NdlPUB80MB-Ntc9vRWo:b6e2c9p) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L268)):*

I'm fine with it, but I liked your original second sentence better. And using the app-info for the second sentence really works, because it sticks to the UX paradigm of always displaying the label while providing supplemental information when desired. There, we are free to put as much explanation as we need, without worrying about space. You could say something like, "Checkers will be required to listen to Scripture audio." But I'm good with whatever you want to do!
</blockquote></details>

How about this.

josephmyers (changes requested):
Yeah, I agree, the flicker indicates we could improve the default GUI state. A skeleton screen makes sense, but it may be easier, and very close to a skeleton screen, to default the player to Audio Unavailable. I believe right now it waits til it receives a status update to show anything there. What do you think?

> Reviewed 11 of 11 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-:-Ne1Z7MQ2YDJhmvEXyaO:bo7x2wd)* status: all files reviewed, 8 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 138 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4IaYW0B08jrC2Et5v:-Ne1UlNZ8KB_FdE7Xrn6:bwe2uc9) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L138)):*

Thank you. I was able to use a similar approach to get the audio player height, and set the splitter area height. I tried to find ways to fetch and apply heights after certain activity had taken place, rather than use setTimeout() or delay(), but in the end I used delay() to wait for the audio player to finish populating its various pieces and get its final height set.
</blockquote></details>

Sounds good. Thanks for going back and reworking it. It's too bad the split-area component doesn't support this (and doesn't intend to), then we wouldn't need all this craziness.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 125 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4JNlGB4dkg3Fv27hJ:-Ne1LIXe6QjSHBipmeIO:bgmzzzj) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L125)):*

Done! I made it in a separate PR at #2047.
</blockquote></details>

Sounds good. Thanks
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 1178 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4Kj5S3fTxYdTzYZLD:-Ne1WA2N79YPc16yNAsd:b9cnz78) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L1178)):*

We might need to chat on the phone for this one :)

> It only shows the player, never hides it.

> isn't responsible for showing and hiding. It's responsible for conditionally showing -- the hiding comes from the fact that the player is hidden by default.

It sounds here like the method is being thought of as being for showing the _audio player_. I agree that this method only _shows_ the audio player, it never _hides_ the audio player. 

But the method does at times show the Scripture text, and at other times hides the Scripture text, which I captured in its name.

The method's primary function is to hide the Scripture text.

Oh, it might be useful to point out that showing the audio player isn't _how_ we hide the Scripture text. The Scripture text is hidden, as can be seen by dragging the slider down, or clicking the upper-right toolbar Play button (as shown in video [text-is-hidden.webm](https://files.reviewableusercontent.io/images/reviewable/7265309/w3IrDm4N66YpDpUnDo_hFenh/text-is-hidden.webm) ).

I think we may need to connect on the phone. Maybe after standup meeting?
</blockquote></details>

Ok, I see what you're saying. I think I wasn't fully understanding everything that was going on here. Thanks for clarifying!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/single-button-audio-player/single-button-audio-player.component.ts` line 16 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Ne1WseT0UO1_cEVtCZ_:-Ne1WseT0UO1_cEVtCZa:b-buy29g) ([raw file](https://github.com/sillsdev/web-xforge/blob/ed14047a9da6ea8584d5ca5c8afaddf7e473b57e/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/single-button-audio-player/single-button-audio-player.component.ts#L16)):*
> ```XML
>   private _source?: string;
>   // Expose enum to template.
>   readonly AudioStatus = AudioStatus;
> ```

I like that you're trying to get away from magic strings in the template, but this solution seems weird. Is this just a limitation of the template?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts` line 55 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd8aG_hDI_Ty-o632kX:-Ne1Xq9-3yX80GGpiTz-:bnyyv85) ([raw file](https://github.com/sillsdev/web-xforge/blob/6bdcab95db428b4ac6b0e8ad1f506f02687b25ff/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts#L55)):*

(Haven't gotten to this one yet. Pushing what I have for now tho.)
</blockquote></details>

Ok, thanks. Were you able to get to this?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts` line 57 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Ne1Y5tWB-v4uRkkLRAv:-Ne1Y5tWB-v4uRkkLRAw:b-zd2ys7) ([raw file](https://github.com/sillsdev/web-xforge/blob/ed14047a9da6ea8584d5ca5c8afaddf7e473b57e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts#L57)):*
> ```XML
>       // 200 ms is a speculative value.
>       const timeoutMs = 200;
>       setTimeout(() => {
> ```

Is there any way to avoid this? Looks like the problem is that, when the source gets unset, the GUI needs some time between isAudioInitComplete = false and isAudioInitComplete = true. What if in this case, we never set isAudioInitComplete to false? Would that help?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 268 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd4C8Cu5oHqFbV-pttP:-Ne1Ys-8CCVXzFmYiYk_:b-5gwryt) ([raw file](https://github.com/sillsdev/web-xforge/blob/45067aaa73561c03550eb0b816061277674d9c8b/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L268)):*

How about this.


</blockquote></details>

Fantastic!

marksvc (commented):
I was rearranging parts of the audio player and it occurred to me that making the audio player appear automatically, and making the slider shrink up higher, are helpful changes on top of what SF-2028 was actually asking for. So I trimmed off some of these pieces and kept the basics of hiding text when the project admin requests it, and am updating the PR with that. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-:-NeFbC24DyWg_OzbY3t4:b-wahwab)* status: all files reviewed, 8 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/single-button-audio-player/single-button-audio-player.component.ts` line 16 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Ne1WseT0UO1_cEVtCZ_:-NeFc4ab6x4xsx6bhy5O:bo2a015) ([raw file](https://github.com/sillsdev/web-xforge/blob/ed14047a9da6ea8584d5ca5c8afaddf7e473b57e/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/single-button-audio-player/single-button-audio-player.component.ts#L16)):*

I like that you're trying to get away from magic strings in the template, but this solution seems weird. Is this just a limitation of the template?
</blockquote></details>

Yes! The template can't seem to use the enum without special help. There are various [solutions](https://stackoverflow.com/a/59289208).

josephmyers (commented):
Ok, but where did all that work go? This feature seems incomplete now. I'm sure you're aware we should only split something out if what remains can stand on its own, and this isn't releasable without auto-sizing, in my opinion. Your code was good -- put it back :D

With regards to having to more than we thought and splitting it off, this isn't necessary in many cases. Work often involves more than we expect, as new tasks always come with unknown. We're taking the opportunity to improve things that are necessary to the addition of this feature. Nothing out of scope so far.

The only thing I would leave off is the setTimeout in the audio-player-base-component. We talked about this last week as a team, and I thought they gave pretty good guidance on how to address that.

> Reviewed 3 of 11 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-:-Neb8u-oC7L6sg7t5L1J:bpmhky2)* status: 15 of 23 files reviewed, 6 unresolved discussions (waiting on @marksvc)

marksvc:
The splitter now shrinks the top up to just the audio player, when hide-text. 
The Checking (parent) component now tells the audio player whether the audio is unavailable.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1967)* status: 10 of 25 files reviewed, 7 unresolved discussions (waiting on @github-code-scanning[bot] and @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 13 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/1967#gh-1329076731:-Nednqi32R6BHhO9Jl_q:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ddfd7f2d8166d9cc4aa4767f5afc4918ac1bad7/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L13)):*

## Unused variable, import, function or class

Unused import AfterViewChecked.

[Show more details](https://github.com/sillsdev/web-xforge/security/code-scanning/640)
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts` line 7 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1967#gh-1310846137:-Nednv-WBqO5mfj7WwIN:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/66709334dd17ae3212b1e71c9465925ef259a4bf/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts#L7)):*

## Unused variable, import, function or class

Unused import from.

[Show more details](https://github.com/sillsdev/web-xforge/security/code-scanning/629)
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts` line 8 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1967#gh-1310846139:-NednvdwC9GWsZy7WZ2-:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/66709334dd17ae3212b1e71c9465925ef259a4bf/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts#L8)):*

## Unused variable, import, function or class

Unused import mergeMap.

[Show more details](https://github.com/sillsdev/web-xforge/security/code-scanning/630)
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts` line 17 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1967#gh-1310846143:-Nednwkl6akaXhLK5d38:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/66709334dd17ae3212b1e71c9465925ef259a4bf/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts#L17)):*

## Unused variable, import, function or class

Unused import RealtimeQuery.

[Show more details](https://github.com/sillsdev/web-xforge/security/code-scanning/631)
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts` line 23 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/1967#gh-1310846144:-NednxEk8Vb1Dj2AJgX-:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/66709334dd17ae3212b1e71c9465925ef259a4bf/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts#L23)):*

## Unused variable, import, function or class

Unused import TextAudioDoc.

[Show more details](https://github.com/sillsdev/web-xforge/security/code-scanning/632)
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts` line 57 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Ne1Y5tWB-v4uRkkLRAv:-NedoPR907dgHarawvsu:bdrur6k) ([raw file](https://github.com/sillsdev/web-xforge/blob/ed14047a9da6ea8584d5ca5c8afaddf7e473b57e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts#L57)):*

Is there any way to avoid this? Looks like the problem is that, when the source gets unset, the GUI needs some time between isAudioInitComplete = false and isAudioInitComplete = true. What if in this case, we never set isAudioInitComplete to false? Would that help?
</blockquote></details>

It seemed like not setting isAudioInitComplete to false isn't properly adjusting the state between switching sources. I would have to go back and play with it to say for sure. 

How about the new design?

marksvc:
Regarding the recent commit, "Don't set audio time until loaded":

We set our HTMLAudioElement to have an inflated currentTime when loading up, to work around another bug. This wasn't a problem before, because the existing/old code would flash "Audio unavailable" when switching to a question in a different chapter. Now it doesn't and so we can see that it flashes a high current time on the slider. The recent commit fixes this behaviour so it does not flash.

marksvc:
@josephmyers: Regarding the code you showed me: I am not too excited to keep track of the audio player's previous height. But I do like that the approach doesn't involve telling the audio-player-base whether `isSourceUnavailable`.

I tried another avenue today for consideration. In branch `task/sf-2028-parent-not-show`, I have a small change that conditionally does or does not allow the audio player into the DOM, depending on whether the CheckingComponent thinks the audio is available or not. If CheckingComponent believes the audio to be unavailable, it instead displays an "Audio unavailable" message itself. You can see the behaviour from the rough implementation in this video:

[parent-audio-unavailable.webm](https://github.com/sillsdev/web-xforge/assets/7265309/b0b65af5-c36a-46d5-9c04-022041861355)

I'm concerned that these remaining aspects will be further delaying the PR from getting merged. What do you think about at least getting these 5 commits merged in, and dealing with the flashing "Audio unavailable" message can happen in followup conversation and code?

```
  Shrink Scripture audio area up if hide text
  Show/hide Scripture text in response to setting
  SF-2028 Add setting to hide community checking text
  dev: test: refactor: allow custom project
  dev: test: refactor: arguments to destructured object
```

marksvc:
I pushed an update of the branch, but I haven't fully gotten things sorted out from a recent rebase on master, so some more adjustment will need to be done. (Though I don't want to do too much more with that until I hear back from you.)

josephmyers:
> I'm concerned that these remaining aspects will be further delaying the PR from getting merged. What do you think about at least getting these 5 commits merged in, and dealing with how and when to show the "Audio unavailable" message can happen in followup conversation and code?
> 
> ```
>   Shrink Scripture audio area up if hide text
>   Show/hide Scripture text in response to setting
>   SF-2028 Add setting to hide community checking text
>   dev: test: refactor: allow custom project
>   dev: test: refactor: arguments to destructured object
> ```

I agree we shouldn't drag this feature out. It is a tough problem, and I feel like I'm so far in the weeds I could be missing an obvious alternative. Simply suppressing the audio player seems like a simple yet solid approach. Do the five commits include that? What's the best way for me to test them out?

marksvc:
I'll throw in the commit about "Don't set audio time until loaded" as well so we don't have a green timing slider flicker.

Have you noticed that when you first load the page, you get an error saying the following?

> Cannot read properties of undefined (reading 'timeUpdated$')

This started happening after I rebased on commit "SF-2231 Stabilized and refactored inconsistent frontend audio tests (#2061)". I haven't looked into solving it yet, though I see that the code where it is getting the error (`this.subscribePlayerVerseChange(this.audioPlayer!.audio!);`) was moved into `ngAfterViewInit()` from the `play()` method. So presumably the code is getting called too soon, before the required pieces are existing. So a solution may be to do something like putting something in ngAfterViewInit() that will watch for `this.audioPlayer.audio` to be non-null, and then do the subscribe. But I haven't examined the rest of the commit to understand yet why the code was moved into `ngAfterViewInit`. 
If it's fresh in your mind and apparent what the trouble is, I am glad to listen to a tip or direction before I start looking into it.

I will update this PR to omit the "Make parent govern showing Audio unavailable message" commit, whose implementation is under discussion.

marksvc:
I see there are yet more conflicts. I plan to work on those tomorrow.

marksvc:
The commit "audio: subscribe when targets are ready" fixes the following error
> Cannot read properties of undefined (reading 'timeUpdated$')

The PR is ready for review. It will benefit from some followup regarding showing "Audio unavailable" when there is no audio.

marksvc:
This does not yet include setting maxSize, which I hope can be done in followup.

josephmyers (changes requested):
> Reviewed 2 of 11 files at r8, 8 of 13 files at r11, 3 of 6 files at r12, 2 of 9 files at r13, 4 of 7 files at r14, 1 of 1 files at r15, 12 of 12 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1967)* status: all files reviewed, 8 unresolved discussions (waiting on @github-code-scanning[bot] and @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts` line 19 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nf9Pgbv0nkDan8TShXG:-Nf9Pgbv0nkDan8TShXH:b-c7zt5o) ([raw file](https://github.com/sillsdev/web-xforge/blob/ed14047a9da6ea8584d5ca5c8afaddf7e473b57e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts#L19)):*
> ```XML
> 
>     this.subscribe(this.isAudioAvailable$, () => {
>       this.audio?.setSeek(0);
> ```

It looks like you moved this to only happen when the audio becomes available. What was your reasoning for changing it? (We'll need to make sure we unit test both cases, since this is of course a behavior/API change.)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts` line 12 at r16](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nf9Oh5-1-yqYZGG7hvR:-Nf9Oh5-1-yqYZGG7hvS:b-c0dss) ([raw file](https://github.com/sillsdev/web-xforge/blob/c5a6bb5537c2e2a41a66377ca83ea3ad2c2fa191/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts#L12)):*
> ```XML
> export abstract class AudioPlayerBaseComponent extends SubscriptionDisposable implements OnDestroy {
>   readonly isAudioAvailable$: BehaviorSubject<boolean> = new BehaviorSubject<boolean>(false);
>   hasProblem: boolean = false;
> ```

Let's make this guy private with a public accessor. We want only the class itself to be able to control changing this property, as opposed to creating classes coming in and changing it. Overall, I think this more holistic property is a step in the right direction. Good job!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.spec.ts` line 48 at r16](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nf9QSuO2xafEIOvYVKd:-Nf9QSuO2xafEIOvYVKe:b6an3d0) ([raw file](https://github.com/sillsdev/web-xforge/blob/c5a6bb5537c2e2a41a66377ca83ea3ad2c2fa191/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.spec.ts#L48)):*
> ```XML
>   }));
> 
>   it('sets isAudioInitComplete when status changes from Init', fakeAsync(() => {
> ```

Don't forget the test name here

josephmyers:
Oh, and don't forget to create those split-off tasks, whenever you can! Thanks

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1967)* status: all files reviewed, 8 unresolved discussions (waiting on @github-code-scanning[bot] and @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts` line 55 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nd8aG_hDI_Ty-o632kX:-NfD-65JCgTQRmsoJ8Xh:b-o6t172) ([raw file](https://github.com/sillsdev/web-xforge/blob/6bdcab95db428b4ac6b0e8ad1f506f02687b25ff/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts#L55)):*

Ok, thanks. Were you able to get to this?
</blockquote></details>

Okay, I think this is done now by "it('clears audio when source is cleared'" in audio-player-base.component.spec.ts, if I'm looking at the right thing.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts` line 19 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nf9Pgbv0nkDan8TShXG:-NfCyHXG69z6x456vyRT:b-3927qw) ([raw file](https://github.com/sillsdev/web-xforge/blob/ed14047a9da6ea8584d5ca5c8afaddf7e473b57e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts#L19)):*

It looks like you moved this to only happen when the audio becomes available. What was your reasoning for changing it? (We'll need to make sure we unit test both cases, since this is of course a behavior/API change.)
</blockquote></details>

My intent was to simplify the code, as I could just set seek to 0 in `set audio` in `if (newVal \=== AudioStatus.Available) {`. But now that I look more closely, I see that was leaving out situations with an invalid source or where an error has occurred. I put the subscription back in the constructor.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts` line 12 at r16](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nf9Oh5-1-yqYZGG7hvR:-NfCxlq7CnrMfK_RFkfV:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/c5a6bb5537c2e2a41a66377ca83ea3ad2c2fa191/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.ts#L12)):*

Let's make this guy private with a public accessor. We want only the class itself to be able to control changing this property, as opposed to creating classes coming in and changing it. Overall, I think this more holistic property is a step in the right direction. Good job!
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.spec.ts` line 48 at r16](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-Nf9QSuO2xafEIOvYVKd:-NfD-due4Gz3HL0CT56V:b-vuj8ut) ([raw file](https://github.com/sillsdev/web-xforge/blob/c5a6bb5537c2e2a41a66377ca83ea3ad2c2fa191/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player-base/audio-player-base.component.spec.ts#L48)):*

Don't forget the test name here
</blockquote></details>

Thank you :)

josephmyers (approved):
There's assuredly more work to be done on this, but what you have looks good. Thanks for sticking with it!

> Reviewed 6 of 6 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-:-NfEB-rk5kUnyAukLgUe:bmxtbil)* status: all files reviewed, 5 unresolved discussions (waiting on @github-code-scanning[bot])

marksvc (commented):
Thanks for looking at the design with me!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1967#-:-NfITkIdDTVm0JXw2q-z:b-qe32px)* status: all files reviewed, 5 unresolved discussions (waiting on @github-code-scanning[bot])

marksvc:
@josephmyers A change to checking-scripture-audio-player.component.spec happened in the mean time, and I needed to make an adjustment to the PR.

The first commit, `audio: subscribe when targets are ready` adjusts checking-scripture-audio-player so that some audio subscriptions happen after `audioPlayer.isAudioAvailable$`. 

With that new arrangement, checking-scripture-audio-player has an initial timing (from 
```ts
beforeEach(
...
  env.component.audioPlayer.timing = ...
```
). When test "emits section heading when timing starts at greater than 0 seconds" runs, it again sets `env.component.audioPlayer.timing`, but _not before the verse change subscription has already run its contents on the original timing data_. 

I modified the checking-scripture-audio-player.component.spec.ts so that we can feed in the desired timing values when we create the test environment. It starts off with the timing values we are going to use. In the test now it starts off with the correct timing values, and emits the expected reference right away (not necessarily when 'play' is clicked). 

Unfortunately, this led to a problem with test "emits verse changed event". I modified that test in light of there already being an emit that happened.

I wanted to run this change thru the review process since it wasn't just a simple adjustment. 

@RaymondLuong3 I see that you wrote the "emits section heading when timing starts at greater than 0 seconds" test. Should I further adjust what I have done? The test now is a little odd since it doesn't really need the play button to be clicked at this time. 
Right now the test reads,
```ts
    const env = new TestEnvironment({
      timings: [
        { textRef: 's', from: 1.0, to: 2.0 },
        { textRef: '1', from: 2.0, to: 3.0 }
      ]
    });
    const verseChangedSpy = jasmine.createSpy('verseChanged');
    env.component.audioPlayer.currentVerseChanged.subscribe(verseChangedSpy);
    env.wait();

    env.playButton.nativeElement.click();
    env.audioPlayer.audio.timeUpdated$.next();
    env.wait();

    expect(verseChangedSpy).toHaveBeenCalledWith('s_1');
```
But it runs just as well as
```ts
    const env = new TestEnvironment({
      timings: [
        { textRef: 's', from: 1.0, to: 2.0 },
        { textRef: '1', from: 2.0, to: 3.0 }
      ]
    });
    const verseChangedSpy = jasmine.createSpy('verseChanged');
    env.component.audioPlayer.currentVerseChanged.subscribe(verseChangedSpy);
    env.wait();

    expect(verseChangedSpy).toHaveBeenCalledWith('s_1');

    env.playButton.nativeElement.click();
    env.audioPlayer.audio.timeUpdated$.next();
    env.wait();
```
or even
```ts
    const env = new TestEnvironment({
      timings: [
        { textRef: 's', from: 1.0, to: 2.0 },
        { textRef: '1', from: 2.0, to: 3.0 }
      ]
    });
    const verseChangedSpy = jasmine.createSpy('verseChanged');
    env.component.audioPlayer.currentVerseChanged.subscribe(verseChangedSpy);
    env.wait();

    expect(verseChangedSpy).toHaveBeenCalledWith('s_1');
```
Maybe I should adjust it to this last example? Do you have any input?
(Actually, I'll go ahead and set it to the last one for now.)

josephmyers:
Thanks for your detailed explanation. This change, and the change to the tests, makes sense. We don't want to have to click play before the verseChanged event starts working. Without it we have to keep the "play on open" feature, as verseChanged will break if we ever remove that. Now, we've uncoupled those behaviors.

The only thing I'm really concerned about in this class is the removal of the `beforeEach()`. Instead of re-introducing manual setup to each test so that you can set up spies in a subset, why not add the spies to the `beforeEach()`? It keeps duplication out, and any future tests that want to spy can do so. And it helps keep the behavior we're talking about, that the verseChanged is fired on init, clearly defined in the test, notably "emits verse changed event"

marksvc:
@josephmyers I started working on putting things back into a `beforeEach()`, but quickly found that if I do that, I won't necessarily know what timings to start the component with. So I kept the `env = new TestEnvironment` in the individual tests.

But I put spy subscriptions into the TestEnvironment, to seek to fulfill some of what you're looking for here.

What do you think of this?

josephmyers:
Alright, sounds good 👍

PR #1993, opened 2023-08-04
SF-2147 Fix notes deleted in PT not deleted in SF

RaymondLuong3:
When multiple notes were deleted from a thread with 3 or more notes in PT, the result was not consistent in SF because when we removed notes from the notes property of a note thread, the operation depends on a static list. But when we remove multiple notes, the list changes for each note removal, and the result is that the wrong notes might be removed from the thread.
This change removes notes from the end of the list to the start of the list so that subsequent removal ops are correct with each note removed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1993)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
@@ -1206,21 +1206,15 @@ public class ParatextSyncRunner : IParatextSyncRunner
                 _syncMetrics.Notes.Added++;
             }
 
-            // Permanently removes a note
-            List<int> removedIndices = new List<int>();
-            for (int i = 0; i < change.NoteIdsRemoved.Count; i++)
-            {
-                string removedId = change.NoteIdsRemoved[i];
-                int index = threadDoc.Data.Notes.FindIndex(n => n.DataId == removedId);
-                if (index >= 0)
-                {
-                    removedIndices.Add(index);
-                }
-            }
+            List<int> removedIndices = change.NoteIdsRemoved
+                .Select(id => threadDoc.Data.Notes.FindIndex(n => n.DataId == id))
+                .Where(index => index >= 0)
+                .ToList();
             // Go through the indices in reverse order so subsequent removal indices are not affected
-            removedIndices.Sort((a, b) => b - a);
+            removedIndices.Sort((a, b) => b.CompareTo(a));
             foreach (int index in removedIndices)
             {
+                // Permanently removes a note
                 op.Remove(td => td.Notes, index);
                 _syncMetrics.Notes.Removed++;
             }
--- a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
@@ -1207,14 +1207,22 @@ public class ParatextSyncRunner : IParatextSyncRunner
             }
 
             // Permanently removes a note
-            foreach (string removedId in change.NoteIdsRemoved)
+            List<int> removedIndices = new List<int>();
+            for (int i = 0; i < change.NoteIdsRemoved.Count; i++)
             {
+                string removedId = change.NoteIdsRemoved[i];
                 int index = threadDoc.Data.Notes.FindIndex(n => n.DataId == removedId);
                 if (index >= 0)
                 {
-                    op.Remove(td => td.Notes, index);
-                    _syncMetrics.Notes.Removed++;
+                    removedIndices.Add(index);
                 }
+            }
+            // Go through the indices in reverse order so subsequent removal indices are not affected
+            removedIndices.Sort((a, b) => b - a);
+            foreach (int index in removedIndices)
+            {
+                op.Remove(td => td.Notes, index);
+                _syncMetrics.Notes.Removed++;
             }
 
             if (change.Position != null)
--- a/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs
+++ b/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs
@@ -2026,6 +2026,50 @@ public class ParatextSyncRunnerTests
     }
 
     [Test]
+    public async Task SyncAsync_RemovesMultipleNotesFromThreadDoc()
+    {
+        var env = new TestEnvironment();
+        string sfProjectId = "project01";
+        string threadId = "thread01";
+        string dataId = "dataId01";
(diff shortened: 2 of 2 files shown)


Nateowami:
@RaymondLuong3 In general, an issue and/or PR should be titled based on what it does, not how it does it. When I read the title of this PR, it's not clear to me why you would need to delete notes in reverse chronological order. Even reading the PR description doesn't really make it clear.

I think a decent title might be something like "SF-2147 Fix notes deleted in PT still being shown in SF"

Then in the description of the PR or commit it can be explained how this was accomplished. Otherwise the title isn't doing its job of introducing the topic.

Nateowami:
(Sorry, didn't mean to close)

RaymondLuong3:
Thanks, I've updated the PR title

pmachapman (changes requested):
A good fix to this bug. Just some minor feedback on changes to make the code a bit more readable.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1993#-:-NbC-D6i8u8gZoZbV7Os:b359rxz)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 1219 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1993#-NbBvz7x7mXImJgjLmr_:-NbBvz7x7mXImJgjLmra:b52524) ([raw file](https://github.com/sillsdev/web-xforge/blob/6050a10266f7203588f82d814389590537a2b18b/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L1219)):*
> ```Smalltalk
>                     removedIndices.Add(index);
>                 }
>             }
> ```

You could simplify this by making it a foreach loop:
```cs
foreach (var id in change.NoteIdsRemoved)
{
    int index = threadDoc.Data.Notes.FindIndex(n => n.DataId == id);
    if (index >= 0)
    {
        removedIndices.Add(index);
    }
}
```
Or perhaps even a one line statement if you like:
```cs
List<int> removedIndices = change.NoteIdsRemoved.Select(id => threadDoc.Data.Notes.FindIndex(n => n.DataId == id)).Where(index => index >= 0).ToList();
```



_Code quote:_
```Smalltalk
            for (int i = 0; i < change.NoteIdsRemoved.Count; i++)
            {
                string removedId = change.NoteIdsRemoved[i];
                int index = threadDoc.Data.Notes.FindIndex(n => n.DataId == removedId);
                if (index >= 0)
                {
                    removedIndices.Add(index);
                }
            }
```
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 1221 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1993#-NbBxr2Z48Fl2MISfBO9:-NbBxr2Z48Fl2MISfBOA:bnwsrz5) ([raw file](https://github.com/sillsdev/web-xforge/blob/6050a10266f7203588f82d814389590537a2b18b/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L1221)):*
> ```Smalltalk
>             }
>             // Go through the indices in reverse order so subsequent removal indices are not affected
>             removedIndices.Sort((a, b) => b - a);
> ```

Usually in C# to sort descending we use this:
```cs
removedIndices.Sort((a, b) => b.CompareTo(a));
```
*CompareTo() is better because b - a can wrap around in cases where `b - a > int.MaxValue` - something which is not likely to occur with array indices, but is still a good pattern to use nonetheless.*

Or you can do it in two statements:
```cs
removedIndices.Sort();
removedIndices.Reverse();
```
*This is more readable, but iterates over the list twice. Not a huge issue given the size of the arrays we are likely dealing with and the frequency of execution, but again up to you which method you prefer.*



_Code quote:_
```Smalltalk
removedIndices.Sort((a, b) => b - a);
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1993)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 1219 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1993#-NbBvz7x7mXImJgjLmr_:-NbLwEwM978SZXon_kQ5:bdx4c24) ([raw file](https://github.com/sillsdev/web-xforge/blob/6050a10266f7203588f82d814389590537a2b18b/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L1219)):*

You could simplify this by making it a foreach loop:
```cs
foreach (var id in change.NoteIdsRemoved)
{
    int index = threadDoc.Data.Notes.FindIndex(n => n.DataId == id);
    if (index >= 0)
    {
        removedIndices.Add(index);
    }
}
```
Or perhaps even a one line statement if you like:
```cs
List<int> removedIndices = change.NoteIdsRemoved.Select(id => threadDoc.Data.Notes.FindIndex(n => n.DataId == id)).Where(index => index >= 0).ToList();
```
</blockquote></details>

I like the one liner. I agree, it is more readable, thanks for the suggestion.
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 1221 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1993#-NbBxr2Z48Fl2MISfBO9:-NbLweuL8jEAjimcr-IU:b-7i1wj3) ([raw file](https://github.com/sillsdev/web-xforge/blob/6050a10266f7203588f82d814389590537a2b18b/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L1221)):*

Usually in C# to sort descending we use this:
```cs
removedIndices.Sort((a, b) => b.CompareTo(a));
```
*CompareTo() is better because b - a can wrap around in cases where `b - a > int.MaxValue` - something which is not likely to occur with array indices, but is still a good pattern to use nonetheless.*

Or you can do it in two statements:
```cs
removedIndices.Sort();
removedIndices.Reverse();
```
*This is more readable, but iterates over the list twice. Not a huge issue given the size of the arrays we are likely dealing with and the frequency of execution, but again up to you which method you prefer.*
</blockquote></details>

Thanks. I think I am happy with `b.CompareTo(a)`. I'm so glad to have a good C# developer on our team!

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1993#-:-NbMiCL99xgQ0dDSD9zR:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #1994, opened 2023-08-04
SF-2115 No projects connected page to material

RaymondLuong3:
Before
No projects connected before

After
No projects connected after

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1994)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/start/start.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/start/start.component.html
@@ -1,9 +1,9 @@
 <ng-container *transloco="let t; read: 'start'">
   <div fxLayout="column" fxLayoutAlign="center center" *ngIf="!isAppLoading">
-    <p mdcBody1>{{ t("not_connected_to_any_projects") }}</p>
+    <p>{{ t("not_connected_to_any_projects") }}</p>
     <p>
-      <button mdc-button [unelevated]="true" type="button" [appRouterLink]="['/connect-project']">
-        <mdc-icon>add</mdc-icon>{{ t("connect_project") }}
+      <button mat-flat-button color="primary" type="button" [appRouterLink]="['/connect-project']">
+        <mat-icon>add</mat-icon><span class="text-padding">{{ t("connect_project") }}</span>
       </button>
     </p>
   </div>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/start/start.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/start/start.component.scss
@@ -1,0 +1,9 @@
+:host ::ng-deep .mat-button-wrapper {
+  display: flex;
+  .mat-icon {
+    align-self: center;
+  }
+  .text-padding {
+    padding: 0 0.5em;
+  }
+}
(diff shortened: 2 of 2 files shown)


josephmyers (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1994)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/start/start.component.scss` line 1 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1994#-NbCwELp6KoLbomLjtUE:-NbCwELp6KoLbomLjtUF:b-13s3l9) ([raw file](https://github.com/sillsdev/web-xforge/blob/4aa248f1d558373fa852ce57e25b884246266274/src/SIL.XForge.Scripture/ClientApp/src/app/start/start.component.scss#L1)):*
> ```SCSS
> :host ::ng-deep .mat-button-wrapper {
> ```

I think we're trying to avoid ::ng-deep when possible. This CSS achieved the same result for me -- can you try it instead?



_Code snippet:_
```SCSS
button {
  .text-padding {
    padding: 0 0.5em;
  }
}
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1994)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/start/start.component.scss` line 1 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1994#-NbCwELp6KoLbomLjtUE:-NbKvKluCZMN43IRSKhU:brxyfqr) ([raw file](https://github.com/sillsdev/web-xforge/blob/4aa248f1d558373fa852ce57e25b884246266274/src/SIL.XForge.Scripture/ClientApp/src/app/start/start.component.scss#L1)):*

I think we're trying to avoid ::ng-deep when possible. This CSS achieved the same result for me -- can you try it instead?
</blockquote></details>

Yes, it is much cleaner. A very subtle difference is that the + icon is ever so slightly lower in comparison to the text. But I don't think it is very noticeable. Here is what it currently looks like.

josephmyers (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/1994)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/start/start.component.scss` line 1 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/1994#-NbCwELp6KoLbomLjtUE:-NbN3gO6914fNiFve0Sb:bsdr92b) ([raw file](https://github.com/sillsdev/web-xforge/blob/4aa248f1d558373fa852ce57e25b884246266274/src/SIL.XForge.Scripture/ClientApp/src/app/start/start.component.scss#L1)):*

Yes, it is much cleaner. A very subtle difference is that the + icon is ever so slightly lower in comparison to the text. But I don't think it is very noticeable. Here is what it currently looks like.


</blockquote></details>

Ah, I see that now. It is subtle, but if you want to keep the old way, that's fine, too!

PR #2000, opened 2023-08-09
SF-2186 Skip syncing notes with invalid content to prevent sync failure

RaymondLuong3:
This change logs the content of notes that are storing xml that is not correctly being encoded to be set to the InnerXml of the Contents note of a Paratext Comment. Since SF notes are not currently being exported to PT, this code is only run to provide warnings in the logs, so it is OK to modify it and skip over updating those comments.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2000)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -2291,8 +2291,15 @@ public class ParatextService : DisposableBase, IParatextService
                     {
                         if (comment.Contents == null)
                             comment.AddTextToContent(string.Empty, false);
-                        comment.Contents.InnerXml = xml;
-                        commentUpdated = true;
+                        try
+                        {
+                            comment.Contents.InnerXml = xml;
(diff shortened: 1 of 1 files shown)


Nateowami (approved):
I meant to comment this before making changes:

> Looks good. Just a couple thoughts:
> 1. I think it would be preferable to have this target master first, and then cherry-pick that to live.
> 2. I think it would be good to leave behind a FIXME to make it easier to find things that still need to be fixed

I'm about to make a new QA release that is a candidate for shipping, so I wanted to get this into master right now. So I've changed the target, added a commit with a FIXME comment, and rebased this on master.

> Reviewed 1 of 1 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2000#-:-NbSIE0t2PFh-NSppeAk:bbnlt2w)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2003, opened 2023-08-10
SF-2024 Highlight current audio text

RaymondLuong3:
* poll the audio at an interval of 500ms
* compare the current seek position of the player to the timing data
* emit the verse reference when it has been changed
* highlight the audio verse when reference changes
* restore highlighting of the question verse text question audio played

I wasn't sure what the best way to determine what the currently playing verse was except by polling the current audio time value and looking up what the textRef of the time would be. There might be a better way that I just have not thought of. Maybe we could set an observable to be the length of the currently playing segment. But then that doesn't work when the user uses the slider to set a new seek position.
Something to improve is that the text doesn't scroll (which I actually like) when the currently highlighted verse goes off the page.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2003)


nigel-wells (commented):
I was excited to try this out but the verses don't seem to be highlighting - unless I've misunderstood how it was supposed to work?

For auto-scroll, we already have a method to scroll to a segment don't we - is it as simple as calling that? On the YouVersion app you can still scroll around and, as long as your finger is on the screen it won't jump to the active segment. If, however, you stop touching, then it will move to the segment of to match the audio - this felt quite good and not annoying for those who wanted to scroll.

There are a bunch of events on the `audio` element that may work for getting an emitted seek value rather than polling: https://www.w3schools.com/tags/ref_av_dom.asp. Some of them sound right but they may only trigger on a user or API interaction rather than constantly which is what you want. Otherwise, we could probably setup a subject on the audio player that emits the current position every 500ms and then you can simply subscribe to that.

There a good bunch of `console.logs` in there too that you may want to remove :)

> Reviewed 2 of 9 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-:-NbXjE85A5Z3wA9bZwQV:b-fvn06c)* status: 2 of 9 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player.ts` line 125 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-NbXgUJ4D_5zTL6_kgxF:-NbXgUJ4D_5zTL6_kgxG:b-pz8vmm) ([raw file](https://github.com/sillsdev/web-xforge/blob/269f2509c07620905a661c23cf7bed03a4612867/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player.ts#L125)):*
> ```XML
>   }
> 
>   play(): void {
> ```

Thanks for moving these down to the right spot

RaymondLuong3 (commented):
Good thinking for moving the polling to the audio player. I like that making that available at that level means other components can make use of it.
Thanks for suggesting for me to look at the Bible app. I see how it is working now, and I do find that the scroll on every verse isn't actually as distracting as I was thinking. So I've updated this to scroll each time a new verse is highlighted.

Hmm, I may know why you weren't seeing the highlighting, I hadn't considered what format the textRef property on Audio timing entries would look like. After speaking with Nathaniel, it looks like we will support a variety of textRef formats. I've updated this PR to handle the most common use case with textRefs formatted like `v1` or `1`. Now it should be working for you.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-:-Nbafe1U0ygTgNArOUk_:b-do8bc9)* status: 1 of 10 files reviewed, all discussions resolved (waiting on @nigel-wells)

josephmyers:
FYI [SF-2078](https://jira.sil.org/browse/SF-2178) I made last week. Do we still need it?

RaymondLuong3:
I looked at SF-2078 and it seems it has to do with removing MDC from editor component. I still see we have mdc in the HTML, so it is still a necessary issue. Was there I different issue you were thinking of?

nigel-wells (changes requested):
This is working great now after those changes. 
Only other observation is that verse one is highlighted as soon as the audio starts playing even if my timing data says it doesn't start until 7 seconds in. I'm not really sure what the right answer is because if we don't highlight anything for timing data we can't match then that may also look wrong? I think for MVP I'd be happy with what you've got unless you feel like you know the solution?

> Reviewed 9 of 9 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-:-NcLO29C64-K93bQreAa:bx67rgg)* status: all files reviewed, 3 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.ts` line 37 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-NcLLiux3qRu8MIblFcV:-NcLLiux3qRu8MIblFcW:b-cbj58k) ([raw file](https://github.com/sillsdev/web-xforge/blob/3ad351059480e1bdd5d1b5f88dd79834ff376bc2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.ts#L37)):*
> ```XML
>     const audioTiming: AudioTiming | undefined = this._timing.find(t => t.to > currentTime);
>     if (audioTiming?.textRef === 's') {
>       return 's_' + this._timing.filter(t => t.textRef === 's' && t.to <= audioTiming.to).length;
> ```

Can you add a comment in here for what we are supporting currently
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.ts` line 170 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-NcLNiSF0Ky8uzGVePoy:-NcLNiSF0Ky8uzGVePoz:b-vmikfx) ([raw file](https://github.com/sillsdev/web-xforge/blob/3ad351059480e1bdd5d1b5f88dd79834ff376bc2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.ts#L170)):*
> ```XML
>   }
> 
>   private scrollToActiveVerse(verseRef: VerseRef | undefined): void {
> ```

Let's rename this method as we're not just scrolling to a verse passed to the method
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player.ts` line 54 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-NcLIrcv0YoTnlH4HLzp:-NcLIrcv0YoTnlH4HLzq:bqo8a1x) ([raw file](https://github.com/sillsdev/web-xforge/blob/3ad351059480e1bdd5d1b5f88dd79834ff376bc2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player.ts#L54)):*
> ```XML
>           return;
>         }
>         playSubscription.unsubscribe();
> ```

This is great. Can we also unsubscribe if the player stops or is destroyed - or anything that isn't playing really.

RaymondLuong3 (commented):
Hmm, are you using the sample audio text data in our shared folder? When I use the text and text audio data, the first verse that gets highlighted is the section heading that gets read. In this situation that seems appropriate.
But you are right, if the timing does not start until 7 seconds in, and the first verse is highlighted, that is something we can improve on further down the road.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-:-NcOyU_45ItWkWOAepDK:b-8h2a32)* status: 5 of 14 files reviewed, 3 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.ts` line 37 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-NcLLiux3qRu8MIblFcV:-NcO1JHDC1WESXoL-6Pq:btyeq5o) ([raw file](https://github.com/sillsdev/web-xforge/blob/3ad351059480e1bdd5d1b5f88dd79834ff376bc2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.ts#L37)):*

Can you add a comment in here for what we are supporting currently
</blockquote></details>

Good thinking. Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.ts` line 170 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-NcLNiSF0Ky8uzGVePoy:-NcO015p2AfhHtlTxDaO:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/3ad351059480e1bdd5d1b5f88dd79834ff376bc2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.ts#L170)):*

Let's rename this method as we're not just scrolling to a verse passed to the method
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player.ts` line 54 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-NcLIrcv0YoTnlH4HLzp:-NcNziYv3BQ9EBUmVjso:b-u95piy) ([raw file](https://github.com/sillsdev/web-xforge/blob/3ad351059480e1bdd5d1b5f88dd79834ff376bc2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player.ts#L54)):*

This is great. Can we also unsubscribe if the player stops or is destroyed - or anything that isn't playing really.
</blockquote></details>

I think that situation is already being handled, since it will unsubscribe when the audio is not currently playing. The SubscriptionDisposable implementation of `subscribe()` handles the case that the current AudioPlayer is destroyed. Let me know if you understand something differently than I do.

nigel-wells (approved):
My test data is in various states from various testing and doesn't match the book data at all. I take it with a grain of salt but it does help highlight error messages from time to time. I'm happy to proceed as is.

:lgtm:

> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-:-NcPMgBL-kGBAHGpvHOu:b7c4hd7)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.ts` line 37 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-NcLLiux3qRu8MIblFcV:-NcPG1bN3JuHUxeqJ3gc:b-zcjvte) ([raw file](https://github.com/sillsdev/web-xforge/blob/3ad351059480e1bdd5d1b5f88dd79834ff376bc2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.ts#L37)):*

Good thinking. Done.
</blockquote></details>

Thanks
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player.ts` line 54 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-NcLIrcv0YoTnlH4HLzp:-NcPM__81xB_ss62o-nM:be93gzo) ([raw file](https://github.com/sillsdev/web-xforge/blob/3ad351059480e1bdd5d1b5f88dd79834ff376bc2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player.ts#L54)):*

I think that situation is already being handled, since it will unsubscribe when the audio is not currently playing. The SubscriptionDisposable implementation of `subscribe()` handles the case that the current AudioPlayer is destroyed. Let me know if you understand something differently than I do.
</blockquote></details>

Yes, you're right, this is all fine. I did some checks and it works as I'd expected it to work.

RaymondLuong3 (commented):
@nigel-wells Could you review the changes I made since you approved this PR?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-:-Ncwtxc0DIy6QRH41qEB:bo0ldcd)* status: 11 of 15 files reviewed, all discussions resolved (waiting on @nigel-wells)

nigel-wells (changes requested):
Just one comment on this

> Reviewed 3 of 4 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-:-Nd2iYlC-V4MLwqxLfUb:b8imfn9)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts` line 120 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-Nd2hvci4QlQXbV6Olr4:-Nd2hvci4QlQXbV6Olr5:b-b31ka6) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4ad7c3da3369b4c24aaf8ec35044a47570043ff/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts#L120)):*
> ```XML
>   });
> 
>   it('can delete audio timing data', async () => {
> ```

This test has 3 concepts of changing timing data: Delete, close, remove. I know there was a recent PR to close the audio player rather than deleting data - is this test still relevant?

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2003)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player.ts` line 57 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-Nd2ssYnCWJD46SfamRN:-Nd2ssYnCWJD46SfamRO:b-lfubjl) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4ad7c3da3369b4c24aaf8ec35044a47570043ff/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player.ts#L57)):*
> ```XML
>         this.setSeek(0);
>       }
>       const playSubscription: Subscription = this.subscribe(interval(this.POLL_INTERVAL), () => {
> ```

This is the spot Nathaniel mentioned in Slack. Could you put this in the handler for 'timeupdate' instead? Unless you specifically want it to only update at 500ms.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2003)* status: 14 of 15 files reviewed, 3 unresolved discussions (waiting on @josephmyers and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts` line 120 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-Nd2hvci4QlQXbV6Olr4:-Nd36hkB67q32tBlI3Q6:b-mm1dpe) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4ad7c3da3369b4c24aaf8ec35044a47570043ff/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts#L120)):*

This test has 3 concepts of changing timing data: Delete, close, remove. I know there was a recent PR to close the audio player rather than deleting data - is this test still relevant?
</blockquote></details>

Yes, the delete audio button is something that is going to be removed now. I think we can remove this test.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player.ts` line 57 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-Nd2ssYnCWJD46SfamRN:-Nd37I4DB1xmyfL4-y3w:b-7vduir) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4ad7c3da3369b4c24aaf8ec35044a47570043ff/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player.ts#L57)):*

This is the spot Nathaniel mentioned in Slack. Could you put this in the handler for 'timeupdate' instead? Unless you specifically want it to only update at 500ms.
</blockquote></details>

OK, I see that is likely the way to go. Let's do that in a follow-up PR so that it can go through the proper testing channel. Thanks for pointing it out.

nigel-wells (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-:-Nd39atB-_WWfjdExMrr:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @github-code-scanning[bot] and @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts` line 120 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2003#-Nd2hvci4QlQXbV6Olr4:-Nd39X328f6ZpYJTHi3Q:b2nhll9) ([raw file](https://github.com/sillsdev/web-xforge/blob/f4ad7c3da3369b4c24aaf8ec35044a47570043ff/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts#L120)):*

Yes, the delete audio button is something that is going to be removed now. I think we can remove this test.
</blockquote></details>

Ok.

PR #2008, opened 2023-08-16
SF-1910 Docker Support

pmachapman:
This Pull Request adds support for running Scripture Forge from a Docker container.

In addition, to get the Migrator working with Docker, minor modifications to allow passing of configuration via environment variables was added.

# Running the Docker Container

## Preliminary Notes
 * Ensure that Docker is installed. The easiest way is with [Docker Desktop](https://www.docker.com/).
 * If the MongoDB service is running on your computer, stop it:
    * **Linux:** `sudo systemctl stop mongod`
    * **Windows:** `sc stop mongodb`

## Linux: Running from the command line
1. Change to the `src/Docker` directory
2. Build the container: `docker compose build`
3. Start the container: `docker compose up`
4. Shutdown the container by pressing Ctrl+C

## Windows/Linux: Run from Visual Studio Code
1. Run the task: docker-compose-up
2. To debug .NET, use the "Docker .NET Attach (Preview)" configuration

## Windows: Run from Visual Studio
1. Change the project dropdown in the Visual Studio toolbar to: "Docker"
3. Run: "Docker Compose"

# Notes
 * By default your user secrets will be imported into the container.
 * Currently only debugging .NET code is supported. I am still investigating ways to allow debugging of the node frontend.
 * The debug configuration "Docker .NET Attach (Preview)" was named by Microsoft's Docker Tools.
 * The .override.yml file contains configuration that is loaded by default on developer machines to help the container run in a dev friendly way.
   * We will create separate .yml files for staging and production when it is time to run the docker container in other environments.
 * The package "Microsoft.VisualStudio.Azure.Containers.Tools.Targets" allows Visual Studio to use "Fast Mode" that makes starting and debugging the container easier by skipping the build phase.
 * I have tested this on Windows 11 and Ubuntu 23.04 using the latest Docker Desktop.
   * I am not sure if this will run on Mac

# Future Ideas
 * Using the docker container for testing configuration changes, such as OS upgrades or MongoDB upgrades.
 * Separating the Node Frontend into a separate service (will require reworking the csproj targets).
   * I believe this might be holding back Node debugging - for dotnet to run effectively the project must be published, which packages the SPA.
 * Automated testing using the container?
 * Apple Silicon support
   * The SIL.Machine dependency is holding this back (and maybe some others?)
   * A Raspberry PI could probably be used to test/develop this?

# End-to-end running on the Vagrant machine
After starting the vagrant machine:
```sh
cd ~/src/web-xforge/deploy/vagranta/sfdev
vagrant up
````
Ensure that the file "warning-not-provisioned.txt" file is no longer on the desktop, then in the virtual machine:

1. Click "Install Now" to the Software Updater dialog
2. When the Updater has finished, click "Restart Now"
3. Open a terminal and run (substituting values where enclosed in `{}` brackets):
```sh
# Update packages
sudo apt update
sudo apt upgrade -y
sudo apt autoremove -y
# Setup Git
git config --global user.name "{my_name}"
git config --global user.email "{my_email}"
# Add Docker GPG key
sudo install -m 0755 -d /etc/apt/keyrings
curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor -o /etc/apt/keyrings/docker.gpg
sudo chmod a+r /etc/apt/keyrings/docker.gpg
# Add Docker repo
echo \
    "deb [arch="$(dpkg --print-architecture)" signed-by=/etc/apt/keyrings/docker.gpg] https://download.docker.com/linux/ubuntu \
    "$(. /etc/os-release && echo "$VERSION_CODENAME")" stable" | \
    sudo tee /etc/apt/sources.list.d/docker.list > /dev/null
# Install Docker
sudo apt-get update
sudo apt-get install -y docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin
# Add vagrant user to docker group
sudo groupadd -f docker
sudo usermod -aG docker $USER
newgrp docker
# Test Docker
docker run hello-world
# Change to the branch
cd ~/src/web-xforge
git fetch --all
git switch feature/SF-1910
# Configure user secrets
cd ~/src/web-xforge/src/SIL.XForge.Scripture
dotnet user-secrets set "Auth:BackendClientSecret" "{backend_client_secret}"
dotnet user-secrets set "Paratext:ClientId" "{paratext_client_id}"
dotnet user-secrets set "Paratext:ClientSecret" "{paratext_client_secret}"
dotnet user-secrets set "Paratext:ResourcePasswordHash" "{paratext_resource_password_hash}"
dotnet user-secrets set "Paratext:ResourcePasswordBase64" "{paratext_resource_password_base64}"
# Create debugger directory
mkdir ~/.vsdbg
```
4. Reboot the vagrant machine (to apply the new docker group)
5. Open a terminal and run:
```sh
# Stop Mongo
sudo systemctl stop mongod
# Build the Docker image
cd ~/src/web-xforge/src/Docker
docker compose build
docker compose up
```
6. Open http://localhost:5000/ using a web browser, and log into Scripture Forge.
7. Run Visual Studio Code and:
	a. Open the folder "~/src/web-xforge"
	b. Click "Yes, I trust the authors"
	c. Click the Extensions button in the sidebar
	d. In the Recommended list, click "Install" next to Docker
	e. Next to the C# extension, click "Reload Required"
	f. Click the Run and Debug button in the sidebar
	g. Click the down arrow in the configuration selector and select "Docker .NET Attach (Preview)"
	h. Click the Play button
	i. Select the process "dotnet" from the dropdown that will appear at the top center of the window

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2008)


marksvc:
I installed docker in Ubuntu 20.04. 
I needed to `systemctl stop mongod.service`
In an incognito window I went to localhost:5000 and logged in. I am in an infinite loop between localhost:5000 and localhost:5000/callback/auth0?code=... . The docker output shows messages about
"""
web-xforge  |           POST https://sil-appbuilder.auth0.com/oauth/token
web-xforge  |           Response:                                                                                                                                                                                                                                                      
web-xforge  |           StatusCode: 401, ReasonPhrase: 'Unauthorized', Version: 1.1, Content: System.Net.Http.HttpConnectionResponseContent, Headers:
[...]
web-xforge  |           {"error":"access_denied","error_description":"Unauthorized"}
web-xforge  |          at SIL.XForge.ExceptionHandler.EnsureSuccessStatusCode(HttpResponseMessage response) in /src/SIL.XForge/ExceptionLogging/ExceptionHandler.cs:line 44
web-xforge  |          at SIL.XForge.Services.AuthService.GetAccessTokenAsync() in /src/SIL.XForge/Services/AuthService.cs:line 142
"""

To attach to docker in vscode I needed to install the docker extension.

I tried to attach to docker in VSCode. 

VSCode had a popup saying "Attaching to container requires .NET debugger in the container. [Do you want to copy .NET debugger into the container?", which I needed to click Yes to quickly before it went away. 

At first the docker output area showed
> 2023-08-16 14:49:57.998 [error] permission denied while trying to connect to the Docker daemon socket at unix:///var/run/docker.sock: Get "http://%2Fvar%2Frun%2Fdocker.sock/v1.24/containers/web-xforge/json": dial unix /var/run/docker.sock: connect: permission denied

So presumably I should put my user in the docker group. I did that and rebooted, but now trying to attach in VSCode says 
> command 'csharp.listRemoteDockerProcess' not found.

I'm still experimenting with it.

marksvc:
I rolled back my system and re-did the docker installation steps, but this time with my user in the "docker" group from earlier in the process, in case that impacts things. 

> Error: EACCES: permission denied, open '/home/vagrant/.vsdbg/getvsdbg.sh'

I fixed the above error on docker attaching with `sudo chown -R vagrant:vagrant ~/.vsdbg`, as it was owned by root.

Docker attach still results in error dialog:

> command 'csharp.listRemoteDockerProcess' not found

I am using VSCode 1.81.0, dated 2023-08-02. Looks like there is an update available to vscode version 1.81.1. But I still get the error after upgrading.

marksvc (changes requested):
Thanks for your work on this. I'm just learning Docker, which of course affects how well I can review this :). I made some comments.

> Reviewed 14 of 14 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-:-Nc-cPHbDEaIgoXKoIgZ:b7tnyoo)* status: all files reviewed, 9 unresolved discussions (waiting on @pmachapman)

___
*[`src/Docker/docker-compose.override.yml` line 36 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-SYr89jK81g_D75s0:-Nc-SYr89jK81g_D75s1:bpngiwp) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/Docker/docker-compose.override.yml#L36)):*
> ```YAML
>       - xforge-data:/var/lib/xforge
>       - ~/.microsoft/usersecrets:/root/.microsoft/usersecrets:ro # Share User Secrets with Linux Host
>       - ~/.vsdbg:/remote_debugger:rw # Visual Studio Code Debugger Support
> ```

Why then does VSCode ask to copy in the vsdbg debugger, if the volume is mounted in in this way?
___
*[`src/Docker/docker-compose.win.yml` line 7 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-bpHO1NB7V6XlXHgo:-Nc-bpHPCHHzfzrW052T:b-jvrzb7) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/Docker/docker-compose.win.yml#L7)):*
> ```YAML
>   web-xforge:
>     environment:
>       - staticWebAssets=DISABLED # Fixes System.ArgumentException: 'The path must be absolute. (Parameter 'root')'
> ```

This is configured both here and in launchSettings.json. Is it needed in both places?



_Code quote:_
```YAML
staticWebAssets=DISABLED
```
___
*[`src/Docker/docker-compose.win.yml` line 9 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-TjXJ27RYh7XBTfKr:-Nc-TjXJ27RYh7XBTfKs:b-ux7hgx) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/Docker/docker-compose.win.yml#L9)):*
> ```YAML
>       - staticWebAssets=DISABLED # Fixes System.ArgumentException: 'The path must be absolute. (Parameter 'root')'
>     volumes:
>       - ${APPDATA}/Microsoft/UserSecrets:/root/.microsoft/usersecrets:ro # Share User Secrets with Windows Host
> ```

This confused me at first. But I understand what it means now. It might be clearer to say "from".



_Code quote:_
```YAML
with 
```
___
*[`src/SIL.XForge.Scripture/Dockerfile` line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-VPwh3SseAfKUBr2R:-Nc-VPwh3SseAfKUBr2S:b-v9aedf) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/SIL.XForge.Scripture/Dockerfile#L13)):*
> ```Dockerfile
> VOLUME /var/lib/xforge
> # Add Ubuntu Mirrors
> RUN sed -i 's/htt[p|ps]:\/\/archive.ubuntu.com\/ubuntu\//mirror:\/\/mirrors.ubuntu.com\/mirrors.txt/g' /etc/apt/sources.list
> ```

In my experience, relying on a mirrors listing results in trying to use mirrors that don't work. I didn't notice any problems when working with it today. Maybe that isn't a problem here. But I'll mention that in case we see a problem with it.

BTW if you're using sed to replace strings with slashes, you can choose a different separator character than `/` to make it a bit easier to read. For example,
```bash
sed 's/\/var\/tmp/\/tmp/'
```
is easier to read when written as
```bash
sed 's|/var/tmp|/tmp|'
```
You can use whatever character is a good idea. `#` stands out well.
___
*[`src/SIL.XForge.Scripture/Dockerfile` line 18 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-XNWF5Rmx5EPnTEa-:-Nc-XNWF5Rmx5EPnTEa0:b-7jam7j) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/SIL.XForge.Scripture/Dockerfile#L18)):*
I was looking at node's instructions and what the different curl options do. They [show](https://github.com/nodesource/distributions#using-ubuntu-3) `curl -fsSL`.  (Tho I see the actual script suggests `curl -sL` :)
I think that's an improvement over what we're doing here. Otherwise, if there's an error page given from the web server, we would pipe it to bash :).

>        -f, --fail
>              (HTTP) Fail silently (no output at all) on server errors. This is mostly done to better enable scripts etc to better deal with failed attempts. In normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also  describes  why  and
>              more). This flag will prevent curl from outputting that and return error 22.
>
>              This method is not fail-safe and there are occasions where non-successful response codes will slip through, especially when authentication is involved (response codes 401 and 407).


>        -S, --show-error
>              When used with -s, --silent, it makes curl show an error message if it fails.


Not that I approve of `curl|bash`, of course :-)



_Code quote:_
```Dockerfile
 curl -sL 
```
___
*[`src/SIL.XForge.Scripture/Dockerfile` line 49 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-ZttF8v6-KwDcwKP5:-Nc-ZttF8v6-KwDcwKP6:b-j56l1l) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/SIL.XForge.Scripture/Dockerfile#L49)):*
> ```Dockerfile
> # Install Node Repository
> RUN update-ca-certificates \
>     && curl -sL https://deb.nodesource.com/setup_16.x | bash -
> ```

(This can be changed, too, as discussed above.)



_Code quote:_
```Dockerfile
-sL
```
___
*[`src/SIL.XForge.Scripture/Program.cs` line 10 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-aFhH9M2iPH_JwcoG:-Nc-aFhH9M2iPH_JwcoH:b-4wwgsn) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/SIL.XForge.Scripture/Program.cs#L10)):*
> ```Smalltalk
> namespace SIL.XForge.Scripture;
> 
> public static class Program
> ```

I'm surprised this wasn't already static. Is this change to tidy this up, or was this significant to docker use?
___
*[`src/SIL.XForge.Scripture/Properties/launchSettings.json` line 5 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-atT6C_WwVvwhaL8d:-Nc-atT7-eN5a95oANFm:b-lt4ei2) ([raw file](https://github.com/sillsdev/web-xforge/blob/51c0c17e23b3c3ee2e2027b0ae6d553821266851/src/SIL.XForge.Scripture/Properties/launchSettings.json#L5)):*
> ```JSON
>     "SIL.XForge.Scripture": {
>       "commandName": "Project",
>       "launchBrowser": false,
> ```

What's the thinking behind removing launchBrowser: false ?
___
*[`src/SIL.XForge.Scripture/Properties/launchSettings.json` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-b9Fk8PhwmvUKRykZ:-Nc-b9Fk8PhwmvUKRyk_:b-hbiwfd) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/SIL.XForge.Scripture/Properties/launchSettings.json#L23)):*
> ```JSON
>       "commandName": "Docker",
>       "launchUrl": "{Scheme}://{ServiceHost}:{ServicePort}/",
>       "DockerfileRunArguments": "-p 5003:5003 -v C:/var/lib/scriptureforge:/var/lib/scriptureforge -v C:/var/lib/xforge:/var/lib/xforge -v \"{ProjectDir}/../RealtimeServer/\":/app/ClientApp/node_modules/realtime-server --name web-xforge --add-host host.docker.internal:host-gateway",
> ```

Okay, this  must be a Windows file for Visual Studio?



_Code quote:_
```JSON
C:/
```
___
*[`src/SIL.XForge.Scripture/Properties/launchSettings.json` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-bOyS8sOZZTZRVTId:-Nc-bOyS8sOZZTZRVTIe:bmjl446) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/SIL.XForge.Scripture/Properties/launchSettings.json#L23)):*
> ```JSON
>       "commandName": "Docker",
>       "launchUrl": "{Scheme}://{ServiceHost}:{ServicePort}/",
>       "DockerfileRunArguments": "-p 5003:5003 -v C:/var/lib/scriptureforge:/var/lib/scriptureforge -v C:/var/lib/xforge:/var/lib/xforge -v \"{ProjectDir}/../RealtimeServer/\":/app/ClientApp/node_modules/realtime-server --name web-xforge --add-host host.docker.internal:host-gateway",
> ```

I'm surprised to see these paths configured here. I would have thought this would have been some of what was already specified in the various docker configuration files. It's needed to set these directories here?

pmachapman (commented):
No problem - I am testing the Docker container on the vagrant image to see if I can recreate some of the problems you have encountered.

I did notice that Docker Desktop is not supported on Ubuntu 20.04, so I installed the [Docker Engine manually](https://docs.docker.com/engine/install/ubuntu/), and ran the SF Docker container this way:
```sh
cd ~/src/web-xforge/src/Docker/
sudo docker compose build
sudo docker compose up
```

Although to do so, I needed to copy my usersecrets to `/root/.microsoft/usersecrets/4d0606c3-0fc7-4d76-b43b-236485004e81/secrets.json` on the host.

The redirect loop is caused by the usersecrets not being accessible in the Docker container.

Also, I have added the Docker extension to the extension recommendations.

> command 'csharp.listRemoteDockerProcess' not found

I think this might also be permission related?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-:-Nc-g9eAAKjbFZtj8ER8:bd1lg7a)* status: all files reviewed, 9 unresolved discussions (waiting on @marksvc)

___
*[`src/Docker/docker-compose.override.yml` line 36 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-SYr89jK81g_D75s0:-Nc-g8379mvNLkqDBhYK:b-odepyu) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/Docker/docker-compose.override.yml#L36)):*

Why then does VSCode ask to copy in the vsdbg debugger, if the volume is mounted in in this way?
</blockquote></details>

I am not sure - I will see if I can create it - it may be a VS Code bug?
___
*[`src/Docker/docker-compose.win.yml` line 7 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-bpHO1NB7V6XlXHgo:-Nc-gts76jvNHOEj2ATJ:b-d5x32j) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/Docker/docker-compose.win.yml#L7)):*

This is configured both here and in launchSettings.json. Is it needed in both places?
</blockquote></details>

No - it was a hangover from debugging. Removed from launchSettings.json - Thanks!
___
*[`src/Docker/docker-compose.win.yml` line 9 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-TjXJ27RYh7XBTfKr:-Nc-h7998nC2XkX1FQNF:b-92n6i0) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/Docker/docker-compose.win.yml#L9)):*

This confused me at first. But I understand what it means now. It might be clearer to say "from".
</blockquote></details>

Good idea - fixed.
___
*[`src/SIL.XForge.Scripture/Dockerfile` line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-VPwh3SseAfKUBr2R:-Nc-hV6RDaAm4b9SaktZ:ba32exm) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/SIL.XForge.Scripture/Dockerfile#L13)):*

In my experience, relying on a mirrors listing results in trying to use mirrors that don't work. I didn't notice any problems when working with it today. Maybe that isn't a problem here. But I'll mention that in case we see a problem with it.

BTW if you're using sed to replace strings with slashes, you can choose a different separator character than `/` to make it a bit easier to read. For example,
```bash
sed 's/\/var\/tmp/\/tmp/'
```
is easier to read when written as
```bash
sed 's|/var/tmp|/tmp|'
```
You can use whatever character is a good idea. `#` stands out well.
</blockquote></details>

Good point. I enabled the mirrors as the default repo was endlessly timing out for me (it might just be a New Zealand problem?). Good idea to flag it, we should remove this if we update the distro, I think?

Thank you RE: sed - this was a copy-paste (and my knowledge of sed is non-existent). Changed.
___
*[`src/SIL.XForge.Scripture/Dockerfile` line 18 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-XNWF5Rmx5EPnTEa-:-Nc-kWZXBCd4-ZYYLOTc:b-hscml6) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/SIL.XForge.Scripture/Dockerfile#L18)):*

I was looking at node's instructions and what the different curl options do. They [show](https://github.com/nodesource/distributions#using-ubuntu-3) `curl -fsSL`.  (Tho I see the actual script suggests `curl -sL` :)
I think that's an improvement over what we're doing here. Otherwise, if there's an error page given from the web server, we would pipe it to bash :).

>        -f, --fail
>              (HTTP) Fail silently (no output at all) on server errors. This is mostly done to better enable scripts etc to better deal with failed attempts. In normal cases when an HTTP server fails to deliver a document, it returns an HTML document stating so (which often also  describes  why  and
>              more). This flag will prevent curl from outputting that and return error 22.
>
>              This method is not fail-safe and there are occasions where non-successful response codes will slip through, especially when authentication is involved (response codes 401 and 407).


>        -S, --show-error
>              When used with -s, --silent, it makes curl show an error message if it fails.


Not that I approve of `curl|bash`, of course :-)
</blockquote></details>

Thank you - fixed!

I think in a later image I will remove the need for this by using separate node services in their own containers for the frontend and realtime server. Not fully sure how I will tackle that one yet, though.
___
*[`src/SIL.XForge.Scripture/Dockerfile` line 49 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-ZttF8v6-KwDcwKP5:-Nc-kvhk4AZHuS7ZAO4k:b4855pb) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/SIL.XForge.Scripture/Dockerfile#L49)):*

(This can be changed, too, as discussed above.)
</blockquote></details>

Thanks! Done.
___
*[`src/SIL.XForge.Scripture/Program.cs` line 10 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-aFhH9M2iPH_JwcoG:-Nc-kyz15SW4d7dit-Wj:b-83zpyf) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/SIL.XForge.Scripture/Program.cs#L10)):*

I'm surprised this wasn't already static. Is this change to tidy this up, or was this significant to docker use?
</blockquote></details>

No just a tidy up I noticed while adding environment variable support.
___
*[`src/SIL.XForge.Scripture/Properties/launchSettings.json` line 5 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-atT6C_WwVvwhaL8d:-Nc-lCF8-i9bA4GEd55b:bc967x8) ([raw file](https://github.com/sillsdev/web-xforge/blob/51c0c17e23b3c3ee2e2027b0ae6d553821266851/src/SIL.XForge.Scripture/Properties/launchSettings.json#L5)):*

What's the thinking behind removing launchBrowser: false ?
</blockquote></details>

This was done automatically by Visual Studio. I kept reverting this line, but every time I opened the dialog in Visual Studio it would remove it...so I caved in...
___
*[`src/SIL.XForge.Scripture/Properties/launchSettings.json` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-b9Fk8PhwmvUKRykZ:-Nc-lTO5EAaVNcUbs15J:b-eo912t) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/SIL.XForge.Scripture/Properties/launchSettings.json#L23)):*

Okay, this  must be a Windows file for Visual Studio?
</blockquote></details>

Yes. For some reason it required an absolute path...I realize this won't work for everyone (I keep /var/lib on my D: drive), but it was a blocker sadly (and most Windows devs will have the directory on the C: drive).
___
*[`src/SIL.XForge.Scripture/Properties/launchSettings.json` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-bOyS8sOZZTZRVTId:-Nc-ljwa7TVVTm5eAG7A:b7kaxiq) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/SIL.XForge.Scripture/Properties/launchSettings.json#L23)):*

I'm surprised to see these paths configured here. I would have thought this would have been some of what was already specified in the various docker configuration files. It's needed to set these directories here?
</blockquote></details>

They have to be configured here as this profile is not using docker compose, but docker run. It is really only used if you want to use the MongoDB instance and SF directories on your host.

I originally thought of removing this profile as docker compose is the preferred method of running the container, but I figured it would also serve as an example of how to run the container using docker run.

When the PR is merged, I will create a wiki page going over all of this...

marksvc:
> No problem - I am testing the Docker container on the vagrant image to see if I can recreate some of the problems you have encountered.

Were you successful at logging in and using SF? What about at attaching the debugger?

> I did notice that Docker Desktop is not supported on Ubuntu 20.04, so I installed the [Docker Engine manually](https://docs.docker.com/engine/install/ubuntu/), 

Here are the commands I used to add Docker into my vagrant machine:
```bash
# https://docs.docker.com/desktop/install/linux-install/
sudo usermod --append --groups kvm vagrant

# https://docs.docker.com/engine/install/ubuntu/#set-up-the-repository
curl --silent --location --fail --show-error https://download.docker.com/linux/ubuntu/gpg | sudo gpg --dearmor --output /etc/apt/trusted.gpg.d/docker.gpg
sudo tee /etc/apt/sources.list.d/docker.list <<< "deb [arch=amd64 signed-by=/etc/apt/trusted.gpg.d/docker.gpg] https://download.docker.com/linux/ubuntu focal stable"
sudo apt-get update
sudo apt-get install --assume-yes docker-ce docker-ce-cli containerd.io docker-buildx-plugin docker-compose-plugin

# This is not a great security solution. https://askubuntu.com/q/477551
sudo usermod --append --groups docker vagrant
```
Then I reboot the machine, and do `docker compose build` and `.. up`, without `sudo`.

> and ran the SF Docker container this way:
> 
> cd ~/src/web-xforge/src/Docker/
> sudo docker compose build
> sudo docker compose up
> 
> Although to do so, I needed to copy my usersecrets to /root/.microsoft/usersecrets/4d0606c3-0fc7-4d76-b43b-236485004e81/secrets.json on the host.

I looked in my docker container today and see that /root/.microsoft/usersecrets/ has the same information as in my /home/vagrant/.microsoft/usersecrets/ on the host. (Presumably it is working from my user account, rather than needing to get them from host's /root/.microsoft/... , because I am using `docker compose` without sudo.)

> The redirect loop is caused by the usersecrets not being accessible in the Docker container.

>>    command 'csharp.listRemoteDockerProcess' not found
> 
> I think this might also be permission related?

I scanned for files in my home that are owned by root. I have some, but none stood out as something that would be a problem. 

I wondered if the remote debugger isn't present in the docker container, but I see that it is, at various paths such as 
```
/remote-Debugger/linux-x64
/remote_debugger/linux-musl-x64
/remote_debugger/.vsdbg/linux-x64
/remote_debugger/.vsdbg/linux-musl-x64
```

I tried a different launch pattern, using pipeTransport, from (a perhaps old page at) https://github.com/dotnet/vscode-csharp/wiki/Attaching-to-remote-processes , but without resulting in a change to the error message.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2008)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Dockerfile` line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-VPwh3SseAfKUBr2R:-Nc3M-QZ8-CfN_U35pL-:b-1e3z5o) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/SIL.XForge.Scripture/Dockerfile#L13)):*
> Good point. I enabled the mirrors as the default repo was endlessly timing out for me (it might just be a New Zealand problem?). Good idea to flag it, we should remove this if we update the distro, I think?

I'm happy with leaving it to use mirrors.txt as long as we don't start to see any problems from it. In a web search I found that someone else's complaint about a similar trouble was 10 years old, so maybe it's been improved since. It is helpful if we can make the mirror work well from different countries. 
___
*[`src/SIL.XForge.Scripture/Properties/launchSettings.json` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-Nc-bOyS8sOZZTZRVTId:-Nc3bNzJ26pRzsLC-fk8:b4kznet) ([raw file](https://github.com/sillsdev/web-xforge/blob/914db54084dc7b994052b31f83946c0c366c2850/src/SIL.XForge.Scripture/Properties/launchSettings.json#L23)):*
> this profile is not using docker compose, but docker run

Ahh, okay, thank you.

marksvc:
Oops I hadn't tried your updated PR. 

Interestingly, if my docker is _not running_, and I try to attach to docker, it
- Asks if it should copy the vsdbg debugger into the docker container. I say Yes.
- It shows for a minute that it is trying to copy it to /remote_debugger
- After a while that goes away and a message appear saying "command 'csharp.listRemoteDockerProcess' not found"
- In the Docker Output area, it says "Error response from daemon: Container abcabc.. is not running".

If the docker container _is_ running, and I try to attach to docker, it
- Goes right to saying "command 'csharp.listRemoteDockerProcess' not found"
- Is not asking about copying vsdbg.
- Does not show an error about container not running.

So it's almost like it can tell that the vsdbg is in the container, but maybe it fails to connect to and use it.

pmachapman (commented):
What version of VS Code and the C# extension are you using? I am currently using 2.0

csharp.listRemoteDockerProcess was added in [1.23.1 of the C# extension](https://github.com/dotnet/vscode-csharp/blob/0ba29bf8fde1a8184f6d4dcb2b454ff65d4c8c69/CHANGELOG.md?plain=1#L351).



> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-:-NcJOBub1e3ivipuY2Nt:bpl0ui6)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

marksvc:
vscode 1.81.1
C# extension ms-dotnettools.csharp 2.0.376

Were you successful at logging in when using the Docker in the sfdev vagrant machine? BTW, you will need to adjust the settings in Virtualbox for the sfdev machine at System - Processor - select "Enable Nested VT-x/AMD-V", if you haven't done so already.

I could spin up an Ubuntu 22.04 machine if you think that would make a meaningful difference? The sfdev vagrant is Ubuntu 20.04.

pmachapman (commented):
I was able to run it on the vagrant machine running on my Linux laptop. The "Enable Nested VT-x/AMD-V" option was disabled as I have an older CPU on this laptop (Intel i5 3rd gen).

After adding myself to the docker group:
```sh
sudo groupadd -f docker
sudo usermod -aG docker vagrant
newgrp docker
```

Then restarting, I could start docker:
```sh
cd ~/src/web-xforge/src/Docker/
docker compose build
docker compose up
```
And was able to log in to SF and debug with VS code.

I have done an end to end test and documented all steps in the PR notes (this will form the basis of the wiki article I imagine).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-:-NcPCBrADNYtqlZjHEAm:bgpbd9n)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

marksvc:
Thank you for your detailed steps. I am able to get it working in a new machine. I may have something awry in my normal machine. Previously, I did not get a drop-down in VSCode in which to select the dotnet process.

marksvc (approved):
Thanks for your work on this!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2008#-:-NcTq9z957Gbr8jvSCGQ:b7tt6lp)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2009, opened 2023-08-16
SF-2113 Convert checking questions to material

siltomato:
- In mobile view, questions are now full screen instead of in a 'drawer'
- Adjusted the look of the questions list (need feedback on visual style choices)
- Moved some global styles from `styles.scss` to component stylesheet that were only used in checking component

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss
@@ -7,10 +7,6 @@
 // https://m3.material.io/components/navigation-drawer/guidelines#56b3fea0-ecc6-4ff0-bb0a-80ff48c74a58
 $questionsPanelBreakpoint: 600px;
 
-@mixin snapBack($direction) {
-  transition: $direction 0.35s cubic-bezier(0.68, -0.55, 0.265, 1.55); // Snap back effect
-}
-
 .panel-content {
   display: flex;
   position: relative;
@@ -86,10 +82,10 @@ $questionsPanelBreakpoint: 600px;
     margin: 0;
     z-index: 7;
     background: #fff;
-    @include snapBack(left);
+    transition: transform 0.4s cubic-bezier(0.68, -0.6, 0.265, 1.6); // Snap back effect
 
     &.overlay-visible {
-      left: 0;
+      transform: translateX(100%);
 
       .questions-overlay-close {
         display: block;
@@ -102,12 +98,10 @@ $questionsPanelBreakpoint: 600px;
 :host-context([dir='rtl']) {
   #questions-panel {
     @include media-breakpoint('<=', sm, $questionsPanelBreakpoint) {
-      left: unset;
-      right: -100%;
-      @include snapBack(right);
+      left: 100%;
 
       &.overlay-visible {
-        right: 0;
+        transform: translateX(-100%);
       }
     }
   }


Nateowami (changes requested):
Overall looks good. A few quick comments from running this locally and looking for any issues.

I think the minimum width of the answer list is not enough:



> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-:-Nc8A8AO-CiZwzYk8yek:bf1dzwq)* status: 0 of 9 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3 and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss` line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc86oqR6o_BLoEoq9nK:-Nc86oqR6o_BLoEoq9nL:b3kokcl) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss#L27)):*
> ```SCSS
>     height: 40px;
>     overflow: hidden;
>     border-bottom: 1px solid #64798f1f;
> ```

I started this code review by checking out the branch and running it. Overall looks quite good.

I'm going to start out by complaining about the most trivial thing possible though:


The reason for it is the transparency in the border bottom color (so it should be a simple fix).

In general I don't like using transparency to specify a color. It's always felt slightly wrong to me, but until recently I didn't have a clear reason why. In certain scripts (such as Arabic), the glyphs can actually overlap, and a partly transparent text color ends up looking very bad (zoomed in to make it easier to see):



That's not to say there's never a time to use transparency, but I do wonder whether there may be other potential issues I haven't discovered yet.

Also, it would be nice if for the sake of consistency we could use a color that's used for dividers elsewhere. I know taken alone it's hard to tell if the divider color used here is the same as elsewhere, but I think if we are not consistent, the differences will start adding up until the application doesn't feel like it has a single consistent theme.

RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009)* status: all files reviewed, 7 unresolved discussions (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-global-vars.scss` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8iVH39HxX8XMXoROx:-Nc8iVH39HxX8XMXoROy:b-p5unyf) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-global-vars.scss#L3)):*
> ```SCSS
> @import 'src/variables';
> 
> $questionReadBGColor: #f9fcff;
> ```

I don't actually think the background color is necessary, the bold text is already enough to tell me that I have not read a question.
If we do remove the background colour, I like the highlight to be the same as the colour for inactive menu icons.




_Code quote:_
```SCSS
$questionReadBGColor: #f9fcff;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 4 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8kiVc0u8nGpbqvZC1:-Nc8kiVc0u8nGpbqvZC2:b-skmz5n) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L4)):*
> ```HTML
>   <div id="projects-app" class="flex-max">
>     <div id="checking-app-container" class="d-flex flex-fill flex-nowrap">
>       <div id="questions-panel" [ngClass]="{ 'overlay-visible': isQuestionsOverlayVisible }">
> ```

The fact that the question panel on mobile view overlays the entire screen is not a big issue, but it doesn't look good when the screen size is small like on a tablet. Is there a way to limit the size of the overlay so it looks a little better and only covers part of the screen?



_Code quote:_
```HTML
<div id="questions-panel" [ngClass]="{ 'overlay-visible': isQuestionsOverlayVisible }">
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.html` line 18 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8pfCUEnNJjyJGsKrs:-Nc8pfCUEnNJjyJGsKrt:b-kxkpg5) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.html#L18)):*
> ```HTML
>         <span class="question-text" *ngIf="questionDoc.data?.text">{{ questionDoc.data?.text }}</span>
>       </div>
>       <a class="view-answers" title="{{ 'checking.view_questions' | transloco }}">
> ```

We should probably change this to view answers, not view questions.



_Code quote:_
```HTML
'checking.view_questions'
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.html` line 21 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8ppKm0euMal32R0Po:-Nc8ppKm0euMal32R0Pp:bvzhp8g) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.html#L21)):*
> ```HTML
>         <mat-icon
>           class="mirror-rtl"
>           [matBadge]="getUnreadAnswers(questionDoc)"
> ```

I am so happy to see material implement this.



_Code quote:_
```HTML
    [matBadge]="getUnreadAnswers(questionDoc)"
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss` line 29 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8rALA7wrVLVWlqcEP:-Nc8rALA7wrVLVWlqcEQ:by68j8r) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss#L29)):*
> ```SCSS
>     border-bottom: 1px solid #64798f1f;
> 
>     ::ng-deep {
> ```

We should ensure that this only applies to the host component by using `:host` before ng-deep otherwise this style will leak out to other components.



_Code quote:_
```SCSS
    ::ng-deep {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss` line 110 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8rvcSCZOZOSE8IplM:-Nc8rvcSCZOZOSE8IplN:b-au9g7u) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss#L110)):*
> ```SCSS
>           background-color: vars.$questionAnswerBubbleBackgroundColor;
>           color: vars.$questionAnswerBubbleColor;
>         }
> ```

It looks like this can be removed.



_Code quote:_
```SCSS
      ::ng-deep {
        .mat-badge-content {
          background-color: vars.$questionAnswerBubbleBackgroundColor;
          color: vars.$questionAnswerBubbleColor;
        }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 89 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8sdiQ7JDPuzmru1s3:-Nc8sdiQ7JDPuzmru1s4:b-lu94gm) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L89)):*
> ```XML
> 
>   @Input()
>   set questionDocs(questionDocs: Readonly<QuestionDoc[]> | undefined) {
> ```

Nit: Setters should actually go after the constructor. Is there a reason for putting them up here?



_Code quote:_
```XML
  set questionDocs(questionDocs: Readonly<QuestionDoc[]> | undefined) {
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009)* status: all files reviewed, 7 unresolved discussions (waiting on @RaymondLuong3 and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss` line 29 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8rALA7wrVLVWlqcEP:-NcDdDeq4UWoRAhKUTLG:bk2z2ht) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss#L29)):*

We should ensure that this only applies to the host component by using `:host` before ng-deep otherwise this style will leak out to other components.
</blockquote></details>

Actually it will not leak. It's already within other selectors. What's important isn't that `:host` comes before it, but that `::ng-deep` isn't the first item in the selector. However, sometimes the thing you want to style is at the root of the component, or you just don't want to select anything else, so using `:host ::ng-deep` makes sense.

``` scss
// bad
::ng-deep .some .thing {
}

// good
.something ::ng-deep {
}

// if you don't want to select .something
:host ::ng-deep {
}
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009)* status: all files reviewed, 6 unresolved discussions (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss` line 29 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8rALA7wrVLVWlqcEP:-NcNmCdt5xIQx5TVQc17:bi1epee) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss#L29)):*

Actually it will not leak. It's already within other selectors. What's important isn't that `:host` comes before it, but that `::ng-deep` isn't the first item in the selector. However, sometimes the thing you want to style is at the root of the component, or you just don't want to select anything else, so using `:host ::ng-deep` makes sense.

``` scss
// bad
::ng-deep .some .thing {
}

// good
.something ::ng-deep {
}

// if you don't want to select .something
:host ::ng-deep {
}
```
</blockquote></details>

Oh, that is good to know. I was not aware that it was safe to use it after other selectors. Thanks for catching that.

siltomato (commented):
It should be 300px, which is wider than the left nav menu, but in the screenshot, it looks smaller.  I couldn't reproduce this.  What is set for the width when you inspect?

Also, in addition to the changes noted below, I swapped out the reference to `MediaObserver` from `@angular/flex-layout` for `BreakpointObserver` from '@angular/cdk/layout'.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-:-NcP9pAC0nFMy5IANz-6:b-cbqj05)* status: 4 of 10 files reviewed, 6 unresolved discussions (waiting on @input, @InputOutput, @Nateowami, and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-global-vars.scss` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8iVH39HxX8XMXoROx:-NcO-x0RDztYIuDrRIEK:ba628ua) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-global-vars.scss#L3)):*

I don't actually think the background color is necessary, the bold text is already enough to tell me that I have not read a question.
If we do remove the background colour, I like the highlight to be the same as the colour for inactive menu icons.

</blockquote></details>

What do you think about a slight contrast for read questions something like gmail?  I lightened it some and made it more gray.

Also, is it $purpleMedium (#5a465f) that you are thinking for the 'selected' highlight?  In the screenshot above, I think you've got something a little different, maybe #5a565f (more gray, less purple).  Personally, I like a little splash of color (like the #5a465f menu icons), but both look pretty good I think.  Which did you have in mind?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 4 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8kiVc0u8nGpbqvZC1:-NcPA7Jn-88TOwLgh7qW:bvm685h) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L4)):*

The fact that the question panel on mobile view overlays the entire screen is not a big issue, but it doesn't look good when the screen size is small like on a tablet. Is there a way to limit the size of the overlay so it looks a little better and only covers part of the screen?
</blockquote></details>

Currently the media breakpoint is set at 767.98px to make the question list full screen.  Do you think this is too wide?  What screen width do you think would be the limit for a full screen list?

If I remember right, it doesn't seem out of the ordinary for a tablet experience to have a full screen selection list.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.html` line 18 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8pfCUEnNJjyJGsKrs:-NcNz6hZ7FnI6qYyQy3-:b6zrrkh) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.html#L18)):*

We should probably change this to view answers, not view questions.
</blockquote></details>

Good catch.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.html` line 21 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8ppKm0euMal32R0Po:-NcNxBdNAafwXuz88Apz:bb77uf3) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.html#L21)):*

I am so happy to see material implement this.
</blockquote></details>

Yes!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss` line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc86oqR6o_BLoEoq9nK:-NcOaROl2ko7abAA04jj:blkmysc) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss#L27)):*

I started this code review by checking out the branch and running it. Overall looks quite good.

I'm going to start out by complaining about the most trivial thing possible though:


The reason for it is the transparency in the border bottom color (so it should be a simple fix).

In general I don't like using transparency to specify a color. It's always felt slightly wrong to me, but until recently I didn't have a clear reason why. In certain scripts (such as Arabic), the glyphs can actually overlap, and a partly transparent text color ends up looking very bad (zoomed in to make it easier to see):



That's not to say there's never a time to use transparency, but I do wonder whether there may be other potential issues I haven't discovered yet.

Also, it would be nice if for the sake of consistency we could use a color that's used for dividers elsewhere. I know taken alone it's hard to tell if the divider color used here is the same as elsewhere, but I think if we are not consistent, the differences will start adding up until the application doesn't feel like it has a single consistent theme.
</blockquote></details>

Good eyes!  Changed to use inset box-shadow instead of border to avoid border showing under the 'answered question' marker.  Neat technique I copied from gmail.

I see what you mean about transparent colors for text.  That is definitely good to file away.  I left the alpha channel here, as it allows the background color to show through, which improves the separator look for 'selected' and 'read' backgrounds (also gmail technique).

Regarding consistent divider color, I could go with rgba(0, 0, 0, 0.12), which is used for dividers elsewhere in the app.  It's a little darker.  The lighter color looks better here to my eyes (a little more subtle), but I don't think rgba(0, 0, 0, 0.12) looks bad either.  What do you think?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss` line 110 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8rvcSCZOZOSE8IplM:-NcOkjGh-erjlj5sLW69:b-wdj8zi) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss#L110)):*

It looks like this can be removed.
</blockquote></details>

Nice, thanks!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 89 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8sdiQ7JDPuzmru1s3:-NcNpvLf8agV5Fv8kN4N:b-l04cd) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L89)):*

Nit: Setters should actually go after the constructor. Is there a reason for putting them up here?
</blockquote></details>

Even if it's an @Input?  All the examples that I can find put the @Input/@Output above the constructor.

RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009)* status: all files reviewed, 5 unresolved discussions (waiting on @input, @InputOutput, @Nateowami, and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-global-vars.scss` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8iVH39HxX8XMXoROx:-NcT-h_tBV1jzXqW5Rgn:bxe49k4) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-global-vars.scss#L3)):*

What do you think about a slight contrast for read questions something like gmail?  I lightened it some and made it more gray.

Also, is it $purpleMedium (#5a465f) that you are thinking for the 'selected' highlight?  In the screenshot above, I think you've got something a little different, maybe #5a565f (more gray, less purple).  Personally, I like a little splash of color (like the #5a465f menu icons), but both look pretty good I think.  Which did you have in mind?
</blockquote></details>

Ok, thanks for making those changes. I like the look much better. We want to minimize non-palette colour to keep our app clean.
You're right, I remember picking a standard colour for the screenshot, but it does look more grey. But what you have currently works for me. I posted this in slack and maybe others will have an opinion.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 4 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8kiVc0u8nGpbqvZC1:-NcT32nOAp3uVzCmmml6:b-g4d2rw) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L4)):*

Currently the media breakpoint is set at 767.98px to make the question list full screen.  Do you think this is too wide?  What screen width do you think would be the limit for a full screen list?

If I remember right, it doesn't seem out of the ordinary for a tablet experience to have a full screen selection list.
</blockquote></details>

I looked at the material design explanation, and it appears that the recommendation is that 600px or greater is where they recommend a navigation rail. https://m3.material.io/components/navigation-drawer/guidelines#56b3fea0-ecc6-4ff0-bb0a-80ff48c74a58 I don't think we really need to do that, but I would say 768px is too wide for a full screen drawer. Is it possible to make the drawer more modal, and limit the width to 360px?
This is not a blocker for me. Just a suggestion.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 521 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NcT4D5W5mxQWYlg7l_t:-NcT4D5XBPYofxLo-8R9:bhal2y2) ([raw file](https://github.com/sillsdev/web-xforge/blob/143ba6a24cffe030db6dc37247ef63408961a201/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L521)):*
> ```XML
> 
>     // Allows scrolling to the active question in the question list once it becomes visible
>     this.subscribe(this.breakpointObserver.observe(['(min-width: 767.98px)']), (state: BreakpointState) => {
> ```

Is it possible to set the breakpoints somewhere, maybe in utils? I wonder if we could even put breakpoint observables into `utils.spec` so it is available for all components that need to subscribe to breakpoints.



_Code quote:_
```XML
'(min-width: 767.98px)'
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 89 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8sdiQ7JDPuzmru1s3:-NcT91e36-q0nnImL4vX:b-8emzyh) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L89)):*

Even if it's an @Input?  All the examples that I can find put the @Input/@Output above the constructor.
</blockquote></details>

Yes, you can see this in `checking-answers.component.ts`.

Nateowami (commented):
I just shrunk the viewport to the smallest point where the sidebar wouldn't disappear.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-:-NcTw5QoDCG4rzPoQnb8:b-lznikc)* status: all files reviewed, 5 unresolved discussions (waiting on @input, @InputOutput, @RaymondLuong3, and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 521 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NcT4D5W5mxQWYlg7l_t:-NcTtYGmChKWpquWSFhD:b-y41nsy) ([raw file](https://github.com/sillsdev/web-xforge/blob/143ba6a24cffe030db6dc37247ef63408961a201/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L521)):*

Is it possible to set the breakpoints somewhere, maybe in utils? I wonder if we could even put breakpoint observables into `utils.spec` so it is available for all components that need to subscribe to breakpoints.
</blockquote></details>

I was going to say the same thing even before seeing Raymond's comment.

I don't think we should be hard coding values like this. I think we should have a service somewhere that defines screen sizes such as xs, sm, md, lg, xl, and then you can observe when those states are active.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 89 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8sdiQ7JDPuzmru1s3:-NcTv931BUHYi6eHCm9j:b-smonjc) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L89)):*

Yes, you can see this in `checking-answers.component.ts`.
</blockquote></details>

...I don't think that's what we've been doing. Does our style guide address this?

I personally think having inputs specified up front improves readability.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009)* status: all files reviewed, 5 unresolved discussions (waiting on @input, @InputOutput, @Nateowami, and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 89 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8sdiQ7JDPuzmru1s3:-NcU5GuZAzmcc3F2xq9l:b-7x221l) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L89)):*

...I don't think that's what we've been doing. Does our style guide address this?

I personally think having inputs specified up front improves readability.
</blockquote></details>

I am in favour of moving inputs/outputs to the top. We should try to remember that so that we can keep our other files consistent with that.

siltomato (commented):
I watched the video you posted in slack, and I still couldn't reproduce it in firefox or brave.  Crazy.  Added `flex-shrink: 0`.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-:-NcZXYF057PrdLAnoHfR:b-91v0sm)* status: 4 of 14 files reviewed, 2 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-global-vars.scss` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8iVH39HxX8XMXoROx:-Nc_9oOmBwevjgjhEOgh:b-6v7n5s) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-global-vars.scss#L3)):*

Ok, thanks for making those changes. I like the look much better. We want to minimize non-palette colour to keep our app clean.
You're right, I remember picking a standard colour for the screenshot, but it does look more grey. But what you have currently works for me. I posted this in slack and maybe others will have an opinion.
</blockquote></details>

I tried something a little more muted, like Nathanial mentioned in Slack.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 4 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8kiVc0u8nGpbqvZC1:-Ncf0DHaAXZZUs360DFs:b64l9ym) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L4)):*

I looked at the material design explanation, and it appears that the recommendation is that 600px or greater is where they recommend a navigation rail. https://m3.material.io/components/navigation-drawer/guidelines#56b3fea0-ecc6-4ff0-bb0a-80ff48c74a58 I don't think we really need to do that, but I would say 768px is too wide for a full screen drawer. Is it possible to make the drawer more modal, and limit the width to 360px?
This is not a blocker for me. Just a suggestion.
</blockquote></details>

Thanks for that link.  I reduced the breakpoint to 600px, but I left the questions list full screen, as I think that is what Nathaniel had in mind?  Also, I added a snap-back transition effect when the question list is shown/hidden.  Maybe too much?  What do you think?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 521 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NcT4D5W5mxQWYlg7l_t:-NcewrNr9dZJS-Al0ax0:b-y9n50) ([raw file](https://github.com/sillsdev/web-xforge/blob/143ba6a24cffe030db6dc37247ef63408961a201/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L521)):*

I was going to say the same thing even before seeing Raymond's comment.

I don't think we should be hard coding values like this. I think we should have a service somewhere that defines screen sizes such as xs, sm, md, lg, xl, and then you can observe when those states are active.
</blockquote></details>

Ok, I created a service to do this.  There is a case where you want a breakpoint that is not in the defined sizes, and you want to coordinate a css media query and a `BreakpointObserver`.  I wrote `MediaBreakpointService` to do this in combination with the scss mixin defined in `xforge-common/media-breakpoints/_breakpoints.scss`.  This makes it possible to override a breakpoint value like 'sm' in the component stylesheet, and the `BreakpointObserver` for the corresponding element in that component that is observing 'SM' will adjust.

The coordination is done via the breakpoints getting written to css variables and the `MediaBreakpointService` reading the css variable values via `window.getComputedStyle()`.

Hopefully this is not overkill, but it maintains a single source of truth for the breakpoints and avoids manual sync of the hard-coding in two locations in the case described above.

RaymondLuong3 (changes requested):
> Reviewed 10 of 10 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009)* status: all files reviewed, 8 unresolved discussions (waiting on @Nateowami and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 2 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nchbmor5jTP3n1qgc6G:-Nchbmor5jTP3n1qgc6H:b-pcz0nk) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L2)):*
> ```SCSS
> @use 'variables';
> @use 'src/xforge-common/media-breakpoints/css-vars';
> ```

Is this being used?



_Code quote:_
```SCSS
@use 'src/xforge-common/media-breakpoints/css-vars';
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-global-vars.scss` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8iVH39HxX8XMXoROx:-Nchc2aL6RQtL2GFJvTQ:b5dofh7) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-global-vars.scss#L3)):*

I tried something a little more muted, like Nathanial mentioned in Slack.
</blockquote></details>

I like it. Thanks.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 4 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8kiVc0u8nGpbqvZC1:-Nchm_Bv8qvZ-TK05nS7:bpfzeq8) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L4)):*

Thanks for that link.  I reduced the breakpoint to 600px, but I left the questions list full screen, as I think that is what Nathaniel had in mind?  Also, I added a snap-back transition effect when the question list is shown/hidden.  Maybe too much?  What do you think?
</blockquote></details>

Yea, the full screen thing was minor. I'm happy for it to be full screen, but I will miss the nice modal effect we had before.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 6 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nchesv6DIctVQYGWQ3t:-Nchesv6DIctVQYGWQ3u:br0exhe) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L6)):*
> ```SCSS
> @import 'bootstrap/scss/mixins/breakpoints';
> 
> $questionsPanelBreakpoint: 600px;
> ```

This value is just somehwere between our small and medium breakpoints. I can see it does look like a good choice. It would be good to leave a comment of why you chose 600px?



_Code quote:_
```SCSS
$questionsPanelBreakpoint: 600px;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 69 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NchdIPTE6yxE66uhMN2:-NchdIPTE6yxE66uhMN3:bcxa88y) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L69)):*
> ```SCSS
>   // Expand the side and bottom to fill space
>   @include media-breakpoint('<', md) {
>     margin-inline-start: -20px;
> ```

I'm not entirely sure the question panel needs to be right up against the left edge of the screen, but not a big deal.



_Code quote:_
```SCSS
    margin-inline-start: -20px;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 84 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NchdivHECzW8WbQX0Xi:-NchdivHECzW8WbQX0Xj:b-kr9nnh) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L84)):*
> ```SCSS
>     z-index: 7;
>     background: #fff;
>     transition: left 0.35s cubic-bezier(0.68, -0.55, 0.265, 1.55); // Snap back effect
> ```

Neat trick. I like that it gives the effect that the menu is coming on top of the scripture panel.



_Code quote:_
```SCSS
    transition: left 0.35s cubic-bezier(0.68, -0.55, 0.265, 1.55); // Snap back effect
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 217 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NchfPfL-aJ5F0jy9Me5:-NchfPfL-aJ5F0jy9Me6:b-m75ca) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L217)):*
> ```SCSS
> 
>     // Widen once the questions panel is hidden
>     @include media-breakpoint('<=', sm, $questionsPanelBreakpoint) {
> ```

Nice work figuring this one out. It took me a while to even know what this was trying to do.



_Code quote:_
```SCSS
    @include media-breakpoint('<=', sm, $questionsPanelBreakpoint) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 521 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NcT4D5W5mxQWYlg7l_t:-Nchmt7w597gF8aaPTct:bmtdc8s) ([raw file](https://github.com/sillsdev/web-xforge/blob/143ba6a24cffe030db6dc37247ef63408961a201/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L521)):*

Ok, I created a service to do this.  There is a case where you want a breakpoint that is not in the defined sizes, and you want to coordinate a css media query and a `BreakpointObserver`.  I wrote `MediaBreakpointService` to do this in combination with the scss mixin defined in `xforge-common/media-breakpoints/_breakpoints.scss`.  This makes it possible to override a breakpoint value like 'sm' in the component stylesheet, and the `BreakpointObserver` for the corresponding element in that component that is observing 'SM' will adjust.

The coordination is done via the breakpoints getting written to css variables and the `MediaBreakpointService` reading the css variable values via `window.getComputedStyle()`.

Hopefully this is not overkill, but it maintains a single source of truth for the breakpoints and avoids manual sync of the hard-coding in two locations in the case described above.
</blockquote></details>

I can see why you wanted to make an override for the breakpoint sizes possible. I might be a little overkill, but I think that the cost to maintain it will be worth the benefit.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 531 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NchhhtE7S4vK-u5AX8h:-NchhhtE7S4vK-u5AX8i:b-ujmoqp) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L531)):*
> ```XML
>       ]),
>       (state: BreakpointState) => {
>         this.calculateScriptureSliderPosition();
> ```

I would have thought that



_Code quote:_
```XML
        this.calculateScriptureSliderPosition();
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/media-breakpoints/_breakpoints.scss` line 41 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nchj5HU4BpeoV7NlUFM:-Nchj5HV6cdF7nsN5Jt7:b-4jzbhh) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/media-breakpoints/_breakpoints.scss#L41)):*
> ```SCSS
>  * Setting this css variable allows the MediaBreakpointService to get the overridden value of the breakpoint.
>  */
> @mixin media-breakpoint($operator, $breakpointName, $override: null) {
> ```

This is pretty cool. Did you learn this online, or was it something you already knew how to do?



_Code quote:_
```SCSS
@mixin media-breakpoint($operator, $breakpointName, $override: null) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/media-breakpoints/media-breakpoint.service.ts` line 47 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NchlcSP1SyZwpzgY06W:-NchlcSP1SyZwpzgY06X:b-h08g5i) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/media-breakpoints/media-breakpoint.service.ts#L47)):*
> ```XML
>    * ```scss
>    * :root { --sf-breakpoint-sm: '576px'}  // Must be set at root level
>    * element-selector { --sf-breakpoint-sm: '600px' }  // Optional override at component level
> ```

Ok, if I understand this correctly, the breakpoint values for the app at the root element is set, and then each component has the option to override the default. That is a neat way to do it. Sorry you had to go through all that work for this service. Hopefully this becomes a useful service.



_Code quote:_
```XML
   * :root { --sf-breakpoint-sm: '576px'}  // Must be set at root level
   * element-selector { --sf-breakpoint-sm: '600px' }  // Optional override at component level
```

siltomato (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 2 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nchbmor5jTP3n1qgc6G:-NciFrRt7xAZ0OZj3PK0:b-7f2ohl) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L2)):*

Is this being used?
</blockquote></details>

Yes, this line is what includes the `--sf-breakpoint-<name>` css variables into the main stylesheet.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 4 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8kiVc0u8nGpbqvZC1:-NciEXXtA0wXWY4LeCDo:b-jur215) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L4)):*

Yea, the full screen thing was minor. I'm happy for it to be full screen, but I will miss the nice modal effect we had before.
</blockquote></details>

Yes, the drawer effect is cool.  I added the snap-back effect to add some candy back into it, and I'm open to other ideas as well.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 6 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nchesv6DIctVQYGWQ3t:-NciENRiEtvCkUTSAjnH:bpjx23j) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L6)):*

This value is just somehwere between our small and medium breakpoints. I can see it does look like a good choice. It would be good to leave a comment of why you chose 600px?
</blockquote></details>

Added a comment to reference the material design link you mentioned.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 69 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NchdIPTE6yxE66uhMN2:-NciD1TABIioBEsnrhmc:b-ncwoss) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L69)):*

I'm not entirely sure the question panel needs to be right up against the left edge of the screen, but not a big deal.
</blockquote></details>

Yeah, I was trying to match with what the question navigation was doing on the other side of the screen.  I figure it gives just a little more space to the text viewer.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 84 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NchdivHECzW8WbQX0Xi:-NciCvYK3WtJw5Qj3Jza:b-ucfknh) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L84)):*

Neat trick. I like that it gives the effect that the menu is coming on top of the scripture panel.
</blockquote></details>

Thanks!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 217 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NchfPfL-aJ5F0jy9Me5:-NciCh4Y2z58tp9SpUut:b-lt4p2u) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L217)):*

Nice work figuring this one out. It took me a while to even know what this was trying to do.
</blockquote></details>

Thanks! Let me know if a change in syntax or extra documentation would be useful.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 521 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NcT4D5W5mxQWYlg7l_t:-NciCUdZAK8gPGlUbqSy:bw8fzik) ([raw file](https://github.com/sillsdev/web-xforge/blob/143ba6a24cffe030db6dc37247ef63408961a201/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L521)):*

I can see why you wanted to make an override for the breakpoint sizes possible. I might be a little overkill, but I think that the cost to maintain it will be worth the benefit.
</blockquote></details>

Yes, I hope it works as expected =).
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 531 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NchhhtE7S4vK-u5AX8h:-NciCJNA6Ug7Hk7RaNvn:bgz61ce) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L531)):*

I would have thought that
</blockquote></details>

I think your thought got cut off!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/media-breakpoints/_breakpoints.scss` line 41 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nchj5HU4BpeoV7NlUFM:-Nci6TRC0_hYhaHqKWze:b-hmouhq) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/media-breakpoints/_breakpoints.scss#L41)):*

This is pretty cool. Did you learn this online, or was it something you already knew how to do?
</blockquote></details>

Learn how to create a sass mixin or this particular mixin online?  This particular mixin I created to solve the problem at hand.  It's a similar idea that we were already doing with the bootstrap mixin, but with the added override that affects `MediaBreakpointService` generated queries due to the changed css variable for the element.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/media-breakpoints/media-breakpoint.service.ts` line 47 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NchlcSP1SyZwpzgY06W:-Nci7qRKAZwjOdw4v_Oj:b-7qtmkv) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/media-breakpoints/media-breakpoint.service.ts#L47)):*

Ok, if I understand this correctly, the breakpoint values for the app at the root element is set, and then each component has the option to override the default. That is a neat way to do it. Sorry you had to go through all that work for this service. Hopefully this becomes a useful service.
</blockquote></details>

Yes, that's right.  A big benefit to the solution is that overrides can happen in one place as long as `BreakpointObserver` uses `MediaBreakpointService` to generate the query for the element with the override.  I think this can be useful where we are currently using `MediaObserver` from the deprecated `@angular/flex-layout`.

RaymondLuong3 (changes requested):
I looked over the changes and everything looks good. Just one suggestion to add a comment, then I'll let Nathaniel go over it one more time.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-:-NcyIVqC952QxhlQStJl:b-86yo6a)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 2 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nchbmor5jTP3n1qgc6G:-NcyIM3S0FsOUUOIyEDM:b-3xq8zr) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L2)):*

Yes, this line is what includes the `--sf-breakpoint-<name>` css variables into the main stylesheet.
</blockquote></details>

Ok, could you leave a comment since it is not obvious.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 531 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NchhhtE7S4vK-u5AX8h:-NcyHzUc4EYQISNNOHfL:bwtv6g5) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L531)):*

I think your thought got cut off!
</blockquote></details>

Oh yes, I was leaving a comment, but then realized this is what I expected to be here. You can ignore this.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/media-breakpoints/_breakpoints.scss` line 41 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nchj5HU4BpeoV7NlUFM:-NcyI91t-o0vCO6cFVos:b2uzpt2) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/media-breakpoints/_breakpoints.scss#L41)):*

Learn how to create a sass mixin or this particular mixin online?  This particular mixin I created to solve the problem at hand.  It's a similar idea that we were already doing with the bootstrap mixin, but with the added override that affects `MediaBreakpointService` generated queries due to the changed css variable for the element.
</blockquote></details>

I mean this particular mixin. Nice.

siltomato (commented):
Ok, thanks!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-:-NcyMAQR2_GMlZXEjV1a:b-54avj1)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 2 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nchbmor5jTP3n1qgc6G:-NcyLGrMDu_BkmMoup85:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/0858d50ee1384c8510fdc06851d7fdd997f1ad21/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L2)):*

Ok, could you leave a comment since it is not obvious.
</blockquote></details>

Done.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009)* status: 13 of 14 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 89 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8sdiQ7JDPuzmru1s3:-NcU5rD87x205-ZYqicc:bvk3jv3) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L89)):*

I am in favour of moving inputs/outputs to the top. We should try to remember that so that we can keep our other files consistent with that.
</blockquote></details>

OK, actually Raymond is right that we've been keeping setters after the constructor in most cases. I didn't think through this very clearly. I don't have a strong opinion.

siltomato (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009)* status: 13 of 14 files reviewed, 2 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 89 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc8sdiQ7JDPuzmru1s3:-NczDcn8CLNtAe7oAOV7:b-czkop4) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L89)):*

OK, actually Raymond is right that we've been keeping setters after the constructor in most cases. I didn't think through this very clearly. I don't have a strong opinion.
</blockquote></details>

I didn't see anything specifically about it in our wiki style guide other than the reference to the angular style guide, in which I couldn't find explicit guidance, so I figured I'd look in the angular material repo to see how they did it.

[Here's a link for the angular material `MatDrawer` component](https://github.com/angular/components/blob/main/src/material/sidenav/drawer.ts).  All the \@Input/Output properties, including getter/setter props are above the constructor.

Nateowami (dismissed):
@siltomato Is this ready for testing and then merging?

> Reviewed 1 of 10 files at r3, 1 of 2 files at r7.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-:-NdGtF6D4XkrDlxBTp67:b-jy71xj)* status: 11 of 14 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss` line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-Nc86oqR6o_BLoEoq9nK:-NdGt7e8A9qOHYrd-gSo:b-176s13) ([raw file](https://github.com/sillsdev/web-xforge/blob/a48909b18e5235dff99e1622f23d6bc063b8d5a3/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss#L27)):*

Good eyes!  Changed to use inset box-shadow instead of border to avoid border showing under the 'answered question' marker.  Neat technique I copied from gmail.

I see what you mean about transparent colors for text.  That is definitely good to file away.  I left the alpha channel here, as it allows the background color to show through, which improves the separator look for 'selected' and 'read' backgrounds (also gmail technique).

Regarding consistent divider color, I could go with rgba(0, 0, 0, 0.12), which is used for dividers elsewhere in the app.  It's a little darker.  The lighter color looks better here to my eyes (a little more subtle), but I don't think rgba(0, 0, 0, 0.12) looks bad either.  What do you think?
</blockquote></details>

@siltomato I'm sorry; I forgot that I had left this as a blocker and then not come back to resolve it.

siltomato (commented):
Yes.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-:-NdHhoQoBilXUHLlF167:bb77ufg)* status: 11 of 14 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

siltomato (commented):
Added css rules to handle rtl environments for the questions overlay.  This addresses issues found by testers.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-:-NdgUTesBTLy9mBzT3Rz:bmadjx8)* status: 5 of 14 files reviewed, all discussions resolved (waiting on @Nateowami and @RaymondLuong3)

RaymondLuong3 (approved):
> Reviewed 1 of 2 files at r7, 2 of 2 files at r8, 6 of 6 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009)* status: all files reviewed, 1 unresolved discussion (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 107 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NdgdDoiBRRoISQ9k_gV:-NdgdDoiBRRoISQ9k_gW:bfbui7b) ([raw file](https://github.com/sillsdev/web-xforge/blob/e221d02d07260ae9930c652bfc3d74f7924216c5/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L107)):*
> ```SCSS
>       left: unset;
>       right: -100%;
>       @include snapBack(right);
> ```

The snapback effect doesn't quite work on RTL. The panel appears and then closes, without the snapback. I wouldn't say it is a blocker, since the snapback is just a nice touch, but it would be nice if we could get it to work.



_Code quote:_
```SCSS
      @include snapBack(right);
```

siltomato (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009)* status: 13 of 14 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 107 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NdgdDoiBRRoISQ9k_gV:-Ndgtibg3ob4J-GBACgz:b-glm8n6) ([raw file](https://github.com/sillsdev/web-xforge/blob/e221d02d07260ae9930c652bfc3d74f7924216c5/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L107)):*

The snapback effect doesn't quite work on RTL. The panel appears and then closes, without the snapback. I wouldn't say it is a blocker, since the snapback is just a nice touch, but it would be nice if we could get it to work.
</blockquote></details>

I switched the transition to be the `translateX` transform instead of `left`, which seems to work better.

RaymondLuong3 (approved):
> Reviewed 4 of 4 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2009)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 107 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2009#-NdgdDoiBRRoISQ9k_gV:-NdkN8Fy0mPePQQ5W3T4:b-i39vd2) ([raw file](https://github.com/sillsdev/web-xforge/blob/e221d02d07260ae9930c652bfc3d74f7924216c5/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L107)):*

I switched the transition to be the `translateX` transform instead of `left`, which seems to work better.
</blockquote></details>

Yep, it looks better now.

PR #2013, opened 2023-08-22
SF-2193 Add bubble splash effect that can be added to various buttons

siltomato:
Prototype bubble splash effect to add to various buttons as an angular directive.  Part of gamification initiative to add visual rewards for app actions.  The directive can be added to angular material buttons.  Currently, the angular directive is only in Storybook for preview and feedback.
```html
<button mat-flat-button sfBubbleButton color="primary">Mat flat button</button>
```
Sample buttons can be seen in Storybook under 'Misc/Bubble Button'.

**Note:**
- Added npm packages `sass-loader`, `css-loader`, and `style-loader` to be able to load the directive stylesheet in its story.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2013)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.directive.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.directive.ts
@@ -1,0 +1,45 @@
+/* eslint-disable @angular-eslint/directive-selector */
+import { Directive, ElementRef, OnInit, Renderer2 } from '@angular/core';
+
+/**
+ * Directive to add bubble animation to a button. Inspired by https://codepen.io/nourabusoud/pen/ypZzMM,
+ * but modified to use an inner `<span>` to attach the pseudo elements to so as not to conflict
+ * with other styles or components that use `::before` or `::after` such as Angular Material components.
+ */
+@Directive({
+  selector: '[sfBubbleButton]'
+})
+export class BubbleButtonDirective implements OnInit {
+  cssInnerSpanStyleClass = 'sf-bubble-button-elements';
+  cssButtonStyleClass = 'sf-bubble-button';
+  cssButtonAnimationClass = 'sf-bubble-animate';
+
+  constructor(private readonly el: ElementRef, private readonly renderer: Renderer2) {}
+
+  ngOnInit(): void {
+    const hostElement = this.el.nativeElement;
+    const innerSpan = this.renderer.createElement('span');
+
+    // Add inner span to host element
+    this.renderer.addClass(innerSpan, this.cssInnerSpanStyleClass);
+    this.renderer.appendChild(hostElement, innerSpan);
+
+    // Add class and click listener to host element
+    this.renderer.addClass(hostElement, this.cssButtonStyleClass);
+    this.renderer.listen(hostElement, 'click', () => {
+      // Add animation class to inner span
+      this.addAnimationClass(innerSpan);
+    });
+  }
+
+  // Adds animation class to the element and removes it after animation is complete
+  addAnimationClass(el: any): void {
+    // Reset animation
+    el.classList.remove(this.cssButtonAnimationClass);
+
+    // Timeout needed to restart animation
+    setTimeout(() => {
+      el.classList.add(this.cssButtonAnimationClass);
+    }, 10);
+  }
+}
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.directive.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.directive.ts
@@ -14,7 +14,10 @@ export class BubbleButtonDirective implements OnInit {
   cssButtonStyleClass = 'sf-bubble-button';
   cssButtonAnimationClass = 'sf-bubble-animate';
 
-  constructor(private readonly el: ElementRef, private readonly renderer: Renderer2) {}
+  constructor(
+    private readonly el: ElementRef,
+    private readonly renderer: Renderer2
+  ) {}
 
   ngOnInit(): void {
     const hostElement = this.el.nativeElement;
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.scss
@@ -1,0 +1,162 @@
+@use 'sass:map';
+@use 'sass:list';
+@use 'src/variables' as vars;
+
+$bubbleColor: vars.$blueMedium;
+$animationDuration: 0.7s;
+
+// Bubble types
+// 1: Large solid bubble
+// 2: Large outlined bubble
+// 3: Small solid bubble
+$bubbleTypes: (
+  1: radial-gradient(circle, $bubbleColor 20%, transparent 20%),
+  2: radial-gradient(circle, transparent 20%, $bubbleColor 20%, transparent 30%),
+  3: radial-gradient(circle, transparent 10%, $bubbleColor 15%, transparent 20%)
+);
+
+// prettier-ignore
+$bubbles: (
+  // Top bubbles
+  (location: 'top', type: 1, size:(10% 10%), pos0:( 5% 90%), pos50:( 0% 80%), pos100:( 0%  70%)),
+  (location: 'top', type: 2, size:(20% 20%), pos0:(10% 90%), pos50:( 0% 20%), pos100:( 0%  10%)),
+  (location: 'top', type: 1, size:(15% 15%), pos0:(10% 90%), pos50:(10% 40%), pos100:(10%  30%)),
+  (location: 'top', type: 1, size:(20% 20%), pos0:(15% 90%), pos50:(20%  0%), pos100:(20% -10%)),
+  (location: 'top', type: 3, size:(18% 18%), pos0:(25% 90%), pos50:(30% 30%), pos100:(30%  20%)),
+  (location: 'top', type: 1, size:(10% 10%), pos0:(25% 90%), pos50:(22% 50%), pos100:(22%  40%)),
+  (location: 'top', type: 1, size:(15% 15%), pos0:(40% 90%), pos50:(50% 50%), pos100:(50%  40%)),
+  (location: 'top', type: 1, size:(10% 10%), pos0:(55% 90%), pos50:(65% 20%), pos100:(65%  10%)),
+  (location: 'top', type: 1, size:(18% 18%), pos0:(70% 90%), pos50:(90% 30%), pos100:(90%  20%)),
+  // Bottom bubbles
+  (location: 'bottom', type: 1, size:(15% 15%), pos0:(10% -10%), pos50:(  0%  80%), pos100:(  0%  90%)),
+  (location: 'bottom', type: 1, size:(20% 20%), pos0:(30%  10%), pos50:( 20%  80%), pos100:( 20%  90%)),
+  (location: 'bottom', type: 3, size:(18% 18%), pos0:(55% -10%), pos50:( 45%  60%), pos100:( 45%  70%)),
+  (location: 'bottom', type: 1, size:(20% 20%), pos0:(70% -10%), pos50:( 60% 100%), pos100:( 60% 110%)),
+  (location: 'bottom', type: 1, size:(15% 15%), pos0:(85% -10%), pos50:( 75%  70%), pos100:( 75%  80%)),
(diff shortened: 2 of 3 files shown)


Nateowami (commented):
Overall looks good. A couple comments.

Curious if you considered using a third-party library to simplify the creation of this? See e.g. the feature/confetti branch from a few months ago.

Was it intentional that the size of bubbles seems to be scaled relative to button size? I found that a bit odd. I thought if it was necessary to scale the size of the buttons, that would make sense as a parameter.

**I am not taking responsibility for reviewing this, just swinging by to add a few comments.**

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2013#-:-NcTntUT4JrhJpuQJYsc:b-ewqtls)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.directive.ts` line 11 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2013#-NcTpduu0DFIWAg6Fa8W:-NcTpduu0DFIWAg6Fa8X:bv9scym) ([raw file](https://github.com/sillsdev/web-xforge/blob/76b45087f81d28b91b1cfcaad0a84ed4eb606d00/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.directive.ts#L11)):*
> ```XML
> @Directive({
>   selector: '[sfBubbleButton]'
> })
> ```

Does Angular not allow specifying CSS for a directive?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.stories.ts` line 26 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2013#-NcToE2u0GF0qlXI05bb:-NcToE2u0GF0qlXI05bc:bcpj61w) ([raw file](https://github.com/sillsdev/web-xforge/blob/76b45087f81d28b91b1cfcaad0a84ed4eb606d00/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.stories.ts#L26)):*
> ```XML
> type Story = StoryObj;
> 
> export const MatRaisedButton: Story = {
> ```

There are a lot of stories here that all do the same thing in just about the same way (so far as I can tell). Chromatic bills us per screenshot. We're not going to go broke due to having six more screenshots than necessary, but if we consistently create several times as many as really necessary, we're likely to put ourselves in a position where we're forced to upgrade to a plan with more screenshots.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.stories.ts` line 28 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2013#-NcTpLdoAMzfMVHVPBmQ:-NcTpLdoAMzfMVHVPBmR:b-xhwjg2) ([raw file](https://github.com/sillsdev/web-xforge/blob/76b45087f81d28b91b1cfcaad0a84ed4eb606d00/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.stories.ts#L28)):*
> ```XML
> export const MatRaisedButton: Story = {
>   render: () => ({
>     template: `<button mat-raised-button sfBubbleButton color="primary">Mat raised button</button>`
> ```

This is a lot more verbose than necessary on all these stories. Instead of specifying a render function on all of them, it seems like it would make more sense to use [args](https://storybook.js.org/docs/angular/writing-stories/args). Maybe that wouldn't work as well as I think in this situation.

marksvc (changes requested):
I like it. I noticed that the short buttons only have a little bit of celebration. I think we should decide on the amount of celebration (for example, the size of the bubbles, and how far out they fly), and make it consistent for large and small buttons.

I think it does make sense to use a 3rd party library. It seems like something that someone else would have already provided. I also haven't looked into it. I also like the bubbles. :)

I think I'm learning here rather than in a position to provide a scrutinous review. But I considered the files and made comments.

Thanks for your work!

> Reviewed 4 of 5 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2013#-:-Nd2JWCY8-1SQ_XN8CEM:b9tcb7n)* status: all files reviewed, 9 unresolved discussions (waiting on @Nateowami and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.scss` line 6 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2013#-Nd2Q0fq9EHKP1BR_gzB:-Nd2Q0fq9EHKP1BR_gzC:b-36nxwo) ([raw file](https://github.com/sillsdev/web-xforge/blob/76b45087f81d28b91b1cfcaad0a84ed4eb606d00/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.scss#L6)):*
> ```SCSS
> 
> /// Gets the min/max css function to constrain the size within a min and max size in px.
> /// @param {number} $sizePercentage - The size of the bubble as a percentage of the button size.
> ```

Yeah, maybe this could be based on something other than the size of the button, so it's uniform across buttons of varying lengths. Maybe it could be based on
- height of the button,
- height or width of the web page's viewport, or
- a constant px or em value, if that looks good on different viewports
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.scss` line 111 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2013#-Nd2T-U07tGx0Ur1aG2L:-Nd2T-U07tGx0Ur1aG2M:b-msjj2j) ([raw file](https://github.com/sillsdev/web-xforge/blob/76b45087f81d28b91b1cfcaad0a84ed4eb606d00/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.scss#L111)):*
> ```SCSS
>       right: -20%;
> 
>       // Needed to make the padding increase the height as the width increases.
> ```

I wonder if it will result in UI elements getting pushed around, if the button is in the midst of other UI elements when this happens?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.scss` line 114 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2013#-Nd2TE_LC8XHP_-fNOOP:-Nd2TE_LC8XHP_-fNOOQ:b-nhuhzl) ([raw file](https://github.com/sillsdev/web-xforge/blob/76b45087f81d28b91b1cfcaad0a84ed4eb606d00/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.scss#L114)):*
> ```SCSS
>       // Long buttons need more height for the animation so the circles don't get cut off.
>       box-sizing: content-box;
>       padding-top: 30%; // Percentage of width of containing element
> ```

Do you mean "height" instead of "width"?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.scss` line 114 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2013#-Nd2TIcJ7J0OKDf-vzQj:-Nd2TIcJ7J0OKDf-vzQk:b-9e0eao) ([raw file](https://github.com/sillsdev/web-xforge/blob/76b45087f81d28b91b1cfcaad0a84ed4eb606d00/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.scss#L114)):*
> ```SCSS
>       // Long buttons need more height for the animation so the circles don't get cut off.
>       box-sizing: content-box;
>       padding-top: 30%; // Percentage of width of containing element
> ```

Because the bubbles fly out at known range of pixels, wouldn't you want to set padding-top to close to the value of the maximum px distance that a bubble might fly out? Or what is the purpose of using a percentage of the size of the containing element?

Also, would you not also need a padding-bottom?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.scss` line 121 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2013#-Nd2TrXO5AjO7fo-wRzW:-Nd2TrXO5AjO7fo-wRzX:b9po9r8) ([raw file](https://github.com/sillsdev/web-xforge/blob/76b45087f81d28b91b1cfcaad0a84ed4eb606d00/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.scss#L121)):*
> ```SCSS
> 
>     &:before {
>       display: none;
> ```

Why is `display` configured both here and above in the following? Is that a mistake, or is there a particular behaviour that causes?
```css
    &:before,
    &:after {
      position: absolute;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.scss` line 129 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2013#-Nd2VVT20jQT0sQWgMR1:-Nd2VVT20jQT0sQWgMR2:b-fiyntz) ([raw file](https://github.com/sillsdev/web-xforge/blob/76b45087f81d28b91b1cfcaad0a84ed4eb606d00/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.scss#L129)):*
> ```SCSS
>     &:after {
>       display: none;
>       top: 65%;
> ```

(And this is not at odds with the following? )

```css
&:before,
&:after {
  ...
  height: 100%;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.stories.ts` line 26 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2013#-NcToE2u0GF0qlXI05bb:-Nd2K2sy2J-ubIeuNJeh:balruyl) ([raw file](https://github.com/sillsdev/web-xforge/blob/76b45087f81d28b91b1cfcaad0a84ed4eb606d00/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.stories.ts#L26)):*

There are a lot of stories here that all do the same thing in just about the same way (so far as I can tell). Chromatic bills us per screenshot. We're not going to go broke due to having six more screenshots than necessary, but if we consistently create several times as many as really necessary, we're likely to put ourselves in a position where we're forced to upgrade to a plan with more screenshots.
</blockquote></details>

@Nateowami, :-/ Can we disconnect some of our storybook tests from Chromatic?
I found it helpful to click thru the different UI cases and see the behaviour for them. It will be too bad if using Chromatic resulted in needing to restrict ourselves.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2013)* status: all files reviewed, 9 unresolved discussions (waiting on @marksvc and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.stories.ts` line 26 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2013#-NcToE2u0GF0qlXI05bb:-Nd2mJK--yr5vr65_ZFU:b-hx3upc) ([raw file](https://github.com/sillsdev/web-xforge/blob/76b45087f81d28b91b1cfcaad0a84ed4eb606d00/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.stories.ts#L26)):*

@Nateowami, :-/ Can we disconnect some of our storybook tests from Chromatic?
I found it helpful to click thru the different UI cases and see the behaviour for them. It will be too bad if using Chromatic resulted in needing to restrict ourselves.
</blockquote></details>

Yes, that can be done. Though I thought args would make more sense, and allow even more flexibility with experimenting with possible variations.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2013)* status: all files reviewed, 9 unresolved discussions (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.stories.ts` line 26 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2013#-NcToE2u0GF0qlXI05bb:-NdpvIaMDhkMPitdI1s3:bsg18yh) ([raw file](https://github.com/sillsdev/web-xforge/blob/76b45087f81d28b91b1cfcaad0a84ed4eb606d00/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/bubble-button/bubble-button.stories.ts#L26)):*

Yes, that can be done. Though I thought args would make more sense, and allow even more flexibility with experimenting with possible variations.
</blockquote></details>

Actually, come to think of it, Chromatic isn't going to be able to do a good job screenshotting an animation anyway, so my point was kind of moot. @siltomato Can you just disable Chromatic for this entire file? https://www.chromatic.com/docs/ignoring-elements

josephmyers:
@Nateowami, do we want to leave this open, as well?

PR #2021, opened 2023-08-24
SF-2195 Close chapter audio player when audio is finished

RaymondLuong3:
* emit finished when the base audio player finishes
* move test audio files to a test audio folder

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2021)


josephmyers (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2021)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.ts` line 52 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2021#-Ncui9Ha6f_6k4fGgT4x:-Ncui9Ha6f_6k4fGgT4y:b-39jaup) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab585273022eae027720119cd276d3d3bc3817d4/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.ts#L52)):*
> ```XML
>     this.audioPlayer?.audio?.play();
>     if (this.audioPlayer?.audio == null) return;
>     const finishedSubscription: Subscription = this.subscribe(this.audioPlayer.audio.finishedPlaying$, () => {
> ```

This results in n-1 extra subscriptions if you play/pause. You'll need to check for a pre-existing subscription if you want to keep it here, but I believe you can just add a listener to AudioPlayer.status$ and do the subscribe when it becomes Available.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts` line 56 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2021#-Ncuk97F5DjJaL3ZSUa8:-Ncuk97F5DjJaL3ZSUa9:boae11) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab585273022eae027720119cd276d3d3bc3817d4/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts#L56)):*
> ```XML
>     expect(env.component.audioPlayer.currentRef).toEqual('verse_1_1');
>   });
> ```

We should add coverage for the above bug, and we should add coverage for playing/finishing the audio a second time.

josephmyers:
> What do you mean by bug above?

The multiple subscriptions. We want to ensure we're only executing that "on finished" code once, correct? Before your fix, play/pause/play would cause an extra subscription and make the "on finished" code execute twice. I would think we'd want to test that that only happens once.

josephmyers (approved):
I'm approving this, but the diff was really weird. It doesn't match your single commit (there are extra files changed in the review). Just be careful when merging in after test!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2021#-:-Ne1EQrx8vhxC5K9KCad:b-9dfa96)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.ts` line 52 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2021#-Ncui9Ha6f_6k4fGgT4x:-Ne1CylD8nVeWIpCXIte:b-pbdqzg) ([raw file](https://github.com/sillsdev/web-xforge/blob/ab585273022eae027720119cd276d3d3bc3817d4/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.ts#L52)):*

That's a good catch. I added a listener to the audioAvailable$ subject as part of the afterViewInit lifecycle hook. This seems to be doing what we want. I think it is better than using the status$ subject on the audioPlayer. Let me know what you think.
</blockquote></details>

Agreed!

Nateowami (dismissed):
I'm going to mark this as "Changes requested" while we discuss it in Slack so it doesn't mistakenly get merged.

Nateowami:
@josephmyers The final commit message that got generated for this was:
```
SF-2195 Close chapter audio player when audio is finished (#2021)

* SF-2195 Close chapter audio player when audio is finished
```
Which is a bit redundant and makes it slower to skim through a list of commits. Not a big deal, but I prefer if we clean up commits messages when we merge, if there's anything like that.

PR #2033, opened 2023-09-01
SF-2116 Convert checking component to material with questions by chapter

siltomato:
- Converted `CheckingComponent` to material and converted attributes and css classes from `@angular/flex-layout` to styles.
- Changed to have 3-way question scope options to load questions for all books, one book, or one chapter.
   - Added checking route that includes chapter.
   - Replaced chapter picker with `BookChapterChooserComponent`.
   - Previous/Next question buttons should take into account the current question filter and should navigate to questions outside the current chapter.
      - Created new query `queryAdjacentQuestion()` to do this and refactored question queries into their own service.
      - Reworked some query types in `query-parameters.ts`
   - Scope selection is added to the filter menu as 'Question range'.  This menu can be accessed by clicking the filter icon or by clicking the filter text description below the questions list title.

### Style Question...
Currently, there is padding beside the questions list.  Would this look better with no padding?

### _With Padding_
image

### _No Padding_
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2033)


Nateowami:
I've been testing it out locally, and this is great! I've been looking forward to this for a long time.

I think there is a major usability issue though in that when you get to the end of a chapter the next question button is disabled. You have to go to the top of the page and select the next chapter from the list, which isn't obvious.

I think making the "next question" button go to the next chapter when you're at the end of this chapter should be pretty simple to implement. (Actually, it shouldn't be the next chapter, but the chapter of the next question. If you're at the end of chapter 1, and the next question is in chapter 3, clicking next should take you to the first question of chapter 3).

siltomato (commented):
Yes, I see what you mean.  Ok, what about if there is a filter on the questions?  It seems like the prev/next question should follow that filter.  If so, perhaps a new db question query would need to be created to get the prev/next question with filter options that mirrored the filter functions?  This would prevent the whole question list from needing to be queried and then filtered, since this was avoided by just fetching the questions for the chapter.  What are your thoughts?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-:-Ne1VfaqC8aWnYsb0O4Y:bmjl38q)* status: 0 of 33 files reviewed, all discussions resolved

src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_ar.json
```diff
@@ -242,7 +242,7 @@
     "questions": "أسئلة"
   },
   "checking_questions": {
-    "no_questions_found": "No questions found to match your filter"
+    "no_questions_found_filtered": "No questions found to match your filter"
```
Nateowami, line 245: This isn't wrong or going to break anything, but I want to make sure you're aware you don't have to update the non-English localization files. The English files get uploaded to Crowdin, and the non-English ones are downloaded and overwrite the files in the repo, so the update can easily get applied without going through file by file.

And before the files get updated, it just falls back to the English strings.

siltomato (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2033)* status: 0 of 48 files reviewed, 2 unresolved discussions (waiting on @github-code-scanning[bot] and @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 973 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2033#gh-1328294053:-NeeCYTw1TcaoZNmVwp2:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/8a16d2ffea220f74e4ffdca2325911a00f9e16fb/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L973)):*

## Useless conditional

This use of variable 'adjacentQuestionInsideChapter' always evaluates to false.

[Show more details](https://github.com/sillsdev/web-xforge/security/code-scanning/639)
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_ar.json` line 245 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2033#gh-1322973417:-NeeCbeB1MkQJeTT6DuS:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/783f6d16af4d6d47d6baf95e4e8db42178a46b60/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_ar.json#L245)):*

This isn't wrong or going to break anything, but I want to make sure you're aware you don't have to update the non-English localization files. The English files get uploaded to Crowdin, and the non-English ones are downloaded and overwrite the files in the repo, so the update can easily get applied without going through file by file.

And before the files get updated, it just falls back to the English strings.
</blockquote></details>

Done.

RaymondLuong3 (changes requested):
Yes, if the list of questions is filtered, the best behaviour would be that the next and previous buttons go to the next and previous in the list of filtered questions. If you are at the end of the filtered list for a chapter, then the next button should go to the next question in the next chapter for the applied filter. Feel free to save that as a follow up commit.

I noticed that if I have all questions selected in the side nav, and I navigate to a different question that is in a different book, the navigation takes me away from "All questions" and takes me to the book for the question that I navigate to. It should stay on the all questions selection.

> Reviewed 1 of 33 files at r1, 5 of 44 files at r2, 1 of 15 files at r3, 5 of 8 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-:-Ney7EbR9W3Bx_iK9Wuj:bvo69cd)* status: 12 of 48 files reviewed, 7 unresolved discussions (waiting on @github-code-scanning[bot], @Nateowami, and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html` line 185 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Nexwbmx7v1_EhpmJQPv:-Nexwbmx7v1_EhpmJQPw:b-mge7c9) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L185)):*
> ```HTML
>             {{ t("overview") }}
>           </a>
>           
> ```

Do we have an open issue to track this? Seems it would be a good thing to do.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 7 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ney6nVNBuTEil050eJj:-Ney6nVNBuTEil050eJk:bynmlpj) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L7)):*
> ```XML
> import { Canon, VerseRef } from '@sillsdev/scripture';
> import { SplitComponent } from 'angular-split';
> import { debounce } from 'lodash-es';
> ```

rxjs also has a version of debounce. I imagine they work similarly, but we should stick with the rxjs one for consistency in our app.



_Code quote:_
```XML
import { debounce } from 'lodash-es';
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 446 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ney1Pck-8vs5NKYhVym:-Ney1Pck-8vs5NKYhVyn:baq4m2i) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L446)):*
> ```XML
>         const prevProjectId = this.projectDoc == null ? '' : this.projectDoc.id;
> 
>         // Reroute to first book in chapter if not specified and not showing all books
> ```

This comment is kind of confusing. Did you mean first chapter in book?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 463 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ney1gvP4dqnvZvlAcdv:-Ney1gvP4dqnvZvlAcdw:bm0ikq5) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L463)):*
> ```XML
> 
>         const bookNum = bookId == null ? 0 : Canon.bookIdToNumber(bookId);
>         const chapterNum = Number(chapter);
> ```

We can use parseInt(chapter) here. Also, add types here.



_Code quote:_
```XML
   const chapterNum = Number(chapter);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 492 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ney2yte7wsN-v8ArMnS:-Ney2yte7wsN-v8ArMnT:b-mcq965) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L492)):*
> ```XML
> 
>           // Select newly created question if any
>           if (this.questionToActivate) {
> ```

Use `this.questionToActivate != null`



_Code quote:_
```XML
         if (this.questionToActivate) {
```

RaymondLuong3 (changes requested):
> Reviewed 2 of 33 files at r1, 18 of 44 files at r2, 13 of 15 files at r3, 3 of 8 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2033)* status: all files reviewed, 11 unresolved discussions (waiting on @github-code-scanning[bot], @Nateowami, and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 986 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-NeyYExP9NQaoEg1jGug:-NeyYExP9NQaoEg1jGuh:b-1j6c9c) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L986)):*
> ```XML
>           this.activeQuestionFilter,
>           prevOrNext
>         )
> ```

I could be mistaken, but you should dispose of the query when you are finished with it so that it can be cleaned up.



_Code quote:_
```XML
        await this.checkingQuestionsService.queryAdjacentQuestion(
          this.projectId!,
          relativeTo,
          this.activeQuestionFilter,
          prevOrNext
        )
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/checking-questions.service.ts` line 1 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-NeyTvy68T1qqrCKMDSm:-NeyTvy68T1qqrCKMDSn:b-14cmiz) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/checking-questions.service.ts#L1)):*
> ```XML
> import { Injectable } from '@angular/core';
> ```

This service should be moved into the checking folder and not in the xforge-common folder. 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/checking-questions.service.ts` line 7 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-NeyOdoO3EKF56LiA-Fa:-NeyOdoO3EKF56LiA-Fb:b9n3oxg) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/checking-questions.service.ts#L7)):*
> ```XML
> import { getQuestionDocId, Question } from 'realtime-server/lib/esm/scriptureforge/models/question';
> import { VerseRefData } from 'realtime-server/lib/esm/scriptureforge/models/verse-ref-data';
> import { QuestionDoc } from 'src/app/core/models/question-doc';
> ```

Use a relative path here



_Code quote:_
```XML
t { QuestionDoc } from 'src/app/core/models/question-doc';
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/checking-questions.service.ts` line 78 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-NeyXQ_yB3zUklSRS4B_:-NeyXQ_yB3zUklSRS4Ba:bnnuy3m) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/checking-questions.service.ts#L78)):*
> ```XML
>     questionFilter: QuestionFilter,
>     prevOrNext: 'prev' | 'next'
>   ): Promise<RealtimeQuery<QuestionDoc>> {
> ```

I'm not sure I really like how complicated it is to just get the next question to navigate to, but I don't really know of a better solution. 



_Code quote:_
```XML
  queryAdjacentQuestion(
    projectId: string,
    relativeTo: Question | VerseRefData,
    questionFilter: QuestionFilter,
    prevOrNext: 'prev' | 'next'
  ): Promise<RealtimeQuery<QuestionDoc>> {
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2033)* status: all files reviewed, 12 unresolved discussions (waiting on @github-code-scanning[bot] and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 1 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-NfU673mDebxDp3ZEHBJ:-NfU673mDebxDp3ZEHBK:b-nl2mok) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L1)):*
> ```SCSS
> @use '@material/icon-button/mixins' as icon-button;
> ```

This `@use` can now be removed. `@material/icon-button` is only installed because it's a dependency of `@angular-mdc/web`, so removing it is critical to removing MDC.

siltomato (commented):
With the new update, there is now 3-way 'scope' (labeled as 'Questions range' in the filter): all | book | chapter.  The route changes to the active book/chapter, with the scope as a query string.  If desired, the user could enter book and chapter in the url and navigate there even if there are no questions, or even if the last active question was in a different book/chapter.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-:-NfyhOh1EE8-DHKXD5aQ:b963hxf)* status: 25 of 52 files reviewed, 10 unresolved discussions (waiting on @github-code-scanning[bot], @Nateowami, and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 1 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-NfU673mDebxDp3ZEHBJ:-NfyLIcdCN1W2oHVj3Nr:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L1)):*

This `@use` can now be removed. `@material/icon-button` is only installed because it's a dependency of `@angular-mdc/web`, so removing it is critical to removing MDC.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 7 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ney6nVNBuTEil050eJj:-NeyQnmY7xjym1_Y0eNT:b-72hd0v) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L7)):*

rxjs also has a version of debounce. I imagine they work similarly, but we should stick with the rxjs one for consistency in our app.
</blockquote></details>

Yes, good point.  The rxjs operator for this use seemed like a lot more code just to set it up.  Do you know a simple way to handle this?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 446 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ney1Pck-8vs5NKYhVym:-NfyMBLo-W0L6-ujbDlc:btetjd8) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L446)):*

This comment is kind of confusing. Did you mean first chapter in book?
</blockquote></details>

Yes, confusing comment.  This section has been reworked as part of the 3-way scope functionality.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 463 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ney1gvP4dqnvZvlAcdv:-NfyMAFR25tMA0uKYcFk:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L463)):*

We can use parseInt(chapter) here. Also, add types here.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 492 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ney2yte7wsN-v8ArMnS:-NfyNeoN0EkRfIk9_1LY:blfzpnp) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L492)):*

Use `this.questionToActivate != null`
</blockquote></details>

Code reworked... this is removed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 986 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-NeyYExP9NQaoEg1jGug:-NfyNqS3CIVyJgrck32S:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L986)):*

I could be mistaken, but you should dispose of the query when you are finished with it so that it can be cleaned up.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html` line 185 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Nexwbmx7v1_EhpmJQPv:-NfyKZho89RAJG9hlulB:b-5z2u0r) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L185)):*

Do we have an open issue to track this? Seems it would be a good thing to do.
</blockquote></details>

If I understand correctly, I think this will be no longer applicable in the new side nav.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/checking-questions.service.ts` line 1 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-NeyTvy68T1qqrCKMDSm:-NfyRCnJ1DTiF2Zf09RU:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/checking-questions.service.ts#L1)):*

This service should be moved into the checking folder and not in the xforge-common folder.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/checking-questions.service.ts` line 7 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-NeyOdoO3EKF56LiA-Fa:-NfyRIzEE-r5esQ24-34:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/checking-questions.service.ts#L7)):*

Use a relative path here
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/checking-questions.service.ts` line 78 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-NeyXQ_yB3zUklSRS4B_:-NfyRKMI170t1BD7z2ZI:buz1f0a) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/checking-questions.service.ts#L78)):*

I'm not sure I really like how complicated it is to just get the next question to navigate to, but I don't really know of a better solution.
</blockquote></details>

True that.

RaymondLuong3 (changes requested):
Phew! That's a lot to cover, and I'm sure I missed some things. Overall, this is a good change to allow users the ability to narrow down questions to the chapter level. I could be wrong, but I'm not sure how much we add by having the chapter number in the path. Maybe I missed why that was important.

> Reviewed 27 of 27 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-:-Ng0N5j2DrLNwue41j11:b-y1r9pj)* status: all files reviewed, 13 unresolved discussions (waiting on @github-code-scanning[bot], @Nateowami, and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.ts` line 1 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng-adDo-wo_fBlc9vYg:-Ng-adDo-wo_fBlc9vYh:bqa0jjh) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.ts#L1)):*
> ```XML
> import { VerseRefData } from 'realtime-server/lib/esm/scriptureforge/models/verse-ref-data';
> ```

I wonder if we should put this into the checking.utils.ts file. Or if you want to keep this separate, a more specific name could be checking-routing.utils.ts.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.ts` line 6 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng-b05vDosU58gxJ-ho:-Ng-b05vDosU58gxJ-hp:byle4mh) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.ts#L6)):*
> ```XML
> export type QuestionScope = (typeof scopes)[number];
> 
> export function isQuestionScope(scope: any): scope is QuestionScope {
> ```

I have trouble understanding the return type here. Is the return type a boolean?



_Code quote:_
```XML
export function isQuestionScope(scope: any): scope is QuestionScope {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 15 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng-ffi53bmAHpG5Otwv:-Ng-ffi53bmAHpG5Otww:b-x62va0) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L15)):*
> ```HTML
>             <mat-selection-list [multiple]="true" class="menu-list">
>               <h2 matSubheader>{{ t("questions_range") }}</h2>
>               <button
> ```

I like how intuitive this is to use. But I'll mention that I was a little lost at first when the chapter select was disabled, and I wanted to see other chapters.
It might be worth adding a tooltip onto the disabled book and chapter nav. But since that does not work on mobile, we might want a better solution than that. Just pointing this out.



_Code quote:_
```HTML
   <button
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 7 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ney6nVNBuTEil050eJj:-Ng0MnLJDXnMDPDARWsw:bf2nri4) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L7)):*

Yes, good point.  The rxjs operator for this use seemed like a lot more code just to set it up.  Do you know a simple way to handle this?
</blockquote></details>

Hmm, I didn't mean for it to add more work, so let's ignore that and keep it as is.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 26 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng-uGEP2fx6zTdTXp6e:-Ng-uGEP2fx6zTdTXp6f:bljs7km) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L26)):*
> ```XML
>   PreCreationQuestionData,
>   QuestionFilter
> } from 'src/app/checking/checking/checking-questions.service';
> ```

Use relative paths here.



_Code quote:_
```XML
om 'src/app/checking/checking/checking-questions.service';
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 475 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng-vSrB80cZXP-uJf9c:-Ng-vSrB80cZXP-uJf9d:b-c4pejz) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L475)):*
> ```XML
>           const routeBookId: string | undefined = params['bookId']?.toLowerCase();
>           const routeScope: QuestionScope | undefined =
>             routeBookId === 'all' ? 'all' : queryParams['scope']?.toLowerCase();
> ```

What is the reason for reading this from the url? Are we expecting to let users share links to get to specific filters on the checking component? I'm trying to understand how this might be useful? Is this saving a user's current filter so that they can navigate back to it after navigating away?



_Code quote:_
```XML
          const routeScope: QuestionScope | undefined =
            routeBookId === 'all' ? 'all' : queryParams['scope']?.toLowerCase();
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 608 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng0OpNW2Cs4CL5qtpIZ:-Ng0OpNW2Cs4CL5qtpI_:b-mdiofa) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L608)):*
> ```XML
>             this.questionsQuery = await this.checkingQuestionsService.queryQuestions(routeProjectId, {
>               bookNum: routeScope === 'all' ? undefined : routeBookNum,
>               chapterNum: routeScope === 'chapter' ? routeChapterNum : undefined,
> ```

I see when I navigate to a chapter without questions by typing in the URL, and the scope is 'chapter', I get redirected to the chapter with the questions. I don't expect anyone to be bothered by this, but just pointing it out.



_Code quote:_
```XML
              chapterNum: routeScope === 'chapter' ? routeChapterNum : undefined,
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 637 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng00Dm-1UHQBiOw8qYE:-Ng00Dm-1UHQBiOw8qYF:bk73dj5) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L637)):*
> ```XML
>               merge(
>                 this.questionsQuery.ready$.pipe(filter(isReady => isReady)),
>                 this.questionsQuery!.remoteChanges$.pipe(map(() => 'remote')),
> ```

It looks like these bangs are unnecessary and can be removed
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 679 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng00b-8DWvwZ5XlOCQT:-Ng00b-8DWvwZ5XlOCQU:b-byxl9c) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L679)):*
> ```XML
> 
>     // Pre-creation question object is no longer needed after question is actually created
>     this.subscribe(this.checkingQuestionsService.afterQuestionCreated$, (data: QuestionDoc) => {
> ```

When a question is created using the question dialog, my guess is that there is no way to detect when the question is actually created, and that's why we need these subscriptions. Is that correct?



_Code quote:_
```XML
    this.subscribe(this.checkingQuestionsService.afterQuestionCreated$, (data: QuestionDoc) => {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts` line 42 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng05-fWBm_PBUaKn5v3:-Ng05-fWBm_PBUaKn5v4:berq8xm) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts#L42)):*
> ```XML
> import { AvatarTestingModule } from 'xforge-common/avatar/avatar-testing.module';
> import { BugsnagService } from 'xforge-common/bugsnag.service';
> import { QuestionFilter } from 'src/app/checking/checking/checking-questions.service';
> ```

Use a relative path here.



_Code quote:_
```XML
src/app/checking/checking/checking-questions.service';
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts` line 18 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng0LZnX4XKFinKTqaRq:-Ng0LZnX4XKFinKTqaRr:b-421w53) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts#L18)):*
> ```XML
>   @Input() chapters: number[] = [];
>   @Input() chapter?: number;
>   @Input() prevNextHidden?: boolean = false;
> ```

Just curious, why hide the controls completely and not just disable them?



_Code quote:_
```XML
  @Input() prevNextHidden?: boolean = false;
```

siltomato (commented):
Thanks for going through this, Raymond.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-:-Ng0saD00ntoXgL2MJAx:bz6002w)* status: all files reviewed, 10 unresolved discussions (waiting on @github-code-scanning[bot], @Nateowami, and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.ts` line 1 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng-adDo-wo_fBlc9vYg:-Ng0p9fl7gS9xIoL_tnu:b-7ypg3m) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.ts#L1)):*

I wonder if we should put this into the checking.utils.ts file. Or if you want to keep this separate, a more specific name could be checking-routing.utils.ts.
</blockquote></details>

Added to `checking.utils.ts`.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.ts` line 6 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng-b05vDosU58gxJ-ho:-Ng0eaUh1RWnpUDTPqrg:b-bdctdq) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.ts#L6)):*

I have trouble understanding the return type here. Is the return type a boolean?
</blockquote></details>

Yeah, interesting question.  It's a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates), but I suppose that once it's javascript, it's just a `boolean`.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 15 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng-ffi53bmAHpG5Otwv:-Ng0phQN68QilZC4GpGb:by1xa2) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L15)):*

I like how intuitive this is to use. But I'll mention that I was a little lost at first when the chapter select was disabled, and I wanted to see other chapters.
It might be worth adding a tooltip onto the disabled book and chapter nav. But since that does not work on mobile, we might want a better solution than that. Just pointing this out.
</blockquote></details>

Yes, that is a good point.  Maybe the book/chapter selectors should be enabled always, and then jump the question list to the first question for the selected chapter, or the first question of the first chapter in the selected book?

Maybe this could be added to a followup if it isn't too big of a usability issue.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 26 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng-uGEP2fx6zTdTXp6e:-Ng0WBGd09TQU5otKo9w:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L26)):*

Use relative paths here.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 475 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng-vSrB80cZXP-uJf9c:-Ng0RgbE0a_QHbm0splf:bn7qw7r) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L475)):*

What is the reason for reading this from the url? Are we expecting to let users share links to get to specific filters on the checking component? I'm trying to understand how this might be useful? Is this saving a user's current filter so that they can navigate back to it after navigating away?
</blockquote></details>

Yeah, good question.  The way I'm thinking of it that query string params seem like the natural place to store state for result filters like scope and, well, filter.  In fact, filter should probably be a query string param, itself.  That way, the filter state would be persisted across page refreshes (like scope is now).

Also, like you mentioned, the history navigation seems like a positive.

Maybe adding 'filter' to the query string would be a good follow-up issue.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 608 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng0OpNW2Cs4CL5qtpIZ:-Ng0U7MtEDuqgv2TjqQ_:b-szlh85) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L608)):*

I see when I navigate to a chapter without questions by typing in the URL, and the scope is 'chapter', I get redirected to the chapter with the questions. I don't expect anyone to be bothered by this, but just pointing it out.
</blockquote></details>

Navigating to a specific book/chapter url should load that chapter, even if there are no questions in that chapter.

If I'm understanding you right, you may be experiencing a bug, but I can't reproduce it.  Can you reproduce this consistently?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 637 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng00Dm-1UHQBiOw8qYE:-Ng0XPPF0Cv5yxRelIgD:b-kg3qx9) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L637)):*

It looks like these bangs are unnecessary and can be removed
</blockquote></details>

Oops.  Good catch, thanks.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 679 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng00b-8DWvwZ5XlOCQT:-Ng0XdVl1qMOYeWfzi73:b2foyeh) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L679)):*

When a question is created using the question dialog, my guess is that there is no way to detect when the question is actually created, and that's why we need these subscriptions. Is that correct?
</blockquote></details>

Yes, I didn't see a way to determine the verse ref for the question to be created before the actual creation triggered `localChanges$` on the questions query.  So, I added pre- and post-creation hooks to notify right before and after a question is created.

Maybe this isn't the best way to handle this, but I wanted to prevent triggering an unnecessary 'visible questions update' (which happens on `localChanges$` emission) when a newly created out-of-range question was about to cause a navigation to a new range, which would trigger a new questions query.

Ideally, `localChanges$` would emit what the actual changes were, similar to `ngOnChanges`, so that listeners could respond appropriately.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts` line 42 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng05-fWBm_PBUaKn5v3:-Ng0bdEb8zJAjhHvw1uw:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts#L42)):*

Use a relative path here.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts` line 18 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng0LZnX4XKFinKTqaRq:-Ng0cDPT7nqhe1rJNciT:b-x2bhcu) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts#L18)):*

Just curious, why hide the controls completely and not just disable them?
</blockquote></details>

Hmm, I don't have a good reason.  Maybe that would be better.  Do you have a strong feeling about it?

Nateowami (commented):
> I'm not sure how much we add by having the chapter number in the path. Maybe I missed why that was important.

I actually really appreciate the presence of the chapter in the path. Some time ago I was talking to a user that wanted to invite an observer to review a chapter, and wanted to send a link to that specific chapter. But at the time, the editor didn't have a chapter in the URL, so it wasn't possible. I could see a similar use-case with community checking.

I also appreciate the presence of the scope variable in the URL, and I think it would be good to have the filter in the URL as well, but that's not worth holding this PR up over.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-:-Ng2Rod9AfaR_htHVT3o:b-w56bq1)* status: 47 of 54 files reviewed, 9 unresolved discussions (waiting on @github-code-scanning[bot], @RaymondLuong3, and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 15 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng-ffi53bmAHpG5Otwv:-Ng2KBcKDZ7K8NM6IRX6:b-bxobpu) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L15)):*

Yes, that is a good point.  Maybe the book/chapter selectors should be enabled always, and then jump the question list to the first question for the selected chapter, or the first question of the first chapter in the selected book?

Maybe this could be added to a followup if it isn't too big of a usability issue.
</blockquote></details>

I, too, was a bit unsure about the disabling of the book/chapter chooser. But then I remember that most users are just going to use it chapter-by-chapter, and will only ever get in that state by specifically choosing a filter, so it's less of an issue than I originally thought.

I think it may be better to keep it enabled, but I think we can merge it as-is and think through it more afterward.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 69 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng2PheE3mGlJkL7ddZ2:-Ng2PheE3mGlJkL7ddZ3:boox7se) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L69)):*
> ```SCSS
> 
>   .active-question-filter {
>     font-size: 0.7em;
> ```

I realize you didn't set the color or font size here, but I think the font size is too small and the text is too light. Not everyone has excellent eyesight, and I'm guessing in certain scripts it will be even harder to read. Also, the use of transparency for setting a color should be avoided, especially for text (in some scripts it causes overlapping glyphs that look bad). This does not need to be addressed before merging this PR, but I wanted to mention it because it was one of the main things I noticed when doing the functional review.


___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts` line 18 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng0LZnX4XKFinKTqaRq:-Ng2Kw5h1Z-JSgeoBl5a:b-cr8bi5) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts#L18)):*

Hmm, I don't have a good reason.  Maybe that would be better.  Do you have a strong feeling about it?
</blockquote></details>

Do we really need the next/previous controls in this context? It was added in this component mainly for use on the editor page, and is they're always hidden on mobile. I think turning them off completely for the checking component may make more sense.

I don't want this to slow down the PR though, so leaving as-is is OK.

siltomato (commented):
Yeah, I agree on the filter in the query params as well.  I'll make a note of it, but leave it for a separate issue.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-:-Ng4e6X00yW-N9YmAgN6:b-2e4ycv)* status: 45 of 54 files reviewed, 6 unresolved discussions (waiting on @github-code-scanning[bot] and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 15 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng-ffi53bmAHpG5Otwv:-Ng4dzbo7OC_x8w2C7zL:b-pmy01s) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L15)):*

I, too, was a bit unsure about the disabling of the book/chapter chooser. But then I remember that most users are just going to use it chapter-by-chapter, and will only ever get in that state by specifically choosing a filter, so it's less of an issue than I originally thought.

I think it may be better to keep it enabled, but I think we can merge it as-is and think through it more afterward.
</blockquote></details>

Ok, I'll leave as-is for now.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 69 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng2PheE3mGlJkL7ddZ2:-Ng4dL-t0DSwq6E45Ypu:b-4c3hda) ([raw file](https://github.com/sillsdev/web-xforge/blob/df44a312d3902bbc957a1f782fcebcc463aea2c2/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L69)):*

I realize you didn't set the color or font size here, but I think the font size is too small and the text is too light. Not everyone has excellent eyesight, and I'm guessing in certain scripts it will be even harder to read. Also, the use of transparency for setting a color should be avoided, especially for text (in some scripts it causes overlapping glyphs that look bad). This does not need to be addressed before merging this PR, but I wanted to mention it because it was one of the main things I noticed when doing the functional review.


</blockquote></details>

Yep, good point.  I made some adjustments to size and color, and I swapped out the 'Filter' text for the icon to squeeze some more space for when the filter string is long.

RaymondLuong3 (approved):
Hmm, I was thinking that the "share url" scenario would be rare. When I am in an app, I only find it useful to share my state of the app when I'm using Google maps. Otherwise, I never have a reason to share my state of an app, maybe Jira? Anyway, I can understand why others might find it useful.
I approve, let's get it to testers.

> Reviewed 5 of 7 files at r6, 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-:-Ng537LAANXP7WPvKBgj:b-xsrsyn)* status: all files reviewed, 1 unresolved discussion (waiting on @github-code-scanning[bot])

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 608 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng0OpNW2Cs4CL5qtpIZ:-Ng53c9M9kJPSos1ASQG:b-od0b51) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L608)):*

Navigating to a specific book/chapter url should load that chapter, even if there are no questions in that chapter.

If I'm understanding you right, you may be experiencing a bug, but I can't reproduce it.  Can you reproduce this consistently?
</blockquote></details>

Sorry, I was mistaken, it is when the scope is at book level that I try to enter a chapter number that has no questions, and I get navigated to the chapter of the first question. I actually think that should be the expected behaviour, so I'll resolve this.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts` line 18 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-Ng0LZnX4XKFinKTqaRq:-Ng54B1r3DJQ1EazBoJ-:b-r0l5ph) ([raw file](https://github.com/sillsdev/web-xforge/blob/2e61549cdc80b5fed0f8a3b1bab1e11a2f2e9029/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts#L18)):*

Do we really need the next/previous controls in this context? It was added in this component mainly for use on the editor page, and is they're always hidden on mobile. I think turning them off completely for the checking component may make more sense.

I don't want this to slow down the PR though, so leaving as-is is OK.
</blockquote></details>

I'm satisfied with this too, and we can always change it later.

RaymondLuong3:
It looks like you have a story-book test failure. You should be able to see it locally by running npm run storybook and in another terminal, run npm run test-storybook.

siltomato (commented):
Moving to testing.  One question... did either of you get a chance to look at the initial PR comment with comparison images?  If a change is desired, it could be for a separate issue, but I'd be curious to get your feedback on that.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2033#-:-Ng5EON_Adjnjs9PHs7E:bfvei44)* status: all files reviewed, 1 unresolved discussion (waiting on @github-code-scanning[bot])

PR #2037, opened 2023-09-04
SF-2197 Chapter Audio Error Handling for Bad Timing Files

nigel-wells:
* Introduced a grace period for skipping to previous verses
* Skipping to the next verse can now skip to the start of the current verse if it hasn't started yet
* Fixed bug when skipping to the start of the first timing data

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2037)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts
@@ -142,6 +142,14 @@ xdescribe('ScriptureAudioComponent', () => {
     expect(pauseSpy).toHaveBeenCalled();
     expect(count).toEqual(1);
   });
+
+  it('can skip to previous verse', async () => {});
+
+  it('skipping to previous verse remains on the current verse if within grace period', async () => {});
+
+  it('skipping to the next verse will skip to the start of the current timing data if it has not started yet', async () => {});
+
+  it('skipping beyond the last timing data will stop the player ', async () => {});
 });
 
 @Component({ selector: 'app-host', template: '' })
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts
@@ -117,10 +117,10 @@ xdescribe('ScriptureAudioComponent', () => {
     env.component.audioPlayer.currentVerseChanged.subscribe(verseChangedSpy);
     env.playButton.nativeElement.click();
     await env.waitForPlayer(1400);
-    expect(env.component.audioPlayer.audioPlayer!.audio!.currentTime).toBeGreaterThan(timings[1].from);
+    expect(env.currentTime).toBeGreaterThan(timings[1].from);
     expect(env.verseLabel.nativeElement.textContent).toEqual('Genesis 1:1');
     await env.waitForPlayer(1400);
-    expect(env.component.audioPlayer.audioPlayer!.audio!.currentTime).toBeGreaterThan(timings[3].from);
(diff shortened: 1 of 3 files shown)


josephmyers:
See video for active verse not updating.
https://github.com/sillsdev/web-xforge/assets/43733878/e81ad7c4-b93d-45e8-9716-5ccddc455572

josephmyers:
And here's some real weirdness:
https://github.com/sillsdev/web-xforge/assets/43733878/c706eae6-2a9e-4c19-b07c-e9eecd5e99b4

nigel-wells:
Can you please post in here what your timing data is as I can't really comment without seeing it. Or, can you send me through the timing data file you're using.

josephmyers:
The next weirdness is still with the Mark 3 test data, but we can explain this one. It may be related to the first video. The "s" entries (subchapter labels) are being counted in the next/prev behavior. This is not necessarily a problem, and there are two ways we can address them. We can either allow users to skip between these labels, treating them as navigable, or not. If you think we should go the other way, as I would assume based on the current code, and allow/force the user to navigate through them, we shouldn't show the active verse label as the previous verse, as it is now. We could do something like "Mark 3," "<Title>," or even just "-"
https://github.com/sillsdev/web-xforge/assets/43733878/403948fb-62cc-4281-b88d-9e06e3b0fd16

josephmyers (changes requested):
Good job on these. Really just seeing the minor issues with active verse display, maybe not even your code.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2037#-:-NdTTO6h8voKSLijqCXj:b-ltgw3r)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts` line 120 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2037#-NdTmutF7TpJOfiovIQL:-NdTmutF7TpJOfiovIQM:bgqpoas) ([raw file](https://github.com/sillsdev/web-xforge/blob/2d5726a7d3bdf5773cfca6c14af3d186272481ba/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts#L120)):*
> ```XML
>   });
> 
>   it('can skip to previous verse', async () => {
> ```

We already have this covered in another test (second in this file). Maybe there's another test we could add here instead? Or you can delete.

nigel-wells (commented):
I've asked for feedback from  @RaymondLuong3 as I know he worked on some code specifically relating to section headings and how they should work.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2037#-:-NdXrgm-42R87gXjt-7o:bmz3lf)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts` line 120 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2037#-NdTmutF7TpJOfiovIQL:-NdXrdkp6O-rtuuaWRgD:birq90c) ([raw file](https://github.com/sillsdev/web-xforge/blob/2d5726a7d3bdf5773cfca6c14af3d186272481ba/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.spec.ts#L120)):*

We already have this covered in another test (second in this file). Maybe there's another test we could add here instead? Or you can delete.
</blockquote></details>

So it is - I have removed this one.

RaymondLuong3:
I think it would make sense to leave the verse number blank (Mark 3: ) when playing a section heading. Previously, I had been treating them as the same as the verse before (Paratext does this) but we don't need to do that.

nigel-wells:
Thanks Raymond. Let's create another Jira card to address that scenario. Did you require anything else on this Joseph?

josephmyers (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2037)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

PR #2048, opened 2023-09-07
SF-1627 Fix audio player in RTL UI

RaymondLuong3:
The audio player uses a mat-slide for UI to indicate the progress of the audio clip. In RTL, we forced the mat-slider to be in LTR mode so that it plays left to right. However, there seems to be some bug and the slider players left to right, but the green fill goes right to left.
This sets the slider to the direction indicated by the locale, but in RTL languages, it rotates the slider so that it plays equivalent to the LTR mode.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2048)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html
@@ -1,8 +1,16 @@
 <ng-container *transloco="let t; read: 'checking_audio_player'">
   <div *ngIf="isAudioInitComplete">
-    <div *ngIf="audio !== undefined && (isAudioAvailable$ | async)" class="content" dir="ltr">
+    <div *ngIf="audio !== undefined && (isAudioAvailable$ | async)" class="content">
       <div class="current-time">{{ audio.currentTime | audioTime }}</div>
-      <mat-slider class="slider" [min]="0" [max]="100" [step]="1" [value]="seek" (input)="onSeek($event)"></mat-slider>
+      <mat-slider
+        class="slider"
+        [min]="0"
+        [max]="100"
+        [step]="1"
+        [value]="seek"
+        (input)="onSeek($event)"
+        [dir]="direction"
+      ></mat-slider>
       <div class="duration">{{ duration | audioTime }}</div>
     </div>
     <div *ngIf="(isAudioAvailable$ | async) !== true" class="audio-not-available">
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html
@@ -1,6 +1,6 @@
 <ng-container *transloco="let t; read: 'checking_audio_player'">
   <div *ngIf="isAudioInitComplete">
-    <div *ngIf="audio !== undefined && (isAudioAvailable$ | async)" class="content">
+    <div *ngIf="audio !== undefined && (isAudioAvailable$ | async)" class="content" dir="ltr">
       <div class="current-time">{{ audio.currentTime | audioTime }}</div>
       <mat-slider
         class="slider"
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html
@@ -9,7 +9,7 @@
         [step]="1"
         [value]="seek"
         (input)="onSeek($event)"
-        [dir]="direction"
+        [dir]="i18n.direction"
       ></mat-slider>
       <div class="duration">{{ duration | audioTime }}</div>
     </div>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.scss
@@ -16,6 +16,9 @@
     ::ng-deep .mat-slider-thumb {
       background-color: #000;
     }
+    &[dir='rtl'] {
+      transform: rotateY(180deg);
+    }
   }
 }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.ts
@@ -22,15 +22,13 @@ export class AudioPlayerComponent extends AudioPlayerBaseComponent {
     return this.audio?.seek ?? 0;
   }
 
-  get direction(): 'ltr' | 'rtl' {
-    return this.i18n.direction;
-  }
-
   onSeek(event: MatSliderChange): void {
     let seek: number | null = event.value;
     if (seek == null) return;
-    if (this.direction === 'rtl') {
-      // The slider is reversed for RTL languages so the seek is the inverse of the value that is emitted
+    if (this.i18n.direction === 'rtl') {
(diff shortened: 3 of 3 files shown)


Nateowami:
If you try to seek, the input is reversed (drag it to the right, and it moves to the left). I think when we update to Material 15 very soon this is going to be fixed.

RaymondLuong3 (commented):
Thanks for finding that. I fixed the issue. If you would rather that we wait to update to material 15, I am ok with that too.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2048#-:-NdqvoQt-kN9CE0MDcfs:be8n5sf)* status: 0 of 3 files reviewed, all discussions resolved

pmachapman (approved):
:lgtm: A cool fix for an interesting bug.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2048#-:-Ne0VdIn3dgqJ8gc912N:b615vqi)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

pmachapman (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2048)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2048)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html` line 12 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2048#-Ne4svKr19aPqhjASGEU:-Ne4svKr19aPqhjASGEV:b-l0he7r) ([raw file](https://github.com/sillsdev/web-xforge/blob/c73704088f9467ed9626369ad2857b8a75e60547/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html#L12)):*
> ```HTML
>         [value]="seek"
>         (input)="onSeek($event)"
>         [dir]="direction"
> ```

Is there a reason this isn't just `[dir]="i18n.direction" like w do most other places? Then no getter is needed. The `i18n` member is already public.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.ts` line 33 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2048#-Ne4x5t_5wqLgMhY1ZQe:-Ne4x5t_5wqLgMhY1ZQf:b6h0e9h) ([raw file](https://github.com/sillsdev/web-xforge/blob/c73704088f9467ed9626369ad2857b8a75e60547/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.ts#L33)):*
> ```XML
>     if (seek == null) return;
>     if (this.direction === 'rtl') {
>       // The slider is reversed for RTL languages so the seek is the inverse of the value that is emitted
> ```

I think both here and in the CSS, a comment is needed to explain *why* (i.e. we're working around what we believe to be a bug in @angular/material@14.x).

We expect this to be removed at some point in the future, and it needs to be obvious why this was here in the first place. (In general, I think when working around some non-obvious problem, an explanation of the rationale is warranted).

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2048)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html` line 12 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2048#-Ne4svKr19aPqhjASGEU:-Ne5MbGw8q1dUg-hSqaZ:b-kes67o) ([raw file](https://github.com/sillsdev/web-xforge/blob/c73704088f9467ed9626369ad2857b8a75e60547/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html#L12)):*

Is there a reason this isn't just `[dir]="i18n.direction" like w do most other places? Then no getter is needed. The `i18n` member is already public.
</blockquote></details>

I was copying what I saw done in other files, but I agree, we can access the direction directly without the getter.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.ts` line 33 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2048#-Ne4x5t_5wqLgMhY1ZQe:-Ne5M_uL9xMqULeYpl7Q:b-mbcpfk) ([raw file](https://github.com/sillsdev/web-xforge/blob/c73704088f9467ed9626369ad2857b8a75e60547/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.ts#L33)):*

I think both here and in the CSS, a comment is needed to explain *why* (i.e. we're working around what we believe to be a bug in @angular/material@14.x).

We expect this to be removed at some point in the future, and it needs to be obvious why this was here in the first place. (In general, I think when working around some non-obvious problem, an explanation of the rationale is warranted).
</blockquote></details>

Good thinking.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2048)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @pmachapman and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html` line 12 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2048#-Ne4svKr19aPqhjASGEU:-Ne5P3kPEfO3ZtHLtNgz:by69jwv) ([raw file](https://github.com/sillsdev/web-xforge/blob/c73704088f9467ed9626369ad2857b8a75e60547/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html#L12)):*

I was copying what I saw done in other files, but I agree, we can access the direction directly without the getter.
</blockquote></details>

Most places already use `i18n.direction`, with only one exception (the editor component)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.ts` line 33 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2048#-Ne4x5t_5wqLgMhY1ZQe:-Ne5PYb039y_I2C2i-VH:b1l288g) ([raw file](https://github.com/sillsdev/web-xforge/blob/c73704088f9467ed9626369ad2857b8a75e60547/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.ts#L33)):*
> prevents the slider from working in RTL environments.

I think this could be clearer. They built in RTL support, and it works fine, but if an ancestor is RTL, it won't let you keep it as LTR. So it's a lack of support for bidirectionality.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2048)* status: 1 of 3 files reviewed, 2 unresolved discussions (waiting on @pmachapman and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html` line 12 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2048#-Ne4svKr19aPqhjASGEU:-Ne5QCIy9dMB9T_ZA1p4:byugr3s) ([raw file](https://github.com/sillsdev/web-xforge/blob/c73704088f9467ed9626369ad2857b8a75e60547/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html#L12)):*

Most places already use `i18n.direction`, with only one exception (the editor component)
</blockquote></details>

This is more of a side note, but when this is done in the editor component, it leaves unnecessary room for confusion:
``` html
 <div class="fab-bottom-sheet" [dir]="direction">
```
In the context of the editor, it's not clear what "direction" would refer to. It could be the direction of the text being edited, or the UI direction. If it instead read `[dir]="i18n.direction"` there would be no ambiguity.

PR #2059, opened 2023-09-12
SF-2211 Do not show the verse too long error when suggestions are disabled

pmachapman:
When translation suggestions are disabled by the user, training still occurs in the background.

Because of this, the "verse too long" error message can appear when the user does not expect it (due to their having disabled translation suggestions in the Translation Settings dialog).

Included in this PR are two unit tests confirming the correct behavior for the "verse too long" notice when a verse is too long.

Please see [SF-1656](https://jira.sil.org/browse/SF-1656) for details on how to recreate this issue.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2059)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -446,6 +446,52 @@ describe('EditorComponent', () => {
       env.dispose();
     }));
 
+    it('should display the verse too long error', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig({ selectedBookNum: 40, selectedChapterNum: 1, selectedSegment: 'verse_1_5' });
+      env.wait();
+      expect(env.component.target!.segmentRef).toBe('verse_1_5');
+      expect(env.component.showSuggestions).toBe(true);
+
+      // Change to the long verse
+      const range = env.component.target!.getSegmentRange('verse_1_6');
+      env.targetEditor.setSelection(range!.index + range!.length, 0, 'user');
+      env.wait();
+
+      // Verify an error displayed
+      expect(env.component.target!.segmentRef).toBe('verse_1_6');
+      expect(env.component.showSuggestions).toBe(false);
+      verify(mockedNoticeService.show(anything())).once();
+
+      env.dispose();
+    }));
+
+    it('should not display the verse too long error if user has suggestions disabled', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig({
+        selectedBookNum: 40,
+        selectedChapterNum: 1,
+        selectedSegment: 'verse_1_5',
+        translationSuggestionsEnabled: false
+      });
+      env.wait();
+      expect(env.component.target!.segmentRef).toBe('verse_1_5');
+      // showSuggestions being true doesn't mean suggestions are shown, only that they could be if visible
+      expect(env.component.showSuggestions).toBe(true);
+
+      // Change to the long verse
+      const range = env.component.target!.getSegmentRange('verse_1_6');
+      env.targetEditor.setSelection(range!.index + range!.length, 0, 'user');
+      env.wait();
+
+      // Verify an error did not display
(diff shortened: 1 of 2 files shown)


RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2059)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1266 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2059#-NeEeBm_2otD3SjC7Hkr:-NeEeBm_2otD3SjC7Hks:b-ejrkeg) ([raw file](https://github.com/sillsdev/web-xforge/blob/12ddd7fc1f58f555e8d7f85d0da282c97d611214/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1266)):*
> ```XML
>   }
> 
>   private async translateSegment(): Promise<void> {
> ```

Is there a reason we call translateSegment when translation suggestions are disabled? Maybe a better fix would be to not translate segments or update suggestions at all if it is disabled.
In my testing, it looks like we generate suggestions even if translation suggestions are disabled. I don't think we need to do this.



_Code quote:_
```XML
  private async translateSegment(): Promise<void> {
```

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2059)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1266 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2059#-NeEeBm_2otD3SjC7Hkr:-NeFbHaJ-yaeH6lCZTD4:bxv14b5) ([raw file](https://github.com/sillsdev/web-xforge/blob/12ddd7fc1f58f555e8d7f85d0da282c97d611214/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1266)):*

Is there a reason we call translateSegment when translation suggestions are disabled? Maybe a better fix would be to not translate segments or update suggestions at all if it is disabled.
In my testing, it looks like we generate suggestions even if translation suggestions are disabled. I don't think we need to do this.
</blockquote></details>

A good question. We need to call `translateSegment()` when suggestions are disabled for a user is so that any changes made to the target text by the user incrementally update the model via the `InteractiveTranslator` object created by this method.

I briefly looked at ways we could optimize this, but it does look like we do need to have the background calls to `getWordGraph` made via machine.js to achieve this - i.e. all of the background work for generating translation suggestions has to take place for us to incrementally update the model, even though we are not displaying the suggestions popup.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2059)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1266 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2059#-NeEeBm_2otD3SjC7Hkr:-NeJFZYz0mfGuOIpdGNH:b57yerx) ([raw file](https://github.com/sillsdev/web-xforge/blob/05d8d1cc423511521c4b8ba7c96e682f3d5fef10/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1266)):*

A good question. We need to call `translateSegment()` when suggestions are disabled for a user is so that any changes made to the target text by the user incrementally update the model via the `InteractiveTranslator` object created by this method.

I briefly looked at ways we could optimize this, but it does look like we do need to have the background calls to `getWordGraph` made via machine.js to achieve this - i.e. all of the background work for generating translation suggestions has to take place for us to incrementally update the model, even though we are not displaying the suggestions popup.
</blockquote></details>

Thanks for looking into that. I was wondering if that is what we were doing, training the model in the background and just not providing the suggestions. I'm not sure what the benefit of that would be... I guess having an incrementally trained model will save time when the team does decide to enable translation suggestions. Ok, seems I talked my way into the use case.

PR #2061, opened 2023-09-13
SF-2231 Flaky audio tests

josephmyers:
This greatly improves performance for these tests. Theoretically it removes the flakiness, as well, since we no longer depend on playing actual audio files.

The new tests use a stub child component, which itself uses (via implicit override) a stub audio player, which itself uses a mocked html audio element. This all was necessary to work around the inabiliity to mock Angular child components and the limitations in ts-mockito surrounding property setters. ts-mockito doesn't allow verification on property set actions, and it frequently fails at runtime often ever reaching the verification. The current setup subclasses (you could use Partial, as well) AudioPlayer and overrides the critical methods and properties. This effectively removes any dependency on realtime audio data.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2061)


pmachapman (approved):
:lgtm: I like how much clearer the tests are - not just much more stable, but easier to understand and maintain. Thank you!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2061#-:-Neiz0zY5KtbWcqrzFMf:br4ajdu)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

PR #2064, opened 2023-09-13
SF-2221 Allow frontend feature flags to be set from the backend

pmachapman:
This PR allows frontend feature flags to be set via the backend.

This means that features could be rolled out to all users, or rolled back if the feature proves problematic.

The infrastructure for this PR could also be extended without too much difficulty to allow rolling out of features to specific users or groups of users.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2064)


Nateowami:
How does this impact how we use feature flags? Right now, we're hiding back translation drafting behind a feature flag the user can toggle on the front end. Which is nice, because then we can tell certain people how to turn it on.

If it's all set on the back end (and not configured on a per-user basis), then are we losing the ability to put the flags in the hands of the users?

It seems like perhaps we want two settings:
- "Has the user turned the flag on" (set on the front end by the user)
- "Is the flag turned on at a global level" (set on the back end by us)

And if either is true, then the feature is shown.

At least, that seems ideal to me for this particular feature. Not all features work that way.

pmachapman:
> How does this impact how we use feature flags? Right now, we're hiding back translation drafting behind a feature flag the user can toggle on the front end. Which is nice, because then we can tell certain people how to turn it on.
> 
> If it's all set on the back end (and not configured on a per-user basis), then are we losing the ability to put the flags in the hands of the users?

No, not at all. With this Pull Request, I am trying to allow the following:

1. **Global Rollout:** If a Feature Flag is enabled on the backend, enable it on the frontend and do not allow users to disable it.
2. **Global Rollback:** If a Feature Flag was enabled on the backend, but is now disabled on the backend, disable it on the frontend for all users. The user can manually reenable on the frontend after this takes place.
3. **Opt In:** If a Feature Flag is disabled on the backend, allow users to manually enable it on the frontend.

Currently, our system only supports step 3. This PR will not break that functionality. Control is only stripped from the user when a Feature Flag is enabled on the backend.

RaymondLuong3 (changes requested):
> Reviewed 37 of 37 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2064)* status: all files reviewed, 8 unresolved discussions (waiting on @pmachapman)

___
*[`scripts/db_tools/parse-version.ts` line 21 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-Ng5irEcCGfznHCIbvO8:-Ng5irEcCGfznHCIbvO9:b-v06uwq) ([raw file](https://github.com/sillsdev/web-xforge/blob/a3800133241e851c2c3b5d2b5eb76345da679ba0/scripts/db_tools/parse-version.ts#L21)):*
> ```XML
> };
> 
> class ParseVersion {
> ```

Neat tool. I was able to test it out and automatically see the feature flags enabled using the version number in SF. Anytime we add a feature flag, we will need to update this script, which is unfortunate. But otherwise, this works nicely.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts` line 32 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-Ng5Xv-dDtlDgUfuyYuA:-Ng5Xv-e2h78H3aI_H81:bxqs5kk) ([raw file](https://github.com/sillsdev/web-xforge/blob/a3800133241e851c2c3b5d2b5eb76345da679ba0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts#L32)):*
> ```XML
>     super();
>     // Cause the flags to be reloaded when coming online
>     if (onlineStatusService.onlineStatus$ != null) {
> ```

This check will always return true, since onlineStatus$ is set at initialization.



_Code quote:_
```XML
    if (onlineStatusService.onlineStatus$ != null) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts` line 95 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-Ng5ZZiR0mXIGatEuCQF:-Ng5ZZiR0mXIGatEuCQG:b-qirjed) ([raw file](https://github.com/sillsdev/web-xforge/blob/a3800133241e851c2c3b5d2b5eb76345da679ba0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts#L95)):*
> ```XML
>     if (this.remoteFlagCacheExpiry <= new Date() && this.onlineStatusService.isOnline) {
>       // Set to the next remote flag cache expiry timestamp for 1 hour so that the null check above returns false
>       this.remoteFlagCacheExpiry = new Date(new Date().getTime() + 360_000);
> ```

I could totally be wrong, but 1000*60*60 gives 3 600 000 as the value in milliseconds. So you would need to add a 0 on this value.



_Code quote:_
```XML
      this.remoteFlagCacheExpiry = new Date(new Date().getTime() + 360_000);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts` line 262 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-Ng5ajrtBik3a8PQkrWU:-Ng5ajrtBik3a8PQkrWV:b-dwpuom) ([raw file](https://github.com/sillsdev/web-xforge/blob/a3800133241e851c2c3b5d2b5eb76345da679ba0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts#L262)):*
> ```XML
>       value => value instanceof FeatureFlagFromStorage || value instanceof ServerOnlyFeatureFlag
>     );
>     featureFlags.sort((a, b) => (a.position < b.position ? -1 : 1));
> ```

I think this could be simplified to `featureFlags.sort((a, b) => a - b)`



_Code quote:_
```XML
    featureFlags.sort((a, b) => (a.position < b.position ? -1 : 1));
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts` line 272 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-Ng5b-flDy15oQ3gCOS_:-Ng5b-flDy15oQ3gCOSa:b-ottia2) ([raw file](https://github.com/sillsdev/web-xforge/blob/a3800133241e851c2c3b5d2b5eb76345da679ba0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts#L272)):*
> ```XML
> 
>   getFeatureFlagVersion(featureFlags: FeatureFlag[]): number {
>     var versionNumber: number = 0;
> ```

You can use let here. We tend to avoid using var.



_Code quote:_
```XML
    var versionNumber: number = 0;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.spec.ts` line 117 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-Ng5eIlR0y6Sf8H1uIWf:-Ng5eIlR0y6Sf8H1uIWg:b-h05no3) ([raw file](https://github.com/sillsdev/web-xforge/blob/a3800133241e851c2c3b5d2b5eb76345da679ba0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.spec.ts#L117)):*
> ```XML
>         { key: 'key_p', description: 'Key p', readonly: true, position: 15, enabled: false }
>       ])
>     ).toEqual(1 + 4 + 16 + 64 + 256 + 1024 + 4096 + 16384);
> ```

I haven't used binary counting in so long. Thanks for the mental exercise :astonished: 



_Code quote:_
```XML
      ])
    ).toEqual(1 + 4 + 16 + 64 + 256 + 1024 + 4096 + 16384);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.spec.ts` line 157 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-Ng5cviX8k_LyDWiEiIu:-Ng5cviX8k_LyDWiEiIv:b-8xs32l) ([raw file](https://github.com/sillsdev/web-xforge/blob/a3800133241e851c2c3b5d2b5eb76345da679ba0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.spec.ts#L157)):*
> ```XML
>         { key: 'key_5', description: 'Key 5', readonly: true, position: 31, enabled: true }
>       ])
>     ).toEqual(2147483648 + 1); // First and last bits
> ```

You can set a variable and use that here just to be extra clear.



_Code quote (i):_
```XML
    ).toEqual(2147483648 + 1); // First and last bits
```



_Code snippet (ii):_
```XML
const twoPower31: number = Math.pow(2,31);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flags-dialog.component.spec.ts` line 44 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-Ng5g8r13lBg4djK58p2:-Ng5g8r13lBg4djK58p3:b5zsvma) ([raw file](https://github.com/sillsdev/web-xforge/blob/a3800133241e851c2c3b5d2b5eb76345da679ba0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flags-dialog.component.spec.ts#L44)):*
> ```XML
> 
>     expect(env.getCheckboxLabel(1)).toBe('disabled flag');
>     expect(env.getMatCheckbox(1).disabled).toBeFalsy();
> ```

I get what you mean here. You are checking disabled as in the checkbox is disabled for a readonly feature flag. It think it is worth a comment.



_Code quote:_
```XML
  expect(env.getMatCheckbox(1).disabled).toBeFalsy();
```

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2064)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

pmachapman (commented):
So for some reason, I forgot to click Publish :-/

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-:-NgPqfuB4y3fZU2f65by:b-4t2gje)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts` line 32 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-Ng5Xv-dDtlDgUfuyYuA:-NgFcjuD7OFFN6VHbFZV:brs1m5m) ([raw file](https://github.com/sillsdev/web-xforge/blob/a3800133241e851c2c3b5d2b5eb76345da679ba0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts#L32)):*

This check will always return true, since onlineStatus$ is set at initialization.
</blockquote></details>

Done. I added the check for some tests that were failing, but have fixed the tests instead.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts` line 95 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-Ng5ZZiR0mXIGatEuCQF:-NgFfzNX94Nhm8LqL0Ut:b-y6ws8e) ([raw file](https://github.com/sillsdev/web-xforge/blob/a3800133241e851c2c3b5d2b5eb76345da679ba0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts#L95)):*

I could totally be wrong, but 1000*60*60 gives 3 600 000 as the value in milliseconds. So you would need to add a 0 on this value.
</blockquote></details>

Done. I forgot a zero. Thanks!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts` line 262 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-Ng5ajrtBik3a8PQkrWU:-NgFgpSK4b1v8JM6vEZc:bugs9as) ([raw file](https://github.com/sillsdev/web-xforge/blob/a3800133241e851c2c3b5d2b5eb76345da679ba0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts#L262)):*

I think this could be simplified to `featureFlags.sort((a, b) => a - b)`
</blockquote></details>

Done. Good spotting.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts` line 272 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-Ng5b-flDy15oQ3gCOS_:-NgFhD9lCZU7-4vrJqlc:bsr9umw) ([raw file](https://github.com/sillsdev/web-xforge/blob/a3800133241e851c2c3b5d2b5eb76345da679ba0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.ts#L272)):*

You can use let here. We tend to avoid using var.
</blockquote></details>

Done. Thanks! I think my mind thought I was programming in Swift for a minute...
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.spec.ts` line 117 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-Ng5eIlR0y6Sf8H1uIWf:-NgFj-GQ-GEnauITN_mJ:br9f8xj) ([raw file](https://github.com/sillsdev/web-xforge/blob/a3800133241e851c2c3b5d2b5eb76345da679ba0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.spec.ts#L117)):*

I haven't used binary counting in so long. Thanks for the mental exercise :astonished:
</blockquote></details>

Done. I have added comments next to each feature flag to help with this.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.spec.ts` line 157 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-Ng5cviX8k_LyDWiEiIu:-NgFiCwz7cQi3FCyyJIi:b-scnqdr) ([raw file](https://github.com/sillsdev/web-xforge/blob/a3800133241e851c2c3b5d2b5eb76345da679ba0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.spec.ts#L157)):*

You can set a variable and use that here just to be extra clear.
</blockquote></details>

Done. Good idea - that is a lot clearer.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flags-dialog.component.spec.ts` line 44 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2064#-Ng5g8r13lBg4djK58p2:-NgFky2HC_rh0YNHHHLZ:b7frlfo) ([raw file](https://github.com/sillsdev/web-xforge/blob/a3800133241e851c2c3b5d2b5eb76345da679ba0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flags-dialog.component.spec.ts#L44)):*

I get what you mean here. You are checking disabled as in the checkbox is disabled for a readonly feature flag. It think it is worth a comment.
</blockquote></details>

Done. I have added some comments and an additional test combination - please let me know if these are clearer?

PR #2066, opened 2023-09-14
SF-2229 Only allow admins and translators to see generate draft page

RaymondLuong3:
Add router guard so only admins and translators see the draft generation page. I know @Nateowami is working on an updated side nav, so maybe this is irrelevant now. If so, please say so and we can close this PR.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2066)


Nateowami (dismissed):
I think it would be easier to make a PR against my branch, rather than merging this separately. Otherwise we'll just end up deleting the change to the app component, and the tests.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2066#-:-NeLE-IJ6Vx3v_F54-RZ:bhijgw2)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app-routing.module.ts` line 25 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2066#-NeLCVmu0fnznaVnQikd:-NeLCVmu0fnznaVnQike:byz3r0y) ([raw file](https://github.com/sillsdev/web-xforge/blob/3892d3db57ddde5872aa72e51dfea1ffad4d148b/src/SIL.XForge.Scripture/ClientApp/src/app/app-routing.module.ts#L25)):*
> ```XML
>   { path: 'projects/:projectId/sync', component: SyncComponent, canActivate: [SyncAuthGuard] },
>   {
>     path: 'projects/:projectId/draft-generation',
> ```

I'm surprised to see a route being added, when a route for this already exists. Is there a reason it's being duplicated?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/project-router.guard.ts` line 94 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2066#-NeLDc6RCvtY06Tw8IpR:-NeLDc6RCvtY06Tw8IpS:b9wzyjy) ([raw file](https://github.com/sillsdev/web-xforge/blob/3892d3db57ddde5872aa72e51dfea1ffad4d148b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/project-router.guard.ts#L94)):*
> ```XML
>     return (
>       projectDoc.data != null &&
>       (projectDoc.data.userRoles[this.userService.currentUserId] === SFProjectRole.ParatextAdministrator ||
> ```

This should be using the rights service, not directly checking roles.

RaymondLuong3 (commented):
Good thinking. I've updated the PR to go against the side nav change.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2066#-:-NeOubgnDdEwGrhMLfrF:b-40vhio)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app-routing.module.ts` line 25 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2066#-NeLCVmu0fnznaVnQikd:-NeOuhmQC1kXQTfR7S-S:boty8ew) ([raw file](https://github.com/sillsdev/web-xforge/blob/3892d3db57ddde5872aa72e51dfea1ffad4d148b/src/SIL.XForge.Scripture/ClientApp/src/app/app-routing.module.ts#L25)):*

I'm surprised to see a route being added, when a route for this already exists. Is there a reason it's being duplicated?
</blockquote></details>

You're right! I totally missed that this was already in the translate routing module.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/project-router.guard.ts` line 94 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2066#-NeLDc6RCvtY06Tw8IpR:-NeOuoCnD2ONzpD4dZf7:bgbd732) ([raw file](https://github.com/sillsdev/web-xforge/blob/3892d3db57ddde5872aa72e51dfea1ffad4d148b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/project-router.guard.ts#L94)):*

This should be using the rights service, not directly checking roles.
</blockquote></details>

I understand what you mean, and I agree with you. I think that as much as we can, we should use the rights service. However, the rights service is designed to determine if a user has the right to access a doc in our database, and it doesn't necessarily give info about what pages a user should be able to see.
Since we can assume users that can edit text are either translators or admins, I updated it to use the rights service for that. So in this case, it is reasonable.

pmachapman (approved):
:lgtm:

> Reviewed 9 of 9 files at r2, all commit messages.
> Dismissed @Nateowami from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2066#-:-NeZg2HpB34kQEbXpNI8:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/project-router.guard.ts` line 94 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2066#-NeLDc6RCvtY06Tw8IpR:-NeZehsQDnA7Gb4EiVAb:bj79i7j) ([raw file](https://github.com/sillsdev/web-xforge/blob/3892d3db57ddde5872aa72e51dfea1ffad4d148b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/project-router.guard.ts#L94)):*

I understand what you mean, and I agree with you. I think that as much as we can, we should use the rights service. However, the rights service is designed to determine if a user has the right to access a doc in our database, and it doesn't necessarily give info about what pages a user should be able to see.
Since we can assume users that can edit text are either translators or admins, I updated it to use the rights service for that. So in this case, it is reasonable.
</blockquote></details>

LGTM I agree, the Edit rights check should be sufficient, even though the backend checks for Translator or Administrator roles.

PR #2068, opened 2023-09-15
Remove unneeded references to MDC top app bar

Nateowami:
This is doing two things:
1. Removing dead code in `src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts`
2. Removing a reference in the SCSS to `top-app-bar.$mobile-row-height`, and replacing it with `56px`, which is what the value was anyway.

I have marked this as "testing not required" because in order to test it the "update available" banner needs to be present. I tested it myself manually by turning the banner on and making sure it worked as expected.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2068)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
@@ -1,9 +1,10 @@
 @use 'src/variables';
 
 @use '@material/icon-button/mixins' as icon-button;
-@use '@material/top-app-bar/variables' as top-app-bar;
 
 @import 'bootstrap/scss/mixins/breakpoints';
+
+$toolbar-height: 56px;
 
 // Set height of the app to 100% so that the app will fill the entire screen.
 // Don't use 100vh because that will cause the app to be too tall on mobile devices when the browser's address bar is
@@ -50,11 +51,11 @@
   }
 
   .top-app-bar-adjust {
-    margin-top: top-app-bar.$mobile-row-height;
+    margin-top: $toolbar-height;
   }
 
   .top-app-bar-adjust-double {
-    margin-top: top-app-bar.$mobile-row-height * 2;
+    margin-top: $toolbar-height * 2;
   }
 
   .app-content {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
@@ -1,6 +1,5 @@
 import { MdcIconRegistry } from '@angular-mdc/web';
 import { MdcSelect } from '@angular-mdc/web/select';
-import { MdcTopAppBar } from '@angular-mdc/web/top-app-bar';
 import { Component, OnDestroy, OnInit, ViewChild } from '@angular/core';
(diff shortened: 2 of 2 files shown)


nigel-wells (approved):
Just one query but doesn't need to hold anything up if you have a good reason for it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2068#-:-NeaZov02yvr-hVOVmyA:b-l3w9yn)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2068#-NeaZW2zCiFbukA2ToFb:-NeaZW3-BiSx2uZtGSVa:b-hh3vix) ([raw file](https://github.com/sillsdev/web-xforge/blob/b6838f04bb04268939b5bf06905a41bba2d75b10/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L8)):*
> ```SCSS
> @import 'bootstrap/scss/mixins/breakpoints';
> 
> $toolbar-height: 56px;
> ```

Why would you not want to reference Material's design system for the height any more - what if Material decide to change these in the future? I realise it probably won't break anything but we wouldn't be following their guidelines anymore for the top bar height.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2068)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2068#-NeaZW2zCiFbukA2ToFb:-NeckL5j7u5XeMdMQr6q:b-n5p6oc) ([raw file](https://github.com/sillsdev/web-xforge/blob/b6838f04bb04268939b5bf06905a41bba2d75b10/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L8)):*

Why would you not want to reference Material's design system for the height any more - what if Material decide to change these in the future? I realise it probably won't break anything but we wouldn't be following their guidelines anymore for the top bar height.
</blockquote></details>

If you run `npm why @material/top-app-bar`, you'll get this:
```
@material/top-app-bar@5.1.0
node_modules/@material/top-app-bar
  @material/top-app-bar@"^5.1.0" from material-components-web@5.1.0
  node_modules/material-components-web
    material-components-web@"^5.1.0" from @angular-mdc/web@5.1.1
    node_modules/@angular-mdc/web
      @angular-mdc/web@"^5.1.1" from the root project
```
Notice that it's only installed because `@angular-mdc/web` requires it, and it's installed at version 5. The latest version is v14 (I'm a bit surprised; I thought it would keep up with Angular versions. A 15.0 canary version is frequently being published though).

The only reason we don't have problems right now, is because `@angular/material` is not based on official Material Components for Web, so the MDC and Material dependencies don't conflict. In the next version of Material, they started switching to the official Material Components for Web, so the dependencies do start to conflict, as they're going to be requiring the packages at around version 15, whereas we're currently installing them at version 5.

PR #2078, opened 2023-09-20
Remove Biblical term definition from form field

Nateowami:
Before | After
-------|------
 | 

Since the term isn't editable, we don't to make it appear that it is. The old design is mirroring Paratext. Maybe they have certain cases where you can edit the definition? I would expect that's probably the case, but we don't.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2078)


pmachapman (changes requested):
Looks good - just a couple of minor questions.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2078#-:-Neoll6B6rTGXQrnoOoT:b5hw143)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.spec.ts` line 180 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2078#-NeolOFk-FFG6vsiq4Al:-NeolOFk-FFG6vsiq4Am:b-bjgzej) ([raw file](https://github.com/sillsdev/web-xforge/blob/fcf046240a5f808b8eb6c8f9feb71c32c2d11b20/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.spec.ts#L180)):*
> ```XML
>   get term(): string | null {
>     return this.overlayContainerElement.querySelector('#term')?.textContent ?? null;
>   }
> ```

I wonder if both this and the tag id should be "definition" for consistency?

Also, is there a reason you use `string | null`, not `string | undefined`?

```ts
get term(): string | undefined {
    return this.overlayContainerElement.querySelector('#term')?.textContent ?? undefined;
  }
```



_Code quote:_
```XML
  get term(): string | null {
    return this.overlayContainerElement.querySelector('#term')?.textContent ?? null;
  }
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2078)* status: 3 of 5 files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.spec.ts` line 180 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2078#-NeolOFk-FFG6vsiq4Al:-Nepi7yl-8GzXd33X3HF:b378jw6) ([raw file](https://github.com/sillsdev/web-xforge/blob/fcf046240a5f808b8eb6c8f9feb71c32c2d11b20/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-term-dialog.component.spec.ts#L180)):*

I wonder if both this and the tag id should be "definition" for consistency?

Also, is there a reason you use `string | null`, not `string | undefined`?

```ts
get term(): string | undefined {
    return this.overlayContainerElement.querySelector('#term')?.textContent ?? undefined;
  }
```
</blockquote></details>

Changed to "definition".

null or undefined both work. As discussed, Element.querySelector returns an Element or null, which is why I opted for null in this case.

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2078#-:-Neq0JyfAnn6CE3hETCm:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2080, opened 2023-09-21
SF-2250 Remove DBL resource access on sync for invalid tokens

pmachapman:
This PR fixes a crash that occurs during synchronization with Paratext when a users refresh tokens have been corrupted and we are attempting to use their credentials to access the DBL to see if they have access to a resource.

Instead of crashing, this PR removes their access to the resource, as there really is no way of knowing if they should have access, and if they log in to Scripture Forge again in the future and sync the project their access will be restored.

This PR also fixes the exception handlers so that they return the HTTP status code in the `HttpRequestException` when  `IExceptionHandler.EnsureSuccessStatusCode(HttpResponseMessage response)` is called.

This PR should be tested by a developer using the steps outlined in the [JIRA ticket](https://jira.sil.org/browse/SF-2250) as the nature of modifying database records to create an inconsistent state is outside of the usual scope of the test team.


Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2080#-:-NesWeOB9RfSlKUyp-qZ:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2094, opened 2023-09-27
Set audio component source intentionally - Proof of concept

marksvc:
This proof of concept removes the audio source from being an `@Input` to the audio player. Instead of the source being set in the template, the source is passed to the audio player when the chapter changes. 

Because of this, we don't set an empty source at page-construction time. Therefore, the audio player can know that source setting instances are intentionally done by the parent/owner.

Here the audio player has the responsibility to say whether the audio is available or not.

This code isn't meant to be super well designed, but showing the idea.

Video showing behaviour:
[set-source-intentionally.webm](https://github.com/sillsdev/web-xforge/assets/7265309/4a63c4c0-6b3b-4037-a72c-8e835501d62e)

Note: This PR is targeting an unmerged commit so that the relevant changes are shown.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2094)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.html
@@ -10,7 +10,7 @@
       <mat-icon class="mirror-rtl">skip_next</mat-icon>
     </button>
   </div>
-  <button mat-icon-button class="close-button" (click)="close()">
+  <button *ngIf="canClose" mat-icon-button class="close-button" (click)="close()">
     <mat-icon>close</mat-icon>
   </button>
   <app-audio-player #audioPlayer [source]="audioSource"></app-audio-player>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.html
@@ -13,6 +13,6 @@
   <button *ngIf="canClose" mat-icon-button class="close-button" (click)="close()">
     <mat-icon>close</mat-icon>
   </button>
-  <app-audio-player #audioPlayer [source]="audioSource"></app-audio-player>
+  <app-audio-player #audioPlayer></app-audio-player>
   <span class="verse-label">{{ verseLabel }}</span>
 </div>
--- a/src/RealtimeServer/scriptureforge/models/checking-config.ts
+++ b/src/RealtimeServer/scriptureforge/models/checking-config.ts
@@ -10,4 +10,5 @@ export interface CheckingConfig {
   shareEnabled: boolean;
   answerExportMethod: CheckingAnswerExport;
   noteTagId?: number;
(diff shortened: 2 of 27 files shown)


Nateowami:
Can you update the PR title to be a bit more descriptive? "Set source intentionally" is very vague. Maybe "Set audio component source ..."? I'm also not sure what "intentionally" is supposed to imply.

Maybe you could describe what you're trying to accomplish, rather than what the solution is? E.g. "Fix audio unavailable message when switching between audio-only questions"?

marksvc:
This draft PR relates to some discussion in chat regarding handling what to do when the source audio is not available, or just not yet set, and showing an "Audio unavailable" message.

In the current design, the audio source is set in the template, and so starts out with an initial value before we know what the real audio source is. This PR shows a way that we could alter the code so that we don't set the source in the template, with a consequence that we only set the audio source based on what we think the source is, rather than angular automatically doing it for us when we sort of aren't ready. 

But the PR is best considered in light of the chat discussion. This is one of a few approaches.

I probably should have put in a message like "See Slack discussion" so viewers wouldn't try to interpret the PR on its own. Sorry :)

Nateowami:
After looking through the discussion in Slack and the changes here, I think I've reverse-engineered what this is trying to do: "Show disabled audio player when chapter audio unavailable"

I just checked what tests are available for the various audio player components, and I see that they do have stories for the state where no audio is set. But they all show up blank when no audio is set, so now I'm still confused what the intent of this is. I thought you were pursing the "show a disabled player" option, but that doesn't appear to be what this does.

src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.html
```diff
@@ -13,6 +13,6 @@
   <button *ngIf="canClose" mat-icon-button class="close-button" (click)="close()">
     <mat-icon>close</mat-icon>
   </button>
-  <app-audio-player #audioPlayer [source]="audioSource"></app-audio-player>
```
Nateowami, line 16: Why can't this be something like:
``` html
<app-audio-player *ngIf="audioSource != null" #audioPlayer [source]="audioSource"></app-audio-player>`
```

marksvc:
Hey @Nateowami , thanks for your comments. I'm not sure if I've been clear what the purpose of this PR is :). It's demonstrating a proof of concept for doing one of the possible implementation directions mentioned in Slack. This is a draft PR to put this code where it can be seen as part of such a discussion. 

The problem at hand is that there are a few needs that we need to triangulate, or walk between successfully: 
- Look nice when we are loading.
- Do something reasonable when there is really no audio.
- Look nice when fully loaded.

Complications involve the timing of when we can know the height of the fully-created audio player, when we can know that there is really no audio, and how we can tell the audio player whether there really is audio or not (if we even should). 

This draft PR shows one possible implementation, which is by only writing to the audio player's `source` when we actually have a source, rather than when we don't (which the behaviour on `master` branch does). We should probably be having this discussion on the phone :)

> Why can't this be something like:
> `<app-audio-player *ngIf="audioSource != null" #audioPlayer [source]="audioSource">`</app-audio-player>`

Yes, it can be. A similar approach is shown in #2095, and is another way to try to reach the 3 needs I mention above, by not putting the audio player in the CheckingComponent DOM at all. I see where you are commenting is at a bit of a different level, but similar.

PR #2102, opened 2023-10-03
Remove backend push SF notes feature flag

RaymondLuong3:
If we need to disable the writing notes to Paratext, we would need to disable writing notes and receiving notes, otherwise notes created in SF will disappear after synchronizing. So this feature flag on its own isn't very helpful.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2102)


marksvc (approved):
Okay. I assume it would be okay to leave this in in case it would be helpful (perhaps in combination with switching off something on the front-end). But we can take it out if you think that's the right direction.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2102#-:-NfqWFH93V5vUjEjX89d:b-et3pc0)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2116, opened 2023-10-10
SF-2276 Hide font size changer when audio-only

josephmyers:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2116)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html
@@ -103,7 +103,7 @@
               >
                 <mat-icon>{{ isAudioPlaying() ? "pause" : "play_arrow" }}</mat-icon>
               </button>
-              <app-font-size (apply)="applyFontChange($event)"></app-font-size>
+              <app-font-size [class.hidden]="hideChapterText" (apply)="applyFontChange($event)"></app-font-size>
               <app-share-button *ngIf="canShare" [defaultRole]="defaultShareRole"></app-share-button>
             </div>
           </div>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss
@@ -66,6 +66,13 @@ header {
     }
   }
 
+  .active-question-filter {
+    font-size: 0.7em;
+    opacity: 0.5;
+    margin-top: -5px;
+    margin-inline-start: 2px;
+  }
+
   .questions-overlay-close {
     display: none;
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss
@@ -73,7 +73,8 @@ header {
     margin-inline-start: 2px;
   }
 
-  .questions-overlay-close {
+  .questions-overlay-close,
+  .hidden {
     display: none;
   }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss
@@ -66,13 +66,6 @@ header {
     }
   }
 
-  .active-question-filter {
-    font-size: 0.7em;
-    opacity: 0.5;
-    margin-top: -5px;
-    margin-inline-start: 2px;
-  }
-
   .questions-overlay-close,
   .hidden {
     display: none;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts
@@ -137,6 +137,7 @@ export class CheckingComponent extends DataLoadingComponent implements OnInit, A
   private questionsSub?: Subscription;
   private textAudioQuery?: RealtimeQuery<TextAudioDoc>;
   private projectDeleteSub?: Subscription;
+  private hideTextSub?: Subscription;
   private projectRemoteChangesSub?: Subscription;
   private questionFilterFunctions: Record<QuestionFilter, (answers: Answer[]) => boolean> = {
     [QuestionFilter.None]: () => true,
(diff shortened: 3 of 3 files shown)


Nateowami:
This is extremely minor, but for consistency, this should be titled "Hide font size changer when audio only". Using imperative mood in the subject line of a commit is considered best practice and makes it easier to read a list of commit messages.

RaymondLuong3 (approved):
I agree, keeping commit messages in the same mood is good.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2116#-:-NgUnZTo3Ew89RnaYfJf:bno9kw2)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 555 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2116#-NgUnIFg3mBrw3GeP5Id:-NgUnIFg3mBrw3GeP5Ie:bx71lst) ([raw file](https://github.com/sillsdev/web-xforge/blob/8cda2de8c4c89af183ad2751f4b201a48f414159/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L555)):*
> ```XML
> 
>                 for (const op of ops) {
>                   if (op.p.length > 1 && op.p[0] === 'checkingConfig' && op.p[1] === 'hideCommunityCheckingText') {
> ```

This might be fragile if we ever decide to change the names of the properties, but considering that is not too likely, I'll let this slide.



_Code quote:_
```XML
                  if (op.p.length > 1 && op.p[0] === 'checkingConfig' && op.p[1] === 'hideCommunityCheckingText') {
```

josephmyers:
> 
> This might be fragile if we ever decide to change the names of the properties, but considering that is not too likely, I'll let this slide.
> 
> _Code quote:_
> 
> ```
>                   if (op.p.length > 1 && op.p[0] === 'checkingConfig' && op.p[1] === 'hideCommunityCheckingText') {
> ```

I agree it's brittle, but it appears to be the API we have for detecting actual changes to the realtime docs. Flexible and scalable, but brittle. I can extract them out to constants, but that would only reduce the brittleness.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2116)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 555 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2116#-NgUnIFg3mBrw3GeP5Id:-NgcffprCALdV1v4FqRj:bcm3rqd) ([raw file](https://github.com/sillsdev/web-xforge/blob/8cda2de8c4c89af183ad2751f4b201a48f414159/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L555)):*
> it appears to be the API we have for detecting actual changes to the realtime docs

This should be improved upon; it's too unmaintainable.

For one thing, I don't think you only want remote changes; I think if the changes were made locally you'd want to pick up on them as well.

Secondly, I can't think of a reason you couldn't just access the data from the project doc rather than access the ops themselves. If you want to make this only run when it changes, you could do something like:

``` ts
this.projectDoc.changes$.pipe(map([
  () => this.projectDoc.data.checkingConfig.hideCommunityCheckingText,
  distinctUntilChanged()
])).subscribe(() => this.calculateScriptureSliderPosition())
```

Don't try to copy what I just wrote; it's a concept and almost certainly not valid RxJS, though I think you can do what I've shown if you know the right operators and how to connect them.

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2116)* status: 1 of 3 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 555 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2116#-NgUnIFg3mBrw3GeP5Id:-NgvZ2byDeu_lOMKnBSs:b-486q75) ([raw file](https://github.com/sillsdev/web-xforge/blob/8cda2de8c4c89af183ad2751f4b201a48f414159/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L555)):*

> it appears to be the API we have for detecting actual changes to the realtime docs

This should be improved upon; it's too unmaintainable.

For one thing, I don't think you only want remote changes; I think if the changes were made locally you'd want to pick up on them as well.

Secondly, I can't think of a reason you couldn't just access the data from the project doc rather than access the ops themselves. If you want to make this only run when it changes, you could do something like:

``` ts
this.projectDoc.changes$.pipe(map([
  () => this.projectDoc.data.checkingConfig.hideCommunityCheckingText,
  distinctUntilChanged()
])).subscribe(() => this.calculateScriptureSliderPosition())
```

Don't try to copy what I just wrote; it's a concept and almost certainly not valid RxJS, though I think you can do what I've shown if you know the right operators and how to connect them.
</blockquote></details>

Thanks, Nathaniel, that's what I was looking for. This is less familiar to me, coming from a C# background, so I'm glad for the help.

The "ops" parsing is also present in checking-answers, though that doesn't mean it's okay. I certainly like your method better.

RaymondLuong3 (changes requested):
A little more tidying up, then it's good to go.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2116#-:-Nh2amow7v80jcsnm1en:b-rtpwny)* status: all files reviewed, 2 unresolved discussions (waiting on @josephmyers and @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 555 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2116#-NgUnIFg3mBrw3GeP5Id:-Nh2_KZQEIk2NFOO2zFN:bt4n9pc) ([raw file](https://github.com/sillsdev/web-xforge/blob/8cda2de8c4c89af183ad2751f4b201a48f414159/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L555)):*

Thanks, Nathaniel, that's what I was looking for. This is less familiar to me, coming from a C# background, so I'm glad for the help.

The "ops" parsing is also present in checking-answers, though that doesn't mean it's okay. I certainly like your method better.
</blockquote></details>

I like this much better.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 554 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2116#-Nh2_r-NAlG5T0xsCqUj:-Nh2_r-NAlG5T0xsCqUk:bq6oz6g) ([raw file](https://github.com/sillsdev/web-xforge/blob/242831ad1726dbc4a84f23974a21e632ac8fb790/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L554)):*
> ```XML
>             });
> 
>             this.projectDoc.changes$
> ```

You will want to assign this to a variable, and then unsubscribe before creating a new subscription when the project doc or query params change. Additionally, use the `subscribe()` method of the SubscriptionDisposible base class. Then, you can add the unsubscribe to the dispose call if one exists for this class, but that is not necessary since it will be disposed automatically.

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2116)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 554 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2116#-Nh2_r-NAlG5T0xsCqUj:-Nh4RkL894qVhQl7Ws0k:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/242831ad1726dbc4a84f23974a21e632ac8fb790/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L554)):*

You will want to assign this to a variable, and then unsubscribe before creating a new subscription when the project doc or query params change. Additionally, use the `subscribe()` method of the SubscriptionDisposible base class. Then, you can add the unsubscribe to the dispose call if one exists for this class, but that is not necessary since it will be disposed automatically.
</blockquote></details>

Done.

PR #2124, opened 2023-10-17
SF-2266 Resize bug when audio-only

nigel-wells:
* Calculate scripture slider position on window resize

This fix relates to changing the split component size and then dynamically resizing the screen.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2124)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
@@ -202,10 +202,11 @@ describe('CheckingComponent', () => {
       const testProject: SFProject = TestEnvironment.generateTestProject();
       testProject.checkingConfig.hideCommunityCheckingText = true;
       const env = new TestEnvironment({ user: CHECKER_USER, testProject });
-      const spySliderPosition = spyOn<any>(env.component, 'calculateScriptureSliderPosition').and.callThrough();
-      window.dispatchEvent(new Event('resize'));
       env.waitForSliderUpdate();
-      expect(spySliderPosition).toHaveBeenCalledTimes(1);
+      (env.component as any)._scriptureAreaMaxSize = 1;
+      expect(env.component.scriptureAreaMaxSize).toEqual(1);
+      window.dispatchEvent(new Event('resize'));
+      expect(env.component.scriptureAreaMaxSize).toBeGreaterThan(1);
       flush();
       discardPeriodicTasks();
     }));
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
@@ -198,6 +198,18 @@ describe('CheckingComponent', () => {
       expect(nextQuestion).toEqual(1);
     }));
 
+    it('should re-calculate scripture slide position on resize', fakeAsync(() => {
+      const testProject: SFProject = TestEnvironment.generateTestProject();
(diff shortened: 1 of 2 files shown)


josephmyers (commented):
Hey, I'm wondering what your reasoning was on this. Master makes sense to me. Can you post the bug steps that this fixes? I think we've departed from the original bug reported by test.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2124#-:-NgwMQjPA2E2vkizXTuI:b-4m0paw)* status: 0 of 2 files reviewed, all discussions resolved

josephmyers:
Master

https://github.com/sillsdev/web-xforge/assets/43733878/ca0c03a9-f60b-4869-a72e-fcf6d1a48762

josephmyers:
Branch:

https://github.com/sillsdev/web-xforge/assets/43733878/0e03e2fc-bcb3-45f0-926e-910f2af3d120

josephmyers:
Branch:
https://github.com/sillsdev/web-xforge/assets/43733878/bf7a3e6e-fee1-4b62-ac13-0f5cb53b4398

nigel-wells (commented):
There is an observable on the media breakpoints changing. If they don't change, however, then there is now check on what the size of everything should be. The gif and images in [SF-2266](https://jira.sil.org/browse/SF-2266) are more targeted at the chapter audio player (with no scripture text) rather than the interaction between scripture text and the Q/A panel.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2124#-:-Nh-CZ7K79iJwW2wUz6m:b-c4nbf8)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (commented):
Right, maybe I wasn't clear. I think this fix makes the user experience less helpful when there's scripture text, in that it's always undoing your pane sizings. Did you test your changes with the setting switched off?

As for resizing with the setting on, it makes sense to me that the audio player might be hidden a little if the user shrinks the window, the problem being that (in master) you actually can't drag the player down any more than what's pictured. The code is limiting the draggability to the audio player size, but it's calculating this size incorrectly when the window resizes.

So to be super clear, you did fix this use case. But this change at the same time makes resizing the window in all the other cases inconvenient. I'll upload a video of a potential use case to demonstrate.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2124#-:-Nh-HqazC7LdjgbpuStf:b-js2ixm)* status: 0 of 2 files reviewed, all discussions resolved

josephmyers:
https://github.com/sillsdev/web-xforge/assets/43733878/acfec81f-5a69-473b-b1b5-5a43c8eed44f

josephmyers:
Phew, I think I got it. Very annoying how complicated this code is, and mostly because the splitter library we're using is not great. The bug I ran into that I mentioned on Slack was due to the splitter component ignoring maxSize changes, e.g. setting it to 10 on resize does nothing. We really need a refactor. Are there other splitter libraries

josephmyers (approved):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2124)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

nigel-wells:
Thanks for the rename @josephmyers

Nateowami:
I hope the bug is smaller, now that it has been resized.

Perhaps the title could be a bit clearer if it were "SF-2266 Fix splitter resize bug when audio-only". Since our commit messages are usually written with imperative mood and start with a verb, when you read through a list of them you're going to see something like:

- Fix xyz
- Update abc
- Convert x to y
- Remove a from b
- Resize bug when audio only

PR #2130, opened 2023-10-18
SF-2291 Add queue depth visibility to Pre-Translation drafting

pmachapman:
This PR passes the Serval queue depth value through to the Scripture Forge front end.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2130)


marksvc:
*[`src/SIL.XForge.Scripture/Models/ServalBuildDto.cs` line 7 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2130#-NhhrEnyCgcqzYyazYlz:-NhhxQd90GcRXogj5Fk_:b-2jvfgy) ([raw file](https://github.com/sillsdev/web-xforge/blob/a01174fa75ba4f225c88407a1d6899fdaba3d99d/src/SIL.XForge.Scripture/Models/ServalBuildDto.cs#L7)):*

I'm surprised this wasn't something in BuildDto.  Hmm, I guess I also haven't seen anywhere were this gets set either :)
</blockquote></details>

Oops. Never mind.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2130)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Models/ServalBuildDto.cs` line 7 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2130#-NhhrEnyCgcqzYyazYlz:-Nhwv6zs2WlqISKvcAv1:b-et79nl) ([raw file](https://github.com/sillsdev/web-xforge/blob/a01174fa75ba4f225c88407a1d6899fdaba3d99d/src/SIL.XForge.Scripture/Models/ServalBuildDto.cs#L7)):*

Oops. Never mind.
</blockquote></details>

`BuildDto` is from the older SIL.Machine version used by the in-process Machine, which does not support `QueueDepth`. In a future build of SF after we will disable the in-process Machine instance, we will remove SIL.Machine completely and implement any models entirely in SF.

marksvc (approved):
> Reviewed 7 of 7 files at r6, 4 of 4 files at r7, 1 of 1 files at r9, 1 of 1 files at r10, 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2130)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html` line 188 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/2130#-NiGEydt7YYTx1vPu5S9:-NiGEydt7YYTx1vPu5SA:bxy7wq) ([raw file](https://github.com/sillsdev/web-xforge/blob/c8ca148a41343a9df221c0e8c6b64151929dc1c5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html#L188)):*
> ```HTML
>                 <p *ngIf="(draftJob?.queueDepth ?? 0) === 0">{{ t("draft_queued_detail") }}</p>
>                 <p *ngIf="(draftJob?.queueDepth ?? 0) > 0">
>                   {{ t("draft_queued_detail_multiple", { count: draftJob.queueDepth }) }}
> ```

I might have thought this needs to be  `draftJob?.queueDepth ?? 0`. But the IDE complains if I make changes like that. But DraftGenerationComponent.draftJob is type `BuildDto | undefined`, so shouldn't it be important to account for draftJob not being defined?
___
*[`src/SIL.XForge.Scripture/Models/ServalBuildDto.cs` line 7 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2130#-NhhrEnyCgcqzYyazYlz:-Ni6i9Q_EeSj7xVXvd_o:bt1kwye) ([raw file](https://github.com/sillsdev/web-xforge/blob/a01174fa75ba4f225c88407a1d6899fdaba3d99d/src/SIL.XForge.Scripture/Models/ServalBuildDto.cs#L7)):*

`BuildDto` is from the older SIL.Machine version used by the in-process Machine, which does not support `QueueDepth`. In a future build of SF after we will disable the in-process Machine instance, we will remove SIL.Machine completely and implement any models entirely in SF.
</blockquote></details>

Thank you, that is a helpful additional explanation.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2130)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html` line 188 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/2130#-NiGEydt7YYTx1vPu5S9:-NiG_7Iy16cPpChaLki7:b-3hhqmt) ([raw file](https://github.com/sillsdev/web-xforge/blob/c8ca148a41343a9df221c0e8c6b64151929dc1c5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html#L188)):*

I might have thought this needs to be  `draftJob?.queueDepth ?? 0`. But the IDE complains if I make changes like that. But DraftGenerationComponent.draftJob is type `BuildDto | undefined`, so shouldn't it be important to account for draftJob not being defined?
</blockquote></details>

The `ngIf*` above ensures that `draftJob` is not null, as `queueDepth` must be over zero to call this block.

marksvc:
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html` line 188 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/2130#-NiGEydt7YYTx1vPu5S9:-NiGprqK8SZj-TkZHH02:bjoutgg) ([raw file](https://github.com/sillsdev/web-xforge/blob/c8ca148a41343a9df221c0e8c6b64151929dc1c5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html#L188)):*
Ah, good point. Though if I change the code (just for learning) to the following,
```html
<p *ngIf="(draftJob?.queueDepth ?? 0) === 0">
{{ t("draft_queued_detail_multiple", { count: draftJob?.queueDepth }) }}
</p>
```
It still complains, saying 
> The left side of this optional chain operation does not include 'null' or 'undefined' in its type, therefore the '?.' operator can be replaced with the '.' operator.

This makes me think its prior complaint is not based on it being really clever and knowing that `draftJob` must not be null since `draftJob?.queueDepth ??0` is `> 0`.

marksvc:
I got some projects set up successfully to be able to test this today. 

I notice that the project that is needing to wait, starts off saying 
> The back translation draft has been queued. Please check back later.

And it doesn't very soon announce that it has a position in the queue. 

Also, I don't know if this is from me experimenting with the template code, but I notice that it says I am number 0 in the queue, rather than number 1 in the queue. I haven't gotten it to tell me what position I am in the queue recently🗡, so I can't confirm if this is a behaviour from your code or not. 

🗡 Because it is still saying "The back translation draft has been queued. Please check back later." Hmm, if it keeps doing this, tho, then the PR isn't really working for me hey :/

marksvc:
Okay, if I start two draft generations at the same time, one of them will act with queueDepth 0, and the other will say I am position 1 in the queue. So that's good.

If I have a back translation generation _in progress_, with a percentage finished, and on another project I request a generation, then it acts with queueDepth 0. 
Presumably if I had yet a 3rd project, and requested a generation, then the 3rd project would say it is in position 1 in the queue.

Regarding the message:

> Draft queued
> The back translation draft has been queued. Please check back later.

It's not super clear to me that the project is in fact queued behind another project that may take _hours_ before my project starts to get generated. 

It's okay like it is for now. But maybe a future improvement might be something like instead saying: "Generation of the back translation draft has been queued and will begin once the server is finished with another project. Please check back later."
And, "The back translation draft generation is number {{ count }} in the queue and will begin once the server is finished with other projects. Please check back later."

pmachapman (commented):
I have updated the wording for the message as I think your wording is much clearer than the current message text. (Thank you!)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2130#-:-NiVhMDJB310yW3tZjHr:bg0qjp6)* status: 5 of 15 files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html` line 188 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/2130#-NiGEydt7YYTx1vPu5S9:-NiVisN84jA22oJPWf_g:b-7y1azk) ([raw file](https://github.com/sillsdev/web-xforge/blob/c8ca148a41343a9df221c0e8c6b64151929dc1c5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html#L188)):*

Ah, good point. Though if I change the code (just for learning) to the following,
```html
<p *ngIf="(draftJob?.queueDepth ?? 0) === 0">
{{ t("draft_queued_detail_multiple", { count: draftJob?.queueDepth }) }}
</p>
```
It still complains, saying 
> The left side of this optional chain operation does not include 'null' or 'undefined' in its type, therefore the '?.' operator can be replaced with the '.' operator.

This makes me think its prior complaint is not based on it being really clever and knowing that `draftJob` must not be null since `draftJob?.queueDepth ??0` is `> 0`.
</blockquote></details>

In the example statement you give of `(draftJob?.queueDepth ?? 0) === 0`, `draftJob` can be null for this condition to be true, so it requires the null-conditional operator in the body - this is a correct warning from the IDE in that case.

For statement `(draftJob?.queueDepth ?? 0) > 0`, there is no way that the statement can be true if `draftJob` is null, so the null-conditional operator is not required in the body.

marksvc (approved):
> Reviewed 9 of 10 files at r12, 4 of 4 files at r13, 1 of 1 files at r14, 1 of 1 files at r15, 1 of 1 files at r16, 2 of 2 files at r17, 1 of 1 files at r18, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2130)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html` line 188 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/2130#-NiGEydt7YYTx1vPu5S9:-NiaVLX52hXy7jCzRlgO:b-3o1nfm) ([raw file](https://github.com/sillsdev/web-xforge/blob/c8ca148a41343a9df221c0e8c6b64151929dc1c5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html#L188)):*
> can be null for this condition to be true, so it requires the null-conditional operator in the body - this is a correct warning from the IDE in that case.

But since `draftJob` can be null, and since one would thereforeuse the `?.` null-conditional operator, doesn't that mean that the IDE is incorrectly saying that `?.` can be replaced with `.`? For a somewhat cleaner situation, consider the following code in the screenshot below:



This is above the `*ngIf` lines, so it is not inside their scope. I'm querying a property of `draftJob` and using the `?.` operator. But the IDE complains, saying I can replace `?.` with `.`. 

From your reply, it sounds like we both are thinking the same way about whether it needs the `?.` operator. And I'm confused why the IDE appears to be saying otherwise (unless I really don't understand what the IDE is saying). If I change `?.` in the screenshot to `.`, then the IDE complaint goes away. I had a look at https://angular.io/extended-diagnostics/NG8107 but without finding what I'm missing.

I guess it's not super important that we get to the bottom of it, but I may learn something here.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2130)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html` line 188 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/2130#-NiGEydt7YYTx1vPu5S9:-NibGeN99qqyQGfv8Mwx:bwhvebm) ([raw file](https://github.com/sillsdev/web-xforge/blob/c8ca148a41343a9df221c0e8c6b64151929dc1c5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html#L188)):*

> can be null for this condition to be true, so it requires the null-conditional operator in the body - this is a correct warning from the IDE in that case.

But since `draftJob` can be null, and since one would thereforeuse the `?.` null-conditional operator, doesn't that mean that the IDE is incorrectly saying that `?.` can be replaced with `.`? For a somewhat cleaner situation, consider the following code in the screenshot below:



This is above the `*ngIf` lines, so it is not inside their scope. I'm querying a property of `draftJob` and using the `?.` operator. But the IDE complains, saying I can replace `?.` with `.`. 

From your reply, it sounds like we both are thinking the same way about whether it needs the `?.` operator. And I'm confused why the IDE appears to be saying otherwise (unless I really don't understand what the IDE is saying). If I change `?.` in the screenshot to `.`, then the IDE complaint goes away. I had a look at https://angular.io/extended-diagnostics/NG8107 but without finding what I'm missing.

I guess it's not super important that we get to the bottom of it, but I may learn something here.
</blockquote></details>

Hmmm. For the example you give, `isDraftQueued` will only return true (and execute the block) if `draftJob` is not null, so again warning against a `?.` is appropriate.

It seems that the [TypeScript language service](https://github.com/microsoft/TypeScript/wiki/Using-the-Language-Service-API) is far more advanced than static analysis tools I have worked with in the past in other languages.

marksvc:
In experimenting with that some. I find that it doesn't seem to be basing it on `isDraftQueued()`. But looking up higher, I see an obvious thing that I had missed before, which is 
```html
  <ng-container *ngIf="draftJob != null">
```
So that explains it :)

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2130)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2136, opened 2023-10-23
Refactor migrations to reuse migrateOp

pmachapman:
Every migration implements an empty op migration:

```ts
  migrateOp(_op: RawOp): void {
    // do nothing
  }
```

This PR creates an abstract class `DocMigration` that implements this empty op migration.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2136)

### DIFF
--- a/src/RealtimeServer/common/migration.ts
+++ b/src/RealtimeServer/common/migration.ts
@@ -25,3 +25,11 @@ export interface Migration {
    */
   migrateOp(op: RawOp): void;
 }
+
+export abstract class DocMigration implements Migration {
+  abstract migrateDoc(doc: Doc): Promise<void>;
+
+  migrateOp(_op: RawOp): void {
+    // do nothing
+  }
+}
--- a/src/RealtimeServer/scriptureforge/services/note-thread-migrations.ts
+++ b/src/RealtimeServer/scriptureforge/services/note-thread-migrations.ts
@@ -1,8 +1,8 @@
-import { Doc, Op, RawOp } from 'sharedb/lib/client';
+import { Doc, Op } from 'sharedb/lib/client';
 import { submitMigrationOp } from '../../common/realtime-server';
-import { Migration, MigrationConstructor } from '../../common/migration';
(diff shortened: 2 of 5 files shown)


Nateowami:
Why does the `migrateOp` method exist at all if we aren't using it?

pmachapman:
> Why does the `migrateOp` method exist at all if we aren't using it?

It acts as a way to override the default op transform during a migration. We haven't needed to do anything that complex yet, so we could remove it if you think that is viable (we could always just restore the code from the PR that removes it if we need this in future).

Nateowami:
That makes sense. I've always wondered about it. Do you know anything more detailed than `We haven't needed to do anything that complex yet`? In what context would we need to? Are there edge cases that we're just not handling and it just hasn't bitten us yet?

pmachapman:
> Do you know anything more detailed than `We haven't needed to do anything that complex yet`? In what context would we need to? Are there edge cases that we're just not handling and it just hasn't bitten us yet?

Unfortunately I don't. That might be a question for @ddaspit? I cannot find much documentation in the PR that created the change, or any examples that might have created the need for it.

ddaspit:
It was intended to deal with offline clients. A normal When an offline client comes online, they might have ops that need to be migrated prior to being applied to the MongoDB in order to avoid data corruption.

Nateowami:
@ddaspit That's what I've always assumed. But none of our data migrations have implemented it, and I don't know what the side-effects of not migrating ops might be, or how we would go about migrating them.

ddaspit:
You would primarily need it if you made a breaking change to a model. An example would be if you renamed a property in a model. If an offline client had an op that updated the renamed property, you would want to migrate the op by renaming the property, otherwise the update would be lost.

nigel-wells (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2136)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2148, opened 2023-10-27
SF-2272 Prevent corrupting text when user deletes all verse text offline

RaymondLuong3:
When editing a text doc offline, the ops for the edits are stored in IndexedDB and sent to the real-time server when coming back online. However, the way that our text component handles ops that include inserting a blank embed has a bug that incorrectly interprets the delta from that op. If only one user is editing a doc while offline, there is no issue. But when two users edit the same doc while offline, the merge event results in duplicated verses.

To prevent this issue from happening, I decided to skip the code that inserts blank embeds when text is deleted from a verse. The negative impact of this fix is that blanks do not get automatically inserted if a user removes all text from a verse, but I think this will have a minimal impact because:

- I don't expect users to be intentionally removing all text from a verse in large quantities
- The blank will automatically be restored when returning online

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2148)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -217,8 +217,9 @@ export class TextViewModel {
    *
    * @param {DeltaStatic} delta The view model delta.
    * @param {Sources} source The source of the change.
+   * @param {boolean} isOnline Whether the user is online.
    */
-  update(delta: DeltaStatic, source: Sources): void {
+  update(delta: DeltaStatic, source: Sources, isOnline: boolean): void {
     const editor = this.checkEditor();
     if (this.textDoc == null) {
       return;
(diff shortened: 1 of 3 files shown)


pmachapman (approved):
:lgtm: This worked well in my tests.

Is this related to the `verse: true` and `note: true` ops noticed on production 14 days ago?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2148#-:-NhxEPcE-zvuNnzDlSRu:by7k2sm)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
It is possible it is related, but the symptom that I have seen is that verses and embeds are duplicated without being modified. It is very possible that this could partially fix the other issue, but for now there is no clear link.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2148#-:-Ni0NTQA5ni8HFH5P1h6:b-caa4gp)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2157, opened 2023-11-02
SF-2215 Reorganized Community Checking

josephmyers:
Removed global padding from app component and removed negative margins where possible. Avoiding those, even if it means each page needs to specify its own padding.

Made questions list an icon, instead of text. When shown, the question nav bar will now be visible. It also slides up now, instead of right. Introduced question toggle button to indicate when the question is shown, and added ability to hide the question when it's clicked. Moved the edit/archive buttons next to it. Moved the add question button to this bar, as well.

These changes better colocate the question operations, as well as reducing space consumption, horizontal and vertical. The result should be a more streamlined experience.

Made answer actions into icons. Much more space friendly, and it doesn't look awkward anymore on mobile.

Reorganized chapter audio, as well, though the play button situation isn't great. I believe we can hide the verse label and reinstate the play button in its original spot, then hide the play button when this player is shown.

A couple small bugs still. Pane autosize is resizing a little small. Toggle nav button selections are simple circles, instead of more professional highlighting.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2157)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
@@ -49,9 +49,11 @@ $toolbar-height: 56px;
 
     > div {
       flex: 1;
-      padding: 14px 20px;
+      padding-block-start: 14px;
+      padding-inline: 20px;
       @include media-breakpoint-only(xs, variables.$grid-breakpoints) {
-        padding: 10px;
+        padding-block-start: 0px;
+        padding-inline: 10px;
       }
       display: flex;
       flex-direction: column;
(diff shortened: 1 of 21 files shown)


Nateowami:
I appreciate your work to try to improve this; however, it somewhat conflicts with (as of yet ill-defined) plans we have for optimizing this page. It needs a much bigger overhaul to work well on mobile, and it needs to be an overhaul that starts with a blank canvas.

The biggest UX problems we have on this page are most apparent when working on an actual mobile device (not dev tools simulating a mobile device). And the ones that surface there are not nearly as easy to solve.

If you haven't tried community checking on a mobile phone, give it a try (preferably on a smaller phone), and slow down and pretend to be a user discovering it for the first time. 

Also, keep in mind that a lot of the actions on this page are for admins, but we don't expect admins to be working on mobile that much. It's nice if it works for admins, but we need to optimize the page with the community checkers as our sole target user.

josephmyers:
@Nateowami Thanks for your input. My goal was to help things while we wait. I think there are a number of solid tweaks we can integrate from here, regardless of us starting over, whenever that may be.

PR #2158, opened 2023-11-02
SF-2311 Allow playing chapter audio immediately after upload

RaymondLuong3:
The text audio query does not update automatically when audio is uploaded. Local update needs to be called on the query so that the locally uploaded audio is added to the query. I would have expected this to be automatic, but in other PRs such as SF-993 it appears to be the responsibility of the component to update the query.
I spend a couple hours trying to add a test for this behaviour, but didn't get far and decided it was not worth the effort.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2158)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
@@ -2504,6 +2504,7 @@ class TestEnvironment {
 
     const query = mock(RealtimeQuery<TextAudioDoc>) as RealtimeQuery<TextAudioDoc>;
     when(query.remoteChanges$).thenReturn(new BehaviorSubject<void>(undefined));
+    when(query.localChanges$).thenReturn(new BehaviorSubject<void>(undefined));
     const doc = mock(TextAudioDoc);
     const textAudio = mock<TextAudio>();
     when(textAudio.audioUrl).thenReturn('test-audio-short.webm');
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts
@@ -1057,6 +1057,7 @@ export class CheckingComponent extends DataLoadingComponent implements OnInit, A
     };
     await this.chapterAudioDialogService.openDialog(dialogConfig);
     this.updateAudioMissingWarning();
+    this.textAudioQuery?.localUpdate();
     this.calculateScriptureSliderPosition();
   }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts
@@ -639,11 +639,14 @@ export class CheckingComponent extends DataLoadingComponent implements OnInit, A
             // TODO (scripture audio) Only fetch the timing data for the currently active chapter
             this.projectService.queryAudioText(routeProjectId).then(query => {
               this.textAudioQuery = query;
-              this.audioChangedSub = this.textAudioQuery.remoteChanges$.subscribe(() => {
-                if (this.chapterAudioSource === '') {
-                  this.hideChapterAudio();
+              this.audioChangedSub = this.subscribe(
+                merge(this.textAudioQuery.remoteChanges$, this.textAudioQuery.localChanges$),
+                () => {
+                  if (this.chapterAudioSource === '') {
+                    this.hideChapterAudio();
+                  }
                 }
-              });
+              );
             });
 
             // TODO: check for remote changes to file data more generically
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts
@@ -140,6 +140,7 @@ export class CheckingComponent extends DataLoadingComponent implements OnInit, A
   private projectDeleteSub?: Subscription;
   private hideTextSub?: Subscription;
   private textAudioSub?: Subscription;
+  private audioChangedSub?: Subscription;
   private projectRemoteChangesSub?: Subscription;
   private questionFilterFunctions: Record<QuestionFilter, (answers: Answer[]) => boolean> = {
     [QuestionFilter.None]: () => true,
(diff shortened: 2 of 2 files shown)


Nateowami (changes requested):
In my testing, this does not actually solve the issue at hand.

And looking at the code, I don't see how it would. The problem isn't that the component doesn't realize there's audio to play; it does realize it and makes the player available. But then it doesn't actually play. I think something should be watching the project document and realizing that `Chapter.hasAudio` has been set to true, and then fetches the audio timing data.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2158#-:-NiH9KjmCBKh9roxHQqH:br96vik)* status: 0 of 1 files reviewed, all discussions resolved

RaymondLuong3 (commented):
Did this change not fix the issue for you? Maybe there are actually 2 issues.

In my testing, it was definitely the textAudioQuery that was not being updated, and the result is that the scripture audio player appears without a valid source. The source actually gets set to the empty string. We know that the `text.chapter.hasAudio` is true because the play button is visible and allows opening the scripture player.
What you are saying contradicts my observations.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2158#-:-NiKrCxS6S-BbvRmo1Gf:bqzs4e9)* status: 0 of 1 files reviewed, all discussions resolved

Nateowami:
Did you watch the video I attached to the issue? The behavior I see from this PR is identical to what that video shows.

Is the problem you see being fixed different?

Nateowami:
I just tested this again to make sure I didn't make some mistake testing it last time. This time around I saw it work correctly twice, but every other time it failed to play the audio.

RaymondLuong3 (commented):
Ok, sounds like there might be some sort of race condition happening. I'll dig around a little more.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2158#-:-NiLCLfU5QkAX3A3-Nro:b3licrg)* status: 0 of 1 files reviewed, all discussions resolved

RaymondLuong3 (commented):
You were right, the reason the query was not being updated was because the query for the text audio was not in the same place as the query for the questions. The query for the questions called `cleanup()` which disposed the text audio query. Now the `cheanup()` method is called before the new question and text audio queries are created.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2158#-:-NifzfNL2_ovyYbO-df4:b-x1b88f)* status: 0 of 1 files reviewed, all discussions resolved

Nateowami:
Thanks. I don't have time to review it at the moment, but it appears to be working now.

nigel-wells (commented):
I couldn't reproduce the issue on `master`. Although re-subscribing to the query works it seems like we shouldn't have to unless the project changes and this query would update anyway. Perhaps its that we're only listening to remote changes and subscribing to `localChanges$` would resolve the issue?
```
this.subscribe(merge(this.textAudioQuery.remoteChanges$, this.textAudioQuery.remoteChanges$),() => {...});
```

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2158#-:-NigwMhsANlZ0bZTbgTW:bvh5zbt)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
Because there is a race-condition in this issue, I have only been able to reproduce it by watching the video that Nathaniel has posted. The number of questions in the project has an effect on reproducing this issue.
Listening to the `localChanges$` on the query is a good idea, however the observer does not actually have an effect on whether the audio is available to play or not, just if the player is visible. I've added it anyway.

There is a note to update the query to only get the text audio for the current chapter. When we do implement that, the audio query will get disposed whenever the router params change. That is why the query needs to be recreated just like the question query.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2158#-:-Nik1yWR2Iclzk51tlgY:bns1mhb)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @nigel-wells)

Nateowami:
> The number of questions in the project has an effect on reproducing this issue.

I see it happen even when I have only two questions in the project, and it happens nearly every single time.

nigel-wells (commented):
Does this still resolve the issue? I suspect the issue is the realtime query isn't `ready` yet when we try to set the value in the audio player. The same could be said of `chapterTextAudioTiming` which also tries to find the audio data.

How about we create a method called `async showScriptureAudioPlayer()` which sets the value to true only after we're confident the realtime query is ready to use. You could add something like to make sure it is ready:
```
await this.textAudioQuery?.ready$  
  .pipe(  
    filter(ready => ready),  
    first()  
  )  
  .toPromise();
```
I still can't replicate the race-condition so I can't confirm anything from my end sorry.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2158#-:-NilfR862LR3WGLE6bWW:b-r3nzl9)* status: 0 of 2 files reviewed, all discussions resolved

nigel-wells (commented):
Thinking about some more - due to some other similar Jira cards - we probably need to do more than just hide the player if there is a change. If another user is waiting for audio to appear they'll never see the player as the we never update the source after the RT query changes. I'm working on SF-2323 so I might try a few things in there.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2158#-:-Niloe6yBA3_1yGkNqYL:b-34uoll)* status: 0 of 2 files reviewed, all discussions resolved

josephmyers:
@RaymondLuong3 @nigel-wells @Nateowami This is one of the very last SA MVP tasks we have left. What can we do to move this guy along?

Nateowami:
@josephmyers If you're able to review it that would be great.

josephmyers:
I had grabbed some other work and am just now seeing this. If no one's touched it by the time I get to work tomorrow, I'll give this a review 👍

josephmyers (approved):
Code looks normal to me. If y'all have tested this, let's pass it on to test

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2158#-:-NkIg9vV4syjzDGDJVB9:bhg5wwu)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

Nateowami (commented):
I just looked over the changes, and the way this has been implemented flies directly in the face of how we designed the model. `Chapter.hasAudio` was created *specifically* to avoid having to create a realtime query for audio timing data, which is much more expensive than watching a property. 

Maybe we should merge anyway, but this can be seriously optimized.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2158#-:-NkHelig7FmEZ4EykEGC:b5ipuho)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
Hmm, I know that we have the project doc, which we are listening to for when `projectDoc.text[x].chapter.hasAudio` gets updated. Then if it does, we would need to then retrieve the textAudio from the realtime server. That does sound less expensive than listening to all of the text audio for the project. I believe that we have a TODO to make the query to be specific to the book and chapter. Then that problem would be fixed too.
Let's get this tested, and then I'll create an issue to remove the textAudio query.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2158#-:-NkJAlFhEvTJ38CwPtkM:b-y92dh2)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2159, opened 2023-11-03
SF-2317 Improve comment FAB position to not cover text

RaymondLuong3:
This change adds padding to the quill editor on the side where the FAB for adding comments is showing. This makes it much less likely for the FAB to cover text as a user is typing or reading.

Translators and Admins
FAB Translator Wide Screen

Commenters
FAB Commenter Wide Screen

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2159)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
@@ -97,6 +97,21 @@ quill-editor {
   }
 }
 
+.comment-enabled-editor {
+  quill-editor:not(.read-only-editor) .ql-editor {
+    padding-inline-start: 10px;
+    padding-inline-end: 40px;
+  }
+
+  // only commenter users on devices with screens larger than large need to account for the add comment FAB
+  quill-editor.read-only-editor .ql-editor {
+    @include media-breakpoint-up(lg) {
+      padding-inline-start: 10px;
+      padding-inline-end: 40px;
+    }
+  }
+}
+
 quill-editor.ltr {
   .question-segment[data-question-count] {
     margin-left: 1.75em;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
@@ -160,6 +160,7 @@
                   [isRightToLeft]="isTargetRightToLeft"
                   [fontSize]="fontSize"
                   [selectableVerses]="showAddCommentUI"
+                  [ngClass]="{ 'comment-enabled-editor': showAddCommentUI }"
                 ></app-text>
                 <app-suggestions
                   class="mat-elevation-z2"
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
@@ -197,7 +197,7 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
   private resizeObserver?: ResizeObserver;
   private scrollSubscription?: Subscription;
   private readonly fabDiameter = 40;
-  private readonly fabHorizMargin = 20;
+  private readonly fabHorizMargin = 15;
 
   constructor(
     private readonly activatedRoute: ActivatedRoute,
@@ -585,7 +585,6 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
   ngAfterViewInit(): void {
     this.subscribe(fromEvent(window, 'resize'), () => {
       this.setTextHeight();
-      this.resetInsertNoteFab(false);
       this.positionInsertNoteFab();
     });
     this.subscribe(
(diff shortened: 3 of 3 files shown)


marksvc (changes requested):
> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2159)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 200 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2159#-NiakM3c7ilpCU20TQTu:-NiakM3c7ilpCU20TQTv:b-21pl5) ([raw file](https://github.com/sillsdev/web-xforge/blob/416df097139dfbcb00efed31c41e1741d85a88e9/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L200)):*
> ```XML
>   private scrollSubscription?: Subscription;
>   private readonly fabDiameter = 40;
>   private readonly fabHorizMargin = 15;
> ```

When viewing as a Commenter, the FAB feels a bit too close to the scrollbar for me. 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 572 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2159#-NiajmvU-imDCtbUlGRX:-NiajmvU-imDCtbUlGRY:b4t8ppw) ([raw file](https://github.com/sillsdev/web-xforge/blob/c081a9b9a34325de6d35dbaddca168a38b9a89b5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L572)):*
> ```XML
>       this.setTextHeight();
>       this.resetInsertNoteFab(false);
> ```

Without this and the change below, I have _similar_, but slightly poorer behaviour. Maybe we can chat on the phone and you can explain what is happening here.

marksvc:
Thank you for your work on this. I am pleased that this fixes the usability problem at hand. But I am not sure that this should be our long-term behaviour. Because it loses a whole column of space down the right side of the text. And may be a bit odd, visually.

Possible alternatives:

- The add note icon could be moved _up above the verse being edited_. This puts it in the way of the _context_, obscuring text of the previous verse, but it's at least not obscuring the text of the word that an editor is trying to type. See screenshot illustrating:
image

- The FAB could instead be put _below_ the current verse. That still obscures context, but only for situations where the user is editing a verse that was already written. In cases where the user is drafting verses, and they haven't written the next verse yet, it would not be obscuring any text (such as text in the verse at hand, or text in the prior verse). Moving the FAB down in this way may be better than moving the FAB up, as described in the preceding bullet.

- When I move between different verses, the FAB moves up and down to be next to my current verse. And it's has a green circle, making it stand out as one of the few most colourful items on the page. I'm not sure an editor needs as much visibility highlighting the action of adding a Note, with the green circle following their insertion point. I assume this was all decided with good reasons in UI/UX design previously. But I will mention that not having a FAB, and providing a toolbar button, is of course a possible solution to the problem at hand.


Interestingly, for a commenter (not editor), if I start to shrink the horizontal size of the window, and the view port changes, the FAB can start to cover the text. I see that if I click a verse then the FAB disappears and there is a control now on the bottom of the page that lets me add a comment (not shown in the video). Though I don't think this needs thought about or anything done with for this PR. Video showing the behaviour I describe:

[add-comment-fab.webm](https://github.com/sillsdev/web-xforge/assets/7265309/dd2451ae-a230-4f9b-8174-877da3d83aa2)

RaymondLuong3 (commented):
Thanks for the comments. I had suggested multiple ways to fix this, and Nathaniel and I agreed this was a quickest solution without causing any further issues. The number of pixels we lose by adding padding on the right side is 20 pixels or less, so it is more that it looks strange with the extra padding on the right that we aren't thrilled about. I'll bring this up in tonight's meeting, and maybe we can come to a consensus as to the solution we are most happy with.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2159#-:-NianHdi7KXUgnx4xgEv:bxuyxmv)* status: 2 of 3 files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 572 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2159#-NiajmvU-imDCtbUlGRX:-NiaoHKjBrej6zbOmlsz:b-dm3oj3) ([raw file](https://github.com/sillsdev/web-xforge/blob/c081a9b9a34325de6d35dbaddca168a38b9a89b5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L572)):*

Without this and the change below, I have _similar_, but slightly poorer behaviour. Maybe we can chat on the phone and you can explain what is happening here.
</blockquote></details>

For a brief moment when the page is loading, there are multiple scrollbars on the page. This has to do with the biblical terms loading. When the page is loaded, then the scrollbars disappears but the left style of the FAB has been pushed to accommodate the extra scroll bars. This change puts the horizontal and vertical calculation of the FAB all in 1 place.

marksvc (approved):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2159)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2183, opened 2023-11-22
SF-2345 Fix MatSlider in audio player in RTL mode

Nateowami:
The issue this is fixing is the jumpyness when seeking in RTL mode. The problem is caused by the Material slider which is supposed to support RTL mode, but since this is an audio player, we want it to stay in LTR mode. But it refuses to cooperate, which led to workarounds that don't fully work.

I updated us to the MDC-based Material slider, removed the workarounds, set the direction to LTR, and things now work fine.

In addition, I fixed two other issues:
1. Previously, as you seek, it would immediately jump to that point in the audio, meaning you hear choppiness while seeking. Instead, we now listen to the dragEnd event, which is how I think most decent audio players work.
2. Previously if you dragged towards the end of the audio, and overshot and brought the slider to the end of the audio, it would immediately close the player, since you had "reached the end." This is not intuitive, and clearly not what the user was expecting (they're clearly trying to seek, not close the player). This is also fixed by listening for the dragEnd event instead.

https://github.com/sillsdev/web-xforge/assets/6140710/8eb6fd81-56f2-46b1-ae72-0c6a64aec878

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.html
@@ -1,16 +1,9 @@
 <ng-container *transloco="let t; read: 'checking_audio_player'">
   <div *ngIf="!hasProblem" class="content" dir="ltr">
     <div class="current-time">{{ audio?.currentTime ?? 0 | audioTime }}</div>
-    <mat-slider
-      class="slider"
-      [disabled]="!isAudioAvailable"
-      [min]="0"
-      [max]="100"
-      [step]="1"
-      [value]="seek"
-      (input)="onSeek($event)"
-      [dir]="i18n.direction"
-    ></mat-slider>
+    <mat-slider class="slider" [disabled]="!isAudioAvailable" [min]="0" [max]="100">
+      <input matSliderThumb [value]="seek" (dragEnd)="onSeek($event)" />
+    </mat-slider>
     <div class="duration">{{ duration | audioTime }}</div>
   </div>
   <div *ngIf="hasProblem" class="audio-not-available">
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio/audio-player/audio-player.component.scss
@@ -8,17 +8,7 @@
     font-size: 0.8em;
   }
   .slider {
-    flex: 1;
-    width: 100%;
-    ::ng-deep .mat-slider-track-background {
-      background-color: #e0e0e0;
-    }
-    ::ng-deep .mat-slider-thumb {
-      background-color: #000;
-    }
-    &[dir='rtl'] {
-      transform: rotateY(180deg);
-    }
+    flex-grow: 1;
   }
 }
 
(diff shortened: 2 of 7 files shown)


josephmyers (changes requested):
Woot! Great job removing old workarounds

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2183#-:-NkDeuz75YrZ1dWlwzCC:b-h8z8h8)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions-settings-dialog.component.html` line 37 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2183#-NkDeIh_-s94BOfhrBKB:-NkDeIh_-s94BOfhrBKC:b-d7bk3b) ([raw file](https://github.com/sillsdev/web-xforge/blob/a933588e7384292568bf62a5a9b1ebe316df2e23/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions-settings-dialog.component.html#L37)):*
> ```HTML
>         </div>
>         <mat-slider id="confidence-threshold-slider" [disabled]="translationSuggestionsDisabled" [min]="0" [max]="100">
>           <input matSliderThumb [(ngModel)]="confidenceThreshold" (dragEnd)="confidenceThreshold = $event.value" />
> ```

ngModel is deprecated. We should update it while we're in here

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2183)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions-settings-dialog.component.html` line 37 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2183#-NkDeIh_-s94BOfhrBKB:-NkDgXZc1KbDDnx5a7gK:b-arao90) ([raw file](https://github.com/sillsdev/web-xforge/blob/a933588e7384292568bf62a5a9b1ebe316df2e23/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions-settings-dialog.component.html#L37)):*

ngModel is deprecated. We should update it while we're in here
</blockquote></details>

Can you provide a source? I see that a particular use of it was deprecated in Angular 6 to be removed in Angular 7, but we're well past that. If it's deprecated, I'd think it would be clearly stated, and we would get the notice in the editor.

josephmyers (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2183)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions-settings-dialog.component.html` line 37 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2183#-NkDeIh_-s94BOfhrBKB:-NkDjsPK0aVwzhG8wGpU:b-tpcf1j) ([raw file](https://github.com/sillsdev/web-xforge/blob/a933588e7384292568bf62a5a9b1ebe316df2e23/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions-settings-dialog.component.html#L37)):*

Can you provide a source? I see that a particular use of it was deprecated in Angular 6 to be removed in Angular 7, but we're well past that. If it's deprecated, I'd think it would be clearly stated, and we would get the notice in the editor.
</blockquote></details>

That was my source. I see now that it's only deprecated for a particular usage with reactive forms. Nevermind! Thanks

PR #2188, opened 2023-11-27
SF-2318 Right align suggestions when there's overflow

josephmyers:
Simplified the overflow repositioning logic a little. Before, it was basing its calculations off of the entire body, which resulted in some incorrect assumptions, namely that that logic was not accounting for margins. The old code only resized when the overflow was greater than the body width, but there could have been and was nonetheless cutoff due to the righthand margin between the editor and the right side of the body.

The updated logic bases its computations off of the editor alone, resulting in a cleaner and more accurate solution.

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2188)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions.component.scss
@@ -48,12 +48,12 @@ mat-selection-list[dense] {
   padding: 0;
 
   > mat-list-option {
-    height: 32px;
+    height: auto;
   }
 }
 
 .mat-list-option ::ng-deep .mat-list-item-content {
-  padding: 0 8px !important;
+  padding: 6px 8px !important;
   .mat-list-text {
     padding: 0 !important;
   }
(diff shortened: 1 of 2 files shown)


nigel-wells (approved):
This works really well.
I tested out @Nateowami RTL points and it appears to work exactly the same as it was before where the suggestions have always been positioned from the bottom right of the segment. Did you want to try and address that in this or create a separate card for it so that we can get this fix in sooner as a user is running into the issue.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2188#-:-NkHxiN-7R6Zge8MTbx_:b2qv9nf)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (commented):
Agreed. Let's address the issue at hand and follow back up for a potential effort to account for RTL, which was not being done before. Assuming the current design needs to change, it will be nontrivial, as it's currently calculating all positioning by hand (for LTR).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2188#-:-NkJR7nH5FrsFs6nPJHD:bmv86lv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

PR #2199, opened 2023-11-30
SF-2381 Support phrase level audio scripture playback on questions

RaymondLuong3:
Timing files can be broken down into phrase level (and even word level) components. The Single button audio player accepts the timing for the section of the audio for the verse that a question pertains to. Previously, the component that extracted the data for the timing range did not account for the phrase level timings, meaning that for a single verse, there may be multiple timings.
This change ensures that all the timings for a timing range is considered so that single button audio player is given the correct timing information for the verse.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2199)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-test.utils.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-test.utils.ts
@@ -21,6 +21,15 @@ export function getAudioTimingWithHeadings(): AudioTiming[] {
   ];
 }
 
+export function getAudioTimingsPhraseLevel(): AudioTiming[] {
+  return [
+    { textRef: '1a', from: 0.0, to: 1.0 },
+    { textRef: '1b', from: 1.0, to: 2.0 },
+    { textRef: '2a', from: 2.0, to: 3.0 },
+    { textRef: '2b', from: 3.0, to: 4.0 }
+  ];
+}
+
 export class AudioPlayerStub extends AudioPlayer {
   htmlAudioMock: HTMLAudioElement = mock(HTMLAudioElement);
   protected override audio: HTMLAudioElement = instance(this.htmlAudioMock);
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.spec.ts
@@ -18,8 +18,9 @@ import { toVerseRef } from 'realtime-server/lib/esm/scriptureforge/models/verse-
 import { QuestionDoc } from '../../../../core/models/question-doc';
 import { TextAudioDoc } from '../../../../core/models/text-audio-doc';
 import { SFProjectService } from '../../../../core/sf-project.service';
-import { SingleButtonAudioPlayerComponent } from '../../single-button-audio-player/single-button-audio-player.component';
 import { SFProjectUserConfigDoc } from '../../../../core/models/sf-project-user-config-doc';
+import { getAudioTimingsPhraseLevel } from '../../../checking-test.utils';
+import { SingleButtonAudioPlayerComponent } from '../../single-button-audio-player/single-button-audio-player.component';
 import { CheckingQuestionComponent } from './checking-question.component';
 
 const mockedSFProjectService = mock(SFProjectService);
@@ -47,8 +48,8 @@ class MockComponent {
     when(mockedQuestion.audioUrl).thenReturn('test-audio-player.webm');
     const verseRef: VerseRefData = {
       bookNum: 8,
-      chapterNum: 22,
-      verseNum: 17
+      chapterNum: 1,
+      verseNum: 1
     };
     when(mockedQuestion.verseRef).thenReturn(verseRef);
     when(mockedQuestionDoc.data).thenReturn(instance(mockedQuestion));
@@ -93,7 +94,7 @@ describe('CheckingQuestionComponent', () => {
 
   it('selects question when scripture audio has already been played', async () => {
     const env = new TestEnvironment();
-    when(mockedSFProjectUserConfig.audioRefsPlayed).thenReturn(['RUT 22:17']);
+    when(mockedSFProjectUserConfig.audioRefsPlayed).thenReturn(['RUT 1:1']);
     await env.wait();
     await env.wait();
 
@@ -104,12 +105,12 @@ describe('CheckingQuestionComponent', () => {
 
   it('selects scripture if not all scripture audio has already been played for question', async () => {
     const env = new TestEnvironment();
-    when(mockedSFProjectUserConfig.audioRefsPlayed).thenReturn(['RUT 22:17']);
+    when(mockedSFProjectUserConfig.audioRefsPlayed).thenReturn(['RUT 1:1']);
     const verseRef: VerseRefData = {
       bookNum: 8,
-      chapterNum: 22,
-      verseNum: 17,
-      verse: '17-18'
+      chapterNum: 1,
+      verseNum: 1,
+      verse: '1-2'
     };
     when(mockedQuestion.verseRef).thenReturn(verseRef);
     await env.wait();
@@ -160,6 +161,19 @@ describe('CheckingQuestionComponent', () => {
 
     await env.wait(1000); //wait for the audio to finish playing
     expect(env.component.question.focusedText).toBe('scripture-audio-label');
+  });
+
+  it('plays the entire verse when timing files are phrase level', async () => {
+    const env = new TestEnvironment();
+    await env.wait();
+    await env.wait();
+
+    env.component.question.playScripture();
+    await env.wait(1200);
+
+    expect(env.component.question.scriptureAudioStart).toBe(0);
+    expect(env.component.question.scriptureAudioEnd).toBe(2);
+    expect(env.component.question.focusedText).toBe('question-audio-label');
   });
 
   it('does not select question if scriptureAudio is set to the same value', async () => {
@@ -269,7 +283,7 @@ class TestEnvironment {
   readonly queryChanged$: Subject<void> = new Subject<void>();
 
   constructor() {
-    const audio1 = this.createTextAudioDoc(getTextAudioId('project01', 8, 22), 'test-audio-player-b.webm');
+    const audio1 = this.createTextAudioDoc(getTextAudioId('project01', 8, 1), 'test-audio-player-b.webm');
     const audio2 = this.createTextAudioDoc(getTextAudioId('project01', 1, 11), 'test-audio-player.webm');
 
     when(this.query.remoteChanges$).thenReturn(this.queryChanged$);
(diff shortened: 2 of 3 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2199)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.ts` line 95 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2199#-NkaSmHa124zJFHmr63x:-NkaSmHa124zJFHmr63y:bi4mxsm) ([raw file](https://github.com/sillsdev/web-xforge/blob/dce33dbeb7e2bf3e52b030a6fd8c42d4f4b0af80/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.ts#L95)):*
> ```XML
>     );
>     if (verseTimings == null || verseTimings.length === 0) return;
>     return verseTimings.sort((a, b) => b.to - a.to)[0].to;
> ```

Are we expecting the timings to be out of order? I think that would be invalid timing data. Why can't we just take the last timing for the verse? For example:
``` ts
return verseTimings[verseTimings.length - 1].to;
```
Tests continue to pass after this change.

josephmyers (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2199)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.spec.ts` line 166 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2199#-Nkn0ziv9qH6cQT1QmdN:-Nkn0ziv9qH6cQT1QmdO:bc38suo) ([raw file](https://github.com/sillsdev/web-xforge/blob/dce33dbeb7e2bf3e52b030a6fd8c42d4f4b0af80/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.spec.ts#L166)):*
> ```XML
>   });
> 
>   it('plays the entire verse when timing files are phrase level', async () => {
> ```

This is basically an integration test. It's not this class's responsibility to verify the entire verse plays, just that it passes the correct start and end to the audio player. We don't have to delete it, but it's a little out of place, and it has a costly 1.2s wait time.

I recommend writing a unit test to verify that scriptureAudioStart and scriptureAudioEnd are correct with phrase-style timing data.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2199)* status: all files reviewed, 2 unresolved discussions (waiting on @josephmyers and @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.ts` line 95 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2199#-NkaSmHa124zJFHmr63x:-NkqVEolAl1GZOm_Zyu8:b-yky4od) ([raw file](https://github.com/sillsdev/web-xforge/blob/dce33dbeb7e2bf3e52b030a6fd8c42d4f4b0af80/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.ts#L95)):*

Are we expecting the timings to be out of order? I think that would be invalid timing data. Why can't we just take the last timing for the verse? For example:
``` ts
return verseTimings[verseTimings.length - 1].to;
```
Tests continue to pass after this change.
</blockquote></details>

Some timing formats have all the headings grouped in one section of the file, and the verse timings grouped in another. So we should not make the assumption that timings are in order.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.spec.ts` line 166 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2199#-Nkn0ziv9qH6cQT1QmdN:-NkqU04s8ZhOEZd2e3Hj:bza8odk) ([raw file](https://github.com/sillsdev/web-xforge/blob/dce33dbeb7e2bf3e52b030a6fd8c42d4f4b0af80/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.spec.ts#L166)):*

This is basically an integration test. It's not this class's responsibility to verify the entire verse plays, just that it passes the correct start and end to the audio player. We don't have to delete it, but it's a little out of place, and it has a costly 1.2s wait time.

I recommend writing a unit test to verify that scriptureAudioStart and scriptureAudioEnd are correct with phrase-style timing data.
</blockquote></details>

That is basically what this does. `env.component.question.scriptureAudioStart` is the propert on the `CheckingQuestionComponent` that is provided to the `SingleButtonAudioPlayerComponent`. If the 1.2s wait seems too costly, it is there to check that the audio is more than 1s long.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2199)* status: all files reviewed, 2 unresolved discussions (waiting on @josephmyers and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.ts` line 95 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2199#-NkaSmHa124zJFHmr63x:-NkrDpC039dRnK_L2dol:b-9gp9hh) ([raw file](https://github.com/sillsdev/web-xforge/blob/dce33dbeb7e2bf3e52b030a6fd8c42d4f4b0af80/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.ts#L95)):*

Some timing formats have all the headings grouped in one section of the file, and the verse timings grouped in another. So we should not make the assumption that timings are in order.
</blockquote></details>

That is extremely surprising to me. Good to know. I would have assumed timing data formats would require that the timing data be chronological. Is this documented anywhere in a comment? I think it may belong here.

In `get scriptureAudioStart()`, there is a built-in assumption that the first matching timing will be the first chronologically, I think?

Also, what do you think about changing this last line to `Math.max(...verseTimings.map(t => t.to))`? We can be fairly certain that the list will always be short enough that sorting and taking the first item will be extremely fast, but it seems odd to do so when there's a more direct solution that doesn't require sorting.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.spec.ts` line 166 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2199#-Nkn0ziv9qH6cQT1QmdN:-NkrmxQA8dmACDewMFpt:b47a7dp) ([raw file](https://github.com/sillsdev/web-xforge/blob/dce33dbeb7e2bf3e52b030a6fd8c42d4f4b0af80/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.spec.ts#L166)):*

That is basically what this does. `env.component.question.scriptureAudioStart` is the propert on the `CheckingQuestionComponent` that is provided to the `SingleButtonAudioPlayerComponent`. If the 1.2s wait seems too costly, it is there to check that the audio is more than 1s long.
</blockquote></details>

I don't think we actually have to play to audio to test that the component correctly figures out when the audio should start and stop. As long as it's figuring it out correctly and sending that information to the child component, we should be fine.

Joseph's comment about 1.2 seconds being costly may seem extreme, but if each of our 1197 Jasmine tests took 1.2 seconds, the entire suite would take 24 minutes. On my machine the tests take about 70 seconds to run, and that's about 58 ms each on average. Being able to get near instant feedback is critical to being able to use the tests as part of the development process.

josephmyers (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2199)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.spec.ts` line 166 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2199#-Nkn0ziv9qH6cQT1QmdN:-NksWFcE9PNkhcsBrmCw:begqjr5) ([raw file](https://github.com/sillsdev/web-xforge/blob/dce33dbeb7e2bf3e52b030a6fd8c42d4f4b0af80/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.spec.ts#L166)):*

I don't think we actually have to play to audio to test that the component correctly figures out when the audio should start and stop. As long as it's figuring it out correctly and sending that information to the child component, we should be fine.

Joseph's comment about 1.2 seconds being costly may seem extreme, but if each of our 1197 Jasmine tests took 1.2 seconds, the entire suite would take 24 minutes. On my machine the tests take about 70 seconds to run, and that's about 58 ms each on average. Being able to get near instant feedback is critical to being able to use the tests as part of the development process.
</blockquote></details>

Thanks, Nathaniel, you said it better. Maybe it's as simple as removing the playScripture() and wait(1200). They don't have an impact on the start and end properties, and we have coverage for the focusedText label enough that we can likely get by without it in this test.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2199)* status: all files reviewed, 2 unresolved discussions (waiting on @josephmyers and @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.ts` line 95 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2199#-NkaSmHa124zJFHmr63x:-NkvM3O9E-B1MJJX1sE3:b-huxwo0) ([raw file](https://github.com/sillsdev/web-xforge/blob/dce33dbeb7e2bf3e52b030a6fd8c42d4f4b0af80/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.ts#L95)):*

That is extremely surprising to me. Good to know. I would have assumed timing data formats would require that the timing data be chronological. Is this documented anywhere in a comment? I think it may belong here.

In `get scriptureAudioStart()`, there is a built-in assumption that the first matching timing will be the first chronologically, I think?

Also, what do you think about changing this last line to `Math.max(...verseTimings.map(t => t.to))`? We can be fairly certain that the list will always be short enough that sorting and taking the first item will be extremely fast, but it seems odd to do so when there's a more direct solution that doesn't require sorting.
</blockquote></details>

Thanks for spotting that. I've added a comment and updated both getters to return the start and end time using the Math function. Why didn't I think of that?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.spec.ts` line 166 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2199#-Nkn0ziv9qH6cQT1QmdN:-NkvL1G5-VxOBEBxytYj:b6slcya) ([raw file](https://github.com/sillsdev/web-xforge/blob/dce33dbeb7e2bf3e52b030a6fd8c42d4f4b0af80/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.spec.ts#L166)):*

Thanks, Nathaniel, you said it better. Maybe it's as simple as removing the playScripture() and wait(1200). They don't have an impact on the start and end properties, and we have coverage for the focusedText label enough that we can likely get by without it in this test.
</blockquote></details>

Yes, it does not do a whole lot other than it shows that at 1.2 seconds in, the audio is still playing and the scripture focus text is still active. But the test coverage already in the document is sufficient.

PR #2182, opened 2023-11-22
SF-2314 Allow opening roles & permissions dialog offline

josephmyers:
The dialog can be opened while offline, but the fields are still not editable. I added some text to indicate this in the dialog.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2182)

PR #2200, opened 2023-12-01
SF-2383 Adjust and add styling options to notice component

siltomato:
Updated notice component to have softer styles and more styling options as mentioned in [this github comment](https://github.com/sillsdev/web-xforge/pull/2182#pullrequestreview-1751463413).

`<app-notice type="normal">` will still work, but the new way is to use 'primary' `<app-notice type="primary">`.
`<app-notice [outline]="true">` will still work, but the new way is to use the `mode` input `<app-notice mode="outline">`.

- Updated Storybook with the new options.
- Converted NoticeComponent to be a standalone component.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2200)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.html
@@ -61,7 +61,7 @@
         >
       </div>
     </div>
-    <app-notice id="offline-error" *ngIf="!isOnline" [type]="'error'" [outline]="true" [icon]="'cloud_off'">
+    <app-notice id="offline-error" *ngIf="!isOnline" type="error" icon="cloud_off">
       {{ t("chapter_audio_dialog.not_upload_audio_offline") }}
     </app-notice>
   </mat-dialog-content>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.html
@@ -44,9 +44,8 @@
     <app-notice
       id="warning-some-actions-unavailable-offline"
       *ngIf="!isOnline && (canCreateScriptureAudio || canDeleteScriptureAudio)"
-      [icon]="'cloud_off'"
-      [type]="'warning'"
-      [outline]="true"
+      icon="cloud_off"
+      type="warning"
       >{{ t("some_actions_unavailable_offline") }}</app-notice
     >
     <ng-container *ngFor="let text of texts">
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.spec.ts
@@ -38,14 +38,15 @@ const mockedSFProjectService = mock(SFProjectService);
 
 describe('JoinComponent', () => {
   configureTestingModule(() => ({
-    declarations: [JoinComponent, NoticeComponent],
+    declarations: [JoinComponent],
     imports: [
       HttpClientTestingModule,
       NoopAnimationsModule,
       TestTranslocoModule,
       TestRealtimeModule.forRoot(SF_TYPE_REGISTRY),
       TestOnlineStatusModule.forRoot(),
-      UICommonModule
+      UICommonModule,
+      NoticeComponent
     ],
     providers: [
       { provide: ActivatedRoute, useMock: mockedActivatedRoute },
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.stories.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.stories.ts
@@ -2,7 +2,7 @@ import { CommonModule } from '@angular/common';
 import { HttpErrorResponse } from '@angular/common/http';
 import { MatLegacyDialogRef as MatDialogRef } from '@angular/material/legacy-dialog';
 import { ActivatedRoute, Router } from '@angular/router';
-import { Meta, moduleMetadata, StoryObj } from '@storybook/angular';
+import { Meta, StoryObj, moduleMetadata } from '@storybook/angular';
 import { userEvent, within } from '@storybook/testing-library';
 import { of } from 'rxjs';
 import { anything, instance, mock, when } from 'ts-mockito';
@@ -71,8 +71,8 @@ const meta: Meta = {
   },
   decorators: [
     moduleMetadata({
-      imports: [UICommonModule, CommonModule, I18nStoryModule],
-      declarations: [NoticeComponent, GenericDialogComponent],
+      imports: [UICommonModule, CommonModule, I18nStoryModule, NoticeComponent],
+      declarations: [GenericDialogComponent],
       providers: [
         { provide: ActivatedRoute, useValue: instance(mockedActivatedRoute) },
         { provide: AnonymousService, useValue: instance(mockedAnonymousService) },
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/info/info.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/info/info.component.ts
@@ -14,6 +14,6 @@ export class InfoComponent {
   constructor() {}
 
   get mirrorRTL(): boolean {
-    return ICONS_TO_MIRROR_RTL.includes(this.icon ?? '');
+    return ICONS_TO_MIRROR_RTL.has(this.icon);
   }
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.html
@@ -1,4 +1,2 @@
-<div [ngClass]="{ outline }" [class]="type">
-  <mat-icon *ngIf="icon" [class.mirror-rtl]="mirrorRTL">{{ icon }}</mat-icon>
-  <span class="notice-content"><ng-content></ng-content></span>
-</div>
+<mat-icon *ngIf="icon" [class.mirror-rtl]="mirrorRtl">{{ icon }}</mat-icon>
+<span class="notice-content"><ng-content></ng-content></span>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss
@@ -1,17 +1,23 @@
 @use 'sass:color';
 @use 'sass:list';
+@use 'src/variables' as sfColors;
 
 // prettier-ignore
 $colors: (
-  primary:   (#cfe2fd, #305993),
-  secondary: (#d6d4e8, #4a4959),
-  success:   (#c4ead9, #125235),
-  warning:   (#fff3d0, #6a5117),
-  error:     (#f8d7da, #952633),
-  info:      (#d0f4fb, #345564),
-  light:     (#f7f7f7, #6a6b6b),
-  dark:      (#d3d3d4, #181a1d)
+  primary:   (sfColors.$blueMedium, #cfe2fd),
+  secondary: (sfColors.$purpleDark, #d6d4e8),
+  success:   (#125235, #c4ead9),
+  warning:   (#6a5117, #fff3d0),
+  error:     (sfColors.$errorColor, #f8d7da),
+  info:      (#345564, #d0f4fb),
+  light:     (#6a6b6b, #f7f7f7),
+  dark:      (#181a1d, #d3d3d4),
 );
+
+// Helper function just to clean up the syntax
+@function scaleLightness($color, $lightness) {
+  @return color.scale($color, $lightness: $lightness);
+}
 
 :host {
   display: flex;
@@ -24,8 +30,8 @@ $colors: (
   color: var(--notice-color-text);
   background-color: var(--notice-color);
 
-  --notice-color-button-bg: var(--notice-color-outline);
   --notice-color-button-text: #fff;
+  --notice-color-button-hover-text: var(--notice-color-button-text);
 
   mat-icon {
     flex-shrink: 0;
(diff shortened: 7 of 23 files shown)


Nateowami (commented):
> Reviewed 2 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2200)* status: 2 of 12 files reviewed, 1 unresolved discussion (waiting on @nigel-wells and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts` line 277 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkauxMX6nMvVEz0SfAp:-NkarewX9C5aWd52hFZG:bvvywgl) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L277)):*
> ```XML
>  * than merely mirroring. This list is ONLY for those icons that can be mirrored in all contexts.
>  */
> export const ICONS_TO_MIRROR_RTL = new Set<string | undefined>([
> ```

I wish TypeScript was a tad smarter.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss` line 16 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-Nkav3GwE3hrtkKYfKOf:-Nkav3GwE3hrtkKYfKOg:b-k8lcg1) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss#L16)):*
> ```SCSS
> );
> 
> :host {
> ```

I'm not sure whether this is a good idea. I've gone from always creating wrapper elements around components for styling reasons, to thinking "hey, I should just style :host", to discovering that this means that either the parent component will be stuck with `display: flex`, or it set display to something else, and inadvertently break the component in a way that isn't noticed until some change happens later.

I recall a Slack thread (I've just mentioned you in it) where there was an issue getting a component to have margin. The issue was that the component was inline, not a block element. An easy solution was to use CSS in the parent to make it a block element.

But now we have the parent element picking how the component is laid out, and it might conflict with assumptions in the component itself. For the `display: block` issue, it should always be possible to pick a value of display that work for both the parent and child, such as `inline-block`, or `inline-flex`. But it's a concern that now has to be shared between the two compnents.

Nateowami (commented):
I agree that having too many options risks losing consistency. But we can probably mitigate that by having some guidelines/standards for what should be used when.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-:-NkvFoh6Eg-Ky26IYlen:bdno3jy)* status: all files reviewed, 7 unresolved discussions (waiting on @nigel-wells and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss` line 5 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkmTKtG8BvaFCqmfmoX:-NkvG9Q73IuYjNhHEY5s:b-6lpb3w) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss#L5)):*

Can the colors match up with our existing ones otherwise it is introducing a bunch of new colors to our palette i.e., purple, green, red, etc.

I also wonder if purple should be our primary as that was the intention of notices created so far?

Also, could we make use of darken and lighten to create the text/border colors rather than supporting two?
</blockquote></details>

As much as possible we should use standardized colors to increase consistency.

Personally I'm not a fan of the purple in this context, and I like the blue a lot better.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss` line 16 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-Nkav3GwE3hrtkKYfKOf:-NkvGLcR3LUeMfsHFfSR:bduq4cv) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss#L16)):*

I'm not sure whether this is a good idea. I've gone from always creating wrapper elements around components for styling reasons, to thinking "hey, I should just style :host", to discovering that this means that either the parent component will be stuck with `display: flex`, or it set display to something else, and inadvertently break the component in a way that isn't noticed until some change happens later.

I recall a Slack thread (I've just mentioned you in it) where there was an issue getting a component to have margin. The issue was that the component was inline, not a block element. An easy solution was to use CSS in the parent to make it a block element.

But now we have the parent element picking how the component is laid out, and it might conflict with assumptions in the component itself. For the `display: block` issue, it should always be possible to pick a value of display that work for both the parent and child, such as `inline-block`, or `inline-flex`. But it's a concern that now has to be shared between the two compnents.
</blockquote></details>

(Resolving and moving to Slack)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.ts` line 16 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkaxlXO87L4sL4KxYxW:-NkaxlXO87L4sL4KxYxX:b-h34brd) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.ts#L16)):*
> ```XML
> export class NoticeComponent implements OnChanges {
>   @Input() type: NoticeType = 'primary';
>   @Input() mode: NoticeMode = 'fill-dark';
> ```

I don't think fill-dark is a good default. I'm guessing you're copying what the default was before, but I don't think it made sense then either.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.ts` line 18 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkvHUIF9RiYiYxGJGnd:-NkvHUIF9RiYiYxGJGne:b8tfqa8) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.ts#L18)):*
> ```XML
>   @Input() mode: NoticeMode = 'fill-dark';
>   @Input() icon?: string;
>   // TODO: remove 'outline' once all components are migrated to use 'mode'
> ```

There aren't that many. Probably easier to just change the default, and then fix any that start to look bad? Most of the ones specifying "outline" are doing so because the default was fill, which is way too strong for most contexts. I think the default should be something that works in most contexts (like fill-light), and then we can just update the few places where that doesn't feel right.

siltomato (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2200)* status: all files reviewed, 5 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts` line 277 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkauxMX6nMvVEz0SfAp:-NlPpM0L80Tg_2MylraY:bsjgzh8) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L277)):*

I wish TypeScript was a tad smarter.
</blockquote></details>

Ah yes, it would be nice if the `undefined` wasn't necessary to avoid `??` when calling `Set.has()`.  I guess it is trying to alert to a potential bug.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss` line 5 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkmTKtG8BvaFCqmfmoX:-NlVWWNs0J59op2P5Y94:by6rf0j) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss#L5)):*

As much as possible we should use standardized colors to increase consistency.

Personally I'm not a fan of the purple in this context, and I like the blue a lot better.
</blockquote></details>

Good idea about the lighten/darken.  I'm not sure that would gain much in this case, though, because the amount to lighten/darken would be different for each notice type (primary, secondary, etc).

If folks experiment with tweaking the colors to something more pleasing, we could still change them pretty easily.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss` line 19 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkmUM05EeRUfEGLzyUe:-NlPta0VAA-wJh5GZSC9:b-kr8znl) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss#L19)):*

Why the larger right padding? If relevant we'll need to adjust for RTL as well
</blockquote></details>

If the notice width is not full width, the extra padding on the right seems nicer to me.  You can see this if you toggle 'inline' in storybook.

Good point about handling RTL.  Fixed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss` line 26 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkmT5G05v-i202n7fct:-NlUYn0a6FiELG_MlNVe:bz83i7g) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss#L26)):*

Some notices have buttons that will need to utilize  these new styles i.e. `users`
</blockquote></details>

Changed notices using `--notice-color` for background to use `--notice-color-button-bg` and `--notice-color-button-text`.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.ts` line 16 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkaxlXO87L4sL4KxYxW:-NlPvFHFCi-kQ_PT1LNu:bjkthbe) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.ts#L16)):*

I don't think fill-dark is a good default. I'm guessing you're copying what the default was before, but I don't think it made sense then either.
</blockquote></details>

Changed to make `fill-light` the default.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.ts` line 18 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkvHUIF9RiYiYxGJGnd:-NlPwVb9ERGhk0m9g26E:b-b87ouy) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.ts#L18)):*

There aren't that many. Probably easier to just change the default, and then fix any that start to look bad? Most of the ones specifying "outline" are doing so because the default was fill, which is way too strong for most contexts. I think the default should be something that works in most contexts (like fill-light), and then we can just update the few places where that doesn't feel right.
</blockquote></details>

Updated references to 'outline' to use the default mode, and update references to 'normal' to use 'primary' type.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.stories.ts` line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkmcPjg3gMUD92CADSa:-NlPzn3P4fXQoeXyNCPA:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.stories.ts#L13)):*

Can you add a story with no icon
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.stories.ts` line 47 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkmcXSiBdk7Ju_iKnW_:-NlR4N1M9ndVJV8l5cEK:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.stories.ts#L47)):*

Can you include a button in each template to ensure they render correctly
</blockquote></details>

Done.

nigel-wells (changes requested):
Just two comments around colors and buttons.

> Reviewed 14 of 16 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-:-NlVeMyo3o554wLtkhR_:b-t4o9w)* status: 21 of 23 files reviewed, 2 unresolved discussions (waiting on @Nateowami and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss` line 5 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkmTKtG8BvaFCqmfmoX:-NlVbeGFEcwXCpR3lNCQ:bqpbdzy) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss#L5)):*

Good idea about the lighten/darken.  I'm not sure that would gain much in this case, though, because the amount to lighten/darken would be different for each notice type (primary, secondary, etc).

If folks experiment with tweaking the colors to something more pleasing, we could still change them pretty easily.
</blockquote></details>

Ok. Can we at least adjust it to match current colors if we have them i.e. purple, green
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss` line 26 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkmT5G05v-i202n7fct:-NlVbpUiDKQuYN3Rgphr:b-xwkfp0) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss#L26)):*

Changed notices using `--notice-color` for background to use `--notice-color-button-bg` and `--notice-color-button-text`.
</blockquote></details>

We'll also need a hover color as they all look green in story book at the moment


Is it possible to force these colors in the notice component rather than relying on the developer to find the relevant variables they need to apply to the component they're working on?

siltomato (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2200)* status: 16 of 23 files reviewed, 2 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss` line 5 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkmTKtG8BvaFCqmfmoX:-NlZlpBS9xizPlLqUxtR:b9xks38) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss#L5)):*

Ok. Can we at least adjust it to match current colors if we have them i.e. purple, green
</blockquote></details>

Ok, I swapped this out for a few of them.  I tried some others, but they didn't look quite right.  I like the palette you created with your original mock-up.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss` line 26 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-NkmT5G05v-i202n7fct:-NlZ9lDS1ig6bScxbKzH:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/83c3504d00536d842a164abad5458fc41cf52c0b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss#L26)):*

We'll also need a hover color as they all look green in story book at the moment


Is it possible to force these colors in the notice component rather than relying on the developer to find the relevant variables they need to apply to the component they're working on?
</blockquote></details>

Done.

nigel-wells (approved):
:lgtm:
Looking good!

> Reviewed 7 of 7 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-:-Nl_rHuDAhp6SjKIMrOY:b-mvoklu)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

Nateowami:
@siltomato What kind of testing were you thinking of doing on this PR?

siltomato (commented):
I was thinking just a check through the notices in the app to make sure they still function, but maybe that isn't necessary if you don't think so.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-:-Nl_s2G253w9dLJvvkj-:b-o4fy0)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

Nateowami:
It's not a bad idea, but given the test team is playing catch-up, and this is relatively low risk, I think it makes sense to just merge. We're going to be doing a full regression run before shipping anyway.

siltomato (commented):
Roger that.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2200#-:-Nl_tSluD7wwNva-NoPk:b-gb7eiv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

PR #2201, opened 2023-12-01
SF-1715 Enable scripture audio feature flag

Nateowami:
@pmachapman I don't know if this is the right approach for turning on a feature flag or not. It's not as clean as I'd like. But I do want feature flags to be able to be turned on by hard coding a new value, so they can be turned on and off by shipping features (rather than having to be separately controlled by changing environment variables). There's a place for changing feature flags using environment variables (e.g. the Serval switchover would be much more of a pain if we had to ship every time). But being able to change them in code and have commits go through our normal shipping process makes the whole testing process so much easier. If something can be handled by version control, keeping it in version control means everything is simpler.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2201)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/feature-flags/feature-flag.service.spec.ts
@@ -68,7 +68,9 @@ describe('FeatureFlagService', () => {
     verify(mockedAnonymousService.featureFlags()).once();
   }));
 
-  it('versionSuffix returns an empty string when no feature flags', fakeAsync(() => {
+  // Disabling test because the feature flag service does not give as a simple way to actually test the state where all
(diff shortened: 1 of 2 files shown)


Nateowami:
@pmachapman The feature flag tests are failing because they assume all feature flags should be disabled by default. I'm leaving this PR as-is for now, since I'm really not sure the best way to untangle this.

RaymondLuong3 (approved):
We have a versioning system in place to tell us which feature flags are turned on from the back end. This adds a new layer to consider, since the Scripture Audio feature can be turned on across the app from the client, but the backend thinks that it is turned off. It is awkward that this implements a work-around to the backend feature flag system. I'm not sure how to make this better, but for the sake of shipping scripture audio, we can move this forward.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2201#-:-NkbEpLv-7R8wH9bZ0E-:b-b9fv0z)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2209, opened 2023-12-04
SF-2382 Fix combined verse support for pre-translation drafts

pmachapman:
This PR allows pre-translations for verses that are combined (i.e. `\v 1-2`) in the source to be correctly added to the draft preview and generated draft when the target does not have a combined verse at that same position.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2209)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.spec.ts
@@ -64,5 +64,18 @@ describe('DraftViewerService', () => {
       const targetOps: DeltaOperation[] = [{ insert: 'existing translation', attributes: { segment: 'verse_1_1' } }];
       expect(service.toDraftOps(draft, targetOps)).toEqual(targetOps);
     });
+
+    it('should allow combined verses in the source that are separated in the target', () => {
+      const draft: DraftSegmentMap = { 'verse_1_1-2': 'In the beginning' };
+      const targetOps: DeltaOperation[] = [
+        { insert: '', attributes: { segment: 'verse_1_1' } },
+        { insert: 'Existing verse 2', attributes: { segment: 'verse_1_2' } }
+      ];
+      const expectedResult: DeltaOperation[] = [
+        { insert: 'In the beginning', attributes: { segment: 'verse_1_1', draft: true } },
+        { insert: 'Existing verse 2', attributes: { segment: 'verse_1_2' } }
+      ];
+      expect(service.toDraftOps(draft, targetOps)).toEqual(expectedResult);
+    });
   });
 });
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts
@@ -47,12 +47,25 @@ export class DraftViewerService {
     }
 
     return targetOps.map(op => {
-      const draftSegmentText: string | undefined = draft[op.attributes?.segment];
-      const isSegmentDraftAvailable = draftSegmentText != null && draftSegmentText.trim().length > 0;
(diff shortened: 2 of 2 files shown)


RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2209)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 61 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2209#-NkvEWKzBqR_fFxwbbHh:-NkvEWKzBqR_fFxwbbHi:b-swi6f6) ([raw file](https://github.com/sillsdev/web-xforge/blob/ea490e5fc54dfedf8a2eafad8a3c5abb79cd4f02/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L61)):*
> ```XML
>         if (combinedVerseNumbers.length > 0) {
>           // Place the combined verse segment in the verse segment
>           draftSegmentText = draft[combinedVerseNumbers[0]];
> ```

Say the draft had verse 2-3 but the target translation did not have verse 2-3 but it did have verse 2 and verse 3. This would put the draft text from verse 2-3 into verse 2. Is that correct? I think that is probably the right approach, and users can modify the text afterwards.

I will point out that if the draft had verse 2-3, and the target had verse 2-3a or verse 2-4, this would not work in those situations. Did you want to fix that here, or create a new card for it since we don't expect that to be common?



_Code quote:_
```XML
          draftSegmentText = draft[combinedVerseNumbers[0]];
```

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2209)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 61 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2209#-NkvEWKzBqR_fFxwbbHh:-NkvNktz2AxXDLeRHJq3:b-5l4rfx) ([raw file](https://github.com/sillsdev/web-xforge/blob/ea490e5fc54dfedf8a2eafad8a3c5abb79cd4f02/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L61)):*
> Say the draft had verse 2-3 but the target translation did not have verse 2-3 but it did have verse 2 and verse 3. This would put the draft text from verse 2-3 into verse 2. Is that correct?

Yes, this is correct.

> I will point out that if the draft had verse 2-3, and the target had verse 2-3a or verse 2-4, this would not work in those situations.

I think I will create a new card for it because it is uncommon and there will likely be other bugs with a-z verses.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2209)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 61 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2209#-NkvEWKzBqR_fFxwbbHh:-NkvPIFO8f61oevz4rMg:b-ucfknh) ([raw file](https://github.com/sillsdev/web-xforge/blob/ea490e5fc54dfedf8a2eafad8a3c5abb79cd4f02/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L61)):*

> Say the draft had verse 2-3 but the target translation did not have verse 2-3 but it did have verse 2 and verse 3. This would put the draft text from verse 2-3 into verse 2. Is that correct?

Yes, this is correct.

> I will point out that if the draft had verse 2-3, and the target had verse 2-3a or verse 2-4, this would not work in those situations.

I think I will create a new card for it because it is uncommon and there will likely be other bugs with a-z verses.
</blockquote></details>

Thanks!

PR #2211, opened 2023-12-05
SF-2387 Convert ResumeCheckingService to be reactive

siltomato:
This should help fix some timing issues surrounding question query and project doc.

The timing issues are evident when a new checking user joins the project and then is incorrectly sent to the 'ALL' book link instead of the book/chapter for the first unanswered question in the project.

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -130,6 +130,7 @@ describe('AppComponent', () => {
     expect(env.menuLength).toEqual(9);
     verify(mockedUserService.setCurrentProjectId(anything(), 'project01')).once();
     tick();
+    discardPeriodicTasks();
   }));
 
   it('navigate to different project', fakeAsync(() => {
@@ -144,6 +145,7 @@ describe('AppComponent', () => {
     // Expect: Community Checking | Manage Questions | Overview | Sync | Settings | Users
     expect(env.menuLength).toEqual(6);
     verify(mockedUserService.setCurrentProjectId(anything(), 'project02')).once();
+    discardPeriodicTasks();
   }));
 
   it('change project', fakeAsync(() => {
@@ -237,6 +239,7 @@ describe('AppComponent', () => {
     env.wait();
     verify(mockedPwaService.activateUpdates()).once();
     tick();
+    discardPeriodicTasks();
   }));
 
   it('user added to project after init', fakeAsync(() => {
@@ -249,6 +252,7 @@ describe('AppComponent', () => {
     env.wait();
     expect(env.isDrawerVisible).toEqual(true);
     expect(env.selectedProjectId).toEqual('project04');
+    discardPeriodicTasks();
   }));
 
   it('user data is set for Bugsnag', fakeAsync(() => {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts
@@ -1,13 +1,13 @@
+import { Injectable } from '@angular/core';
+import { Canon } from '@sillsdev/scripture';
+import { Operation } from 'realtime-server/lib/esm/common/models/project-rights';
+import { SFProjectDomain, SF_PROJECT_RIGHTS } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-rights';
+import { asyncScheduler, from, merge, Observable, of } from 'rxjs';
+import { map, shareReplay, switchMap, throttleTime } from 'rxjs/operators';
+import { QuestionDoc } from 'src/app/core/models/question-doc';
+import { SFProjectProfileDoc } from 'src/app/core/models/sf-project-profile-doc';
 import { ActivatedProjectService } from 'xforge-common/activated-project.service';
 import { UserService } from 'xforge-common/user.service';
-import { SubscriptionDisposable } from 'xforge-common/subscription-disposable';
-import { QuestionDoc } from 'src/app/core/models/question-doc';
-import { RealtimeQuery } from 'xforge-common/models/realtime-query';
-import { Injectable } from '@angular/core';
-import { Operation } from 'realtime-server/lib/esm/common/models/project-rights';
-import { Canon } from '@sillsdev/scripture';
-import { SFProjectDomain, SF_PROJECT_RIGHTS } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-rights';
-import { SFProjectProfileDoc } from 'src/app/core/models/sf-project-profile-doc';
 import { CheckingQuestionsService } from './checking-questions.service';
 
 /**
@@ -16,77 +16,88 @@ import { CheckingQuestionsService } from './checking-questions.service';
  *
  * At present, this means sending the user to the first question that is unanswered, and limiting the scope of questions
  * to the chapter level.
- *
- * TODO Consider making question and link observable.
  */
 @Injectable({ providedIn: 'root' })
-export class ResumeCheckingService extends SubscriptionDisposable {
-  firstUnansweredQuestionQuery?: RealtimeQuery<QuestionDoc>;
+export class ResumeCheckingService {
+  private readonly questionLink$: Observable<string[] | undefined> = this.createLink().pipe(shareReplay(1));
 
   constructor(
     private readonly userService: UserService,
     private readonly activatedProjectService: ActivatedProjectService,
     private readonly questionService: CheckingQuestionsService
-  ) {
-    super();
-    this.subscribe(this.activatedProjectService.projectDoc$, projectDoc => this.refreshQuestionQuery(projectDoc));
-  }
+  ) {}
 
   /**
-   * @returns The question for the user to resume community checking on, or undefined, if no project is selected, the
-   * question has not yet been loaded, or no question matches the criteria (e.g. user has answered all questions'
-   * already)
+   * Gets a path to navigate to the community checking component and resume checking. In general, this will lead
+   * the user to first unanswered question.  If there are no unanswered questions, the path will be to the first
+   * book/chapter in the project.
+   * @returns The path tokens, or undefined if project has no texts.
    */
-  getQuestion(): QuestionDoc | undefined {
-    return this.firstUnansweredQuestionQuery?.docs[0];
+  getLink(): Observable<string[] | undefined> {
+    return this.questionLink$;
   }
 
-  /**
-   * @returns An path to navigate to the community checking component and resume checking. In general this will lead
-   * the user to first unanswered question.
-   */
-  getLink(): string[] {
-    const verseRef = this.getQuestion()?.data?.verseRef;
-    const projectId = this.activatedProjectService.projectId;
+  private createLink(): Observable<string[] | undefined> {
+    return this.activatedProjectService.projectDoc$.pipe(
+      switchMap(projectDoc =>
+        this.getQuestion(projectDoc).pipe(map(question => this.getLinkTokens(projectDoc, question)))
+      )
+    );
+  }
 
-    // TODO Make the code that ultimately consumes this value capable of handling the case where there is no link at
-    // this time without having to resort to this workaround.
-    if (projectId == null) return [''];
+  private getLinkTokens(
+    projectDoc: SFProjectProfileDoc | undefined,
+    questionDoc: QuestionDoc | undefined
+  ): string[] | undefined {
+    if (projectDoc == null) {
+      return undefined;
+    }
 
-    let book: string;
-    let chapter: number;
+    const verseRef = questionDoc?.data?.verseRef;
+    let bookNum: number;
+    let chapterNum: number;
+
     if (verseRef != null) {
-      book = Canon.bookNumberToId(verseRef.bookNum);
-      chapter = verseRef.chapterNum;
-    } else if (this.activatedProjectService.projectDoc != null) {
-      const projectDoc = this.activatedProjectService.projectDoc;
-      const firstText = projectDoc.data?.texts[0];
-      book = Canon.bookNumberToId(firstText?.bookNum ?? Canon.firstBook);
-      chapter = firstText?.chapters[0]?.number ?? 1;
+      bookNum = verseRef.bookNum;
+      chapterNum = verseRef.chapterNum;
     } else {
-      book = Canon.bookNumberToId(Canon.firstBook);
-      chapter = 1;
+      // No questions, or all questions already answered.  Send user to first book/chapter.
+      const firstTextWithChapters = projectDoc.data?.texts.find(t => t.chapters.length > 0);
+
+      if (firstTextWithChapters == null) {
+        return undefined;
+      }
+
+      bookNum = firstTextWithChapters.bookNum;
+      chapterNum = firstTextWithChapters.chapters[0].number;
     }
-    return ['projects', projectId, 'checking', book, String(chapter)];
+
+    return ['projects', projectDoc.id, 'checking', Canon.bookNumberToId(bookNum), chapterNum.toString()];
   }
 
-  private async refreshQuestionQuery(projectDoc: SFProjectProfileDoc | undefined): Promise<void> {
-    this.firstUnansweredQuestionQuery = undefined;
+  /**
+   * Gets the question for the user to resume community checking on.
+   * @returns A question doc, or undefined if no project is selected
+   * or no question matches the criteria (e.g. user has answered all questions' already)
+   */
+  private getQuestion(projectDoc: SFProjectProfileDoc | undefined): Observable<QuestionDoc | undefined> {
     if (projectDoc?.data == null) {
-      return;
+      return of(undefined);
     }
 
-    // ensure user has permission to view questions
+    // Ensure user has permission to view questions
     const userId = this.userService.currentUserId;
     if (!SF_PROJECT_RIGHTS.hasRight(projectDoc.data, userId, SFProjectDomain.Answers, Operation.View)) {
-      return;
+      return of(undefined);
     }
 
-    const query = await this.questionService.queryFirstUnansweredQuestion(projectDoc.id, userId);
-
-    // ensure the project has not changed while awaiting the query
-    if (this.activatedProjectService.projectId === projectDoc.id) {
-      this.firstUnansweredQuestionQuery = query;
-    }
+    return from(this.questionService.queryFirstUnansweredQuestion(projectDoc.id, userId)).pipe(
+      switchMap(query =>
+        merge(query.ready$, query.remoteChanges$, query.localChanges$, query.remoteDocChanges$).pipe(
+          throttleTime(100, asyncScheduler, { leading: false, trailing: true }),
+          map(() => query?.docs[0])
+        )
+      )
+    );
   }
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.html
@@ -30,7 +30,7 @@
         <mat-icon mat-list-icon>bar_chart</mat-icon>
         {{ t("my_progress") }}
       </a>
-      <a mat-list-item [appRouterLink]="answerQuestionsLink" [class.active]="answerQuestionsActive">
+      <a mat-list-item [appRouterLink]="(answerQuestionsLink$ | async) ?? ['']" [class.active]="answerQuestionsActive">
         <mat-icon mat-list-icon class="mirror-rtl">question_answer</mat-icon>
         {{ t("questions_answers") }}
       </a>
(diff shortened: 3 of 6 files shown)


Nateowami (changes requested):
Good work. Some concerns and questions.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-:-NkrugxqB-Qpxl3I3jQ_:b-pmy1pw)* status: all files reviewed, 5 unresolved discussions (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts` line 36 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-NkrqZ9WApvp00l7MDfe:-NkrqZ9WApvp00l7MDff:beb1ysr) ([raw file](https://github.com/sillsdev/web-xforge/blob/d0e3d1e96114b2ab62b17d295be08045b8e29824/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts#L36)):*
> ```XML
>    * @returns The path tokens, or undefined if project has no texts.
>    */
>   getLink(): Observable<string[] | undefined> {
> ```

My only complaint about this is that it's not obvious (without looking at the types) that this is an observable and not just a simple getter. Maybe `get checkingLink$()`? Or make a private `_questionLink$` and expose a public `questionLink$`?

Or maybe you're trying to prevent the query from being run if there's no observer? I'm not familiar enough with RxJS to fully understand how you prevent unnecessary subscriptions.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.ts` line 75 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-NkrsBB6CU7MwgYWBXan:-NkrsBB6CU7MwgYWBXao:b-a0qah2) ([raw file](https://github.com/sillsdev/web-xforge/blob/d0e3d1e96114b2ab62b17d295be08045b8e29824/src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.ts#L75)):*
> ```XML
> 
>     try {
>       const [projectUserConfigDoc, projectDoc] = await Promise.all([
> ```

Thanks for removing a hideous serial await.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.ts` line 105 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-NkruAyzB0o8fzrdvHay:-NkruAyzB0o8fzrdvHaz:bl55tvd) ([raw file](https://github.com/sillsdev/web-xforge/blob/d0e3d1e96114b2ab62b17d295be08045b8e29824/src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.ts#L105)):*
> ```XML
>       }
>     } finally {
>       this.loadingFinished();
> ```

You're calling this well before we've even redirected the user to checking, since there's no awaiting `this.navigateToTranslate`. Shouldn't it return a promise that can be awaited?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.ts` line 122 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-NkrtYcG0hVm18Kbdj8z:-NkrtYcG0hVm18Kbdj9-:b2flkb2) ([raw file](https://github.com/sillsdev/web-xforge/blob/d0e3d1e96114b2ab62b17d295be08045b8e29824/src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.ts#L122)):*
> ```XML
> 
>       const chapterNum: number | undefined =
>         projectUserConfig.selectedChapterNum ?? project.texts[bookNum]?.chapters[0].number;
> ```

`chapters[0]` could be undefined, so `chapters[0].number` could cause an exception. TypeScript just isn't strict enough to force us to check for an array not having a value at a given index. If you change it to `chapters[0]?.number` it would be safe.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.ts` line 132 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-Nkrt6aO3mHybokEh3I_:-Nkrt6aO3mHybokEh3Ia:b-a1fio0) ([raw file](https://github.com/sillsdev/web-xforge/blob/d0e3d1e96114b2ab62b17d295be08045b8e29824/src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.ts#L132)):*
> ```XML
>   }
> 
>   private navigateToChecking(projectId: string, task: TaskType = 'checking'): void {
> ```

Why is the task hard-coded here, but not for `navigateToTranslate`?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.ts` line 135 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-NkrseL32FHTvTM3Xdwz:-NkrseL32FHTvTM3Xdx-:b46xook) ([raw file](https://github.com/sillsdev/web-xforge/blob/d0e3d1e96114b2ab62b17d295be08045b8e29824/src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.ts#L135)):*
> ```XML
>     const defaultCheckingLink: string[] = ['/projects', projectId, task];
> 
>     this.resumeCheckingService.getLink().subscribe(link => {
> ```

I'm I misreading this? Won't this cause the user to be redirected every time a new link is generated by the ResumeCheckingService? Don't we just want the very first result, navigate, and then finish?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2211)* status: all files reviewed, 2 unresolved discussions (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts` line 36 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-NkrqZ9WApvp00l7MDfe:-NkvCfb--VJaIQWSEiRC:bfa0yr3) ([raw file](https://github.com/sillsdev/web-xforge/blob/d0e3d1e96114b2ab62b17d295be08045b8e29824/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts#L36)):*

Changed to getter with $.  I'm torn in regards to appending '$' for functions that return an observable.  I like the idea, and I've used it myself, but I'm still not sure how to apply that convention in regard to http services.  It feels too much to append $ to all the api methods.
</blockquote></details>

I see what you mean. I think it's a bit odd to have a method end with `$`, but maybe if it's a getter it looks OK?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts` line 65 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-NkvBkTuDhdTazpBPKyv:-NkvBkTuDhdTazpBPKyw:b-swtc9a) ([raw file](https://github.com/sillsdev/web-xforge/blob/ec58b03c9493d7fdb114a30a06b913739e63f4f6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts#L65)):*
> ```XML
>     } else {
>       // No questions, or all questions already answered.  Send user to first book/chapter.
>       const firstTextWithChapters = projectDoc.data?.texts.find(t => t.chapters.length > 0);
> ```

Strictly speaking, it's possible the project document has loaded, but doesn't have any chapters yet, so this function would return undefined, but then there's nothing to trigger a recalculation of the link.

I'm not sure whether that case is far-fetched or not.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts` line 97 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-Nkv8tRcEUDo5r8EHs7k:-Nkv8tRcEUDo5r8EHs7l:b-ec6wxv) ([raw file](https://github.com/sillsdev/web-xforge/blob/ec58b03c9493d7fdb114a30a06b913739e63f4f6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts#L97)):*
> ```XML
>       switchMap(query =>
>         merge(query.ready$, query.remoteChanges$, query.localChanges$, query.remoteDocChanges$).pipe(
>           throttleTime(100, asyncScheduler, { leading: false, trailing: true }),
> ```

I'm concerned this `throttleTime()` is buying us less than it would appear, and costing us more than it would appear.

Firstly, the only thing it's buying us is:
- Reducing how often `getLinkTokens()` is called (which is a pretty quick method).
- Reducing how often the `checkingLink$` observable emits (which a consumer could throttle if necessary).

What it's costing us is 100ms of delay in redirecting the user, since we're waiting for this to emit before doing the redirect. If we did this a lot of places, all these delays would add up in addition to the network delays. I think we're better off using `distinctUntilChanged()` at some point (probably after calculating the link) to prevent this from emitting too often. It might also prevent the tests from having to discard periodic tasks.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.ts` line 105 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-NkruAyzB0o8fzrdvHay:-NkvCInNBc5ZeM9TOvKY:b-ucfkn4) ([raw file](https://github.com/sillsdev/web-xforge/blob/d0e3d1e96114b2ab62b17d295be08045b8e29824/src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.ts#L105)):*

Good point.  Fixed.
</blockquote></details>

Thanks.

siltomato (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2211)* status: 4 of 9 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts` line 65 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-NkvBkTuDhdTazpBPKyv:-Nl-AV4O9Wjw6C_Dr7sm:bp84623) ([raw file](https://github.com/sillsdev/web-xforge/blob/ec58b03c9493d7fdb114a30a06b913739e63f4f6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts#L65)):*

Strictly speaking, it's possible the project document has loaded, but doesn't have any chapters yet, so this function would return undefined, but then there's nothing to trigger a recalculation of the link.

I'm not sure whether that case is far-fetched or not.
</blockquote></details>

Is there an event that we could listen for?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts` line 97 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-Nkv8tRcEUDo5r8EHs7k:-NkzsnyO725z9xDJccii:br7im31) ([raw file](https://github.com/sillsdev/web-xforge/blob/ec58b03c9493d7fdb114a30a06b913739e63f4f6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts#L97)):*

I'm concerned this `throttleTime()` is buying us less than it would appear, and costing us more than it would appear.

Firstly, the only thing it's buying us is:
- Reducing how often `getLinkTokens()` is called (which is a pretty quick method).
- Reducing how often the `checkingLink$` observable emits (which a consumer could throttle if necessary).

What it's costing us is 100ms of delay in redirecting the user, since we're waiting for this to emit before doing the redirect. If we did this a lot of places, all these delays would add up in addition to the network delays. I think we're better off using `distinctUntilChanged()` at some point (probably after calculating the link) to prevent this from emitting too often. It might also prevent the tests from having to discard periodic tasks.
</blockquote></details>

Yes, that makes sense.  Changed to use `distinctUntilChanged()`.

siltomato (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2211)* status: 7 of 9 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts` line 65 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-NkvBkTuDhdTazpBPKyv:-Nl-MeQ51YUu2uAX7hzY:bcd5jrd) ([raw file](https://github.com/sillsdev/web-xforge/blob/ec58b03c9493d7fdb114a30a06b913739e63f4f6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts#L65)):*

ActivatedProjectService.changes$ emits every time the project changes.
</blockquote></details>

Ah, I remember you mentioning this now.  Changed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts` line 111 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-Nl-BduA-MUrm7XxRw_g:-Nl-FrBsDv9bRi274aip:b-8ud65w) ([raw file](https://github.com/sillsdev/web-xforge/blob/0af72fce0856ddc913991aab97a19c0a1aa874d0/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts#L111)):*

Does `ready$` ever fire offline? My recollection is that it does not.
</blockquote></details>

Fixed in a manner similar to how we handle in checking component.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/util/string-util.ts` line 4 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-Nl-BtdaD19NBTxOqGF9:-Nl-F2HwEiS-qy1x2D5u:bglv99d) ([raw file](https://github.com/sillsdev/web-xforge/blob/0af72fce0856ddc913991aab97a19c0a1aa874d0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/util/string-util.ts#L4)):*

Does Lodash not have a method for this?

I'm not against Lodash; my hesitance to use certain functions in the past was due `isString()` in Lodash not being the same as `typeof value === 'string'`, and wanting to be able to look at a type check and know *exactly* what it is checking. Also, the types don't work out the same in TS.

Writing this manually is also OK; I just want to make sure my hesitance to use a Lodash function in a particular case in the past wasn't keeping you from ever using Lodash again.
</blockquote></details>

Ha, it did play into it a bit =).  Lodash has [`isEqual`](https://lodash.com/docs/4.17.15#isEqual) that would work.  I'm fine either way.  How do you lean?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/util/string-util.spec.ts` line 17 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-Nl-CRqfB_mvqW2NoyDa:-Nl-GK1A7sVXpyiuJE4s:bnl9wms) ([raw file](https://github.com/sillsdev/web-xforge/blob/0af72fce0856ddc913991aab97a19c0a1aa874d0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/util/string-util.spec.ts#L17)):*

You already have a check for different lengths; maybe this was supposed to be the same length but different contents? You don't have any examples where the length is the same and the content is different.
</blockquote></details>

Copy and paste error.  Good catch.

Nateowami (approved):
:lgtm: I still want to test this, which I haven't done yet.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-:-Nl-NI0K7Nxv570k-DBz:bjy5qq8)* status: all files reviewed, 1 unresolved discussion (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts` line 111 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-Nl-BduA-MUrm7XxRw_g:-Nl-N779EoyBAoCwj-B3:b-abi9eo) ([raw file](https://github.com/sillsdev/web-xforge/blob/0af72fce0856ddc913991aab97a19c0a1aa874d0/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts#L111)):*

Fixed in a manner similar to how we handle in checking component.
</blockquote></details>

We need a better solution so we don't run into this exact problem everywhere, but this is an acceptable solution for this PR.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/util/string-util.ts` line 4 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-Nl-BtdaD19NBTxOqGF9:-Nl-NN31AoTdsfkQbZ2N:b-djq12u) ([raw file](https://github.com/sillsdev/web-xforge/blob/0af72fce0856ddc913991aab97a19c0a1aa874d0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/util/string-util.ts#L4)):*

Ha, it did play into it a bit =).  Lodash has [`isEqual`](https://lodash.com/docs/4.17.15#isEqual) that would work.  I'm fine either way.  How do you lean?
</blockquote></details>

I don't have a strong opinion.

siltomato (commented):
Acceptance tests added.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-:-Nl-gJbKE6--7oH0OBNg:bxy3r75)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts` line 111 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-Nl-BduA-MUrm7XxRw_g:-Nl-gH7i8cSquBwrXGGT:bwygxsx) ([raw file](https://github.com/sillsdev/web-xforge/blob/0af72fce0856ddc913991aab97a19c0a1aa874d0/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts#L111)):*

We need a better solution so we don't run into this exact problem everywhere, but this is an acceptable solution for this PR.
</blockquote></details>

Noted.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/util/string-util.ts` line 4 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-Nl-BtdaD19NBTxOqGF9:-Nl-g9x575Apc5hOAdBu:b-k46qyw) ([raw file](https://github.com/sillsdev/web-xforge/blob/0af72fce0856ddc913991aab97a19c0a1aa874d0/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/util/string-util.ts#L4)):*

I don't have a strong opinion.
</blockquote></details>

I'll leave it as-is.

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2211)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @siltomato)

Nateowami (changes requested):
Unfortunately I'm seeing problems with this in my basic testing. I'm hitting a MaxListenersExceededWarning, and worse, seeing incorrect behavior from the user's perspective.

It refuses to let me switch from a translate only project (it's a resource) to a project where I have community checking, and last time I was there, I was on community checking.

It may be a good idea to add a spec for this file, though I think doing so could be challenging.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-:-Nl-ogLQ-IAMIPMPCnPs:blwd3uv)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @siltomato)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2211)* status: all files reviewed, 1 unresolved discussion (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts` line 24 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-NlR-rdoEV9jBnOY-J5a:-NlOuDJT1zAlHsuRVXiM:br4rygj) ([raw file](https://github.com/sillsdev/web-xforge/blob/64e62e165a865c4ddc9628f362349047a1a7f820/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts#L24)):*
> ```XML
> @Injectable({ providedIn: 'root' })
> export class ResumeCheckingService {
>   private readonly questionLink$: Observable<string[] | undefined> = this.createLink().pipe(shareReplay(1));
> ```

I think removing this has the unfortunate effect of having no value emitted until the value changes. Some components might need it to act like a BehaviorSubject, and then not have any link. Maybe the project component should watch the projectId, and then wait for this to emit with a link that has the ID of that project?

Nateowami:
I'm going to move this to ready to test nonetheless, since although I still have a bit of concern that this might have a bit of an issue, it could always be addressed in a followup, and I may be wrong or it may be a minor issue.

siltomato (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2211)* status: 7 of 9 files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts` line 24 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-NlR-rdoEV9jBnOY-J5a:-NlTzIHG1880Z-ckGRCi:bhcgb0a) ([raw file](https://github.com/sillsdev/web-xforge/blob/64e62e165a865c4ddc9628f362349047a1a7f820/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/resume-checking.service.ts#L24)):*

I think removing this has the unfortunate effect of having no value emitted until the value changes. Some components might need it to act like a BehaviorSubject, and then not have any link. Maybe the project component should watch the projectId, and then wait for this to emit with a link that has the ID of that project?
</blockquote></details>

I went with a similar idea, but I moved the check to the service.

Nateowami (approved):
> Reviewed 2 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2211)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

Nateowami (commented):
:lgtm:

> Reviewed 4 of 4 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2211#-:-NlZ1U3hE3QvUtU0fxA9:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

PR #2212, opened 2023-12-05
SF-2289 Fixed rightward runoff at medium resolutions

josephmyers:
Appearance at given range (992px-1080px) after fix:
image

Though I didn't verify on an actual Ipad Mini, I was able to see the bug on my laptop before fixing it.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2212)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html
@@ -63,7 +63,7 @@
               <ng-template matTabLabel>
                 <span class="tab-label">
                   <mat-icon>mic</mat-icon>
-                  {{ media.isActive("xs") ? t("record") : t("record_upload") }}
+                  {{ t("record") }}
                 </span>
               </ng-template>
               <!-- Lazy loading causes heavy components like this to reload when switching tabs -->
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.scss
@@ -2,6 +2,11 @@
 @use '../checking-global-vars' as checking-vars;
 @use 'bootstrap/scss/mixins/breakpoints';
 
+:host ::ng-deep .mat-tab-label {
+  min-width: unset;
+  padding-inline: 16px;
+}
+
 .answers-component {
   display: flex;
   flex-direction: column;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss
@@ -7,6 +7,11 @@ $questionsPanelBreakpoint: 700px;
 
 $questionsHeaderInlineStartPadding: 8px;
 
+:host {
+  display: flex;
+  flex: 1;
+}
+
 header {
   display: flex;
   padding: 8px 0;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss
@@ -7,11 +7,6 @@ $questionsPanelBreakpoint: 700px;
 
 $questionsHeaderInlineStartPadding: 8px;
 
-:host {
-  flex: 1;
-  display: flex;
-}
-
 header {
   display: flex;
   padding: 8px 0;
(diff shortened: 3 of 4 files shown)


Nateowami (changes requested):
I don't really like the UI this produces. The tab system is bad to begin with (and we want to remove it), and this makes it even worse in this case.

In my investigation, the width was caused by these two declarations from Material:

```
.mat-tab-label {
    padding: 0 24px;
    min-width: 160px;
}
```

And there's never been a good reason we've had "Upload" in the audio tab. Uploading audio recordings of answers seems like a very odd thing to do (terrible workflow) and certainly doesn't need to be in the tab.


Changing the tab label to just "Record" and adding this CSS override fixes the layout entirely:
```
:host ::ng-deep .mat-tab-label {
  min-width: unset;
  padding-inline: 20px;
}
```
And we end up with this:


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2212#-:-NkujU8P4Ar4_YrudX_E:b2428ro)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 11 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2212#-Nkum2eM28SxXLgmOu9d:-Nkum2eN1xbnnrDX5O2i:b8uydn6) ([raw file](https://github.com/sillsdev/web-xforge/blob/f40bf69b8991e9713a3b9b6cdb806a0272c0d9c0/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L11)):*
> ```SCSS
> 
> :host {
>   flex: 1;
> ```

The removal of this declaration completely breaks the UI.

josephmyers (commented):
Sure, I have no problem changing the tab size here and removing Record/Upload. Unfortunately this only masks the problem, which is that the dynamic tab sizing logic from Material isn't great. It looks at the overall screen width, instead of the visible space. And it certainly doesn't help that their tab scrolling solution is pretty ugly.

Your change still exposes tab scrolling at 1000px. I used 16px instead, which fixes it, assuming we're wanting to avoid tab scrolling altogether. Let me know if that's an issue. Looking forward to being rid of tabs!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2212#-:-Nkx00WY5iMd9oz47cJz:b-2fpmof)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 11 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2212#-Nkum2eM28SxXLgmOu9d:-Nkx0s38398UAP9JuDm_:b-mk74rp) ([raw file](https://github.com/sillsdev/web-xforge/blob/f40bf69b8991e9713a3b9b6cdb806a0272c0d9c0/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L11)):*

The removal of this declaration completely breaks the UI.
</blockquote></details>

I'm not sure what completely broken means, as I'm not seeing that in my testing. This comment isn't constructive, given that my changes fix the reported bug, as shown in my screenshot.

If you're referring to the padding at the bottom of the page, that has long been present, though in my mind it looks weird and should be removed. It'd be more helpful to indicate what page or portion of a page is broken, or even better to post a screenshot.

That said, this isn't necessary when using your fix, so I undid it. Thanks for the help.

Nateowami (approved):
:lgtm:

> Reviewed 3 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2212#-:-NkzFpqR1f0VqalxA0jo:bnfp4nl)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 11 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2212#-Nkum2eM28SxXLgmOu9d:-NkzF6lG4kS7Ttu5_d_T:bmd8m7e) ([raw file](https://github.com/sillsdev/web-xforge/blob/f40bf69b8991e9713a3b9b6cdb806a0272c0d9c0/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L11)):*

I'm not sure what completely broken means, as I'm not seeing that in my testing. This comment isn't constructive, given that my changes fix the reported bug, as shown in my screenshot.

If you're referring to the padding at the bottom of the page, that has long been present, though in my mind it looks weird and should be removed. It'd be more helpful to indicate what page or portion of a page is broken, or even better to post a screenshot.

That said, this isn't necessary when using your fix, so I undid it. Thanks for the help.
</blockquote></details>

Sorry; it was so glaring in my testing I thought it was just accidentally included by mistake. After a bit more testing, I see this only happens when the chapter document is very short (notice the massive gap of space at the bottom).

PR #2214, opened 2023-12-05
Fix clipping of icons in nav component

Nateowami:
The Q&A icon, and the users icon, were both being clipped by the mat icon border-radius.

This could be fixed by just setting border-radius to 0, but the reason this was a problem is because I had set the padding on the icons to 0. I did this to get the spacing how I wanted it. However, there's a more straightforward way to get the spacing than removing the padding.

Before, with padding removed from icon:

After, with padding restored to icon, padding reduced on left side of icon, and gap between icon and the text adjusted.

See Storybook for a comprehensive understanding of what actually changed visually.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2214)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss
@@ -9,7 +9,6 @@ mat-nav-list .mat-list-item {
     color: variables.$purpleMedium;
     fill: variables.$purpleMedium;
     transition: color 0.15s;
-    padding: 0;
   }
   &:hover {
     mat-icon {
@@ -23,7 +22,8 @@ mat-nav-list .mat-list-item {
   }
 
   ::ng-deep .mat-list-item-content {
-    column-gap: 16px;
+    padding-inline: 12px;
+    column-gap: 12px;
   }
 
   &.activated-nav-item,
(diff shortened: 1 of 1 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2214)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss` line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2214#-Nkv1cct9QL4IGo29Fup:-Nkv1cct9QL4IGo29Fuq:bnnybqd) ([raw file](https://github.com/sillsdev/web-xforge/blob/09a62af3782ccb8e4dbf1d8f84fe02f9f3dbaaed/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss#L68)):*
> ```SCSS
>   font-size: 1.4rem;
>   inset-inline-end: -7px;
>   top: -7px;
> ```

I initially missed that this change was needed as well. Chromatic caught it:

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2214)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2216, opened 2023-12-05
SF-2301 Enable resolving note threads in SF

RaymondLuong3:
* specify status when note dialog is closed
* update note thread status to be the status of the saved note
* allow note content to be optional
* add trigger for save options menu next to save button

The PR introduces the logic to allow Paratext users with permission to add notes to resolve a note thread. Users with the Commenter role are unable to resolve note threads. When a thread is resolved, they can no longer be viewed in SF unless a user synchronizes their project to PT and the note is unresolved in PT. I don't expect that it will be often that a user will need to unresolve a note, but it can be done.

Resolve button translator view
Resolve a note options

Save button commenter view
Resolve a note commenter disabled

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2216)

### DIFF
--- a/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts
+++ b/src/RealtimeServer/scriptureforge/models/sf-project-rights.ts
@@ -69,8 +69,8 @@ const rightsByRole: Record<SFProjectRole, { [domain in `${SFProjectDomain}`]?: `
     answer_comments: ['view'],
     likes: ['view'],
     biblical_terms: ['view'],
-    pt_note_threads: ['view', 'create', 'delete_own'],
-    sf_note_threads: ['view', 'create', 'delete_own'],
+    pt_note_threads: ['view', 'create', 'edit', 'delete_own'],
+    sf_note_threads: ['view', 'create', 'edit', 'delete_own'],
     notes: ['view', 'create', 'edit_own', 'delete_own'],
     text_audio: ['view']
   },
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -2827,6 +2827,32 @@ describe('EditorComponent', () => {
       env.dispose();
     }));
 
+    it('allows resolving a note', fakeAsync(() => {
+      const projectId: string = 'project01';
+      const threadDataId: string = 'dataid01';
+      const content: string = 'This thread is resolved.';
+      const env = new TestEnvironment();
+      let noteThread: NoteThreadDoc = env.getNoteThreadDoc(projectId, threadDataId);
+      expect(noteThread.data!.notes.length).toEqual(3);
+
+      env.setProjectUserConfig();
+      env.wait();
+      let noteThreadIconElem: HTMLElement | null = env.getNoteThreadIconElement('verse_1_1', threadDataId);
+      noteThreadIconElem!.click();
+      verify(mockedMatDialog.open(NoteDialogComponent, anything())).once();
+      env.mockNoteDialogRef.close({ noteContent: content, status: NoteStatus.Resolved });
+      env.wait();
+      noteThread = env.getNoteThreadDoc(projectId, threadDataId);
+      expect(noteThread.data!.notes.length).toEqual(4);
+      expect(noteThread.data!.notes[3].content).toEqual(content);
+      expect(noteThread.data!.notes[3].status).toEqual(NoteStatus.Resolved);
+
+      // the icon should be hidden from the editor
+      noteThreadIconElem = env.getNoteThreadIconElement('verse_1_1', threadDataId);
+      expect(noteThreadIconElem).toBeNull();
+      env.dispose();
+    }));
+
     it('can open dialog of the second note on the verse', fakeAsync(() => {
       const env = new TestEnvironment();
       env.setProjectUserConfig();
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
@@ -108,11 +108,12 @@ import { TranslateMetricsSession } from './translate-metrics-session';
 export const UPDATE_SUGGESTIONS_TIMEOUT = 100;
 
 export interface SaveNoteParameters {
-  content: string;
+  content?: string;
   dataId?: string;
   threadDataId?: string;
   verseRef?: VerseRef;
   biblicalTermId?: string;
+  status?: NoteStatus;
 }
 
 const PUNCT_SPACE_REGEX = /^(?:\p{P}|\p{S}|\p{Cc}|\p{Z})+$/u;
@@ -1142,7 +1143,7 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
       content: params.content,
       conflictType: NoteConflictType.DefaultValue,
       type: NoteType.Normal,
-      status: NoteStatus.Todo,
+      status: params.status ?? NoteStatus.Todo,
       deleted: false,
       editable: true,
       versionNumber: 1
@@ -1182,7 +1183,11 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
         }
       } else {
         note.threadId = threadDoc.data!.threadId;
-        await threadDoc.submitJson0Op(op => op.add(t => t.notes, note));
+        await threadDoc.submitJson0Op(op => {
+          op.add(t => t.notes, note);
+          // also set the status of the thread to be the status of the note
+          op.set(t => t.status, note.status);
+        });
         await this.updateNoteReadRefs(note.dataId);
       }
     }
@@ -1257,12 +1262,13 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
     const result: NoteDialogResult | undefined = await dialogRef.afterClosed().toPromise();
 
     if (result != null) {
-      if (result.noteContent != null) {
+      if (result.noteContent != null || result.status != null) {
         await this.saveNote({
           content: result.noteContent,
           threadDataId: threadDataId,
           dataId: result.noteDataId,
-          verseRef: currentVerseRef
+          verseRef: currentVerseRef,
+          status: result.status
         });
       }
       this.toggleNoteThreadVerseRefs$.next();
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
@@ -1167,12 +1167,12 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
       };
       await this.projectService.createNoteThread(this.projectId, noteThread);
     } else {
-      // updated the existing note
       const threadDoc: NoteThreadDoc = await this.projectService.getNoteThread(
         getNoteThreadDocId(this.projectId, params.threadDataId)
       );
       const noteIndex: number = threadDoc.data!.notes.findIndex(n => n.dataId === params.dataId);
       if (noteIndex >= 0) {
+        // updated the existing note
         if (threadDoc.data?.notes[noteIndex].editable === true) {
           await threadDoc!.submitJson0Op(op => {
             op.set(t => t.notes[noteIndex].content, params.content);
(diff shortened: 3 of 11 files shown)


josephmyers (changes requested):
I was able to unresolve in Paratext a resolve note. Cool! Here's what it looks like:


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2216#-:-NkyS1uA6cHemY5IsnIu:b-66d9ic)* status: all files reviewed, 6 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1170 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2216#-Nky4TnZ9QU4gbwchCnf:-Nky4TnZ9QU4gbwchCng:b-2lpcp) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7a9db1092aaad30cf19e996ad3c86390c73d9f2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1170)):*
> ```XML
>       await this.projectService.createNoteThread(this.projectId, noteThread);
>     } else {
>       // updated the existing note
> ```

This comment was confusing to me. Can you move it closer to the note updating code?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1189 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2216#-Nky6U7QAkSttkh1jvM1:-Nky6U7R4iPrIKbOyFBP:b-o6xvv7) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7a9db1092aaad30cf19e996ad3c86390c73d9f2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1189)):*
> ```XML
>           op.add(t => t.notes, note);
>           // also set the status of the thread to be the status of the note
>           op.set(t => t.status, note.status);
> ```

Is there any need to also update the status of the other notes in the thread?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html` line 59 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2216#-NkyBjXJCxmMZ3hm2w1_:-NkyBjXJCxmMZ3hm2w1a:b8gxztw) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7a9db1092aaad30cf19e996ad3c86390c73d9f2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html#L59)):*
> ```HTML
>   <mat-dialog-actions fxLayoutAlign="end">
>     <mat-button-toggle-group>
>       <mat-button-toggle class="close-button" (click)="close()">{{ t("close") }}</mat-button-toggle>
> ```

Potentially disregard: Can we move this button out of the toggle group? It makes the Save button have a border radius only on the right side :D
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html` line 60 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2216#-NkyFjyZ1NBn4qDq4xu8:-NkyFjyZ1NBn4qDq4xu9:b-bn1bmo) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7a9db1092aaad30cf19e996ad3c86390c73d9f2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html#L60)):*
> ```HTML
>     <mat-button-toggle-group>
>       <mat-button-toggle class="close-button" (click)="close()">{{ t("close") }}</mat-button-toggle>
>       <mat-button-toggle *ngIf="canInsertNote" class="save-button save-options" (click)="submit()">{{
> ```

So Close closes without saving any text, and Save saves the added text, of course. Having a Save and Resolve button to me is strange, because it doesn't make sense for folks just wanting to Resolve. The button resolves whether there's text or not, but it's confusing that it's associated with Save, which involves adding content. The JIRA task says users need this to reduce clutter, so it's all the more likely the form will be empty. They want to open the dialog and resolve the thread.

It doesn't seem a safe assumption that they will always want to add a note and resolve/disappear the thread in one click. What if they made a typo or otherwise need to go back and edit something? It seems safer to me, at first glance, to have three buttons: Close, Save, and Resolve, all visible up front. Save would be disabled while the form is empty, and clicking it would add the note without closing the dialog. Hang with me. Resolve is the opposite, enabled while the form is empty.

Reasons this works:
- Resolve button is not hidden behind a small icon and menu
- Save button re-enables Resolve button
- Saving will often be followed by other actions nearby, e.g. Save and Resolve.
- Confirms they're happy with the thread as visible before resolving.
- Resolve disabling is noticeably associated with entering onto the form

I know you've already done a bunch of work on this, so let me know your thoughts!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html` line 64 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2216#-NkyC-lb7hpnUS7nmN64:-NkyC-lcEVRscX4uUfh2:bu4g1l) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7a9db1092aaad30cf19e996ad3c86390c73d9f2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html#L64)):*
> ```HTML
>       }}</mat-button-toggle>
>       <mat-button-toggle *ngIf="canResolve" class="save-options-trigger save-options" [matMenuTriggerFor]="saveMenu">
>         <mat-icon>expand_less</mat-icon>
> ```

Potentially disregard: I think we need a separator between the button text and chevron, because it's not clear at first that part of the button takes the action and the other part opens a menu.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.scss` line 111 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2216#-NkyCp3N5VUGVClEL-4C:-NkyCp3N5VUGVClEL-4D:b-284l5) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7a9db1092aaad30cf19e996ad3c86390c73d9f2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.scss#L111)):*
> ```SCSS
> }
> 
> .save-options {
> ```

Potentially disregard: This is really picky, but can you move the chevron up a couple pixels? It looks a little low by default.

josephmyers (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2216)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html` line 60 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2216#-NkyFjyZ1NBn4qDq4xu8:-Nl1aUgm3gsIhGQ1A6v2:b-et7kqh) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7a9db1092aaad30cf19e996ad3c86390c73d9f2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html#L60)):*

Mark was here and we thought about is best in terms of options for the action buttons on this dialog. We discussed if it would make sense to have all the buttons upfront (which is what I initially proposed to Nigel). We decided since resolving a note makes it disappear from SF, we want there to be an intentional choice to resolve a note, which is why we decided to not have the resolve button upfront.
Since resolving notes is reserved only for paratext users, they will be familiar with the resolving feature which is possible with or without text content involved. I think I will leave it as-is for that reason.

Mark suggested that if we choose "resolve" from the menu, it takes action immediately rather than change the save button behaviour. I am open that that. Do you have a suggestion whether it is better to update the save button to be a "Resolve button", or for the Resolve action to be immediate upon selecting it in the menu.
</blockquote></details>

If you guys have already discussed it at length, I'm fine giving this a pass. I still think it will go mostly undiscovered. As you said, our users will be from PT, and it's visible up front there. I do like the idea of Resolve taking effect when they click the menu button, but again, that's a "no going back" button, so I think it would frustrate some people that way.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.scss` line 111 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2216#-NkyCp3N5VUGVClEL-4C:-Nl1b_ru6PV75PRsk3ab:b1wu0ix) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7a9db1092aaad30cf19e996ad3c86390c73d9f2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.scss#L111)):*

Hmm, it is really hard to notice. I'll leave it for now unless someone else notices
</blockquote></details>

Yeah, with the other changes you've made, it looks pretty good. Thanks
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 273 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2216#-Nl1bhnUCg6Ji3Hsf2LI:-Nl1bhnUCg6Ji3Hsf2LJ:bsv6jrk) ([raw file](https://github.com/sillsdev/web-xforge/blob/c31c65f81b1609f1d0ef5f11ddfb9b451342838d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L273)):*
> ```JSON
>     "permanently_delete_note": "Permanently delete this comment?",
>     "reattached": "(Re-attached)",
>     "resolve": "Resolve",
> ```

Cool, I like where I think you're going. But now it's confusing as to whether any text in the form will be saved or not. Once again, you run into issues with the structure of these buttons. You could try to disable the form while Resolve is selected, but that's not perfect. You could go back to the previous wording?

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2216)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html` line 60 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2216#-NkyFjyZ1NBn4qDq4xu8:-Nl52RJZ6auMMzSQEFuS:b-a5dn60) ([raw file](https://github.com/sillsdev/web-xforge/blob/a7a9db1092aaad30cf19e996ad3c86390c73d9f2/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.html#L60)):*

If you guys have already discussed it at length, I'm fine giving this a pass. I still think it will go mostly undiscovered. As you said, our users will be from PT, and it's visible up front there. I do like the idea of Resolve taking effect when they click the menu button, but again, that's a "no going back" button, so I think it would frustrate some people that way.
</blockquote></details>

That's right, we want to make any action about resolve a note require intentionality. If a user selects the "save and resolve" option, then we update the button text and still require that a user clicks the button.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 273 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2216#-Nl1bhnUCg6Ji3Hsf2LI:-Nl52K2p1ikShUnG42MC:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/c31c65f81b1609f1d0ef5f11ddfb9b451342838d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L273)):*

Cool, I like where I think you're going. But now it's confusing as to whether any text in the form will be saved or not. Once again, you run into issues with the structure of these buttons. You could try to disable the form while Resolve is selected, but that's not perfect. You could go back to the previous wording?
</blockquote></details>

Done.

josephmyers (approved):
I'm not sold on the UX here, but I do like the way the button looks. This is optional, but how difficult would it be to style the menu option to look more like the button itself? Thanks

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2216#-:-NlMfy3j9wPUEPFopLTL:b-skanrk)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
Right, I've been flip-flopping on how to provide this feature in the most useable way, and part of that is providing something that is not too different from Paratext since this will be visible to Paratext users only. But how it is now can always be improved on as we use it and get feedback from users and testers.
I'm never tried styling the menu before. Do you think the font should match the button? I don't see a way to configure it on the documentation for the material component, so we would have to provide our own css.



> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2216#-:-NlPhzUi-bdfL3JlvDzr:budmakq)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

josephmyers:
> Right, I've been flip-flopping on how to provide this feature in the most useable way, and part of that is providing something that is not too different from Paratext since this will be visible to Paratext users only. But how it is now can always be improved on as we use it and get feedback from users and testers. I'm never tried styling the menu before. Do you think the font should match the button? I don't see a way to configure it on the documentation for the material component, so we would have to provide our own css.

Yeah, it would be providing our own CSS. To me, it looks out of place with the slick, black "Save" button. It's not critical, just affects the overall impression we give off

PR #2219, opened 2023-12-06
SF-2145 Hide the PWA installation dialog on mobile

nigel-wells:
* Capture chromium install prompt and cancel default prompt
* Added new install button to avatar menu
* Added install icon to avatar

Please note that this PR only addresses the issue on Chromium devices as it is the only engine that supports the `beforeinstallprompt` event. It is also the only engine that offers additional methods to help determine if an app has been installed or not.

To test this on localhost you will need to build the PWA version.

By default on Android, the prompt would display like this in the footer:
image

Now the user will see an icon in the header over their avatar - this only displays until clicked and is then hidden for a period of time:
image

The, "Install on device", button is always available for Chromium based browsers:
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2219)


Nateowami (dismissed):
> Reviewed 4 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2219)* status: 4 of 6 files reviewed, 15 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html` line 95 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nkx3l4N1JVIu6eQYBkH:-Nkx3l4N1JVIu6eQYBkI:b-4hosi5) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L95)):*
> ```HTML
>           </button>
>           <button *ngIf="canInstallOnDevice$ | async" mat-menu-item (click)="installOnDevice()" class="install-button">
>             Install on device
> ```

Not internationalized?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html` line 96 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nkx9mpa79GB0wXRgFCg:-Nkx9mpa79GB0wXRgFCh:b-lt2ld6) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html#L96)):*
> ```HTML
>           <button *ngIf="canInstallOnDevice$ | async" mat-menu-item (click)="installOnDevice()" class="install-button">
>             Install on device
>             <mat-icon>install_mobile</mat-icon>
> ```

I really don't like the positioning of the icon up against the text.


Maybe pull it to the right and get rid of the 16px margin on the icon?

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 20 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nkx8ldwAKbg5d_D6Jfy:-Nkx8ldwAKbg5d_D6Jfz:btxylbw) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L20)):*
> ```SCSS
> :host {
>   display: flex;
>   flex-direction: column;
> ```

Why was this needed?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nkx7a_P63zXwRW4mAAb:-Nkx7a_P63zXwRW4mAAc:b-8o6e39) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L165)):*
> ```SCSS
> 
>   &.user-menu-btn {
>     mat-icon {
> ```

This is hard to read because you have to figure out what context we're in (the toolbar, in the header, in the app component), and that `.user-menu-btn` is the avatar icon in the top right, and then that the icon is actually the install badge.

It would be a lot easier to follow and less indented if you just wrote:
``` scss
.install-badge {
  // ...
}
```
... and didn't nest it inside anything else. It would also survive refactors a lot better.

Not sure how long it took me to figure out what you were targeting, but I was confused for a little while even while thinking about it (maybe 30 seconds).
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 167 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-NkxAEi63Ux1I5SmYxK0:-NkxAEi63Ux1I5SmYxK1:b-lr4wgg) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L167)):*
> ```SCSS
>     mat-icon {
>       position: absolute;
>       left: 3px;
> ```

Is `left: 3px;` failure to consider RTL, or is it bug-for-bug compatibility with the avatar component which does not shift the offline indicator (but which does in my WIP PR)?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 174 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-NkxAxKv01jRQTT1oCAm:-NkxAxKv01jRQTT1oCAn:blpsijk) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L174)):*
> ```SCSS
>       border-radius: 50%;
>       height: 18px;
>       width: 16px;
> ```

It doesn't look too bad as on oval?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 137 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nkx3tr_7avxf38ZBVLd:-Nkx3tr_7avxf38ZBVLe:b1sfyz9) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L137)):*
> ```XML
>   }
> 
>   get canShowInstallIconOnAvatar$(): Observable<boolean> {
> ```

"Can install" makes sense. "Can show" doesn't make much sense. It's more of "should show", but that can be simplified to just `show InstallIconOnAvatar$`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 138 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nkx43Oe9k8fXxV8uSrv:-Nkx43Oe9k8fXxV8uSrw:b-b0jcf5) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L138)):*
> ```XML
> 
>   get canShowInstallIconOnAvatar$(): Observable<boolean> {
>     return this.canInstallOnDevice$.pipe(filter(() => this.pwaService.getLastPromptSeen < Date.now()));
> ```

When I read "getLastPromptSeen", the word "get" says to me that this is a method, but there are no parentheses, so it looks to me like you've just referenced a function, rather than called it. Probably get's a getter? Getters generally don't start with "get" since their entire point is to look like properties.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 343 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nkx4T01E61g8ge0-hL4:-Nkx4T020-5rW2rJIJLC:bm6apgz) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L343)):*
> ```XML
> 
>   hideInstallIcon(): void {
>     this.localSettings.set(PWA_LAST_PROMPT_SEEN, Date.now() + 86400 * 7);
> ```

How about dismiss instead of "hide"?

I read this line as "write down that the last time the PWA prompt was seen was 7 days in the future."

But what you're actually doing is storing when you plan to next show the prompt. Storing when it was shown is a better idea and will be easier to reason about, and we can easily change how often we show it if we want.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts` line 14 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-NkxCMQYAkBQcIRbhNF7:-NkxCMQYAkBQcIRbhNF8:b-3ed6hj) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts#L14)):*
> ```XML
>   prompt: () => Promise<InstallPromptOutcome>;
> }
> export interface InstallPromptOutcome {
> ```

Is this interface trying to implement something that already exists in the browser, but we don't have types for, so we're specifying them ourselves? If so, I think it should say what it is, and link to some official documentation. APIs change, especially the ones that aren't super standard. Same for BeforeInstallPromptEvent.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts` line 56 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-NkxB9873MFSJ1rSaI-0:-NkxB9873MFSJ1rSaI-1:b-le5jww) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts#L56)):*
> ```XML
>   }
> 
>   get canInstall$(): BehaviorSubject<boolean> {
> ```

You've made this a getter, which follows a common pattern of exposing a BehaviorSubject with an API that does not allow the caller to modify the values inside it. But then you're exposing it as a BehaviorSubject, instead of as an Observable, which seems to defeat the purpose?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts` line 61 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-NkxBcIe7WsUoVfq4wD7:-NkxBcIe7WsUoVfq4wD8:bmodw1b) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts#L61)):*
> ```XML
> 
>   get getLastPromptSeen(): number {
>     return this.localSettings.get(PWA_LAST_PROMPT_SEEN) ?? 0;
> ```

How about `installPromptLastShownTime`?

`getLastPromptSeen` looks like a method name for something that might been a boolean, or the prompt itself, or something.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts` line 71 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nkx644L1wumPn-dChd_:-Nkx644L1wumPn-dChda:b-2zknqi) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts#L71)):*
> ```XML
> 
>   get isRunningInstalledApp(): boolean {
>     return window.matchMedia('(display-mode: standalone)').matches;
> ```

Is this only supported in Chromium as well? If so, there should be a comment. If not, it might still deserve a comment since some other things are not supported.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts` line 81 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nkx6wOSA7lWN6AoC97E:-Nkx6wOSA7lWN6AoC97F:b61cd7i) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts#L81)):*
> ```XML
>   install(): void {
>     if (this.promptEvent != null) {
>       this.promptEvent?.prompt().then((result: InstallPromptOutcome) => {
> ```

How about using `await`?
``` ts
  async install(): Promise<void> {
    if (this.promptEvent != null) {
      const result: InstallPromptOutcome = await this.promptEvent.prompt();
      if (result.outcome === 'accepted') {
        this.canInstall$.next(!this.isRunningInstalledApp);
      }
    }
  }
```
If you for some reason need to not make this async, at least drop the double null check on `this.promptEvent`.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts` line 83 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nkx7416-UJCc59694BA:-Nkx7416-UJCc59694BB:bdzcvwu) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts#L83)):*
> ```XML
>       this.promptEvent?.prompt().then((result: InstallPromptOutcome) => {
>         if (result.outcome === 'accepted') {
>           this.canInstall$.next(!this.isRunningInstalledApp);
> ```

What if it's installed, but the user opened it in the browser? Are we properly handling that? This line to me reads like it would prompt the user to install the app if the user already installed it. That might not be correct, but it appears that way.

Nateowami (commented):
> Reviewed 1 of 6 files at r1, 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2219)* status: all files reviewed, 3 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nkx7a_P63zXwRW4mAAb:-Nl0K3ix4biT0EIPrUHY:b-ip9b31) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L165)):*

I've renamed. As you know, specificity is very important to CSS which is always my default. If someone is only intended to be used on a specific place then it makes sense to have it there rather than higher up which may unintendedly get applied to something else on the page.
</blockquote></details>

Given styles are scoped to the component, how likely is it that there would be more than one element with `.install-badge`? To me it seems very unique.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 174 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-NkxAxKv01jRQTT1oCAm:-Nl0KJON1lK2ftU6eYCH:b-z41r85) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L174)):*

Have gone back to a circle - it was getting a little fiddley with how different displays would render the white space around the icon. How does it look on your screen now?
</blockquote></details>

It wasn't really bad, but I found it a little odd.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 127 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nl0C6N41_H3VW5Jfbhy:-Nl0C6N41_H3VW5Jfbhz:b-g11w40) ([raw file](https://github.com/sillsdev/web-xforge/blob/3eee50a016f30ec4d54e07d00b326de0784a0312/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L127)):*
> ```SCSS
>     .install-button {
>       display: flex;
>       column-gap: 16px;
> ```

Is there a reason you don't just do `justify-content: space-between;` to right-align the icon? To me that makes more sense than picking an arbitrary offset.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts` line 14 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-NkxCMQYAkBQcIRbhNF7:-Nl0KfcT0SOHuOsbuibf:b-gxxkwa) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts#L14)):*

Correct. I've added some comments and also removed `platform` for now as I didn't end up supporting multiple platforms.
</blockquote></details>

Thanks. Appreciate the comments. They make clear what these interfaces are supposed to match.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts` line 71 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nkx644L1wumPn-dChd_:-Nl09Q-w3UDKCT39LcSJ:b-uv4ta) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts#L71)):*

This one actually works fine across all browsers and devices. I've added a comment.
</blockquote></details>

Thanks for the clarification.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts` line 83 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nkx7416-UJCc59694BA:-Nl0KxRA6gwqfvT-Yu4j:b-uv4ta) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.ts#L83)):*

If it is installed, and the user opened their browser instead of the installed app, then the `beforeinstallprompt` event won't fire and the default `canInstall$` value is `false` so no prompts or buttons will appear. I could have simply set the next value to `false` but I used `isRunningInstalledApp` again as a sanity check that it installed correctly.
</blockquote></details>

Thanks for the clarification.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.spec.ts` line 47 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nl0MGDk1f1YUaUouUJ_:-Nl0MGDk1f1YUaUouUJa:b-vwajlc) ([raw file](https://github.com/sillsdev/web-xforge/blob/3eee50a016f30ec4d54e07d00b326de0784a0312/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.spec.ts#L47)):*
> ```XML
>     });
>     expect(canInstall).toEqual(false);
>     window.dispatchEvent(new Event('beforeinstallprompt'));
> ```

Directly mocking the globals has serious disadvantages:
1. Tests aren't necessarily isolated (they might leave behind state that impact other tests)
2. You're at the mercy of the browser, and it allowing you to fake events and calls.

This is a perfect use case for dependency injection. If you inject the window (see example in `src/SIL.XForge.Scripture/ClientApp/src/xforge-common/route.directive.ts`), then you can mock things instead of mutating the global object.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2219)* status: 1 of 7 files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 165 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nkx7a_P63zXwRW4mAAb:-Nl0wpex5RoHppILb-5M:b-rx0zcw) ([raw file](https://github.com/sillsdev/web-xforge/blob/571f82706e3147fa9f87020cf1c922d1a41688cb/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L165)):*

Given styles are scoped to the component, how likely is it that there would be more than one element with `.install-badge`? To me it seems very unique.
</blockquote></details>

I agree that at the moment it would be slim. I could imagine a different install prompt being introduced in the future perhaps in the name of gamification. Besides specificity, I also like the context that nesting provides especially when some components start to grow.
I'll move it if you really want it moved but the above is my preference anyway :)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 127 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nl0C6N41_H3VW5Jfbhy:-Nl0wBf76lLfw93mlL1h:blerccb) ([raw file](https://github.com/sillsdev/web-xforge/blob/3eee50a016f30ec4d54e07d00b326de0784a0312/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L127)):*

Is there a reason you don't just do `justify-content: space-between;` to right-align the icon? To me that makes more sense than picking an arbitrary offset.
</blockquote></details>

`justify-content` works well if content around it is larger. As this button is typically the longest then `justify-content` visually does nothing which is why I went with ensuring there was a consistent gap between the elements.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.spec.ts` line 47 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nl0MGDk1f1YUaUouUJ_:-NlW2jO83PP5bimFRERc:b-qiszwq) ([raw file](https://github.com/sillsdev/web-xforge/blob/3eee50a016f30ec4d54e07d00b326de0784a0312/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.spec.ts#L47)):*

Directly mocking the globals has serious disadvantages:
1. Tests aren't necessarily isolated (they might leave behind state that impact other tests)
2. You're at the mercy of the browser, and it allowing you to fake events and calls.

This is a perfect use case for dependency injection. If you inject the window (see example in `src/SIL.XForge.Scripture/ClientApp/src/xforge-common/route.directive.ts`), then you can mock things instead of mutating the global object.
</blockquote></details>

Rabbit hole code complete! Let me know if you think I've taken a wrong turn with how I implemented this. It works. What I found is when we bring in `WINDOW` it has zero methods or properties available so we can't use ts-mockito like we normally would and required me to recreate all the relevant methods.

Nateowami (commented):
> Reviewed 5 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2219)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 127 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nl0C6N41_H3VW5Jfbhy:-Nla-z7X2F2cH95zPcIo:b-s7hldt) ([raw file](https://github.com/sillsdev/web-xforge/blob/3eee50a016f30ec4d54e07d00b326de0784a0312/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L127)):*

`justify-content` works well if content around it is larger. As this button is typically the longest then `justify-content` visually does nothing which is why I went with ensuring there was a consistent gap between the elements.
</blockquote></details>

I see what you mean now. On my machine it's not the largest thing, which is why I felt it looked a bit odd without `justify-content`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.spec.ts` line 47 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nl0MGDk1f1YUaUouUJ_:-Nl_vlyG6gEg6StsBGCJ:b-l0g5bz) ([raw file](https://github.com/sillsdev/web-xforge/blob/3eee50a016f30ec4d54e07d00b326de0784a0312/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.spec.ts#L47)):*

Rabbit hole code complete! Let me know if you think I've taken a wrong turn with how I implemented this. It works. What I found is when we bring in `WINDOW` it has zero methods or properties available so we can't use ts-mockito like we normally would and required me to recreate all the relevant methods.
</blockquote></details>

Wow, I didn't intend to send you on something this challenging. Are you sure it wouldn't be possible to mock it in the same way we mock the document in the following files?

- `text-chooser-dialog.component.spec.ts`
- `i18n.service.spec.ts`

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2219)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.spec.ts` line 47 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-Nl0MGDk1f1YUaUouUJ_:-Nluof6PCfylW40BJNot:bcnnz9a) ([raw file](https://github.com/sillsdev/web-xforge/blob/3eee50a016f30ec4d54e07d00b326de0784a0312/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/pwa.service.spec.ts#L47)):*

Wow, I didn't intend to send you on something this challenging. Are you sure it wouldn't be possible to mock it in the same way we mock the document in the following files?

- `text-chooser-dialog.component.spec.ts`
- `i18n.service.spec.ts`
</blockquote></details>

I've tried a number of different things and always get the same result `mockedWindow.matchMedia is not a function` along with similar errors for the other methods I had to implement.

nigel-wells:
Any further feedback on this one @Nateowami ?

nigel-wells:
@josephmyers are you able to take a look over this as well when you're free.

josephmyers (commented):
Since Nathaniel's already done a code review, I just did basic testing. Overall, this work looks really good. Nice job. The install badge is small, but I think it works really well in leading users to the (larger) menu option. I didn't test how long it takes to be prompted again after the first time.

The only issue I did notice is that installing from the browser three-dot doesn't remove the option from the avatar menu, until you refresh. The install button in the avatar menu just doesn't do anything. The opposite case works as expected. It's minor, but we should fix it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-:-NqW3gne6C_AFJXv2mm5:b24hqmg)* status: 1 of 7 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

nigel-wells (commented):
AH, well spotted!

I've changed the logic to listen to a different event which will trigger from either place now.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-:-Nqjdac61v2q8rNoAUOK:b-x3dc06)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

josephmyers (approved):
Very good, thank you. Let's send to test!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-:-Nqzk5X7087OA9BinBhT:b-eev3vg)* status: 0 of 7 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

nigel-wells (commented):
@Nateowami This is ready to merge if you're happy to give your approval.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2219#-:-NrXs10_0aUmkkn9kB4T:bs6nujg)* status: 0 of 7 files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2221, opened 2023-12-07
SF-1916 Allow admins to send a general share link when sharing disabled

RaymondLuong3:
Admins were originally restricted from sharing general links where anyone with the link can join unless sharing was enabled. This change allows admin users to share 'anyone' links while keeping a project un-shareable by non-admin users.

Unfortunately, users who received a link to the project can still pass that link on to other users, so being un-shareable is not accurate. A follow-up for this would be to allow admins to revoke a share link. Maybe this change should not be submitted before that is complete. I'm not sure what the security concerns may be in the real world.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2221)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.spec.ts
@@ -204,15 +204,15 @@ describe('ShareDialogComponent', () => {
     expect(roles).toContain(SFProjectRole.Commenter);
   }));
 
-  it('admin users can not share with anyone if sharing is disabled', fakeAsync(() => {
-    env = new TestEnvironment({ userId: TestUsers.Admin, checkingEnabled: false, translateShareEnabled: false });
-    expect(env.component.shareRole).toEqual(SF_DEFAULT_TRANSLATE_SHARE_ROLE);
-    expect(env.canChangeLinkUsage).toBeFalse();
+  it('admin users can share with anyone if sharing is disabled', fakeAsync(() => {
+    env = new TestEnvironment({ userId: TestUsers.Admin, checkingShareEnabled: false, translateShareEnabled: false });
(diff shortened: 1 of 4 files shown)


Nateowami (commented):
I'm pretty sure you asked me something about this recently, and I didn't get back with you.

I don't know what a good solution to this problem is. The rationale for the current behavior is sound, and I don't know what how we can improve it. I think the solution may not be to change the requirements for creating such a link, but find a way to help the user discover that that's what they need to do to be able to share a general invitation link.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2221#-:-Nl0i8495MtKfC1nVfgR:b9vbvbr)* status: 0 of 4 files reviewed, all discussions resolved

nigel-wells:
What if we put an expiry date (7 days?) on anyone links that are shared when sharing is disabled? The dialog is clear that the link can be used by anyone so I don't think we need to worry about someone further sharing it. If admin has concerns then don't share a link with anyone.

RaymondLuong3 (commented):
The expiration idea might really be a good option. Currently a link will persist indefinitely unless the admin disables 'share with anyone'. If we make the link expire in 7 days, we would have to rework the back-end to produce a new link if the a user requests it and it has expired. Not a big task, but not trivial. We can come back to this topic in the new year.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2221#-:-NlZdxfL7uxY8WaYFiRM:bu3xn3n)* status: 0 of 4 files reviewed, all discussions resolved

marksvc:
I as looking for something to review, and this was the oldest. Have there been any thoughts regarding direction for this task?

nigel-wells:
We actually discussed this one yesterday and a new Jira card is being created. Is that right @Nateowami or is SF-1916 going to be updated with the new requirements? Either way, this PR should probably be closed as the requirements are quite different now.

Nateowami:
Closing in favor of #2527

PR #2225, opened 2023-12-07
SF-2399 Sort books in community checking book list

RaymondLuong3:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2225)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
@@ -205,6 +205,13 @@ describe('CheckingComponent', () => {
       expect(nextQuestion).toEqual(1);
     }));
 
+    it('should display books in canonical order', fakeAsync(() => {
+      const env = new TestEnvironment({ user: ADMIN_USER });
+      env.waitForSliderUpdate();
+      expect(env.component.books).toEqual([40, 43]);
+      discardPeriodicTasks();
+    }));
+
     it('should re-calculate scripture slide position on resize', fakeAsync(() => {
       const testProject: SFProject = TestEnvironment.generateTestProject();
       testProject.checkingConfig.hideCommunityCheckingText = true;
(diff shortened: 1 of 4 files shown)


Nateowami (changes requested):
The BookChapterChooserComponent is used in three components. This fixes the problem in one of them. If the sorting happened in the component that actually lists them for the user, the problem would be solved for all these components, and any future components that use it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2225#-:-Nl52hlX3BvTnUEiJESX:b9zb9se)* status: 0 of 2 files reviewed, all discussions resolved

RaymondLuong3 (commented):
I actually just went down that route, and then discovered that each component that uses the book chapter select also requires their cache of books to be in order. So the change needs to be there. I added the logic inside the book chapter select also.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2225#-:-Nl5P57J1-Jeb_p12EcY:bs72zvx)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami (approved):
:lgtm:

> Reviewed 2 of 2 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2225#-:-NmD2VO3EGpmOUZke3Mn:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts` line 29 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2225#-NmD214fDrCRI9HT7zJB:-Nl9_oDi4Z_uX1KNTZbB:b-sebju8) ([raw file](https://github.com/sillsdev/web-xforge/blob/9aa9cf0af9dc3aaf99de02623d63b4b597bd53c9/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-chapter-chooser/book-chapter-chooser.component.ts#L29)):*
> ```XML
> 
>   @Input() set books(value: number[]) {
>     this._books = Object.values(value).sort((a, b) => a - b);
> ```

Good job recognizing that we don't want to mutate the original array. Newer browsers have `toSorted()` which doesn't mutate the original array, but unfortunately not all of the browsers we want to target.

PR #2226, opened 2023-12-07
Refer to draft generation instead of pre-translation

Nateowami:
"Generating drafts" may not be the most clear to users, but I think it's a lot clearer than pre-translation, and it's consistent with how we talk about it elsewhere, which will help with comprehension.

I also converted a couple related settings to sentence case. I know there are others on this page that could be adjusted as well, but I'm trying to limit the scope to just this group of settings.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2226)


RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2226)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2231, opened 2023-12-11
SF-2309 Support for combined verses in the target text

pmachapman:
This PR adds support for combined verse in the target text that are separate in the source text.

In this case, the verses that are in the draft that are separated, are placed into the combined verse in the target.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2231)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.spec.ts
@@ -77,5 +77,21 @@ describe('DraftViewerService', () => {
       ];
       expect(service.toDraftOps(draft, targetOps)).toEqual(expectedResult);
     });
+
+    it('should allow combined verses in the target that are separated in the source', () => {
+      const draft: DraftSegmentMap = {
+        verse_1_1: 'In the beginning ',
+        verse_1_2: 'The earth was formless',
+        verse_1_3: 'Then God said'
+      };
+      const targetOps: DeltaOperation[] = [{ insert: '', attributes: { segment: 'verse_1_1-3' } }];
+      const expectedResult: DeltaOperation[] = [
+        {
+          insert: 'In the beginning The earth was formless Then God said',
+          attributes: { segment: 'verse_1_1-3', draft: true }
+        }
+      ];
+      expect(service.toDraftOps(draft, targetOps)).toEqual(expectedResult);
+    });
   });
 });
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.spec.ts
@@ -139,5 +139,20 @@ describe('DraftViewerService', () => {
       ];
       expect(service.toDraftOps(draft, targetOps)).toEqual(expectedResult);
     });
+
+    it('should allow combined verses in the target that only one exists in the source', () => {
+      const draft: DraftSegmentMap = {
+        verse_150_1: 'Praise ye the Lord. ',
+        verse_150_2: 'Praise him for his mighty acts: '
+      };
+      const targetOps: DeltaOperation[] = [{ insert: '', attributes: { segment: 'verse_150_1-3' } }];
+      const expectedResult: DeltaOperation[] = [
+        {
+          insert: 'Praise ye the Lord. Praise him for his mighty acts: ',
+          attributes: { segment: 'verse_150_1-3', draft: true }
+        }
+      ];
+      expect(service.toDraftOps(draft, targetOps)).toEqual(expectedResult);
+    });
   });
 });
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.spec.ts
@@ -80,15 +80,61 @@ describe('DraftViewerService', () => {
 
     it('should allow combined verses in the target that are separated in the source', () => {
       const draft: DraftSegmentMap = {
-        verse_1_1: 'In the beginning ',
-        verse_1_2: 'The earth was formless',
-        verse_1_3: 'Then God said'
+        verse_150_1: 'Praise ye the Lord. ',
+        verse_150_2: 'Praise him for his mighty acts: ',
+        verse_150_3: 'Praise him with the sound of the trumpet: '
       };
-      const targetOps: DeltaOperation[] = [{ insert: '', attributes: { segment: 'verse_1_1-3' } }];
+      const targetOps: DeltaOperation[] = [{ insert: '', attributes: { segment: 'verse_150_1-3' } }];
       const expectedResult: DeltaOperation[] = [
         {
-          insert: 'In the beginning The earth was formless Then God said',
-          attributes: { segment: 'verse_1_1-3', draft: true }
+          insert: 'Praise ye the Lord. Praise him for his mighty acts: Praise him with the sound of the trumpet: ',
+          attributes: { segment: 'verse_150_1-3', draft: true }
+        }
+      ];
+      expect(service.toDraftOps(draft, targetOps)).toEqual(expectedResult);
+    });
+
+    it('should allow combined verses in the target that are combined in the source', () => {
+      const draft: DraftSegmentMap = {
+        'verse_150_1-3':
+          'Praise ye the Lord. Praise him for his mighty acts: Praise him with the sound of the trumpet: '
+      };
+      const targetOps: DeltaOperation[] = [{ insert: '', attributes: { segment: 'verse_150_1-3' } }];
+      const expectedResult: DeltaOperation[] = [
+        {
+          insert: 'Praise ye the Lord. Praise him for his mighty acts: Praise him with the sound of the trumpet: ',
+          attributes: { segment: 'verse_150_1-3', draft: true }
+        }
+      ];
+      expect(service.toDraftOps(draft, targetOps)).toEqual(expectedResult);
+    });
+
+    it('should allow partially combined verses in the target that are combined in the source', () => {
+      const draft: DraftSegmentMap = {
+        'verse_150_1-3':
+          'Praise ye the Lord. Praise him for his mighty acts: Praise him with the sound of the trumpet: '
+      };
+      const targetOps: DeltaOperation[] = [{ insert: '', attributes: { segment: 'verse_150_1-2' } }];
+      const expectedResult: DeltaOperation[] = [
+        {
+          insert: 'Praise ye the Lord. Praise him for his mighty acts: Praise him with the sound of the trumpet: ',
+          attributes: { segment: 'verse_150_1-2', draft: true }
+        }
+      ];
+      expect(service.toDraftOps(draft, targetOps)).toEqual(expectedResult);
+    });
+
+    it('should allow partially combined verses in the target that are combined in the source', () => {
+      const draft: DraftSegmentMap = {
+        'verse_150_1-2': 'Praise ye the Lord. Praise him for his mighty acts: ',
+        // Known issue: This verse will not be merged into 1-3 in the target
+        verse_150_1_3: 'Praise him with the sound of the trumpet: '
+      };
+      const targetOps: DeltaOperation[] = [{ insert: '', attributes: { segment: 'verse_150_1-3' } }];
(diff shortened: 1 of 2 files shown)


RaymondLuong3 (changes requested):
Thanks for adding the new functionality. The logic is starting to look clunky and could probably use a refactor. I left some comments here.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2231#-:-NlVKcrJ7V7ys8ypo_Yx:b90u3q4)* status: all files reviewed, 5 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 51 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2231#-NlVHEDa8BFzwnVnxk3P:-NlVHEDa8BFzwnVnxk3Q:b-z9n1pr) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfb4604b15351b6cd3f9b56be1eb0a41321e12e4/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L51)):*
> ```XML
>     }
> 
>     return targetOps.map(op => {
> ```

This may be outside the scope of this PR, but say you have target reference `verse_1_1-3` and a draft reference of `verse_1_1-2`, you would expect that the draft reference would be included in `verse_1_1-3`, but it would not with the current logic. It might be enough just to leave a comment for this function of this limitation, instead of try to support it (IMO we should map draft to target references when it is obvious, but it is OK if we drop one here and there)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 60 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2231#-NlVIjOQBlQBoSFzZ2sF:-NlVIjOQBlQBoSFzZ2sG:b8a14bh) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfb4604b15351b6cd3f9b56be1eb0a41321e12e4/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L60)):*
> ```XML
>       if (!isSegmentDraftAvailable) {
>         // See if the source verse is combined
>         const combinedVerseNumbers: string[] = Object.keys(draft).filter(key => key.startsWith(segmentRef + '-'));
> ```

If you wanted to solve the problem above, you can replace this line like so.



_Code snippet:_
```XML
        const combinedVerseNumbers: string[] = Object.keys(draft).filter(key => key.split('-')[0] === segmentRef.split('-')[0]);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 65 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2231#-NlVA7m1-BhFTEglB9rQ:-NlVA7m1-BhFTEglB9rR:b8if27d) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfb4604b15351b6cd3f9b56be1eb0a41321e12e4/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L65)):*
> ```XML
>           draftSegmentText = draft[combinedVerseNumbers[0]];
>           isSegmentDraftAvailable = draftSegmentText != null && draftSegmentText.trim().length > 0;
>         } else if (segmentRef.startsWith('verse_') && segmentRef.indexOf('-') > -1) {
> ```

You can use `VERSE_REGEX.test(segmentRef)` here.



_Code quote:_
```XML
segmentRef.startsWith('verse_') 
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2231#-NlVBLtu9uX6sv4rxSj_:-NlVBLtu9uX6sv4rxSja:b-e35qgg) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfb4604b15351b6cd3f9b56be1eb0a41321e12e4/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L68)):*
> ```XML
>           // Otherwise, if the target verse is combined
>           // Get the book number from the segment ref
>           const bookNum: number = parseInt(segmentRef.split('_')[1]);
> ```

Unfortunately, this is not the book num, it is the chapter number. In this case, it works because the book num does not play a part in calculating the verses here. Maybe you can create a new method called `getSegmentsInRange(segmentRef)` for `utils.ts` which provides all the segmentRefs for a range. So the input could be `verse_1_1-3` and the output is `verse_1_1, verse_1_2, verse_1_3` and you could put the logic that you have here in that function.



_Code quote:_
```XML
          const bookNum: number = parseInt(segmentRef.split('_')[1]);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 73 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2231#-NlVB-I8CllNy3NxF_Ww:-NlVB-I8CllNy3NxF_Wx:b-ycqw7k) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfb4604b15351b6cd3f9b56be1eb0a41321e12e4/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L73)):*
> ```XML
>           if (segmentVerseRef != null) {
>             // Add the drafts for all of the verses in the segment
>             for (var verseRef of segmentVerseRef?.allVerses()) {
> ```

Then you could go `for (const ref of segmentRefsInRange)`

pmachapman:
My correcting the branch name had the unintended consequence of closing this PR. I have recreated the PR, and updated to match your feedback.

pmachapman (commented):
I agree with the growing clunk. There will be a major refactor next quarter when all of the code is reworked to accept USFM directly from Serval.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2231#-:-Nl_zBlb-p_DLcnQJtZO:bbw8emz)* status: 0 of 2 files reviewed, 4 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 51 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2231#-NlVHEDa8BFzwnVnxk3P:-Nl_wZyf1joSH6FBJLwR:bs7385y) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfb4604b15351b6cd3f9b56be1eb0a41321e12e4/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L51)):*

This may be outside the scope of this PR, but say you have target reference `verse_1_1-3` and a draft reference of `verse_1_1-2`, you would expect that the draft reference would be included in `verse_1_1-3`, but it would not with the current logic. It might be enough just to leave a comment for this function of this limitation, instead of try to support it (IMO we should map draft to target references when it is obvious, but it is OK if we drop one here and there)
</blockquote></details>

Good idea have added your suggestion below and documented with a comment.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 60 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2231#-NlVIjOQBlQBoSFzZ2sF:-Nl_whAJ3_AihiDyC9Un:bo04yz5) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfb4604b15351b6cd3f9b56be1eb0a41321e12e4/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L60)):*

If you wanted to solve the problem above, you can replace this line like so.
</blockquote></details>

Done. I've had to tweak it a little bit to work, and added unit tests to cover the valid scenarios.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 65 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2231#-NlVA7m1-BhFTEglB9rQ:-Nl_xC1C81sWuHI3EJlx:b-tmxo3o) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfb4604b15351b6cd3f9b56be1eb0a41321e12e4/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L65)):*

You can use `VERSE_REGEX.test(segmentRef)` here.
</blockquote></details>

`VERSE_REGEX` only supports verses that are numbers - it doesn't support combined verses or partial verses. I am not sure on how it is used in `text.component.ts` to modify it without unintended consequences?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 68 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2231#-NlVBLtu9uX6sv4rxSj_:-Nl_xSno7wtDnm36gsNV:b-2f412v) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfb4604b15351b6cd3f9b56be1eb0a41321e12e4/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L68)):*

Unfortunately, this is not the book num, it is the chapter number. In this case, it works because the book num does not play a part in calculating the verses here. Maybe you can create a new method called `getSegmentsInRange(segmentRef)` for `utils.ts` which provides all the segmentRefs for a range. So the input could be `verse_1_1-3` and the output is `verse_1_1, verse_1_2, verse_1_3` and you could put the logic that you have here in that function.
</blockquote></details>

Thank you for picking that up. I really don't know why I thought it did!

I have updated it to feed in the book number of Genesis as we do not use the book numbers from the resulting verserefs.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 73 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2231#-NlVB-I8CllNy3NxF_Ww:-Nl_z8Xr3J7nm3cR1WsY:b-q64pkt) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfb4604b15351b6cd3f9b56be1eb0a41321e12e4/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L73)):*

Then you could go `for (const ref of segmentRefsInRange)`
</blockquote></details>

See above.

RaymondLuong3 (approved):
That is good to know.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2231#-:-NldeeN7Bzv5PmJlhiCF:b6jh6gv)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 65 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2231#-NlVA7m1-BhFTEglB9rQ:-NldbJZKDI4T9NJJ5ZWe:bumz05g) ([raw file](https://github.com/sillsdev/web-xforge/blob/dfb4604b15351b6cd3f9b56be1eb0a41321e12e4/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L65)):*

`VERSE_REGEX` only supports verses that are numbers - it doesn't support combined verses or partial verses. I am not sure on how it is used in `text.component.ts` to modify it without unintended consequences?
</blockquote></details>

Hmm, that is a good point. Let's leave this as is.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.spec.ts` line 131 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2231#-NldeJ6-CJVL8FpZJMau:-NldeJ6-CJVL8FpZJMav:ba8bugm) ([raw file](https://github.com/sillsdev/web-xforge/blob/b416929e8891929a248a035bf298fb8639630beb/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.spec.ts#L131)):*
> ```XML
>         'verse_150_1-2': 'Praise ye the Lord. Praise him for his mighty acts: ',
>         // Known issue: This verse will not be merged into 1-3 in the target
>         verse_150_1_3: 'Praise him with the sound of the trumpet: '
> ```

Yes, the never-ending number of different ways that verses can be structures. I'm hopeful that we can make it work smart enough to cover 99% of cases. Thanks for your work on this.

PR #2234, opened 2023-12-12
SF-2390 Allow multiple roles per user

pmachapman:
This Pull Request changes the current "one role per user" system to allow multiple roles per user.

This PR does not add any additional roles, but lays the foundation for this to be done in a completely backwards compatible manner. Backwards compatibility is required to stop the breaking of any current tokens in use, and because if Auth0 was updated to always return the Role claim as an array, it would break all other developer environments not running this branch.

**Adding multiple roles to a user**

Although this will be able to be done in the future using the Auth0 API when system wide roles need to be edited for many users, currently this must be done by editing the `app_metadata` for the user in the Auth0 dashboard.

To test multiple roles for a user:

1. Log into the Auth0 Dashboard
2. Click User Management -> Users
3. Click on the user you wish to edit
4. Change the `app_metadata` from:
```json
{
  "xf_role": "user",
  "xf_user_id": "your_user_id_here"
}
```
to:
```json
{
  "xf_role": ["system_admin", "user"],
  "xf_user_id": "your_user_id_here"
}
```
5. Click Save.
6. Log out then in to Scripture Forge to use the new roles

***Note:** When you are finished testing this branch, be sure to log off, and revert the change made to `app_metadata` in Auth0.*

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2234)

### DIFF
--- a/src/RealtimeServer/common/connect-session.ts
+++ b/src/RealtimeServer/common/connect-session.ts
@@ -3,6 +3,6 @@
  */
 export interface ConnectSession {
   userId: string;
-  role?: string;
+  roles: string[];
   isServer: boolean;
 }
--- a/src/RealtimeServer/common/models/user-test-data.ts
+++ b/src/RealtimeServer/common/models/user-test-data.ts
@@ -15,7 +15,7 @@ function testUser(ordinal: number): User {
     ...testUserProfile(ordinal),
     name: `Name of test user ${ordinal}`,
     email: `user${ordinal}@example.com`,
-    role: SystemRole.User,
+    roles: [SystemRole.User],
     isDisplayNameConfirmed: false,
     interfaceLanguage: 'en',
     authId: `authId${ordinal}`,
--- a/src/RealtimeServer/common/models/user.ts
+++ b/src/RealtimeServer/common/models/user.ts
@@ -48,7 +48,7 @@ export interface User extends UserProfile {
   name: string;
   email: string;
   paratextId?: string;
-  role: string;
+  roles: string[];
   isDisplayNameConfirmed: boolean;
   interfaceLanguage?: string;
   authId: string;
--- a/src/RealtimeServer/common/realtime-server.ts
+++ b/src/RealtimeServer/common/realtime-server.ts
@@ -475,11 +475,11 @@ export class RealtimeServer extends ShareDB {
   private async setConnectSession(context: ShareDB.middleware.ConnectContext): Promise<void> {
     let session: ConnectSession;
     if (context.req != null && context.req.user != null) {
-      const userId = context.req.user[XF_USER_ID_CLAIM];
-      const role = context.req.user[XF_ROLE_CLAIM];
+      const userId: string = context.req.user[XF_USER_ID_CLAIM];
+      const roles: string[] = [context.req.user[XF_ROLE_CLAIM]];
       session = {
         userId,
-        role,
+        roles,
         isServer: false
       };
     } else {
@@ -487,7 +487,7 @@ export class RealtimeServer extends ShareDB {
       if (context.req != null && context.req.userId != null) {
         userId = context.req.userId;
       }
-      session = { isServer: true, userId };
+      session = { isServer: true, userId, roles: [] };
     }
     context.agent.connectSession = session;
   }
--- a/src/RealtimeServer/common/realtime-server.ts
+++ b/src/RealtimeServer/common/realtime-server.ts
@@ -476,7 +476,8 @@ export class RealtimeServer extends ShareDB {
     let session: ConnectSession;
     if (context.req != null && context.req.user != null) {
       const userId: string = context.req.user[XF_USER_ID_CLAIM];
-      const roles: string[] = [context.req.user[XF_ROLE_CLAIM]];
+      const role: string | undefined = context.req.user[XF_ROLE_CLAIM];
+      const roles: string[] = role?.split(',') ?? [];
       session = {
         userId,
         roles,
--- a/src/RealtimeServer/common/realtime-server.ts
+++ b/src/RealtimeServer/common/realtime-server.ts
@@ -476,8 +476,8 @@ export class RealtimeServer extends ShareDB {
     let session: ConnectSession;
     if (context.req != null && context.req.user != null) {
       const userId: string = context.req.user[XF_USER_ID_CLAIM];
-      const role: string | undefined = context.req.user[XF_ROLE_CLAIM];
-      const roles: string[] = role?.split(',') ?? [];
+      const role: string | string[] | undefined = context.req.user[XF_ROLE_CLAIM];
+      const roles: string[] = typeof role === 'string' ? [role] : role || [];
       session = {
         userId,
         roles,
--- a/src/RealtimeServer/common/services/project-service.spec.ts
+++ b/src/RealtimeServer/common/services/project-service.spec.ts
@@ -123,7 +123,7 @@ class TestEnvironment {
       'systemAdmin',
       createTestUser(
         {
-          role: SystemRole.SystemAdmin
+          roles: [SystemRole.SystemAdmin]
         },
         1
       )
--- a/src/RealtimeServer/common/services/project-service.ts
+++ b/src/RealtimeServer/common/services/project-service.ts
@@ -53,7 +53,7 @@ export abstract class ProjectService<T extends Project = Project> extends JsonDo
   };
 
   protected allowRead(_docId: string, doc: T, session: ConnectSession): boolean {
-    if (session.isServer || session.role === SystemRole.SystemAdmin || Object.keys(doc).length === 0) {
+    if (session.isServer || session.roles.includes(SystemRole.SystemAdmin) || Object.keys(doc).length === 0) {
       return true;
     }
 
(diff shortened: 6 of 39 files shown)


RaymondLuong3 (changes requested):
When I look at the issue SF-2390, I was assuming that we would create a new System Role where a user can specifically be a "Serval Technician" and they would have all the privileges of a User, but they can be added to projects with a SF specific technician role with the ability to synchronize projects and modify pre-translation configs.

But then the issue specifics "A SF specific roles system". I don't even understand what that means. Does that mean a roles system that is specific to SF from Auth0's point of view? Anyway, I'm sure you can clarify so that I understand what this PR is trying to accomplish.

With that said, how does having multiple system roles help to support serval? Maybe it does so by allowing a user to be both a `SystemRole.User` and a `SystemRole.ServalTech`?

> Reviewed 39 of 39 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2234#-:-NnkT5pP02rEZLYHIzAn:b-i1kgmw)* status: all files reviewed, 3 unresolved discussions (waiting on @pmachapman)

___
*[`src/RealtimeServer/common/models/user.ts` line 51 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2234#-NnkEnwK-EnVb5ayQLa5:-NnkEnwK-EnVb5ayQLa6:brnn1y5) ([raw file](https://github.com/sillsdev/web-xforge/blob/77dff84322e347c723d4140481952be3fb8dcc08/src/RealtimeServer/common/models/user.ts#L51)):*
> ```XML
>   email: string;
>   paratextId?: string;
>   roles: string[];
> ```

Why do we need to change the role property. This indicates whether a user is a system admin, or just a regular user. Since system admins can do anything a user can do, there is no reason for them to have the normal user role.



_Code quote:_
```XML
  roles: string[];
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 658 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2234#-NnkMCWu9h5qcbPD5dWB:-NnkMCWu9h5qcbPD5dWC:b-ofq3tt) ([raw file](https://github.com/sillsdev/web-xforge/blob/77dff84322e347c723d4140481952be3fb8dcc08/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L658)):*
> ```XML
>     this.localSettings.set(USER_ID_SETTING, userId);
>     this.localSettings.remove(ROLE_SETTING);
>     this.localSettings.set(ROLES_SETTING, typeof role === 'string' ? [role] : role || []);
> ```

Could we not assume that the role would be an array at this point. Since we would only set `ROLES_SETTING` only if the current code has been deployed? I do see that the claim will return a `string | string[]` so it appears we do need to account for the role being a string.
___
*[`src/SIL.XForge.Scripture/Controllers/MachineApiController.cs` line 2 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2234#-NnkNSvd1RbCUyJd_HmB:-NnkNSvd1RbCUyJd_HmC:br7pz1m) ([raw file](https://github.com/sillsdev/web-xforge/blob/77dff84322e347c723d4140481952be3fb8dcc08/src/SIL.XForge.Scripture/Controllers/MachineApiController.cs#L2)):*
> ```Smalltalk
> using System;
> using System.Linq;
> ```

Nit: looks like linq is not being used.



_Code quote:_
```Smalltalk
using System.Linq;
```

pmachapman (commented):
> When I look at the issue [SF-2390](https://jira.sil.org/browse/SF-2390), I was assuming that we would create a new System Role where a user can specifically be a "Serval Technician" and they would have all the privileges of a User, but they can be added to projects with a SF specific technician role with the ability to synchronize projects and modify pre-translation configs.

That will come in the next PR. To make this PR clearer, I made this just the plumbing changes required to have multiple roles per user. The next PR will have the new role defined and permission checks where the role is used, and be a much simpler PR.

> But then the issue specifics "A SF specific roles system". I don't even understand what that means. Does that mean a roles system that is specific to SF from Auth0's point of view? Anyway, I'm sure you can clarify so that I understand what this PR is trying to accomplish.

I have renamed the JIRA ticket to "An application wide user roles system" and clarified it a bit. What this PR is trying to achieve is to have the ability to have multiple system wide roles for a user. This will mean we can have additional roles in future (like ServalTech or SupportStaff or ContentModerator), and grant permissions based on that role, without needing code like `IF SystemAdmin OR ServalTech`. These system wide roles will be limited in scope but allow the ability to affect specific, limited changes to projects across the system.

> With that said, how does having multiple system roles help to support serval? Maybe it does so by allowing a user to be both a `SystemRole.User` and a `SystemRole.ServalTech`?

It means that permission checks can be made based if an application level role is assigned to the user, rather than checking is the user's role in our list of allowed roles (like we do for project level permission checks). In addition, the system administrator could be both a SystemAdmin and a ServalTech.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2234#-:-NnkdrQBAx__lbmImFSj:buo6sz3)* status: 36 of 39 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/RealtimeServer/common/models/user.ts` line 51 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2234#-NnkEnwK-EnVb5ayQLa5:-NnkhRTbBaYEIyKm2OGp:b-2fad15) ([raw file](https://github.com/sillsdev/web-xforge/blob/77dff84322e347c723d4140481952be3fb8dcc08/src/RealtimeServer/common/models/user.ts#L51)):*

Why do we need to change the role property. This indicates whether a user is a system admin, or just a regular user. Since system admins can do anything a user can do, there is no reason for them to have the normal user role.
</blockquote></details>

Correct, but when we add future roles, we can allow users to have any combination of them, granting the specific permissions they need, rather than the hierarchical system currently used for roles, where each role includes the permissions of the roles below it.

The next PR will enable users to be ["SystemAdmin", "ServalTech"] or ["User", "ServalTech"].

There is a likelihood for more roles to be added in future as the scope of what SF does expands, and the need to assign global but limited powers to specific users.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 658 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2234#-NnkMCWu9h5qcbPD5dWB:-NnkiISU5IdCM-r3Uebs:bkpat4f) ([raw file](https://github.com/sillsdev/web-xforge/blob/77dff84322e347c723d4140481952be3fb8dcc08/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L658)):*

Could we not assume that the role would be an array at this point. Since we would only set `ROLES_SETTING` only if the current code has been deployed? I do see that the claim will return a `string | string[]` so it appears we do need to account for the role being a string.
</blockquote></details>

Yes, this was a tricky one for me. I had to settle on allowing a string or an array of strings for the roles in values loaded from auth0, loaded from tokens, and loaded from local storage to ensure maximum compatibility with I hope as little complexity as possible.

This code will only write an array of strings, but I still had to work around the problem of existing sessions having just a string written to the local setting.
___
*[`src/SIL.XForge.Scripture/Controllers/MachineApiController.cs` line 2 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2234#-NnkNSvd1RbCUyJd_HmB:-NnkjIXaCPmY9MXyk-_L:boyexb3) ([raw file](https://github.com/sillsdev/web-xforge/blob/77dff84322e347c723d4140481952be3fb8dcc08/src/SIL.XForge.Scripture/Controllers/MachineApiController.cs#L2)):*

Nit: looks like linq is not being used.
</blockquote></details>

Done. Thanks!

RaymondLuong3 (approved):
@Nateowami It has been a while since the meeting that made it necessary to create this issue. Peter can't recall exactly why it seemed that this was necessary to fulfill SF-2390. If you have the notes or a better memory, can you do a higher level look over this PR. I have approved it based on the information I have, but you know more about our plan for the system wide roles.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2234#-:-NnocBKD2RVUp2NHpQCm:b-9ac4kx)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/RealtimeServer/common/models/user.ts` line 51 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2234#-NnkEnwK-EnVb5ayQLa5:-NnoXy02BGIZodzZvqah:bxg89o) ([raw file](https://github.com/sillsdev/web-xforge/blob/77dff84322e347c723d4140481952be3fb8dcc08/src/RealtimeServer/common/models/user.ts#L51)):*

Correct, but when we add future roles, we can allow users to have any combination of them, granting the specific permissions they need, rather than the hierarchical system currently used for roles, where each role includes the permissions of the roles below it.

The next PR will enable users to be ["SystemAdmin", "ServalTech"] or ["User", "ServalTech"].

There is a likelihood for more roles to be added in future as the scope of what SF does expands, and the need to assign global but limited powers to specific users.
</blockquote></details>

Hmm interesting. In our discussion, it sounds like this gives more flexibility so that we do not need to rely so much on permission hierarchies and make it more simple to introduce new system wide roles. It is hard to know now whether that will actually come to anything, but planning for it earlier is not a bad idea.

Nateowami (commented):
@RaymondLuong3 Just chatted with Peter about this. Conclusion: it's not strictly necessary, but it does give us a lot more flexibility to manage user permissions, which makes it worth doing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2234#-:-NoJrORZCowynfE7uwz8:bc6ndsk)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

RaymondLuong3:
I think we can pass this along to testers now, I'll move it to me tested.

PR #2235, opened 2023-12-12
SF-2406 Support timing files with headings

RaymondLuong3:
The timing files that are exported from HearThis likely have headers. This change assumes that any timing file with some row with 3 columns is audacity style, the same style that HearThis uses. It then allows parsing the entries. Previously, if the first entry does not have 3 columns, it would mark the files as invalid.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2235)


nigel-wells (approved):
Awesome - nice simple fix.
Only comment from me is when I first thought of headings I assumed you were referring to column headings but your referring to metadata about the data in the file aren't you. If you want to tweak the test file or add something in that might be useful - doesn't need to hold up the critical path though so I'll send it through to testing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2235#-:-NlVfphv0CJJ_WRhH5PE:bzav2uu)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2238, opened 2023-12-13
SF-2395 Make Mobile Great Again

josephmyers:
- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2238)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
@@ -7,7 +7,7 @@ $toolbar-height: 56px;
 // Set height of the app to 100% so that the app will fill the entire screen.
 // Don't use 100vh because that will cause the app to be too tall on mobile devices when the browser's address bar is
 // visible.
-// Newer browsers support dvh (dynamic viewport height) units, starting with Safari 15.4 and Chrome 108.
+// Newer browsers support dvh (dynamic viewport height) units, but they are not supported in Safari 15.4 and Chrome 108.
 // Setting the height to 100% seems to work just as well.
 ::ng-deep html,
 ::ng-deep body,
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
@@ -50,6 +50,7 @@ $toolbar-height: 56px;
     > div {
       flex: 1;
       padding: 14px 20px;
+      height: 100%;
       @include media-breakpoint-only(xs, variables.$grid-breakpoints) {
         padding: 10px;
       }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
@@ -7,7 +7,7 @@ $toolbar-height: 56px;
 // Set height of the app to 100% so that the app will fill the entire screen.
 // Don't use 100vh because that will cause the app to be too tall on mobile devices when the browser's address bar is
 // visible.
-// Newer browsers support dvh (dynamic viewport height) units, but they are not supported in Safari 15.4 and Chrome 108.
+// Newer browsers support dvh (dynamic viewport height) units, starting with Safari 15.4 and Chrome 108.
 // Setting the height to 100% seems to work just as well.
 ::ng-deep html,
 ::ng-deep body,
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.scss
@@ -1,6 +1,7 @@
 @import '../checking-global-vars';
 
 :host {
+  position: absolute;
   height: 100%;
   display: flex;
   overflow-y: auto;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.scss
@@ -3,6 +3,7 @@
 :host {
   position: absolute;
   height: 100%;
+  width: 100%;
   display: flex;
   overflow-y: auto;
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html
@@ -135,7 +135,18 @@
             <!-- Splitter area sizes are controlled programmatically. -->
             <as-split-area [size]="100" [maxSize]="scriptureAreaMaxSize" [dir]="i18n.direction">
               <div class="panel-content" #scripturePanelContainer>
-                <!-- TODO (scripture audio) Instead of covering scripture panel, move the audio player below it -->
+                <div class="scripture">
+                  <app-checking-text
+                    #textPanel
+                    [id]="textDocId"
+                    [fontSize]="scriptureFontSize"
+                    [activeVerse]="activeQuestionVerseRef"
+                    [questionVerses]="questionVerseRefs"
+                    (questionVerseSelected)="verseRefClicked($event)"
+                    [isRightToLeft]="isRightToLeft"
+                    [class.hidden]="hideChapterText"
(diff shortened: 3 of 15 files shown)


Nateowami (commented):
Thanks for your work on this. A large number of our users are on older, lower-end mobile devices, which may not be getting updates anymore. We don't want to spend a large chunk of time trying to support the oldest devices, but I do think we need to support them as much as we can. 

See `src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts` for the browsers we're targeting. When it makes sense, we bump these versions, but I avoid doing so as much as possible.

(It's worth noting that Angular only officially supports the last two major versions of each browser (approximately; their support policy is a little more complicated), but in practice I've seen it work quite well in older versions of Chrome).

See https://en.wikipedia.org/wiki/Template:Google_Chrome_release_compatibility for which Chrome versions are supported on which versions of Android. This jumps the requirement to Android 7. Looking at our stats, 6% of users on Android are on Android 6, though as projects in other regions come on board, it wouldn't surprise me to see that go up a lot. Also, developers occasionally opening SF are likely skewing that number somewhat.

I think there should be a way to write CSS so that unsupported values fall back to supported values?

Also, a lot of the rules you've updated appear to be qualitative, rather than quantitative. That is to say, it appears the intent is to prevent a menu from being taller than the viewport, but there is no need to have a precise value. Can we write it in such a way that we're not dependent on dvh?

> Reviewed 1 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2238#-:-NlYlNkaEvHg-zegs3cm:bd0l37d)* status: 1 of 8 files reviewed, all discussions resolved

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss` line 10 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2238#-NlYmwFj4usdQF-ePspU:-NlYmwFj4usdQF-ePspV:b-ailmn7) ([raw file](https://github.com/sillsdev/web-xforge/blob/c2a13ce005c653a82488e2fcf18dac4995fdc048/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss#L10)):*
> ```SCSS
> // Don't use 100vh because that will cause the app to be too tall on mobile devices when the browser's address bar is
> // visible.
> // Newer browsers support dvh (dynamic viewport height) units, starting with Safari 15.4 and Chrome 108.
> ```

Oops, thanks for catching my mistake.

josephmyers (commented):
Thanks for this writeup, it was really helpful. I absolutely agree that we don't want to drop support unnecessarily. To me, the alternative is JS-based. Unless you're suggesting we intend the older versions to fall back to the bugged CSS values? Can you explain a little bit what you're thinking?

With regards to the qualitative settings, I think there are only two, unless you're thinking some of the "100's" are. I can certainly update (or revert) these.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2238#-:-NlaAy6f1SG8k8gBH6ld:bcujrit)* status: 1 of 8 files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami (commented):
Falling back to less than ideal behavior is one option, but I think it may also be possible to preserve the behavior from before I broke it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2238#-:-NlaF7cQ8AEiVfUJS6j3:bej8mo1)* status: 1 of 8 files reviewed, all discussions resolved

josephmyers (commented):
Yeah, that would be awesome. Are you suggesting trying to get something to work with "position: absolute"? That's what was there before your change, and it seems to work correctly with percentage height's on mobile and desktop. Absolute positioning is in line with my second link, but I wasn't able to re-fix the bug with chapter audio overlapping the scrollbar.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2238#-:-Nlw5Oo7-i2lgd5Miikm:b-f174ks)* status: 1 of 8 files reviewed, all discussions resolved (waiting on @Nateowami)

nigel-wells (changes requested):
This does work a whole lot better on my Android phone - I've not tested an older device or iPhone yet.
Just a couple of things I've noticed:

* If a question has an answer with audio then the bottom bar creates a scroll
* Translate drafting still causes a scroll

I'll try some other tasks in the translate app once you get that working

> Reviewed 2 of 2 files at r2, 6 of 8 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2238#-:-NoKKhGdBd6KDW1UQfSI:btl8ek8)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

josephmyers (commented):
Thanks. I just remembered that Nathaniel recently merged in a refactor that removed a lot of these hard-coded values, e.g. calc(100vh - 236px). He literally removed the code where I made my fix for Translate, though what he did was an absolute improvement. "It used to work." - Joseph Myers :D

You'll actually see the same buggy behavior on master. Translate is too tall on mobile, and it needs the same width-specific CSS that we have for Community Checking. I've uploaded what fixes it on my devices, but absolutely check me.

For Community Checking, I'm not seeing any runoff/scroll issues. My device does have a scroll for the bottom pane, but that's because there's a lot of content there. I'll post a screenshot. Maybe our devices are working differently? That would be annoying.

I do see an odd scroll on desktop, but only when the browser width is <576px. This is due to the exact problem I'm addressing, where desktop doesn't correctly simulate mobile devices. Same for Translate.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2238#-:-NoKoJcK7dp36IcvKRoj:b-yhvbph)* status: 10 of 11 files reviewed, all discussions resolved (waiting on @Nateowami)

josephmyers:
Here it is with correct runoff on my old Pixel 3

nigel-wells (commented):
I think it may be if there is chapter audio available that is pushing it down - if I remove that then it lines up nicely. p.s. the whitespace at the very bottom is my phone UI and not a bug 


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2238#-:-NoPCCy99yZDL9QHWXYp:b-e86fqk)* status: 10 of 11 files reviewed, all discussions resolved (waiting on @Nateowami)

nigel-wells (commented):
This is what I get for translate - it is better than yesterday though:


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2238#-:-NoPCn0N-mtGAGrQUpwI:b-utywbo)* status: 10 of 11 files reviewed, all discussions resolved (waiting on @Nateowami)

nigel-wells (commented):
I see in `editor.component.scss` there is a rule on `.content` to change the height by specific pixels which is what causes the different height on mobile. Adjusting this doesn't fix the issue above on a device but it does make it more consistent on smaller desktop screens.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2238#-:-NoPGDWCDyz1oH9_c9Rz:b-qf8cs9)* status: 10 of 11 files reviewed, all discussions resolved (waiting on @Nateowami)

josephmyers:
I'm fine using your branch. It seems to fix the problem with CC. My changes in `editor.component.scss` still appear to be needed.

Nateowami:
@nigel-wells I think the editor component is partly broken on mobile (possible to scroll the toolbar away, though it's there by default).

nigel-wells:
Sorry for missing this one, I'll take another look at to see if the translate editor can be improved as well.

nigel-wells:
I've pushed through an update which makes the translate app work better on mobile. Let me know what you think.

nigel-wells (commented):
> Reviewed 1 of 1 files at r8, 1 of 1 files at r9, 2 of 2 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2238)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

josephmyers:
Looks awesome. I ran through the various pages on both desktop and my Android, and the bug seems quite squashed. Thanks for the assistance

nigel-wells (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2238#-:-Nr2wg6A0j01hkUWoBcx:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2241, opened 2023-12-14
SF-2411 Restrict resolve option on dialog when restricted tags used

RaymondLuong3:
* synchronize note tag create resolve property
* add logic to detect if a user is allowed to resolve a thread
* hide the resolve note trigger when user is not allowed to resolve

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2241)

### DIFF
--- a/src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts
+++ b/src/RealtimeServer/scriptureforge/services/sf-project-migrations.spec.ts
@@ -325,6 +325,24 @@ describe('version 12', () => {
       expect(projectDoc.data.translateConfig.draftConfig.lastSelectedTrainingBooks).toEqual([1, 2, 3]);
     });
   });
+
+  describe('version 15', () => {
+    it('adds create resolve property to note tag', async () => {
+      const env = new TestEnvironment(14);
+      const conn = env.server.connect();
+      await createDoc(conn, SF_PROJECTS_COLLECTION, 'project01', {
+        noteTags: [{ tagId: 1, name: 'Tag 01', icon: '01flag1', creatorResolve: false }]
+      });
+
+      let projectDoc = await fetchDoc(conn, SF_PROJECTS_COLLECTION, 'project01');
+      expect(projectDoc.data.noteTags[0].creatorResolve).toBe(false);
+
+      await env.server.migrateIfNecessary();
+
+      projectDoc = await fetchDoc(conn, SF_PROJECTS_COLLECTION, 'project01');
+      expect(projectDoc.data.noteTags[0].creatorResolve).toBe(true);
+    });
+  });
 });
 
 class TestEnvironment {
--- a/src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts
+++ b/src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts
@@ -1,6 +1,7 @@
 import { Doc, Op } from 'sharedb/lib/client';
 import { DocMigration, MigrationConstructor } from '../../common/migration';
 import { submitMigrationOp } from '../../common/realtime-server';
+import { NoteTag } from '../models/note-tag';
 import { SFProjectRole } from '../models/sf-project-role';
 import { TextInfoPermission } from '../models/text-info-permission';
 import { TranslateShareLevel } from '../models/translate-config';
@@ -252,6 +253,20 @@ class SFProjectMigration14 extends DocMigration {
   }
 }
 
+class SFProjectMigration15 extends DocMigration {
+  static readonly VERSION = 15;
+
+  async migrateDoc(doc: Doc): Promise<void> {
+    const ops: Op[] = [];
+    for (let i = 0; i < doc.data.noteTags.length; i++) {
+      const noteTag: NoteTag = doc.data.noteTags[i];
+      ops.push({ p: ['noteTags', i, 'creatorResolve'], od: noteTag.creatorResolve });
+      ops.push({ p: ['noteTags', i, 'creatorResolve'], oi: true });
+    }
+    await submitMigrationOp(SFProjectMigration13.VERSION, doc, ops);
+  }
+}
+
 export const SF_PROJECT_MIGRATIONS: MigrationConstructor[] = [
   SFProjectMigration1,
   SFProjectMigration2,
@@ -266,5 +281,6 @@ export const SF_PROJECT_MIGRATIONS: MigrationConstructor[] = [
   SFProjectMigration11,
   SFProjectMigration12,
   SFProjectMigration13,
-  SFProjectMigration14
+  SFProjectMigration14,
+  SFProjectMigration15
 ];
--- a/src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts
+++ b/src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts
@@ -260,10 +260,12 @@ class SFProjectMigration15 extends DocMigration {
     const ops: Op[] = [];
     for (let i = 0; i < doc.data.noteTags.length; i++) {
       const noteTag: NoteTag = doc.data.noteTags[i];
-      ops.push({ p: ['noteTags', i, 'creatorResolve'], od: noteTag.creatorResolve });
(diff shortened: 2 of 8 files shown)


pmachapman (commented):
I needed to perform a synchronization for this change to work on my test project, due to `CreatorResolve` not being pulled through previously.

Do you think we will need to have a migrator that would synchronize affected projects, or perhaps create the op with the `creatorResolve` value for affected `noteTags` in the `sf_projects` collection?

To get an idea of the scope of the problem, running this command on QA & Production might help:

```bash
grep -r 'CreatorResolve="true"' /var/lib/scriptureforge/sync/*/target/CommentTags.xml
```

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2241#-:-NltJPMf0dxlLCqm0d29:b-ccgb97)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
Thanks for that feedback! I should have thought this through more thoroughly. Synchronizing those projects might be the best option to enforce the creatorResolve property. I decided to go with a simpler approach, we start be migrating all note tags to be `creatorResolve = true` and then when the project synchronizes, the up-to-date note tag properties will be applied.
It is not ideal since some users will be wondering why one day they could resolve a note, and the next day they cannot. But since this feature is still relatively new, I do not see a big problem with it (and admins will always be able to resolve notes, so this only affects translators and consultants).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2241#-:-Nm27pLKEFK5HJ2TJiLt:b4o864h)* status: 4 of 8 files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (changes requested):
Hmmmm. We could write a migrator in C# to migrate the values?

Were you able to run the grep command on QA and Prod? I would be interested to know how many tags are `true` vs. `false`.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2241#-:-Nm2rf_J9MmjovcAc2mP:b-yyelos)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts` line 263 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2241#-Nm2s6du8jEezVUEfHoX:-Nm2s6du8jEezVUEfHoY:b-rf5q4q) ([raw file](https://github.com/sillsdev/web-xforge/blob/304961253219b22a4ac0699a1980d6f5d4edcfec/src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts#L263)):*
> ```XML
>     for (let i = 0; i < doc.data.noteTags.length; i++) {
>       const noteTag: NoteTag = doc.data.noteTags[i];
>       ops.push({ p: ['noteTags', i, 'creatorResolve'], od: noteTag.creatorResolve });
> ```

Could we wrap this in an if statement, in case the migrator is rerun for some reason?
```ts
if (!noteTag.creatorResolve) {
```



_Code quote:_
```XML
ops.push({ p: ['noteTags', i, 'creatorResolve'], od: noteTag.creatorResolve });
```
___
*[`src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts` line 266 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2241#-Nm2fFa_2Op-Mdc0Yt5r:-Nm2fFa_2Op-Mdc0Yt5s:b2ek6mb) ([raw file](https://github.com/sillsdev/web-xforge/blob/304961253219b22a4ac0699a1980d6f5d4edcfec/src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts#L266)):*
> ```XML
>       ops.push({ p: ['noteTags', i, 'creatorResolve'], oi: true });
>     }
>     await submitMigrationOp(SFProjectMigration13.VERSION, doc, ops);
> ```

This should be: `SFProjectMigration15.VERSION`



_Code quote:_
```XML
await submitMigrationOp(SFProjectMigration13.VERSION, doc, ops);
```

RaymondLuong3 (commented):
What is the advantage of migrating the values with the c# migrator? I haven't looked at it to know what benefits it gives us.
I did run the  grep command on live and there were 225 note tags with `creatorResolve = true`

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2241#-:-Nm2tiTQ3KKsZe8hakYb:bvrhar4)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts` line 263 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2241#-Nm2s6du8jEezVUEfHoX:-Nm2ufhW64YHSbgRfP55:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/304961253219b22a4ac0699a1980d6f5d4edcfec/src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts#L263)):*

Could we wrap this in an if statement, in case the migrator is rerun for some reason?
```ts
if (!noteTag.creatorResolve) {
```
</blockquote></details>

Done.
___
*[`src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts` line 266 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2241#-Nm2fFa_2Op-Mdc0Yt5r:-Nm2uKiY0e9MPPz-KxX5:b-dla35x) ([raw file](https://github.com/sillsdev/web-xforge/blob/304961253219b22a4ac0699a1980d6f5d4edcfec/src/RealtimeServer/scriptureforge/services/sf-project-migrations.ts#L266)):*

This should be: `SFProjectMigration15.VERSION`
</blockquote></details>

Nice catch. Copy pasta...

pmachapman (approved):
The advantage would be we could accurately populate the creatorResolve value from the ScrText, like we do for a sync but without syncing. I could write this if you want me to.

Alternately, (probably easier to deploy, too) we could have `SFProjectMigration15` in TypeScript open the /var/lib/scriptureforge/{paratextId}/target/CommentTags.xml file and populate the value? I think we might need to include a third party XML library in the realtime server if we are not already, though.

Up to you. If you want to write a better migrator, great. If you think the support burden from users who want to resolve who previously could will be minimal, then :lgtm:.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2241#-:-Nm2vS2p5pJBaIY8h8jp:bzfsfsn)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
Wow, that is really neat. Give me a chance to look into that, if it is not very difficult, let's do it. If not, then we can decide if it is worth the effort to properly migrate the projects.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2241#-:-Nm2ytqxBuygK-9OsPpQ:br84tpb)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
Thanks for the suggestions. I think I will move this along to testing. With how permissions work in SF, we depend on the project to be in sync with SF as much as possible. Project roles aren't updated in SF until a synchronization is done. The `creatorResolve` property similarly depends on note tags to be in-sync which is up to the admins to do. So have a temporarily non in-sync note tag is acceptable.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2241#-:-Nm7hTLS7uplimuLh9B-:b-ox99ep)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

pmachapman (approved):
:lgtm_strong: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2241#-:-Nm7lCF1DPJGyPxkBh3u:b-uwkowd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2255, opened 2023-12-20
Force Chromatic Rebuild for Storybook

pmachapman:
As storybook tests require the `storybookUrl` value from Chromatic, this PR forces Chromatic to rebuild, even if it thinks there are no changes.

For the reason behind this change, please see the "Publish to Chromatic" and "Run Storybook tests" sections of https://github.com/sillsdev/web-xforge/actions/runs/7277973553/job/19836485161

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2255)

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -69,6 +69,8 @@ jobs:
           onlyChanged: true
           externals: |
             - '*.scss'
+          # Force a rebuild even if Chromatic thinks there are no changes
+          forceRebuild: true
       - name: Run Storybook tests
         run: cd src/SIL.XForge.Scripture/ClientApp && npm run test-storybook -- --no-index-json
         env:


Nateowami:
It seems like we should be able to serve the storybook ourselves, i.e. not run the tests against what Chromatic is serving.

I'm a bit concerned doing the rebuild could have drawbacks with what Chromatic uses as the baseline for comparisons. We've had issues with that already.

pmachapman:
> I'm a bit concerned doing the rebuild could have drawbacks with what Chromatic uses as the baseline for comparisons. We've had issues with that already.

Is it only commits on `master` suffering this problem, or PRs as well? If it is PRs as well, we could put this on ice until it occurs again, then see if this fixes it then. If it is just on `master`, I think we could skip this PR if you think the risk is not worth the fix?

Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2255#-:-OJOh9qH-wf8fmZmvI7J:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2256, opened 2023-12-21
SF-2428 Allow drafting with no training data when target and source in NLLB

siltomato:
Selecting training books is optional (and discouraged) when both source and target languages are in the NLLB.  Translation book selection is still required.

This PR modifies the UI in the draft generation stepper so that when both source and target are in the NLLB:
- User is shown notice encouraging skipping training book selection.
- User is shown heavier warning if training books are selected.
- User is allowed to click "Generate draft" without error if no training books are selected, and an empty array of training books is passed to the API.


josephmyers (changes requested):
Cool, I think this looks fine overall. In one place, though, the workflow for me was a little misleading, in that it recommends you skip the training step and click "Generate draft" but defaults books to be selected. For me, it had Mark selected; it was the only book there. You can pretty easily envision a scenario where the user sees "skip this" and does so, without really making the connection that they need to uncheck things before launching the generation. I recommend we default all books to deselected if training is optional.

My only other comment to the front end is that "training book selection step" is a little too wordy. I think it would be clearer, with no information lost, to say, _We recommend that you skip this step and continue to "Generate draft"._

> Reviewed 2 of 5 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2256#-:-NmiTfri9fzt5NYXhszl:b-1c4m9i)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

siltomato (commented):
Interesting point about the pre-selection of training books.  Let me give you the current logic for training book pre-selection, and then I'd love to hear your thoughts regarding the UX.

The way it works now is that it will pre-select books for training that the user had previously selected for training, which means that the user would have already ignored the warning for a previous build.  With this in mind, do you think it should clear the previous selection each time, or pre-select?

I modified the warning to recommend clearing the current selection if any training books are selected.  Does this help or increase confusion?

Also, I like your wording of the warning better.  Changed!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2256#-:-Nn9yn84-wbNK5J8XtjS:b7n652p)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

josephmyers (approved):
Ah, so it must've been a previous run of mine causing that. The workflow makes more sense now, and I like the updates you made. Thank you!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2256#-:-NnC5sC570KZ2tEJe6ly:bwb92pm)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

PR #2268, opened 2024-01-03
SF-2334 Update to Angular 16

siltomato:
Here are the workarounds used to get everything building with tests passing:
- npm package `@triviality/logger` was causing a test build error that look like it was due to missing types.  The package was old and we only used it for an abstract class, interface, and enum.  I copied these into the file with the only class that used them and I removed the package reference.
- `ng build` was showing an error in creating the `index.html` file.  This was fixed by deleting and recreating the `package-lock.json` files.
- Froze npm package `@sillsdev/scripture` at version 1.4.1, as a change in 1.4.2 was breaking some tests due to the added `instanceof` check in `VerseRef.equals()`.  **This will need to be investigated further.**
- Added `npx playwright install` before `test-storybook` command to ensure playwright browsers are installed.  This fixed an error when running storybook tests.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2268)


Nateowami (approved):
> Reviewed 4 of 9 files at r1, 11 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2268)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

PR #2272, opened 2024-01-03
SF-2432 Fix quill editor double scroll bar

RaymondLuong3:
The angular host appears to be injecting a style that affects the quill editor, causing the display to be set to inline-block. This was causing the quill editor to overflow its flex container and creating a double scroll bar. This PR forces the quill editor element to have the display set to unset, keeping the entire quill editor within its flex container.
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2272)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
@@ -4,6 +4,7 @@
 
 quill-editor {
   height: 100%;
+  display: unset !important;
   > .ql-container {
     letter-spacing: 0.03125em;
     > .ql-editor {


Nateowami (changes requested):
In my testing, the editor is now sometimes very tall (a lot longer than the page). It's somewhat random. 

Also, instead of setting the display to "unset !important`, I think it makes more sense to add component styles to the text component, and then in `text.component.scss`, add:
``` scss
quill-editor {
  display: inline;
}
```



> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2272#-:-NnGGBCF4pLwVZuTLf9T:b-cegcjk)* status: 0 of 1 files reviewed, all discussions resolved

Nateowami:
See #2273 for an attempt to completely rework the editor styles.

PR #2281, opened 2024-01-08
Fix layout of editors when only one editor is shown

Nateowami:
This is a fix for the same problem as #2280.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2281)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss
@@ -131,47 +131,36 @@ app-text {
 .content {
   height: calc(100vh - 84px);
   display: grid;
-  grid-template-areas:
-    'toolbar toolbar'
-    'source target';
   column-gap: 10px;
 
   grid-template-rows: 72px minmax(0, 1fr);
   grid-template-columns: 1fr 1fr;
-
-  &.reverse-columns {
-    grid-template-areas:
-      'toolbar toolbar'
-      'target source';
-  }
-}
-
-@include media-breakpoint-only(xs) {
-  .content {
-    grid-template-columns: 1fr;
-    grid-template-areas: 'toolbar' 'target';
-  }
 }
 
 .toolbar {
-  grid-area: toolbar;
+  grid-column: 1 / span 2;
 }
 
-#source-text-area {
-  grid-area: source;
+.text-area.text-area-full-width {
+  grid-column: 1 / span 2;
 }
 
-#target-text-area {
-  grid-area: target;
+.text-area {
+  grid-row-start: 2;
+}
+
+.reverse-columns {
+  #source-text-area {
+    grid-column-start: 2;
+  }
+  #target-text-area {
+    grid-column-start: 1;
+  }
 }
 
 .text-container {
   height: 100%;
   position: relative;
-}
-
-.text-area.text-area-full-width {
-  grid-column: 1 / span 2;
 }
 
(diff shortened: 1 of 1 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2281)* status: 0 of 1 files reviewed, 3 unresolved discussions

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss` line 134 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2281#-NndI5N47DhPiZk0X9q2:-NndI5N47DhPiZk0X9q3:bb2ajrr) ([raw file](https://github.com/sillsdev/web-xforge/blob/2ec46ab732ce7f8dceb743f58552569d9229a65f/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss#L134)):*
> ```SCSS
>   height: calc(100vh - 84px);
>   display: grid;
>   grid-template-areas:
> ```

Previously I was using the grid areas to specify where things should be placed, but unfortunately that wasn't very flexible. I've completely changed the strategy to avoid specifying named areas.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss` line 149 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2281#-NndJ374CLa8ayYbK92B:-NndJ374CLa8ayYbK92C:burflh0) ([raw file](https://github.com/sillsdev/web-xforge/blob/f1e6c4901f248c3bece024a0599b1f8108e944a1/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss#L149)):*
> ```SCSS
> 
> .text-area {
>   grid-row-start: 2;
> ```

This is specifying that the editors should always be in the second row. Usually it would do this by default, but when the source and target are inverted, the source is placed in column 2, and since there are only two columns, the target gets moved to a third row. This is specifying that the editors should always be in the second row.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss` line 154 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2281#-NndIV-Z4WAuUcD2AzW7:-NndIV-Z4WAuUcD2AzW8:buf4vj9) ([raw file](https://github.com/sillsdev/web-xforge/blob/f1e6c4901f248c3bece024a0599b1f8108e944a1/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss#L154)):*
> ```SCSS
> .reverse-columns {
>   #source-text-area {
>     grid-column-start: 2;
> ```

Unfortunately, CSS grid doesn't have a way to tell it to lay out from the right instead of the left, so far as I can tell. It can be done by setting the text direction, but that has other side effects. So instead we're just specifying that when the editors need to be reversed, the source starts in column 2, and the target starts in column 1.

pmachapman (approved):
:lgtm: A great fix - it worked well in my testing, and I think the CSS is clearer now. Thank you!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2281#-:-NneI8tUEbkRKwiBY42E:bhe58r5)* status: all files reviewed, 3 unresolved discussions

pmachapman (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2281)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2291, opened 2024-01-10
SF-2457 Avoid incorrect plurals in checking overview labels

callumlaw17:
Under the manage questions section each book/verse has a question count label and each book/verse/question has an answer count label.

In the case that there is only one question or answer we are still using the plural e.g. "1 questions" where we should be using the non-plural e.g. "1 question"

This PR adds two new translations to use when there is only one question/answer and uses them when applicable.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2291)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts
@@ -416,7 +416,7 @@ describe('CheckingOverviewComponent', () => {
       env.waitForQuestions();
       expect(env.textRows.length).toEqual(2);
       expect(env.textArchivedRows.length).toEqual(1);
-      expect(env.getArchivedQuestionsCountTextByRow(0)).toContain('1 questions');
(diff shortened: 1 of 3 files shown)


Nateowami:
Localizing numbers with proper plurals is a lot more complicated, so much so that I haven't really thought we should try to tackle it, since it seems like such a major undertaking: https://www.unicode.org/cldr/charts/44/supplemental/language_plural_rules.html

Crowdin also has a section on it: https://support.crowdin.com/icu-message-syntax/#plural

Furthermore, our localizers will need to understand the ICU syntax (see the screenshot on the Crowdin page; it shows the localizer inputting the markup). But most likely they won't, so we'll have to coach each of them through the process.

callumlaw17:
@Nateowami Alright, I suspected that was the case but thought that this small change may be worth it as a stepping stone - happy to close this off if you would rather it was revisited with a more comprehensive solution.

RaymondLuong3:
I agree with Nathaniel, the plural forms of a word when it is incorrect is not ideal, but to really solve the problem for the other languages is not simple. So we can push this off unless there is a pressing reason to need to solve it.

PR #2293, opened 2024-01-10
SF-2384 Fix flicker after project book is deleted

RaymondLuong3:
When a user deletes a book in PT, the project component is not made aware and tries to navigate to the delete book. This causes an infinite redirect when the editor component redirects back to the project component when it realizes that the book does not exist. This PR updates the project component to navigate to a book only if the book exists on the project.

To reproduce this issue easily, update the SFProjectUserConfig doc entry in mongo so the selectedBookNum is a book that does not exist on the project.

Selected Book Num Flicker

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2293)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.spec.ts
@@ -70,6 +70,18 @@ describe('ProjectComponent', () => {
     expect().nothing();
   }));
 
+  it('navigate to first text when last selected text missing', fakeAsync(() => {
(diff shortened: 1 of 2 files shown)


nigel-wells (approved):
That redirect loop is brutal - thanks for the fix!

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2293#-:-NnqXKjMAgMl_7PYQeB6:byqdfqo)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2300, opened 2024-01-18
SF-2463 Fix Paratext 7 Resources being marked as English

pmachapman:
**Problem Overview**
Paratext 7 Resources store the language definition in the file ldml.xml, while Paratext 8+ store this data in the file [lang_id].ldml. One such project on prod and dev is DYU, or just in dev (i.e. is correct in prod), NNRV.

When ParatextData reads the installed resource in Scripture Forge, it shows the Resources language as being the default, English.

This was previously not much of a problem for Resources in Scripture Forge, but as we send this language code to Serval, we need to ensure that this code is correct so that pre-translation drafting can occur.

**Solution Overview**
The Paratext migrators are quite complex, and locked in the Paratext source code (i.e. they are not in ParatextData.dll). Because of this, I have implemented the minimum required migrator to move the shipped LDML file to a location where ParatextData can read the correct language code.

**Migration Notes**
To fix any existing installed resources, run the script `trigger-resources-need-sync.ts` in PR #1764.

**Testing Notes**
This PR should be tested by a developer as its changes are predominantly under the hood.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2300)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -2335,7 +2335,7 @@ public class ParatextService : DisposableBase, IParatextService
         if (target is ParatextResource resource)
         {
             // If the target is a resource, install it
-            InstallResource(resource, target.ParatextId, targetNeedsCloned);
+            InstallResource(username, resource, target.ParatextId, targetNeedsCloned);
         }
         else if (targetNeedsCloned)
         {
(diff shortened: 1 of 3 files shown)


RaymondLuong3 (approved):
Thanks for finding and fixing this. I can see that if I delete the repo locally and reconnect to NNRV, the `ldml.xml` file is renamed to `npi.ldml`. It seems to be working as expected. I am glad we do not need to implement the PT 7 to PT 8 migrator. The minimal approach is probably the right one in this case.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2300#-:-NoTb1Xt5vurpygcHdar:b-z2k9c1)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2304, opened 2024-01-19
SF-2469 Inform user when no mic is found

Nateowami:
I couldn't figure out why my machine wouldn't let me record, despite having given permission. It just kept saying permission denied, which I figured was related to being on localhost (an insecure origin).

Actually I just didn't have a mic connected. Probably not a common situation, but even I couldn't figure that out without stepping through the debugger. :facepalm:

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2304)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-recorder/checking-audio-recorder.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-recorder/checking-audio-recorder.component.spec.ts
@@ -126,7 +126,7 @@ class TestEnvironment {
     });
     when(mockedNavigator.mediaDevices).thenReturn({
       getUserMedia: (mediaConstraints: MediaStreamConstraints) =>
-        this.rejectUserMedia ? Promise.reject() : navigator.mediaDevices.getUserMedia(mediaConstraints)
+        this.rejectUserMedia ? Promise.reject({}) : navigator.mediaDevices.getUserMedia(mediaConstraints)
     } as MediaDevices);
     this.fixture.detectChanges();
   }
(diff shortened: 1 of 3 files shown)


pmachapman (approved):
:lgtm: Just one comment about a potential use case when the microphones are disabled.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2304#-:-NomZ9EPD2ERsa7Z07X_:b-rnduzn)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-recorder/checking-audio-recorder.component.ts` line 111 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2304#-NomVTGxAJ6VwYPO-xyC:-NomVTGxAJ6VwYPO-xyD:b638v4q) ([raw file](https://github.com/sillsdev/web-xforge/blob/e180cbd4eb59e795739ec980708787a503cb6b30/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-audio-recorder/checking-audio-recorder.component.ts#L111)):*
> ```XML
>     this.status.emit({ status: 'denied' });
> 
>     if (error.code === DOMException.NOT_FOUND_ERR) {
> ```

I tested this locally by disabling all of my microphones, and had the `access_denied` message appear because the code returned was `NotReadableError`/`"Could not start audio source"` (https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia#notreadableerror), which does not have a legacy code, so has to be checked for with:
```ts
if (error.code === DOMException.NOT_FOUND_ERR || error.name === 'NotReadableError') {
```

I tested on another machine that had no physical microphone, and the code worked as expected (the code returned was `DOMException.NOT_FOUND_ERR`), so up to you whether you think it is necessary to check for the case of a user disabling their microphones.



_Code quote:_
```XML
if (error.code === DOMException.NOT_FOUND_ERR) {
```

Nateowami:
I wasn't aware microphones could be disabled. If all mics are disabled, does that mean there are no mics found, or that access was denied? It sounds like right now it tells the user that access was denied in that situation, which seems correct to me.

pmachapman (approved):
Sounds good. I think access denied is the best way to describe it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2304#-:-NowoY959W6gNbsUFwux:bcditnb)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2315, opened 2024-01-29
SF-2478 Support nb paragraph marker

marksvc:
- Add nb as valid paragraph marker
- test: add ASV sample data
- test: DeltaUsxMapper roundtrips ASV Bible

I am okay with the PR being squashed as a single commit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2315)


marksvc:
This commit adds support for nb as a paragraph marker. Without support for nb, 1 Cor 11:1 in GTt shows invalid formatting, as seen in this screenshot: 

GTt-1Cor11 1-invalid-nb--Screenshot from 2024-01-24 15-19-03

Some ASV Bible sample data is added, showing that the ASV sample data, which includes the \nb marker, roundtrips successfully.

Nateowami:
I'm a bit unsure about why this paragraph marker in particular was chosen. It looks like there are [roughly 15 paragraph markers in the USFM spec](https://ubsicap.github.io/usfm/paragraphs/index.html), and we only support `\p` and `\b`.

Looking at the definition of `\nb`, it only indicates that there shouldn't be a space between paragraphs here. The semantics are such that if we basically treat it like a regular paragraph we'll be fine. We could update our CSS to display it just so, but we don't *have* to do that to round-trip this, and make a decent editing experience.

Looking at several other paragraph markers, I see similar things; they mostly have stylistic differences, not differences in how we would handle the data.

The point is, the same logic that leads me to look at this and think "yes, we can safely treat \nb just like we treat \p" leads me to think we can safely add most of the other paragraph markers as well, so I don't really understand why we would add this one and not similar markers.

I had thought the reason more USFM markers weren't supported was mainly because some of them required more complex handling than our system currently supports, so we wouldn't be able to add them without additional development effort and added complexity. It appears that was a mistaken assumption.

marksvc:
Thanks for having a look. You ask a good question.

I didn't approach this thinking "I want to support more paragraph markers", but rather "If I load some Bibles into this, do they roundtrip the data successfully, or do we corrupt it?"

If we load in the ASV Bible, process it to USX, make ChapterDeltas out of that, and convert them back to USX, we find that our resulting USX is missing huge sections of data. For example, Genesis chapter 28 exists, but has no content, as seen in this screenshot from a visual diff:

image

(You can reproduce this on your machine if you comment out `"nb",` in DeltaUsxMapper.cs, run the test in debugger and break on "All Exceptions" (or perhaps put a breakpoint on `Assert.IsTrue(XNode.DeepEquals(roundTrippedUsx, bookUsx), $"Trouble in {entry.Name}");`. 
(Actually, I'm rearranging the sample data a bit, so it may be different than this, but should be the same idea.)
)

Adding support for the `\nb` marker results in successfully being able to read and roundtrip the ASV data (or at least the Scripture books of it). 

I notice that the lack of Chapter 28 content seems a bit different than the GTt 1 Cor 11:1 situation, so there's probably a bit more to it than just missing Chapter 28 content.

I had only loaded in two Bibles (Hebrew and ASV) before giving my attention to other issues. When I came back to this code, I cleaned it up to propose, but haven't tried loading in other sample data sets to see what other markers should be supported, or other problems that we may be having that we don't know about yet. 

For some additional context, last year I asked Damien, 
"""
in DeltaUsxMapper.cs, there is a list of [ParagraphPoetryListStyles](https://github.dev/sillsdev/web-xforge/blob/master/src/SIL.XForge.Scripture/Services/DeltaUsxMapper.cs#L25) (and also a list in text-view-model.ts [PARA_STYLES](https://github.dev/sillsdev/web-xforge/blob/master/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts#L11)). This list is much smaller than the [Para.para.style.enum](https://github.dev/sillsdev/web-xforge/blob/master/src/SIL.XForge.Scripture/usx-sf.rnc#L139) in usx-sf.rnc. How can I determine whether a given paragraph style in usx-sf.rnc is not present in `ParagraphPoetryListStyles` because it _shouldn't_ be there, or because it just hasn't been added yet?
Or maybe another way to ask is, why are some items in usx-sf.rnc not present in the list of `ParagraphPoetryListStyles`?

[...] I haven't found any information regarding why `ParagraphPoetryListStyles` only contains some of the paragraph styles and not others, to know whether I can just add another style [\nb] to the list and not worry about it, or what I should be taking account of, or maybe some styles are something that shouldn't be added for some reason.
"""
Damien said that 
"""
When I originally wrote DeltaUsxMapper, I based it  on the USX documentation. There is no mention of the paragraph style nb in the documentation, which is probably why it is not included in ParagraphPoetryListStyles. I simply wasn't aware it existed. [...] It looks like nb can contain verse text, so you can safely add it. If you do add support for it, I would check PT's stylesheet to see if it is formatted differently than standard paragraphs. If so, add a style for it _usx.scss.
"""

Should we add in support for more paragraph markers? Probably yes, if Bibles are going to be using them. Should we proactively add support for paragraph markers before finding that a Bible is using it? Maybe, and that can be balanced among our various issues.
Would you like me to write up an issue about adding support for more paragraph markers?

I did find that not only in the ASV sample data, but also in our GTt test project, there is the `\nb` marker. So part of GTt (1 Cor 11:1) works better with support for `\nb`. (Otherwise it says the formatting is invalid and the chapter can not be edited, as shown in an earlier image.)

> and we only support \p and \b.

Hmm. I haven't looked closely to see what happens with other markers, but comparing the list in DeltaUsxMapper.cs ParagraphPoetryListStyles and the list on the USFM web site of paragraph markers, it looks like we account for p, m, po, pr, cls, pmo, pm, pmc, pmr, maybe pi, mi, now nb, pc, and maybe ph. And I would need to test 'b' to say.

> The point is, the same logic that leads me to look at this and think "yes, we can safely treat \nb just like we treat \p" leads me to think we can safely add most of the other paragraph markers as well, so I don't really understand why we would add this one and not similar markers.

That could well be. `nb` was added here rather than others because of the goal of roundtripping some data sets, rather than a goal of supporting more paragraph markers. But yes we could look into supporting more paragraph markers.

marksvc:
Ah, having a JIRA issue more-so is emphasizing the intent to support `\nb` specifically. I made a JIRA issue so I would have a ticket to write an acceptance test on :)

marksvc:
Okay, if you comment out `"nb",` in DeltaUsxMapper.cs and debug test "RoundTrip_Asv", you can see that the roundtripped USX is missing 1CO11 content, as seen in this image. And oops, the diff is a bit reversed, with the original content on the right, and roundtripped content on the left.

image

Nateowami:
Just to clarify, the current behavior wouldn't result in data loss to the actual project, right? We just mark the chapter as invalid, and then don't ever send the data back to PT?

marksvc:
> Just to clarify, the current behavior wouldn't result in data loss to the actual project, right? We just mark the chapter as invalid, and then don't ever send the data back to PT?

Yes, I think that's true.

pmachapman (approved):
:lgtm: I tested locally, and syncing GTt after applying this fix works correctly, and makes the chapter valid.

Also, the formatting matches how PT displays `\nb` (A new line with no indent)

> Reviewed 2 of 7 files at r1, 1 of 1 files at r2, 5 of 5 files at r4, all commit messages.
> Dismissed @github-advanced-security[bot] from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2315#-:-Npzt2K097cx8Dp2PWht:b2wo3xg)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #2316, opened 2024-01-30
SF-2477 Menus do not automatically close when opening another menu

nigel-wells:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2316)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
@@ -24,6 +24,14 @@ $toolbar-height: 56px;
     z-index: 8;
   }
 
+  > header {
+    position: absolute;
+
+    &.overlay-drawer {
+      z-index: 6;
+    }
+  }
+
   .top-app-bar-adjust {
     margin-top: $toolbar-height;
   }
@@ -56,9 +64,8 @@ $toolbar-height: 56px;
 }
 
 header {
-  position: absolute;
   width: 100%;
-  z-index: 1001; // Needs to be larger than overlay containers
+  z-index: 7;
 
   .title {
     cursor: pointer;


Nateowami:
If I understand correctly, the problem we're trying to solve is this:



It's kind of odd that the list of projects is on top of the header, but you can really only get to this situation by scrolling while this is open, which isn't something a user should need to do. Nothing's really broken, and clicking anywhere closes the overlay.

On the other hand, the solution creates weird edge cases. Here we have a dialog opened (which creates a dark backdrop), but when the menu opens in the dialog, the header loses the dark backdrop (it's more obvious if you see the transition happen, than just looking at a screenshot):



If you have an unrealistically sized viewport, the header can cover the dialog and menu. 



If you rotate a device sideways, this becomes more likely, though nothing really works well sideways. But with all the possible permutations of localizations, dialogs that exist now or in the future, and device sizes and orientations, this will probably break in a real-world scenario.

To me this looks like it's going to bite us.

I'm not really bothered by the original bug description (I'd be fine with leaving it as a won't fix), but if anyone is bothered by it I think there are several other approaches that could be taken that would address the issue more narrowly and with less risk of unintended side-effects.

The MatAutoComplete accepts a ScrollStrategy:
- See https://v16.material.angular.io/components/autocomplete/api#MAT_AUTOCOMPLETE_SCROLL_STRATEGY. I think this is used to specify the ScrollStrategy used by the overlay: https://v16.material.angular.io/cdk/overlay/overview#scroll-strategies. I think this can be used to tell it to just close the overlay when the user scrolls, or to not scroll, and the user can only close the autocomplete.
- I think it's possible to create an overlay that isn't at the root of the DOM, allowing it to be under the header as would be more expected. See https://v16.material.angular.io/cdk/overlay/overview#creating-overlays. I'm not sure if the autocomplete component would provide options for creating the overlay. This to me seems like a complicated option that's probably more work than it's worth, but it's probably also the most "correct" solution.

nigel-wells:
The autocomplete component is one aspect - we could live with that.

The scroll strategy isn't a complete solution either as the select overlay map  go up rather than down and cover the header without user getting a chance to scroll.

I hadn't fully considered two containers stacked. I'm sure we could adjust the CSS rules to work better for that situation.

Is it worth writing a rule to "fix" one known situation while potentially creating x number of other situations we're not aware of yet - probably not. I can't actually see anyway around it - you can't have an overlay over the entire screen and then put something on top of that but expect it to go under the header that the overlay is on top of....

In saying all this, if this was a native control, it would absolutely behave as it is and float on top of everything on the page.

I'll just revert the z-index change.

Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2316#-:-NpXL59nCuUWu9cfghg3:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

PR #2317, opened 2024-01-30
SF-2360 Update Material legacy checkbox

callumlaw17:
This PR updates mat-checkbox from legacy to the MDC version.

Styling changes were done to keep the style in line with before this change however there are some slight changes e.g. check-marks are now black instead of white for better contrast and letter-spacing on labels is slightly larger.

Some styling was also removed as it was no longer required with the new version.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2317)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.spec.ts
@@ -3,7 +3,7 @@ import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { NgModule } from '@angular/core';
 import { ComponentFixture, fakeAsync, flush, TestBed, tick } from '@angular/core/testing';
 import { FormsModule, ReactiveFormsModule, UntypedFormControl } from '@angular/forms';
-import { MatLegacyCheckbox as MatCheckbox } from '@angular/material/legacy-checkbox';
+import { MatCheckbox } from '@angular/material/checkbox';
 import { MatLegacyDialog as MatDialog, MatLegacyDialogRef as MatDialogRef } from '@angular/material/legacy-dialog';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { Canon, VerseRef } from '@sillsdev/scripture';
@@ -670,7 +670,7 @@ class TestEnvironment {
   }
 
   getRowReference(row: HTMLElement): string {
-    return row.querySelector('td .mat-checkbox-label')?.textContent?.trim() || '';
+    return row.querySelector('td mat-checkbox .mdc-label')?.textContent?.trim() || '';
   }
 
   getRowQuestion(row: HTMLElement): string {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.ts
@@ -1,6 +1,6 @@
 import { Component, ElementRef, Inject, NgZone, OnDestroy, ViewChild } from '@angular/core';
 import { AbstractControl, UntypedFormControl, UntypedFormGroup } from '@angular/forms';
-import { MatLegacyCheckbox as MatCheckbox } from '@angular/material/legacy-checkbox';
+import { MatCheckbox } from '@angular/material/checkbox';
 import {
   MatLegacyDialogConfig as MatDialogConfig,
   MatLegacyDialogRef as MatDialogRef,
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.scss
@@ -62,10 +62,6 @@
       mat-divider + .card-content {
         padding-top: 16px;
       }
-
-      mat-checkbox {
-        padding-bottom: 12px;
-      }
     }
 
     #connect-submit-button {
@@ -92,15 +88,6 @@
       font-size: 14px;
       margin-top: 10px;
     }
-  }
-}
-::ng-deep .checkbox-placeholder {
-  .mat-checkbox-inner-container {
-    visibility: hidden;
-  }
-
-  .mat-checkbox-label {
-    color: initial;
   }
 }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.scss
@@ -98,22 +98,12 @@ mat-icon {
   }
 }
 
-.mat-checkbox {
+.mat-mdc-checkbox {
   padding: 3px 0px;
 }
 
-::ng-deep .mat-checkbox-label {
+::ng-deep .mat-mdc-checkbox .mdc-label {
   line-height: normal !important;
-}
-
-::ng-deep .checkbox-placeholder {
-  .mat-checkbox-inner-container {
-    visibility: hidden;
-  }
-
-  .mat-checkbox-label {
-    color: initial;
-  }
 }
 
 #delete-project {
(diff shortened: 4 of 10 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2317)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @callumlaw17)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 67 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2317#-NpRL0LJ6G_Fia3WEL8L:-NpRL0LJ6G_Fia3WEL8M:b-m9bveh) ([raw file](https://github.com/sillsdev/web-xforge/blob/ae3ebd7ee332648248d99865ab94892d7382c811/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L67)):*
> ```SCSS
> 
> // Make checkboxes only 16x16px instead of the default 40x40px
> .mat-mdc-checkbox {
> ```

Can we not just configure the density? https://material.angular.io/guide/mdc-migration#checkbox

callumlaw17 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2317)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 67 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2317#-NpRL0LJ6G_Fia3WEL8L:-NpRSBsBCqqh3zcMx60u:b-3e1g5c) ([raw file](https://github.com/sillsdev/web-xforge/blob/ae3ebd7ee332648248d99865ab94892d7382c811/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L67)):*

Can we not just configure the density? https://material.angular.io/guide/mdc-migration#checkbox
</blockquote></details>

Unfortunately not - the migration documentation is incorrect. The only supported values for density are -1, -2, -3, -4 which are 40, 36, 32 and 28px respectively. As far as I can tell there is no way to use the density to get 16px.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2317)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @callumlaw17)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 67 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2317#-NpRL0LJ6G_Fia3WEL8L:-NpRZYKEAPVThpMACfI1:b-mi0e54) ([raw file](https://github.com/sillsdev/web-xforge/blob/ae3ebd7ee332648248d99865ab94892d7382c811/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L67)):*
<details><summary><i>Previously, callumlaw17 wrote…</i></summary><blockquote>

Unfortunately not - the migration documentation is incorrect. The only supported values for density are -1, -2, -3, -4 which are 40, 36, 32 and 28px respectively. As far as I can tell there is no way to use the density to get 16px.
</blockquote></details>

Ah, OK. Is there a significant reason why we need to keep them at 16px though? 40px actually looks fairly good to me, though it does look like some parts of our CSS would need to be updated to prevent the layout on the settings page from having too much margin between items.

I'm not opposed to customizing MDC components when needed, but when we can avoid it, we make things easier for ourselves in the future. Right now we're converting from Material legacy to MDC-based components, but not very long ago we were converting from an old MDC-based library ([@angular-mdc/web](https://npmjs.com/package/@angular-mdc/web)), and I'm sure at some point in the future Angular Material will have changes that we'll need to adjust to again (e.g. the update to Material v3). So if we can avoid making customizations, we're potentially saving ourselves work in the future when the internal implementation changes.

(Again, not against custom styles, just so long as there's a reason, which generally should be included as a comment so someone else can judge when that reason no longer applies).

As a side note, it looks to me like the ripple doesn't work very well (not centered, and overlaps text):

callumlaw17 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2317)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 67 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2317#-NpRL0LJ6G_Fia3WEL8L:-NpSmrNV-7f_xttZ0xKb:b-fki3on) ([raw file](https://github.com/sillsdev/web-xforge/blob/ae3ebd7ee332648248d99865ab94892d7382c811/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L67)):*

Ah, OK. Is there a significant reason why we need to keep them at 16px though? 40px actually looks fairly good to me, though it does look like some parts of our CSS would need to be updated to prevent the layout on the settings page from having too much margin between items.

I'm not opposed to customizing MDC components when needed, but when we can avoid it, we make things easier for ourselves in the future. Right now we're converting from Material legacy to MDC-based components, but not very long ago we were converting from an old MDC-based library ([@angular-mdc/web](https://npmjs.com/package/@angular-mdc/web)), and I'm sure at some point in the future Angular Material will have changes that we'll need to adjust to again (e.g. the update to Material v3). So if we can avoid making customizations, we're potentially saving ourselves work in the future when the internal implementation changes.

(Again, not against custom styles, just so long as there's a reason, which generally should be included as a comment so someone else can judge when that reason no longer applies).

As a side note, it looks to me like the ripple doesn't work very well (not centered, and overlaps text):

</blockquote></details>

Regardless of the density the checkbox will always be 18px (not 16px as I mistakenly said) and changing the density only affects the padding around the checkbox. Since the old checkbox didn't have this padding we already have spacing applied in other ways (i.e. row-gap or padding on containing elements) meaning that it can start to get a bit excessive, for instance see the screenshot of the feature flags dialog at 18px compared to 40px.




I have fixed the off-center ripple and decreased the size (6px off each edge rather than 12px) so that it doesn't overlap text.

nigel-wells:
@Nateowami is this one ready to move to testing or does it need another review?

nigel-wells (changes requested):
Quick lookover and test so that this can move along. 
Only observation is the check is no longer white - was this intentional or a side effect from the migration? Let's keep it white for now to avoid it coming up during testing.

> Reviewed 9 of 10 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2317#-:-NrJ-izf10x-Igrn0Xbx:b-v2s0gn)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

callumlaw17 (commented):
The check is now either black or white depending on which provides the most contrast against the background color - I think this is a change we want as it is an improvement in terms of both styling and accessibility so in my opinion we shouldn't override this new behaviour. Happy to change it back to always be white if you disagree.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2317#-:-NrYGm365bulzmdV1E9S:bk102ga)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

nigel-wells (approved):
So Material is deciding it as part of the newer version? Nice. I'm happy with it. It will need to be pointed out in the acceptance tests to the testers.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2317#-:-NrYMidv3tYib0_4LRAl:b-g48lig)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

nigel-wells (approved):
Can you do a rebase before sending it on to the testers

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2317#-:-NrYNPC6AAvcYVZDUUdK:b-bxspm3)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

callumlaw17 (commented):
Rebased and noted the checkmark colour change in the acceptance tests on the Jira ticket.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2317#-:-NrYYz7BBW7QFDkony-c:b8zpyc0)* status: 9 of 10 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @nigel-wells)

PR #2326, opened 2024-02-04
SF-2493 Fix training source language for back translations

pmachapman:
This PR fixes the language tags for alternate source and alternate training sources for back translations by updating these values after the initial sync of the alternate source or alternate training source.

This bug occurred because the Paratext API does not return the language tag for back translation projects, and so to retrieve this value the project must first be synced with Scripture Forge.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2326)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/IMachineProjectService.cs
+++ b/src/SIL.XForge.Scripture/Services/IMachineProjectService.cs
@@ -43,4 +43,5 @@ public interface IMachineProjectService
         bool preTranslate,
         CancellationToken cancellationToken
     );
+    Task UpdateTranslationSourcesAsync(string curUserId, string sfProjectId);
 }
--- a/src/SIL.XForge.Scripture/Services/MachineProjectService.cs
+++ b/src/SIL.XForge.Scripture/Services/MachineProjectService.cs
@@ -827,6 +827,71 @@ public class MachineProjectService : IMachineProjectService
         }
     }
 
+    public async Task UpdateTranslationSourcesAsync(string curUserId, string sfProjectId)
+    {
+        // Get the user secret
+        if (!(await _userSecrets.TryGetAsync(curUserId)).TryResult(out UserSecret userSecret))
+        {
+            throw new DataNotFoundException("The user secret does not exist.");
+        }
+
+        // Load the project from the realtime service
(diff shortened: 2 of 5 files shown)


josephmyers (approved):
This certainly fixes the bug, so thanks for that. It is unfortunate that the sync doesn't update these values. Why isn't it doing that? Patching it after the sync seems like a workaround to this more concerning problem, which this PR doesn't address. The JIRA description suggests not storing writingSystem in draftConfig at all, since it's difficult to keep in sync, as shown here. Should we take action on this? Thanks

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2326#-:-Nq16Q7m0Gllmx0VL34G:b-g76l1i)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

josephmyers:
To be clear, what I'm asking is, wouldn't it be better (more future-proof) to make your UpdateTranslationSourcesAsync call in the SyncService, as part of the sync itself? Is this possible?

pmachapman:
> To be clear, what I'm asking is, wouldn't it be better (more future-proof) to make your UpdateTranslationSourcesAsync call in the SyncService, as part of the sync itself? Is this possible?

We are syncing a source project (i.e. either an alternate source or training source), but updating the value in the target project. This is a one way relationship (target -> source), so the only way to do that would be by traversing all projects that reference the source project being synced, and update them.

We probably could look at updating all references to a project when it is synced (i.e. search for source, alternate training source, and training source), and update them all of changes in language, etc, but I think that is outside of the scope of this PR?

josephmyers:
Yeah that's probably out of scope. I'm not the database expert, but I would think we'd want to update all references. Otherwise, we've got one project correct and all others that reference it incorrect.

Let's say we pass this one to test, and if you think a fix for the above is important, would you create a JIRA issue for it? Thanks for your work on this

pmachapman:
> I would think we'd want to update all references. Otherwise, we've got one project correct and all others that reference it incorrect.
> Let's say we pass this one to test, and if you think a fix for the above is important, would you create a JIRA issue for it?

Done. Created [SF-2525](https://jira.sil.org/browse/SF-2525).

PR #2335, opened 2024-02-12
SF-2499 Add quick buttons for selecting books when drafting

josephmyers:
Here's a screenshot of one of the selection options activated:
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2335)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.html
@@ -2,7 +2,7 @@
   <div *ngIf="availableBooks.length > 0 && !readonly">
     <mat-card class="bulk-select">
       <div>
-        <button mat-button *ngIf="selectedBooks.length > 0" (click)="clear()">{{ t("book_select.reset") }}</button>
+        <button mat-button *ngIf="selectedBooks.length > 0" (click)="clear()">{{ t("book_select.clear") }}</button>
         <span>Select:</span>
         <mat-button-toggle-group (change)="select($event.value)" [(ngModel)]="selection">
           <mat-button-toggle value="All">{{ t("book_select.ot") }}</mat-button-toggle>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.html
@@ -2,7 +2,6 @@
   <div *ngIf="availableBooks.length > 0 && !readonly">
     <mat-card class="bulk-select">
       <div>
-        <button mat-button *ngIf="selectedBooks.length > 0" (click)="clear()">{{ t("book_select.clear") }}</button>
         <span>Select:</span>
         <mat-button-toggle-group (change)="select($event.value)" [(ngModel)]="selection">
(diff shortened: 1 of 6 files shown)


Nateowami:
Something is broken.

pmachapman:
Should we have a DC (Deuterocanon) option, like Paratext does in the Books tab of Project Properties dialog?

Nateowami:
@josephmyers I had originally specified that there should be button for All, OT, and NT. But perhaps it makes more sense to only have Clear, OT, NT, DC, and _not_ clear the selection when a button is clicked. So if you wanted NT + OT, you could clear any selection, then click the two buttons, and NT + OT would be selected. I think that's more flexible than the behavior where clicking "NT" un-selects all non-NT books. If un-selecting them is actually desired, the user can clear all, then select "NT".

josephmyers:
Sounds good, I'll get on that today

Nateowami (commented):
> Reviewed 1 of 6 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2335)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 34 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2335#-Nqc0WTj8zlR_FRou00a:-Nqc0WTk0yRcSc_4ebFi:bkb4sq4) ([raw file](https://github.com/sillsdev/web-xforge/blob/edf7199ac4a9d328d8525978e7f800b8766a4867/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L34)):*
> ```JSON
>   "book_select": {
>     "clear": "Clear",
>     "dc": "DC",
> ```

In context, these abbreviations make sense, but without context, it's hard to tell what they refer to (and therefore to translate). Maybe you could make the keys say be `deuterocanon`, `new_testament`, and `old_testament`?

pmachapman (changes requested):
Some feedback on the design:


1. This gap is gigantic!
2.  I had trouble finding this on my 1920x1024 monitor - it took me 10 or so seconds to actually hunt it down. 
  a. Maybe it would be more visible if it was inline with the bar above it (you would have to hide at smaller screen sizes), or perhaps left aligned (and right aligned for RTL languages), or may even floating right inside the list of chips (perhaps a more unorthodox way - I am not sure if it would work?).
  b. I think I didn't see it because my eye scanned from the phrase "Choose books to translate" looked down, saw the gap then moved down further to the list of chips due to their bold color
3. This top border is invisible on my monitor, which I think did not help with me finding the widget.
4. It is not clear that this is a button. I like the show/hide aspect of it, but perhaps if it was a button and to the right of NT/DC
Also, I like how DC only appears if you have Deuterocanonical/Apocryphal books.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2335#-:-NqdSn9t-3v9eveYFptG:b-umjjzq)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers)

pmachapman (approved):
:lgtm: That looks heaps better. I kind of miss the toggle on/off functionality you had previously, but I think your method is more usable.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2335#-:-Nr2iMQyC3GWrLFDuVDy:b-x6eg8r)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

PR #2338, opened 2024-02-13
SF-2524 Prevent applying drafts when editing is not permitted

pmachapman:
The SF editor interface prevents editing a chapter if:

 * The USFM is invalid
 * Editing the project is disabled
 * The ops are corrupted
 * The project is not in sync
 * The user does not have project level edit permission
 * The user does not have chapter level edit permission

This PR updates the draft preview screen to use these same checks.

As the draft viewer will be removed in the next few months, this PR does not display any warning messages about why the user cannot see the apply draft button (they could view these on the editor).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2338)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.component.html
@@ -6,7 +6,7 @@
   </section>
 
   <div [class.offline]="!isOnline">
-    <p><transloco key="draft_viewer.select_apply_to_project"></transloco></p>
+    <p *ngIf="canEdit"><transloco key="draft_viewer.select_apply_to_project"></transloco></p>
     <p><transloco key="draft_viewer.select_edit"></transloco></p>
 
     <div class="draft-viewer-body" [ngClass]="{ 'no-source': !bookHasSource }">
@@ -56,7 +56,7 @@
             {{ isDraftApplied ? t("draft_indicator_applied") : t("draft_indicator_not_applied") }}
             <mat-icon *ngIf="isDraftApplied" class="check-icon">check</mat-icon>
           </span>
-          <div *ngIf="!isDraftApplied && hasDraft" class="apply-draft-button-container">
+          <div *ngIf="!isDraftApplied && hasDraft && canEdit" class="apply-draft-button-container">
             <button mat-flat-button color="primary" (click)="applyDraft()">
               <mat-icon>auto_awesome</mat-icon>
               {{ t("apply_to_project") }}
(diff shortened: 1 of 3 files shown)


pmachapman (commented):
> I am also assuming, by your changes, that we don't want to prevent the actual draft generation. Is this something we should consider? It seems kind of pointless to allow them to generate and not apply. Aren't they just busying up the server? What do you think?

I thought on that too. I think since the new UI will allow viewing of the AI draft as a form of separate text (this is something that has been requested a lot), it is OK to allow people to view it here even if they won't be able to apply it, especially since this code will likely be removed in a few months.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2338#-:-NqYez_47yMDgLmVeWOQ:b-jq7l6o)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.component.ts` line 73 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2338#-NqWY74A7J4SPNs_SgsS:-NqYfOnoBPjnNZPbN3PE:bi6irp0) ([raw file](https://github.com/sillsdev/web-xforge/blob/2d29a4aa26b644853adba532b6fc26d3cdc29d08/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.component.ts#L73)):*

It's good that you're recognizing the duplication in these four areas. However, isn't there something we can do about this? Maybe not a base class, but these aren't that far from static methods. We also have a PermissionsService, which is easier to test than static's. It doesn't seem that difficult to move those methods there -- these mostly just need the targetProject as input. Maybe you can use it in the editor component, as well?
</blockquote></details>

I agonized over whether to refactor this. The comments hide my frustration at choosing the copy-pasta approach!

I chose to duplicate this code, as this component will be removed with the new interface, and the new interface will be using what is in the editor. I think refactoring of this code should take into account the architecture of the new UI (which I am still unfamiliar with).

If this code was staying, I would have implemented a shared component that would have displayed the warning messages on the editor, and perhaps had the logic there.

If you think I should still refactor, let me know, and I will - it was difficult for me to weigh the future considerations when doing this (i.e., if Colin's PR had the new draft UI, I would not be writing this code, and instead waiting for that PR to get merged).

josephmyers (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2338)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.component.ts` line 73 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2338#-NqWY74A7J4SPNs_SgsS:-Nq_ajZv8KpLDe_yHyxE:b-j9o2k4) ([raw file](https://github.com/sillsdev/web-xforge/blob/0c86d9f712bf8545d476677785f7b572f436c70d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.component.ts#L73)):*

I agonized over whether to refactor this. The comments hide my frustration at choosing the copy-pasta approach!

I chose to duplicate this code, as this component will be removed with the new interface, and the new interface will be using what is in the editor. I think refactoring of this code should take into account the architecture of the new UI (which I am still unfamiliar with).

If this code was staying, I would have implemented a shared component that would have displayed the warning messages on the editor, and perhaps had the logic there.

If you think I should still refactor, let me know, and I will - it was difficult for me to weigh the future considerations when doing this (i.e., if Colin's PR had the new draft UI, I would not be writing this code, and instead waiting for that PR to get merged).
</blockquote></details>

Alright, that makes sense. Thanks!

PR #2346, opened 2024-02-19
SF-2636 Add preliminary font support

Nateowami:
This loads fonts from fonts.languagetechnology.org, based on the font specified on the Paratext project. I only added support for the most used fonts on Scripture Forge, as I want to keep this change small and make sure it works well before expanding the list too much.

Things I did not address:
- Preview draft page
- Question dialog
- Text chooser dialog

Initially I was going to add support for them, but they were going to take slightly more work, and I wanted to keep this change as small as possible, so I can easily address any changes that need to be made, without it being held up. We can follow up this PR with fixing those other components.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
@@ -6,7 +6,7 @@ quill-editor {
   height: 100%;
   > .ql-container {
     > .ql-editor {
-      font-family: Roboto, 'Noto Sans Kayah Li', sans-serif;
+      font-family: Roboto, sans-serif;
       line-height: 1.6;
       width: 100%;
       word-break: break-word;
@@ -33,7 +33,7 @@ quill-editor {
 .read-only-editor > .ql-container {
   > .ql-editor {
     usx-segment {
-      padding: 0.2em 0;
+      line-height: 1.8;
     }
 
     .highlight-segment {
@@ -278,4 +278,12 @@ quill-editor.rtl .highlight-para usx-segment:first-child:before {
 usx-segment.draft {
   color: $draft-color;
   font-style: italic;
+}
+
+usx-segment {
+  font-family: var(--project-font);
+  &::before,
+  &::after {
+    font-family: Roboto, sans-serif;
+  }
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.html
@@ -6,4 +6,5 @@
   (loaded)="onLoaded()"
   [isRightToLeft]="isRightToLeft"
   [fontSize]="fontSize"
+  [style.--project-font]="fontService.getFontFamilyFromProject(projectDoc)"
 ></app-text>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.ts
@@ -5,6 +5,8 @@ import isEqual from 'lodash-es/isEqual';
 import { Canon, VerseRef } from '@sillsdev/scripture';
 import { fromEvent, Subscription } from 'rxjs';
 import { SubscriptionDisposable } from 'xforge-common/subscription-disposable';
+import { SFProjectProfileDoc } from 'src/app/core/models/sf-project-profile-doc';
+import { FontService } from 'xforge-common/font.service';
 import { TextDocId } from '../../../core/models/text-doc';
 import { TextComponent } from '../../../shared/text/text.component';
 import { getVerseStrFromSegmentRef, verseRefFromMouseEvent } from '../../../shared/utils';
@@ -19,6 +21,7 @@ export class CheckingTextComponent extends SubscriptionDisposable {
   @Output() questionVerseSelected = new EventEmitter<VerseRef>();
   @Input() isRightToLeft: boolean = false;
   @Input() fontSize?: string;
+  @Input() projectDoc?: SFProjectProfileDoc;
 
   private clickSubs: Subscription[] = [];
   private _activeVerse?: VerseRef;
@@ -26,6 +29,10 @@ export class CheckingTextComponent extends SubscriptionDisposable {
   private _id?: TextDocId;
   private _questionVerses?: VerseRef[];
   private _placeholder?: string;
+
+  constructor(readonly fontService: FontService) {
+    super();
+  }
 
   @Input() set placeholder(value: string) {
     this._placeholder = value;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html
@@ -145,6 +145,7 @@
                     (questionVerseSelected)="verseRefClicked($event)"
                     [isRightToLeft]="isRightToLeft"
                     [class.hidden]="hideChapterText"
+                    [projectDoc]="projectDoc"
                   ></app-checking-text>
                 </div>
                 <div class="scripture-audio-player-wrapper" *ngIf="showScriptureAudioPlayer">
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
@@ -94,6 +94,7 @@
                     (updated)="onSourceUpdated($event.delta != null)"
                     [isRightToLeft]="isSourceRightToLeft"
                     [fontSize]="sourceFontSize"
+                    [style.--project-font]="fontService.getFontFamilyFromProject(sourceProjectDoc)"
                   ></app-text>
                 </div>
               </as-split-area>
@@ -177,6 +178,7 @@
                     [fontSize]="fontSize"
                     [selectableVerses]="showAddCommentUI"
                     [ngClass]="{ 'comment-enabled-editor': showAddCommentUI }"
+                    [style.--project-font]="fontService.getFontFamilyFromProject(projectDoc)"
                   ></app-text>
                   <app-suggestions
                     class="mat-elevation-z2"
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
@@ -91,6 +91,7 @@ import { OnlineStatusService } from 'xforge-common/online-status.service';
 import { UserService } from 'xforge-common/user.service';
 import { getLinkHTML, issuesEmailTemplate, objectId } from 'xforge-common/utils';
 import { XFValidators } from 'xforge-common/xfvalidators';
+import { FontService } from 'xforge-common/font.service';
 import { environment } from '../../../environments/environment';
 import { isString } from '../../../type-utils';
 import { defaultNoteThreadIcon, NoteThreadDoc, NoteThreadIcon } from '../../core/models/note-thread-doc';
@@ -204,7 +205,7 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
   private insertSuggestionEnd: number = -1;
   private bottomSheetRef?: MatBottomSheetRef;
   private currentUserDoc?: UserDoc;
-  private projectDoc?: SFProjectProfileDoc;
+  projectDoc?: SFProjectProfileDoc;
   private projectUserConfigDoc?: SFProjectUserConfigDoc;
   private paratextUsers: ParatextUserProfile[] = [];
   private projectUserConfigChangesSub?: Subscription;
@@ -212,7 +213,7 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
   private targetLabel?: string;
   private text?: TextInfo;
   private sourceText?: TextInfo;
-  private sourceProjectDoc?: SFProjectProfileDoc;
+  sourceProjectDoc?: SFProjectProfileDoc;
   private sourceLoaded: boolean = false;
   private targetLoaded: boolean = false;
   private _targetFocused: boolean = false;
@@ -248,6 +249,7 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
     private readonly onlineStatusService: OnlineStatusService,
     private readonly translationEngineService: TranslationEngineService,
     readonly i18n: I18nService,
+    readonly fontService: FontService,
     readonly featureFlags: FeatureFlagService,
     private readonly reportingService: ErrorReportingService,
     private readonly activatedProject: ActivatedProjectService,
--- a/src/SIL.XForge.Scripture/ClientApp/src/index.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/index.html
@@ -12,10 +12,7 @@
     </script>
 
     <link href="https://fonts.googleapis.com/icon?family=Material+Icons|Material+Icons+Outlined" rel="stylesheet" />
-    <link
-      href="https://fonts.googleapis.com/css2?family=Noto+Sans+Kayah+Li&family=Roboto:wght@300;400;500&display=swap"
-      rel="stylesheet"
-    />
+    <link href="https://fonts.googleapis.com/css?family=Roboto:300,400,500" rel="stylesheet" />
     <meta charset="utf-8" />
     <title>Scripture Forge</title>
     <base href="/" />
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.spec.ts
@@ -1,0 +1,58 @@
+import { FontService, FONT_FACE_DEFINITIONS, FONT_FACE_FALLBACKS } from './font.service';
+
+// Mocking the document with ts-mockito does not work because the type definitions for document.fonts does not include
+// the add method
+class FakeDocument {
+  addCount = 0;
+  fonts = {
+    add: (_: any) => {
+      this.addCount++;
+    }
+  };
+}
+
+describe('FontService', () => {
+  let fontService: FontService;
+
+  beforeEach(() => {
+    fontService = new FontService(new FakeDocument() as any);
+  });
+
+  it('should default to Charis SIL when font is not specified', () => {
+    expect(fontService.getCSSFontName(undefined)).toEqual('Charis SIL');
+    expect(fontService.getCSSFontName('')).toEqual('Charis SIL');
+  });
+
+  it('should default to Charis SIL when font is not recognized', () => {
+    expect(fontService.getCSSFontName('zyz123')).toEqual('Charis SIL');
+  });
+
+  it('should default to Charis SIL for proprietary serif fonts', () => {
+    expect(fontService.getCSSFontName('Times New Roman')).toEqual('Charis SIL');
+    expect(fontService.getCSSFontName('Cambria')).toEqual('Charis SIL');
+    expect(fontService.getCSSFontName('Sylfaen')).toEqual('Charis SIL');
+  });
+
+  it('should default to Andika for proprietary sans-serif fonts', () => {
+    expect(fontService.getCSSFontName('Arial')).toEqual('Andika');
+    expect(fontService.getCSSFontName('Verdana')).toEqual('Andika');
+    expect(fontService.getCSSFontName('Tahoma')).toEqual('Andika');
+  });
+
+  it('should fall back from Annapurna SIL Nepal to Annapurna SIL', () => {
+    expect(fontService.getCSSFontName('Annapurna SIL Nepal')).toEqual('Annapurna SIL');
+  });
+
+  it('should not have any broken font fallbacks', () => {
+    // make sure all fallbacks are actually defined
+    expect(Object.values(FONT_FACE_FALLBACKS).filter(fallback => FONT_FACE_DEFINITIONS[fallback] == null)).toEqual([]);
+  });
+
+  it('should only load each font once', () => {
+    fontService.getCSSFontName('Charis SIL');
+    fontService.getCSSFontName('Charis SIL');
+    fontService.getCSSFontName('Charis SIL');
+
+    expect(((fontService as any).document as FakeDocument).addCount).toEqual(1);
+  });
+});
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts
@@ -1,0 +1,135 @@
+import { Inject, Injectable } from '@angular/core';
+import { SFProjectProfileDoc } from 'src/app/core/models/sf-project-profile-doc';
+import { DOCUMENT } from './browser-globals';
+
+const DEFAULT_SERIF_FONT_FAMILY = 'Charis SIL';
+const DEFAULT_SANS_SERIF_FONT_FAMILY = 'Andika';
+const DEFAULT_FONT_FAMILY = DEFAULT_SERIF_FONT_FAMILY;
+
+export const FONT_FACE_FALLBACKS: { [key: string]: string } = {
+  // Proprietary fonts typically used for English (where we suspect the specified font isn't that critical)
+  Arial: DEFAULT_SANS_SERIF_FONT_FAMILY,
+  'Times New Roman': DEFAULT_SERIF_FONT_FAMILY,
+  Verdana: DEFAULT_SANS_SERIF_FONT_FAMILY,
+  'Arial Unicode MS': DEFAULT_SANS_SERIF_FONT_FAMILY,
+  Cambria: DEFAULT_SERIF_FONT_FAMILY,
+  Sylfaen: DEFAULT_SERIF_FONT_FAMILY,
+  Tahoma: DEFAULT_SANS_SERIF_FONT_FAMILY,
+  Calibri: DEFAULT_SANS_SERIF_FONT_FAMILY,
+  'Maiandra GD': DEFAULT_SANS_SERIF_FONT_FAMILY,
+
+  // Freely available fonts we have yet to properly support (temporary fallbacks)
+  'Suranna UI': DEFAULT_SERIF_FONT_FAMILY, // Suranna is available via Google Fonts
+  'Andika New Basic': 'Andika',
+  'Andika New Basic Compact': 'Andika',
+  'Andika Eng-Lit': 'Andika',
+  'Andika SEB': 'Andika',
+  'Annapurna SIL Nepal': 'Annapurna SIL',
+  'Annapurna SIL Thami': 'Annapurna SIL',
+  Gentium: 'Gentium Plus',
+  'Gentium Basic': 'Gentium Plus',
+  'Gentium Book Basic': 'Gentium Book Plus',
+  'Lateef Haz Kas Low': 'Lateef',
+  'Lateef Sindhi': 'Lateef',
+  'Scheherazade New Persian': 'Scheherazade New',
+
+  // Unclear category
+  'Karenni Unicode': 'Kay Pho Du',
+  'Adobe Arabic': 'Scheherazade New',
+  'Simplified Arabic': 'Scheherazade New'
+};
+
+export const FONT_FACE_DEFINITIONS: { [key: string]: string } = {
+  'Abyssinica SIL': 'https://fonts.languagetechnology.org/fonts/sil/abyssinicasil/web/AbyssinicaSIL-Regular.woff2',
+  Akatab: 'https://fonts.languagetechnology.org/fonts/sil/akatab/web/Akatab-Regular.woff2',
+  Alkalami: 'https://fonts.languagetechnology.org/fonts/sil/alkalami/web/Alkalami-Regular.woff2',
+  Andika: 'https://fonts.languagetechnology.org/fonts/sil/andika/web/Andika-Regular.woff2',
+  'Annapurna SIL': 'https://fonts.languagetechnology.org/fonts/sil/annapurnasil/web/AnnapurnaSIL-Regular.woff2',
+  'Awami Nastaliq': 'https://fonts.languagetechnology.org/fonts/sil/awaminastaliq/web/AwamiNastaliq-Regular.woff2',
+  Badami: 'https://fonts.languagetechnology.org/fonts/other/badami/woff/Badami-Regular.woff2',
+  Bailey: 'https://fonts.languagetechnology.org/fonts/other/bailey/web/Bailey-Regular.woff2',
+  'Charis SIL': 'https://fonts.languagetechnology.org/fonts/sil/charissil/web/CharisSIL-Regular.woff2',
+  'Dai Banna SIL': 'https://fonts.languagetechnology.org/fonts/sil/daibannasil/web/DaiBannaSIL-Regular.woff2',
+  'Doulos SIL': 'https://fonts.languagetechnology.org/fonts/sil/doulossil/web/DoulosSIL-Regular.woff2',
+  Eeyek: 'https://fonts.languagetechnology.org/fonts/sil/eeyek/web/Eeyek-Regular.woff',
+  'Ezra SIL': 'https://fonts.languagetechnology.org/fonts/sil/ezrasil/web/SILEOT.woff',
+  'Ezra SIL SR': 'https://fonts.languagetechnology.org/fonts/sil/ezrasilsr/web/SILEOTSR.woff',
+  'Galatia SIL': 'https://fonts.languagetechnology.org/fonts/sil/galatiasil/web/GalSILR.woff',
+  'Gentium Book Plus':
+    'https://fonts.languagetechnology.org/fonts/sil/gentiumbookplus/web/GentiumBookPlus-Regular.woff2',
+  'Gentium Plus': 'https://fonts.languagetechnology.org/fonts/sil/gentiumplus/web/GentiumPlus-Regular.woff2',
+  Harmattan: 'https://fonts.languagetechnology.org/fonts/sil/harmattan/web/Harmattan-Regular.woff2',
+  'Japa Sans Oriya': 'https://fonts.languagetechnology.org/fonts/sil/japasansoriya/woff/JapaSansOriya-Regular.woff2',
+  Kanchenjunga: 'https://fonts.languagetechnology.org/fonts/sil/kanchenjunga/web/Kanchenjunga-Regular.woff2',
+  'Kay Pho Du': 'https://fonts.languagetechnology.org/fonts/sil/kayphodu/web/KayPhoDu-Regular.woff2',
+  Lateef: 'https://fonts.languagetechnology.org/fonts/sil/lateef/web/Lateef-Regular.woff2',
+  'Lisu Bosa': 'https://fonts.languagetechnology.org/fonts/sil/lisubosa/web/LisuBosa-Regular.woff2',
+  Mingzat: 'https://fonts.languagetechnology.org/fonts/sil/mingzat/web/Mingzat-Regular.woff2',
+  Namdhinggo: 'https://fonts.languagetechnology.org/fonts/sil/namdhinggo/web/Namdhinggo-Regular.woff2',
+  Narnoor: 'https://fonts.languagetechnology.org/fonts/sil/narnoor/web/Narnoor-Regular.woff2',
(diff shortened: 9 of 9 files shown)


nigel-wells (changes requested):
This works pretty well! There was a small rendering adjustment the first time the editor loaded but was fine after that.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-:-NrIgtF05PtLNYi0ytKd:b1o8p52)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 35 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NrIgmzX0eIWuwUiuO0_:-NrIgmzX0eIWuwUiuO0a:bxvf1g) ([raw file](https://github.com/sillsdev/web-xforge/blob/02a558dcead4726fca676980961dab753efd6c39/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L35)):*
> ```XML
> export class FontService {
>   // Map of project font names to CSS font families
>   loadedFontFamilies: Map<string, string> = new Map();
> ```

Could be `private`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 53 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NrIcmFnA7YfgMz8lE6B:-NrIcmFnA7YfgMz8lE6C:bb3b07v) ([raw file](https://github.com/sillsdev/web-xforge/blob/02a558dcead4726fca676980961dab753efd6c39/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L53)):*
> ```XML
>     const fontDefinition = FONT_FACE_DEFINITIONS[projectFont];
>     if (fontDefinition == null) {
>       console.error(`No font definition for ${projectFont}`);
> ```

A warning might be more appropriate as nothing actually broke. Also, the console is littered with this message if the font isn't found.

Would it be worth doing a silent call to bugsnag as well so we can keep an eye on fonts users are trying to use and see if we can support them.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NrIeKg5E1sYU4v9koQT:-NrIeKg5E1sYU4v9koQU:b-yuiene) ([raw file](https://github.com/sillsdev/web-xforge/blob/02a558dcead4726fca676980961dab753efd6c39/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L54)):*
> ```XML
>     if (fontDefinition == null) {
>       console.error(`No font definition for ${projectFont}`);
>       return DEFAULT_FONT_FAMILY;
> ```

Are you sure this is a good idea? We're basically saying SF only supports 5 fonts regardless of what may be installed on the device. Some of my projects are default windows ones (Arial) but now they all render as Charis SIL. Perhaps return the project font first and `DEFAULT_FONT_FAMILY` second for the fallback. This may be more of an issue for different scripts that Charis doesn't suport at all.

Nateowami:
Oh, and I didn't actually test my latest changes, since I'm in the middle of a workshop, and my laptop froze and had to be restarted, some I'm doubly hampered.

nigel-wells (changes requested):
> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 53 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NrIcmFnA7YfgMz8lE6B:-NrXnfcK4jTCkYpNWKyU:bu3otc4) ([raw file](https://github.com/sillsdev/web-xforge/blob/02a558dcead4726fca676980961dab753efd6c39/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L53)):*

Good point. Changed to warning, and only showing once.

Instead of logging to Bugsnag, I think we can just monitor what fonts are specified on projects. We have a Mongo query for that. See the spreadsheet "SF live font usage" which I updated about an hour ago.
</blockquote></details>

They mongo query will mainly show you what is set but not what users may be trying to use more regularly. Doesn't need to hold this up but I do think bugsnag would give you a greater picture of how often an unsupported font is trying to be used.

It would be great for your charts to exclude the now support fonts so it is easier to see what is next popular to tackle.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NrIeKg5E1sYU4v9koQT:-NrXpC-V1Av_CZ0d0dJt:bj6x570) ([raw file](https://github.com/sillsdev/web-xforge/blob/02a558dcead4726fca676980961dab753efd6c39/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L54)):*

This is a good point, though I don't particularly like that it means behavior will be different between devices. I don't like the idea that something will work as expected on one device, but then behave differently for e.g. community checkers that are invited.

My medium-term plan was to map proprietary fonts to the most similar freely licensed font. For sans-serif fonts, that's likely to be Liberation Sans, or if we want to stick with SIL fonts, [Andika](https://software.sil.org/andika/) (since it's hosted on fonts.languagetechnology.org). For most projects that have chosen common proprietary fonts, I don't think it's because there's some important technical reason why they need that font (there are probably some exceptions for specialized proprietary fonts, but we'll have to handle those as they come up).

Then longer-term I'd thought we should add a warnings/recommendations section to the project overview page, where there would be notices for things that aren't set up correctly. We would then add a warning when someone has chosen a font that we don't fully support exactly as they specified it, and indicate what we're using as a fallback. Users would be advised to contact us if that fallback does not satisfy their needs.
</blockquote></details>

I get that a device without the font could be different from a device with the font. I'd say this is still better than both devices having a bad font when one of the devices should work fine.

As this change is going to effect every single project once it is merged in I think we need to tread carefully. Could we feature flag it and get some specific users to test it out?

I like your longer-term plan - maybe some of that could come into the immediate-plan. Maybe we have a setting to turn on/off font support for those who want to try or those who want to disable it as a "quick fix" if they find there is an issue for them. This would be better than us having to roll out a hotfix. I guess we could delete the font files as a quick fix as well.

Nateowami:
I think you are considering this a regression for some projects, but I don't see that. Can you give an example setup where it would be worse after this change than before? Browsers will fall back to local fonts if the glyphs aren't in Charis. The only change is the default is now Charis instead of Roboto.

nigel-wells (approved):
I'm thinking of any project who was very intentional about changing the font to something other than the 5 setup in this PR. 

I've done some more reading on missing glyphs and, you're right, this won't be as big of an issue as I first thought. I knew fonts had a fallback when missing but I didn't realise missing glyphs also would use the same fallback - that's great!


:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-:-NrYGdIK4K0sqV0CBC0y:ba8xi2v)* status: 6 of 7 files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NrIeKg5E1sYU4v9koQT:-NrYLfn-593Btjfp04lk:bcq2zn4) ([raw file](https://github.com/sillsdev/web-xforge/blob/02a558dcead4726fca676980961dab753efd6c39/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L54)):*

I get that a device without the font could be different from a device with the font. I'd say this is still better than both devices having a bad font when one of the devices should work fine.

As this change is going to effect every single project once it is merged in I think we need to tread carefully. Could we feature flag it and get some specific users to test it out?

I like your longer-term plan - maybe some of that could come into the immediate-plan. Maybe we have a setting to turn on/off font support for those who want to try or those who want to disable it as a "quick fix" if they find there is an issue for them. This would be better than us having to roll out a hotfix. I guess we could delete the font files as a quick fix as well.
</blockquote></details>

As per other comments, happy to resolve this now. It would be good to get the medium/long term ideas capture in Jira so they don't get lost.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346)* status: 3 of 7 files reviewed, 2 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/_text.scss` line 287 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NrcPnSB3BwdU_oR4llV:-NrcPnSB3BwdU_oR4llW:bhlfl2k) ([raw file](https://github.com/sillsdev/web-xforge/blob/7480fd4ad2238ebc89a6fdd5cb4f0d6635d69d7f/src/SIL.XForge.Scripture/ClientApp/src/_text.scss#L287)):*
> ```SCSS
>   &::before,
>   &::after {
>     font-family: Roboto, 'Noto Sans Kayah Li', sans-serif;
> ```

I've added this to make the USFM popups revert to the normal font.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 53 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NrIcmFnA7YfgMz8lE6B:-NrcPhSR2pCzyFe1kZYI:bu76i95) ([raw file](https://github.com/sillsdev/web-xforge/blob/02a558dcead4726fca676980961dab753efd6c39/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L53)):*

They mongo query will mainly show you what is set but not what users may be trying to use more regularly. Doesn't need to hold this up but I do think bugsnag would give you a greater picture of how often an unsupported font is trying to be used.

It would be great for your charts to exclude the now support fonts so it is easier to see what is next popular to tackle.
</blockquote></details>

We may want to do that eventually, but a naive solution would create a ton of noise, and it would be difficult to process the information. Also, I think we will add support for a lot of fonts in the near future. So it would make more sense to do once the unsupported number is smaller.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NrIeKg5E1sYU4v9koQT:-NrcQN6J3-sFF3BUvFvs:bc9bse4) ([raw file](https://github.com/sillsdev/web-xforge/blob/02a558dcead4726fca676980961dab753efd6c39/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L54)):*

As per other comments, happy to resolve this now. It would be good to get the medium/long term ideas capture in Jira so they don't get lost.
</blockquote></details>

I've added some TODOs to this file. Do you think that's sufficient?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 20 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NrcIo5v8egvMw2CIAzm:-NrcIo5v8egvMw2CIAzn:b-v7tjxp) ([raw file](https://github.com/sillsdev/web-xforge/blob/7480fd4ad2238ebc89a6fdd5cb4f0d6635d69d7f/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L20)):*
> ```XML
>     "url('https://fonts.languagetechnology.org/fonts/sil/annapurnasil/web/AnnapurnaSIL-Regular.woff2') format('woff2')"
>   ),
>   ['Awami Nastaliq']: new FontFace(
> ```

I added this font after I realized it was one that was specifically requested by a user.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346)* status: 3 of 8 files reviewed, 2 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 35 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NrcoFsNDUUnxN7qiKHE:-NrcoFsOCIV_GonieAxE:bpopf93) ([raw file](https://github.com/sillsdev/web-xforge/blob/277c8a0d8a29e250b9e4c03f064549e274689001/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L35)):*
> ```XML
>   // TODO Warn that this is a fallback font
>   // This font was only added to prevent a regression, since a Karenni font had previously been added at user request
>   ['Karenni Unicode']: new FontFace(
> ```

The longer term plan would be to create a separate list of font fallbacks, like:
``` js
{
  'Karenni Unicode': 'Kay Pho Du'
}
```
...and then use it to map user-specified to fonts to actual fonts. When a fallback is used, that would be the time to show the user a warning.

Nateowami (commented):
@nigel-wells Ready for re-review, I think. Now with support for a Karenni typeface.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-:-NrcqTTsCiPH_Mrknqpk:b-93bsxh)* status: 3 of 8 files reviewed, 2 unresolved discussions (waiting on @nigel-wells)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346)* status: 3 of 9 files reviewed, 2 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 41 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-Nrhn_wM6tfyC44bK7YR:-Nrhn_wM6tfyC44bK7YS:b976gy) ([raw file](https://github.com/sillsdev/web-xforge/blob/6f4e89d12d19a40c7db71322c39c681d2fb56c71/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L41)):*
> ```XML
> };
> 
> const FONT_FACE_DEFINITIONS: { [key: string]: string } = {
> ```

I created this list by running this hideous script in the root of https://github.com/silnrsi/fonts, saved as `index.mjs`, run with Node v18.17.0.

It's a mess; I'm just leaving it here for anyone in the future who wants to quickly re-generate the list as the list of published fonts changes.

``` js
import * as fontFamililes from './families.json' assert { type: 'json' };

const fonts = fontFamililes.default;

const results = {};

for (const [key, value] of Object.entries(fonts)) {
  if (value.distributable !== true) {
    continue;
  }

  const defaultFileName = value.defaults?.['woff2'] ?? value.defaults?.['woff'];
  const defaultFileUrl = value.files?.[defaultFileName]?.flourl;

  if (results[value.family] == null) {
    results[value.family] = [];
  }
  results[value.family].push({
    fileName: defaultFileName,
    fileUrl: defaultFileUrl
  });
}

const results2 = {};

for (const [key, value] of Object.entries(results)) {
  if (value.length === 1) {
    results2[key] = value[0].fileUrl;
  } else if (value.length > 1) {
    const matchingFiles = value.filter(file => /Regular/.test(file.fileName));
    if (matchingFiles.length === 1) {
      results2[key] = matchingFiles[0].fileUrl;
    } else if (matchingFiles.length > 1) {
      console.log(`Multiple Regular files found for ${key}: ${matchingFiles.map(file => file.fileName).join(', ')}`)
    }
  } else if (value.length === 0) {
    console.log(`No matching files found for ${key}`)
  }
}

console.log(JSON.stringify(results2, null, 2))  
```

Nateowami:
I should mention that we're not handling anything other than the regular version of the font, so italics and bold are going to be a bit off. This is left as future work. Hence "preliminary" in the title.

nigel-wells (changes requested):
Some great improvements here. Just one query around tests.

On a side note, I've noticed the editor performance is not great - can you check that at your end. It seems fine on `master`.

> Reviewed 3 of 5 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-:-Nrhu5xS2MWALzbqa7iE:b203uhq)* status: all files reviewed, 4 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 138 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-Nrhsorc81UfvEleKnee:-Nrhsorc81UfvEleKnef:b-tugbcb) ([raw file](https://github.com/sillsdev/web-xforge/blob/02a558dcead4726fca676980961dab753efd6c39/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L138)):*
> ```HTML
>             <as-split-area [size]="100" [maxSize]="scriptureAreaMaxSize" [dir]="i18n.direction">
>               <div class="panel-content" #scripturePanelContainer>
>                 
> ```

This should be a separate PR don't you think? 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 54 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NrIeKg5E1sYU4v9koQT:-NrhsHwHDaagoaj2omih:b-l78ngl) ([raw file](https://github.com/sillsdev/web-xforge/blob/02a558dcead4726fca676980961dab753efd6c39/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L54)):*

I've added some TODOs to this file. Do you think that's sufficient?
</blockquote></details>

I was thinking adding them to the Jira backlog - they may be forgotten in there but more easier to find than here. In saying that, you've added a lot more support to this since the PR was first opened so not as big a deal.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 41 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-Nrhn_wM6tfyC44bK7YR:-Nrhqd9fAsNYsgYK5p56:b3svi4w) ([raw file](https://github.com/sillsdev/web-xforge/blob/6f4e89d12d19a40c7db71322c39c681d2fb56c71/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L41)):*

I created this list by running this hideous script in the root of https://github.com/silnrsi/fonts, saved as `index.mjs`, run with Node v18.17.0.

It's a mess; I'm just leaving it here for anyone in the future who wants to quickly re-generate the list as the list of published fonts changes.

``` js
import * as fontFamililes from './families.json' assert { type: 'json' };

const fonts = fontFamililes.default;

const results = {};

for (const [key, value] of Object.entries(fonts)) {
  if (value.distributable !== true) {
    continue;
  }

  const defaultFileName = value.defaults?.['woff2'] ?? value.defaults?.['woff'];
  const defaultFileUrl = value.files?.[defaultFileName]?.flourl;

  if (results[value.family] == null) {
    results[value.family] = [];
  }
  results[value.family].push({
    fileName: defaultFileName,
    fileUrl: defaultFileUrl
  });
}

const results2 = {};

for (const [key, value] of Object.entries(results)) {
  if (value.length === 1) {
    results2[key] = value[0].fileUrl;
  } else if (value.length > 1) {
    const matchingFiles = value.filter(file => /Regular/.test(file.fileName));
    if (matchingFiles.length === 1) {
      results2[key] = matchingFiles[0].fileUrl;
    } else if (matchingFiles.length > 1) {
      console.log(`Multiple Regular files found for ${key}: ${matchingFiles.map(file => file.fileName).join(', ')}`)
    }
  } else if (value.length === 0) {
    console.log(`No matching files found for ${key}`)
  }
}

console.log(JSON.stringify(results2, null, 2))  
```
</blockquote></details>

Could save this as a local script in the repo - it is nice to be able to quickly produce it though.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 130 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NrhsCT9DIAV77AEEKD9:-NrhsCT9DIAV77AEEKDA:b-b0nr6u) ([raw file](https://github.com/sillsdev/web-xforge/blob/6f4e89d12d19a40c7db71322c39c681d2fb56c71/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L130)):*
> ```XML
> 
>   // Basic check to make sure no fallbacks weren't actually defined. To be called by tests.
>   getBrokenFontFallbacks(): string[] {
> ```

Can't this just be in the test file and export the consts?

Nateowami (commented):
Looking into the performance issues. It seems like having the mix of multiple fonts is what makes it laggy. I'm not sure why.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-:-NrlQjE_7eEBziqNIJNQ:bvc51li)* status: 7 of 9 files reviewed, 3 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 138 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-Nrhsorc81UfvEleKnee:-Nrk54jYAsjYaEIzH3va:bdgvaxu) ([raw file](https://github.com/sillsdev/web-xforge/blob/02a558dcead4726fca676980961dab753efd6c39/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L138)):*

This should be a separate PR don't you think?
</blockquote></details>

You're seeing a diff from a rebase. Look at the whole range of changes and you'll see I've added just one line.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 130 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NrhsCT9DIAV77AEEKD9:-NrlQb2i8eIt1MAXpe8B:b8e62m8) ([raw file](https://github.com/sillsdev/web-xforge/blob/6f4e89d12d19a40c7db71322c39c681d2fb56c71/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L130)):*

Can't this just be in the test file and export the consts?
</blockquote></details>

I thought about that but it seemed weird to export them. But this is odd too. Changed.

nigel-wells (commented):
Ok. Let me know when ready for me to test it out again - everything else is good to go.

> Reviewed 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-:-NrmSUxu5JvT-dQMls4C:bebelie)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 138 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-Nrhsorc81UfvEleKnee:-NrmRd5_7CEKm1HqtLPM:b5k1zsq) ([raw file](https://github.com/sillsdev/web-xforge/blob/02a558dcead4726fca676980961dab753efd6c39/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L138)):*

You're seeing a diff from a rebase. Look at the whole range of changes and you'll see I've added just one line.
</blockquote></details>

Weird - its no longer in the diff.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 20 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NrcIo5v8egvMw2CIAzm:-NrmSfZ05A1dMZ99mb99:b-6kx4ph) ([raw file](https://github.com/sillsdev/web-xforge/blob/7480fd4ad2238ebc89a6fdd5cb4f0d6635d69d7f/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L20)):*

I added this font after I realized it was one that was specifically requested by a user.
</blockquote></details>

Reviewable wants me to respond to this. There, I have responded :)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346)* status: 7 of 9 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 116 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-Nrv44IrBPDi6RleM9tA:-Nrv44IrBPDi6RleM9tB:b-lvhjz9) ([raw file](https://github.com/sillsdev/web-xforge/blob/505eac5f27febf549dc23d28046cde66c8f07eec/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L116)):*
> ```XML
>     let fontUrl = FONT_FACE_DEFINITIONS[projectFont];
>     this.addFontFamilyToDocument(projectFont, fontUrl);
>     this.loadedFontFamilies.set(projectFont, projectFont);
> ```

:facepalm: I used a set to make sure we only add the font once, but I was only ever checking whether the font had been added to the set, and never actually adding to it. Which means every Angular digest cycle was adding the font to the document all over again. Fixed.

Nateowami:
There are some issues that come up when we start using a custom font, specifically related to the height of each line. I'm not sure how critical we should consider this.

nigel-wells (changes requested):
Is it any better if you apply the font to everything in the editor - so that the font height is applied correctly - and then adjust elements you specifically want to be Roboto?

> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-:-NsGdH-97iQjpg5pB-Uo:bqcsfw6)* status: all files reviewed, 6 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/font.service.spec.ts` line 0 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NsGctLpAW4WAZtu5PIW:-NsGctLpAW4WAZtu5PIX:b-7lj3jd) ([raw file](https://github.com/sillsdev/web-xforge/blob/505eac5f27febf549dc23d28046cde66c8f07eec/src/SIL.XForge.Scripture/ClientApp/src/font.service.spec.ts#L0)):*
Just noticed that this test file is in the wrong folder
___
*[`src/SIL.XForge.Scripture/ClientApp/src/font.service.spec.ts` line 56 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NsGd7EA9gco8tvwOz4U:-NsGd7EA9gco8tvwOz4V:b-dnk3n0) ([raw file](https://github.com/sillsdev/web-xforge/blob/505eac5f27febf549dc23d28046cde66c8f07eec/src/SIL.XForge.Scripture/ClientApp/src/font.service.spec.ts#L56)):*
> ```XML
>     fontService.getCSSFontName('Charis SIL');
> 
>     expect(((fontService as any).document as FakeDocument).addCount).toEqual(1);
> ```

Would a spy have worked easier for this?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 100 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NsGbewYA6uvuTBUbTCq:-NsGbewYA6uvuTBUbTCr:b7d79ca) ([raw file](https://github.com/sillsdev/web-xforge/blob/505eac5f27febf549dc23d28046cde66c8f07eec/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L100)):*
> ```XML
>    * @returns A font family that can be used in CSS to specify the font.
>    */
>   getCSSFontName(projectFont: string | undefined): string {
> ```

nit  Can be private
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 116 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-Nrv44IrBPDi6RleM9tA:-NsGbxA8B4iyD-itW5aQ:brjz3y4) ([raw file](https://github.com/sillsdev/web-xforge/blob/505eac5f27febf549dc23d28046cde66c8f07eec/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L116)):*

:facepalm: I used a set to make sure we only add the font once, but I was only ever checking whether the font had been added to the set, and never actually adding to it. Which means every Angular digest cycle was adding the font to the document all over again. Fixed.
</blockquote></details>

Ah, that'll do it! Performance is much better now.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 120 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NsGbleHCu8QNeW17z0v:-NsGbleHCu8QNeW17z0w:bop1s1b) ([raw file](https://github.com/sillsdev/web-xforge/blob/505eac5f27febf549dc23d28046cde66c8f07eec/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L120)):*
> ```XML
>   }
> 
>   getFontFamilyFromProject(projectDoc: SFProjectProfileDoc | undefined): string {
> ```

nit  Move public method above private methods
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 131 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NsGbkGM-DazlnTZU0R5:-NsGbkGM-DazlnTZU0R6:b7d79ca) ([raw file](https://github.com/sillsdev/web-xforge/blob/505eac5f27febf549dc23d28046cde66c8f07eec/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L131)):*
> ```XML
>   }
> 
>   addFontFamilyToDocument(fontFamily: string, fontUrl: string): void {
> ```

nit  Can be private

Nateowami (commented):
> Is it any better if you apply the font to everything in the editor

It was much, much worse.

Sorry; PR response is terse as I'm boarding a flight _right now_.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-:-Nt7Ia3s5jqa1J2emnC2:bwces49)* status: 7 of 9 files reviewed, 5 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 100 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NsGbewYA6uvuTBUbTCq:-Nt7IO08B_OGnSJFFSAX:b6zrrkh) ([raw file](https://github.com/sillsdev/web-xforge/blob/505eac5f27febf549dc23d28046cde66c8f07eec/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L100)):*

nit  Can be private
</blockquote></details>

Good catch.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 120 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NsGbleHCu8QNeW17z0v:-Nt7IQbKB4ihygJcdBAY:b6zrrkh) ([raw file](https://github.com/sillsdev/web-xforge/blob/505eac5f27febf549dc23d28046cde66c8f07eec/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L120)):*

nit  Move public method above private methods
</blockquote></details>

Good catch.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 131 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NsGbkGM-DazlnTZU0R5:-Nt7IRDd3gq9lGNkeuhY:b6zrrkh) ([raw file](https://github.com/sillsdev/web-xforge/blob/505eac5f27febf549dc23d28046cde66c8f07eec/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L131)):*

nit  Can be private
</blockquote></details>

Good catch.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/font.service.spec.ts` line 56 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NsGd7EA9gco8tvwOz4U:-Nt7ITja5QL2F1EBpXv3:b-bxr0t5) ([raw file](https://github.com/sillsdev/web-xforge/blob/505eac5f27febf549dc23d28046cde66c8f07eec/src/SIL.XForge.Scripture/ClientApp/src/font.service.spec.ts#L56)):*

Would a spy have worked easier for this?
</blockquote></details>

I don't think so. And I don't want to actually add fonts to the document, since that potentially creates side-effects for other tests.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/font.service.spec.ts` line  at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NsGctLpAW4WAZtu5PIW:-Nt7ISKDEZFoZ4iKsReG:b-k2rwtz) ([raw file](https://github.com/sillsdev/web-xforge/blob/505eac5f27febf549dc23d28046cde66c8f07eec/src/SIL.XForge.Scripture/ClientApp/src/font.service.spec.ts#L)):*

Just noticed that this test file is in the wrong folder
</blockquote></details>

Good catch. Fixed.

nigel-wells (commented):
All good. What did you want to do about the height regarding this PR? Let it in or find a solution?

> Reviewed 1 of 2 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-:-NtE9obA6OpMXJNB7C6W:b-7f4p5g)* status: 8 of 9 files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 100 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NsGbewYA6uvuTBUbTCq:-NtE9ilxDDfiKk08FbA0:b-madojf) ([raw file](https://github.com/sillsdev/web-xforge/blob/505eac5f27febf549dc23d28046cde66c8f07eec/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L100)):*

Good catch.
</blockquote></details>

Did you want to make this private or leave as public?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/font.service.spec.ts` line 56 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NsGd7EA9gco8tvwOz4U:-NtE9bmDAJVWwfh83Ujb:bpfu2hz) ([raw file](https://github.com/sillsdev/web-xforge/blob/505eac5f27febf549dc23d28046cde66c8f07eec/src/SIL.XForge.Scripture/ClientApp/src/font.service.spec.ts#L56)):*

I don't think so. And I don't want to actually add fonts to the document, since that potentially creates side-effects for other tests.
</blockquote></details>

Good point about side-effects for other tests.

Nateowami (commented):
I've done some work on the text height, and I think it's now much, much, better. There are some situations where I think there's now more of a gap between lines than would be ideal, but I think it's acceptable. 

There's a link in the Jira issues to an article about line height and different fonts. It can probably help us fully solve the problem later.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-:-NtSCZk17buT_DyYr9nT:b-2afx7z)* status: 7 of 9 files reviewed, all discussions resolved (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 100 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-NsGbewYA6uvuTBUbTCq:-NtSMVoI-up0oq42U3BX:b7d979c) ([raw file](https://github.com/sillsdev/web-xforge/blob/505eac5f27febf549dc23d28046cde66c8f07eec/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L100)):*

Did you want to make this private or leave as public?
</blockquote></details>

Public. `getFontFamilyFromProject` only exists as a convenience wrapper around this method, because it makes things simpler in the template.

nigel-wells (approved):
Extra line height does the trick. What did you want to do about feedback from the testers? This seems good apart from their comments.

> Reviewed 1 of 2 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2346#-:-NtYb3Lu60rkY90skrrk:b-71d5cn)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami:
@nigel-wells I think that was referring to the fact that there were some places where the font wasn't applied? I deliberately skipped over some components that were less trivial, in an effort to get this completely more quickly.

nigel-wells:
Correct. Shall we let them know to create cards for them after this is merged in?

Nateowami:
@nigel-wells I've created a bunch of cards in the epic.

nigel-wells:
That's great. Let's update this PR title to match the new [SF-2636](https://jira.sil.org/browse/SF-2636) which I assume has taken its place.

Nateowami:
Done. Thought the branch name will continue to reference SF-531.

nigel-wells:
Thanks. Do the testers know they need to now look at this new card - they might think they are already done from when they were testing SF-531? I think they are done and it just needs to be moved out of testing now and then we can merge.

Nateowami:
@nigel-wells I think they'll know, since a new card has been added to the column. Do you think they finished testing it on the old card though? I didn't really think so.

nigel-wells:
I see they are testing it again :)

PR #2347, opened 2024-02-19
SF-2551 Throttle Syncs to one at a time

pmachapman:
This PR throttles Hangfire jobs to run one at a time, per job type.

These job types are:

 * Syncs
 * Draft Builds
 * Updating of Translation Sources

This was achieved by modifying the Mutex to make the `resource` parameter optional. When `resource` is not configured, the method name is used as the mutex key.

It was determined to allow this approach rather than allow just one job at a time for any job type to stop any potential dependency deadlocks (i.e. when a sync triggers a build, and that build may require a sync, etc). In addition, the three job types are different enough that I do not that that a Resource conflict is likely due to concurrent access, as the shared surface is the MongoDB, the Realtime server, and Paratext Settings.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2347)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/MachineApiService.cs
+++ b/src/SIL.XForge.Scripture/Services/MachineApiService.cs
@@ -894,7 +894,7 @@ public class MachineApiService : IMachineApiService
         if (!string.IsNullOrWhiteSpace(jobId))
         {
             // Run the training after the sync has completed
-            jobId = _backgroundJobClient.ContinueJobWith<IMachineProjectService>(
+            jobId = _backgroundJobClient.ContinueJobWith<MachineProjectService>(
                 jobId,
                 r => r.BuildProjectForBackgroundJobAsync(curUserId, buildConfig, true, CancellationToken.None)
             );
@@ -902,7 +902,7 @@ public class MachineApiService : IMachineApiService
         else
         {
             // No sync required, just run the training
-            jobId = _backgroundJobClient.Enqueue<IMachineProjectService>(
+            jobId = _backgroundJobClient.Enqueue<MachineProjectService>(
                 r => r.BuildProjectForBackgroundJobAsync(curUserId, buildConfig, true, CancellationToken.None)
             );
         }
--- a/src/SIL.XForge.Scripture/Services/MachineProjectService.cs
+++ b/src/SIL.XForge.Scripture/Services/MachineProjectService.cs
@@ -391,6 +391,7 @@ public class MachineProjectService : IMachineProjectService
         return null;
     }
 
+    [Mutex]
     public async Task BuildProjectForBackgroundJobAsync(
         string curUserId,
         BuildConfig buildConfig,
@@ -847,6 +848,7 @@ public class MachineProjectService : IMachineProjectService
         }
     }
 
+    [Mutex]
     public async Task UpdateTranslationSourcesAsync(string curUserId, string sfProjectId)
     {
         // Get the user secret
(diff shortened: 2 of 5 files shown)


josephmyers (approved):
Great, I tested this out with sync's, and it's working as expected. I noticed two usages of Mutex in the migrator code. They're supplying a resource string, which you removed from ParatextSyncRunner, so I'm wondering if we need to remove it from those places, as well. But I don't know the migrator code well

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2347#-:-Nr47ko94l0uxhL6FEw5:b-45aneo)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (commented):
> I'm wondering if we need to remove it from those places, as well.
I tried to build and run the two Migrators that have Mutexes, but they are designed for a much older codebase and reference classes and methods which have been renamed or no longer exist.

I think we can get away with keeping them "as-is" for historical purposes, and could probably even remove them at a later time.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2347#-:-Nr6nKFG-ZEpKTk9_QHB:b953tiq)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

marksvc (commented):
> When `resource` is not configured, the method name is used as the mutex key.

Peter, could you record that in the MutexAttribute class comment? That might help me next time I look at this :)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2347#-:-NrBowT5BclG6BKaRz8S:b-1kxv7d)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 147 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2347#-NrBoGNm8qmm9yqEDmlY:-NrBoGNm8qmm9yqEDmlZ:begcozb) ([raw file](https://github.com/sillsdev/web-xforge/blob/633ba6628abe875aea8e6fb24250b4c0700f76cf/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L147)):*
> ```Smalltalk
>     /// <param name="token">The cancellation token.</param>
>     /// <remarks>
>     /// Do not allow multiple sync jobs to run in parallel on the same project by creating a hangfire mutex on the
> ```

Can you update this to clarify what the mutex is over? Thanks!

pmachapman (commented):
@marksvc Good idea. I have updated the comment. Please let me know if it is clearer.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2347#-:-NrBrcLV4grsLdkfxsWm:b-ns06sw)* status: 3 of 5 files reviewed, all discussions resolved (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 147 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2347#-NrBoGNm8qmm9yqEDmlY:-NrBrZy146BCbuuRZr0E:b-q66hu7) ([raw file](https://github.com/sillsdev/web-xforge/blob/633ba6628abe875aea8e6fb24250b4c0700f76cf/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L147)):*

Can you update this to clarify what the mutex is over? Thanks!
</blockquote></details>

Done. Good spotting - thank you!

marksvc (commented):
That's helpful, thank you.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2347#-:-NrCcvuG31mXGSQClo5T:buyrutp)* status: 3 of 5 files reviewed, all discussions resolved (waiting on @josephmyers)

PR #2354, opened 2024-02-26
Pass client ID to auth0 when changing password

nigel-wells:
This fixes an issue that causes our password reset emails to be sent out from Auth0 under a default "application" rather than our Scripture Forge specific one.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2354)


josephmyers (approved):
I'm actually not able to see the difference this makes. I'm sure this is simply because I'm not familiar enough with Auth0 functionality. The emails were from the same source, before and after, and appear to have the same message body, as well. But I'm giving this a pass, under the assumption that I'm just missing something. Thanks

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2354#-:-NrZ-5TK2aLOnUnFydRi:b-9m0md0)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

nigel-wells (commented):
I have another PR open to change our Auth0 settings which requires accessing application metadata. Without this change the metadata will not be picked up. You'll be able to spot the difference in the footer of the "Change Password" email:


On `master` "Support Center" will not have a valid link. On this branch it will now point to help.scriptureforge.org.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2354#-:-Nrbvr2l7GJeBj-LLTx9:b-382jbo)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

josephmyers:
Wow! In my emails, that footer is collapsed by default into a tiny, flat three-dot menu. I never would've noticed it if you hadn't told me there was a footer.

PR #2358, opened 2024-02-28
SF-2583 Improve Note Speed

pmachapman:
This PR:

 * Adds indices that match the querying patterns of the `note_threads` collection by `sf-project.service.ts`. This will speed up the realtime server's access of notes.
  **Note:** Indexes cannot be created on existing object stores during a version change transaction, so the indexes will only be created locally on log out/log on.
 * Optimizes of `ParatextSyncRunner.GetAndUpdateParatextBooksAndNotes()` to only update notes that are editable. This can speed up the pre-PT sync phase for projects with little or no notes created in SF.
 * Optimizes retrieval of chapter deltas from the USFM on disk to only be loaded once per book (the conversion from USFM to USX to Deltas is expensive). This is a minor performance improvement, but one that results in a reasonable code clean up.
 * Optimizes use of `CommentManager.FindThreads()` in `ParatextService` to query the VerseRefString, rather than the VerseRef object which must be instantiated for every note if used in `CommentManager.FindThreads()`.

(Each of these optimizations have a corresponding commit).

These changes resulted in a modest speed increase for one particularly large project with a note on every verse, but this PR does not provide massive improvements.

**Caveat:** I was unable to find any multiple iteration type bugs that could be optimized. An optimization I tried was to update all of the notes in one go - which turned out to be slower than the current process. Perhaps only substantial optimization could be achieved by creating a new MongoDB structure for notes that perhaps mimics Paratext's structure for notes more closely to enable a much quicker "one pass" note change detection algorithm (the scope of that kind of change is too large for this issue, and requires careful consideration).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2358)

### DIFF
--- a/src/RealtimeServer/common/services/doc-service.ts
+++ b/src/RealtimeServer/common/services/doc-service.ts
@@ -1,4 +1,4 @@
-import { Db } from 'mongodb';
+import { Db, IndexSpecification } from 'mongodb';
 import { ConnectSession } from '../connect-session';
 import { Migration, MigrationConstructor } from '../migration';
 import { ValidationSchema } from '../models/validation-schema';
@@ -54,7 +54,7 @@ export abstract class DocService<T = any> {
   }
 
   abstract get collection(): string;
-  protected abstract get indexPaths(): string[];
+  protected abstract get indexPaths(): (string | IndexSpecification)[];
   validationSchema: ValidationSchema | undefined = undefined;
 
   init(server: RealtimeServer): void {
@@ -78,7 +78,11 @@ export abstract class DocService<T = any> {
   async createIndexes(db: Db): Promise<void> {
     for (const path of this.indexPaths) {
       const collection = db.collection(this.collection);
-      await collection.createIndex({ [path]: 1 });
+      if (typeof path === 'string') {
+        await collection.createIndex({ [path]: 1 });
+      } else {
+        await collection.createIndex(path);
+      }
     }
   }
 
--- a/src/RealtimeServer/scriptureforge/models/note-thread.ts
+++ b/src/RealtimeServer/scriptureforge/models/note-thread.ts
@@ -1,11 +1,26 @@
 import { ProjectData, PROJECT_DATA_INDEX_PATHS } from '../../common/models/project-data';
+import { obj } from '../../common/utils/obj-path';
 import { BiblicalTermNoteHeadingInfo } from './biblical-term-note-heading-info';
 import { Note } from './note';
 import { TextAnchor } from './text-anchor';
 import { VerseRefData } from './verse-ref-data';
 
 export const NOTE_THREAD_COLLECTION = 'note_threads';
-export const NOTE_THREAD_INDEX_PATHS = PROJECT_DATA_INDEX_PATHS;
+export const NOTE_THREAD_INDEX_PATHS = [
+  ...PROJECT_DATA_INDEX_PATHS,
+  // Index for SFProjectService.queryNoteThreads()
+  {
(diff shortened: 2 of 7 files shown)


nigel-wells (commented):
These look like some nice little improvements. Just one query for you. Everything appears to work fine testing this.

> Reviewed 4 of 4 files at r1, 2 of 2 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2358#-:-NtU-0oh18DPpKFowzQq:b-3ji7p5)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2454 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2358#-NtTzYy-3axupJHIxmWr:-NtTzYy-3axupJHIxmWs:bsxsybt) ([raw file](https://github.com/sillsdev/web-xforge/blob/28dd26a338ce9d4a75338d8aae89d6f035310b98/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2454)):*
> ```Smalltalk
>     {
>         // We will query the VerseRefStr value, rather than VerseRef to speed up this query
>         string verseRefBook = bookNum is not null ? Canon.BookNumberToId(bookNum.Value) + " " : string.Empty;
> ```

Should we be concerned about finding and matching threads if this ended up as an empty string?

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2358)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2454 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2358#-NtTzYy-3axupJHIxmWr:-NtU5Smx8RdzK-_txlNx:biqtro2) ([raw file](https://github.com/sillsdev/web-xforge/blob/28dd26a338ce9d4a75338d8aae89d6f035310b98/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2454)):*

Should we be concerned about finding and matching threads if this ended up as an empty string?
</blockquote></details>

No, not that I am aware of - it is mostly there so I can make sure the variable is not null. Notice that the lambda code in manager.FindThreads() only uses `verseRefBook` if `bookNum` is not null. If `bookNum` is not null, `verseRefBook` will have a valid value.

nigel-wells (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2358#-:-NtYYgc5A43ghJ5wLBVJ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2454 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2358#-NtTzYy-3axupJHIxmWr:-NtYYe9sC40j1XoZbVZ9:b-c3tlw3) ([raw file](https://github.com/sillsdev/web-xforge/blob/28dd26a338ce9d4a75338d8aae89d6f035310b98/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2454)):*

No, not that I am aware of - it is mostly there so I can make sure the variable is not null. Notice that the lambda code in manager.FindThreads() only uses `verseRefBook` if `bookNum` is not null. If `bookNum` is not null, `verseRefBook` will have a valid value.
</blockquote></details>

Good points.

PR #2359, opened 2024-02-28
SF-2518 Fix sync failure when multiple threads have the same Id

RaymondLuong3:
In rare cases, multiple comment threads will be listed that are actually a single thread. The most likely cause is that the thread loading algorithm in Paratext.Data has a problem that only occurs in rare situations. When it happens, Paratext shows two icons for the same thread, but this prevents SF from synchronizing with PT.

This change uses `CommentManager.FindThread()` to extract a single thread from all comments with the same thread Id where as previously an error would occur when querying from a list. Unfortunately, I could not get a test to successfully demonstrate the problem state.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2359)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -1250,7 +1250,7 @@ public class ParatextService : DisposableBase, IParatextService
         Dictionary<string, ParatextUserProfile> ptProjectUsers
     )
     {
-        IEnumerable<CommentThread>? commentThreads = GetCommentThreads(userSecret, paratextId, bookNum);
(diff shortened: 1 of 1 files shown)


pmachapman (approved):
:lgtm: I also attempted to cause the problem state to occur, but could not. Based on my inspection of the code in ParatextData, it looks like your fix will improve the code for cases that work in PT but not SF.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2359#-:-Ns553Da96l0PgxgFcXU:bgiov8d)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2363, opened 2024-03-04
SF-2528 Editor tabs drag/drop reorder

siltomato:
Add drag/drop reordering and moving of tabs between source/target tab groups.  Main source and target project tabs should remain fixed in place as first item in their respective groups.

This PR is currently based off `feature/sf-2439-editor-tabs`, which is not yet merged into master.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2363)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/base-services/tab-menu.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/base-services/tab-menu.service.ts
@@ -1,12 +1,12 @@
 import { Observable } from 'rxjs';
 
-export interface NewTabMenuItem {
+export interface TabMenuItem {
   type: string;
   text: string;
   icon?: string;
   disabled?: boolean;
 }
 
-export abstract class TabMenuService {
-  abstract getMenuItems(tabGroup: string): Observable<NewTabMenuItem[]>;
+export abstract class TabMenuService<TGroupId extends string> {
+  abstract getMenuItems(groupId?: TGroupId): Observable<TabMenuItem[]>;
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/sf-tabs.module.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/sf-tabs.module.ts
@@ -1,3 +1,4 @@
+import { CdkDrag, CdkDropList, CdkDropListGroup } from '@angular/cdk/drag-drop';
 import { CommonModule } from '@angular/common';
 import { NgModule } from '@angular/core';
 import { MatButtonModule } from '@angular/material/button';
@@ -23,7 +24,17 @@ import { TabComponent } from './tab/tab.component';
     TabScrollButtonComponent,
     TabBodyComponent
   ],
-  imports: [CommonModule, MatTabsModule, MatButtonModule, MatIconModule, MatMenuModule, TranslocoModule],
+  imports: [
+    CommonModule,
+    MatTabsModule,
+    MatButtonModule,
+    MatIconModule,
+    MatMenuModule,
+    CdkDrag,
+    CdkDropList,
+    CdkDropListGroup,
+    TranslocoModule
+  ],
   exports: [TabGroupComponent, TabComponent, TabHeaderDirective]
 })
 export class SFTabsModule {}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/sf-tabs.module.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/sf-tabs.module.ts
@@ -1,4 +1,4 @@
-import { CdkDrag, CdkDropList, CdkDropListGroup } from '@angular/cdk/drag-drop';
+import { DragDropModule } from '@angular/cdk/drag-drop';
 import { CommonModule } from '@angular/common';
 import { NgModule } from '@angular/core';
 import { MatButtonModule } from '@angular/material/button';
@@ -30,9 +30,7 @@ import { TabComponent } from './tab/tab.component';
     MatButtonModule,
     MatIconModule,
     MatMenuModule,
-    CdkDrag,
-    CdkDropList,
-    CdkDropListGroup,
+    DragDropModule,
     TranslocoModule
   ],
   exports: [TabGroupComponent, TabComponent, TabHeaderDirective]
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/sf-tabs.types.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/sf-tabs.types.ts
@@ -5,3 +5,13 @@ export interface TabEvent {
 export interface TabHeaderMouseEvent extends TabEvent {
   mouseEvent: MouseEvent;
 }
+
+export interface TabLocation<TGroupId> {
+  groupId: TGroupId;
+  index: number;
+}
+
+export interface TabMoveEvent<TGroupId> {
+  from: TabLocation<TGroupId>;
+  to: TabLocation<TGroupId>;
+}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/sf-tabs.types.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/sf-tabs.types.ts
@@ -2,8 +2,8 @@ export interface TabEvent {
   index: number;
 }
 
-export interface TabHeaderMouseEvent extends TabEvent {
-  mouseEvent: MouseEvent;
+export interface TabHeaderPointerEvent extends TabEvent {
+  pointerEvent: MouseEvent | TouchEvent;
 }
 
 export interface TabLocation<TGroupId> {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.html
@@ -21,7 +21,8 @@
     [closeable]="tab.closeable"
     [movable]="tab.movable"
     [active]="i === selectedIndex"
-    (tabClick)="tabClick.next({ index: i, mouseEvent: $event })"
+    (tabPress)="tabPress.next({ index: i, pointerEvent: $event })"
+    (tabClick)="tabClick.next({ index: i, pointerEvent: $event })"
     (closeClick)="closeClick.next(i)"
     cdkDrag
     [cdkDragDisabled]="!tab.movable"
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.html
@@ -5,13 +5,27 @@
   (scrollStop)="stopButtonScroll()"
 ></app-tab-scroll-button>
 
-<div class="tabs">
+<div
+  [id]="groupId!"
+  class="tabs"
+  cdkDropList
+  cdkDropListLockAxis="x"
+  cdkDropListOrientation="horizontal"
+  [cdkDropListDisabled]="!allowDragDrop"
+  [cdkDropListConnectedTo]="connectedTo"
+  [cdkDropListSortPredicate]="movablePredicate"
+  (cdkDropListDropped)="onTabDrop($event)"
+>
   <app-tab-header
     *ngFor="let tab of tabs; let i = index"
     [closeable]="tab.closeable"
+    [movable]="tab.movable"
     [active]="i === selectedIndex"
     (tabClick)="tabClick.next({ index: i, mouseEvent: $event })"
     (closeClick)="closeClick.next(i)"
+    cdkDrag
+    [cdkDragDisabled]="!tab.movable"
+    [cdkDragData]="tab"
   >
     <ng-container *ngTemplateOutlet="tab.tabHeaderTemplate"></ng-container>
   </app-tab-header>
@@ -22,6 +36,9 @@
     [closeable]="false"
     class="add-tab"
     (click)="onAddTabClicked()"
+    cdkDrag
+    cdkDragDisabled
+    [cdkDragData]="{ isAddTab: true }"
   >
     <mat-icon>add</mat-icon>
   </app-tab-header>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.html
@@ -13,7 +13,7 @@
   cdkDropListOrientation="horizontal"
   [cdkDropListDisabled]="!allowDragDrop"
   [cdkDropListConnectedTo]="connectedTo"
-  [cdkDropListSortPredicate]="movablePredicate"
+  [cdkDropListSortPredicate]="movablePredicate.bind(this)"
   (cdkDropListDropped)="onTabDrop($event)"
 >
   <app-tab-header
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.scss
@@ -52,6 +52,7 @@ app-tab-scroll-button {
   align-items: flex-end;
   gap: 0.1em;
   border-radius: var(--sf-tab-header-border-radius) var(--sf-tab-header-border-radius) 0 0;
+  width: 100%;
 
   // Hide scroll bar
   scrollbar-width: none;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.scss
@@ -59,6 +59,11 @@ app-tab-scroll-button {
     width: 0;
     height: 0;
   }
+
+  // Transition the slide of the tab headers when making place for drop
+  &.cdk-drop-list-dragging :not(app-tab-header.cdk-drag-placeholder) {
+    transition: transform 250ms cubic-bezier(0, 0, 0.2, 1);
+  }
 }
 
 .add-tab {
@@ -68,6 +73,23 @@ app-tab-scroll-button {
   &:hover {
     color: inherit !important;
     background-color: var(--sf-tab-header-button-hover-background-color);
+  }
+}
+
+app-tab-header {
+  // Shadow under the dragged tab
+  &.cdk-drag-preview {
+    box-shadow: 0 4px 4px -3px rgba(0, 0, 0, 0.2);
+  }
+
+  // Empty spot for drop target
+  &.cdk-drag-placeholder {
+    opacity: 0;
+  }
+
+  // Transition the drop of the dragged tab into its place
+  &.cdk-drag-animating {
+    transition: transform 250ms cubic-bezier(0, 0, 0.2, 1);
   }
 }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.spec.ts
@@ -1,3 +1,4 @@
+import { CdkDrag, CdkDropList } from '@angular/cdk/drag-drop';
 import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { of } from 'rxjs';
 import { TestTranslocoModule } from 'xforge-common/test-utils';
@@ -250,5 +251,34 @@ describe('TabGroupHeaderComponent', () => {
       expect(scrollIntoViewSpy).toHaveBeenCalled();
       expect(scrollToEndSpy).not.toHaveBeenCalled();
     }));
+  });
+
+  describe('movablePredicate', () => {
+    let tab: any;
+    let cdkDrag = {} as CdkDrag;
+    let cdkDropList: CdkDropList;
+
+    beforeEach(() => {
+      tab = {};
+      cdkDropList = { getSortedItems: () => [{ data: tab } as CdkDrag] } as CdkDropList;
+    });
+
+    it('should return true if the tab at index is an "add" tab', () => {
+      const index = 0;
+      tab.isAddTab = true;
+      expect(component.movablePredicate(index, cdkDrag, cdkDropList)).toBe(true);
+    });
+
+    it('should return true if the tab at index is movable', () => {
+      const index = 0;
+      tab.movable = true;
+      expect(component.movablePredicate(index, cdkDrag, cdkDropList)).toBe(true);
+    });
+
+    it('should return false if the tab at index is not an "add" and not movable', () => {
+      const index = 0;
+      tab.movable = false;
+      expect(component.movablePredicate(index, cdkDrag, cdkDropList)).toBe(false);
+    });
   });
 });
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.spec.ts
@@ -254,31 +254,112 @@ describe('TabGroupHeaderComponent', () => {
   });
 
   describe('movablePredicate', () => {
-    let tab: any;
-    let cdkDrag = {} as CdkDrag;
+    let tabToMove: any;
+    let cdkDraggingTab = {} as CdkDrag;
     let cdkDropList: CdkDropList;
 
-    beforeEach(() => {
-      tab = {};
-      cdkDropList = { getSortedItems: () => [{ data: tab } as CdkDrag] } as CdkDropList;
+    describe('move within same tab group', () => {
+      beforeEach(() => {
+        // Dragging tab is in the drop list (not group transfer)
+        cdkDropList = { getSortedItems: () => [{ data: tabToMove } as CdkDrag, cdkDraggingTab] } as CdkDropList;
+        tabToMove = {};
+      });
+
+      it('should return true when direction is ltr and the tab at index is an "add" tab', () => {
+        const index = 0;
+        tabToMove.isAddTab = true;
+        component.direction = 'ltr';
+        expect(component.movablePredicate(index, cdkDraggingTab, cdkDropList)).toBe(true);
+      });
+
+      it('should return true when direction is rtl and the tab at index is an "add" tab', () => {
+        const index = 1;
+        tabToMove.isAddTab = true;
+        component.direction = 'rtl';
+        expect(component.movablePredicate(index, cdkDraggingTab, cdkDropList)).toBe(true);
+      });
+
+      it('should return true when direction is ltr and the tab at index is movable', () => {
+        const index = 0;
+        tabToMove.movable = true;
+        component.direction = 'ltr';
+        expect(component.movablePredicate(index, cdkDraggingTab, cdkDropList)).toBe(true);
+      });
+
+      it('should return true when direction is rtl and the tab at index is movable', () => {
+        const index = 1;
+        tabToMove.movable = true;
+        component.direction = 'rtl';
+        expect(component.movablePredicate(index, cdkDraggingTab, cdkDropList)).toBe(true);
+      });
+
+      it('should return false when direction is ltr and the tab at index is not an "add" and not movable', () => {
+        const index = 0;
+        tabToMove.movable = false;
+        component.direction = 'ltr';
+        expect(component.movablePredicate(index, cdkDraggingTab, cdkDropList)).toBe(false);
+      });
+
+      it('should return false when direction is rtl and the tab at index is not an "add" and not movable', () => {
+        const index = 1;
+        tabToMove.movable = false;
+        component.direction = 'rtl';
+        expect(component.movablePredicate(index, cdkDraggingTab, cdkDropList)).toBe(false);
+      });
     });
 
-    it('should return true if the tab at index is an "add" tab', () => {
-      const index = 0;
-      tab.isAddTab = true;
-      expect(component.movablePredicate(index, cdkDrag, cdkDropList)).toBe(true);
-    });
+    describe('transfer to another tab group', () => {
+      beforeEach(() => {
+        // Dragging tab is not in the drop list (meaning group transfer)
+        cdkDropList = {
+          getSortedItems: () => [{ data: tabToMove } as CdkDrag, { data: {} } as CdkDrag]
+        } as CdkDropList;
+        tabToMove = {};
+      });
 
-    it('should return true if the tab at index is movable', () => {
-      const index = 0;
-      tab.movable = true;
-      expect(component.movablePredicate(index, cdkDrag, cdkDropList)).toBe(true);
-    });
+      it('should return true when direction is ltr and the tab at index is an "add" tab', () => {
+        const index = 0;
+        tabToMove.isAddTab = true;
+        component.direction = 'ltr';
+        expect(component.movablePredicate(index, cdkDraggingTab, cdkDropList)).toBe(true);
+      });
 
-    it('should return false if the tab at index is not an "add" and not movable', () => {
-      const index = 0;
-      tab.movable = false;
-      expect(component.movablePredicate(index, cdkDrag, cdkDropList)).toBe(false);
+      it('should return true when direction is rtl and the tab at index is an "add" tab', () => {
+        // As of (v16), CDK drag and drop seems to have some issues transferring horizontally-oriented items in RTL.
+        // The library passes an index that references the wrong item when transferring groups in RTL.
+        const index = 2;
+        tabToMove.isAddTab = true;
+        component.direction = 'rtl';
+        expect(component.movablePredicate(index, cdkDraggingTab, cdkDropList)).toBe(true);
+      });
+
+      it('should return true when direction is ltr and the tab at index is movable', () => {
+        const index = 0;
+        tabToMove.movable = true;
+        component.direction = 'ltr';
+        expect(component.movablePredicate(index, cdkDraggingTab, cdkDropList)).toBe(true);
+      });
+
+      it('should return true when direction is rtl and the tab at index is movable', () => {
+        const index = 2;
+        tabToMove.movable = true;
+        component.direction = 'rtl';
+        expect(component.movablePredicate(index, cdkDraggingTab, cdkDropList)).toBe(true);
+      });
+
+      it('should return false when direction is ltr and the tab at index is not an "add" and not movable', () => {
+        const index = 0;
+        tabToMove.movable = false;
+        component.direction = 'ltr';
+        expect(component.movablePredicate(index, cdkDraggingTab, cdkDropList)).toBe(false);
+      });
+
+      it('should return false when direction is rtl and the tab at index is not an "add" and not movable', () => {
+        const index = 2;
+        tabToMove.movable = false;
+        component.direction = 'rtl';
+        expect(component.movablePredicate(index, cdkDraggingTab, cdkDropList)).toBe(false);
+      });
     });
   });
 });
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.ts
@@ -26,7 +26,7 @@ import {
   Subscription
 } from 'rxjs';
 import { TabMenuItem, TabMenuService } from 'src/app/shared/sf-tab-group';
-import { TabHeaderMouseEvent, TabMoveEvent } from '../sf-tabs.types';
+import { TabHeaderPointerEvent, TabLocation, TabMoveEvent } from '../sf-tabs.types';
 import { TabHeaderComponent } from '../tab-header/tab-header.component';
 import { TabComponent } from '../tab/tab.component';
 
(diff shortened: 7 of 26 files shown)


RaymondLuong3 (changes requested):
This is so slick! When I drag a tab from one group to another, I expect that the tab I dragged would become the active one in the tab group. This is not the case currently. Do you think that is worth changing? In VSCode, that is the behaviour I see.

> Reviewed 23 of 23 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-:-NsKjYY10oar2Jl-F6eb:b-9dbrli)* status: all files reviewed, 11 unresolved discussions (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.stories.ts` line 12 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKmHwj9wI-ZJvEn6Nf:-NsKmHwj9wI-ZJvEn6Ng:b-onz1hw) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.stories.ts#L12)):*
> ```XML
>   TabMenuService,
>   TabStateService
> } from 'src/app/shared/sf-tab-group';
> ```

Nit: This should be a relative path.



_Code quote:_
```XML
} from 'src/app/shared/sf-tab-group';
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.stories.ts` line 37 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKm7m6ArwhRLkYh6OH:-NsKm7m6ArwhRLkYh6OI:b-xt2n7j) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.stories.ts#L37)):*
> ```XML
>     >
>       <app-tab
>         *ngFor="let tab of tabGroup.value.tabs; let i = index"
> ```

I don't see I being used anywhere.



_Code quote:_
```XML
let i = index
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.scss` line 80 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKqmEa3CQc6614FlYO:-NsKqmEa3CQc6614FlYP:brene6i) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.scss#L80)):*
> ```SCSS
> 
> app-tab-header {
>   // Shadow under the dragged tab
> ```

Thanks for the comments here.



_Code quote:_
```SCSS
  // Shadow under the dragged tab
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.ts` line 130 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKrdwgES-31tsuvsdi:-NsKrdwgES-31tsuvsdj:bzavz5u) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.ts#L130)):*
> ```XML
>     // Convert CdkDragDrop event to TabMoveEvent
>     const from = { groupId: event.previousContainer.id, index: event.previousIndex };
>     const to = { groupId: event.container.id, index: event.currentIndex };
> ```

Can you add types here.



_Code quote:_
```XML
    const from = { groupId: event.previousContainer.id, index: event.previousIndex };
    const to = { groupId: event.container.id, index: event.currentIndex };
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts` line 107 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKuDHm71L56LO7Ud_Z:-NsKuDHm71L56LO7Ud__:bjob9hc) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts#L107)):*
> ```XML
> 
>   moveTab(from: TabLocation<TGroupId>, to: TabLocation<TGroupId>): void {
>     const fromGroup = this.groups.get(from.groupId);
> ```

Nit: add the type here.



_Code quote:_
```XML
    const fromGroup = this.groups.get(from.groupId);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts` line 118 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKvY9S2nPQGG1q11Gk:-NsKvY9S2nPQGG1q11Gl:b5tlw0h) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts#L118)):*
> ```XML
> 
>         // Update selected tab index if necessary
>         switch (true) {
> ```

Why did you use a switch here? Maybe this would be clearer as a separate function call `updateSelectedTab`



_Code quote:_
```XML
        switch (true) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts` line 134 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKwCAi-qKnphFA9EvM:-NsKwCAi-qKnphFA9EvN:b-kwe3vk) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts#L134)):*
> ```XML
>       } else {
>         // Tab move across groups
>         const toGroup = this.groups.get(to.groupId);
> ```

Nit: Add the type here.



_Code quote:_
```XML
  const toGroup = this.groups.get(to.groupId);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts` line 138 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKwUnyEV3Q95dIM7ZL:-NsKwUnyEV3Q95dIM7ZM:b-5j7b85) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts#L138)):*
> ```XML
>         if (toGroup) {
>           // Add bounds in case tab is dropped after 'add tab'
>           to.index = Math.min(toGroup.tabs.length, to.index);
> ```

If 'add tab' is at index 1, and we drag a tab past index 1, then the `to.index` would be 2, when it should be 1? Maybe I'm not understanding correctly.



_Code quote:_
```XML
          to.index = Math.min(toGroup.tabs.length, to.index);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts` line 143 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKxDEq4XO6PNXhH1UX:-NsKxDEq4XO6PNXhH1UY:ba3t1is) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts#L143)):*
> ```XML
> 
>           // Update 'from group' selected tab index if necessary
>           switch (true) {
> ```

This should just be an if, else if statement rather than a switch.



_Code quote:_
```XML
          // Update 'from group' selected tab index if necessary
          switch (true) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.spec.ts` line 125 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsL-1xE8LLZ-_a7ok6Y:-NsL-1xE8LLZ-_a7ok6Z:b-m4tf01) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.spec.ts#L125)):*
> ```XML
>     });
> 
>     describe('moveTab', () => {
> ```

Could you add a test where the selected tab is the one that is moved to a new group, and when that happens, a new tab in the old group becomes the selected tab.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.spec.ts` line 189 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKzghM2mRpSyh_Psgh:-NsKzghM2mRpSyh_Psgi:b-kpnb7c) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.spec.ts#L189)):*
> ```XML
>         service.moveTab({ groupId: fromGroupId, index: 0 }, { groupId: toGroupId, index: 1 });
>         expect(service['groups'].get(fromGroupId)!.selectedIndex).toBe(0);
>         expect(service['groups'].get(toGroupId)!.selectedIndex).toBe(2);
> ```

If we did choose to have the tab moved be the newest selected tab, this would need to be updated.



_Code quote:_
```XML
        expect(service['groups'].get(fromGroupId)!.selectedIndex).toBe(0);
        expect(service['groups'].get(toGroupId)!.selectedIndex).toBe(2);
```

siltomato (commented):
Good idea.  I was hesitant to select the tab on `mousedown` as opposed to click, but when I see that that is how VS Code, Chrome, and Firefox do it, it makes sense!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-:-NtSxFWm1Pb97BYgscxT:b-i27in5)* status: 12 of 24 files reviewed, 7 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.stories.ts` line 12 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKmHwj9wI-ZJvEn6Nf:-NtRrVZlEl59bdUv5f4D:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.stories.ts#L12)):*

Nit: This should be a relative path.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.stories.ts` line 37 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKm7m6ArwhRLkYh6OH:-NtRrQhcA0yDMsqXlOg8:bss4a2b) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.stories.ts#L37)):*

I don't see I being used anywhere.
</blockquote></details>

Removed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.scss` line 80 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKqmEa3CQc6614FlYO:-NtRrgu69PpiluXAp1_Y:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.scss#L80)):*

Thanks for the comments here.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.ts` line 130 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKrdwgES-31tsuvsdi:-NtRsYN75uZy9Y6jZl_f:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.ts#L130)):*

Can you add types here.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts` line 107 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKuDHm71L56LO7Ud_Z:-NtRszVD1jUnQjXc_qKK:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts#L107)):*

Nit: add the type here.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts` line 118 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKvY9S2nPQGG1q11Gk:-NtSvA9FEWyXsAlGaxY0:b-lzgeq8) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts#L118)):*

Why did you use a switch here? Maybe this would be clearer as a separate function call `updateSelectedTab`
</blockquote></details>

Sometimes I prefer the switch structure instead of several 'else if' statements, but no problem to change it.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts` line 134 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKwCAi-qKnphFA9EvM:-NtRt2hTEF1ZE4jb_bFq:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts#L134)):*

Nit: Add the type here.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts` line 138 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKwUnyEV3Q95dIM7ZL:-NtSna4V81M-Ypz51Y7e:bhlxte8) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts#L138)):*

If 'add tab' is at index 1, and we drag a tab past index 1, then the `to.index` would be 2, when it should be 1? Maybe I'm not understanding correctly.
</blockquote></details>

This check is necessary for tab movement within the same group, as dropping past the 'add tab' could happen, but it doesn't look like this can happen when moving tabs across groups, so I'll remove this check here.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts` line 143 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKxDEq4XO6PNXhH1UX:-NtSska5BIaFw3Ynq8Lk:blyphgp) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts#L143)):*

This should just be an if, else if statement rather than a switch.
</blockquote></details>

No problem.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.spec.ts` line 125 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsL-1xE8LLZ-_a7ok6Y:-NtSx88_4n6J1INvKsj5:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.spec.ts#L125)):*

Could you add a test where the selected tab is the one that is moved to a new group, and when that happens, a new tab in the old group becomes the selected tab.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.spec.ts` line 189 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKzghM2mRpSyh_Psgh:-NtSxALjEl_gBK7TU4_b:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.spec.ts#L189)):*

If we did choose to have the tab moved be the newest selected tab, this would need to be updated.
</blockquote></details>

Done.

RaymondLuong3 (changes requested):
Nice! I few more minor comments! I won't get a chance to re-review this until monday, so I am happy to let someone else complete the review on this

> Reviewed 12 of 12 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-:-NtX8T1W3idzu4BISPSK:b48v43u)* status: all files reviewed, 4 unresolved discussions (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts` line 47 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NtX2ATa5H72Ljb4xSUl:-NtX2ATa5H72Ljb4xSUm:b-7ghqa7) ([raw file](https://github.com/sillsdev/web-xforge/blob/d1915f1478bf5f94958e06fd3ad46d19d773b8bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts#L47)):*
> ```XML
>     // Select tab on left mouse button press or mobile touch
>     if (e.pointerEvent instanceof TouchEvent || e.pointerEvent.button === 0) {
>       if (e.index < this.tabs.length) {
> ```

does the "add tab" tab count toward the length? If it does, I assume this should be `e.index < this.tabs.length - 1` if it is a 0 based index.



_Code quote:_
```XML
      if (e.index < this.tabs.length) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.ts` line 121 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NtX40n-Cg-XmdtZZZPd:-NtX40n-Cg-XmdtZZZPe:b-lxxpxd) ([raw file](https://github.com/sillsdev/web-xforge/blob/d1915f1478bf5f94958e06fd3ad46d19d773b8bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.ts#L121)):*
> ```XML
>   movablePredicate(index: number, _draggingTab: CdkDrag<TabComponent>, dropList: CdkDropList): boolean {
>     const tabToMove = dropList.getSortedItems()[index];
>     return tabToMove && (tabToMove.data.isAddTab || (tabToMove.data as TabComponent).movable);
> ```

Nit: I would be explicit and write `tabToMove != null` Can you add a type for `tabToMove` above.



_Code quote:_
```XML
tabToMove 
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-header/tab-header.component.html` line 11 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NtX4dJ7AO2XIYZBffEf:-NtX4dJ7AO2XIYZBffEg:b5k879q) ([raw file](https://github.com/sillsdev/web-xforge/blob/d1915f1478bf5f94958e06fd3ad46d19d773b8bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-header/tab-header.component.html#L11)):*
> ```HTML
>   (click)="onCloseClick($event)"
>   (mousedown)="onClosePress($event)"
>   (touchstart)="onClosePress($event)"
> ```

Curious why `(click)` isn't sufficient to handle the closing event. 



_Code quote:_
```HTML
  (mousedown)="onClosePress($event)"
  (touchstart)="onClosePress($event)"
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts` line 118 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NsKvY9S2nPQGG1q11Gk:-NtX7Au-8qmkUwnS5Thd:btd8men) ([raw file](https://github.com/sillsdev/web-xforge/blob/103040973819eddf02a4fac5c10161f01f5d45bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.ts#L118)):*

Sometimes I prefer the switch structure instead of several 'else if' statements, but no problem to change it.
</blockquote></details>

Ah, yes. I can understand that. This is different than how I am used to seeing switch statements, but I see the advantage you mentioned. Thank for the update.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.spec.ts` line 162 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NtX7yKS43xSlqUKxXGk:-NtX7yKTDPZ6DzGBKJqH:b-7o32qw) ([raw file](https://github.com/sillsdev/web-xforge/blob/d1915f1478bf5f94958e06fd3ad46d19d773b8bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.spec.ts#L162)):*
> ```XML
>         group.selectedIndex = 1;
>         service.moveTab({ groupId, index: 2 }, { groupId, index: 1 });
>         expect(service['groups'].get(groupId)!.selectedIndex).toBe(2);
> ```

If we moved this tab to index 1, I thought it would be the tab moved to index 1 that would be selected.



_Code quote:_
```XML
       expect(service['groups'].get(groupId)!.selectedIndex).toBe(2);
```

siltomato (commented):
- added explicit type and null check
- removed unused component ViewChild field

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-:-NtXoJsh8Ljwas5_6rza:b-flgj3d)* status: 23 of 24 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts` line 47 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NtX2ATa5H72Ljb4xSUl:-NtXixHl7MYLd53FSwlR:b-3y6lri) ([raw file](https://github.com/sillsdev/web-xforge/blob/d1915f1478bf5f94958e06fd3ad46d19d773b8bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts#L47)):*

does the "add tab" tab count toward the length? If it does, I assume this should be `e.index < this.tabs.length - 1` if it is a 0 based index.
</blockquote></details>

Good question.  The 'add tab' is header-only and doesn't  count toward the length of `this.tabs`, which is the list of `<app-tab>` content children that would be nested inside an `<app-tab-group>`  So the check ensures that the index is within this range.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.ts` line 121 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NtX40n-Cg-XmdtZZZPd:-NtXgOoA5tSpeDZGX9zD:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/d1915f1478bf5f94958e06fd3ad46d19d773b8bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group-header/tab-group-header.component.ts#L121)):*

Nit: I would be explicit and write `tabToMove != null` Can you add a type for `tabToMove` above.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-header/tab-header.component.html` line 11 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NtX4dJ7AO2XIYZBffEf:-NtXeugQ9p60KaoD6QdH:b-ljkg1q) ([raw file](https://github.com/sillsdev/web-xforge/blob/d1915f1478bf5f94958e06fd3ad46d19d773b8bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-header/tab-header.component.html#L11)):*

Curious why `(click)` isn't sufficient to handle the closing event.
</blockquote></details>

Now that the tab header is listening for `mousedown` and `touchstart` to select the tab, I need to listen for those events on the close button to stop those events from bubbling up to the containing tab header.

In `onClosePress(e)`, I call `e.stopPropagation()`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.spec.ts` line 162 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NtX7yKS43xSlqUKxXGk:-NtXbSXS2dUIDpzup3SG:b-oen9p) ([raw file](https://github.com/sillsdev/web-xforge/blob/d1915f1478bf5f94958e06fd3ad46d19d773b8bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.spec.ts#L162)):*

If we moved this tab to index 1, I thought it would be the tab moved to index 1 that would be selected.
</blockquote></details>

What you describe is correct in practice because `mousedown` to start the reorder now selects the dragged tab, but the tab state service doesn't assume this.  The test indicates that the selected item is at index 1, but the dragged tab (which isn't selected) moves from index 2 to index 1, which pushes the selected tab to index 2.

In practice now, beginning to drag a tab will select it, so the selected index will always follow the index of the dragged tab (this would be simulated in the test by setting `group.selectedIndex` to whatever the 'from' tab index is), but that behavior is not the responsibility of the tab state service, so I set up the tests to handle either scenario.

RaymondLuong3 (approved):
Awesome. My comments seem to be addressed and we can move this to testing. I mention it here since it is related. There is no indication to the user why the source and target text tabs cannot be dragged. We can improve that in the future, but I imagine some users might be tripped up about that.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-:-Ntsjf0n82hnA5c4pGnk:b-xy2ixa)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-header/tab-header.component.html` line 11 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NtX4dJ7AO2XIYZBffEf:-NtsfQjE-sGMbTvrtHnI:b-d3vm0y) ([raw file](https://github.com/sillsdev/web-xforge/blob/d1915f1478bf5f94958e06fd3ad46d19d773b8bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-header/tab-header.component.html#L11)):*

Now that the tab header is listening for `mousedown` and `touchstart` to select the tab, I need to listen for those events on the close button to stop those events from bubbling up to the containing tab header.

In `onClosePress(e)`, I call `e.stopPropagation()`
</blockquote></details>

Aha, I have run into that scenario before. Maybe it would have been better to create a method called stopPropagation() for the mousedown and touchstart events.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.spec.ts` line 162 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NtX7yKS43xSlqUKxXGk:-NtsbU6PBfBzwn9U0BxE:b-29utvj) ([raw file](https://github.com/sillsdev/web-xforge/blob/d1915f1478bf5f94958e06fd3ad46d19d773b8bf/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-state/tab-state.service.spec.ts#L162)):*

What you describe is correct in practice because `mousedown` to start the reorder now selects the dragged tab, but the tab state service doesn't assume this.  The test indicates that the selected item is at index 1, but the dragged tab (which isn't selected) moves from index 2 to index 1, which pushes the selected tab to index 2.

In practice now, beginning to drag a tab will select it, so the selected index will always follow the index of the dragged tab (this would be simulated in the test by setting `group.selectedIndex` to whatever the 'from' tab index is), but that behavior is not the responsibility of the tab state service, so I set up the tests to handle either scenario.
</blockquote></details>

Ah yes. That makes sense. Thanks for the explanation.

Nateowami:
I haven't reviewed any of the code, but I tested this out locally, and I just want to say it looks great! The dragging is a lot easier than I initially imagined, since I envisioned it being more like VS Code and Paratext, where things are snapping into place as the user drags them.

siltomato (commented):
Good to hear!  Curious if you have any ideas regarding adding a more intuitive visual indicator for tabs that are not movable?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-:-Ntwze9AAn31H4JmaTfo:by55gqa)* status: 23 of 24 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (approved):
I am satisfied with the functionality and the code. If testers confirm it is good, let's merge it in.

> Reviewed 1 of 1 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-:-NuY8x4J0Tr6_vG8rroN:bz5afev)* status: all files reviewed, 1 unresolved discussion (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts` line 85 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NuY8GEv5cHbDZlXGiCz:-NuY8GEwBrDS5ki12PTh:b1d58e4) ([raw file](https://github.com/sillsdev/web-xforge/blob/dbe5e644283a0d1c0a38ac2f0eb04d89c846526b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts#L85)):*
> ```XML
>   }
> 
>   private isTouchEvent(event: MouseEvent | TouchEvent): event is TouchEvent {
> ```

I am not familiar with this kind of syntax for a return type. It just sounds like a boolean. What is the advantage of this kind of syntax. The method name already makes it clear that we are expecting to learn if this "event" is a touch event.



_Code quote:_
```XML
  private isTouchEvent(event: MouseEvent | TouchEvent): event is TouchEvent {
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2363)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts` line 85 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NuY8GEv5cHbDZlXGiCz:-NuZxCuo4terc5I51eec:bqpsihy) ([raw file](https://github.com/sillsdev/web-xforge/blob/dbe5e644283a0d1c0a38ac2f0eb04d89c846526b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts#L85)):*

I am not familiar with this kind of syntax for a return type. It just sounds like a boolean. What is the advantage of this kind of syntax. The method name already makes it clear that we are expecting to learn if this "event" is a touch event.
</blockquote></details>

@RaymondLuong3 This is a TypeScirpt type guard. It is telling the TypeScript compiler that this function checks whether the value is a TouchEvent.

See https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

That means later you can do:
``` ts
if (isTouchEvent(value)) {
  // we can now treat value as a TouchEvent
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts` line 86 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NuZy-Ah0tBj7_DeyB5g:-NuZy-Ah0tBj7_DeyB5h:bl6vzxv) ([raw file](https://github.com/sillsdev/web-xforge/blob/dbe5e644283a0d1c0a38ac2f0eb04d89c846526b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts#L86)):*
> ```XML
> 
>   private isTouchEvent(event: MouseEvent | TouchEvent): event is TouchEvent {
>     return typeof TouchEvent !== 'undefined' && event instanceof TouchEvent;
> ```

@siltomato Usually we do null checks with `event != null`, which guards against both `undefined` and `null`. 

You also can't misspell `null`, since it's a keyword (at least not without it being kind of obvious), but you can misspell a string and accidentally have an incorrect check (though the compiler actually does catch it if you make this mistake, so maybe this doesn't make much of a difference. But whenever I see undefined in a string I always feel compelled to check it super closely to make sure it was spelled correctly).

Nateowami:
@siltomato One of the things I find a little odd is that if I drag a tab from one side to the other and let go, it goes back to where it came from, unless I drag it past the <kbd>+</kbd> tab. In this screenshot, I've dragged the history from the right side to the left, but when I let go, it snaps back to the right side, instead of finding its place on the left side.

siltomato (commented):
Yes, that is a good point about the drop zone.  I added a fix to address that.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-:-NuaE3Om4FxB-b7sKRNL:b-8zy47p)* status: 16 of 25 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts` line 85 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NuY8GEv5cHbDZlXGiCz:-NuZO7Sm0UOltJb7VBIR:b-fyel52) ([raw file](https://github.com/sillsdev/web-xforge/blob/dbe5e644283a0d1c0a38ac2f0eb04d89c846526b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts#L85)):*

@RaymondLuong3 This is a TypeScirpt type guard. It is telling the TypeScript compiler that this function checks whether the value is a TouchEvent.

See https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

That means later you can do:
``` ts
if (isTouchEvent(value)) {
  // we can now treat value as a TouchEvent
```
</blockquote></details>

I was drafting a comment, but Nathaniel beat me to it!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts` line 86 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NuZy-Ah0tBj7_DeyB5g:-NuaDRouBuF8Gly0I1SX:b-14vja4) ([raw file](https://github.com/sillsdev/web-xforge/blob/dbe5e644283a0d1c0a38ac2f0eb04d89c846526b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts#L86)):*

@siltomato Usually we do null checks with `event != null`, which guards against both `undefined` and `null`. 

You also can't misspell `null`, since it's a keyword (at least not without it being kind of obvious), but you can misspell a string and accidentally have an incorrect check (though the compiler actually does catch it if you make this mistake, so maybe this doesn't make much of a difference. But whenever I see undefined in a string I always feel compelled to check it super closely to make sure it was spelled correctly).
</blockquote></details>

Just checking `TouchEvent != null` throws a reference error, which is why I was checking the type, but I could change it to `window.TouchEvent`.

Nateowami (commented):
:+1: Drop zone fix works well.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-:-Nuf2GN23xqcmey9hUX9:b-n9bqvh)* status: 16 of 25 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts` line 86 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NuZy-Ah0tBj7_DeyB5g:-Nuf2LTe3hJ0ZitAyvEe:b-32ot6f) ([raw file](https://github.com/sillsdev/web-xforge/blob/dbe5e644283a0d1c0a38ac2f0eb04d89c846526b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts#L86)):*

Just checking `TouchEvent != null` throws a reference error, which is why I was checking the type, but I could change it to `window.TouchEvent`.
</blockquote></details>

I just realized that this function isn't doing what I assumed it was doing, which is checking if *the event* is undefined. I was expecting to see that so much that I didn't notice that wasn't what you were actually checking.

siltomato (commented):
Cool, that is much better.  Also, I included a fix for the editor history scrolling, which isn't really relevant to this PR, but the tester mentioned it, and it was a quick fix.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-:-NufEUCz0ph-KgaGLbaD:bi8bgpm)* status: 16 of 26 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts` line 86 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2363#-NuZy-Ah0tBj7_DeyB5g:-NufEQKP-tadf4YSU-US:bntlroy) ([raw file](https://github.com/sillsdev/web-xforge/blob/dbe5e644283a0d1c0a38ac2f0eb04d89c846526b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts#L86)):*

I just realized that this function isn't doing what I assumed it was doing, which is checking if *the event* is undefined. I was expecting to see that so much that I didn't notice that wasn't what you were actually checking.
</blockquote></details>

No prob!

PR #2369, opened 2024-03-11
SF-2598 Stop unnecessary ops created on sync

pmachapman:
This PR fixes a bug where unnecessary ops that modify the `cid` are created when those ops are preceded by ops that have no properties.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2369)

### DIFF
--- a/src/SIL.XForge/Realtime/RichText/Delta.cs
+++ b/src/SIL.XForge/Realtime/RichText/Delta.cs
@@ -357,8 +357,10 @@ public class Delta
     /// <summary>
     /// Test for value equality of two JTokens while ignoring the cid object property in char nodes.
     /// </summary>
-    static bool JTokenDeepEqualsIgnoreCharId(JToken a, JToken b)
+    private static bool JTokenDeepEqualsIgnoreCharId(JToken? a, JToken? b)
     {
+        if (a == null && b == null)
+            return true;
         if (b == null)
             return false;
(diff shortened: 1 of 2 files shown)


nigel-wells (commented):
The fix seems simple enough.

If I remove the fix, however, and run your tests then only `Diff_CharIdShouldNotChangeWhenAnPrecedingObjectHasNoAttributes` fails. Is that the only test targeting this fix and the other tests where just added for coverage or are they not quite testing what is needed?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2369#-:-NsuWYZ40WW5e0zmoAjR:b-577wm3)* status: 0 of 2 files reviewed, all discussions resolved

pmachapman (commented):
I added the other tests to cover existing scenarios that worked (and that I felt were not covered), and to ensure that I did not break them. I couldn't think of a good way to document that rationale in the tests, at least in a way that would make sense when we looked at the tests in 1 or 2 years time.

It is correct behavior that only one failed when you reverted the fix.

If you think the additional tests are overkill, I can remove them?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2369#-:-NsuZSt-4kOlvkhtFXcI:b-en7413)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @nigel-wells)

PR #2390, opened 2024-03-29
SF-2643 Fix editor jump to top on paste

siltomato:
When pasting text in the target editor in chrome (no issue in firefox), the editor would jump the scroll to the top.  This PR fixes this issue as well as adds tests to ensure the note FAB is tracking the selected verse while staying within the visible scroll area.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2390)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
@@ -81,10 +81,10 @@
               <span class="copyright-more-info" (click)="showCopyrightNotice('source')">{{ t("more_info") }}</span>
             </div>
           </app-notice>
-          <div #sourceSplitContainer class="container-for-split">
+          <div class="container-for-split">
             <!-- dir must be set on as-split-area because as-split is always set with dir=ltr -->
             <as-split direction="vertical" [style.height]="sourceSplitHeight" [dir]="i18n.direction">
-              <as-split-area #sourceScrollContainer [size]="75" [minSize]="20">
+              <as-split-area [size]="75" [minSize]="20">
                 <div class="text-container">
                   <app-text
                     #source
@@ -149,16 +149,10 @@
           <app-notice *ngIf="showNoEditPermissionMessage" type="info" icon="info" class="no-edit-permission-message">
             {{ t("no_permission_edit_chapter", { userRole: userRoleStr }) }}
           </app-notice>
-          <div #targetSplitContainer class="container-for-split">
+          <div class="container-for-split">
             <!-- dir must be set on as-split-area because as-split is always set with dir=ltr -->
             <as-split direction="vertical" [style.height]="targetSplitHeight">
-              <as-split-area
-                #targetScrollContainer
-                [size]="75"
-                [minSize]="20"
-                [dir]="i18n.direction"
-                [class.has-draft]="showPreviewDraft"
-              >
+              <as-split-area [size]="75" [minSize]="20" [dir]="i18n.direction" [class.has-draft]="showPreviewDraft">
                 <div class="text-container" [dir]="isTargetRightToLeft ? 'rtl' : 'ltr'">
                   <app-text
                     #target
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss
@@ -114,7 +114,7 @@ app-text {
 .insert-note-fab {
   position: absolute;
   visibility: hidden;
-  inset-inline-end: 2px;
+  inset-inline-end: 10px;
 }
 
 .fab-bottom-sheet {
@@ -171,14 +171,18 @@ app-text {
   overflow: hidden;
 }
 
-as-split-area.has-draft {
-  border: 4px dashed vars.$draft-color;
+as-split-area {
+  overflow: hidden !important;
+
+  &.has-draft {
+    border: 4px dashed vars.$draft-color;
+  }
 }
 
 .text-container {
   display: flex;
   position: relative;
-  min-height: 100%;
+  height: 100%;
 }
 
 .text-area {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -3200,6 +3200,108 @@ describe('EditorComponent', () => {
       expect(noteThreadDoc.data!.position).toEqual({ start: originalPos.start, length: originalPos.length + 1 });
       env.dispose();
     }));
+
+    it('should position FAB beside selected segment', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig();
+      env.wait();
+
+      const segmentRef = 'verse_1_1';
+
+      env.clickSegmentRef(segmentRef);
+      env.wait();
+
+      const segmentElRect = env.getSegmentElement(segmentRef)!.getBoundingClientRect();
+      const fabRect = env.insertNoteFab.nativeElement.getBoundingClientRect();
+      expect(segmentElRect.top).toEqual(fabRect.top);
+
+      env.dispose();
+    }));
+
+    it('should position FAB beside selected segment when scrolling segment in view', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig();
+      env.wait();
+
+      // Set window size to be narrow to test scrolling
+      const contentContainer: HTMLElement = document.getElementsByClassName('content')[0] as HTMLElement;
+      Object.assign(contentContainer.style, { width: '360px', height: '300px' });
+
+      const segmentRef = 'verse_1_1';
+
+      // Select segment
+      env.clickSegmentRef(segmentRef);
+      env.wait();
+
+      // Scroll, keeping selected segment in view
+      const scrollContainer: Element = env.component['targetScrollContainer'] as Element;
+      scrollContainer.scrollTop = 20;
+      scrollContainer.dispatchEvent(new Event('scroll'));
+
+      const segmentElRect = env.getSegmentElement(segmentRef)!.getBoundingClientRect();
+      const fabRect = env.insertNoteFab.nativeElement.getBoundingClientRect();
+      expect(fabRect.top).toEqual(segmentElRect.top);
+
+      env.dispose();
+    }));
+
+    it('should position FAB within scroll container when scrolling segment above view', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.setProjectUserConfig();
+      env.wait();
+
+      // Set window size to be narrow to test scrolling
+      const contentContainer: HTMLElement = document.getElementsByClassName('content')[0] as HTMLElement;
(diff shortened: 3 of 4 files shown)


Nateowami (changes requested):
I've partially reviewed this. I have one concern so far, and I'm leaving my comments before finishing the review.

> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2390#-:-NuQe4mm3lwrk8KL0hkS:b-hbvyuc)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 770 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2390#-NuQcIQX064ytcLe0IgL:-NuQcIQY1p9ELVEuW7jG:b7qnekd) ([raw file](https://github.com/sillsdev/web-xforge/blob/4106e49039213d34ebc49a52b3080f039e01b502/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L770)):*
> ```XML
> 
>     // Throttle bursts of sync scroll requests
>     this.syncScrollRequested$.pipe(takeUntilDestroyed(this.destroyRef), throttleTime(200)).subscribe(() => {
> ```

`throttleTime` by default throws away trailing values. That is to say, if a sync scroll request happens, it will execute this function, but if another request happens within that 200ms window of time, it will not trigger anything. I'm pretty sure that would result in incorrect behavior, right? I think you probably want:
``` ts
throttleTime(100, asyncScheduler, { leading: true, trailing: true })
```
Where `asyncScheduler` is imported from `'rxjs'`:
``` ts
import { asyncScheduler } from 'rxjs'; 
```
See the documentation: https://rxjs.dev/api/index/function/throttleTime
The marble diagram on that page *is not for the default behavior, it's for with a config set to `{ leading: true, trailing: true }`.*

Nateowami (commented):
> Reviewed 1 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2390)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 228 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2390#-NuQf6e9ETplgnbhOMs6:-NuQf6e9ETplgnbhOMs7:b-8c9uu3) ([raw file](https://github.com/sillsdev/web-xforge/blob/4106e49039213d34ebc49a52b3080f039e01b502/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L228)):*
> ```XML
>   private scrollSubscription?: Subscription;
>   private readonly fabDiameter = 40;
>   private readonly fabCushion = 5;
> ```

Before reading through this file, I thought this was referring to the spacing between the right side of the editor, and the FAB. But it's actually the vertical cushion. Maybe it could be a bit clearer if it were called `fabVerticalCushion` or something? Just a minor point; doesn't need to hold up this PR.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3228 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2390#-NuQh3Zh-2fvwhT1_u-O:-NuQh3Zh-2fvwhT1_u-P:b-rbo4en) ([raw file](https://github.com/sillsdev/web-xforge/blob/4106e49039213d34ebc49a52b3080f039e01b502/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3228)):*
> ```XML
>       // Set window size to be narrow to test scrolling
>       const contentContainer: HTMLElement = document.getElementsByClassName('content')[0] as HTMLElement;
>       Object.assign(contentContainer.style, { width: '360px', height: '300px' });
> ```

Curious why the use of `Object.assign` here? Doesn't this mean you're removing any styles that might have been set directly on this element? For example, if we had set a the overflow property directly on this element, aren't you now (inadvertently) removing it? Is there a reason I'm not seeing?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3272 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2390#-NuQj2XY-M_ZB1TzOD2v:-NuQj2XY-M_ZB1TzOD2w:b-urrpzk) ([raw file](https://github.com/sillsdev/web-xforge/blob/4106e49039213d34ebc49a52b3080f039e01b502/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3272)):*
> ```XML
> 
>       const fabRect = env.insertNoteFab.nativeElement.getBoundingClientRect();
>       expect(fabRect.top).toEqual(scrollContainerRect.top + env.component['fabCushion']);
> ```

Making `fabCushion` private and doing this works fine, but I personally think making it public is simpler.
Pros:
- Easy to find references (useful when seeing if something can be removed)
- Easy to rename
- Easier to write tests without misspelling the name by accident

Cons:
1. Someone somehow writes a component in the future that references `fabCushion` and becomes more tightly coupled to the editor component than necessary.
2. ...
3. Bad things happen because complexity is too great.

I have trouble figuring out what step 2 would be here. It just seems like the OOP principle of encapsulation doesn't particularly matter for this type of thing.

Nateowami (commented):
> Reviewed 1 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2390)* status: all files reviewed, 4 unresolved discussions (waiting on @siltomato)

siltomato (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2390)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 228 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2390#-NuQf6e9ETplgnbhOMs6:-NuR9wRZ-Ge1gwUNvAaM:b-g7lpia) ([raw file](https://github.com/sillsdev/web-xforge/blob/4106e49039213d34ebc49a52b3080f039e01b502/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L228)):*

Before reading through this file, I thought this was referring to the spacing between the right side of the editor, and the FAB. But it's actually the vertical cushion. Maybe it could be a bit clearer if it were called `fabVerticalCushion` or something? Just a minor point; doesn't need to hold up this PR.
</blockquote></details>

Good point. Changed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 770 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2390#-NuQcIQX064ytcLe0IgL:-NuRDMkm24u1EyEGQ0cs:byvrm5q) ([raw file](https://github.com/sillsdev/web-xforge/blob/4106e49039213d34ebc49a52b3080f039e01b502/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L770)):*

`throttleTime` by default throws away trailing values. That is to say, if a sync scroll request happens, it will execute this function, but if another request happens within that 200ms window of time, it will not trigger anything. I'm pretty sure that would result in incorrect behavior, right? I think you probably want:
``` ts
throttleTime(100, asyncScheduler, { leading: true, trailing: true })
```
Where `asyncScheduler` is imported from `'rxjs'`:
``` ts
import { asyncScheduler } from 'rxjs'; 
```
See the documentation: https://rxjs.dev/api/index/function/throttleTime
The marble diagram on that page *is not for the default behavior, it's for with a config set to `{ leading: true, trailing: true }`.*
</blockquote></details>

Makes sense.  I was trying to avoid calling sync twice every time, but it should be fast.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3228 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2390#-NuQh3Zh-2fvwhT1_u-O:-NuR74kaAijxq6-lqaK8:blo1fw4) ([raw file](https://github.com/sillsdev/web-xforge/blob/4106e49039213d34ebc49a52b3080f039e01b502/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3228)):*

Curious why the use of `Object.assign` here? Doesn't this mean you're removing any styles that might have been set directly on this element? For example, if we had set a the overflow property directly on this element, aren't you now (inadvertently) removing it? Is there a reason I'm not seeing?
</blockquote></details>

If I'm not mistaken, `Object.assign` will only overwrite the properties in the target object that have the same name as properties in the source object.  My experimenting in browser dev tools appears to confirm this.  If this is correct, then my idea is to express the styles more naturally than

```
container.style.width = '360px';
container.style.height= '300px';
```

At least, it feels more natural (css-like syntax) to me.  
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3272 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2390#-NuQj2XY-M_ZB1TzOD2v:-NuR8OgL0HmhOlR0TaRJ:b-icfesn) ([raw file](https://github.com/sillsdev/web-xforge/blob/4106e49039213d34ebc49a52b3080f039e01b502/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3272)):*

Making `fabCushion` private and doing this works fine, but I personally think making it public is simpler.
Pros:
- Easy to find references (useful when seeing if something can be removed)
- Easy to rename
- Easier to write tests without misspelling the name by accident

Cons:
1. Someone somehow writes a component in the future that references `fabCushion` and becomes more tightly coupled to the editor component than necessary.
2. ...
3. Bad things happen because complexity is too great.

I have trouble figuring out what step 2 would be here. It just seems like the OOP principle of encapsulation doesn't particularly matter for this type of thing.
</blockquote></details>

Makes sense.  The refactoring pro is a big one.  I suppose I didn't like the idea of the test driving the access modifier, but it probably doesn't matter on a practical level.

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2390)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3228 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2390#-NuQh3Zh-2fvwhT1_u-O:-NuREptP8sGFKC-RTFMV:b-wqnmt0) ([raw file](https://github.com/sillsdev/web-xforge/blob/4106e49039213d34ebc49a52b3080f039e01b502/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3228)):*

If I'm not mistaken, `Object.assign` will only overwrite the properties in the target object that have the same name as properties in the source object.  My experimenting in browser dev tools appears to confirm this.  If this is correct, then my idea is to express the styles more naturally than

```
container.style.width = '360px';
container.style.height= '300px';
```

At least, it feels more natural (css-like syntax) to me.
</blockquote></details>

Ah, makes sense. `Object.assign` is just a convenience method. For some reason I was confusing it with the behavior of `Object.defineproperty`, which has interesting capabilities (like making a property read-only).

PR #2391, opened 2024-04-01
SF-2443 Apply button for draft tabs

josephmyers:
So now that SF-2589 is in, the task originally planned for SF-2445 has been merged into this task.

-------------------------------------------------

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.spec.ts
@@ -1,0 +1,68 @@
+import { fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { DeltaStatic } from 'quill';
+import { TextData } from 'realtime-server/lib/esm/scriptureforge/models/text-data';
+import * as RichText from 'rich-text';
+import { mock, when } from 'ts-mockito';
+import { TestRealtimeModule } from 'xforge-common/test-realtime.module';
+import { TestRealtimeService } from 'xforge-common/test-realtime.service';
+import { configureTestingModule } from 'xforge-common/test-utils';
+import { getCombinedVerseTextDoc, getPoetryVerseTextDoc, getTextDoc } from '../shared/test-utils';
+import { SF_TYPE_REGISTRY } from './models/sf-type-registry';
+import { TextDoc, TextDocId } from './models/text-doc';
+import { SFProjectService } from './sf-project.service';
+import { TextDocService } from './text-doc.service';
+
+const mockProjectService = mock(SFProjectService);
+
+describe('TextDocService', () => {
+  configureTestingModule(() => ({
+    imports: [TestRealtimeModule.forRoot(SF_TYPE_REGISTRY)],
+    providers: [{ provide: SFProjectService, useMock: mockProjectService }]
+  }));
+
+  it('should overwrite text doc', fakeAsync(() => {
+    const env = new TestEnvironment();
+    const newDelta: DeltaStatic = getCombinedVerseTextDoc(env.textDocId) as DeltaStatic;
+
+    env.textDocService.overwrite(env.textDocId, newDelta);
+    tick();
+
+    expect(env.getTextDoc(env.textDocId).data?.ops).toEqual(newDelta.ops);
+  }));
+
+  it('should emit diff', fakeAsync(() => {
+    const env = new TestEnvironment();
+    const origDelta: DeltaStatic = env.getTextDoc(env.textDocId).data as DeltaStatic;
+    const newDelta: DeltaStatic = getPoetryVerseTextDoc(env.textDocId) as DeltaStatic;
+    const diff: DeltaStatic = origDelta.diff(newDelta);
+
+    env.textDocService.getLocalSystemChanges$(env.textDocId).subscribe(emittedDiff => {
+      expect(emittedDiff.ops).toEqual(diff.ops);
+    });
+
+    env.textDocService.overwrite(env.textDocId, newDelta);
+    tick();
+  }));
+});
+
+class TestEnvironment {
+  readonly textDocId = new TextDocId('project01', 40, 1);
+  readonly textDocService: TextDocService = TestBed.inject(TextDocService);
+  private readonly realtimeService: TestRealtimeService = TestBed.inject(TestRealtimeService);
+
+  constructor() {
+    this.realtimeService.addSnapshot<TextData>(TextDoc.COLLECTION, {
+      id: this.textDocId.toString(),
+      data: getTextDoc(this.textDocId),
+      type: RichText.type.name
+    });
+
+    when(mockProjectService.getText(this.textDocId)).thenCall(id =>
+      this.realtimeService.subscribe(TextDoc.COLLECTION, id.toString())
+    );
+  }
+
+  getTextDoc(textId: TextDocId): TextDoc {
+    return this.realtimeService.get<TextDoc>(TextDoc.COLLECTION, textId.toString());
+  }
+}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts
@@ -1,0 +1,62 @@
+import { Injectable } from '@angular/core';
+import { DeltaStatic } from 'quill';
+import { TextData } from 'realtime-server/lib/esm/scriptureforge/models/text-data';
+import { Delta } from 'rich-text';
+import { Observable, Subject } from 'rxjs';
+import { TextDoc, TextDocId } from './models/text-doc';
+import { SFProjectService } from './sf-project.service';
+
+@Injectable({
+  providedIn: 'root'
+})
+export class TextDocService {
+  private localSystemTextDocChangesMap = new Map<string, Subject<TextData>>();
+
+  constructor(private readonly projectService: SFProjectService) {}
+
+  /**
+   * Overwrites the specified text doc with the specified delta and then notifies listeners of the changes.
+   * @param {TextDocId} textDocId The id for text doc.
+   * @param {DeltaStatic} newDelta The ops to overwrite the text doc with.
+   */
+  async overwrite(textDocId: TextDocId, newDelta: DeltaStatic): Promise<void> {
+    const textDoc: TextDoc = await this.projectService.getText(textDocId);
+
+    if (textDoc.data?.ops == null) {
+      throw new Error(`No TextDoc data for ${textDocId}`);
+    }
+
+    const origDelta: DeltaStatic = new Delta(textDoc.data.ops);
+    const diff: DeltaStatic = origDelta.diff(newDelta);
+
+    // Update text doc directly
+    await textDoc.submit(diff);
+
+    // Notify so that TextViewModels can update
+    this.getLocalSystemChangesInternal$(textDocId).next(diff);
+  }
+
+  /**
+   * Gets an observable that emits the diff when the local system has made changes to the specified text doc.
+   *
+   * This is useful for updating the editor to reflect changes that are not made by user edits,
+   * such as when a draft or history revision is applied.
+   * @param {TextDocId} textDocId The id for text doc to listen to.
+   * @returns {Observable<TextData>} An observable that emits the diff ops.
+   */
+  getLocalSystemChanges$(textDocId: TextDocId): Observable<TextData> {
+    return this.getLocalSystemChangesInternal$(textDocId).asObservable();
+  }
+
+  private getLocalSystemChangesInternal$(textDocId: TextDocId): Subject<TextData> {
+    const key: string = textDocId.toString();
+    let subject: Subject<TextData> | undefined = this.localSystemTextDocChangesMap.get(key);
+
+    if (subject == null) {
+      subject = new Subject<TextData>();
+      this.localSystemTextDocChangesMap.set(key, subject);
+    }
+
+    return subject;
+  }
+}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.scss
@@ -13,6 +13,8 @@
     --sf-tab-header-inactive-button-hover-background-color: #c0c0c0;
     --sf-tab-header-button-hover-background-color: #e0e0e0;
     --sf-tab-header-group-scroll-button-background-color: #f7f7f7;
+
+    --sf-tab-toolbar-height: 57px;
   }
 }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text-view-model.ts
@@ -1,9 +1,11 @@
+import { DestroyRef, Injectable } from '@angular/core';
+import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { VerseRef } from '@sillsdev/scripture';
 import cloneDeep from 'lodash-es/cloneDeep';
 import Quill, { DeltaOperation, DeltaStatic, RangeStatic, Sources, StringMap } from 'quill';
-import { Subscription } from 'rxjs';
-
+import { merge, Subscription } from 'rxjs';
 import { Delta, TextDoc, TextDocId } from '../../core/models/text-doc';
+import { TextDocService } from '../../core/text-doc.service';
 import { getVerseStrFromSegmentRef, isBadDelta } from '../utils';
 import { getAttributesAtPosition } from './quill-scripture';
 import { USFM_STYLE_DESCRIPTIONS } from './usfm-style-descriptions';
@@ -117,11 +119,12 @@ class SegmentInfo {
  * consistent and correct.
  * See text.component.spec.ts for some unit tests.
  */
+@Injectable()
 export class TextViewModel {
   editor?: Quill;
 
   private readonly _segments: Map<string, RangeStatic> = new Map<string, RangeStatic>();
-  private remoteChangesSub?: Subscription;
+  private changesSub?: Subscription;
   private onCreateSub?: Subscription;
   private textDoc?: TextDoc;
   private textDocId?: TextDocId;
@@ -130,6 +133,8 @@ export class TextViewModel {
    * These elements are in addition to the text data i.e. Note threads
    */
   private _embeddedElements: Map<string, EmbedPosition> = new Map<string, EmbedPosition>();
+
+  constructor(private readonly destroyRef: DestroyRef, private readonly textDocService: TextDocService) {}
 
   get segments(): IterableIterator<[string, RangeStatic]> {
     return this._segments.entries();
@@ -182,13 +187,17 @@ export class TextViewModel {
     this.textDoc = textDoc;
     editor.setContents(this.textDoc.data as DeltaStatic);
     editor.history.clear();
+
     if (subscribeToUpdates) {
-      this.remoteChangesSub = this.textDoc.remoteChanges$.subscribe(ops => {
-        const deltaWithEmbeds: DeltaStatic = this.addEmbeddedElementsToDelta(ops as DeltaStatic);
-        editor.updateContents(deltaWithEmbeds, 'api');
-      });
+      this.changesSub = merge(this.textDocService.getLocalSystemChanges$(textDocId), this.textDoc.remoteChanges$)
+        .pipe(takeUntilDestroyed(this.destroyRef))
+        .subscribe(ops => {
+          const deltaWithEmbeds: DeltaStatic = this.addEmbeddedElementsToDelta(ops as DeltaStatic);
+          editor.updateContents(deltaWithEmbeds, 'api');
+        });
     }
-    this.onCreateSub = this.textDoc.create$.subscribe(() => {
+
+    this.onCreateSub = this.textDoc.create$.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(() => {
       if (textDoc.data != null) {
         editor.setContents(textDoc.data as DeltaStatic);
       }
@@ -198,17 +207,11 @@ export class TextViewModel {
 
   /** Break the association of the editor with the currently associated textdoc. */
   unbind(): void {
-    if (this.remoteChangesSub != null) {
-      this.remoteChangesSub.unsubscribe();
-    }
-    if (this.onCreateSub != null) {
-      this.onCreateSub.unsubscribe();
-    }
+    this.changesSub?.unsubscribe();
+    this.onCreateSub?.unsubscribe();
     this.textDoc = undefined;
 
-    if (this.editor != null) {
-      this.editor.setText('', 'silent');
-    }
+    this.editor?.setText('', 'silent');
     this._segments.clear();
     this._embeddedElements.clear();
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -83,7 +83,8 @@ export interface EmbedsByVerse {
 @Component({
   selector: 'app-text',
   templateUrl: './text.component.html',
-  styleUrls: ['./text.component.scss']
+  styleUrls: ['./text.component.scss'],
+  providers: [TextViewModel] // New instance for each text component
 })
 export class TextComponent extends SubscriptionDisposable implements AfterViewInit, OnDestroy {
   @ViewChild('quillEditor', { static: true, read: ElementRef }) quill!: ElementRef;
@@ -219,7 +220,6 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
   private _isRightToLeft: boolean = false;
   private _modules: any = this.DEFAULT_MODULES;
   private _editor?: Quill;
-  private viewModel = new TextViewModel();
   private _segment?: Segment;
   private initialTextFetched: boolean = false;
   private initialSegmentRef?: string;
@@ -249,7 +249,8 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     private readonly projectService: SFProjectService,
     private readonly onlineStatusService: OnlineStatusService,
     private readonly transloco: TranslocoService,
-    private readonly userService: UserService
+    private readonly userService: UserService,
+    private readonly viewModel: TextViewModel
   ) {
     super();
     let localCursorColor = localStorage.getItem(this.cursorColorStorageKey);
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.service.spec.ts
@@ -303,4 +303,37 @@ describe('DraftGenerationService', () => {
       });
     });
   });
+
+  describe('draftExists', () => {
+    it('should return true if draft exists', done => {
+      const preTranslationData = {
+        data: {
+          preTranslations: [
+            { reference: 'verse_3_16', translation: 'For God so loved the world' },
+            { reference: 'verse_1_1', translation: 'In the beginning was the Word' }
+          ]
+        }
+      };
+
+      httpClient.get = jasmine.createSpy().and.returnValue(of(preTranslationData));
+      service.draftExists(projectId, 43, 3).subscribe(result => {
+        expect(result).toBe(true);
+        done();
+      });
+    });
+
+    it('should return false if draft does not exist', done => {
+      const preTranslationData = {
+        data: {
+          preTranslations: []
+        }
+      };
+
+      httpClient.get = jasmine.createSpy().and.returnValue(of(preTranslationData));
+      service.draftExists(projectId, 43, 3).subscribe(result => {
+        expect(result).toBe(false);
+        done();
+      });
+    });
+  });
 });
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.service.ts
@@ -176,6 +176,17 @@ export class DraftGenerationService {
   }
 
   /**
+   * Determines if a draft exists for the specified book/chapter.
+   * @param projectId The SF project id for the target translation.
+   * @param book The book number.
+   * @param chapter The chapter number.
+   * @returns An observable indicating if a draft exists.
+   */
+  draftExists(projectId: string, book: number, chapter: number): Observable<boolean> {
+    return this.getGeneratedDraft(projectId, book, chapter).pipe(map(draft => Object.keys(draft).length > 0));
+  }
+
+  /**
    * Calls the machine api to start a pretranslation build job.
    * This should only be called if no build is currently active.
    * @param buildConfig The build configuration.
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts
@@ -1,12 +1,18 @@
 import { Injectable } from '@angular/core';
 import { VerseRef } from '@sillsdev/scripture';
-import { DeltaOperation } from 'quill';
+import { DeltaOperation, DeltaStatic } from 'quill';
+import { TextDocId } from 'src/app/core/models/text-doc';
 import { isString } from '../../../../type-utils';
 import { getVerseRefFromSegmentRef, verseSlug } from '../../../shared/utils';
 import { DraftSegmentMap } from '../draft-generation';
 
 export interface DraftMappingOptions {
   overwrite?: boolean;
+}
+
+export interface DraftDiff {
+  id: TextDocId;
+  ops: DeltaStatic;
 }
 
 @Injectable({
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.html
@@ -13,6 +13,24 @@
   </app-notice>
 </ng-container>
 
+<div class="toolbar">
+  <div class="apply-draft-button-container">
+    <span *ngIf="isDraftApplied && isDraftReady" class="draft-indicator">
+      <mat-icon class="check-icon">check</mat-icon>
+      <transloco key="editor_draft_tab.draft_indicator_applied"></transloco>
+    </span>
+    <button
+      *ngIf="!isDraftApplied && isDraftReady && canApplyDraft"
+      mat-flat-button
+      color="primary"
+      (click)="applyDraft()"
+    >
+      <mat-icon>auto_awesome</mat-icon>
+      <transloco key="editor_draft_tab.apply_to_project"></transloco>
+    </button>
+  </div>
+</div>
+
 <app-text
   [isReadOnly]="true"
   [subscribeToUpdates]="false"
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss
@@ -1,3 +1,10 @@
+@use 'src/variables';
+
+:host {
+  display: flex;
+  flex-direction: column;
+}
+
 .draft-empty,
 .draft-unknown {
   display: none;
@@ -5,4 +12,31 @@
 
 app-notice {
   margin: 1em;
+}
+
+.toolbar {
+  position: sticky;
+  top: 0;
+  background: #fff;
+  z-index: 1;
+  display: flex; //fill vertical space
+}
+
+.apply-draft-button-container {
+  flex-grow: 1;
+  display: flex;
+  justify-content: flex-end;
+  padding: 0.5em;
+  border-bottom: 1px solid var(--sf-tab-group-border-color);
+  height: var(--sf-tab-toolbar-height);
+}
+
+.draft-indicator {
+  color: variables.$greenDark;
+  font-style: italic;
+  font-size: 18px;
+  font-weight: 400;
+  line-height: 24px;
+  display: inline-flex;
+  align-items: center;
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.spec.ts
@@ -7,6 +7,7 @@ import { Delta } from 'rich-text';
 import { of, throwError } from 'rxjs';
 import { anything, mock, verify, when } from 'ts-mockito';
 import { ActivatedProjectService } from 'xforge-common/activated-project.service';
+import { DialogService } from 'xforge-common/dialog.service';
 import { I18nService } from 'xforge-common/i18n.service';
 import { OnlineStatusService } from 'xforge-common/online-status.service';
 import { TestOnlineStatusModule } from 'xforge-common/test-online-status.module';
@@ -14,6 +15,7 @@ import { TestOnlineStatusService } from 'xforge-common/test-online-status.servic
 import { TestRealtimeModule } from 'xforge-common/test-realtime.module';
 import { configureTestingModule, TestTranslocoModule } from 'xforge-common/test-utils';
 import { SF_TYPE_REGISTRY } from '../../../core/models/sf-type-registry';
+import { TextDocService } from '../../../core/text-doc.service';
 import { SharedModule } from '../../../shared/shared.module';
 import { EDITOR_READY_TIMEOUT } from '../../../shared/text/text.component';
 import { DraftSegmentMap } from '../../draft-generation/draft-generation';
@@ -25,6 +27,8 @@ const mockDraftGenerationService = mock(DraftGenerationService);
 const mockActivatedProjectService = mock(ActivatedProjectService);
 const mockDraftViewerService = mock(DraftViewerService);
 const mockI18nService = mock(I18nService);
+const mockDialogService = mock(DialogService);
+const mockTextDocService = mock(TextDocService);
 
 describe('EditorDraftComponent', () => {
   let fixture: ComponentFixture<EditorDraftComponent>;
@@ -46,7 +50,9 @@ describe('EditorDraftComponent', () => {
       { provide: DraftGenerationService, useMock: mockDraftGenerationService },
       { provide: DraftViewerService, useMock: mockDraftViewerService },
       { provide: I18nService, useMock: mockI18nService },
-      { provide: OnlineStatusService, useClass: TestOnlineStatusService }
+      { provide: OnlineStatusService, useClass: TestOnlineStatusService },
+      { provide: DialogService, useMock: mockDialogService },
+      { provide: TextDocService, useMock: mockTextDocService }
     ]
   }));
 
@@ -60,6 +66,7 @@ describe('EditorDraftComponent', () => {
     component.bookNum = 1;
     component.chapter = 1;
     component.isRightToLeft = false;
+    component.ngOnChanges();
   });
 
   it('should handle offline when component created', fakeAsync(() => {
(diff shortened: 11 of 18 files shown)


Nateowami (changes requested):
Not a full review, but leaving a few comments.

> Reviewed 2 of 10 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-:-NvAX7LaCmcWlxtivJaw:b-vy4gzo)* status: 2 of 10 files reviewed, 4 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.html` line 20 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-Nv8K1py5rY2aXkM4h7A:-Nv8K1pz0ynPIj_9tR7V:b-m50hfr) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.html#L20)):*
> ```HTML
>     <span *ngIf="isDraftApplied && hasDraft" class="draft-indicator">
>       <transloco key="editor_draft_tab.draft_indicator_applied"></transloco>
>       <mat-icon class="check-icon">check</mat-icon>
> ```

Can we put the check _before_ the text?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss` line 16 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-Nv8Kuq2A8daLw2qJk_3:-Nv8Kuq2A8daLw2qJk_4:b-3b4sxh) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss#L16)):*
> ```SCSS
> 
> .toolbar {
>   position: sticky;
> ```

I don't think this is the right approach. The problem isn't that it needs to be sticky, have a white background, and have a z-index of 1. The problem is that _the wrong element is scrolling_.

But, it's a minor point, and I don't really want it to hold up this PR. Just something I wanted to point out.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 242 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-Nv8HdUOESU3gn4PFTMc:-Nv8HdUOESU3gn4PFTMd:b6vy0hu) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L242)):*
> ```JSON
>   },
>   "editor_draft_tab": {
>     "apply_to_project": "Apply to project",
> ```

Let's change this to "Add to project"
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 243 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-Nv8IV8zBJtKJXyWGDDK:-Nv8IV9-Cuol8FxZI6G_:bjlwp43) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L243)):*
> ```JSON
>   "editor_draft_tab": {
>     "apply_to_project": "Apply to project",
>     "draft_indicator_applied": "DRAFT APPLIED",
> ```

All caps do not work in all languages and writing systems. I would suggest we avoid it. Also, draft applied is less clear than it could be. How about "Draft has been added to the project"? I think it will be clearer not just to users that are not native English speakers, but also for localizers to understand what it means and how to translate it.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 246 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-Nv8I-N-9LBTkgXT5rc8:-Nv8I-N-9LBTkgXT5rc9:b644p55) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L246)):*
Instead of explaining followed by a "continue" question, how about:

> Adding the draft will overwrite the current chapter. Are you sure you want to continue?
> [ Cancel ] - or - [ Overwrite chapter ]

Nateowami:
> The diff used for applying should also match Draft Previewer, but it is pretty easy to break. I'm leaning against putting a great amount of effort into this part, given that we'll be changing it when the USFM roundtripping ([SF-2589](https://jira.sil.org/browse/SF-2589)) is in.

It's already in. Have you updated for that already? I would think yes, but the PR description hasn't been updated, so that makes me think no. I haven't read through all the changes yet.

josephmyers:
Yes. I'll pull that part out of the description.

nigel-wells (changes requested):
This seems to work pretty well - good job.
Just minor thing and one bigger question about how we could tackle the duplicated methods.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-:-NvFIU6NERBSOVt7zJ3i:bpjy3rh)* status: all files reviewed, 12 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFBi7N5p5OkUMj9-Zw:-NvFBi7N5p5OkUMj9-Zx:b-cdu67w) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L13)):*
> ```XML
> }
> 
> export interface DraftDiff {
> ```

It's not really a diff is it? What about `AppliedDraftOps` or something similar?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 749 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvF6ciECJB5ya7jHIns:-NvF6ciECJB5ya7jHInt:b-mpmst8) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L749)):*
> ```XML
> 
>           this.draftAppliedSub = this.draftViewerService.draftApplied.subscribe(e => {
>             if (this.target?.id?.toString() === e.id.toString()) {
> ```

You could `pipe` and `filter` this
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3743 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFGhkTCvcBUjwjYot_:-NvFGhkTCvcBUjwjYota:b-ejhyhp) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3743)):*
> ```XML
>     }));
> 
>     it('updates editor when a draft is applied', fakeAsync(() => {
> ```

Can you also assert that a `hasDraft` was `true` and that the existing contents were different from the new ones.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss` line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFAbvj2xtK1zkUonrI:-NvFAbvj2xtK1zkUonrJ:b-y8o1o4) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss#L33)):*
> ```SCSS
> 
> .draft-indicator {
>   color: green;
> ```

Let's use one of our existing colors - `$greenDark` is closest to what you have.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 130 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFCUNp1uJHPAn_rXm6:-NvFCUNp1uJHPAn_rXm7:b1komuo) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L130)):*
> ```XML
>         // Set the draft editor with the pre-translation segments
>         this.draftText.editor?.setContents(new Delta(draftOps), 'api');
>         this.isDraftApplied = (await this.getDiff()).length() === 0;
> ```

Just noting here that I have a book/chapter which this registered as `true` even though I hadn't applied anything. The source had text but the draft only had verse numbers. 

This could easily be bad data on my local machine (probably) but wanted to point it out in case there is more to it.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 175 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFDEDl01r5yXeQz9_s:-NvFDEDl01r5yXeQz9_t:b-t7v7mh) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L175)):*
> ```XML
>       }
> 
>       const isInsertBlank = (isString(op.insert) && op.insert.trim().length === 0) || op.insert.blank === true;
> ```

Can just inline this without creating the variable.

If the existing target was full of blanks then do we care about warning for that?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 184 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFFcwM8-LzOHTWyiT1:-NvFFcwM8-LzOHTWyiT2:bw7xvvp) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L184)):*
> ```XML
>    * This code is reimplemented from editor.component.ts
>    */
>   get canEdit(): boolean {
> ```

Is there another way we can share this logic across both components - 4 seems a bit too many to duplicate. Could they move refactored into a service?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.spec.ts` line 178 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFIOcs87-yT8Rk4Qtd:-NvFIOcs87-yT8Rk4Qte:b4sr5ws) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.spec.ts#L178)):*
> ```XML
>       )
>     ).once();
>     expect(component.isDraftApplied).toBe(true);
> ```

Let's assert this was `false` to begin with.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391)* status: all files reviewed, 12 unresolved discussions (waiting on @josephmyers and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss` line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFAbvj2xtK1zkUonrI:-NvFNhHm7Wuw3t3E7Z5R:b-8fx2u7) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss#L33)):*

Let's use one of our existing colors - `$greenDark` is closest to what you have.
</blockquote></details>

:+1: 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 175 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFDEDl01r5yXeQz9_s:-NvFNxbe1xa4qovfZyW8:b-6uv9jj) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L175)):*

Can just inline this without creating the variable.

If the existing target was full of blanks then do we care about warning for that?
</blockquote></details>

Creating the variable isn't necessary, but can often be more readable. In this case, I can read from the beginning of the line and know what we're trying to figure out (whether the op is blank) and then read the logic and make sure it is doing the right thing. If we just returned an expression, it would be necessary to spend time looking at the logic to find out what it's checking.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 184 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFFcwM8-LzOHTWyiT1:-NvFOQ0B2d4JaqeSVyIB:bix593q) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L184)):*

Is there another way we can share this logic across both components - 4 seems a bit too many to duplicate. Could they move refactored into a service?
</blockquote></details>

:+1: although I don't see an obvious solution.

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391)* status: 4 of 10 files reviewed, 11 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFBi7N5p5OkUMj9-Zw:-NvU4_pc0xOkCLysVqel:bjgzp8) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L13)):*

It's not really a diff is it? What about `AppliedDraftOps` or something similar?
</blockquote></details>

Well, it kind of is. Maybe I'm not seeing what you're saying. To me, a DeltaStatic is a diff. It's what's output by DeltaStatic.diff(), which is what I'm supplying to these ops.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 749 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvF6ciECJB5ya7jHIns:-NvU7nqQ3Var0kM8KP1k:b-af183f) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L749)):*

You could `pipe` and `filter` this
</blockquote></details>

Yep, thanks!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3743 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFGhkTCvcBUjwjYot_:-NvU9lt69noFkWbJ-6yX:b7mofu2) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3743)):*

Can you also assert that a `hasDraft` was `true` and that the existing contents were different from the new ones.
</blockquote></details>

That would be difficult in this test. I haven't set up the editor's contents or the draft contents, because it's not required by or the point of this test. This test demonstrates simply that when the editor receives the draftApplied event, it takes the passed diff and applies it. It has no bearing on the contents of either text, as that was determined in the draft tab.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.html` line 20 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-Nv8K1py5rY2aXkM4h7A:-NvUHhwF0jLXhRdEw_8R:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.html#L20)):*

Can we put the check _before_ the text?
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss` line 16 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-Nv8Kuq2A8daLw2qJk_3:-NvUHnSn4ZJDbeQeZYvs:b-l5hd0f) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss#L16)):*

I don't think this is the right approach. The problem isn't that it needs to be sticky, have a white background, and have a z-index of 1. The problem is that _the wrong element is scrolling_.

But, it's a minor point, and I don't really want it to hold up this PR. Just something I wanted to point out.
</blockquote></details>

Thanks. I'm following the pattern from the history tab. So if we want to refactor things, we'd need to get all occurrences.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss` line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFAbvj2xtK1zkUonrI:-NvUJkkL2d8HmP7x_JK6:b-h2szml) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss#L33)):*

:+1:
</blockquote></details>

I certainly can
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 130 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFCUNp1uJHPAn_rXm6:-NvUKeNH2u-uO_NVmAoO:b-g9ct8k) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L130)):*

Just noting here that I have a book/chapter which this registered as `true` even though I hadn't applied anything. The source had text but the draft only had verse numbers. 

This could easily be bad data on my local machine (probably) but wanted to point it out in case there is more to it.
</blockquote></details>

That is interesting! I can put a check in here, but how did you get an empty draft? You'll have to be the one to test this change, haha
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 175 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFDEDl01r5yXeQz9_s:-NvUNLA85tRWXbKuVv0k:bdt528l) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L175)):*

Creating the variable isn't necessary, but can often be more readable. In this case, I can read from the beginning of the line and know what we're trying to figure out (whether the op is blank) and then read the logic and make sure it is doing the right thing. If we just returned an expression, it would be necessary to spend time looking at the logic to find out what it's checking.
</blockquote></details>

Yep, thanks, Nathaniel!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 184 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFFcwM8-LzOHTWyiT1:-NvUNStYAFbbT3tPmKC2:b-23ntrq) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L184)):*

:+1: although I don't see an obvious solution.
</blockquote></details>

It will go down to 3 when we remove the draft previewer :)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.spec.ts` line 178 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFIOcs87-yT8Rk4Qtd:-NvUYhp82vUyRWcbESBB:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.spec.ts#L178)):*

Let's assert this was `false` to begin with.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 242 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-Nv8HdUOESU3gn4PFTMc:-NvU_scS05jEUa1QBrCY:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L242)):*

Let's change this to "Add to project"
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 243 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-Nv8IV8zBJtKJXyWGDDK:-NvVkX9h7awbFXvN_Z63:b4rsj1f) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L243)):*

All caps do not work in all languages and writing systems. I would suggest we avoid it. Also, draft applied is less clear than it could be. How about "Draft has been added to the project"? I think it will be clearer not just to users that are not native English speakers, but also for localizers to understand what it means and how to translate it.
</blockquote></details>

That's a lot of words. Now I do think the current wording should change, given the change from Apply to Add. This wording was okay when the Draft Previewer was done, so is it not okay now? That said, just having Added will make sense to users, who just selected Add to project. Plus, it's less to localize.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 246 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-Nv8I-N-9LBTkgXT5rc8:-NvVnNl-1yebFzOKn8V_:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L246)):*

Instead of explaining followed by a "continue" question, how about:

> Adding the draft will overwrite the current chapter. Are you sure you want to continue?
> [ Cancel ] - or - [ Overwrite chapter ]
</blockquote></details>

Done.

nigel-wells (changes requested):
I have a state here - which may be bad data - but I have a book (which I didn't ask for a draft on) that has an apply button but there is no source available. I'm guessing this isn't related to this PR but wanted  to point it out just in case:


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-:-NvZpiwMAeJAVCAy1H2F:b-nae5dc)* status: all files reviewed, 6 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFBi7N5p5OkUMj9-Zw:-NvZch1YABEruI3deGIW:bshtq2) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L13)):*

Well, it kind of is. Maybe I'm not seeing what you're saying. To me, a DeltaStatic is a diff. It's what's output by DeltaStatic.diff(), which is what I'm supplying to these ops.
</blockquote></details>

DeltaStatic can help provide a diff but it isn't its only purpose. `DraftDiff` only contains a diff because those were the ops set - it could have any ops if it wanted to. Doesn't need to hold anything up.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3743 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFGhkTCvcBUjwjYot_:-NvZbx9e1je64KZLlAko:b-61qij9) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3743)):*

That would be difficult in this test. I haven't set up the editor's contents or the draft contents, because it's not required by or the point of this test. This test demonstrates simply that when the editor receives the draftApplied event, it takes the passed diff and applies it. It has no bearing on the contents of either text, as that was determined in the draft tab.
</blockquote></details>

Fair point - thanks.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss` line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFAbvj2xtK1zkUonrI:-NvZD8W2-waKykB7piIC:b-8riw0z) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss#L33)):*

I certainly can
</blockquote></details>

Thanks.
Can you change from `@import` to `@use`. It won't overly affect anything for this PR but it is good practice that we've started introducing to the repo.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 130 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFCUNp1uJHPAn_rXm6:-NvZZO_K1OzX1et5rAdv:b5j37kl) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L130)):*

That is interesting! I can put a check in here, but how did you get an empty draft? You'll have to be the one to test this change, haha
</blockquote></details>

Yep, that fixed it thanks!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 175 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFDEDl01r5yXeQz9_s:-NvZZo1S53JbC9dJmsgl:bjwdd6v) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L175)):*

Yep, thanks, Nathaniel!
</blockquote></details>

Fair point.
What about the white space? My question is in relation to showing a dialog to confirm overwriting content and I think that isn't needed if all the content is blanks.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 184 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFFcwM8-LzOHTWyiT1:-NvZcBqI3tR73gN6veQd:b9jye7) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L184)):*

It will go down to 3 when we remove the draft previewer :)
</blockquote></details>

Well, if it's 3 then go for it :)
Can we at least add the same comment on the methods in `editor.component.ts` as a reminder to update them here as well for now.

josephmyers (commented):
This is quite interesting, because I thought we had just covered in this a comment below. If you're saying (again?) that a blank draft chapter, however you got there, has the button, then the doesDeltaHaveContent method I just made to prevent that isn't fully working. Can you step into that method and see what's causing it to think that there's content? Whenever I have a chapter with no draft data, I get the blue "x has no draft" blurb. I would think the `op.insert.trim().length === 0` check would cover this.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-:-Nv_cSC73yP-QqINx4di:b-w5c4y3)* status: 7 of 10 files reviewed, 6 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts` line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFBi7N5p5OkUMj9-Zw:-Nv_XjZuBFePh8okS0Ln:brr9t9b) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-viewer/draft-viewer.service.ts#L13)):*

DeltaStatic can help provide a diff but it isn't its only purpose. `DraftDiff` only contains a diff because those were the ops set - it could have any ops if it wanted to. Doesn't need to hold anything up.
</blockquote></details>

Right, I kinda see what you're saying. Given that a set of ops could be (for example) a whole document, I wanted to specify that we should only be passing a diff, which importantly will then be used to update the contents of the target editor. I think my reasoning is still sound, but maybe my implementation wasn't. I thought that specifying that in the name itself would be sufficient to prevent misuse, but maybe we should be doing something else. I'm open to ideas!

(To be clear, I think naming it more generically is the wrong direction to go. Given that DeltaStatic can be more than a diff, we need to be all the more explicit in how we want it to be used here.)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss` line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFAbvj2xtK1zkUonrI:-Nv_cK7-5avPVD9TCWn1:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss#L33)):*

Thanks.
Can you change from `@import` to `@use`. It won't overly affect anything for this PR but it is good practice that we've started introducing to the repo.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 175 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFDEDl01r5yXeQz9_s:-Nv_efHcEm8kSLJqtoPo:bde6zwq) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L175)):*

Fair point.
What about the white space? My question is in relation to showing a dialog to confirm overwriting content and I think that isn't needed if all the content is blanks.
</blockquote></details>

Let's continue this discussion up where your screenshot is.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 184 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFFcwM8-LzOHTWyiT1:-Nv_g3zP7tfiq9eJIkF4:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L184)):*

Well, if it's 3 then go for it :)
Can we at least add the same comment on the methods in `editor.component.ts` as a reminder to update them here as well for now.
</blockquote></details>

Done.

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391)* status: 8 of 10 files reviewed, 6 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss` line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFAbvj2xtK1zkUonrI:-Nv_hc7P0vDC9IscTVM-:b9fv38x) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss#L33)):*

Done.
</blockquote></details>

Um this broke the build. Reverting until we figure out why.

nigel-wells (changes requested):
The previous comment was to fix "DRAFT APPLIED" from appearing when I hadn't applied anything - that is fixed. This instance I wasn't expecting to be able to get into and it may not be related to what you're doing. The source project (I think Mark created?) Only has Genesis Chapter 1 available. When `getGeneratedDraftDeltaOperations`is called it does return valid ops so I must be wrong and had requested a draft at some point. I guess this is still correct and the add button should be available even if it seems like an odd state?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-:-NveDXAQ8xuEFuaiUZNI:b-5ft9v3)* status: all files reviewed, 6 unresolved discussions (waiting on @josephmyers and @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss` line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFAbvj2xtK1zkUonrI:-NveGVen1tW54fW2CRaw:b-sffk6n) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss#L33)):*

Um this broke the build. Reverting until we figure out why.
</blockquote></details>

What exactly broke?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 161 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NveGN4X4VyCcqXVV79n:-NveGN4X4VyCcqXVV79o:b-tk7cb5) ([raw file](https://github.com/sillsdev/web-xforge/blob/af789d16cc0df367fa7a30d57cebc0f82338fe5d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L161)):*
> ```XML
>       if (!proceed) return;
>     } else {
>       return;
> ```

Did you mean to commit this as now nothing happens?

josephmyers (commented):
Yeah, that is odd, but if it's got content, then the button should be there. We can keep an eye on it

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-:-Nvexz_QD_lPkXT236YM:b-lh2019)* status: 9 of 10 files reviewed, 6 unresolved discussions (waiting on @Nateowami, @nigel-wells, and @use)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss` line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFAbvj2xtK1zkUonrI:-NveyAebBwrEbsRan8Ff:bb6s7nw) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss#L33)):*

What exactly broke?
</blockquote></details>

@use
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 161 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NveGN4X4VyCcqXVV79n:-NveyJkA5QIR4mfJVVvw:b-834n6v) ([raw file](https://github.com/sillsdev/web-xforge/blob/af789d16cc0df367fa7a30d57cebc0f82338fe5d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L161)):*

Did you mean to commit this as now nothing happens?
</blockquote></details>

Oops!

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391)* status: 9 of 10 files reviewed, 6 unresolved discussions (waiting on @Nateowami, @nigel-wells, and @use)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss` line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFAbvj2xtK1zkUonrI:-Nvez6A58L5sNfzxt7d8:b-nomyfp) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss#L33)):*

@use
</blockquote></details>

`@use` -- sorry, Will Hall?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391)* status: 9 of 10 files reviewed, 3 unresolved discussions (waiting on @nigel-wells and @use)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss` line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFAbvj2xtK1zkUonrI:-NvgjtBH1zFC2gOb7QTD:bi9e6fk) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss#L33)):*

`@use` -- sorry, Will Hall?
</blockquote></details>

`@use` is not identical to `@import`. You will need to make this change as well. Actually, this is part of why `@use` is preferable.
``` diff
-  color: $greenDark;
+  color: variables.$greenDark;
```

nigel-wells (changes requested):
Agreed.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-:-Nvhm_pw-uhazFxxOgNW:b-pbdqz3)* status: all files reviewed, 1 unresolved discussion (waiting on @use)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss` line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFAbvj2xtK1zkUonrI:-NvhnI_g3Oij1vUmHQcL:b-a036yy) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss#L33)):*

`@use` is not identical to `@import`. You will need to make this change as well. Actually, this is part of why `@use` is preferable.
``` diff
-  color: $greenDark;
+  color: variables.$greenDark;
```
</blockquote></details>

Sorry, I should have been more specific. Changing `@import 'src/variables';` to `@use 'src/variables';` is the first part. You then need to reference the namespace for the variable as Nathaniel has indicated.

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391)* status: 9 of 10 files reviewed, 1 unresolved discussion (waiting on @Nateowami, @nigel-wells, and @use)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss` line 33 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-NvFAbvj2xtK1zkUonrI:-NvieZm8AMRyw7uDd50G:b-nvr5un) ([raw file](https://github.com/sillsdev/web-xforge/blob/884247983ba124757258a715565a66dda1e4925d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.scss#L33)):*

Sorry, I should have been more specific. Changing `@import 'src/variables';` to `@use 'src/variables';` is the first part. You then need to reference the namespace for the variable as Nathaniel has indicated.
</blockquote></details>

Ah, good to know.

nigel-wells (approved):
:lgtm:

> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-:-NvimAj37kXZMdME84tJ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

nigel-wells (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami:
I am seeing a chapter become corrupted after applying a draft. I'm trying to work out what in particular makes this chapter get corrupted, but the build queue is full, so I can't re-generate a draft. I will try again tomorrow.

josephmyers:
> I am seeing a chapter become corrupted after applying a draft. I'm trying to work out what in particular makes this chapter get corrupted, but the build queue is full, so I can't re-generate a draft. I will try again tomorrow.

I saw this early on in development, when I still had legacy drafts. I haven't seen it with any new drafts. Test hasn't mentioned anything like this, though they did see an error in the History tab which I believe to be related to the new format, and potentially on master, as well. I'm not doing any special processing for the new style (or the old). Maybe that's causing an issue?

Nateowami:
@josephmyers I can consistently reproduce the problem. It appears it requires there to be two notes in the verse, and that's what triggers it.

josephmyers:
@Nateowami That is bizarre. Any idea as to the cause or fix? I won't have time this week to do a full research and implementation, so feel free to hijack this PR if time is critical.

josephmyers:
Yes, I'm seeing this, as well. My chapter looks _really_ messed up.

image

josephmyers:
So it sounds like @siltomato has the latest information on this and is the best one suited to finish this bug out. From the brief investigation I did yesterday, you just need to match what Peter's done in the latest Draft Viewer code. It doesn't have this issue.

siltomato (commented):
I made some changes to fix the draft detection issue and the corruption issue.

**To fix corruption**
- Added `TextDocService`
  - Has method to overwrite a `TextDoc` with the supplied delta.
    - This avoids going through the editor and having to handle 'view model' ops like note embeds.
    - Could possibly be used in the future for applying history revisions.
    - Since this is a local change, `TextViewModel` does not respond to update the editor with the changes.
    - So, a new event is emitted that `TextViewModel` can listen to.
  - Has method to get an observable that emits local 'system' changes for a `TextDocId` (emitted by the `overwrite()` method).
- `TextViewModel` is now `@Injectable()`, and is provided at the `TextComponent` level
  - Allows injection of `TextDocService` to listen for local 'system' changes emissions.

**To fix draft existence detection**
- Added request to legacy draft API endpoint to check for draft existence (until better way provided by serval).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-:-Nx-8Uq78lObRB5cgykb:b-2kv4kp)* status: 3 of 18 files reviewed, 1 unresolved discussion (waiting on @josephmyers, @Nateowami, and @nigel-wells)

Nateowami:
Excellent work @siltomato! This is working well in my testing.

@nigel-wells or @josephmyers can you review @siltomato's changes?

josephmyers (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-:-NxPi14B2vlRsn4oFDjf:bnfp4nl)* status: 3 of 18 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @nigel-wells)

nigel-wells (approved):
> Reviewed 11 of 15 files at r8, 3 of 3 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers and @Nateowami)

nigel-wells (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2391#-:-NxRkEU8-EHKsJ266vnV:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers and @Nateowami)

PR #2394, opened 2024-04-02
SF-2615 Assign Additional Sources for Training

pmachapman:
This PR adds support for mixing in additional source for drafting.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394)


josephmyers:
This is still needing a review, right? I'll do it, but can you rebase to latest first? Thanks

pmachapman:
> This is still needing a review, right? I'll do it, but can you rebase to latest first? Thanks

Great idea - done! Thank you!!!

josephmyers:
> If you think this isa bad idea, I can rename it all, but I thought I would submit this first version with `mixSources` for consideration. The `mixSources` code was all written before the name change.

"mixSources" is confusing. There's so much other data already being mixed in potentially! Let's do additionalTrainingSource, to match alternateTrainingSource. I'm also wondering, generally, how this differs from additionalTrainingData. The Settings items are just that much more confusing now. We have an option "Allow additional training data" and now "Use an additional source for training"

josephmyers (changes requested):
Thanks for bearing with my ignorance in some of the questions I ask. I've gotten through the code and will do some light testing now and into next week. Thanks for your work on this!

> Reviewed 1 of 14 files at r1, 26 of 26 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-:-NwIQloUA8wO-XNHfSCq:bpjcr4m)* status: all files reviewed, 7 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts` line 501 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NwE5V0_2IcVE8AUHb_9:-NwE5V0aAMPYCbLuyWe1:bp2hbo6) ([raw file](https://github.com/sillsdev/web-xforge/blob/d8dc2e758ab60ed4517b06b2ae351819cfec304e/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts#L501)):*
> ```XML
> 
>     // If mix sources is enabled, send all segments is disabled, and vice-versa
>     if (this.projectDoc?.data?.translateConfig?.draftConfig?.sendAllSegments) {
> ```

I realize there are other places here that are following this pattern, but isn't this an overly verbose way of handling two mutually exclusive values? Couldn't we set their disabled states in the HTML in fewer lines?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts` line 504 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NwE5sp43rAU1wb9W8cC:-NwE5sp43rAU1wb9W8cD:b-h42ho) ([raw file](https://github.com/sillsdev/web-xforge/blob/d8dc2e758ab60ed4517b06b2ae351819cfec304e/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts#L504)):*
> ```XML
>       this.mixSourcesEnabled.disable({ onlySelf: true });
>       this.sendAllSegments.enable({ onlySelf: true });
>     } else if (this.projectDoc?.data?.translateConfig?.draftConfig?.mixSourcesEnabled) {
> ```

Do you think we need some sort of explanation (on the page) for why they can't do both?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html` line 82 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NwHpx0O-YPEmw2GKgZN:-NwHpx0O-YPEmw2GKgZO:b-6x25ow) ([raw file](https://github.com/sillsdev/web-xforge/blob/d8dc2e758ab60ed4517b06b2ae351819cfec304e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html#L82)):*
> ```HTML
>       >
>         <transloco
>           key="draft_generation.info_alert_different_mix_and_source_language"
> ```

I don't think we need to do anything for your current task, but having warnings at this point is pretty inconvenient for the user. They have to interpret this message and know to navigate all the way back to Settings, where we can't see the message anymore, to hopefully remember what the problem was and then change things. They don't even know they've fixed it until they leave Settings and go back to auto-drafting.

We could do better (in future UI refactor tasks) if we showed them the warning when they make the error.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources.service.ts` line 87 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NwIGJDq0FBDcDNa2JbQ:-NwIGJDr1aUS_-IYVVyL:b-287tee) ([raw file](https://github.com/sillsdev/web-xforge/blob/d8dc2e758ab60ed4517b06b2ae351819cfec304e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources.service.ts#L87)):*
> ```XML
>         // If the user cannot access the a mix source project,
>         // populate using the target's mix source information, if enabled
>         let mixSourceProjectIds: string[] | undefined;
> ```

Hey, I'm a little confused as to what the software should be doing, based on this code, contrasted with the `info_alert_no_mix_source_access` warning. Are we allowing them to auto-draft if they don't have access to one of the sources, or not? The message seems to say they can't, but this code makes it seem like we're including it/them anyway.

I also don't see a unit test that explicitly mentions "no access" for this feature, though I didn't look much harder past the test names. Maybe I'm just not interpreting the names well enough?
___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 824 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NwIKdgx0zmg3Nu4ozxN:-NwIKdgx0zmg3Nu4ozxO:bozpn7w) ([raw file](https://github.com/sillsdev/web-xforge/blob/d8dc2e758ab60ed4517b06b2ae351819cfec304e/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L824)):*
> ```Smalltalk
>         }
> 
>         // If we have any mixed sources, sync those next, but only if we are uploading a Paratext project
> ```

What actually checks in the code below that we're uploading a PT project?

I also only see a unit test for one of the cases.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 1076 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NwIOpMV4VTKlidly4D9:-NwIOpMV4VTKlidly4DA:bwhdyua) ([raw file](https://github.com/sillsdev/web-xforge/blob/d8dc2e758ab60ed4517b06b2ae351819cfec304e/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L1076)):*
> ```Smalltalk
> 
>         // If there are mix in sources, ensure that their writing system and RTL is correct
>         if (projectDoc.Data.TranslateConfig.DraftConfig.MixSourcesEnabled)
> ```

Non-blocking question: Why is this necessary? Why aren't these two fields correct already? Wouldn't it be better/safer to update the respective field when the change happens or we first find out?

josephmyers:
My draft is completing suspiciously:

**Diagnostic Information**
Build Id: 6629d9d5501872989c1938b2
Corpora Ids: 66288212501872989c1933d3
Date Finished: 2024-04-25T04:27:22.537+00:00
Message: Completed
Percent Completed: 1
Revision: 9
Queue Depth: 0
State: COMPLETED
Step: 0
Translation Engine Id: 6628820f501872989c1933d0

But the draft content is all empty verses. I'm seeing this on other branches, too, but it's hard to actually test this while things are broken.

josephmyers:
So my auto-draft is still blank. But today I realized it might be because some books are blank for the source. My "mixin" source has all books present, but the resource I'm "based on" has the relevant books blank. I was initially thinking that would be a valid/common use case for mixing in sources, but maybe I'm wrong? I'll try another project setup.

pmachapman (commented):
Thank you for taking the time to review this. The Serval environment has changed (for the better) since I wrote this PR, so I wonder if some serious redesign is required (see the answers to the questions below and further down).

I am thinking I may need to make SF-2707 a pre-requisite for this PR.

Also, in answer to your questions:

> So my auto-draft is still blank. But today I realized it might be because some books are blank for the source. My "mixin" source has all books filled out, but the resource I'm "based on" has the relevant books blank. I was initially thinking that would be a valid/common use case for mixing in sources, but maybe I'm wrong? I'll try another project setup.

Mixing of sources involves Serval splitting them 50/50 (or 33/33/33 if more than one, etc). It looks like Serval chose the blank books or chapters. My understanding is blank content should not be used for training.

> "mixSources" is confusing. There's so much other data already being mixed in potentially! Let's do additionalTrainingSource, to match alternateTrainingSource. I'm also wondering, generally, how this differs from additionalTrainingData. The Settings items are just that much more confusing now. We have an option "Allow additional training data" and now "Use an additional source for training"

> Edit: I know what additionalTrainingData is, but its option is much too similar to the one you're adding. How are users supposed to be able to distinguish between "Allow additional training data," "Use an additional source for training," and "Use a different source for training"?

I wonder if we need to implement the proposed redesign from Nathaniel before embarking on this PR? https://balsamiq.cloud/sghq53/pejrlz8/r232F

If we were to do that, there would be no mixSources or alternateTrainingSource, just trainingSources.

> Dismissed @github-advanced-security[bot] from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-:-Nwg7ln61DM2P2mEZ6Kf:b-xh21cj)* status: all files reviewed, 5 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts` line 501 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NwE5V0_2IcVE8AUHb_9:-Nwg933S79S6tF2peInz:b-tqzvsh) ([raw file](https://github.com/sillsdev/web-xforge/blob/d8dc2e758ab60ed4517b06b2ae351819cfec304e/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts#L501)):*

I realize there are other places here that are following this pattern, but isn't this an overly verbose way of handling two mutually exclusive values? Couldn't we set their disabled states in the HTML in fewer lines?
</blockquote></details>

Serval has just added support for translating all segments with PT zip files, so we can remove this logic once SF-2707 is done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts` line 504 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NwE5sp43rAU1wb9W8cC:-Nwg8WhZ6SyQ4NLt8xdw:b-am03r) ([raw file](https://github.com/sillsdev/web-xforge/blob/d8dc2e758ab60ed4517b06b2ae351819cfec304e/src/SIL.XForge.Scripture/ClientApp/src/app/settings/settings.component.ts#L504)):*

Do you think we need some sort of explanation (on the page) for why they can't do both?
</blockquote></details>

See reply above.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html` line 82 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NwHpx0O-YPEmw2GKgZN:-Nwg9GPA7IgqzwIHugyI:b-hlq69n) ([raw file](https://github.com/sillsdev/web-xforge/blob/d8dc2e758ab60ed4517b06b2ae351819cfec304e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html#L82)):*

I don't think we need to do anything for your current task, but having warnings at this point is pretty inconvenient for the user. They have to interpret this message and know to navigate all the way back to Settings, where we can't see the message anymore, to hopefully remember what the problem was and then change things. They don't even know they've fixed it until they leave Settings and go back to auto-drafting.

We could do better (in future UI refactor tasks) if we showed them the warning when they make the error.
</blockquote></details>

Yes, I agree. We need to do better.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources.service.ts` line 87 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NwIGJDq0FBDcDNa2JbQ:-Nwg9s4b3-3VEO36foui:bj0p82c) ([raw file](https://github.com/sillsdev/web-xforge/blob/d8dc2e758ab60ed4517b06b2ae351819cfec304e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources.service.ts#L87)):*

Hey, I'm a little confused as to what the software should be doing, based on this code, contrasted with the `info_alert_no_mix_source_access` warning. Are we allowing them to auto-draft if they don't have access to one of the sources, or not? The message seems to say they can't, but this code makes it seem like we're including it/them anyway.

I also don't see a unit test that explicitly mentions "no access" for this feature, though I didn't look much harder past the test names. Maybe I'm just not interpreting the names well enough?
</blockquote></details>

We get them here so the frontend can display that there is no access to them. If they were absent from the returned object the frontend would be oblivious to whether they have access or not.
___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 824 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NwIKdgx0zmg3Nu4ozxN:-NwgA_dT-jpfyygXzdxo:barickl) ([raw file](https://github.com/sillsdev/web-xforge/blob/d8dc2e758ab60ed4517b06b2ae351819cfec304e/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L824)):*

What actually checks in the code below that we're uploading a PT project?

I also only see a unit test for one of the cases.
</blockquote></details>

I named this comment wrong. Also it will be redundant after SF-2707
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 1076 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NwIOpMV4VTKlidly4D9:-NwgB-nu3ChKF6wYWLGk:b-u8jqcg) ([raw file](https://github.com/sillsdev/web-xforge/blob/d8dc2e758ab60ed4517b06b2ae351819cfec304e/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L1076)):*

Non-blocking question: Why is this necessary? Why aren't these two fields correct already? Wouldn't it be better/safer to update the respective field when the change happens or we first find out?
</blockquote></details>

If the mix source was added to the project and it was not already in Scripture Forge, these values will be incorrect. There are also older projects with these values not set properly, which this will fix.

josephmyers (changes requested):
I'm good with whatever decision you and potentially Nathaniel want to make. I do see below that many, many of my comments will be invalidated or superseded by SF-2707, so it would make sense to me.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-:-NwgX_yp1GUhumnHPq5J:b-4bic6g)* status: all files reviewed, 4 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources.service.ts` line 87 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NwIGJDq0FBDcDNa2JbQ:-NwgYE1V-uyKh3gBHsUp:b-dtbv65) ([raw file](https://github.com/sillsdev/web-xforge/blob/d8dc2e758ab60ed4517b06b2ae351819cfec304e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources.service.ts#L87)):*

We get them here so the frontend can display that there is no access to them. If they were absent from the returned object the frontend would be oblivious to whether they have access or not.
</blockquote></details>

Thanks, I was wondering if that was the case. Do we need a unit test for this?

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394)* status: all files reviewed, 4 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources.service.ts` line 87 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NwIGJDq0FBDcDNa2JbQ:-NwgfzN_D97EhYKBCBfr:bugxdwk) ([raw file](https://github.com/sillsdev/web-xforge/blob/d8dc2e758ab60ed4517b06b2ae351819cfec304e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources.service.ts#L87)):*

Thanks, I was wondering if that was the case. Do we need a unit test for this?
</blockquote></details>

Does 'should pass the values from the target project if no access' cover this, or am I misunderstanding your question?

josephmyers (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394)* status: 22 of 27 files reviewed, 4 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources.service.ts` line 87 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NwIGJDq0FBDcDNa2JbQ:-Nwgp6Xg7SEL1D-7CXjJ:b4qg7ij) ([raw file](https://github.com/sillsdev/web-xforge/blob/d8dc2e758ab60ed4517b06b2ae351819cfec304e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources.service.ts#L87)):*

Does 'should pass the values from the target project if no access' cover this, or am I misunderstanding your question?
</blockquote></details>

Yes, I see that now, this works

josephmyers:
This was mentioned as a request yesterday at EMDC. Are we any closer to getting it through?

pmachapman:
> This was mentioned as a request yesterday at EMDC. Are we any closer to getting it through?

> "mixSources" is confusing. There's so much other data already being mixed in potentially! Let's do additionalTrainingSource, to match alternateTrainingSource. I'm also wondering, generally, how this differs from additionalTrainingData. The Settings items are just that much more confusing now. We have an option "Allow additional training data" and now "Use an additional source for training"

@josephmyers I have made a refactor of the code to use `additionalTrainingSource` instead of `mixSources`. Also, this is now a single object, not an array. I initially tried to combine the alternate training source and additional training source into a collection (called `trainingSources`), but given the current UI, it was impossible to do properly (and keep track of the values), and I had to start again.

The currently plan is to treat this as an interim (but production ready) implementation of "mix sources" feature. When we create the new UI (as designed at https://balsamiq.cloud/sghq53/pejrlz8/r232F), we can combine `additionalTrainingSource` and `alternateTrainingSource` into `trainingSources`, and remove the `additionalTrainingSource`, `alternateTrainingSource`, `additionalTrainingSourceEnabled`, and `alternateTrainingSourceEnabled` properties.

Implementing the new design is a bit more involved that this PR (and I may not be the most proficient to implement the frontend UI changes required - the backend changes will be straightforward), so I think it should be done after this PR is tested/merged.

josephmyers:
As I'm starting on this review again, I recommend we hide/disable-with-comment this option if either the source or target is a back translation, given it still doesn't work if either is the case.

josephmyers (changes requested):
Looks good for this task. Just a tiny comment below, and the one I made above. I have bigger UX concerns with this workflow, but I think they're solvable when we implement the new UI.

> Reviewed 27 of 28 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-:-Nzb45MtAjSaI_f0ezkI:b-sy284i)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.ts` line 397 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NzauoXV1fdPJqY7oFjR:-NzauoXWEBUYSAbXUVBC:b-62vzwz) ([raw file](https://github.com/sillsdev/web-xforge/blob/661f4132cc8680cc8181a3933dc640a3544096e3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.ts#L397)):*
> ```XML
>    * @returns true if the user has access to the source, or if there is no source.
>    */
>   canAccessDraftSourceIfAvailable(source: DraftSource | undefined): boolean {
> ```

Update comment

josephmyers (commented):
We should improve the warning message for when sources are different languages. You could easily get into a situation where it appears that they're the same (from their titles), and the user doesn't know how to fix the error without opening Paratext. Warning messages should always be solution-focused, or at least give enough information that the solution is obvious.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-:-Nzb7Dt21wOgFy_g65wD:b-jenk9d)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

pmachapman (commented):
> We should improve the warning message for when sources are different languages.

Let me know if the changes are an improvement, and if it is correct that I cleared the translations for the updated notices?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-:-NzdyeD6EZhGd6OBq3H3:b-xmzpgu)* status: 20 of 41 files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.ts` line 397 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-NzauoXV1fdPJqY7oFjR:-NzddPzj2IwAqL12wVZU:b-ioxi27) ([raw file](https://github.com/sillsdev/web-xforge/blob/661f4132cc8680cc8181a3933dc640a3544096e3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.ts#L397)):*

Update comment
</blockquote></details>

Done. Thank you!

josephmyers (changes requested):
Thanks, this is looking better. I do see one small error. Are you populating the source language when there's an alternate source selected? I had NIV11UK as an alternate training source, NIV11 as an additional training source, and NIV84 as my source:


This does bring up a larger issue, one that we may not be able to fix. It's really annoying, to me, when I can't use a resource because of its setting in Paratext, when it's basically the same language. For example, I can't use `en-US` with `English`. This is what's happening in the scenario above, and it's pretty silly. Is there no way we can get Serval to accept cases like this?

> Reviewed 21 of 21 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-:-NzfoY5c73Ozkt18A_XL:b-ak6tfv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (commented):
> Are you populating the source language when there's an alternate source selected?

Good bug spotting - I have fixed this.

> or example, I can't use `en-US` with `English`. This is what's happening in the scenario above, and it's pretty silly. Is there no way we can get Serval to accept cases like this?

Serval isn't blocking it - we could not have this check, and upload without issue (for this case). I think a problem comes when we have languages with the same ISO-639 language code, but a different ISO 15924 script code. For example, arb_Arab and arb_Latn. I think we might also have a similar issue with Chinese Simplified and Traditional (although I bypassed the block in code and was able to generate a draft - I an unsure if it was sensible though)? I am not sure whether regional variations in language (i.e. pt and pt-BR) would cause draft issues either.

We could perhaps show the warning, and allow the user to proceed (and ignore the warning if they wish)?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-:-NzysTKRB23jUM_v0MvD:b-ou0qbl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

josephmyers (approved):
I think that'd work, good idea. We'd need to doctor the message, of course, to warn them that they may get some wild output.

> Reviewed 6 of 6 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-:-NzzqjaT87hZEHXjCSHJ:b-na0g0u)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

josephmyers (changes requested):
I didn't mean to approve this

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-:-NzzrUuS6TQk-t1TjLFF:b-symrkt)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (commented):
> I think that'd work, good idea. We'd need to doctor the message, of course, to warn them that they may get some wild output.

OK, let me know if this was along the lines of what you were thinking:


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-:-O-DskWiAUwvRUKAASNs:b2dd79o)* status: 33 of 41 files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (approved):
This looks solid. Thanks for switching this up

> Reviewed 8 of 8 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-:-O-DzyPu7ng6f5-oTOSA:baon8xd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (commented):
@josephmyers Just a heads up - I demoed this screen to Bethany and JoEllen, and the consensus was that the warning should be blocking, so I have reverted the change that made it unblocking. Also, some tweaks were made to the wording based on feedback given.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2394#-:-O-cCS-s07RoJQm_9Loj:bfweyvq)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

josephmyers:
> @josephmyers Just a heads up - I demoed this screen to Bethany and JoEllen, and the consensus was that the warning should be blocking, so I have reverted the change that made it unblocking. Also, some tweaks were made to the wording based on feedback given.

Interesting! Well, I'm glad somebody closer to the users got eyes on it.

PR #2405, opened 2024-04-10
Add Mongo query for errors that caused projects to fail last sync

Nateowami:
This query is an attempt to answer the question "What errors have caused the most projects to fail their most recent sync?" It does this by looking up the most recent sync failure for each project that has failed its last sync, then finding the most recent sync metrics for that project. Then it groups by error message, which somewhat works to find common errors, though sometimes the same error shows up with a different stack trace and doesn't get grouped properly. This is considered acceptable because we're just looking for most common errors.

This query is inefficient and slow to run.

The current top result, with a count of 17, is this:
```
System.Net.Http.HttpRequestException: HTTP Request error:
    POST https://registry.paratext.org/api8/token
    Response:
    StatusCode: 429, ReasonPhrase: 'Too Many Requests',
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2405)

### DIFF
--- a/mongodb/SyncMetrics/FindCurrentSyncErrors.mongodb
+++ b/mongodb/SyncMetrics/FindCurrentSyncErrors.mongodb
@@ -1,0 +1,66 @@
+use("xforge");
+
+// Find errors that have caused projects to fail their most recent sync
+
+db.sf_projects.aggregate([
+  {
+    $match: {
+      "sync.lastSyncSuccessful": false
+    }
+  },
+  {
+    $project: {
+      _id: 1,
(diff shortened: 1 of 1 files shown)


pmachapman (approved):
:lgtm: The query works great.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2405#-:-Nv8XwycBRoDxHhSW1nA:b93pdn9)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2406, opened 2024-04-10
SF-2636b Fallback from Awami to Scheherazade when Graphite unsupported

Nateowami:
A Graphite font, when used in a browser engine that doesn't support it, renders completely wrong. It turns out that for Urdu, using the Arabic script is significantly better, even though it's less readable than the Awami Nastaliq font. I've added it as a fallback in browsers that don't support Graphite.

I've added a story that shows all the fonts we support. This change can be tested by opening the story in both Chrome and Firefox, and using the dev tools to check that the proper font was used in each browser.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2406)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts
@@ -1,10 +1,13 @@
 import { Inject, Injectable } from '@angular/core';
 import { SFProjectProfileDoc } from 'src/app/core/models/sf-project-profile-doc';
 import { DOCUMENT } from './browser-globals';
+import { isGecko } from './utils';
 
 const DEFAULT_SERIF_FONT_FAMILY = 'Charis SIL';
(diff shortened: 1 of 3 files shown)


pmachapman (approved):
:lgtm:

Normally we want to detect the features, not the browser engine, but it seems for graphite there is no real choice. You can fallback from a graphite font to another using CSS, but I think that is too complex to implement for our needs: https://graphite.sil.org/graphite_firefox.html

Also, it looks like Chrome will never support graphite: https://issues.chromium.org/issues/40883477

**Results of my testing**

Awami Nastaliq on Firefox:


Awami Nastaliq on Chrome without the fallback:


Awami Nastaliq on Chrome with the fallback to Scheherazade New:


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2406#-:-Nv8_a5f-ILh0G8eLSPM:besebwa)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2412, opened 2024-04-15
SF-2580 Add API backend for chapter selection UI

pmachapman:
This PR enables support for specifying chapters for drafting or training when generating a draft. This is done in a backwards compatible way with the current "specify book only" method.

See https://github.com/sillsdev/serval/wiki/Filtering-Paratext-Project-Data-with-a-Scripture-Range for the syntax used.

When the frontend implements the corresponding change, it will need to generate the string using that syntax, and feed it through the updated `BuildConfig` object to the backend.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2412)

### DIFF
--- a/src/RealtimeServer/scriptureforge/models/translate-config.ts
+++ b/src/RealtimeServer/scriptureforge/models/translate-config.ts
@@ -34,7 +34,9 @@ export interface DraftConfig {
   alternateTrainingSource?: TranslateSource;
   lastSelectedTrainingBooks: number[];
   lastSelectedTrainingDataFiles: string[];
+  lastSelectedTrainingScriptureRange?: string;
   lastSelectedTranslationBooks: number[];
+  lastSelectedTranslationScriptureRange?: string;
   sendAllSegments: boolean;
   servalConfig?: string;
 }
--- a/src/RealtimeServer/scriptureforge/services/sf-project-service.ts
+++ b/src/RealtimeServer/scriptureforge/services/sf-project-service.ts
@@ -201,11 +201,17 @@ export class SFProjectService extends ProjectService<SFProject> {
                   bsonType: 'string'
                 }
               },
+              lastSelectedTrainingScriptureRange: {
+                bsonType: 'string'
+              },
               lastSelectedTranslationBooks: {
                 bsonType: 'array',
                 items: {
                   bsonType: 'int'
                 }
+              },
+              lastSelectedTranslationScriptureRange: {
+                bsonType: 'string'
               },
               sendAllSegments: {
                 bsonType: 'bool'
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.ts
@@ -25,7 +25,9 @@ import { TrainingDataService } from '../training-data/training-data.service';
 export interface DraftGenerationStepsResult {
   trainingBooks: number[];
   trainingDataFiles: string[];
+  trainingScriptureRange?: string;
   translationBooks: number[];
+  translationScriptureRange?: string;
   fastTraining: boolean;
 }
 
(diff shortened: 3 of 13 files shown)


pmachapman:
This is on hold because of https://github.com/sillsdev/serval/issues/364

pmachapman:
Non-blocking bug found in dev testing: https://github.com/sillsdev/serval/issues/365

josephmyers (changes requested):
Good job -- one small thing!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2412#-:-NwhczbA2d3XAeimxHsY:b6lgnlx)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Models/BuildConfig.cs` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2412#-Nwhbf1X-A75HPpaZXl4:-Nwhbf1X-A75HPpaZXl5:b-ozhm7s) ([raw file](https://github.com/sillsdev/web-xforge/blob/c179fcb5d4eb20ecce515f7e2bc70208ed33645a/src/SIL.XForge.Scripture/Models/BuildConfig.cs#L23)):*
> ```Smalltalk
>     /// <remarks>
>     /// You should not set this property and <see cref="TrainingScriptureRange"/> at the same time.
>     /// </remarks>
> ```

This is a code smell to me, but it's not really our fault. Is the plan for them to deprecate this property on their end?
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 1192 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2412#-Nwhc0PoCPK1C42tHl92:-Nwhc0PoCPK1C42tHl93:b-i8mdfe) ([raw file](https://github.com/sillsdev/web-xforge/blob/c179fcb5d4eb20ecce515f7e2bc70208ed33645a/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L1192)):*
> ```Smalltalk
>                 )
>                     ? buildConfig.TranslationScriptureRange
>                     : string.Join(';', buildConfig.TranslationBooks.Select(Canon.BookNumberToId));
> ```

I think it'd be valuable to have an explicit check that both aren't populated, as mentioned in the API, though having both would almost certainly be a bug on our side. The motive with this is to fail fast.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2412)* status: 9 of 13 files reviewed, 2 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/Models/BuildConfig.cs` line 23 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2412#-Nwhbf1X-A75HPpaZXl4:-NwlohsY6lQ3Vexu0S0c:bucxmc) ([raw file](https://github.com/sillsdev/web-xforge/blob/c179fcb5d4eb20ecce515f7e2bc70208ed33645a/src/SIL.XForge.Scripture/Models/BuildConfig.cs#L23)):*

This is a code smell to me, but it's not really our fault. Is the plan for them to deprecate this property on their end?
</blockquote></details>

Yes. TrainingBooks and TranslationBooks will be removed after the frontend has bedded in. There will likely be a transition period while we wait for clients to update to the latest version of Scripture Forge.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 1192 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2412#-Nwhc0PoCPK1C42tHl92:-NwlpIvJ4akXnhohzofY:b-v1jp4y) ([raw file](https://github.com/sillsdev/web-xforge/blob/c179fcb5d4eb20ecce515f7e2bc70208ed33645a/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L1192)):*

I think it'd be valuable to have an explicit check that both aren't populated, as mentioned in the API, though having both would almost certainly be a bug on our side. The motive with this is to fail fast.
</blockquote></details>

Done. I have implemented this in `MachineApiService` as that was the earliest in the execution chain after the Web API endpoint.

josephmyers (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2412#-:-NwrRgJb5tgWguVoBios:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2414, opened 2024-04-17
SF-2673 Gracefully handle Serval outages in the editor

pmachapman:
This PR updates our use of Serval to handle more gracefully the following scenarios:

 * Serval being offline (but Scripture Forge online)
 * Scripture Forge being offline
 * The user's connection being offline

This is achieved by limiting querying of the machine-api to projects that have pre-translation drafting enabled, and by handling most errors thrown by the machine-api when communicating with the draft generation service. When an error occurs with machine-api, a small red notice is displayed at the bottom of the screen:

image

Also, the draft generate service's tests have been rewritten to mock the HTTP requests correctly, as this was required for the online status service to function correctly in tests.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2414)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.spec.ts
@@ -146,6 +146,7 @@ describe('DraftGenerationComponent', () => {
             draftConfig: {
               alternateTrainingSourceEnabled: false
             },
+            preTranslate: true,
             projectType: ProjectType.BackTranslation,
             source: {
               projectRef: 'testSourceProjectId',
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.ts
@@ -242,11 +242,15 @@ export class DraftGenerationComponent extends SubscriptionDisposable implements 
     );
 
     this.subscribe(
-      this.activatedProject.projectId$.pipe(
+      this.activatedProject.projectDoc$.pipe(
         filterNullish(),
-        switchMap(projectId =>
-          this.draftGenerationService.getLastCompletedBuild(projectId).pipe(map(build => !isEmpty(build)))
-        )
+        switchMap(projectDoc => {
+          // Pre-translation must be enabled for the project
+          if (!(projectDoc.data?.translateConfig.preTranslate ?? false)) {
+            return of(false);
+          }
+          return this.draftGenerationService.getLastCompletedBuild(projectDoc.id).pipe(map(build => !isEmpty(build)));
+        })
       ),
       (hasAnyCompletedBuild: boolean) => {
         this.hasAnyCompletedBuild = hasAnyCompletedBuild;
@@ -415,17 +419,21 @@ export class DraftGenerationComponent extends SubscriptionDisposable implements 
   private pollBuild(): void {
     this.jobSubscription?.unsubscribe();
     this.jobSubscription = this.subscribe(
-      this.activatedProject.projectId$.pipe(
+      this.activatedProject.projectDoc$.pipe(
         filterNullish(),
-        switchMap(projectId =>
-          this.draftGenerationService
-            .getBuildProgress(projectId)
+        switchMap(projectDoc => {
+          // Pre-translation must be enabled for the project
+          if (!(projectDoc.data?.translateConfig.preTranslate ?? false)) {
+            return of(undefined);
+          }
+          return this.draftGenerationService
+            .getBuildProgress(projectDoc.id)
             .pipe(
               switchMap((job: BuildDto | undefined) =>
-                this.isDraftInProgress(job) ? this.draftGenerationService.pollBuildProgress(projectId) : of(job)
+                this.isDraftInProgress(job) ? this.draftGenerationService.pollBuildProgress(projectDoc.id) : of(job)
               )
-            )
-        )
+            );
+        })
       ),
       (job?: BuildDto) => {
         this.draftJob = job;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.service.spec.ts
@@ -1,21 +1,52 @@
 import { HttpErrorResponse, HttpStatusCode } from '@angular/common/http';
 import { HttpClientTestingModule, HttpTestingController } from '@angular/common/http/testing';
-import { TestBed } from '@angular/core/testing';
+import { fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { Canon } from '@sillsdev/scripture';
-import { of, throwError } from 'rxjs';
+import { of } from 'rxjs';
 import { first } from 'rxjs/operators';
+import { mock } from 'ts-mockito';
+import { NoticeService } from 'xforge-common/notice.service';
+import { OnlineStatusService } from 'xforge-common/online-status.service';
+import { TestOnlineStatusModule } from 'xforge-common/test-online-status.module';
+import { TestOnlineStatusService } from 'xforge-common/test-online-status.service';
+import { configureTestingModule, TestTranslocoModule } from 'xforge-common/test-utils';
+import { UICommonModule } from 'xforge-common/ui-common.module';
 import { BuildDto } from '../../machine-api/build-dto';
 import { BuildStates } from '../../machine-api/build-states';
-import { HttpClient } from '../../machine-api/http-client';
+import { MACHINE_API_BASE_URL } from '../../machine-api/http-client';
 import { BuildConfig } from './draft-generation';
 import { DraftGenerationService } from './draft-generation.service';
 
 describe('DraftGenerationService', () => {
   let service: DraftGenerationService;
-  let httpClient: HttpClient;
   let httpTestingController: HttpTestingController;
+  let mockNoticeService: NoticeService;
+  let testOnlineStatusService: TestOnlineStatusService;
+
+  configureTestingModule(() => ({
+    imports: [
+      HttpClientTestingModule,
+      NoopAnimationsModule,
+      TestOnlineStatusModule.forRoot(),
+      TestTranslocoModule,
+      UICommonModule
+    ],
+    providers: [
+      DraftGenerationService,
+      { provide: OnlineStatusService, useClass: TestOnlineStatusService },
+      { provide: NoticeService, useMock: mockNoticeService }
+    ]
+  }));
 
   const projectId = 'testProjectId';
+  const buildConfig: BuildConfig = {
+    projectId,
+    trainingBooks: [],
+    trainingDataFiles: [],
+    translationBooks: [],
+    fastTraining: false
+  };
   const buildDto: BuildDto = {
     id: 'testId',
     href: 'testHref',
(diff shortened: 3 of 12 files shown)


Nateowami (commented):
I have some concern with how the message is presented to the user.

1. "Drafting" has more than one meaning, and is traditionally a process done by translators. I think it would be more useful to have a message like "Computer generated drafts couldn't be fetched at the moment" or "Generating drafts is temporarily unavailable."
2. The placement of the message in a snackbar at the bottom of the page seems less helpful than putting it in context with the individual components that aren't available.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-:-Nvj7dmx5VC9ixKg27ot:b-tcwghd)* status: 0 of 12 files reviewed, all discussions resolved

pmachapman (commented):
Thank for raising your concerns, Nathaniel.

> "Drafting" has more than one meaning, and is traditionally a process done by translators. I think it would be more useful to have a message like "Computer generated drafts couldn't be fetched at the moment" or "Generating drafts is temporarily unavailable."

I have updated the message to "Generating drafts is temporarily unavailable."

We really need to agree cross teams on the nomenclature. It is being called "drafting" to the users, and in SF we have "generate draft", "auto draft" and "Generate translation drafts".

> The placement of the message in a snackbar at the bottom of the page seems less helpful than putting it in context with the individual components that aren't available.

I umm'ed and ahh'ed for a bit before choosing the snackbar. Two factors made me choose the snackbar:

1. Bubbling the error to the frontend, instead of managing in the service, would have resulted in a lot of duplicate error handling code in the different components that use the service.
2. Most often this is used for the display of toolbar icons or add tab icons, so there wasn't really the place there to display an inline message.

Ideally, we would have a more modern error notification area, like an icon in the top right to collect all of these types of messages?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-:-Nwg0xsGCvsbi6KUjqlP:bu4ty1b)* status: 0 of 12 files reviewed, all discussions resolved

marksvc (changes requested):
> Reviewed 6 of 12 files at r1, 6 of 6 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2414)* status: all files reviewed, 3 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.ts` line 436 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-NwlCU6d7ckwHIVv3M-p:-NwlCU6d7ckwHIVv3M-q:b-g33l3t) ([raw file](https://github.com/sillsdev/web-xforge/blob/08a244e18b8a3e3fee6d6866e1380e575e29fba6/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.ts#L436)):*
> ```XML
>         })
>       ),
>       (job?: BuildDto) => {
> ```

(I wondered about if it is okay when returning `of(undefined)` to then proceed and set `this.isDraftJobFetched = true`. I see that `draftGenerationService.getBuildProgress` can return undefined when offline or "if no build has ever been started", so I expect it to be okay.)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.service.spec.ts` line 105 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-Nwou2yC63I4lzpgqfyx:-Nwou2yC63I4lzpgqfyy:bj75uam) ([raw file](https://github.com/sillsdev/web-xforge/blob/08a244e18b8a3e3fee6d6866e1380e575e29fba6/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.service.spec.ts#L105)):*
> ```XML
>         .pipe(first())
>         .subscribe(result => {
>           expect(result).toBeUndefined();
> ```

Regarding this pattern in the test file, don't we not know for sure that the expect inside the subscribe gets called before the test finishes? Don't we need to do something else to ensure this expect is called and that we know it?

I see that in the previous version of the code, it was calling `done()` after the expect, for example.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts` line 128 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-Nwl4CX8DDO3VLXYcqHS:-Nwl4CX8DDO3VLXYcqHT:bdonrba) ([raw file](https://github.com/sillsdev/web-xforge/blob/08a244e18b8a3e3fee6d6866e1380e575e29fba6/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts#L128)):*
> ```XML
>       try {
>         this.projectDoc = await this.projectService.getProfile(projectId);
>         if (this.isOnline) {
> ```

Would there be significance for this to happen when the user _comes online_? For example, we could write something like
```ts
await this.onlineStatusService.online();
this.setupTranslationEngine();
```
Though that would cause some other code here to not run right away at page load when offline, so maybe instead something like
```ts
this.onlineStatusService.online().then( ()=> this.setupTranslationEngine() );
```
?

Oh, from an initial look it looks like maybe `this.updateEngineStats()`(a couple lines below)  might be significant to do _after_ `setupTranslationEngine()`, so maybe
```ts
this.onlineStatusService.online().then( async ()=> {
  this.setupTranslationEngine();
  await this.updateEngineStats();
} );
```
?

If it's significant to be online for running `setupTranslationEngine` here, should it also be done where `setupTranslationEngine` is called a dozen lines below? I know it says _remote_ changes$, but I'm guessing that might not always mean via Internet. (Maybe another window? Maybe it will always be via Internet.)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.spec.ts` line 361 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-Nwl8Y5R6tReVw6Toj2f:-Nwl8Y5R6tReVw6Toj2g:b5huwei) ([raw file](https://github.com/sillsdev/web-xforge/blob/08a244e18b8a3e3fee6d6866e1380e575e29fba6/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.spec.ts#L361)):*
> ```XML
>     this.testOnlineStatusService.setIsOnline(value);
>     this.fixture.detectChanges();
>     tick();
> ```

Was there significance in having `tick()` below `detectChanges()`? Because test-online-status.service.ts setIsOnline() calls `window.dispatchEvent()` (which I assume benefits from tick or flush), and the UI needs to respond to differences in mocked services and perhaps this event, it may be necessary to use `detectChanges` after `tick` to adequately simulate the various things that might happen on the page as a result of changing online status.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2414)* status: all files reviewed, 3 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.ts` line 436 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-NwlCU6d7ckwHIVv3M-p:-Nwqd6Dq3LiFhDI8o0Ly:biye8wf) ([raw file](https://github.com/sillsdev/web-xforge/blob/08a244e18b8a3e3fee6d6866e1380e575e29fba6/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.ts#L436)):*

(I wondered about if it is okay when returning `of(undefined)` to then proceed and set `this.isDraftJobFetched = true`. I see that `draftGenerationService.getBuildProgress` can return undefined when offline or "if no build has ever been started", so I expect it to be okay.)
</blockquote></details>

Yes - I tested a few scenarios with disabling pre-translate (and triggering `return of(undefined);` and the UI displayed the correct buttons, etc,i for the scenarios. One of those scenarios I tested was back translations, which will not have `preTranslate` set until their first build, but can still have drafts generated without the project being approved for AI drafting.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.service.spec.ts` line 105 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-Nwou2yC63I4lzpgqfyx:-Nwqe3566qjUK3j01dQR:been9oz) ([raw file](https://github.com/sillsdev/web-xforge/blob/08a244e18b8a3e3fee6d6866e1380e575e29fba6/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.service.spec.ts#L105)):*

Regarding this pattern in the test file, don't we not know for sure that the expect inside the subscribe gets called before the test finishes? Don't we need to do something else to ensure this expect is called and that we know it?

I see that in the previous version of the code, it was calling `done()` after the expect, for example.
</blockquote></details>

`done()` is not needed because `tick()` handles running the asynchronous code, and ensures that the `expect()` in the `subscribe()` is called.

If it wasn't called, the test would log a warning: `Spec 'DraftGenerationService getBuildProgress should return undefined if offline' has no expectations.`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts` line 128 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-Nwl4CX8DDO3VLXYcqHS:-Nwqkesk5uNviEdn0umA:b-aj68ol) ([raw file](https://github.com/sillsdev/web-xforge/blob/08a244e18b8a3e3fee6d6866e1380e575e29fba6/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts#L128)):*

Would there be significance for this to happen when the user _comes online_? For example, we could write something like
```ts
await this.onlineStatusService.online();
this.setupTranslationEngine();
```
Though that would cause some other code here to not run right away at page load when offline, so maybe instead something like
```ts
this.onlineStatusService.online().then( ()=> this.setupTranslationEngine() );
```
?

Oh, from an initial look it looks like maybe `this.updateEngineStats()`(a couple lines below)  might be significant to do _after_ `setupTranslationEngine()`, so maybe
```ts
this.onlineStatusService.online().then( async ()=> {
  this.setupTranslationEngine();
  await this.updateEngineStats();
} );
```
?

If it's significant to be online for running `setupTranslationEngine` here, should it also be done where `setupTranslationEngine` is called a dozen lines below? I know it says _remote_ changes$, but I'm guessing that might not always mean via Internet. (Maybe another window? Maybe it will always be via Internet.)
</blockquote></details>

Hmm you are right - this is a bug. I have reworked the logic to run when the user comes online, but also only run offline-ready methods (i.e. `calculateProgress()`) if offline.

`remoteChanges$` is always changes from the server (i.e. over the Internet). That said, there may be no changes on the server when the user comes online, so I have used the `online.then()` approach you outlined above as well.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.spec.ts` line 361 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-Nwl8Y5R6tReVw6Toj2f:-NwqhJbAE1z8R2Q6KNQI:bmckdu2) ([raw file](https://github.com/sillsdev/web-xforge/blob/08a244e18b8a3e3fee6d6866e1380e575e29fba6/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.spec.ts#L361)):*

Was there significance in having `tick()` below `detectChanges()`? Because test-online-status.service.ts setIsOnline() calls `window.dispatchEvent()` (which I assume benefits from tick or flush), and the UI needs to respond to differences in mocked services and perhaps this event, it may be necessary to use `detectChanges` after `tick` to adequately simulate the various things that might happen on the page as a result of changing online status.
</blockquote></details>

No, that was copy-pasta'ed from `suggestions-settings-dialog-component.spec.ts`. The `tick()` technically isn't necessary, and should be the other way around (as you state). For consistency with `edit-name-dialog.component.spec.ts` I have changed it both occurrences to use `this.wait();`.

marksvc (changes requested):
> Reviewed 3 of 6 files at r4, 1 of 1 files at r5, 2 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2414)* status: 12 of 13 files reviewed, 4 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.service.spec.ts` line 105 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-Nwou2yC63I4lzpgqfyx:-NwrO9kE-NFiKpOegLlh:b9zsb74) ([raw file](https://github.com/sillsdev/web-xforge/blob/08a244e18b8a3e3fee6d6866e1380e575e29fba6/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.service.spec.ts#L105)):*

`done()` is not needed because `tick()` handles running the asynchronous code, and ensures that the `expect()` in the `subscribe()` is called.

If it wasn't called, the test would log a warning: `Spec 'DraftGenerationService getBuildProgress should return undefined if offline' has no expectations.`
</blockquote></details>

I did some experimenting.

Suppose we modify a test below so that it reads as follows, with `// req.flush(buildDto);` commented out.
```ts
    fit('should get last completed build and return an observable of BuildDto', fakeAsync(() => {
      // SUT
      service.getLastCompletedBuild(projectId).subscribe(result => {
        expect(result).toEqual(buildDto);
        expect(2 + 2).toEqual(5);
      });
      tick();

      // Setup the HTTP request
      const req = httpTestingController.expectOne(
        `${MACHINE_API_BASE_URL}translation/engines/project:${projectId}/actions/getLastCompletedPreTranslationBuild`
      );
      expect(req.request.method).toEqual('GET');
      // req.flush(buildDto);
      tick();
    }));
```
This test passes, without `expect`ing that 2+2===5.

Now, we can observe whether we remember to call `req.flush` in each test, so that needn't be a problem. I  tried to leave `req.flush` uncommented and yet get the test to incorrectly say it passed, but was not successful. As long as `getLastCompletedBuild` returns, it runs the lambda in `subscribe`. 

It seems like it should be possible to break it :). But I see that it is certainly being more reliable than I was concerned about.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts` line 96 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-NwrIedl4lIA2jAmNoJ7:-NwrIedl4lIA2jAmNoJ8:b9369ss) ([raw file](https://github.com/sillsdev/web-xforge/blob/892e7f568fd82a3f6f72ae81f8d284b3a07ee0c9/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts#L96)):*
> ```XML
>   get canTrainSuggestions(): boolean {
>     // 9 is the minimum number found in testing, but we will use 10 to be safe
>     // We want this to return true if offline, as we do not know whether the user really can train in that instance
> ```

I don't understand the purpose of saying that yes, the user can train suggestions, if the user is offline. 
But I'm also confused about why when this method is called, we do something if canTrainSuggestions===false.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts` line 133 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-NwrGDTe8ozNac8Z4kFi:-NwrGDTe8ozNac8Z4kFj:bamimkb) ([raw file](https://github.com/sillsdev/web-xforge/blob/892e7f568fd82a3f6f72ae81f8d284b3a07ee0c9/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts#L133)):*
> ```XML
>           this.calculateProgress();
>         } finally {
>           this.loadingFinished();
> ```

And maybe we should have a `this.loadingStarted();` before the `try`?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts` line 141 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-NwrHbWYDUwjTfVHNdov:-NwrHbWYDUwjTfVHNdow:b-uf3a5x) ([raw file](https://github.com/sillsdev/web-xforge/blob/892e7f568fd82a3f6f72ae81f8d284b3a07ee0c9/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts#L141)):*
> ```XML
>         this.loadingStarted();
>         try {
>           if (this.translationEngine == null || !this.translationSuggestionsEnabled) {
> ```

What's the purpose of running setupTranslationEngine() if translationSuggestionsEnabled is false? In setupTranslationEngine, it soon returns if translationSuggestionsEnabled is false.
I see this same check is done a dozen lines below.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2414)* status: 10 of 13 files reviewed, 3 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts` line 96 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-NwrIedl4lIA2jAmNoJ7:-Nwrv6SjEUVoaZoQoiMu:b-1fjd04) ([raw file](https://github.com/sillsdev/web-xforge/blob/892e7f568fd82a3f6f72ae81f8d284b3a07ee0c9/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts#L96)):*

I don't understand the purpose of saying that yes, the user can train suggestions, if the user is offline. 
But I'm also confused about why when this method is called, we do something if canTrainSuggestions===false.
</blockquote></details>

Hmmm I think I overloaded the use of this function when I shouldn't have. I have refactored for clarity.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts` line 133 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-NwrGDTe8ozNac8Z4kFi:-NwrvXF11MpcAWmRbUPR:bfc3hct) ([raw file](https://github.com/sillsdev/web-xforge/blob/892e7f568fd82a3f6f72ae81f8d284b3a07ee0c9/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts#L133)):*

And maybe we should have a `this.loadingStarted();` before the `try`?
</blockquote></details>

Thank you! Fixed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts` line 141 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2414#-NwrHbWYDUwjTfVHNdov:-NwrvrZGCMbky6ayc5mx:b-sv7qix) ([raw file](https://github.com/sillsdev/web-xforge/blob/892e7f568fd82a3f6f72ae81f8d284b3a07ee0c9/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.ts#L141)):*

What's the purpose of running setupTranslationEngine() if translationSuggestionsEnabled is false? In setupTranslationEngine, it soon returns if translationSuggestionsEnabled is false.
I see this same check is done a dozen lines below.
</blockquote></details>

Thank you - we don't need the check here, but we do need it below. The reason we need it below, is if translation suggestions gets disabled, it will clear the translation engine. That behavior is not needed on first load.

marksvc (approved):
> Reviewed 1 of 3 files at r7, 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2414)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2423, opened 2024-04-18
Log response times from Paratext registry

Nateowami:
The registry seems to be slow a lot of the time, but not all of the time. We don't have concrete data. This PR adds logging regarding response times so we can analyze whether this is the cause of our problems, and determine if/when the problem is actually fixed.

This is not an ideal solution, but it's quick.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2423)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/InternetSharedRepositorySourceProvider.cs
+++ b/src/SIL.XForge.Scripture/Services/InternetSharedRepositorySourceProvider.cs
@@ -1,4 +1,5 @@
 using System;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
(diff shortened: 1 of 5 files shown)


marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2423)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2423)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2433, opened 2024-04-28
SF-2713 Remove MarbleResourceLookup exception

pmachapman:
When a project has its own set of Biblical Terms with the Versification defined, an error can occur because ParatextData checks for a Marble Resource with the short name of the project.

This PR resolves this issue by defaulting to the ParatextData implementation of `ScrTextCollection.MarbleResourceLookup()` which simply returns null.

The implementation in ParatextData contains the following comment on why it returns null:
```cs
// Since Marble projects are password locked, ParatextData alone can't open them.
```


Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2433)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2447, opened 2024-05-07
SF-2656 Fix Audio Highlight Bug

josephmyers:
Since rxjs calls its event handlers immediately after subscribing to them, the verse change subscription value gets initially set to -1, given the currentTime will be 0 and not resolve to any segment in the timing data. And since we're using distinctUntilChanged(), the handler isn't called again until we reach a found index, e.g. the chapter header/title.

This was causing the software to keep the previous highlight when playing chapter audio until the first timing index was reached. This was confusing for users.

My change ignores the handle called on subscribe, when the time was 0 and the audio hasn't fully loaded yet. Now, the first time the event is handled, the audio will be playing, and thus the checking component will update/clear the highlight correctly.

---

The above comment was made in reference to the scripture audio player. The event handler is called on subscribe, on creation, just after the play button has been pushed, but just before the audio has loaded. Simultaneously, the distinctUntilChanged() call means that it won't be called again until the value changes. Simultaneously to this, the checking component ignores its processing (i.e. the highlight) if the audio isn't playing. So we skip the first event, knowing that the audio isn't loaded yet, which allows the initial "distinct-ified" value to be processed after the audio has loaded.

---

Note that the software before and after this change does not update the highlight while audio is paused. I thought this was weird, but probably intended. Just be aware in testing.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2447)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-test.utils.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-test.utils.ts
@@ -5,7 +5,7 @@ import { AudioPlayer } from '../shared/audio/audio-player';
 
 export function getAudioTimings(): AudioTiming[] {
   return [
-    { textRef: '1', from: 0.0, to: 1.0 },
+    { textRef: '1', from: 0.7, to: 1.0 },
     { textRef: '2', from: 1.0, to: 2.0 },
     { textRef: '3-4', from: 2.0, to: 3.0 }
   ];
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking.utils.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking.utils.spec.ts
@@ -16,6 +16,7 @@ describe('CheckingUtils', () => {
   });
 
   it('can parse audio text ref', () => {
+    expect(CheckingUtils.parseAudioRefByTime(env.audioTimingBasic, 0)).toEqual(undefined);
     expect(CheckingUtils.parseAudioRefByTime(env.audioTimingBasic, 0.9)).toEqual({ verseStr: '1' });
     expect(CheckingUtils.parseAudioRefByTime(env.audioTimingBasic, 1.9)).toEqual({ verseStr: '2' });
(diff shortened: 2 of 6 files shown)


pmachapman (changes requested):
I am not sure skipping is the best option? From what I understand, `BehaviorSubject` will always emit the latest value when subscribed to. In the case of the component initialization, the undefined value will be emitted (and so it is useful to skip, as per your change). Your skip is relying on the `AudioPlayer` to have just been initialized, rather than perhaps mid playback when `timeUpdated$` is subscribed to.

What if we changed `timeUpdated$` to a `Subject<void>`? This would mean that when it is subscribed to, it will not emit the previous event, just any following `this.timeUpdated$.next();` calls.

If I am misunderstanding things (rxjs is not an area I am strong in), let me know, then I will happily approve, as the test, etc look good.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2447#-:-NxjzTev-4qV_zTW6PeJ:b-n0gxz4)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (commented):
Sure, thanks for the feedback. I like your idea better!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2447#-:-NxkS_RG8ywCn9uzjtIb:b856x55)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2447#-:-NxnShse3h4rknE1DFhp:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

pmachapman (approved):
> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2447)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

PR #2457, opened 2024-05-14
SF-2444 Add editor resource tabs

siltomato:
This PR adds the ability for the user to create "resource" tabs of other projects or DBL resources.

- Adds 'Resource' tab menu option when online.
- Clicking 'Resource...' in 'add tab' menu brings up dialog to select a project or resource to add.
- Selecting the project or resource will create an SF project, add the user to it, and display a tab.
  - If SF project already exists, and project has texts, a sync is started in the background and the dialog is closed immediately.
  - Otherwise, the dialog waits for the sync to complete before closing and creating the tab.
- Resource tabs should persist.

**Questions to be resolved:**
  - What user role should be used when adding user to SF project?
  - What icon should be used for the tab, and should the icon be different for projects vs DBL resources?
  - How/when should resources be synced?

### Menu
image

### Dialog
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2457)

### DIFF
--- a/src/RealtimeServer/scriptureforge/models/editor-tab.ts
+++ b/src/RealtimeServer/scriptureforge/models/editor-tab.ts
@@ -1,3 +1,4 @@
-export const editorTabTypes = ['history', 'draft', 'project-source', 'project'] as const;
-export type EditorTabType = typeof editorTabTypes[number];
+export const editorTabTypes = ['history', 'draft', 'project-source', 'project-target', 'project-resource'] as const;
+type EditorTabTypes = typeof editorTabTypes;
+export type EditorTabType = EditorTabTypes[number];
 export type EditorTabGroupType = 'source' | 'target';
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.module.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.module.ts
@@ -29,7 +29,6 @@ import { CoreModule } from './core/core.module';
 import { JoinComponent } from './join/join.component';
 import { NavigationProjectSelectorComponent } from './navigation-project-selector/navigation-project-selector.component';
 import { NavigationComponent } from './navigation/navigation.component';
-import { ProjectSelectComponent } from './project-select/project-select.component';
 import { ProjectComponent } from './project/project.component';
 import { ScriptureChooserDialogComponent } from './scripture-chooser-dialog/scripture-chooser-dialog.component';
 import { DeleteProjectDialogComponent } from './settings/delete-project-dialog/delete-project-dialog.component';
@@ -37,7 +36,6 @@ import { SettingsComponent } from './settings/settings.component';
 import { SharedModule } from './shared/shared.module';
 import { TextNoteDialogComponent } from './shared/text/text-note-dialog/text-note-dialog.component';
 import { StartComponent } from './start/start.component';
-import { SyncProgressComponent } from './sync/sync-progress/sync-progress.component';
 import { SyncComponent } from './sync/sync.component';
 import { TranslateModule } from './translate/translate.module';
 import { UsersModule } from './users/users.module';
@@ -57,8 +55,6 @@ import { UsersModule } from './users/users.module';
     ErrorDialogComponent,
     EditNameDialogComponent,
     FeatureFlagsDialogComponent,
-    ProjectSelectComponent,
-    SyncProgressComponent,
     TextNoteDialogComponent,
     JoinComponent
   ],
@@ -90,7 +86,10 @@ import { UsersModule } from './users/users.module';
     defaultTranslocoMarkupTranspilers(),
     { provide: ErrorHandler, useClass: ExceptionHandlingService },
     { provide: OverlayContainer, useClass: InAppRootOverlayContainer },
-    { provide: MAT_FORM_FIELD_DEFAULT_OPTIONS, useValue: { appearance: 'outline' } }
+    {
+      provide: MAT_FORM_FIELD_DEFAULT_OPTIONS,
+      useValue: { appearance: 'outline' }
+    }
   ],
   bootstrap: [AppComponent]
 })
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.html
@@ -56,7 +56,7 @@
             placeholder="{{ t('source_text_placeholder') }}"
             [projects]="projects"
             [resources]="resources"
-            [hideProjectId]="paratextIdControl.value"
+            [hiddenParatextIds]="[paratextIdControl.value]"
           ></app-project-select>
           <mat-error *ngIf="showResourcesLoadingFailedMessage">{{ t("error_fetching_resources") }}</mat-error>
           <ng-container *ngIf="isBasedOnProjectSet">
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/paratext.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/paratext.service.spec.ts
@@ -1,0 +1,35 @@
+import { HttpClient } from '@angular/common/http';
+import { TestBed } from '@angular/core/testing';
+import { mock } from 'ts-mockito';
+import { AuthService } from 'xforge-common/auth.service';
+import { configureTestingModule } from 'xforge-common/test-utils';
+import { ParatextService } from './paratext.service';
+
+describe('ParatextService', () => {
+  let service: ParatextService;
+  const mockHttpClient = mock(HttpClient);
+  const mockAuthService = mock(AuthService);
+
+  configureTestingModule(() => ({
+    providers: [
+      { provide: HttpClient, useMock: mockHttpClient },
+      { provide: AuthService, useMock: mockAuthService }
+    ]
+  }));
+
+  beforeEach(() => {
+    service = TestBed.inject(ParatextService);
+  });
+
+  describe('isResource', () => {
+    it('should return true for a resource id', () => {
+      const id = '1234567890abcdef';
+      expect(service.isResource(id)).toBe(true);
+    });
+
+    it('should return false for a project id', () => {
+      const id = '123456781234567890abcdef1234567890abcdef1234567890abcdef';
+      expect(service.isResource(id)).toBe(false);
+    });
+  });
+});
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/paratext.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/paratext.service.ts
@@ -8,6 +8,9 @@ import { Snapshot } from 'xforge-common/models/snapshot';
 import { PARATEXT_API_NAMESPACE } from 'xforge-common/url-constants';
 import { ParatextProject } from './models/paratext-project';
 
+/** Length of paratext ids for DBL resources. */
+export const RESOURCE_IDENTIFIER_LENGTH = 16;
+
 /**
  * A point-in-time revision of a document.
  */
@@ -95,5 +98,14 @@ export class ParatextService {
       Accept: 'application/json',
       'Content-Type': 'application/json'
     });
+  }
+
+  /**
+   * Determines if a Paratext id refers to a resource.
+   * @param paratextId The Paratext identifier.
+   * @returns True if the Paratext identifier is a resource identifier.
+   */
+  isResource(paratextId: string): boolean {
+    return paratextId.length === RESOURCE_IDENTIFIER_LENGTH;
   }
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.spec.ts
@@ -4,11 +4,12 @@ import { RouterModule } from '@angular/router';
 import { User } from '@bugsnag/js';
 import { cloneDeep } from 'lodash-es';
 import { createTestUser } from 'realtime-server/lib/esm/common/models/user-test-data';
+import { RecursivePartial } from 'realtime-server/lib/esm/common/utils/type-utils';
 import { SFProjectProfile } from 'realtime-server/lib/esm/scriptureforge/models/sf-project';
-import { SFProjectRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-role';
+import { isParatextRole, SFProjectRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-role';
 import { createTestProjectProfile } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-test-data';
 import { TextInfoPermission } from 'realtime-server/lib/esm/scriptureforge/models/text-info-permission';
-import { anything, instance, mock, when } from 'ts-mockito';
+import { anything, instance, mock, verify, when } from 'ts-mockito';
 import { UserDoc } from 'xforge-common/models/user-doc';
 import { TestRealtimeModule } from 'xforge-common/test-realtime.module';
 import { TestRealtimeService } from 'xforge-common/test-realtime.service';
@@ -18,6 +19,7 @@ import { UserService } from 'xforge-common/user.service';
 import { SFProjectProfileDoc } from '../core/models/sf-project-profile-doc';
 import { SF_TYPE_REGISTRY } from './models/sf-type-registry';
 import { TextDocId } from './models/text-doc';
+import { RESOURCE_IDENTIFIER_LENGTH } from './paratext.service';
 import { PermissionsService } from './permissions.service';
 import { SFProjectService } from './sf-project.service';
 
@@ -41,60 +43,60 @@ describe('PermissionsService', () => {
 
   it('allows commenters to access Translate', fakeAsync(() => {
     const env = new TestEnvironment();
-    when(mockedUserService.currentUserId).thenReturn('commenter');
+    when(mockedUserService.currentUserId).thenReturn(SFProjectRole.Commenter);
 
     expect(env.service.canAccessTranslate(env.projectDoc)).toBe(true);
-    expect(env.service.canAccessTranslate(env.projectDoc, 'commenter')).toBe(true);
+    expect(env.service.canAccessTranslate(env.projectDoc, SFProjectRole.Commenter)).toBe(true);
   }));
 
   it('allows checkers to access Community Checking if enabled', fakeAsync(() => {
     const env = new TestEnvironment();
-    when(mockedUserService.currentUserId).thenReturn('checker');
+    when(mockedUserService.currentUserId).thenReturn(SFProjectRole.CommunityChecker);
 
     expect(env.service.canAccessCommunityChecking(env.projectDoc)).toBe(true);
-    expect(env.service.canAccessCommunityChecking(env.projectDoc, 'checker')).toBe(true);
+    expect(env.service.canAccessCommunityChecking(env.projectDoc, SFProjectRole.CommunityChecker)).toBe(true);
   }));
 
   it('allows admins to access both', fakeAsync(() => {
     const env = new TestEnvironment();
-    when(mockedUserService.currentUserId).thenReturn('projectAdmin');
+    when(mockedUserService.currentUserId).thenReturn(SFProjectRole.ParatextAdministrator);
 
     expect(env.service.canAccessTranslate(env.projectDoc)).toBe(true);
-    expect(env.service.canAccessTranslate(env.projectDoc, 'projectAdmin')).toBe(true);
+    expect(env.service.canAccessTranslate(env.projectDoc, SFProjectRole.ParatextAdministrator)).toBe(true);
     expect(env.service.canAccessCommunityChecking(env.projectDoc)).toBe(true);
-    expect(env.service.canAccessCommunityChecking(env.projectDoc, 'projectAdmin')).toBe(true);
+    expect(env.service.canAccessCommunityChecking(env.projectDoc, SFProjectRole.ParatextAdministrator)).toBe(true);
   }));
 
-  it('doesnt allow checkers to access Translate', fakeAsync(() => {
+  it('does not allow checkers to access Translate', fakeAsync(() => {
     const env = new TestEnvironment();
-    when(mockedUserService.currentUserId).thenReturn('checker');
+    when(mockedUserService.currentUserId).thenReturn(SFProjectRole.CommunityChecker);
 
     expect(env.service.canAccessTranslate(env.projectDoc)).toBe(false);
-    expect(env.service.canAccessTranslate(env.projectDoc, 'checker')).toBe(false);
+    expect(env.service.canAccessTranslate(env.projectDoc, SFProjectRole.CommunityChecker)).toBe(false);
   }));
 
-  it('doesnt allow commenters to access Community Checking', fakeAsync(() => {
+  it('does not allow commenters to access Community Checking', fakeAsync(() => {
     const env = new TestEnvironment();
-    when(mockedUserService.currentUserId).thenReturn('commenter');
+    when(mockedUserService.currentUserId).thenReturn(SFProjectRole.Commenter);
 
     expect(env.service.canAccessCommunityChecking(env.projectDoc)).toBe(false);
-    expect(env.service.canAccessCommunityChecking(env.projectDoc, 'commenter')).toBe(false);
+    expect(env.service.canAccessCommunityChecking(env.projectDoc, SFProjectRole.Commenter)).toBe(false);
   }));
 
-  it('doesnt allow checkers to access Community Checking if not enabled', fakeAsync(() => {
+  it('does not allow checkers to access Community Checking if not enabled', fakeAsync(() => {
     const env = new TestEnvironment(false);
-    when(mockedUserService.currentUserId).thenReturn('checker');
+    when(mockedUserService.currentUserId).thenReturn(SFProjectRole.CommunityChecker);
 
     expect(env.service.canAccessCommunityChecking(env.projectDoc)).toBe(false);
-    expect(env.service.canAccessCommunityChecking(env.projectDoc, 'checker')).toBe(false);
+    expect(env.service.canAccessCommunityChecking(env.projectDoc, SFProjectRole.CommunityChecker)).toBe(false);
   }));
 
-  it('doesnt allow admins to access Community Checking if not enabled', fakeAsync(() => {
+  it('does not allow admins to access Community Checking if not enabled', fakeAsync(() => {
     const env = new TestEnvironment(false);
-    when(mockedUserService.currentUserId).thenReturn('projectAdmin');
+    when(mockedUserService.currentUserId).thenReturn(SFProjectRole.ParatextAdministrator);
 
     expect(env.service.canAccessCommunityChecking(env.projectDoc)).toBe(false);
-    expect(env.service.canAccessCommunityChecking(env.projectDoc, 'projectAdmin')).toBe(false);
+    expect(env.service.canAccessCommunityChecking(env.projectDoc, SFProjectRole.ParatextAdministrator)).toBe(false);
   }));
 
   it('allows access to text if user is on project and has permission', fakeAsync(async () => {
@@ -105,7 +107,7 @@ describe('PermissionsService', () => {
     expect(await env.service.canAccessText(cloneDeep(textDoc) as TextDocId)).toBe(true);
   }));
 
-  it('doesnt allow access to text if user is not on project', fakeAsync(async () => {
+  it('does not allow access to text if user is not on project', fakeAsync(async () => {
     const env = new TestEnvironment();
     env.setCurrentUser('other');
 
@@ -114,7 +116,7 @@ describe('PermissionsService', () => {
     expect(await env.service.canAccessText(cloneDeep(textDoc) as TextDocId)).toBe(false);
   }));
 
-  it('doesnt allow access to text if user has no access', fakeAsync(async () => {
+  it('does not allow access to text if user has no access', fakeAsync(async () => {
     const env = new TestEnvironment();
     env.setupProjectData(TextInfoPermission.None);
 
@@ -122,28 +124,78 @@ describe('PermissionsService', () => {
 
     expect(await env.service.canAccessText(cloneDeep(textDoc) as TextDocId)).toBe(false);
   }));
+
+  describe('canSync', () => {
+    it('returns false when projectDoc.data is undefined', fakeAsync(() => {
+      const env = new TestEnvironment();
+      when(mockedProjectDoc.data).thenReturn(undefined);
+      expect(env.service.canSync(env.projectDoc, SFProjectRole.ParatextAdministrator)).toBe(false);
+    }));
+
+    it('uses current user id if userId is not provided', () => {
+      const env = new TestEnvironment();
+      env.service.canSync(env.projectDoc);
+      verify(mockedUserService.currentUserId).once();
+    });
+
+    it('does not use current user id if userId is provided', () => {
+      const env = new TestEnvironment();
+      env.service.canSync(env.projectDoc, SFProjectRole.ParatextAdministrator);
+      verify(mockedUserService.currentUserId).never();
+    });
+
+    it('allows PT admin role to sync projects', () => {
+      const env = new TestEnvironment();
+      env.setProjectType('project');
+      expect(env.service.canSync(env.projectDoc, SFProjectRole.ParatextAdministrator)).toBe(true);
+    });
+
+    it('allows PT translator role to sync projects', () => {
+      const env = new TestEnvironment();
+      env.setProjectType('project');
+      expect(env.service.canSync(env.projectDoc, SFProjectRole.ParatextTranslator)).toBe(true);
+    });
+
+    it('allows any PT role to sync resources', () => {
+      const env = new TestEnvironment();
+      env.setProjectType('resource');
+
+      Object.values(SFProjectRole).forEach(role => {
+        if (!isParatextRole(role)) {
+          return;
+        }
+
+        expect(env.service.canSync(env.projectDoc, role)).toBe(true);
+      });
+    });
+
+    it('disallows non- PT admin/translator roles to sync projects', () => {
+      const env = new TestEnvironment();
+      env.setProjectType('project');
+
+      Object.values(SFProjectRole).forEach(role => {
+        if (role === SFProjectRole.ParatextAdministrator || role === SFProjectRole.ParatextTranslator) {
+          return;
+        }
+
+        expect(env.service.canSync(env.projectDoc, role)).toBe(false);
+      });
+    });
+  });
 });
 class TestEnvironment {
   readonly service: PermissionsService;
   readonly projectDoc: SFProjectProfileDoc = instance(mockedProjectDoc);
   private readonly realtimeService: TestRealtimeService = TestBed.inject<TestRealtimeService>(TestRealtimeService);
 
-  constructor(checkingEnabled = true) {
+  constructor(readonly checkingEnabled = true) {
     this.service = TestBed.inject(PermissionsService);
-    const data = createTestProjectProfile({
-      userRoles: {
-        projectAdmin: SFProjectRole.ParatextAdministrator,
-        checker: SFProjectRole.CommunityChecker,
-        commenter: SFProjectRole.Commenter
-      },
-      checkingConfig: {
-        checkingEnabled: checkingEnabled
-      }
-    });
-    when(mockedProjectDoc.data).thenReturn(data);
+
     when(mockedProjectService.getProfile(anything())).thenCall(id =>
       this.realtimeService.subscribe(SFProjectProfileDoc.COLLECTION, id)
     );
+
+    this.setProjectProfile();
     this.setupProjectData();
     this.setCurrentUser();
     this.setupUserData();
@@ -202,5 +254,31 @@ class TestEnvironment {
         }
       })
     });
+  }
+
+  setProjectProfile(overrides?: RecursivePartial<SFProjectProfile>): void {
+    const userRoles = Object.values(SFProjectRole).reduce((roles, role) => ({ ...roles, [role]: role }), {});
+    const config = {
+      userRoles,
+      checkingConfig: {
+        checkingEnabled: this.checkingEnabled
+      },
+      ...overrides
+    };
+    const data = createTestProjectProfile(config);
+    when(mockedProjectDoc.data).thenReturn(data);
+  }
+
+  setProjectType(projectType: 'resource' | 'project'): string {
+    // DBL resources can be identified by length of paratext id
+    const paratextId = Array(projectType === 'resource' ? RESOURCE_IDENTIFIER_LENGTH : 40)
+      .fill('a')
+      .join('');
+
+    this.setProjectProfile({
+      paratextId: paratextId
+    });
+
+    return paratextId;
   }
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@angular/core';
 import { Operation } from 'realtime-server/lib/esm/common/models/project-rights';
 import { SF_PROJECT_RIGHTS, SFProjectDomain } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-rights';
-import { SFProjectRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-role';
+import { isParatextRole, SFProjectRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-role';
 import { Chapter } from 'realtime-server/lib/esm/scriptureforge/models/text-info';
 import { TextInfoPermission } from 'realtime-server/lib/esm/scriptureforge/models/text-info-permission';
 import { UserService } from 'xforge-common/user.service';
@@ -9,11 +9,16 @@ import { environment } from '../../environments/environment';
 import { SFProjectProfileDoc } from './models/sf-project-profile-doc';
 import { roleCanAccessCommunityChecking, roleCanAccessTranslate } from './models/sf-project-role-info';
 import { TextDocId } from './models/text-doc';
+import { ParatextService } from './paratext.service';
 import { SFProjectService } from './sf-project.service';
 
 @Injectable({ providedIn: 'root' })
 export class PermissionsService {
-  constructor(private readonly userService: UserService, private readonly projectService: SFProjectService) {}
+  constructor(
+    private readonly userService: UserService,
+    private readonly projectService: SFProjectService,
+    private readonly paratextService: ParatextService
+  ) {}
 
   canAccessCommunityChecking(project: SFProjectProfileDoc, userId?: string): boolean {
     if (project.data == null) return false;
@@ -58,5 +63,21 @@ export class PermissionsService {
     }
 
     return false;
+  }
+
+  canSync(projectDoc: SFProjectProfileDoc, userId?: string): boolean {
+    if (projectDoc.data == null) {
+      return false;
+    }
+
+    const role: string = projectDoc.data.userRoles[userId ?? this.userService.currentUserId];
+
+    // Any paratext user role can sync DBL resources
+    if (this.paratextService.isResource(projectDoc.data.paratextId)) {
+      return isParatextRole(role);
+    }
+
+    // Only PT admin and PT translator can sync non-resource projects
+    return role === SFProjectRole.ParatextAdministrator || role === SFProjectRole.ParatextTranslator;
   }
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts
@@ -11,7 +11,6 @@ import { getSFProjectUserConfigDocId } from 'realtime-server/lib/esm/scripturefo
 import { TextAudio } from 'realtime-server/lib/esm/scriptureforge/models/text-audio';
 import { Subject } from 'rxjs';
 import { CommandService } from 'xforge-common/command.service';
-import { FileService } from 'xforge-common/file.service';
 import { LocationService } from 'xforge-common/location.service';
 import { RealtimeQuery } from 'xforge-common/models/realtime-query';
 import { ProjectService } from 'xforge-common/project.service';
@@ -42,7 +41,6 @@ export class SFProjectService extends ProjectService<SFProject, SFProjectDoc> {
   constructor(
     realtimeService: RealtimeService,
     commandService: CommandService,
-    private readonly fileService: FileService,
     private readonly locationService: LocationService,
     protected readonly retryingRequestService: RetryingRequestService
   ) {
@@ -51,6 +49,47 @@ export class SFProjectService extends ProjectService<SFProject, SFProjectDoc> {
 
   async onlineCreate(settings: SFProjectCreateSettings): Promise<string> {
     return (await this.onlineInvoke<string>('create', { settings }))!;
+  }
+
+  /**
+   * Creates an SF project/resource with the given paratext id.
+   * @param paratextId The paratext id of the project or resource.
+   * @returns The SF project id.
+   */
+  async onlineCreateResourceProject(paratextId: string): Promise<string | undefined> {
+    return this.onlineInvoke<string>('createResourceProject', { paratextId });
+  }
+
+  async getRealtimeProject(paratextId: string): Promise<SFProjectDoc | undefined> {
+    const query: RealtimeQuery<SFProjectDoc> = await this.realtimeService.onlineQuery<SFProjectDoc>(this.collection, {
+      paratextId
+    });
+    return query.docs[0];
+  }
+
+  /**
+   * Gets the SF project Id for the project/resource with the given paratext id.  Creates the project if needed.
+   * @param paratextId The paratext id of the project or resource.
+   * @returns The SF project id.
+   */
+  async getOrCreateRealtimeProject(paratextId: string, role?: string): Promise<string | undefined> {
+    // First check if project exists
+    const project: SFProjectDoc | undefined = await this.getRealtimeProject(paratextId);
+
+    // If SF project exists, return its id
+    if (project != null) {
+      return Promise.resolve(project.id);
+    }
+
+    // Otherwise, create realtime project
+    const projectId: string | undefined = await this.onlineCreateResourceProject(paratextId);
+
+    // Add current user to project
+    if (projectId != null) {
+      await this.onlineAddCurrentUser(projectId, role);
+    }
+
+    return projectId;
   }
 
   /**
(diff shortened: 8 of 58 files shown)


RaymondLuong3 (changes requested):
I am working my way through reviewing this PR. I posted my initial comments.

> Reviewed 27 of 41 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-:-Ny7fDiv4uyUiHiKEo7y:b-dwaxbn)* status: 27 of 41 files reviewed, 10 unresolved discussions (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts` line 61 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7TEUE3S8CbZgpfIwN:-Ny7TEUE3S8CbZgpfIwO:b-ofs21c) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts#L61)):*
> ```XML
>   async onlineCreateResourceProject(paratextId: string): Promise<string | undefined> {
>     return this.onlineInvoke<string>('createResourceProject', { paratextId });
>   }
> ```

It looks like there is already a method called `onlineCreate` in `SFProjectService. I would think we would go the same route as what is on the connect project page so we don't have to introduce a new method.



_Code quote:_
```XML
  async onlineCreateResourceProject(paratextId: string): Promise<string | undefined> {
    return this.onlineInvoke<string>('createResourceProject', { paratextId });
  }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts` line 65 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7VmXMBEhOzsCDg4U6:-Ny7VmXMBEhOzsCDg4U7:b-ijff0c) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts#L65)):*
> ```XML
>   }
> 
>   onTabAddRequest(newTabType: string): void {
> ```

Can you add a method comment. From looking at it, I cannot tell what it is accomplishing.



_Code quote:_
```XML
  onTabAddRequest(newTabType: string): void {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 636 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7ZuSU6K05Nj0Or35Z:-Ny7ZuSU6K05Nj0Or35_:biy7rl0) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L636)):*
> ```XML
>   ngOnInit(): void {
>     this.activatedProject.projectDoc$
>       .pipe(takeUntilDestroyed(this.destroyRef), filterNullish(), distinctUntilKeyChanged('id'))
> ```

Out of curiosity, is the `distinctUntilKeyChanged` used here just to be safe? My understanding is that `projectDoc$` only emits if the user changes projects.



_Code quote:_
```XML
      .pipe(takeUntilDestroyed(this.destroyRef), filterNullish(), distinctUntilKeyChanged('id'))
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1184 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7_jb68aOkVspAbALm:-Ny7_jb68aOkVspAbALn:br377xn) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1184)):*
> ```XML
>             persistedTabs.map(async tabData => ({
>               ...tabData,
>               projectDoc: tabData.projectId ? await this.projectService.getProfile(tabData.projectId) : undefined
> ```

Nit: Be explicit where possible i.e. `tabData.projectId != null`



_Code quote:_
```XML
              projectDoc: tabData.projectId
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1232 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7aHAH616JNIMZdEdc:-Ny7aHAH616JNIMZdEdd:b-o12t44) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1232)):*
> ```XML
>         take(1)
>       )
>       .subscribe();
> ```

Is it necessary to call subscribe here with no observer?



_Code quote:_
```XML
      .subscribe();
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.html` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7bBZuCps8d0SqOhTP:-Ny7bBZv1lYC1OpPJw4n:bqg5fme) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.html#L3)):*
> ```HTML
> <app-text
>   *ngIf="projectId"
>   [id]="projectId | textDocId : bookNum : chapter"
> ```

Am I understanding this correctly? The project ID is being piped to the textDocId pipe that also uses the booknum and chapternum? This is neat :)



_Code quote:_
```HTML
  [id]="projectId | textDocId : bookNum : chapter"
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.scss` line 0 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7beDX1pvjqdPpTECU:-Ny7beDX1pvjqdPpTECV:bc8r4q7) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.scss#L0)):*
You can delete this file.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.ts` line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7bpEQE31FeeoiDwqS:-Ny7bpEQE31FeeoiDwqT:bnbq142) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.ts#L13)):*
> ```XML
>   selector: 'app-editor-resource',
>   templateUrl: './editor-resource.component.html',
>   styleUrls: ['./editor-resource.component.scss']
> ```

This line can be removed



_Code quote:_
```XML
  styleUrls: ['./editor-resource.component.scss']
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.ts` line 42 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7cDGP2oQwM9qWwGc6:-Ny7cDGP2oQwM9qWwGc7:b-4oh6ij) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.ts#L42)):*
> ```XML
>   }
> 
>   initProjectDetails(): void {
> ```

Typically, we wouldn't want to create public methods that would otherwise be private just so that we can expose them to testing. Can we make this private and call it in tests like `component['initProjectDetails']()`? That should be possible?



_Code quote:_
```XML
  initProjectDetails(): void {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-request.service.ts` line 31 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7e2OPCBgqzGeDCGxa:-Ny7e2OPCBgqzGeDCGxb:bf9xcqk) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-request.service.ts#L31)):*
> ```XML
>         return this.getParatextIdsForOpenTabs().pipe(
>           switchMap(paratextIds => {
>             return this.dialogService
> ```

How about creating a new method called `openDialog()` and putting this logic there. It will make this rxjs chain easier to read.



_Code quote:_
```XML
            return this.dialogService
```

RaymondLuong3 (changes requested):
> Reviewed 8 of 41 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2457)* status: 35 of 41 files reviewed, 15 unresolved discussions (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-request.service.spec.ts` line 64 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWKhPOCLGrorspkVqb:-NyWKhPOCLGrorspkVqc:bfk1bm3) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-request.service.spec.ts#L64)):*
> ```XML
> 
>     when(tabStateService.tabs$).thenReturn(of([{ projectId: projectDoc1.id }, {}, { projectId: projectDoc2.id }]));
>     when(projectService.get(projectDoc1.id)).thenReturn(Promise.resolve(projectDoc1));
> ```

I think a shorthand for this is `thenResolve(...)`. You might like it.



_Code quote:_
```XML
    when(projectService.get(projectDoc1.id)).thenReturn(Promise.resolve(projectDoc1));
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.ts` line 50 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWLom1DDK37m0pN5OE:-NyWLom1DDK37m0pN5OF:bmtg4c7) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.ts#L50)):*
> ```XML
>             type: tabType,
>             icon: 'book',
>             headerText: tabOptions?.headerText ?? 'Project',
> ```

"Project" should be localizable right?



_Code quote:_
```XML
           headerText: tabOptions?.headerText ?? 'Project',
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.ts` line 66 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWLu5UA6DebLWUjlwb:-NyWLu5UA6DebLWUjlwc:bjasocj) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.ts#L66)):*
> ```XML
>             type: tabType,
>             icon: 'library_books',
>             headerText: tabOptions?.headerText ?? 'Resource',
> ```

"Resources" should be localizable right?



_Code quote:_
```XML
            headerText: tabOptions?.headerText ?? 'Resource',
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-menu.service.ts` line 136 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWO4iy1PWc-yIpqkJ9:-NyWO4izBJnBKLW6Iy8G:bw46sh9) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-menu.service.ts#L136)):*
> ```XML
> 
>   private canShowResource(_projectDoc: SFProjectProfileDoc | undefined): boolean {
>     // TODO: what role check is needed here?
> ```

I imagine we need to check that a user is a Paratext user at a minimum. Viewers and Commenters will not be able to open up resources at all.



_Code quote:_
```XML
   // TODO: what role check is needed here?
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html` line 2 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWT-lW27vt-TXzTFzb:-NyWT-lW27vt-TXzTFzc:b-ksomp4) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html#L2)):*
> ```HTML
> <mat-progress-bar *ngIf="isLoading && !isSyncActive" mode="indeterminate" color="accent"></mat-progress-bar>
> <app-sync-progress
> ```

This is neat! I see the sync progress happening right in this dialog.

RaymondLuong3 (changes requested):
Nice work. I have a few more suggestions. I noticed that when I connect to a new project or resource, the dialog will close after the project is synchronized, but it still takes about 10 seconds for the text docs to be visible. I wonder if users will be confused that the text for the project is not visible for a few seconds after the dialog closes.
That could be something to improve on after this PR.

> Reviewed 4 of 41 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-:-NyWk0VJEj7oTejHRygj:bnw0kpa)* status: 39 of 41 files reviewed, 18 unresolved discussions (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts` line 61 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7TEUE3S8CbZgpfIwN:-NyWmiAm65-I2qGbwa7U:b-nm23f2) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project.service.ts#L61)):*

It looks like there is already a method called `onlineCreate` in `SFProjectService. I would think we would go the same route as what is on the connect project page so we don't have to introduce a new method.
</blockquote></details>

Ah yes, I see now. The backend needed to expose a new API endpoint to allow creating a resource project that does not already exist. Thanks.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 64 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWeEJ31NxWdJtN9Ivj:-NyWeEJ31NxWdJtN9Ivk:b-3spsrq) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L64)):*
> ```XML
>   }
> 
>   async getProjectsAndResources(): Promise<void> {
> ```

This method could be made private.



_Code quote:_
```XML
  async getProjectsAndResources(): Promise<void> {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 139 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWgj4Q9UHYuP9z7q9r:-NyWgj4Q9UHYuP9z7q9s:bc3hm9l) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L139)):*
> ```XML
>   }
> 
>   cancelSync(): void {
> ```

Looks like this an all methods below it can be made private.



_Code quote:_
```XML
  cancelSync(): void {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 160 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWgxO20aQpeMe-k2sS:-NyWgxO20aQpeMe-k2sT:b-d4n3hz) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L160)):*
> ```XML
>     const selectedProjectId: string | undefined = await this.projectService.getOrCreateRealtimeProject(
>       paratextId,
>       SFProjectRole.ParatextAdministrator // TODO: what role should be used?
> ```

I think they should be the role that they have on the project, or an observer if the project is a resource.



_Code quote:_
```XML
      SFProjectRole.ParatextAdministrator // TODO: what role should be used?
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 459 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWjb1AEPeFVgSHUmRa:-NyWjb1AEPeFVgSHUmRb:b-4qkps3) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L459)):*
> ```JSON
>     "error_loading_resource": "There was an error loading the selected resource.",
>     "loading": "Loading",
>     "placeholder_loading": "{{ editor_add_tab_resource_dialog.loading }}...",
> ```

Interesting, so this is a way of using an existing string and interpolating the localized string for a new string. Nice



_Code quote:_
```JSON
    "placeholder_loading": "{{ editor_add_tab_resource_dialog.loading }}...",
```

siltomato (commented):
I haven't experienced that delay yet, myself.  I wonder where the slow down is after it has finished syncing?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-:-Nyh3ZxqELFMWPR7S7bh:bxqo4z5)* status: 39 of 41 files reviewed, 13 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts` line 65 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7VmXMBEhOzsCDg4U6:-NyVeaE47namCFoUAbGg:b-nl1nbk) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.ts#L65)):*

Can you add a method comment. From looking at it, I cannot tell what it is accomplishing.
</blockquote></details>

For 'project-resource' tabs, more prompting and processing needs to be done before the tab can be created, so I abstracted the concept to pass all tab 'add' requests through a service to determine and perform any further processing.

In the case of 'project-resource' tabs, `TabAddRequestService` (via the provided `EditorTabAddRequestService`) will return the user-selected projectId for the resource in the returned `tabOptions`.

Let me know if my added comment isn't descriptive enough.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 636 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7ZuSU6K05Nj0Or35Z:-Nyh3-Ci5_3Y8g6bLctQ:bdd71t) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L636)):*

Out of curiosity, is the `distinctUntilKeyChanged` used here just to be safe? My understanding is that `projectDoc$` only emits if the user changes projects.
</blockquote></details>

Good point.  I think I had it on there when I was experimenting, and I didn't remove it.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1184 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7_jb68aOkVspAbALm:-NyVvVWXCs8Vg_aNUPE6:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1184)):*

Nit: Be explicit where possible i.e. `tabData.projectId != null`
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1232 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7aHAH616JNIMZdEdc:-NyWGiDP0-_eBE5_tNlK:bg9qjyj) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1232)):*

Is it necessary to call subscribe here with no observer?
</blockquote></details>

Subscribe is necessary to trigger the execution of the observable.  The bulk of the work here is done in the 'tap', which isn't really necessary.  It may be left over from a different pipe chain I was experimenting with.  I'll move it to subscribe.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.html` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7bBZuCps8d0SqOhTP:-NyWJr8PDH4T30c_qZW-:bj6tf5j) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.html#L3)):*

Am I understanding this correctly? The project ID is being piped to the textDocId pipe that also uses the booknum and chapternum? This is neat :)
</blockquote></details>

Yes!  The pipe is nice in angular templates because is is treated as 'pure' by angular, whereas calling a function (or getter) in the template will cause angular change detection to call the function a *ton* (unless you specify 'OnPush' change detection strategy) because it doesn't know if the output of the function will change.

It would be nice if angular had a way to mark a component function as pure so that angular change detection would know that if the function input didn't change, the output would also not change.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.scss` line  at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7beDX1pvjqdPpTECU:-NyWNs4J4cvegub63hHX:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.scss#L)):*

You can delete this file.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.ts` line 13 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7bpEQE31FeeoiDwqS:-NyWNt0_0fdRQ0KYv6zb:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.ts#L13)):*

This line can be removed
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.ts` line 42 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7cDGP2oQwM9qWwGc6:-NybK7L_DVcLB1Bb31ix:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-resource/editor-resource.component.ts#L42)):*

Typically, we wouldn't want to create public methods that would otherwise be private just so that we can expose them to testing. Can we make this private and call it in tests like `component['initProjectDetails']()`? That should be possible?
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-request.service.ts` line 31 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Ny7e2OPCBgqzGeDCGxa:-NybOiO-0BvW0NKD0VFy:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-request.service.ts#L31)):*

How about creating a new method called `openDialog()` and putting this logic there. It will make this rxjs chain easier to read.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-request.service.spec.ts` line 64 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWKhPOCLGrorspkVqb:-NybPFVjCbU7jbKZg1i7:b-jqcx6n) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-request.service.spec.ts#L64)):*

I think a shorthand for this is `thenResolve(...)`. You might like it.
</blockquote></details>

Ah, much better, thanks!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.ts` line 50 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWLom1DDK37m0pN5OE:-NybQTz06pul-F_9HESv:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.ts#L50)):*

"Project" should be localizable right?
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.ts` line 66 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWLu5UA6DebLWUjlwb:-NybQUK45h6lYbFuHZBc:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.ts#L66)):*

"Resources" should be localizable right?
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-menu.service.ts` line 136 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWO4iy1PWc-yIpqkJ9:-Nyh0NHEE-BUrwSvqobu:b-ci7kh2) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-menu.service.ts#L136)):*

I imagine we need to check that a user is a Paratext user at a minimum. Viewers and Commenters will not be able to open up resources at all.
</blockquote></details>

The user will need to sync to load the texts, which I believe means that the user will need to be a PT admin or PT translator?  I added this check as `canSync` to `PermissionsService`.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html` line 2 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWT-lW27vt-TXzTFzb:-NybQb-X54KgXGlhMBa1:b-yxsjzj) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html#L2)):*

This is neat! I see the sync progress happening right in this dialog.
</blockquote></details>

Yes, I figured I shouldn't close the dialog until the texts were available, so the user needs some progress updates!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 64 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWeEJ31NxWdJtN9Ivj:-NybRnIa6bY2RRXouomx:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L64)):*

This method could be made private.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 139 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWgj4Q9UHYuP9z7q9r:-NybRnxK8KAhEFXcyZCa:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L139)):*

Looks like this an all methods below it can be made private.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 160 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWgxO20aQpeMe-k2sS:-Nyh0qe-4LVg3YgeVhBS:br6fby8) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L160)):*

I think they should be the role that they have on the project, or an observer if the project is a resource.
</blockquote></details>

I tried leaving this blank, which appears to add the user as a PT observer, but that doesn't allow the user to sync.  Any ideas?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 459 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWjb1AEPeFVgSHUmRa:-NybRvmk3qUMu8Xp8_GB:b-ivufer) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L459)):*

Interesting, so this is a way of using an existing string and interpolating the localized string for a new string. Nice
</blockquote></details>

Yes, Peter showed me that one.  It may be useful for all our instances of "Scripture Forge" in the translation files.

siltomato (commented):
While working on the "adding user to project/resource" changes, I needed a utility function to check for paratext id length, and I noticed that @pmachapman recently created `isResource`  in `ServalAdministrationService`.  It seemed out of place there, so I moved the `isResource` function to a new utility service: `SFProjectUtil`.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-:-Nz4fOCf8RF6c24Z5fnw:blu496o)* status: 34 of 54 files reviewed, 4 unresolved discussions (waiting on @github-advanced-security[bot], @pmachapman, and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.spec.ts` line 8 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2457#gh-1614170914:-NyzUSyP1L8hMnl6DYRK:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/e5e8d11d76741c11caf17ebf84a6dd9ad6d02061/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.spec.ts#L8)):*
<details><summary><i>Previously, github-advanced-security[bot] wrote…</i></summary><blockquote>

## Expression has no effect

This expression has no effect.

[Show more details](https://github.com/sillsdev/web-xforge/security/code-scanning/807)
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 160 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWgxO20aQpeMe-k2sS:-Nz4dzxa695RzHV-OFXg:b-fd87n) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L160)):*

You will want to make changes to the back-end. In `SFProjectServices.cs`, you can allow projects that are resource projects to be synchronized by `pt_observers`. Currently, a project can only be synchronized by administrators or translators.
You can distinguish resources from Paratext projects because resources have IDs that are 16 characters and can be identified by `ParatextService.IsResource`.
For me, that would be the best way to allow syncing Paratext resources.
Then I think that if a user connects to an SF project, they automatically get set to the user role they have on the project. But that would be good to confirm.
</blockquote></details>

Ok, the user should now be added to the project/resource with the role that they have on the project, or an observer if the project is a resource.  I modified `SyncAsync` in `SFProjectService` to allow resources to sync with any paratext role.  Projects still require admin or translator.  @pmachapman, is this kosher?

pmachapman (commented):
> I noticed that [@pmachapman](https://github.com/pmachapman) recently created `isResource` in `ServalAdministrationService`. It seemed out of place there, so I moved the `isResource` function to a new utility service: `SFProjectUtil`.

Sounds good. I have been wanting to use it in other components, so have contemplated the same change.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-:-Nz4vi8s44rDEHwZFNUZ:beho7oy)* status: 34 of 54 files reviewed, 4 unresolved discussions (waiting on @github-advanced-security[bot], @RaymondLuong3, and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 160 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NyWgxO20aQpeMe-k2sS:-Nz4vwdf55p0BLPwy8R_:b7jzzr6) ([raw file](https://github.com/sillsdev/web-xforge/blob/548842739c4425a6e69c1a4b971dd520c3c6e9f5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L160)):*
> [@pmachapman](https://github.com/pmachapman), is this kosher?

Yes. That sounds sensible given no one can administer or translate a resource.

RaymondLuong3 (changes requested):
> Reviewed 2 of 20 files at r5, 18 of 18 files at r7, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2457)* status: all files reviewed, 8 unresolved discussions (waiting on @github-advanced-security[bot] and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.spec.ts` line 247 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Nz5GRvr9nJE_CTfC6Rc:-Nz5GRvr9nJE_CTfC6Rd:b-6rgck4) ([raw file](https://github.com/sillsdev/web-xforge/blob/ef3607c76d7a10fc8d6bd821773aef6abb854995/src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.spec.ts#L247)):*
> ```XML
>   }
> 
>   setProjectProfile(overrides?: RecursivePartial<SFProjectProfile>): void {
> ```

This can be made private, unless you think it will be useful for future tests.



_Code quote:_
```XML
  setProjectProfile(overrides?: RecursivePartial<SFProjectProfile>): void {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project-util.service.ts` line 10 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Nz5Grb8A6eUABhDpfnB:-Nz5Grb8A6eUABhDpfnC:b-9iyygl) ([raw file](https://github.com/sillsdev/web-xforge/blob/ef3607c76d7a10fc8d6bd821773aef6abb854995/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project-util.service.ts#L10)):*
> ```XML
>  */
> @Injectable({ providedIn: 'root' })
> export class SFProjectUtilService {
> ```

This should be moved into the `paratext.service.ts` file. We want to keep Paratext related functions there and not have it leak into services across our codebase. I know this already existed in the `serval-administration.service`. I would argue it still belongs within `paratext.service`.



_Code quote:_
```XML
export class SFProjectUtilService {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project-util.service.spec.ts` line 3 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Nz5HFDE71ht79BoqTtS:-Nz5HFDE71ht79BoqTtT:bkc4q9b) ([raw file](https://github.com/sillsdev/web-xforge/blob/ef3607c76d7a10fc8d6bd821773aef6abb854995/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project-util.service.spec.ts#L3)):*
> ```XML
> import { SFProjectUtilService } from './sf-project-util.service';
> 
> describe('SFProjectUtilService', () => {
> ```

Then this file can be renamed paratext.service.spec.ts.



_Code quote:_
```XML
describe('SFProjectUtilService', () => {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/serval-administration/serval-projects.component.ts` line 28 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Nz5Hmzr21SIvhhqB6cm:-Nz5Hmzr21SIvhhqB6cn:b-k1ggz4) ([raw file](https://github.com/sillsdev/web-xforge/blob/ef3607c76d7a10fc8d6bd821773aef6abb854995/src/SIL.XForge.Scripture/ClientApp/src/app/serval-administration/serval-projects.component.ts#L28)):*
> ```XML
> 
>   get alternateSourceId(): string | undefined {
>     const paratextId = this.projectDoc.data?.translateConfig.draftConfig.alternateSource?.paratextId;
> ```

Nit: add a type here.



_Code quote:_
```XML
    const paratextId = this.projectDoc.data?.translateConfig.draftConfig.alternateSource?.paratextId;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/serval-administration/serval-projects.component.ts` line 76 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Nz5HqkU5dat6HoaBjGe:-Nz5HqkU5dat6HoaBjGf:b-k1ggz4) ([raw file](https://github.com/sillsdev/web-xforge/blob/ef3607c76d7a10fc8d6bd821773aef6abb854995/src/SIL.XForge.Scripture/ClientApp/src/app/serval-administration/serval-projects.component.ts#L76)):*
> ```XML
> 
>   get sourceId(): string | undefined {
>     const paratextId = this.projectDoc.data?.translateConfig.source?.paratextId;
> ```

Nit: add a type here.



_Code quote:_
```XML
    const paratextId = this.projectDoc.data?.translateConfig.source?.paratextId;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-menu.service.spec.ts` line 189 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Nz5ISTM2ZZJ_yCiG7Qx:-Nz5ISTM2ZZJ_yCiG7Qy:bx7ldi0) ([raw file](https://github.com/sillsdev/web-xforge/blob/4fa995225c33d665ebb3fb738c16029a8b69b88a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-menu.service.spec.ts#L189)):*
> ```XML
>     });
> 
>     it('should return true only if the user is an administrator or translator', () => {
> ```

Why was this test removed? It seems that the resource editor tab should not affect the canShowHistory feature.



_Code quote:_
```XML
    it('should return true only if the user is an administrator or translator', () => {
```

siltomato (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2457)* status: all files reviewed, 6 unresolved discussions (waiting on @github-advanced-security[bot] and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.spec.ts` line 247 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Nz5GRvr9nJE_CTfC6Rc:-Nz9UvIQ4i-2cUtBJwSC:bbj17v8) ([raw file](https://github.com/sillsdev/web-xforge/blob/ef3607c76d7a10fc8d6bd821773aef6abb854995/src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.spec.ts#L247)):*

This can be made private, unless you think it will be useful for future tests.
</blockquote></details>

I imagine the permissions service will grow, and this method seems pretty useful.  I'll leave it public unless you insist =).
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/serval-administration/serval-projects.component.ts` line 28 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Nz5Hmzr21SIvhhqB6cm:-Nz9RVUkAKIkX4kUcMzq:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/ef3607c76d7a10fc8d6bd821773aef6abb854995/src/SIL.XForge.Scripture/ClientApp/src/app/serval-administration/serval-projects.component.ts#L28)):*

Nit: add a type here.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/serval-administration/serval-projects.component.ts` line 76 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Nz5HqkU5dat6HoaBjGe:-Nz9RUaU9-y44p6n_pfI:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/ef3607c76d7a10fc8d6bd821773aef6abb854995/src/SIL.XForge.Scripture/ClientApp/src/app/serval-administration/serval-projects.component.ts#L76)):*

Nit: add a type here.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-menu.service.spec.ts` line 189 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Nz5ISTM2ZZJ_yCiG7Qx:-Nz9OAi_22rj1KapRGzd:b-hlrg5t) ([raw file](https://github.com/sillsdev/web-xforge/blob/4fa995225c33d665ebb3fb738c16029a8b69b88a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-menu.service.spec.ts#L189)):*

Why was this test removed? It seems that the resource editor tab should not affect the canShowHistory feature.
</blockquote></details>

`service['canShowHistory']` should have been `service['canShowResource']` in this test (copy/paste mistake).

The reason I removed it completely instead of just changing it to call the correct method is because I was thinking that the requirement for showing the resource option in the menu may really be whether the user can sync.  Since this check is delegated to another service and tested there, I figured just making sure that the service method is called would be sufficient.  Am I thinking about this wrong?
___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 493 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2457#gh-1619366796:-Nz9Qj6F8HKSsuXxVGMI:b-rewi74) ([raw file](https://github.com/sillsdev/web-xforge/blob/be81ab9c8a12ffc8dd5f42d994d2ab3c1ecba3e2/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L493)):*
<details><summary><i>Previously, github-advanced-security[bot] wrote…</i></summary><blockquote>

## Nested 'if' statements can be combined

These 'if' statements can be combined.

[Show more details](https://github.com/sillsdev/web-xforge/security/code-scanning/809)
</blockquote></details>

The nested 'if' statements here seem more readable to me.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project-util.service.ts` line 10 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Nz5Grb8A6eUABhDpfnB:-Nz9UDTrAsk0bzBaxQIc:b-dxcbts) ([raw file](https://github.com/sillsdev/web-xforge/blob/ef3607c76d7a10fc8d6bd821773aef6abb854995/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project-util.service.ts#L10)):*

This should be moved into the `paratext.service.ts` file. We want to keep Paratext related functions there and not have it leak into services across our codebase. I know this already existed in the `serval-administration.service`. I would argue it still belongs within `paratext.service`.
</blockquote></details>

Makes sense.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project-util.service.spec.ts` line 3 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Nz5HFDE71ht79BoqTtS:-Nz9UH4q8AVmC66SOr93:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/ef3607c76d7a10fc8d6bd821773aef6abb854995/src/SIL.XForge.Scripture/ClientApp/src/app/core/sf-project-util.service.spec.ts#L3)):*

Then this file can be renamed paratext.service.spec.ts.
</blockquote></details>

Done.

RaymondLuong3 (changes requested):
This is working nicely! One last thing to work out. I think we should notify the user when they are trying to open a Paratext project in the resource tab that some projects cannot be added.

Currently, if an Observer tries to open a project before it is connected, the PT project will get connected to SF. Then when it tries to sync, it fails because the user is not a translator or admin.

My suggestion is to disable projects that are not yet connected when a user is not an admin, and if the project is connected, skip synchronizing it if the user is not an admin or translator. We do not want PT projects to be connected by anyone other than the admin.

> Reviewed 15 of 15 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-:-NzEMl2yBrtvltt9f4vh:b-50tmqh)* status: all files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot])

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.spec.ts` line 247 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Nz5GRvr9nJE_CTfC6Rc:-NzEGI8x6gMhDS3MKv4t:b-la5ssl) ([raw file](https://github.com/sillsdev/web-xforge/blob/ef3607c76d7a10fc8d6bd821773aef6abb854995/src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.spec.ts#L247)):*

I imagine the permissions service will grow, and this method seems pretty useful.  I'll leave it public unless you insist =).
</blockquote></details>

You are probably right. We can leave it as-in.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-menu.service.spec.ts` line 189 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-Nz5ISTM2ZZJ_yCiG7Qx:-NzEGmgR2gl-LVDkxbpm:b-6w3psk) ([raw file](https://github.com/sillsdev/web-xforge/blob/4fa995225c33d665ebb3fb738c16029a8b69b88a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-menu.service.spec.ts#L189)):*

`service['canShowHistory']` should have been `service['canShowResource']` in this test (copy/paste mistake).

The reason I removed it completely instead of just changing it to call the correct method is because I was thinking that the requirement for showing the resource option in the menu may really be whether the user can sync.  Since this check is delegated to another service and tested there, I figured just making sure that the service method is called would be sufficient.  Am I thinking about this wrong?
</blockquote></details>

That sounds good to me.

siltomato (commented):
- Filtered out Paratext projects that either do not have a corresponding SF project or are not connectable.
- Attempt sync after project selection only if user `canSync` according to `PermissionsService`.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-:-NzF_92VAU2_2ch4Mkxz:b95rbcf)* status: 53 of 58 files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot] and @RaymondLuong3)

RaymondLuong3 (approved):
Yes, that looks like what was needed. Thanks for the update. Let's get this to testers.

I did find an issue when I tried to connect to an existing PT project for a project that I was an observer on. This can be addressed as a follow-up, so I'll open a ticket for it. This won't be perfect when it is merged, but the main functionality should be working.


> Reviewed 5 of 5 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-:-NzULKDgD2TLC7uHNvnW:bt6m91e)* status: all files reviewed, 3 unresolved discussions (waiting on @github-advanced-security[bot] and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.spec.ts` line 73 at r10](https://reviewable.io/reviews/sillsdev/web-xforge/2457#-NzUKyj-9zTzUS7guNHe:-NzUKyj-9zTzUS7guNHf:b-ogg1oz) ([raw file](https://github.com/sillsdev/web-xforge/blob/d310f517e089cc74af33f0639d8af031bc866c1c/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.spec.ts#L73)):*
> ```XML
>       ];
>       const projects = [
>         ...expectedResult,
> ```

I really like how you made expected results so clear in this test setup. Nice!



_Code quote:_
```XML
        ...expectedResult,
```

PR #2458, opened 2024-05-15
SF-2708 Add Restore History Button

pmachapman:
This PR adds support for reverting a chapter to a previous version.

It also includes a refactor of the permission checks to see if a user can edit/update a text, and an update to the history and draft tabs to show text in the correct font and size.

**Known Issues**

*These bugs also occur when the document open while it is updated on sync, so are not limited to this feature. Changing chapter or navigating to another page and back fixes these bugs (they are display only not data).*

 * Formatting is temporarily lost when restoring a chapter with different USFM tags.
 * Notes are temporarily not clickable after restoring history.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2458)


Nateowami (changes requested):
Partially reviewed. I moved to user support before I could get further. 

> Reviewed 5 of 19 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-:-NyN5yKF0jyykHhTvlWP:b-zaznhi)* status: 5 of 19 files reviewed, 4 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts` line 53 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-NyMc_gv0u4nGGhA-rRx:-NyMc_gv0u4nGGhA-rRy:b1nlqgq) ([raw file](https://github.com/sillsdev/web-xforge/blob/8a0b738fde9361d857e409bea38ebae1bfe7529a/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts#L53)):*
> ```XML
>    * @returns {boolean} A value indicating whether the chapter can be edited by the current  user.
>    */
>   canEdit(project: SFProjectProfile | undefined, bookNum: number | undefined, chapterNum: number | undefined): boolean {
> ```

I have some concerns about the return type here being `boolean` rather than `boolean | undefined`. I started to write them up, but moved to some user support. I plan to come back to this.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts` line 55 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-NyMb_74BoqB3T0Gs_Uu:-NyMb_74BoqB3T0Gs_Uv:bnzz6ug) ([raw file](https://github.com/sillsdev/web-xforge/blob/8a0b738fde9361d857e409bea38ebae1bfe7529a/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts#L55)):*
> ```XML
>   canEdit(project: SFProjectProfile | undefined, bookNum: number | undefined, chapterNum: number | undefined): boolean {
>     return (
>       this.isUsfmValid(project, bookNum, chapterNum) &&
> ```

One thing to keep in mind: The reason we check if the USFM is "invalid" is because we can't let the users edit the text and guarantee that it will round-trip correctly. But if we're applying a draft, and in so doing *completely replacing* the content of the chapter, then this consideration because a non-issue.

It would be annoying if we didn't allow users to apply a draft to a project because the target project had "invalid" USFM, when that USFM would be completely blown away anyway by the applying of the draft.

Allowing the user to apply the draft in that situation might be a future improvement.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.spec.ts` line 213 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-NyM7l3yC4e4bZmS7xsC:-NyM7l3z8ScOKV5BgB9g:bno0bhr) ([raw file](https://github.com/sillsdev/web-xforge/blob/8a0b738fde9361d857e409bea38ebae1bfe7529a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.spec.ts#L213)):*
> ```XML
> 
>     get revertHistoryButton(): HTMLElement {
>       return this.fixture.nativeElement.querySelectorAll('#revert-history')[0] as HTMLElement;
> ```

- `querySelectorAll` finds all matching elements
- `querySelector` finds the first matching element
- `querySelectorAll(...)[0]` is just a slightly more convoluted and slightly less performant way of doing `querySelector`

Probably the most normal way to do this would be `this.fixture.nativeElement.getElementById('revert-history')`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 106 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-NyM91NjEnaxsxkbNBQL:-NyM91NjEnaxsxkbNBQM:b-l7a0zw) ([raw file](https://github.com/sillsdev/web-xforge/blob/8a0b738fde9361d857e409bea38ebae1bfe7529a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L106)):*
> ```XML
>   getFontFamilyFromProject(projectDoc: SFProjectProfileDoc | SFProjectProfile | undefined): string {
>     if (projectDoc != null && 'data' in projectDoc) {
>       projectDoc = projectDoc.data;
> ```

Can we rename `projectDoc` to `project`? I think it makes more sense to put a document in a variable named `project`, then to have a variable that specifically specifies that it's a doc, but in fact is not.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2458)* status: 4 of 19 files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts` line 53 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-NyMc_gv0u4nGGhA-rRx:-NyNpSNd319e6hc7NNp7:bexl0bj) ([raw file](https://github.com/sillsdev/web-xforge/blob/8a0b738fde9361d857e409bea38ebae1bfe7529a/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts#L53)):*

I have some concerns about the return type here being `boolean` rather than `boolean | undefined`. I started to write them up, but moved to some user support. I plan to come back to this.
</blockquote></details>

I will wait for your comments.

Just so you understand my thinking, my primary concern in the method signatures and logic was compatibility with the methods I was refactoring.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts` line 55 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-NyMb_74BoqB3T0Gs_Uu:-NyNpuZa2CyVCu2XT6Q9:b62slzs) ([raw file](https://github.com/sillsdev/web-xforge/blob/8a0b738fde9361d857e409bea38ebae1bfe7529a/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts#L55)):*

One thing to keep in mind: The reason we check if the USFM is "invalid" is because we can't let the users edit the text and guarantee that it will round-trip correctly. But if we're applying a draft, and in so doing *completely replacing* the content of the chapter, then this consideration because a non-issue.

It would be annoying if we didn't allow users to apply a draft to a project because the target project had "invalid" USFM, when that USFM would be completely blown away anyway by the applying of the draft.

Allowing the user to apply the draft in that situation might be a future improvement.
</blockquote></details>

It is possible to correct corrupt USFM with correct USFM from a draft (or history), but the "invalid USFM" message is displayed based on the `isValid` property of the `Chapter`, rather than the validity of deltas in the current text doc. This means they still can't edit the doc after applying the new draft or snapshot.

Currently the project document can only be updated via the backend, so I think perhaps a new PR (or expand this PR if you think preferable) to enable updating of the `isValid` property on the frontend when an invalid chapter is replace with valid a USFM Delta.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.spec.ts` line 213 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-NyM7l3yC4e4bZmS7xsC:-NyNqr7G5d--S-MaNyD3:boyexb3) ([raw file](https://github.com/sillsdev/web-xforge/blob/8a0b738fde9361d857e409bea38ebae1bfe7529a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.spec.ts#L213)):*

- `querySelectorAll` finds all matching elements
- `querySelector` finds the first matching element
- `querySelectorAll(...)[0]` is just a slightly more convoluted and slightly less performant way of doing `querySelector`

Probably the most normal way to do this would be `this.fixture.nativeElement.getElementById('revert-history')`
</blockquote></details>

Done. Thanks!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts` line 106 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-NyM91NjEnaxsxkbNBQL:-NyNquATBuvajcxz4_Ur:boyexb3) ([raw file](https://github.com/sillsdev/web-xforge/blob/8a0b738fde9361d857e409bea38ebae1bfe7529a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/font.service.ts#L106)):*

Can we rename `projectDoc` to `project`? I think it makes more sense to put a document in a variable named `project`, then to have a variable that specifically specifies that it's a doc, but in fact is not.
</blockquote></details>

Done. Thanks!

nigel-wells (commented):
Looking forward to getting this feature in. 
Here are a few initial observations:
* The latest revision I assume is the current revision (it is ticked in the drop down) - The wouldn't expect restore revision button to be available for this revision
* RTL isn't respected in the history panel
* I picked a revision to restore and my target and my history panels lost their data - refreshing or logging out and back in didn't help :(
* Following on from the above issue, I was able to revert to the most recent version (before the restore attempt above) which brought my data back!
* Restoring appeared to leave notes in place - or, were the notes in my revision but just not showing in the history panel?
* I know smaller devices isn't our target for this tool but the buttons could drop their labels to be just icons which would help. Even at 1024 it isn't great which is an iPad
* Some of my chapters have invalid formatting and the restore button doesn't appear - is it possible to restore if the current state is broken? Can we restore to get ourselves out of bad data if a revision has good data? Is there a check to make sure we're not accidently restoring a version that includes bad data?

> Reviewed 3 of 19 files at r1, 4 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-:-NzBj4Dm-rRqg8sYBPQb:bjplgoo)* status: 7 of 19 files reviewed, 3 unresolved discussions (waiting on @Nateowami)

pmachapman (commented):
Hi @nigel-wells thanks for your comments. My responses are below:

> The latest revision I assume is the current revision (it is ticked in the drop down) - The wouldn't expect restore revision button to be available for this revision

This is a possible scenario, because after the history panel is loaded, the user may edit the target text, or other edits may flow in from a sync or another user editing it. I think the UI consistency of always having the button visible instead of calculating "is the current target different to the open revision" (and associated calculation time) and show/hiding the button is helpful. But yes, 99%+ of the time, people will not be restoring the current revision.

> RTL isn't respected in the history panel

Good catch - thanks.

> I picked a revision to restore and my target and my history panels lost their data - refreshing or logging out and back in didn't help :(

Can you send me details on how to recreate this? I can't seem to make it reoccur for me.

> Following on from the above issue, I was able to revert to the most recent version (before the restore attempt above) which brought my data back!

Cool.

> Restoring appeared to leave notes in place - or, were the notes in my revision but just not showing in the history panel?

Notes are not shown in the history panel, as these are not in the USFM. Footnotes are in the USFM, and are shown in both.

> I know smaller devices isn't our target for this tool but the buttons could drop their labels to be just icons which would help. Even at 1024 it isn't great which is an iPad

I had that initially, but as the icons are very similar, it got quite confusing. Technical showing/hiding was also difficult when I tried to demo it because of the two column and 1 column tab layouts. Also, the ellipsis will show the first few letters of the caption, it helps to differentiate the buttons and encourages hovering to show the tooltip. If you know other ways of really differentiating between the buttons, we can make them icons though.

> Some of my chapters have invalid formatting and the restore button doesn't appear - is it possible to restore if the current state is broken? Can we restore to get ourselves out of bad data if a revision has good data? Is there a check to make sure we're not accidently restoring a version that includes bad data?

Do the chapters have the `isValid: false` property set? If this is the case, reverting is disabled, because that flag cannot be reset on the client side. For the other case, when the chapter is corrupt, it is not possible to revert as a corrupt document cannot be diff'ed. If there is some other case, let me know the scenario and I can see what it will take to allow reverting. Probably the only way I can think of allowing these cases would be backend revert history logic, but I think that should be addressed in a separate issue, as it would also require a rewrite for the apply draft logic.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-:-NzV5_i6-ZVC59tJaJEN:b-aiprzw)* status: 7 of 19 files reviewed, 3 unresolved discussions (waiting on @Nateowami and @nigel-wells)

nigel-wells (commented):
I think I figured out the revert issue losing my data. I had picked the earliest version I had which looked populated due to the diff but in hindsight it was actually blank. My bad :)

The chapters I've looked at do have `isValid: false` set so that will be why. The use cases I can think of is when we (SF) introduce data corruption. There was the cases we worked through the other day that were caused by the apply draft button. There have been numerous copy and paste issues in the past. It would be nice to completely revert back to a previous working state - this was my expectation as a user though I do get why it isn't that simple if we're applying ops they way we do.

The notes may be an interesting one as they may have been positioned differently at a certain point in time from where they are now. Sometime for us to be aware of in case it crops up.

I'll comment further down regarding the buttons - I'm sure we can do some improvements on this though:


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-:-Nz_yUn56dSQ1V-ogF4C:b-fej67a)* status: 5 of 19 files reviewed, 3 unresolved discussions (waiting on @Nateowami)

nigel-wells (changes requested):
Bit of an overhall, but what about this idea? It will work a lot better across different viewports. Have a slider to quickly move around as well as dedicated previous/next buttons for fine tuning. Material includes tickmarks which could be nice to turn on. I'm not 100% if we can style specific ticks but it would be nice to have some stand out if it was a special event i.e. sync, restore, draft applied, etc. This might be better suited for a future PR :) Here is the [Balsamiq link](https://balsamiq.cloud/sghq53/pejrlz8/r689A)


> Reviewed 5 of 19 files at r1, 2 of 2 files at r2, 7 of 7 files at r4.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-:-NzaHuuSE30WDAJaq2np:bezmhzb)* status: all files reviewed, 9 unresolved discussions (waiting on @Nateowami and @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts` line 70 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza0fKhDtt2iU2gm5IU:-Nza0fKhDtt2iU2gm5IV:b-l0jtqt) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts#L70)):*
> ```XML
>    */
>   isDataInSync(project: SFProjectProfile | undefined): boolean {
>     return project?.sync?.dataInSync !== false;
> ```

Can you confirm if this is the same sync check done on the navigation bar as the logic is different?

```
get syncInProgress(): boolean {
    return this.selectedProjectDoc?.data != null && this.selectedProjectDoc.data.sync.queuedCount > 0;
  }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 866 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza3mKKEAon8mcUkVVI:-Nza3mKKEAon8mcUkVVJ:b-x9fmjr) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L866)):*
> ```XML
>   setContents(delta: DeltaStatic, source?: Sources): void {
>     this?.editor?.setContents(delta, source);
>     this.contentSet = true;
> ```

Should `contentSet` still be set to `true` if `editor` is `null|undefined`?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/editor-history.component.ts` line 6 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza3IGg-_Td0BbSJmYd:-Nza3IGg-_Td0BbSJmYe:bmggv59) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/editor-history.component.ts#L6)):*
> ```XML
> import { combineLatest, startWith, tap } from 'rxjs';
> import { OnlineStatusService } from 'xforge-common/online-status.service';
> import { FontService } from '../../../../xforge-common/font.service';
> ```

nit  no need for all the `../`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.html` line 32 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza66KjEUOrLivXNueP:-Nza66KjEUOrLivXNueQ:bgymlqg) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.html#L32)):*
> ```HTML
>       type="button"
>       (click)="revertToSnapshot()"
>       [disabled]="!canRestoreSnapshot"
> ```

Is this state ever hit as the button will only show if this is true
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.ts` line 47 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza4lhvA8VEpZ6IuXdT:-Nza4lhvA8VEpZ6IuXdU:blcc4k1) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.ts#L47)):*
> ```XML
> 
>   historyRevisions: Revision[] = [];
>   selectedRevision: Revision | undefined;
> ```

Any reason for changing this from the optional `selectedRevision?: Revision` ?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.ts` line 151 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza7GMm2sWuckWbbaOg:-Nza7GMm2sWuckWbbaOh:b-fns75s) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.ts#L151)):*
> ```XML
>       !this.textDocService.canEdit(this.projectDoc?.data, this.bookNum, this.chapter)
>     ) {
>       this.noticeService.showError(translate('history_chooser.error'));
> ```

I'd be inclined to do this check before the confirmation that way confirming the dialog means its good to go. Also, are we able to assist in the error as to what failed in your checks above so that they can try and fix the issue themselves rather than contacting support. If not then can we log details to bugsnag so we know what to look at.

nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2458)* status: all files reviewed, 9 unresolved discussions (waiting on @Nateowami and @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.ts` line 151 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza7GMm2sWuckWbbaOg:-NzaJMGe6tofadAScdZw:b-qyg8es) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.ts#L151)):*

I'd be inclined to do this check before the confirmation that way confirming the dialog means its good to go. Also, are we able to assist in the error as to what failed in your checks above so that they can try and fix the issue themselves rather than contacting support. If not then can we log details to bugsnag so we know what to look at.
</blockquote></details>

Could this logic be better used in the `canRestoreSnapshot` getter?

pmachapman (commented):
> Bit of an overhaul, but what about this idea?

That is definitely possible - I had originally envisaged a UI similar to that, but chose the dropdown to keep the UI understandable across a broad range of users. The C# API is designed to allow that kind of "any point in time" restore (the dates in the dropdown are fairly arbitrary, although there is some calculations to ensure they reflect periods when the chapter was edited). I think it warrants further investigation, with my only real request is that we have broad enough usability testing across our userbase.

> This might be better suited for a future PR :)

Yes, please!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-:-NzdKhcO7lFaaG3sYIVi:b-68bx6o)* status: 12 of 19 files reviewed, 8 unresolved discussions (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts` line 70 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza0fKhDtt2iU2gm5IU:-NzdNFxS5zkcgn5J_i4Z:b-4ny1ik) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts#L70)):*

Can you confirm if this is the same sync check done on the navigation bar as the logic is different?

```
get syncInProgress(): boolean {
    return this.selectedProjectDoc?.data != null && this.selectedProjectDoc.data.sync.queuedCount > 0;
  }
```
</blockquote></details>

No this is a different check - the `syncInProgress` is determining if a sync is in progress (for the navigation UI), `isDataInSync` determines if the data is in sync (if the data is out of sync, we do not allow editing to stop problems careening out of control).
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 866 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza3mKKEAon8mcUkVVI:-NzdOHRV6jx73tEkZV97:b4wsm16) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L866)):*

Should `contentSet` still be set to `true` if `editor` is `null|undefined`?
</blockquote></details>

No it shouldn't. Good catch!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/editor-history.component.ts` line 6 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza3IGg-_Td0BbSJmYd:-NzdOiqL7kGs7MPFuuci:b-ioxi27) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/editor-history.component.ts#L6)):*

nit  no need for all the `../`
</blockquote></details>

Done. Thank you!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.html` line 32 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza66KjEUOrLivXNueP:-NzdQTFv-exp1xQX4Zd1:baz7vsb) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.html#L32)):*

Is this state ever hit as the button will only show if this is true
</blockquote></details>

No - good spotting - thanks!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.ts` line 47 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza4lhvA8VEpZ6IuXdT:-NzdSaqa2K_pSGKJCA35:bz7qg08) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.ts#L47)):*

Any reason for changing this from the optional `selectedRevision?: Revision` ?
</blockquote></details>

Just making the type more specific. If you think it is a bad idea, I can revert it.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.ts` line 151 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza7GMm2sWuckWbbaOg:-NzdW51j4gE6wc6Sn_OF:bcfpse1) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.ts#L151)):*
> I'd be inclined to do this check before the confirmation that way confirming the dialog means its good to go.

It is more of a sanity check in case the state changes while the dialog is left open. I don't expect this message to appear under normal use - it is unlikely to appear, but theoretically possible. I could not trigger the conditions to display it via the frontend (I am only able to trigger it via a unit test).

> Could this logic be better used in the `canRestoreSnapshot` getter?

I have updated `canRestoreSnapshot` to use the same core logic, but annoyingly the type guards wouldn't let swap out any of the null checks with just `!this.canRestoreSnapshot`, so I have left them in.

nigel-wells (approved):
:lgtm:

> Reviewed 7 of 7 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-:-NzekjyW0OrPrZ5q9tRQ:bnfp4nl)* status: all files reviewed, 5 unresolved discussions (waiting on @Nateowami and @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.ts` line 47 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza4lhvA8VEpZ6IuXdT:-Nzehpy8B8jb_hL7jJVw:b-f8na49) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.ts#L47)):*

Just making the type more specific. If you think it is a bad idea, I can revert it.
</blockquote></details>

I wouldn't say a bad idea, it's just more to type out for almost the same result :)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.ts` line 151 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza7GMm2sWuckWbbaOg:-NzeiVi25pQ1e9JuF6CS:bw4w96) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.ts#L151)):*

> I'd be inclined to do this check before the confirmation that way confirming the dialog means its good to go.

It is more of a sanity check in case the state changes while the dialog is left open. I don't expect this message to appear under normal use - it is unlikely to appear, but theoretically possible. I could not trigger the conditions to display it via the frontend (I am only able to trigger it via a unit test).

> Could this logic be better used in the `canRestoreSnapshot` getter?

I have updated `canRestoreSnapshot` to use the same core logic, but annoyingly the type guards wouldn't let swap out any of the null checks with just `!this.canRestoreSnapshot`, so I have left them in.
</blockquote></details>

Thanks - those type guards annoy me as well when I know for sure something is set from another previously run method.

nigel-wells (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2458)* status: all files reviewed, 5 unresolved discussions (waiting on @Nateowami and @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts` line 70 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-Nza0fKhDtt2iU2gm5IU:-Nzeknxz4bY6_09wlFny:b-4gor3o) ([raw file](https://github.com/sillsdev/web-xforge/blob/3e3d76700aaf042e243d535e34236a40ee26af79/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts#L70)):*

No this is a different check - the `syncInProgress` is determining if a sync is in progress (for the navigation UI), `isDataInSync` determines if the data is in sync (if the data is out of sync, we do not allow editing to stop problems careening out of control).
</blockquote></details>

Gotcha!

nigel-wells:
Sorry, just changed this back to review as I see there is still a blocking comment from @Nateowami

Nateowami (commented):
> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2458)* status: all files reviewed, 4 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts` line 53 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-NyMc_gv0u4nGGhA-rRx:-Nzi5RsC40aqWMPi9Epf:b-tsm5ln) ([raw file](https://github.com/sillsdev/web-xforge/blob/8a0b738fde9361d857e409bea38ebae1bfe7529a/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts#L53)):*

I will wait for your comments.

Just so you understand my thinking, my primary concern in the method signatures and logic was compatibility with the methods I was refactoring.
</blockquote></details>

Let's just leave this as-is. We can discuss this kind of thing later.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts` line 55 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2458#-NyMb_74BoqB3T0Gs_Uu:-Nzi5pVzB38mewj16axi:b-h7i7zh) ([raw file](https://github.com/sillsdev/web-xforge/blob/8a0b738fde9361d857e409bea38ebae1bfe7529a/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts#L55)):*

It is possible to correct corrupt USFM with correct USFM from a draft (or history), but the "invalid USFM" message is displayed based on the `isValid` property of the `Chapter`, rather than the validity of deltas in the current text doc. This means they still can't edit the doc after applying the new draft or snapshot.

Currently the project document can only be updated via the backend, so I think perhaps a new PR (or expand this PR if you think preferable) to enable updating of the `isValid` property on the frontend when an invalid chapter is replace with valid a USFM Delta.
</blockquote></details>

Good point; was not thinking of that.

PR #2470, opened 2024-05-22
Fix crash when paging in Serval Administration area

pmachapman:
On production, due to the very large number of projects, paging to projects > 300 in the Serval Administration area or System Administration projects tab can result in the following crash:

image

The query parameters that cause this query to crash are in `ServalProjectsComponent.getQueryParameters()`.

To resolve this, an index is added to the sf_projects name field to enable to the sort operation in the query used in these Administration areas to use an index. See https://www.mongodb.com/docs/manual/tutorial/sort-results-with-indexes/

Because this only occurs on production, it is not testably locally or on QA, although my local tests confirmed that the index is used when paging through projects in the Serval Administration area.

This error is reproducible in MongoDB compass, so is not caused by a restriction of the Realtime Server or MongoDB Node driver:

image

When the backend is run, the following index is created in `sf_projects`:

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2470)

### DIFF
--- a/src/RealtimeServer/scriptureforge/models/sf-project.ts
+++ b/src/RealtimeServer/scriptureforge/models/sf-project.ts
@@ -1,5 +1,6 @@
 import { Project } from '../../common/models/project';
 import { WritingSystem } from '../../common/models/writing-system';
+import { obj } from '../../common/utils/obj-path';
 import { BiblicalTermsConfig } from './biblical-terms-config';
 import { CheckingConfig } from './checking-config';
 import { NoteTag } from './note-tag';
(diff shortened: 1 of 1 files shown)


Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2470#-:-NyXD7R4Dh5pHxosfZj8:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

Nateowami (approved):
Thanks for explaining everything in detail.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2470#-:-NyXDD4c1a6kb_Tf43OZ:b-ynxw65)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #777, opened 2020-08-25
Migrator: PtdaSyncAll: Specific admin, Reword, Inspect doesn't sync

marksvc:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/777)

PR #2472, opened 2024-05-22
SF-2365 Upgrade from legacy list

josephmyers:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -580,7 +580,7 @@ class TestEnvironment {
   }
 
   get menuListItems(): DebugElement[] {
-    return this.fixture.debugElement.queryAll(By.css('#menu-drawer .mat-list-item'));
+    return this.fixture.debugElement.queryAll(By.css('#menu-drawer .mdc-list-item'));
   }
 
   get userMenu(): DebugElement {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.html
@@ -154,6 +154,7 @@
                   <mat-selection-list [multiple]="false">
                     <mat-list-item
                       *ngFor="let questionDoc of getQuestionDocs(getTextDocIdType(text.bookNum, chapter.number))"
+                      disableRipple
                     >
                       <div fxFlex="grow" fxLayout="row" fxLayoutAlign="start center">
                         <button mat-icon-button (click)="questionDialog(questionDoc)" class="edit-btn">
@@ -338,6 +339,7 @@
                 <mat-selection-list [multiple]="false">
                   <mat-list-item
                     *ngFor="let questionDoc of getQuestionDocs(getTextDocIdType(text.bookNum, chapter.number), true)"
+                    disableRipple
                   >
                     <div fxFlex="grow" fxLayout="row" fxLayoutAlign="start center">
                       v{{ questionDoc.data?.verseRef?.verseNum }} -&nbsp;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.scss
@@ -29,7 +29,7 @@
   }
 }
 
-.mat-list-base {
+.mat-mdc-list-base {
   padding-block-start: 0;
 }
 
@@ -38,7 +38,8 @@ app-donut-chart {
   height: 36px;
 }
 
-.mat-list-item {
+.mat-mdc-list-item {
+  cursor: auto;
   .audio-icon {
     margin-right: 5px;
   }
@@ -49,9 +50,14 @@ app-donut-chart {
 }
 
 #reviewer-questions-list {
-  .mat-list-item {
+  .mat-mdc-list-item {
     cursor: pointer;
+    height: 50px;
     margin-inline-start: 16px;
+  }
+  .mat-mdc-list-item:hover {
+    background: var(--mat-expansion-header-hover-state-layer-color);
+    border-radius: 4px;
   }
 }
 
@@ -61,9 +67,13 @@ app-donut-chart {
   overflow: hidden;
 }
 
+.mat-expansion-panel-header {
+  --mat-expansion-header-expanded-state-height: var(--mat-expansion-header-collapsed-state-height);
+}
+
 .book-expander ::ng-deep .mat-expansion-panel-content .mat-expansion-panel-body {
   padding-inline-end: 0 !important;
-  padding-block-start: 0 !important;
+  padding-block: 0 !important;
 }
 
 #no-questions-label,
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.scss
@@ -50,13 +50,13 @@ mat-icon {
   font-style: italic;
 }
 
-.mat-list {
+.mat-mdc-list {
   width: 100%;
   display: flex;
   flex-direction: column;
   padding: 0;
 
-  .mat-list-item {
+  .mat-mdc-list-item {
     cursor: pointer;
     height: auto;
 
@@ -99,6 +99,10 @@ mat-icon {
     &.selected {
       background-color: vars.$questionSelectedColor;
       color: #fff;
+
+      ::ng-deep .mdc-list-item__primary-text {
+        color: unset;
+      }
 
       .view-answers {
         mat-icon {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts
@@ -12,13 +12,13 @@ import {
   SimpleChanges,
   ViewChildren
 } from '@angular/core';
-import { MatLegacyListItem as MatListItem } from '@angular/material/legacy-list';
+import { MatListItem } from '@angular/material/list';
 import sortBy from 'lodash-es/sortBy';
 import { Operation } from 'realtime-server/lib/esm/common/models/project-rights';
 import { Answer } from 'realtime-server/lib/esm/scriptureforge/models/answer';
 import { Comment } from 'realtime-server/lib/esm/scriptureforge/models/comment';
 import { SFProjectProfile } from 'realtime-server/lib/esm/scriptureforge/models/sf-project';
-import { SFProjectDomain, SF_PROJECT_RIGHTS } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-rights';
+import { SF_PROJECT_RIGHTS, SFProjectDomain } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-rights';
 import { SFProjectUserConfig } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-user-config';
 import { toVerseRef, VerseRefData } from 'realtime-server/lib/esm/scriptureforge/models/verse-ref-data';
 import { Subject, Subscription } from 'rxjs';
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.spec.ts
@@ -2486,7 +2486,7 @@ class TestEnvironment {
   }
 
   get questions(): DebugElement[] {
-    return this.fixture.debugElement.queryAll(By.css('app-checking-questions .mat-list-item'));
+    return this.fixture.debugElement.queryAll(By.css('app-checking-questions .mdc-list-item'));
   }
 
   get quillEditor(): HTMLElement {
@@ -2782,7 +2782,7 @@ class TestEnvironment {
 
   selectQuestion(/** indexed starting at 1 */ questionNumber: number, includeReadTimer: boolean = true): DebugElement {
     const question = this.fixture.debugElement.query(
-      By.css('app-checking-questions .mat-list-item:nth-child(' + questionNumber + ')')
+      By.css('app-checking-questions .mdc-list-item:nth-child(' + questionNumber + ')')
     );
     question.nativeElement.click();
     tick(1);
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.html
@@ -4,7 +4,7 @@
       <mat-list-item disableRipple class="navigation-header">
         {{ t("translate") }}
       </mat-list-item>
-      <a mat-list-item class="mat-list-item" [appRouterLink]="getProjectLink('translate')">
+      <a mat-list-item [appRouterLink]="getProjectLink('translate')">
         <mat-icon mat-list-icon>apps</mat-icon>
         {{ t("overview") }}
       </a>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.html
@@ -4,7 +4,7 @@
       <mat-list-item disableRipple class="navigation-header">
         {{ t("translate") }}
       </mat-list-item>
-      <a mat-list-item [appRouterLink]="getProjectLink('translate')">
+      <a mat-list-item class="mat-list-item" [appRouterLink]="getProjectLink('translate')">
         <mat-icon mat-list-icon>apps</mat-icon>
         {{ t("overview") }}
       </a>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.html
@@ -5,15 +5,15 @@
         {{ t("translate") }}
       </mat-list-item>
       <a mat-list-item [appRouterLink]="getProjectLink('translate')">
-        <mat-icon mat-list-icon>apps</mat-icon>
+        <mat-icon>apps</mat-icon>
         {{ t("overview") }}
       </a>
       <a mat-list-item [appRouterLink]="(translateLink$ | async) || ['']" [class.active]="draftReviewActive">
-        <mat-icon mat-list-icon>edit_note</mat-icon>
+        <mat-icon>edit_note</mat-icon>
         {{ t("edit_review") }}
       </a>
       <a *ngIf="canGenerateDraft$ | async" mat-list-item [appRouterLink]="draftGenerationLink">
-        <mat-icon mat-list-icon class="mirror-rtl">auto_awesome</mat-icon>
+        <mat-icon class="mirror-rtl">auto_awesome</mat-icon>
         {{ t("generate_draft") }} <span class="nav-label">{{ t("beta") }}</span>
       </a>
     </ng-container>
@@ -22,15 +22,15 @@
         {{ t("community_checking") }}
       </mat-list-item>
       <a *ngIf="canManageQuestions" mat-list-item [appRouterLink]="getProjectLink('checking')">
-        <mat-icon mat-list-icon class="mirror-rtl">list</mat-icon>
+        <mat-icon class="mirror-rtl">list</mat-icon>
         {{ t("manage_questions") }}
       </a>
       <a *ngIf="!canManageQuestions" mat-list-item [appRouterLink]="getProjectLink('checking')">
-        <mat-icon mat-list-icon class="mirror-rtl">bar_chart</mat-icon>
+        <mat-icon class="mirror-rtl">bar_chart</mat-icon>
         {{ t("my_progress") }}
       </a>
       <a mat-list-item [appRouterLink]="(answerQuestionsLink$ | async) ?? ['']" [class.active]="answerQuestionsActive">
-        <mat-icon mat-list-icon class="mirror-rtl">question_answer</mat-icon>
+        <mat-icon class="mirror-rtl">question_answer</mat-icon>
         {{ t("questions_answers") }}
       </a>
     </ng-container>
@@ -45,7 +45,6 @@
     <a *ngIf="canSync$ | async" mat-list-item [appRouterLink]="getProjectLink('sync')">
       <mat-icon
         aria-hidden="false"
-        mat-list-icon
         matBadge="error"
         [matBadgeHidden]="!lastSyncFailed || syncInProgress"
         [class.sync-in-progress]="syncInProgress && !featureFlags.stillness.enabled"
@@ -55,11 +54,11 @@
       {{ t("synchronize") }}
     </a>
     <a *ngIf="canSeeUsers$ | async" mat-list-item [appRouterLink]="getProjectLink('users')">
-      <mat-icon mat-list-icon class="mirror-rtl">people</mat-icon>
+      <mat-icon class="mirror-rtl">people</mat-icon>
       {{ t("users") }}
     </a>
     <a *ngIf="canSeeSettings$ | async" mat-list-item [appRouterLink]="getProjectLink('settings')">
-      <mat-icon mat-list-icon>settings</mat-icon>
+      <mat-icon>settings</mat-icon>
       {{ t("settings") }}
     </a>
   </mat-nav-list>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss
@@ -85,10 +85,10 @@ mat-nav-list .mdc-list-item {
   margin-top: 2em;
 }
 
-.mat-nav-list.disabled-offline .mdc-list-item {
+.mat-mdc-nav-list.disabled-offline ::ng-deep .mat-mdc-list-item-unscoped-content {
   &,
   mat-icon {
-    color: variables.$disabled-text-color;
+    color: variables.$disabled-text-color !important;
   }
 }
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss
@@ -9,8 +9,7 @@ mat-nav-list .mdc-list-item {
     fill: variables.$purpleMedium;
     transition: color 0.15s;
     padding: 4px;
-    height: unset;
-    width: unset;
+    box-sizing: content-box;
   }
   &:hover {
     mat-icon {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss
@@ -94,7 +94,8 @@ mat-nav-list .mdc-list-item {
 
 .nav-label {
   color: white;
-  padding: 0px 6px;
+  padding: 2px 6px;
+  line-height: 1rem;
   border-radius: 100px;
   font-size: 0.95em;
   background: #4678d6;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss
@@ -3,6 +3,7 @@
 mat-nav-list .mdc-list-item {
   font-weight: 500;
   font-size: 14px;
+  padding-inline: 12px;
 
   mat-icon {
     color: variables.$purpleMedium;
@@ -23,9 +24,9 @@ mat-nav-list .mdc-list-item {
     display: flex;
     align-items: center;
     column-gap: 12px;
-    font-size: unset;
-    font-weight: unset;
-    letter-spacing: unset;
+    font-size: inherit;
+    font-weight: inherit;
+    letter-spacing: normal;
   }
 
   &.navigation-header ::ng-deep .mat-mdc-list-item-unscoped-content {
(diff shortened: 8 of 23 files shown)


pmachapman (changes requested):
Are you able to define the Acceptance Test in the JIRA card?

Also, here are some graphical errors I found while testing:

**Navigation Menu**


**Suggestions Popup**


**Share Dialog**


**Checking Overview - Questions**


**Checking Overview - Archived Questions**


> Reviewed 17 of 18 files at r1, 4 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-:-NyXBZoz8RIARBla8mAS:bourbe2)* status: all files reviewed, 3 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions.component.scss` line 66 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NyXQ_fI8itkIGHRXRF-:-NyXQ_fI8itkIGHRXRF0:bju5ii0) ([raw file](https://github.com/sillsdev/web-xforge/blob/9bddf2f34e65aa051f6f7cbe628719981fbd5725/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions.component.scss#L66)):*
> ```SCSS
>   justify-content: space-between;
> 
>   font-size: 14px;
> ```

This should be 12px to match the previous style - unless you think it really needs to be bigger (see screenshot)



_Code quote:_
```SCSS
font-size: 14px;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.html` line 28 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NyXIHuj4UfKkomkezTj:-NyXIHuj4UfKkomkezTk:bxudf9d) ([raw file](https://github.com/sillsdev/web-xforge/blob/9bddf2f34e65aa051f6f7cbe628719981fbd5725/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.html#L28)):*
> ```HTML
>           <mat-icon matListItemIcon class="mirror-rtl">book</mat-icon>
>           <span matListItemTitle>{{ getBookName(textProgress.text) }}</span>
>           <span matListItemLine>{{
> ```

BTW It is great how you have restored the light grey for the segment counts. I'm sure it used to do that, but got lost along the way.



_Code quote:_
```HTML
 <span matListItemLine>
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/ui-common.module.ts` line 22 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NyXDGOPDgpUbNXU-9Qi:-NyXDGOPDgpUbNXU-9Qj:b-ltnu62) ([raw file](https://github.com/sillsdev/web-xforge/blob/9bddf2f34e65aa051f6f7cbe628719981fbd5725/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/ui-common.module.ts#L22)):*
> ```XML
> import { MatMenuModule } from '@angular/material/menu';
> import { MatPaginatorIntl, MatPaginatorModule } from '@angular/material/paginator';
> import { MatListModule } from '@angular/material/list';
> ```

When you rebase this, are you able to re-save the file, which should sort these imports?



_Code quote:_
```XML
import { MatListModule } from '@angular/material/list';
```

Nateowami:
@josephmyers Please look at the visual changes Chromatic shows. Some of the problems Peter found are there as well. It doesn't provide us anywhere close to full coverage, but it's a good start. And keep in mind just because something has changed doesn't mean it's wrong and needs to be changed. We're just trying to avoid making anything worse.

josephmyers (commented):
Thanks for the review, Peter. I think I was trying to get this in before EMDC, so I appreciate you combing through things. I fixed all items from the screenshots, except for the last two.

The former is intentional. The hover state for individual questions adds contrast, which will help users position themselves on the page, and it balances out the headers, which are clickable but don't have a hover state when expanded.

The latter I couldn't reproduce. I'm not sure how you got the question so small while maintaining the Archived date text. Mine smartly hides this text when shrinking the window, and I can't find a size where there's overlap.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-:-NzRpQhkDFnJ97zy1b77:bxxmvbz)* status: 14 of 21 files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions.component.scss` line 66 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NyXQ_fI8itkIGHRXRF-:-NzS09f25ofcfv_nWHtZ:b-vahccc) ([raw file](https://github.com/sillsdev/web-xforge/blob/9bddf2f34e65aa051f6f7cbe628719981fbd5725/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions.component.scss#L66)):*

This should be 12px to match the previous style - unless you think it really needs to be bigger (see screenshot)
</blockquote></details>

Done. I thought the 14px was easier to read, but they can adjust the page zoom if they want. Also, the 12px looks sleeker
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/ui-common.module.ts` line 22 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NyXDGOPDgpUbNXU-9Qi:-NzS2H1v1AuTNeCN7fDa:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/9bddf2f34e65aa051f6f7cbe628719981fbd5725/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/ui-common.module.ts#L22)):*

When you rebase this, are you able to re-save the file, which should sort these imports?
</blockquote></details>

Done.

pmachapman (approved):
> The former is intentional. The hover state for individual questions adds contrast, which will help users position themselves on the page, and it balances out the headers, which are clickable but don't have a hover state when expanded.

I think its a lil confusing having hover without a corresponding click capability, but will approve anyway, as this is always something we can revisit if we received feedback that users are confused.

> The latter I couldn't reproduce. I'm not sure how you got the question so small while maintaining the Archived date text. Mine smartly hides this text when shrinking the window, and I can't find a size where there's overlap.

I can't recreate it either. Let's mark it as fixed!

> Reviewed 7 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-:-NzUndCsAGxR2qDg8vfW:b-jjltae)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472)* status: all files reviewed, 3 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss` line 12 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NzVFGWv6wDvcr639Nm9:-NzVFGWv6wDvcr639NmA:b-usv0tv) ([raw file](https://github.com/sillsdev/web-xforge/blob/3072d6772794ddf030106d82cccf7513dd3f67de/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss#L12)):*
MDN explains the meaning of `unset`:
> The **`unset`** CSS keyword resets a property to its inherited value if the property naturally inherits from its parent, and to its [initial value](https://developer.mozilla.org/en-US/docs/Web/CSS/initial_value) if not. In other words, it behaves like the [`inherit`](https://developer.mozilla.org/en-US/docs/Web/CSS/inherit) keyword in the first case, when the property is an [inherited property](https://developer.mozilla.org/en-US/docs/Web/CSS/Inheritance#inherited_properties), and like the [`initial`](https://developer.mozilla.org/en-US/docs/Web/CSS/initial) keyword in the second case, when the property is a [non-inherited property](https://developer.mozilla.org/en-US/docs/Web/CSS/Inheritance#non-inherited_properties).

In order to know what a value set to `unset` will do, you need to:
1. Know that this is the meaning of the `unset` keyword
2. Know whether the property is an inherited property
3. If it's not an inherited property (as in this case), what the initial value is (`auto` in this case)

It's a lot easier to read a CSS rule if you spell out the value you intend. In this case, that would be `height: auto`. Of course, in some cases you want a value to be inherited, in which case, you could use the `inherit` keyword, but 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss` line 25 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NzVB7fC8S5tST3V-b3r:-NzVB7fC8S5tST3V-b3s:bmitjq9) ([raw file](https://github.com/sillsdev/web-xforge/blob/4f6809695438662f0c127248e0e0ca193bd29963/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss#L25)):*
> ```SCSS
> 
>   ::ng-deep .mat-list-item-content {
>     padding-inline: 12px;
> ```

This amount of padding was deliberately chosen as a good balance. I don't think removing it is a good idea. It reduces the horizontal space for each navigation item by 8px, which significantly increases the chance that a localized version will have the text wrap. 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss` line 27 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NzVC58gE3m_u_45FqZI:-NzVC58gE3m_u_45FqZJ:banmsvw) ([raw file](https://github.com/sillsdev/web-xforge/blob/3072d6772794ddf030106d82cccf7513dd3f67de/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss#L27)):*
> ```SCSS
>     align-items: center;
>     column-gap: 12px;
>     font-size: unset;
> ```

[I wrote this comment before the one above, so if it seems redundant that's why]

Setting the font size to "unset" feels really odd to me. There's nothing incorrect about this, but it's the less direct way of saying "just inherit the value of the parent element" while not directly saying "inherit", which seems like the more obvious thing to write. 

Similarly with the font-weight, when I initially ready `font-weight: unset;`, I thought "something is making the font weight be bold, and this is setting it back to its default value." But actually it's a round-about way of setting it to inherit, which is making it bold. Setting it to "inherit" wouldn't make it obvious that this is making it bold, rather than un-bolding it, but it would be clearer.

I think the following is much clearer:
``` scss
    font-size: inherit;
    font-weight: inherit;
    letter-spacing: normal;
```

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472)* status: all files reviewed, 3 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss` line 12 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NzVFGWv6wDvcr639Nm9:-Nza0_chD8mWWD0fRBy7:bjh0qam) ([raw file](https://github.com/sillsdev/web-xforge/blob/09accdfb4d1e11defb2ca3044a488b75ee2d8f8a/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss#L12)):*

MDN explains the meaning of `unset`:
> The **`unset`** CSS keyword resets a property to its inherited value if the property naturally inherits from its parent, and to its [initial value](https://developer.mozilla.org/en-US/docs/Web/CSS/initial_value) if not. In other words, it behaves like the [`inherit`](https://developer.mozilla.org/en-US/docs/Web/CSS/inherit) keyword in the first case, when the property is an [inherited property](https://developer.mozilla.org/en-US/docs/Web/CSS/Inheritance#inherited_properties), and like the [`initial`](https://developer.mozilla.org/en-US/docs/Web/CSS/initial) keyword in the second case, when the property is a [non-inherited property](https://developer.mozilla.org/en-US/docs/Web/CSS/Inheritance#non-inherited_properties).

In order to know what a value set to `unset` will do, you need to:
1. Know that this is the meaning of the `unset` keyword
2. Know whether the property is an inherited property
3. If it's not an inherited property (as in this case), what the initial value is (`auto` in this case)

It's a lot easier to read a CSS rule if you spell out the value you intend. In this case, that would be `height: auto`. Of course, in some cases you want a value to be inherited, in which case, you could use the `inherit` keyword, but
</blockquote></details>

Actually I found the real root of the problem. It wasn't that the height and width had been changed in this new version. The `box-sizing` property changed from `content-box` to `border-box`, because the mat-list-icon isn't present/applying anymore. I didn't see a replacement for this, so I struck it from those elements and set the property myself. Even `matListItemIcon` makes things worse initially, and likely would require more involved restructuring to utilize. Even if we did that, we'd still have to either set the `box-sizing` or the `height/width`, so it doesn't seem worth it.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss` line 25 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NzVB7fC8S5tST3V-b3r:-Nza8vk52kRkp3bFQh2G:b-24k2g4) ([raw file](https://github.com/sillsdev/web-xforge/blob/174f33e6154b81a42c0059f5eeeb0dff855cd202/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss#L25)):*

This amount of padding was deliberately chosen as a good balance. I don't think removing it is a good idea. It reduces the horizontal space for each navigation item by 8px, which significantly increases the chance that a localized version will have the text wrap.
</blockquote></details>

Actually I had to remove it, because it was being doubly applied, once here and once at the anchor. To achieve the original padding, I'll override the mdc-list-item setting
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss` line 27 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NzVC58gE3m_u_45FqZI:-NzaC_xn5m7Los8_lUBb:b-uz0c68) ([raw file](https://github.com/sillsdev/web-xforge/blob/09accdfb4d1e11defb2ca3044a488b75ee2d8f8a/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss#L27)):*

[I wrote this comment before the one above, so if it seems redundant that's why]

Setting the font size to "unset" feels really odd to me. There's nothing incorrect about this, but it's the less direct way of saying "just inherit the value of the parent element" while not directly saying "inherit", which seems like the more obvious thing to write. 

Similarly with the font-weight, when I initially ready `font-weight: unset;`, I thought "something is making the font weight be bold, and this is setting it back to its default value." But actually it's a round-about way of setting it to inherit, which is making it bold. Setting it to "inherit" wouldn't make it obvious that this is making it bold, rather than un-bolding it, but it would be clearer.

I think the following is much clearer:
``` scss
    font-size: inherit;
    font-weight: inherit;
    letter-spacing: normal;
```
</blockquote></details>

Done. My intent was to indicate, "Unset the value set just before," but after seeing the documentation, I think I wasn't understanding `unset` perfectly.

Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472)* status: 18 of 21 files reviewed, 3 unresolved discussions (waiting on @josephmyers and @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss` line 12 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NzVFGWv6wDvcr639Nm9:-Nzahh-8-VzEBkDLvMgl:b-dg5zcn) ([raw file](https://github.com/sillsdev/web-xforge/blob/09accdfb4d1e11defb2ca3044a488b75ee2d8f8a/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss#L12)):*

Actually I found the real root of the problem. It wasn't that the height and width had been changed in this new version. The `box-sizing` property changed from `content-box` to `border-box`, because the mat-list-icon isn't present/applying anymore. I didn't see a replacement for this, so I struck it from those elements and set the property myself. Even `matListItemIcon` makes things worse initially, and likely would require more involved restructuring to utilize. Even if we did that, we'd still have to either set the `box-sizing` or the `height/width`, so it doesn't seem worth it.
</blockquote></details>

Good catch seeing it was about box sizing. I prefer to *always* use `border-box`, so I'd prefer a solution that didn't change it, just because having a different box sizing is an easy source of confusion. But it's not wrong.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss` line 25 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NzVB7fC8S5tST3V-b3r:-Nzai-tgDiXbE8DPVbdK:b44gu5s) ([raw file](https://github.com/sillsdev/web-xforge/blob/174f33e6154b81a42c0059f5eeeb0dff855cd202/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss#L25)):*

Actually I had to remove it, because it was being doubly applied, once here and once at the anchor. To achieve the original padding, I'll override the mdc-list-item setting
</blockquote></details>

OK. You can easily see how it was before by opening QA in one tab, and dev in another, and switching between them. Any time something looks off, you can then use dev tools to find out why.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss` line 27 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NzVC58gE3m_u_45FqZI:-Nzab0FGAtG0hA-QkT5c:bartbvb) ([raw file](https://github.com/sillsdev/web-xforge/blob/09accdfb4d1e11defb2ca3044a488b75ee2d8f8a/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss#L27)):*

Done. My intent was to indicate, "Unset the value set just before," but after seeing the documentation, I think I wasn't understanding `unset` perfectly.
</blockquote></details>

Yeah, CSS doesn't really let you unset something, only override something.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss` line 35 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-Nzabrec35FoOdwtL3cM:-Nzabrec35FoOdwtL3cN:b7x7gs2) ([raw file](https://github.com/sillsdev/web-xforge/blob/e7d0a2acc1745fa695f7112e63102bbae1d834ea/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss#L35)):*
> ```SCSS
>     font-size: 0.95em;
>     color: #777;
>     padding-inline-start: 2px;
> ```

The 18px that was used previously was also carefully chosen to make the headings align with the icons. This is targeting a different element now, so if you change it to 6px it should be pretty well aligned.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions.component.scss` line 56 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NzaepuL2dgRke4qKdIH:-NzaepuL2dgRke4qKdII:bpcaqj) ([raw file](https://github.com/sillsdev/web-xforge/blob/e7d0a2acc1745fa695f7112e63102bbae1d834ea/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions.component.scss#L56)):*
> ```SCSS
> 
> ::ng-deep .mdc-list-item__content {
>   overflow: visible;
> ```

There are two problems with this:
1. `::ng-deep` should not be used as a top-level selector, since doing so makes the style global. If you use `:host ::ng-deep .mdc-list-item__content`, then it will apply only to children of this component.
2. This change is not working, because Material is setting `overflow: hidden; text-overflow: ellipsis;` on all `.mdc-list-item__content`. However, conveniently, if you change it to `:host ::ng-deep .mdc-list-item__content`, that increases the specificity enough to take precedence over the Material rule.

This gives me some concern that the hiding of overflow in list items could be an issue elsewhere, but the only places I could find where it made a difference was on the translate overview page, and this PR actually fixes it.

(Sorry; I know you made this exact code change at my suggestion. The rule is correct, but I had only applied it via dev tools and hadn't noticed that something else would override it).

To see the overflow issues clearly, I suggest running storybook, opening the RTL story for the suggestions component, opening the canvas in a new tab, and then zooming in all the way and opening dev tools. It gives you very clear insight into what is going on, and how much is getting cut off:

Nateowami (commented):
I forgot to mention in my review that I would prefer for you to leave Chromatic for a reviewer to approve, just like with code changes. Though I do suggest looking through it carefully to make sure nothing unintended slipped in.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-:-NzainZF7DFXfGf3yopt:b2v88d4)* status: 18 of 21 files reviewed, 3 unresolved discussions (waiting on @josephmyers and @pmachapman)

josephmyers:
> I forgot to mention in my review that I would prefer for you to leave Chromatic for a reviewer to approve, just like with code changes. Though I do suggest looking through it carefully to make sure nothing unintended slipped in.
> 
> > _[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-:-NzainZF7DFXfGf3yopt:b2v88d4)_ status: 18 of 21 files reviewed, 3 unresolved discussions (waiting on @josephmyers and @pmachapman)

Oh right. I had gotten confused and thought there was a check for me (optional) and a check for a reviewer (required).

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472)* status: 18 of 21 files reviewed, 2 unresolved discussions (waiting on @Nateowami and @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss` line 25 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NzVB7fC8S5tST3V-b3r:-Nzeu_pL7qAVWx2W1OMN:b-af183f) ([raw file](https://github.com/sillsdev/web-xforge/blob/174f33e6154b81a42c0059f5eeeb0dff855cd202/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss#L25)):*

OK. You can easily see how it was before by opening QA in one tab, and dev in another, and switching between them. Any time something looks off, you can then use dev tools to find out why.
</blockquote></details>

Yep, thanks!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss` line 35 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-Nzabrec35FoOdwtL3cM:-NzevRsx78enocff4Bos:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/e7d0a2acc1745fa695f7112e63102bbae1d834ea/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss#L35)):*

The 18px that was used previously was also carefully chosen to make the headings align with the icons. This is targeting a different element now, so if you change it to 6px it should be pretty well aligned.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions.component.scss` line 56 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-NzaepuL2dgRke4qKdIH:-NzfZMAeBD9CMWgoPtP_:b-5ylzh1) ([raw file](https://github.com/sillsdev/web-xforge/blob/e7d0a2acc1745fa695f7112e63102bbae1d834ea/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/suggestions.component.scss#L56)):*

There are two problems with this:
1. `::ng-deep` should not be used as a top-level selector, since doing so makes the style global. If you use `:host ::ng-deep .mdc-list-item__content`, then it will apply only to children of this component.
2. This change is not working, because Material is setting `overflow: hidden; text-overflow: ellipsis;` on all `.mdc-list-item__content`. However, conveniently, if you change it to `:host ::ng-deep .mdc-list-item__content`, that increases the specificity enough to take precedence over the Material rule.

This gives me some concern that the hiding of overflow in list items could be an issue elsewhere, but the only places I could find where it made a difference was on the translate overview page, and this PR actually fixes it.

(Sorry; I know you made this exact code change at my suggestion. The rule is correct, but I had only applied it via dev tools and hadn't noticed that something else would override it).

To see the overflow issues clearly, I suggest running storybook, opening the RTL story for the suggestions component, opening the canvas in a new tab, and then zooming in all the way and opening dev tools. It gives you very clear insight into what is going on, and how much is getting cut off:


</blockquote></details>

Dang! I thought I saw that mine was on top of their rule. Thanks for catching this. Storybook helped with the verification.

Nateowami:
I've rebased, and for some reason that shortened the beta label. I've adjusted its size a bit to make it taller.

Nateowami (dismissed):
> Reviewed 1 of 3 files at r4, 3 of 5 files at r6, 1 of 1 files at r7, 1 of 1 files at r9.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472)* status: 18 of 21 files reviewed, 1 unresolved discussion (waiting on @josephmyers and @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.html` line 55 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-O-7Q1QOAzOgOP_mGZFD:-O-7Q1QOAzOgOP_mGZFE:b-nyb8ba) ([raw file](https://github.com/sillsdev/web-xforge/blob/e045fbc17a6c47c7cd9c3b38b2261a3975f57924/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.html#L55)):*
> ```HTML
>         </div>
>         <mat-menu #linkUsageMenu>
>           <mat-selection-list [multiple]="false" *ngIf="shareLinkUsageOptions.length > 1">
> ```

Missed one here. Updated myself just to speed things up.

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.scss` line 52 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-O-7TqjxAfGqIOgA9s_h:-O-7TqjxAfGqIOgA9s_i:bug290z) ([raw file](https://github.com/sillsdev/web-xforge/blob/426684c1769ce9ce6839b0588ee01fac76ca84d6/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.scss#L52)):*
> ```SCSS
> 
> ::ng-deep .mat-mdc-list-base {
>   padding-block: 0;
> ```

You shouldn't be starting a line with `::ng-deep` if it's at all avoidable. Try adding `background-color: red !important` to all your `::ng-deep` rules and you can see it's styling entirely different components (e.g. navigation component) as well.

It looks to me like this rule, and the one below, do not actually need `::ng-deep`.

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472)* status: 18 of 21 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.html` line 55 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-O-7Q1QOAzOgOP_mGZFD:-O-8swqDAkXWiK9HTppp:bk0py46) ([raw file](https://github.com/sillsdev/web-xforge/blob/e045fbc17a6c47c7cd9c3b38b2261a3975f57924/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.html#L55)):*

Missed one here. Updated myself just to speed things up.

</blockquote></details>

Thanks, missed that, too
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.scss` line 52 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-O-7TqjxAfGqIOgA9s_h:-O-8t1xJ1KJOdjrGHpRJ:bk2e73b) ([raw file](https://github.com/sillsdev/web-xforge/blob/426684c1769ce9ce6839b0588ee01fac76ca84d6/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-dialog.component.scss#L52)):*

You shouldn't be starting a line with `::ng-deep` if it's at all avoidable. Try adding `background-color: red !important` to all your `::ng-deep` rules and you can see it's styling entirely different components (e.g. navigation component) as well.

It looks to me like this rule, and the one below, do not actually need `::ng-deep`.
</blockquote></details>

Done. I forgot to go back and check those were needed

Nateowami (commented):
> Reviewed 5 of 7 files at r10.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472)* status: 17 of 21 files reviewed, all discussions resolved (waiting on @pmachapman)

Nateowami (commented):
To fix the issue the testers found, I suggest adding `[panelWidth]="80"` to the chapter-select-menu in `book-chapter-chooser.component.html`. I initially thought having the width set to auto would be better, but then it makes the menu very narrow when a book has less than 10 books. 80px is wide enough for Psalms.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-:-O-veNAw9aRW68B7QYkP:bd9ives)* status: 17 of 21 files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (commented):
> Reviewed 3 of 3 files at r4, 1 of 2 files at r5, 1 of 1 files at r8, 2 of 7 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (commented):
Done. Though I did a bit of custom code to help cases where there aren't 100+ chapters.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472#-:-O06pWzT7wYjbqN_xOci:b-iwis08)* status: 18 of 22 files reviewed, all discussions resolved (waiting on @Nateowami and @pmachapman)

pmachapman (approved):
> Reviewed 4 of 4 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami:
The test team found an issue with archived questions in RTL, but since that's a fairly edge case (I doubt there are any project admins using Arabic for managing community checking questions), we're going to merge and fix that in a followup.

Nateowami (approved):
> Reviewed 2 of 2 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2472)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers:
@Nateowami the issue they found is similar to one that Peter found when first reviewing this task. He saw overlap for Archived Questions in English.

josephmyers:
#2552

PR #2476, opened 2024-05-22
Improve layout of suggestion engine card

Nateowami:
The recent upgrade from the legacy MatCard introduced some changes that I think make the suggestion engine card layout worse.

Here's the timeline:

Before MatCard update | Current master | This branch
----------------------|----------------|------------
suggestion_card_before_mat_update | suggestion_card_master | suggestion_card_branch

This branch does not cause a pixel-perfect match with the original, but that wasn't a goal. Instead, it fixes two problems:
- "Trained segments" always wraps, which looks bad
- The background for the stars didn't span the whole width, which makes it look odd

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2476)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.scss
@@ -73,7 +73,8 @@ mat-card {
   .engine-card-content {
     display: flex;
     flex-direction: column;
-    margin-bottom: 0;
+    padding: 0;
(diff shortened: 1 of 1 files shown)


pmachapman (approved):
:lgtm: That looks much better for one of my test projects with 26286 trained segments!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2476#-:-NyX9vKECqoNM5z3q-Cu:b-yr4vkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2479, opened 2024-05-23
SF-2358 Update mat-button to mdc based mat-button

RaymondLuong3:
The PR updates the legacy material button to the mdc-based button component. This used `ng generate @angular/material:mdc-migration` as the starting point. Additionally, the material 16 version was bumped up to v16.2.14.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2479)


Nateowami:
Good work. Here are some remaining problems I see:

Broken (Some CSS needs to be removed I think. Check for similar CSS elsewhere. I'm pretty sure this is not the only component that specifically styled the buttons for spacing the icon, even if it doesn't look as bad elsewhere)


These buttons (edit and archive) were deliberately styled to be shorter and therefore take up less vertical space
Screenshot from 2024-05-24 00-06-52

This filter button also was designed to be shorter
Screenshot from 2024-05-24 00-06-42

Finally, the icon buttons used to have a ripple when hovered. I thought that was good and made it more obvious that it was a button. When we last upgraded from @angular-mdc/web to @angular/material, we lost the ripple on hover, and only got it on mouse down. I thought we were going to get the ripple on hover back with this upgrade, based on looking at [the examples for icon buttons](https://v17.material.angular.io/components/button/overview), but on this branch it's lost entirely. Why is that?

RaymondLuong3 (commented):
Thanks for the careful review. I've updated the components that you noticed were problematic.

Nice find. Fixed.


The 8px column gap looks good. I thought about applying it app-wide, but it appears that other buttons look good without the extra gap, so I left it to just the import questions dialog.


It seems that the color of the button was getting overridden by the angular theme. I updated the css vars that the theme uses to determine the button color. I think that is the better way of doing it.


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-:-Nz-PrZx7rW6NkxafZ6E:bzfj54d)* status: 0 of 26 files reviewed, all discussions resolved

Nateowami:
`@include mat.icon-button-theme($material-app-theme);` is needed in `material-styles.scss`. That should fix the lack of ripple.

The dividers on the editor toolbar are not supposed to be full height. I think 6px of padding on top and bottom looks about right.

Nateowami:
There's also some customizing of spacing for the icons in a button in the share dialog that can now be removed.

RaymondLuong3 (commented):
Thanks! The icon buttons have their hover ripples back! Actually, I don't think that ripples refer to the change in background colour on hover, it might refer to the effect when a button is clicked. But anyway, the hover effect works now.

Here is the toolbar with the dividers.


Here is the share dialog action buttons. I removed the css that I could see was not having an effect. I converted the "copy link" button to a regular mat-button since it is not our normal design to have 2 mat-flat-buttons on a dialog.


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-:-Nz0P6RFBtpQsqfTt_yB:bup1q9u)* status: 0 of 30 files reviewed, all discussions resolved

Nateowami (changes requested):
Toolbar looks much better, thanks! And yes, I think those are technically not ripples, because they don't, well, ripple. 

> I converted the "copy link" button to a regular mat-button since it is not our normal design to have 2 mat-flat-buttons on a dialog.

That's an improvement in some environments, and a detriment in others. Here's how it looks on my machine:



> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-:-Nz3kEBV-mH1jvmv__yv:b9t9una)* status: 0 of 32 files reviewed, 4 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/material-styles.scss` line 183 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-Nz3kn6u5fqUYF67F20t:-Nz3kn6u5fqUYF67F20u:b-xzdne1) ([raw file](https://github.com/sillsdev/web-xforge/blob/6eb323b282a835c18688e03b0350f5f2579dacbc/src/SIL.XForge.Scripture/ClientApp/src/material-styles.scss#L183)):*
> ```SCSS
> 
> button {
>   --mdc-icon-button-disabled-icon-color: rgba(0, 0, 0, 0.2);
> ```

Why are you setting this on `button`? I'm guessing maybe you were copying the example rule you see above it, but I'm guessing that one is in error as well.

If you want this to be global, it should be set on the `html` element like most other variables. Right now this is affecting the color of icons in button elements, but not those in `<a>` elements. See https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties if you need an explanation for how CSS custom properties work.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 51 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-Nz3mQPd7pibsXqLIwuj:-Nz3mQPd7pibsXqLIwuk:b54wva8) ([raw file](https://github.com/sillsdev/web-xforge/blob/6eb323b282a835c18688e03b0350f5f2579dacbc/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L51)):*
> ```SCSS
> .mat-mdc-unelevated-button,
> .mat-button-toggle-button {
>   transition: background-color 0.15s, color 0.15s;
> ```

This doesn't appear to do anything anymore. It's even more obvious if you change the transition time to more than a second and still can't see a transition.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 313 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-Nz3wGsX-RxxOL_8-QFF:-Nz3wGsX-RxxOL_8-QFG:b12x1f0) ([raw file](https://github.com/sillsdev/web-xforge/blob/6eb323b282a835c18688e03b0350f5f2579dacbc/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L313)):*
> ```SCSS
> }
> 
> .active-question-scope-button {
> ```

This button used to have proper vertical alignment of the text, but now looks like this.

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html` line 90 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-Nz3qm491ASH75ZWPs-G:-Nz3qm491ASH75ZWPs-H:bna1tll) ([raw file](https://github.com/sillsdev/web-xforge/blob/6eb323b282a835c18688e03b0350f5f2579dacbc/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html#L90)):*
> ```HTML
>               <button
>                 mat-button
>                 [class]="isScreenSmall ? 'mdc-icon-button' : 'mdc-button'"
> ```

If you open the answer panel and then adjust the width of the page, you can end up in this state:



Also, the cancel button doesn't look like an icon button on hover:



I realize you're just trying to make minimal changes to make this work with the new button. However, I think the entire concept here of trying to make an element switch between being a regular button an and icon button is flawed and will keep giving us trouble. Instead, it's much simpler to just create two completely separate states:

``` html
            <div class="form-actions" *ngIf="isScreenSmall">
              <button mat-icon-button type="button" (click)="hideAnswerForm()" id="cancel-answer">
                <mat-icon class="material-icons-outlined">cancel</mat-icon>
              </button>
              <button mat-icon-button color="primary" id="save-answer" (click)="submit()" [disabled]="submittingAnswer">
                <mat-icon>check_circle</mat-icon>
              </button>
            </div>
            <div class="form-actions" *ngIf="!isScreenSmall">
              <button mat-button type="button" (click)="hideAnswerForm()" id="cancel-answer">
                <mat-icon>close</mat-icon>
                {{ t("cancel") }}
              </button>
              <button mat-flat-button color="primary" id="save-answer" (click)="submit()" [disabled]="submittingAnswer">
                <mat-icon>check</mat-icon>
                {{ t("save_answer") }}
              </button>
            </div>
```

And then remove this CSS from `checking-answers.component.scss`:
``` css
    mat-icon {
      vertical-align: middle;
    }
```
And check for any other CSS being used on these buttons that may not be needed anymore.

Nateowami:
The header element above the chapter on the community checking page is now 8px taller than before. It has 8px of padding on top and bottom. If you drop that to 4px, the spacing ends up the same as before, and I think it's still sufficient. This may seem insignificant, but on small screens unnecessarily taking up 8px of vertical space makes a difference.

You can look for differences by opening QA in one tab and this branch in another, and then switching between the tabs to see what has changed. To be clear, we're not trying to avoid having anything change in appearance; we're trying to avoid making anything worse. So just because there's a difference doesn't mean something needs to be changed.

Additionally, please look through the changes that Chromatic shows and make sure they all make sense. I'm not asking you to approve/review the changes in Chromatic, just look over them and make sure they're all appropriate.

Sorry to load more on you; I know this PR is a big undertaking.

RaymondLuong3 (commented):
Nice, thanks for pointing those out. I see the additional header padding and have removed that.

I looked at Chromatic, and it appears that the Add comment button is taller by about 12 pixels, but when I tried to see that on in dev-tools, I couldn't find that it was any different except for the touch area. The actual button height was 36px like I expected. I think it is OK as-is.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-:-Nz5BZ6cCJR5BoNt2qR9:b-gtbt87)* status: 0 of 32 files reviewed, 4 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/material-styles.scss` line 183 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-Nz3kn6u5fqUYF67F20t:-Nz5CEo04yW7e61R692Y:b-twgyr5) ([raw file](https://github.com/sillsdev/web-xforge/blob/6eb323b282a835c18688e03b0350f5f2579dacbc/src/SIL.XForge.Scripture/ClientApp/src/material-styles.scss#L183)):*

Why are you setting this on `button`? I'm guessing maybe you were copying the example rule you see above it, but I'm guessing that one is in error as well.

If you want this to be global, it should be set on the `html` element like most other variables. Right now this is affecting the color of icons in button elements, but not those in `<a>` elements. See https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties if you need an explanation for how CSS custom properties work.
</blockquote></details>

Thanks! Now that we have the `mat-icon-button` theme in place, this was no longer needed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 51 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-Nz3mQPd7pibsXqLIwuj:-Nz5CMm60-P9XVl5XErY:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/6eb323b282a835c18688e03b0350f5f2579dacbc/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L51)):*

This doesn't appear to do anything anymore. It's even more obvious if you change the transition time to more than a second and still can't see a transition.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss` line 313 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-Nz3wGsX-RxxOL_8-QFF:-Nz5CNiW0gFlSI-K1PZh:b-wmodlt) ([raw file](https://github.com/sillsdev/web-xforge/blob/6eb323b282a835c18688e03b0350f5f2579dacbc/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.scss#L313)):*

This button used to have proper vertical alignment of the text, but now looks like this.

</blockquote></details>

The button component's internal elements css needed to be modified to get the text to center again. Not ideal given this could break when we update the button library.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html` line 90 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-Nz3qm491ASH75ZWPs-G:-Nz5Cgtl3En8tF1WG6mD:b-xkvza9) ([raw file](https://github.com/sillsdev/web-xforge/blob/6eb323b282a835c18688e03b0350f5f2579dacbc/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html#L90)):*

If you open the answer panel and then adjust the width of the page, you can end up in this state:



Also, the cancel button doesn't look like an icon button on hover:



I realize you're just trying to make minimal changes to make this work with the new button. However, I think the entire concept here of trying to make an element switch between being a regular button an and icon button is flawed and will keep giving us trouble. Instead, it's much simpler to just create two completely separate states:

``` html
            <div class="form-actions" *ngIf="isScreenSmall">
              <button mat-icon-button type="button" (click)="hideAnswerForm()" id="cancel-answer">
                <mat-icon class="material-icons-outlined">cancel</mat-icon>
              </button>
              <button mat-icon-button color="primary" id="save-answer" (click)="submit()" [disabled]="submittingAnswer">
                <mat-icon>check_circle</mat-icon>
              </button>
            </div>
            <div class="form-actions" *ngIf="!isScreenSmall">
              <button mat-button type="button" (click)="hideAnswerForm()" id="cancel-answer">
                <mat-icon>close</mat-icon>
                {{ t("cancel") }}
              </button>
              <button mat-flat-button color="primary" id="save-answer" (click)="submit()" [disabled]="submittingAnswer">
                <mat-icon>check</mat-icon>
                {{ t("save_answer") }}
              </button>
            </div>
```

And then remove this CSS from `checking-answers.component.scss`:
``` css
    mat-icon {
      vertical-align: middle;
    }
```
And check for any other CSS being used on these buttons that may not be needed anymore.
</blockquote></details>

Yes, that is much more stable for when we might have to update the button component in the future, then it should just work. I've removed most of the css that impacted this css.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2479)* status: 0 of 32 files reviewed, all discussions resolved

___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 51 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-Nz3mQPd7pibsXqLIwuj:-Nz5pSU6942YoyqMUnqH:bphptdz) ([raw file](https://github.com/sillsdev/web-xforge/blob/6eb323b282a835c18688e03b0350f5f2579dacbc/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L51)):*

Done.
</blockquote></details>

I actually thought having the transition was a good idea and was suggesting fixing it, not removing it.

The Material default is to not use not use a transition, but the colors they use are close to each other, so there's not a sudden change between two very different colors, like grey and green. I'd like to move to a theme that's more similar to that, but where not there now.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2479)* status: 0 of 32 files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 51 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-Nz3mQPd7pibsXqLIwuj:-Nz9eyOW4nzP8gfEIfPa:b-s11zih) ([raw file](https://github.com/sillsdev/web-xforge/blob/6eb323b282a835c18688e03b0350f5f2579dacbc/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L51)):*

I actually thought having the transition was a good idea and was suggesting fixing it, not removing it.

The Material default is to not use not use a transition, but the colors they use are close to each other, so there's not a sudden change between two very different colors, like grey and green. I'd like to move to a theme that's more similar to that, but where not there now.
</blockquote></details>

Sorry, I definitely misunderstood. It looks like I could get it working again with a simple update. You will need to slow down the transition time to see it.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2479)* status: 0 of 32 files reviewed, all discussions resolved

___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 51 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-Nz3mQPd7pibsXqLIwuj:-Nz9hWaHB511qBva7U6H:b-h28qet) ([raw file](https://github.com/sillsdev/web-xforge/blob/6eb323b282a835c18688e03b0350f5f2579dacbc/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L51)):*

Sorry, I definitely misunderstood. It looks like I could get it working again with a simple update. You will need to slow down the transition time to see it.
</blockquote></details>

This isn't the right way to do a transition. If you want the color of a button to change on hover, you do something like this:
``` css
button {
  color: red;
  transition: color 0.3s ease;
}
button:hover {
  color: blue;
}
```
What you're doing here is making the transition conditional based on whether the button is being hovered. The transition should always be present.

See https://developer.mozilla.org/en-US/docs/Web/CSS/transition

Nateowami:
I can't reply in Reviewable, because it deletes every character right after I type it. The code block in my comment above was not a suggestion, it was an example of the most basic use of a transition that is triggered on hover, as a way of showing how transitions are supposed to be used.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2479)* status: 0 of 32 files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 51 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-Nz3mQPd7pibsXqLIwuj:-NzAIejO7jLCq_7ZkwUh:b-ab4aef) ([raw file](https://github.com/sillsdev/web-xforge/blob/6eb323b282a835c18688e03b0350f5f2579dacbc/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L51)):*

This isn't the right way to do a transition. If you want the color of a button to change on hover, you do something like this:
``` css
button {
  color: red;
  transition: color 0.3s ease;
}
button:hover {
  color: blue;
}
```
What you're doing here is making the transition conditional based on whether the button is being hovered. The transition should always be present.

See https://developer.mozilla.org/en-US/docs/Web/CSS/transition
</blockquote></details>

Thanks. I can understand why we want the transition to apply to the button, and not just the state of the button when it is hovered over.
I found it interesting that unless I specify that the button is not in the disabled state, I cannot get the transition time to apply.

There must be something I'm not understanding with icon buttons, the transition animation doesn't work there at all if I make it apply to the button like this. I left it as is.



_Code snippet:_
```SCSS
.mdc-icon-button {
  transition: color 0.15s
}
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2479)* status: 0 of 32 files reviewed, all discussions resolved

___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 51 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-Nz3mQPd7pibsXqLIwuj:-NzEJTwn8eME2FZL8eKv:b-sjxq8f) ([raw file](https://github.com/sillsdev/web-xforge/blob/6eb323b282a835c18688e03b0350f5f2579dacbc/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L51)):*

Thanks. I can understand why we want the transition to apply to the button, and not just the state of the button when it is hovered over.
I found it interesting that unless I specify that the button is not in the disabled state, I cannot get the transition time to apply.

There must be something I'm not understanding with icon buttons, the transition animation doesn't work there at all if I make it apply to the button like this. I left it as is.
</blockquote></details>

Setting one property after another like this makes the first one do nothing, as dev tools shows:



This doesn't have anything to do with transitions; it's just how CSS works.

The reason you couldn't get the transitions to work without specifying `:not(:disabled)` is because there's another transition specified on the button, and it's overriding this transition. You can see that this is happening in dev tools:



It's the transition of the box shadow, which elevated buttons have. Since this isn't an elevated button, overriding that transition has no actual visual effect, since it was transitioning something that doesn't exist.

Here's what I consider the fixed SCSS (obviously times should be reduced; I added an extra second to make the transition obvious):
``` scss
.mdc-button.mat-mdc-unelevated-button,
.mdc-button.mat-button-toggle-button {
  transition: background-color 1.15s, color 1.15s;

  &:not(:disabled):hover {
    background-color: variables.$greenLight;
    color: variables.$sf_grey;
  }
}

button.mdc-icon-button {
  transition: color 1.15s;

  &:not([disabled]):hover {
    color: variables.$greenLight;
  }
}
```

I changed `.mat-mdc-unelevated-button` and `.mat-button-toggle-button` to additionally specify `.mdc-button`, which increases the specificity of the rule, and allows it to have higher priority over the box shadow transition.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2479)* status: 0 of 32 files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/styles.scss` line 51 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-Nz3mQPd7pibsXqLIwuj:-NzEd1MP3m7a4Pj2vU5X:bak10qt) ([raw file](https://github.com/sillsdev/web-xforge/blob/6eb323b282a835c18688e03b0350f5f2579dacbc/src/SIL.XForge.Scripture/ClientApp/src/styles.scss#L51)):*

Setting one property after another like this makes the first one do nothing, as dev tools shows:



This doesn't have anything to do with transitions; it's just how CSS works.

The reason you couldn't get the transitions to work without specifying `:not(:disabled)` is because there's another transition specified on the button, and it's overriding this transition. You can see that this is happening in dev tools:



It's the transition of the box shadow, which elevated buttons have. Since this isn't an elevated button, overriding that transition has no actual visual effect, since it was transitioning something that doesn't exist.

Here's what I consider the fixed SCSS (obviously times should be reduced; I added an extra second to make the transition obvious):
``` scss
.mdc-button.mat-mdc-unelevated-button,
.mdc-button.mat-button-toggle-button {
  transition: background-color 1.15s, color 1.15s;

  &:not(:disabled):hover {
    background-color: variables.$greenLight;
    color: variables.$sf_grey;
  }
}

button.mdc-icon-button {
  transition: color 1.15s;

  &:not([disabled]):hover {
    color: variables.$greenLight;
  }
}
```

I changed `.mat-mdc-unelevated-button` and `.mat-button-toggle-button` to additionally specify `.mdc-button`, which increases the specificity of the rule, and allows it to have higher priority over the box shadow transition.
</blockquote></details>

Wow, I must not be looking very closely at the CSS in the dev tools, or I would have seen that myself. Thanks for the tip. I had not realized that increasing the specificity of a rule increases a rules priority, that is news to me!

Nateowami (changes requested):
Great work. Just one change I see needs to be made (and a merge or rebase).

> Reviewed 14 of 26 files at r1, 6 of 8 files at r2, 3 of 6 files at r3, 4 of 5 files at r4, 4 of 5 files at r5, 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-:-O-77aeW3T1ueFWRkfIT:bo33iwt)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html` line 222 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-O-75Jzq2hQbSq1-oQWL:-O-75Jzq2hQbSq1-oQWM:bsey7s0) ([raw file](https://github.com/sillsdev/web-xforge/blob/74bf8f3fdb8b7b6ce3a9c97009867d61f7633578/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html#L222)):*
> ```HTML
>             class="next-question"
>           >
>             {{ t("next") }}<mat-icon iconPositionEnd>chevron_{{ i18n.forwardDirectionWord }}</mat-icon>
> ```

This clearly works, but I can't find where it's documented. In fact, all documentation for buttons with icons and text seems sparse. Where did you find it?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss` line 101 at r8](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-O-6y_KY05KK1YCY8smc:-O-6y_KY05KK1YCY8smd:b-9v87ta) ([raw file](https://github.com/sillsdev/web-xforge/blob/74bf8f3fdb8b7b6ce3a9c97009867d61f7633578/src/SIL.XForge.Scripture/ClientApp/src/app/shared/notice/notice.component.scss#L101)):*
> ```SCSS
> 
>     &:hover {
>       background-color: var(--notice-color-button-hover-bg);
> ```

This is not working because the hover effect from styles.scss is overriding it. I don't see a great solution to that problem, other than moving away from directly specifying the hover color and instead using Material's theme solution:
``` scss
      background-color: var(--notice-color-button-hover-bg) !important;
      color: var(--notice-color-button-hover-text) !important;
```
Using `!important` is ugly, but it's an acceptable solution until we can move away from directly specifying the hover effect and using Material theming directly

Nateowami (approved):
:lgtm:

> Reviewed 13 of 13 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-:-O-7MJYeBSZWdt1jubI1:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/material-styles.scss` line 130 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-O-7LHS5BdqByr_62RLB:-O-7LHS5BdqByr_62RLC:bkvqjb8) ([raw file](https://github.com/sillsdev/web-xforge/blob/096eb4c880a7e4a70858a36cb1c53a48f29729b3/src/SIL.XForge.Scripture/ClientApp/src/material-styles.scss#L130)):*
> ```SCSS
> @include mat.bottom-sheet-theme($material-app-theme);
> @include mat.icon-button-theme($material-app-theme);
> @include mat.button-theme($material-app-theme);
> ```

Do we need fab-theme?

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2479)* status: all files reviewed, 1 unresolved discussion

___
*[`src/SIL.XForge.Scripture/ClientApp/src/material-styles.scss` line 130 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2479#-O-7LHS5BdqByr_62RLB:-O-Bp5h22yoE3L6Kdfjx:b78kgel) ([raw file](https://github.com/sillsdev/web-xforge/blob/096eb4c880a7e4a70858a36cb1c53a48f29729b3/src/SIL.XForge.Scripture/ClientApp/src/material-styles.scss#L130)):*

Do we need fab-theme?
</blockquote></details>

@RaymondLuong3  I've merged, but I still think this needs to be considered.

RaymondLuong3:
Thanks. It is likely we will need to revisit the themes for all of our components.

PR #2480, opened 2024-05-24
SF-2658 Audio player randomly scrolls back to active question

nigel-wells:
* Stop highlighting and scrolling when setting an active verse to the same value

The underlaying cause is when a remote user makes a change to the question doc query it triggers an update for all users subscribed to that query. This change triggers the active verse to get set again which then highlights and scrolls to the verse being set.

### Acceptance tests

* Configure a project for community checking
* Load in questions and attach chapter audio and timing data
* Grant 2 users access to community checking on the project

**Browser 1**

1. Select the first question
2. Start playing chapter audio
3. Wait for the highlighted verse change away from the verse of the selected question

**Browser 2**

1. Select the first question
2. Interact in some way i.e. answer the question, like an answer, etc.

**Observe**
When a remote browser is interacting with questions it shouldn't affect the other browser whose active question isn't changing.


pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2480#-:-Nyqd4IW4-KzG437LCbq:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

PR #2492, opened 2024-05-28
SF-2729 Fix ShareDB apply ops crash due to array out of order

pmachapman:
This PR fixes the "Referenced element not an object (it was undefined)." error that occurs when syncing updates the texts array of the project.

This because occurs because the reference to the `texts` array for an `sf_projects_profile` in IndexedDB is sorted by `bookNum`.

To resolve this, I ensure that the copy of the texts array used for drafting is `slice()`'ed from the `sf_project_profile` object.

I also updated the code that performs the sort on this array to `slice()` from the object it was passed, as we should only `sort()` on a `slice()`'ed copy of an array, not the array the method was passed by reference.

Testing may be done by a developer, but given the change is very small with very low risk, I do not think testing is required. Full testing details are in the JIRA ticket, however, for those interested.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2492)


RaymondLuong3 (approved):
Dev testing this issue and it appears to be working as expected as per the acceptance criteria. Nice work!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2492#-:-NyzyYyqE8wZ1fx_B09_:b-4kxyyb)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2497, opened 2024-05-29
Add Nepali locale and localizations

RaymondLuong3:
This will require that the Auth0 localizations are added to the Auth0Config repo first merging this so that selecting Nepali from the home page does not trigger an error when navigating to the Auth0 login screen.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2497)


Nateowami:
The RESX files should be included as well.

Nateowami:
@RaymondLuong3 How about including the RESX files, marking the locale as non-production, and then merging it in? That will allow it to start getting tested, and when we get Auth0 updated it will be a simple matter of flipping a switch to enable the locale.

RaymondLuong3:
Thanks, I think that is reasonable and there is no reason not to include the Nepali localizations before it is production ready.

Nateowami (approved):
> Reviewed 2 of 3 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2497)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2498, opened 2024-05-29
SF-2750 Do not train on all books when no books are specified

pmachapman:
Currently, when no books are selected for training, due to the way SF sends null values for both `textIds` and `scriptureRange` for the `trainOn` object, Serval ends up training on all books. This results in only blank verses being translated, instead of all verses.

This PR, based on the change https://github.com/sillsdev/serval/issues/383, specifies an empty array for train `textIds` collection when no books are selected for training. This means that all verses in the target will be translated by Serval, instead of just blank verses.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2498)


Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2498)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2499, opened 2024-05-30
SF-2753 Shorten email template

Nateowami:
Here's how the email template now looks on my local dev machine:
```
Please explain the problem.




--- Technical details ---
Scripture Forge: 9.9.9
Chrome: 125.0.0.0
Brave: Yes
Linux: Unknown
URL: http://localhost:5000/projects/65eb63b90a55c8594e4e9f17/translate/GEN/1
```

### Changes
1. The request to explain the problem is the only thing at the beginning of the email so it's harder to miss. New lines separate it from the technical details.
2. A default subject is no longer provided
   a. This prevents all our support requests from having the same "Scripture Forge issue" subject
   b. Most mail clients ask a user if they're sure they want to send an email with no subject, so the user will at least have to write *something* about what the problem is
3. The old email template dictated what technical details should be provided. This means we couldn't change the technical details without updating a localization string. The technical details can now be updated with no localization impact.
4. We now indicate whether the user is using Brave or not. Some of our users are more security conscious, and therefore it's likely an outsized number of users using Brave, and it might be the reason why some error IDs we get through email don't have matches in Bugsnag. This will help us understand whether this is the case or not.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2499)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json
+++ b/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json
@@ -326,10 +326,9 @@
     "text_size": "Text size"
   },
   "issue_email": {
-    "body": "Thanks for reporting the issue!\nIt would help us if you fill out some of the information below, but please submit even if you can't fill out much.\nIf you are requesting a feature many of the fields may not be applicable.\n\nBug report\nA clear and concise description of what the bug is.\n\nSteps to reproduce\nFor example:\n1. Go to ...\n2. Click on ...\n3. Scroll down to ...\n4. See error\n\nActual behavior\nPlease describe actual behavior of the issue you are observing.\n\nExpected behavior\nA clear and concise description of what you expected to happen.\n\nScreenshots\nIf applicable, add screenshots to help explain your issue.\n\nAdditional context\nAdd any other context about the problem here.\n\nPossible solution\nAdd any possible solutions to the problem here.\n\nYour environment:\nSoftware                 Version\n-----------------------------------------\n{{ siteName }} - {{ siteVersion }}\n{{ browserName }} - {{ browserVersion }}\n{{ operatingSystem }} - {{ operatingSystemVersion }}\n\nURL: {{ url }}\nError id: {{ errorId }}",
-    "not_applicable": "not applicable",
-    "subject": "{{ siteName }} issue",
-    "unknown": "unknown"
+    "heading": "Please explain the problem.",
+    "technical_details": "Technical details",
+    "unknown": "Unknown"
   },
   "join": {
     "error_occurred_login": "An error occurred trying to join the project",
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts
@@ -50,6 +50,13 @@ export function isIosDevice(): boolean {
   return BROWSER.getOSName(true) === 'ios';
 }
 
+export function isBrave(): boolean {
+  // The officially supported way of detecting Brave is to call navigator.brave.isBrave(), which returns
+  // Promise<boolean>. See https://github.com/brave/brave-browser/wiki/Detecting-Brave-(for-Websites)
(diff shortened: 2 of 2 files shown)


RaymondLuong3 (approved):
I like that it is a simpler email template. The previous one was rarely ever being filled. This will suite our needs.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2499#-:-NzUkXO9A9ORsS4oIaxd:bi4u632)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json` line 330 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2499#-NzUk0Jf43LmXwXIrglh:-NzUk0Jf43LmXwXIrgli:b22oipj) ([raw file](https://github.com/sillsdev/web-xforge/blob/d49d1f644ff6b085769fd189ae690cf155f62c9c/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json#L330)):*
> ```JSON
>   "issue_email": {
>     "heading": "Please explain the problem.",
>     "technical_details": "Technical details",
> ```

I'm not sure we even need to localized this. We do not expect the user to really need to know that the following are technical details, but it's also easy to do. Let's leave this as-is.



_Code quote:_
```JSON
    "technical_details": "Technical details",
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts` line 57 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2499#-NzUgEtM--y5kYae1Ip0:-NzUgEtM--y5kYae1Ip1:bkkibf8) ([raw file](https://github.com/sillsdev/web-xforge/blob/d49d1f644ff6b085769fd189ae690cf155f62c9c/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts#L57)):*
> ```XML
>   // Promise<boolean>. See https://github.com/brave/brave-browser/wiki/Detecting-Brave-(for-Websites)
>   // Just checking for the presence of this property works fine though and doesn't require awaiting a promise.
>   return 'brave' in window.navigator;
> ```

That is a neat trick. We can check without actually awaiting the promise :)



_Code quote:_
```XML
  return 'brave' in window.navigator;
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2499)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json` line 330 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2499#-NzUk0Jf43LmXwXIrglh:-NzV16oiAkhLT99Ka0x1:b-5squdc) ([raw file](https://github.com/sillsdev/web-xforge/blob/d49d1f644ff6b085769fd189ae690cf155f62c9c/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json#L330)):*

I'm not sure we even need to localized this. We do not expect the user to really need to know that the following are technical details, but it's also easy to do. Let's leave this as-is.
</blockquote></details>

I wondered the same thing. But I thought perhaps if they knew what it was they would be more likely to leave as-is instead of trying to edit it or something. 

It's also part of why I added the `---`, so even if it's in a language we don't understand, it will still be clearly delineated.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts` line 57 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2499#-NzUgEtM--y5kYae1Ip0:-NzV1In_2DzirTVkQQ-W:b-u61m7y) ([raw file](https://github.com/sillsdev/web-xforge/blob/d49d1f644ff6b085769fd189ae690cf155f62c9c/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts#L57)):*

That is a neat trick. We can check without actually awaiting the promise :)
</blockquote></details>

I'm *hoping* it's actually a neat trick. I'm guessing there's some reason they made it a promise, but I can't find a reason.

PR #2502, opened 2024-06-03
SF-2751 Fix NG0911 error

siltomato:
This PR attempts to fix NG0911 errors that were occasionally being thrown when navigating away from the editor component.

A recent discussion in an open angular issue (https://github.com/angular/angular/issues/54527) suggests that NG0911 is happening when `takeUntilDestroyed` is used with a `DestroyRef` after component has been destroyed, such as when:
 - subscription happens after awaiting an earlier promise,
 - then navigation destroys component,
 - then promise resolves and code flow subscribes with `takeUntilDestroyed`

I suspect that the observables subscribed in `initEditorTabs()` are the offenders.  The solution in this PR is to combine those observables and return them from `initEditorTabs()` so that they can be piped from projectDoc changes, allowing a single call to `subscribe` before any awaited promises.

Note: Some comments in the above angular issue included suggestions to the angular team to change the behavior of `takeUntilDestroyed` to just terminate the subscription rather than throw an error.  This is worth keeping an eye on.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2502)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -3734,38 +3734,37 @@ describe('EditorComponent', () => {
   }));
 
   describe('initEditorTabs', () => {
-    it('should add source tab group when sourceLabel is defined', fakeAsync(() => {
+    it('should add source tab group when source is defined', fakeAsync(() => {
       const env = new TestEnvironment();
-      flush();
+      const projectDoc = env.getProjectDoc('project01');
       const spyCreateTab = spyOn(env.tabFactory, 'createTab').and.callThrough();
-      const sourceLabel = 'source label';
-      env.component['sourceLabel'] = sourceLabel;
-      env.component['targetLabel'] = 'target label';
-      env.component.initEditorTabs();
-      tick(100);
-      expect(spyCreateTab).toHaveBeenCalledWith('project-source', { headerText: sourceLabel });
+      env.wait();
+
+      expect(spyCreateTab).toHaveBeenCalledWith('project-source', {
+        headerText: projectDoc.data?.translateConfig.source?.shortName
+      });
+      discardPeriodicTasks();
     }));
 
-    it('should not add source tab group when sourceLabel is undefined', fakeAsync(() => {
+    it('should not add source tab group when source is undefined', fakeAsync(() => {
       const env = new TestEnvironment();
-      flush();
       const spyCreateTab = spyOn(env.tabFactory, 'createTab').and.callThrough();
-      env.component['sourceLabel'] = undefined;
-      env.component['targetLabel'] = 'target label';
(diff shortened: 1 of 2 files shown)


pmachapman (changes requested):
@siltomato Just one question!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2502#-:-NzUqlOc45SIT9woyzRo:b7jjxs8)* status: all files reviewed, 1 unresolved discussion (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 638 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2502#-NzUuc5i5qsqXfUmmP0_:-NzUuc5i5qsqXfUmmP0a:b-5guak7) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d6d1b947e5cc33b8a654a48c5508ddfaeaa4725/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L638)):*
> ```XML
>         switchMap(() => this.initEditorTabs())
>       )
>       .subscribe();
> ```

Does this subscription need to be unsubscribed from? i.e. should be subscribe via `this.Subscribe()`?



_Code quote:_
```XML
.subscribe();
```

siltomato (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2502)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 638 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2502#-NzUuc5i5qsqXfUmmP0_:-NzUvJTo7axMf5GULKlg:b-5fug1w) ([raw file](https://github.com/sillsdev/web-xforge/blob/4d6d1b947e5cc33b8a654a48c5508ddfaeaa4725/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L638)):*

Does this subscription need to be unsubscribed from? i.e. should be subscribe via `this.Subscribe()`?
</blockquote></details>

```typescript
takeUntilDestroyed(this.destroyRef),
```
should take care of unsubscribing.  This was introduced in angular 16 and should allow us to move away from extending the `SubscriptionDisposable` class entirely.

Unfortunately, it looks like [there is a caveat (which caused the NG0911 error)](https://github.com/angular/angular/issues/54527).  Perhaps (as suggested in the linked issue comments) they will adjust the behavior to not throw the error when the component has already been destroyed and just terminate the subscription.

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2502#-:-NzUyXkqCx7lOTam-Mqp:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

PR #2509, opened 2024-06-06
SF-2754 Update user profile interface language only when needed

RaymondLuong3:
Each time the angular app is reloaded, a call is made to Auth0 to update the interface language. This was not necessarily a big problem, but it was unnecessary if the interface language still matched what was current in the browsing session.

The intended behaviour of the PR is as follows

- when the user logs in, the user's language is set to the language of the landing page (unchanged)
- upon opening the app (or on refresh) and the user is logged in, the user's language is updated if needed (changed)
- upon opening the app in 2 browsers, the most recently reloaded browser sets the user's language (unchanged)
- the user cannot change interface language while offline (unchanged)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2509)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -193,7 +193,7 @@ describe('AppComponent', () => {
 
     tick();
     env.fixture.detectChanges();
-    verify(mockedI18nService.trySetLocale(anything(), anything())).never();
+    verify(mockedAuthService.updateInterfaceLanguage(anything())).never();
   }));
 
   it('sets user locale when stored locale does not match the browsing session', fakeAsync(() => {
@@ -204,7 +204,7 @@ describe('AppComponent', () => {
 
     tick();
     env.fixture.detectChanges();
-    verify(mockedI18nService.trySetLocale('es', anything())).once();
+    verify(mockedAuthService.updateInterfaceLanguage('es')).once();
 
     env.component.setLocale('pt-BR');
     tick();
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -4,7 +4,6 @@ import { ComponentFixture, fakeAsync, flush, TestBed, tick } from '@angular/core
 import { By } from '@angular/platform-browser';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { Route, Router, RouterModule } from '@angular/router';
-import { CookieService } from 'ngx-cookie-service';
 import { SystemRole } from 'realtime-server/lib/esm/common/models/system-role';
 import { User } from 'realtime-server/lib/esm/common/models/user';
 import { createTestUser } from 'realtime-server/lib/esm/common/models/user-test-data';
@@ -13,14 +12,17 @@ import { SFProjectRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-
 import { createTestProject } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-test-data';
 import { TextInfo } from 'realtime-server/lib/esm/scriptureforge/models/text-info';
 import { BehaviorSubject, firstValueFrom, Subject } from 'rxjs';
+import { filter } from 'rxjs/operators';
 import { anything, capture, mock, verify, when } from 'ts-mockito';
 import { AuthService, LoginResult } from 'xforge-common/auth.service';
 import { AvatarComponent } from 'xforge-common/avatar/avatar.component';
 import { BugsnagService } from 'xforge-common/bugsnag.service';
 import { DialogService } from 'xforge-common/dialog.service';
 import { ErrorReportingService } from 'xforge-common/error-reporting.service';
+import { ExternalUrlService } from 'xforge-common/external-url.service';
 import { createTestFeatureFlag, FeatureFlagService } from 'xforge-common/feature-flags/feature-flag.service';
 import { FileService } from 'xforge-common/file.service';
+import { I18nService } from 'xforge-common/i18n.service';
 import { LocationService } from 'xforge-common/location.service';
 import { UserDoc } from 'xforge-common/models/user-doc';
 import { NoticeService } from 'xforge-common/notice.service';
@@ -33,7 +35,6 @@ import { TestRealtimeService } from 'xforge-common/test-realtime.service';
 import { configureTestingModule, TestTranslocoModule } from 'xforge-common/test-utils';
 import { UICommonModule } from 'xforge-common/ui-common.module';
 import { UserService } from 'xforge-common/user.service';
-import { filter } from 'rxjs/operators';
 import { AppComponent, CONNECT_PROJECT_OPTION } from './app.component';
 import { SFProjectProfileDoc } from './core/models/sf-project-profile-doc';
 import { SF_TYPE_REGISTRY } from './core/models/sf-type-registry';
@@ -52,10 +53,11 @@ const mockedNmtDraftAuthGuard = mock(NmtDraftAuthGuard);
 const mockedUsersAuthGuard = mock(UsersAuthGuard);
 const mockedSFProjectService = mock(SFProjectService);
 const mockedBugsnagService = mock(BugsnagService);
-const mockedCookieService = mock(CookieService);
 const mockedLocationService = mock(LocationService);
 const mockedNoticeService = mock(NoticeService);
 const mockedPwaService = mock(PwaService);
+const mockedI18nService = mock(I18nService);
+const mockedUrlService = mock(ExternalUrlService);
 const mockedFileService = mock(FileService);
 const mockedErrorReportingService = mock(ErrorReportingService);
 const mockedDialogService = mock(DialogService);
@@ -101,10 +103,11 @@ describe('AppComponent', () => {
       { provide: UsersAuthGuard, useMock: mockedUsersAuthGuard },
       { provide: SFProjectService, useMock: mockedSFProjectService },
       { provide: BugsnagService, useMock: mockedBugsnagService },
-      { provide: CookieService, useMock: mockedCookieService },
       { provide: LocationService, useMock: mockedLocationService },
       { provide: NoticeService, useMock: mockedNoticeService },
       { provide: PwaService, useMock: mockedPwaService },
+      { provide: I18nService, useMock: mockedI18nService },
+      { provide: ExternalUrlService, useMock: mockedUrlService },
       { provide: OnlineStatusService, useClass: TestOnlineStatusService },
       { provide: FeatureFlagService, useMock: mockedFeatureFlagService },
       { provide: FileService, useMock: mockedFileService },
@@ -112,6 +115,11 @@ describe('AppComponent', () => {
       { provide: DialogService, useMock: mockedDialogService }
     ]
   }));
+
+  afterEach(() => {
+    // suppress no expectations warning
+    expect(1).toEqual(1);
+  });
 
   it('navigate to last project', fakeAsync(() => {
     const env = new TestEnvironment();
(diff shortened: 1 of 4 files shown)


josephmyers:
> Each time the angular app is reloaded, a call is made to Auth0 to update the interface language. This was not necessarily a big problem, but it was unnecessary if the interface language still matched what was current in the browsing session.

and the new behavior

> * upon opening the app (or on refresh) and the user is logged in, the user's language is updated if needed (changed)

I'm just starting to look into this, but the Before and After here seem very similar. I'm not sure what actual high-level behavior change you've made. Can you elaborate? Thanks

josephmyers (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2509)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 236 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2509#-O--0nwo3z37JoKMW8jk:-O--0nwo3z37JoKMW8jl:b-t9szou) ([raw file](https://github.com/sillsdev/web-xforge/blob/6b8e65455aafa39223d0f328f2198046c743cd9a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L236)):*
> ```XML
>     const languageTag: string | undefined = this.currentUserDoc.data!.interfaceLanguage;
>     if (languageTag != null && I18nService.getLocale(languageTag)?.canonicalTag !== this.i18n.localeCode) {
>       this.i18n.trySetLocale(this.i18n.localeCode, this.authService);
> ```

I feel like, based on the comment, this should be using `languageTag` from the `currentUserDoc`, as before. Is that equivalent to the `localeCode`? It's strange that we're setting i18n with a property from i18n.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts` line 189 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2509#-O--05no9dok9zwUIROO:-O--05no9dok9zwUIROP:b-jrrump) ([raw file](https://github.com/sillsdev/web-xforge/blob/6b8e65455aafa39223d0f328f2198046c743cd9a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts#L189)):*
> ```XML
>   }));
> 
>   it('does not set user locale', fakeAsync(() => {
> ```

Nit: Test names should specify when this does not set user locale, and when it does, below.

josephmyers:
So your acceptance tests do work, but I'm confused about another case.

image

On master, refreshing one of these pages will get them using the same language. On your branch, the language discrepancy persists.

RaymondLuong3 (commented):
> I'm just starting to look into this, but the Before and After here seem very similar. I'm not sure what actual high-level behavior change you've made. Can you elaborate? Thanks

The change is that previously there was no check and the app would call the updateInterfaceLanguage Auth0 API every time the app is reloaded. Now it will only make the call if the language is different from the current browsing session.

> On master, refreshing one of these pages will get them using the same language. On your branch, the language discrepancy persists.

Yes, that was intentional. We are assuming that the cookie set in a browser is the authority on which language the user wants. So if that user chooses to have 2 languages on 2 different browsers, we will not try to fight them. The language stored in Auth0 will be based on the browser that was most recently reloaded (which will usually coincided with which language the users most recently used the app in).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2509#-:-O-2Ie5g7HwUvtIrxeHi:b-1rryej)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 236 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2509#-O--0nwo3z37JoKMW8jk:-O-2K50-5UjXvt3Aiyb9:b-jcc66u) ([raw file](https://github.com/sillsdev/web-xforge/blob/6b8e65455aafa39223d0f328f2198046c743cd9a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L236)):*

I feel like, based on the comment, this should be using `languageTag` from the `currentUserDoc`, as before. Is that equivalent to the `localeCode`? It's strange that we're setting i18n with a property from i18n.
</blockquote></details>

Yes, that does feel strange to set `i18n` to something within `i18n`.
The change I am proposing is for SF to rely solely on `i18n` and our browsing session to determine the user's interface language. We do not want to depend on the `user.interfaceLanguage` for that info, which is just a shortcut to what is recorded in Auth0.

josephmyers (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2509)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 236 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2509#-O--0nwo3z37JoKMW8jk:-O-4ZeJS8HTd_e8NQ4Sh:bhwwab9) ([raw file](https://github.com/sillsdev/web-xforge/blob/6b8e65455aafa39223d0f328f2198046c743cd9a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L236)):*
I'm assuming that you have more knowledge about this and have determined this to be the best fix. But it looks odd. Why does it have to be a manual update step here? Is there no way that `i18n` can update itself, e.g. as it starts up? This is from its constructor:

>     // Note that if the user is already logged in, and the user has a different interface language specified in their
    // Auth0 profile, then the locale from the URL will end up being overridden.
    const urlLocale \= new URLSearchParams(locationService.search).get('locale');
    if (urlLocale != null) {
      this.trySetLocale(urlLocale);
    } else if (!ignoreCookieLocale) {
      const cookieLocale \= this.cookieService.get(ASP\_CULTURE\_COOKIE\_NAME);
      if (cookieLocale != null) {
        this.trySetLocale(getAspCultureCookieLanguage(cookieLocale));
      }
    }

It appears to be trying to let the cookie take precedence over what's in the URL. In my ideal world, we put your fix, the real fix, here, instead of non-obviously overwriting (in some cases) what this code is indicating it's doing. Whether we need both or just yours, it still should probably go around here.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2509)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 236 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2509#-O--0nwo3z37JoKMW8jk:-O-8Cgfy6qeFZafQhDZ2:b-ihiaqf) ([raw file](https://github.com/sillsdev/web-xforge/blob/6b8e65455aafa39223d0f328f2198046c743cd9a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L236)):*

I'm assuming that you have more knowledge about this and have determined this to be the best fix. But it looks odd. Why does it have to be a manual update step here? Is there no way that `i18n` can update itself, e.g. as it starts up? This is from its constructor:

>     // Note that if the user is already logged in, and the user has a different interface language specified in their
    // Auth0 profile, then the locale from the URL will end up being overridden.
    const urlLocale \= new URLSearchParams(locationService.search).get('locale');
    if (urlLocale != null) {
      this.trySetLocale(urlLocale);
    } else if (!ignoreCookieLocale) {
      const cookieLocale \= this.cookieService.get(ASP\_CULTURE\_COOKIE\_NAME);
      if (cookieLocale != null) {
        this.trySetLocale(getAspCultureCookieLanguage(cookieLocale));
      }
    }

It appears to be trying to let the cookie take precedence over what's in the URL. In my ideal world, we put your fix, the real fix, here, instead of non-obviously overwriting (in some cases) what this code is indicating it's doing. Whether we need both or just yours, it still should probably go around here.
</blockquote></details>

Yes, it would be reasonable to make a change in the constructor on the `i18nService`, but we quickly get into a tangle...
- The `i18nService` is a dependency for `authService`, so we cannot inject it here.

Also, the comment about the URL being overridden is no longer true. The result of the change will require that the `app.component` updates the user's interface language in `Auth0` when loaded, but only when it is necessary. Perhaps the `i18nService` should not require that from the `app.component`, but the circular dependency problem is forcing us to do it differently.

josephmyers (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2509)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 236 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2509#-O--0nwo3z37JoKMW8jk:-O-9LNsd4kKmP3BeiNHn:b32o8sa) ([raw file](https://github.com/sillsdev/web-xforge/blob/6b8e65455aafa39223d0f328f2198046c743cd9a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L236)):*

Yes, it would be reasonable to make a change in the constructor on the `i18nService`, but we quickly get into a tangle...
- The `i18nService` is a dependency for `authService`, so we cannot inject it here.

Also, the comment about the URL being overridden is no longer true. The result of the change will require that the `app.component` updates the user's interface language in `Auth0` when loaded, but only when it is necessary. Perhaps the `i18nService` should not require that from the `app.component`, but the circular dependency problem is forcing us to do it differently.
</blockquote></details>

Thanks, I do see the circular dependency now. We can chalk that up to `i18n` being responsible for both managing locale and translating strings. If we split this up, we wouldn't have the circular dependency. Also, the dependency in `authService` is a tiny one; we could do something with that.

But the main point here: it's still weird to call an update on `i18n` with its own property. We should fix that. And given the code I posted above, we know that `trySetLocale` has already gotten called on startup, whether with cookie or URL. So in the case where auth0 has a different value than the cookie, we call `trySetLocale` with the same value twice, the second time being redundant but for the fact that the second call passes the `authService`, which calls `updateInterfaceLanguage`. This is the critical part.

I'm sure you know all this already, but what I'm getting at is that we should be able to swap out the `trySetLocale` with
`this.authService.updateInterfaceLanguage(this.i18n.localeCode);`

It even lines up better with your code comment. Thoughts?

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2509)* status: 2 of 4 files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 236 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2509#-O--0nwo3z37JoKMW8jk:-O-D9thl7g1ta2WF4-C_:bflcgtw) ([raw file](https://github.com/sillsdev/web-xforge/blob/6b8e65455aafa39223d0f328f2198046c743cd9a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L236)):*

Thanks, I do see the circular dependency now. We can chalk that up to `i18n` being responsible for both managing locale and translating strings. If we split this up, we wouldn't have the circular dependency. Also, the dependency in `authService` is a tiny one; we could do something with that.

But the main point here: it's still weird to call an update on `i18n` with its own property. We should fix that. And given the code I posted above, we know that `trySetLocale` has already gotten called on startup, whether with cookie or URL. So in the case where auth0 has a different value than the cookie, we call `trySetLocale` with the same value twice, the second time being redundant but for the fact that the second call passes the `authService`, which calls `updateInterfaceLanguage`. This is the critical part.

I'm sure you know all this already, but what I'm getting at is that we should be able to swap out the `trySetLocale` with
`this.authService.updateInterfaceLanguage(this.i18n.localeCode);`

It even lines up better with your code comment. Thoughts?
</blockquote></details>

I think you are right. Since `i18n.service` will initialize and set the current language without updating it in `auth.service`, we can already assume there is no need to call `i18n.trySetLocale`. We can go right to called `authService.updateInterfaceLanguage` directly from the `AppComponent`. Thanks for pointing that out. I've updated the PR with your suggestion.

josephmyers (approved):
Cool, this looks good. Thanks for working with me on this

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2509#-:-O-EWeJR-5jtwLZpWL_N:b-6a1g35)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2511, opened 2024-06-06
SF-2758 Stack import questions dialog buttons on small devices

RaymondLuong3:
Before

Import questions dialog before

After

Import Questions Dialog After

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2511)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss
@@ -57,6 +57,11 @@
     margin-block-end: 16px;
   }
 
+  mat-card-actions {
+    flex-wrap: wrap;
+    row-gap: 8px;
+  }
+
   ol {
     padding-left: 1.25em;
   }


nigel-wells (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2511)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss` line 61 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2511#-Nzk_Zr707Xv_Z9A5ZS4:-Nzk_Zr707Xv_Z9A5ZS5:b-yyz3jt) ([raw file](https://github.com/sillsdev/web-xforge/blob/cbebb0f3e6f16f2e52180d44fdd6efe5d51f529e/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss#L61)):*
> ```SCSS
> 
>   mat-card-actions {
>     display: unset;
> ```

I'd suggest we use `flex-wrap: wrap;` instead. The card actions is using flex which includes some nice alignment features which we'd want to keep.

You could optional add a `gap` if you felt the buttons were too close together as they are technically touching but the background on one is white so it looks ok.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2511)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss` line 61 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2511#-Nzk_Zr707Xv_Z9A5ZS4:-NznpObLBJc1_lSwrowQ:b-k6u8gq) ([raw file](https://github.com/sillsdev/web-xforge/blob/cbebb0f3e6f16f2e52180d44fdd6efe5d51f529e/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss#L61)):*
`display: unset` will virtually never be the best option. See the definition of unset at MDN https://developer.mozilla.org/en-US/docs/Web/CSS/unset:

> The **`unset`** CSS keyword resets a property to its inherited value if the property naturally inherits from its parent, and to its [initial value](https://developer.mozilla.org/en-US/docs/Web/CSS/initial_value) if not. In other words, it behaves like the [`inherit`](https://developer.mozilla.org/en-US/docs/Web/CSS/inherit) keyword in the first case, when the property is an [inherited property](https://developer.mozilla.org/en-US/docs/Web/CSS/Inheritance#inherited_properties), and like the [`initial`](https://developer.mozilla.org/en-US/docs/Web/CSS/initial) keyword in the second case, when the property is a [non-inherited property](https://developer.mozilla.org/en-US/docs/Web/CSS/Inheritance#non-inherited_properties).

Understanding what this line does requires:
1. Knowing the meaning of `unset` defined above
2. Knowing whether the property at hand is inherited or not ([MDN says it's not](https://developer.mozilla.org/en-US/docs/Web/CSS/display#formal_definition))
3. Knowing what the default value is (`inline`, [according to MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/display#formal_definition))

So, this will be the same as setting `display: inline`. Except that it won't; if you check dev tools you'll find it ends up being a block element, even though the default value for `display` is `inline`. I *think* this is because being a child of a flex container changes the default display value, but I can't even find the documentation that says that.

Tl;dr: Be as explicit about what value a CSS property should be set to as possible. 

I would concur with Nigel's suggestion, and add that when you hover the button you can see that there's no gap with the other button, so adding a 8px gap would probably be a good idea.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2511)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss` line 61 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2511#-Nzk_Zr707Xv_Z9A5ZS4:-O-2DxxQAdefk0ERp6im:bg44ixj) ([raw file](https://github.com/sillsdev/web-xforge/blob/cbebb0f3e6f16f2e52180d44fdd6efe5d51f529e/src/SIL.XForge.Scripture/ClientApp/src/app/checking/import-questions-dialog/import-questions-dialog.component.scss#L61)):*

`display: unset` will virtually never be the best option. See the definition of unset at MDN https://developer.mozilla.org/en-US/docs/Web/CSS/unset:

> The **`unset`** CSS keyword resets a property to its inherited value if the property naturally inherits from its parent, and to its [initial value](https://developer.mozilla.org/en-US/docs/Web/CSS/initial_value) if not. In other words, it behaves like the [`inherit`](https://developer.mozilla.org/en-US/docs/Web/CSS/inherit) keyword in the first case, when the property is an [inherited property](https://developer.mozilla.org/en-US/docs/Web/CSS/Inheritance#inherited_properties), and like the [`initial`](https://developer.mozilla.org/en-US/docs/Web/CSS/initial) keyword in the second case, when the property is a [non-inherited property](https://developer.mozilla.org/en-US/docs/Web/CSS/Inheritance#non-inherited_properties).

Understanding what this line does requires:
1. Knowing the meaning of `unset` defined above
2. Knowing whether the property at hand is inherited or not ([MDN says it's not](https://developer.mozilla.org/en-US/docs/Web/CSS/display#formal_definition))
3. Knowing what the default value is (`inline`, [according to MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/display#formal_definition))

So, this will be the same as setting `display: inline`. Except that it won't; if you check dev tools you'll find it ends up being a block element, even though the default value for `display` is `inline`. I *think* this is because being a child of a flex container changes the default display value, but I can't even find the documentation that says that.

Tl;dr: Be as explicit about what value a CSS property should be set to as possible. 

I would concur with Nigel's suggestion, and add that when you hover the button you can see that there's no gap with the other button, so adding a 8px gap would probably be a good idea.
</blockquote></details>

Great idea. Thanks for the info. Yes, I agree that whenever we can be explicit, it is best to do that because it will limit the chance of misunderstanding. I'll make these improvements.

nigel-wells (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2511#-:-O-3ntt8DZXQuq_Rhp4m:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2512, opened 2024-06-06
SF-2761 Remove checkmark when selected books for drafting and training

RaymondLuong3:
This is how the chips look without the checkmark
image

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.scss
@@ -5,6 +5,12 @@
     &.mat-mdc-chip-selected,
     &.mat-mdc-chip-highlighted {
       --mdc-chip-elevated-container-color: #4a79c9;
+      ::ng-deep button {
+        padding: 0 12px;
+        .mat-mdc-chip-graphic {
+          display: none;
+        }
+      }
     }
   }
 }


nigel-wells:
I assume there was no setting to turn it off other than CSS? No other component we should use instead?

RaymondLuong3 (commented):
That's right. Looking through the components, chips appear to be the best option because they natively have a selected state. I don't see a better option, but definitely open to ideas.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2512#-:-O-2D5KI0nIa1EGZszIZ:b-qaz8bf)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @nigel-wells)

Nateowami:
@RaymondLuong3 Please read the instructions in the Jira issue.

RaymondLuong3 (commented):
I misunderstood the instructions and missed the additional training data chip from the `training-data-multi-select` component. It is added now.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2512#-:-O-7TA1V-jvvdlZu5WXk:b-c95f7b)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @nigel-wells)

nigel-wells (commented):
I see there is this @Input available `hideSingleSelectionIndicator` - can you use that instead of hiding it with CSS. This is only allowed when not using `multiple`. It can work if you use the `ngFor` on the `mat-chip-listbox` instead of `mat-chip-option` which then gives you x number of single selection listboxes. It works, and doesn't involve hacking it away with CSS, but I'm not convinced it is better as it is also a workaround. It does keep us within their APIs though which is a positive.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2512#-:-O-8lykhEOiYAkvEVMVI:brg46k)* status: 0 of 2 files reviewed, all discussions resolved

RaymondLuong3 (commented):
Thanks for finding that. Using the `hideSingleSelectionIndicator` is nice if they ever update the component, it is less likely to break our chips here. I've updated the components so it is using x number of `mat-chip-listbox` for x number of options. Some of the logic for `(changes)` event needed to be handled differently, but maybe that made it slightly simpler to understand. Here are the updated images.




> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2512#-:-O-D4qFP7iHg3qJB_bto:b-fqnfw9)* status: 0 of 8 files reviewed, all discussions resolved (waiting on @nigel-wells)

nigel-wells (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2512#-:-O-Z1bLG0sx2dJepua3S:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2513, opened 2024-06-07
SF-2770 Fix message dialog resolving before user closes dialog

Nateowami:
Some places we show the message dialog and don't need to wait for the user to close it, but in some places we take an action after the user closes it, for example, when a share key is invalid.

To test this, on master, in a browser where you are not logged in, open this bad share link: http://localhost:5000/join/1234567890
A message dialog will briefly open, but before you can read it you will be redirected to the home page. Then test again on this branch.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2513)

PR #2514, opened 2024-06-07
SF-2771 Inform user when link has been used already

Nateowami:
This is _kind of_ dependent on #2513, in the sense that the dialog won't properly show without that change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2514)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/AnonymousService.cs
+++ b/src/SIL.XForge.Scripture/Services/AnonymousService.cs
@@ -29,7 +29,7 @@ public class AnonymousService : IAnonymousService
         // Ensure the key hasn't been used by another recipient
         if (validShareKey.ShareKey.RecipientUserId != null)
         {
-            throw new ForbiddenException();
+            throw new DataNotFoundException("key_already_used");
         }
 
         return new AnonymousShareKeyResponse
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts
@@ -157,8 +157,8 @@ export class JoinComponent extends DataLoadingComponent {
     this.name.setValidators([Validators.required, XFValidators.someNonWhitespace]);
     try {
       this.joiningResponse = await this.anonymousService.checkShareKey(shareKey);
-    } catch {
-      await this.informInvalidShareLinkAndRedirect();
+    } catch (error) {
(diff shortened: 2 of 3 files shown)


src/SIL.XForge.Scripture/Services/AnonymousService.cs
```diff
@@ -29,7 +29,7 @@ public async Task<AnonymousShareKeyResponse> CheckShareKey(string shareKey)
         // Ensure the key hasn't been used by another recipient
         if (validShareKey.ShareKey.RecipientUserId != null)
         {
-            throw new ForbiddenException();
+            throw new DataNotFoundException("key_already_used");
```
Nateowami, line 32: I'm using `DataNotFoundException` because it's being used at https://github.com/sillsdev/web-xforge/blob/3498a653361fa67cddffb7a6035c17d44bace2fc/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L771 for the same error message.

Nateowami:
@pmachapman or @nigel-wells No pressure, but if it's possible to get this reviewed today, we can merge it before testing. There are some teams running into a "link invalid" error dialog, and I think it's likely the key has just been used already, and this change would allow us to see if that is indeed the case.

pmachapman (approved):
:lgtm: I tested this locally, and it worked as expected.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2514#-:-Nzz6WBb3HQtyi5Pw1Mq:b1h1atb)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/Services/AnonymousService.cs` line 32 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2514#gh-1631747376:-Nzz85Np9HnqF_vN1T3c:bt6fllm) ([raw file](https://github.com/sillsdev/web-xforge/blob/1fac320114bda14887c90bb84b7ae2e5f87edc66/src/SIL.XForge.Scripture/Services/AnonymousService.cs#L32)):*

I'm using `DataNotFoundException` because it's being used at https://github.com/sillsdev/web-xforge/blob/3498a653361fa67cddffb7a6035c17d44bace2fc/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L771 for the same error message.
</blockquote></details>

Looks good - as `SFProjectService.CheckShareKeyValidity()` only throws `DataNotFoundException`, this is a good change.

PR #2518, opened 2024-06-12
SF-2778 Fix digits in USFM filenames

pmachapman:
The filename for USFM files is in the format 01GENPROJ.SFM.

However, for New Testament books, the number is off by one. For example, Revelation should be 67REVPROJ.SFM, not 66REVPROJ.SFM (as is currently the generated file name when downloading a draft as a zip file).

This PR corrects the numbers in the download draft zip file names to reflect the same numbering convention as Paratext.

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.spec.ts
@@ -2,7 +2,14 @@ import { VerseRef } from '@sillsdev/scripture';
 import { SFProject } from 'realtime-server/lib/esm/scriptureforge/models/sf-project';
 import { DeltaOperation } from 'rich-text';
 import { SelectableProject } from '../core/paratext.service';
-import { compareProjectsForSorting, getVerseNumbers, isBadDelta, projectLabel, XmlUtils } from './utils';
+import {
+  compareProjectsForSorting,
+  getBookFileNameDigits,
+  getVerseNumbers,
+  isBadDelta,
+  projectLabel,
+  XmlUtils
+} from './utils';
 
 describe('shared utils', () => {
   describe('projectLabel function', () => {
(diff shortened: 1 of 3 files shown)


nigel-wells (commented):
Interesting - why the specific logic around changing the book numbers?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2518#-:-O-9YTZkDTzRQqQdnlOv:bbon9pw)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts` line 64 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2518#-O-9Y2-LBHA-JB9OxRJM:-O-9Y2-LBHA-JB9OxRJN:b-ikxowe) ([raw file](https://github.com/sillsdev/web-xforge/blob/1fab77531551ab1240c804bd13568510e15a4897/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L64)):*
> ```XML
>  * Gets the digit code for Paratext filenames, such as 01 for GEN, 39 for MAL, and 41 for MAT.
>  *
>  * These are all 2 digits, except for xxg which is "100":
> ```

So does "xxg" book 99 as that returns 100? Or should book 100 return "xxg"?

pmachapman (commented):
> Interesting - why the specific logic around changing the book numbers?

I am not sure of the history of the logic, but when testing Serval, I noticed that the number was off by one for the NT book I was drafting. Further investigation revealed this logic in Paratext, which I have ported to TypeScript.

This numbering is important if someone wishes copy-paste this file into their Paratext project directory, and have the old file overwritten.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2518#-:-O-Cpuwd-TBj34b88FNp:b-a52291)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts` line 64 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2518#-O-9Y2-LBHA-JB9OxRJM:-O-CqV6R2FWCd4NYeLr2:b-j2g47r) ([raw file](https://github.com/sillsdev/web-xforge/blob/1fab77531551ab1240c804bd13568510e15a4897/src/SIL.XForge.Scripture/ClientApp/src/app/shared/utils.ts#L64)):*

So does "xxg" book 99 as that returns 100? Or should book 100 return "xxg"?
</blockquote></details>

Yes, XXG (book number 99) will return 100. This comment is from the Paratext code - let me know if it can be clearer.

PR #2531, opened 2024-06-19
Add index to sf_project_secrets.shareKeys.key

pmachapman:
This PR adds an index to the `shareKeys.key` field of `sf_project_secrets`.

This index will speed up checking for share keys when users attempt to log in via a share key.

When running this PR, you will see a new index appear in the `sf_project_secrets` collection:
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2531)

### DIFF
--- a/src/SIL.XForge.Scripture/DataAccess/SFDataAccessServiceCollectionExtensions.cs
+++ b/src/SIL.XForge.Scripture/DataAccess/SFDataAccessServiceCollectionExtensions.cs
@@ -1,6 +1,7 @@
 using Microsoft.Extensions.Configuration;
 using MongoDB.Driver;
 using SIL.XForge.DataAccess;
+using SIL.XForge.Models;
 using SIL.XForge.Scripture.Models;
 
 namespace Microsoft.Extensions.DependencyInjection;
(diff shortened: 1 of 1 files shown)


nigel-wells:
@pmachapman is there another half to this PR for how the frontend handles this? I'm getting errors on the users page as well as the share dialog. I've cleared all local data by logging out and back in. I've also completely removed `sf_project_secrets` in case that assumed no old data was there.

pmachapman:
>@pmachapman is there another half to this PR for how the frontend handles this?

No, not at all. This is purely a backend change. Something else might be a miss - have you run the [cleaning procedure](https://github.com/sillsdev/web-xforge/wiki/Cleaning)?

nigel-wells:
Yes I had done cleanup. I also tried switching projects in case there was bad data but that didn't help either.

I was able to resolve the issue by completely dropping the `xforge` DB from Mongo. Bit excessive but maybe I did have some bad data in there.

It is working as I'd expect now. It wouldn't hurt to get another dev to checkout the branch just to confirm it was a me problem and nothing more serious. @Nateowami are you able to see if it works ok for you?

nigel-wells:
@josephmyers do you mind doing a sanity test on this as well please just to confirm that the issue I encountered was only related to me.

josephmyers (approved):
Yep, I can see the index without dropping the db.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2531#-:-O0vUwYy9uPVaOUXKmrB:b-ozacle)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

PR #2540, opened 2024-06-21
SF-2319a Remove asterisk from required form fields

Nateowami:
James Post pointed out that having an asterisk here is not particularly helpful and doesn't have a corresponding asterisk:

I understand the usefulness of denoting required form fields in a long form, but I don't think it's needed in the context of a one-field form, as it should be pretty obvious to the user that the field is required.

The longest form I could find is this:

And I don't think the asterisks are particularly meaningful here either. And if a user is at all confused and tries to continue, it's made explicit that these fields are required.

I think a proper use of asterisks to denote required form fields is in a long(ish) form where it is stated at the beginning that the asterisks denote required fields.

In addition to reviewing code changes, please make sure to review visual changes in Chromatic.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2540)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.module.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.module.ts
@@ -2,7 +2,6 @@ import { OverlayContainer } from '@angular/cdk/overlay';
 import { DatePipe } from '@angular/common';
 import { HttpClientModule } from '@angular/common/http';
 import { ErrorHandler, NgModule } from '@angular/core';
-import { MAT_FORM_FIELD_DEFAULT_OPTIONS } from '@angular/material/form-field';
 import { BrowserModule } from '@angular/platform-browser';
 import { BrowserAnimationsModule } from '@angular/platform-browser/animations';
 import { ServiceWorkerModule } from '@angular/service-worker';
@@ -85,11 +84,7 @@ import { UsersModule } from './users/users.module';
     translocoMarkupRouterLinkRenderer(),
     defaultTranslocoMarkupTranspilers(),
     { provide: ErrorHandler, useClass: ExceptionHandlingService },
-    { provide: OverlayContainer, useClass: InAppRootOverlayContainer },
-    {
-      provide: MAT_FORM_FIELD_DEFAULT_OPTIONS,
-      useValue: { appearance: 'outline' }
-    }
+    { provide: OverlayContainer, useClass: InAppRootOverlayContainer }
   ],
   bootstrap: [AppComponent]
 })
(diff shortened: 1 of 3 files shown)


Nateowami:
Actually, none of the stories changed in appearance, because the defaults are injected into the AppComponent, which doesn't impact the stories. This seems like a flaw in how we have our stories configured. We should probably have some defaults that we inject globally into all stories, and the app component.

pmachapman (approved):
:lgtm:

I did a search for `Validators.required` to find other controls affected by this, and the omission of the asterisk in the other places I tested was logical, and did not make affected form fields feel "less required".

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2540#-:-O06E_k4Cuv742r43YEn:b-fz81c7)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

pmachapman:
> Actually, none of the stories changed in appearance, because the defaults are injected into the AppComponent, which doesn't impact the stories. This seems like a flaw in how we have our stories configured. We should probably have some defaults that we inject globally into all stories, and the app component.

Should we inject the form field defaults into `preview.ts` or `I18nStoryModule`?
```ts
    {
      provide: MAT_FORM_FIELD_DEFAULT_OPTIONS,
      useValue: { appearance: 'outline', hideRequiredMarker: true }
    }
```

Nateowami (commented):
I just moved it to the UICommonModule, which does get imported into the stories. I think that's where it makes the most sense to put a UI config anyway.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2540#-:-O0ABg_t3t1QyOpsVNmR:b-2culoy)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts` line 473 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2540#-O0AQxEcCZq6M8BnyYd7:-O0AQxEcCZq6M8BnyYd8:b8pxm5d) ([raw file](https://github.com/sillsdev/web-xforge/blob/f86c4223490854b3dff4e39106f387d8f88c81fb/src/SIL.XForge.Scripture/ClientApp/src/app/shared/share/share-control.component.spec.ts#L473)):*
> ```XML
>     flush();
>     this.fixture.detectChanges();
>     flush();
> ```

I don't really understand why this was necessary, but 11 tests fail without it. I would have thought the NoopAnimationsModule would have mean that there wouldn't be any meaningful difference between testing the default and outline versions of the form fields.

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2540#-:-O0AgpXe5MUHCoePveAf:bnfp4nl)* status: all files reviewed, 1 unresolved discussion

PR #2545, opened 2024-06-25
Make error ID in error dialog monospace

Nateowami:
This PR could also be title *Add new i18n interpolation function*, as that's what most of the change ended up being.

---

Users often send us screenshots of the error dialog, requiring me to use an OCR to extract the error ID from the error dialog. The OCR often makes mistakes, and I think making the error ID monospaced may help with that.

Trying to add that sent me down a rabbit hole to add better support for mixing localized strings with HTML elements.

`I18nService` has `translateTextAroundTemplateTags` which works with strings like this:
``` js
"Administrators can always invite people from the {{ templateTagBoundary }}Users{{ templateTagBoundary }} page."
```
The drawbacks are:
1. Requires planning ahead of time to use `templateTagBoundary` prior to translators translating the string.
2. Only works for a single item in the string (it returns an object of the type `{ before: string; templateTagText: string; after: string; }`)

`I18nService` also has `interpolate`, which can work with strings of the form:
``` js
"For more information, { 1 } see the help site{ 2 }.",
```
This is great for when the text you want to format also needs to be translated. However, we don't have a good way (before this PR) to handle this very common type of string:
``` js
"Email {{ email }} for help."
```

A future improvement may be to combine `I18nService.interpolate` with the new `I18nService.interpolateVariables`, so this format of string will be supported:
``` js
"Write to {{ email }} if you need help { 1 }connecting a project.{ 2 }"
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2545)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-dialog/error-dialog.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-dialog/error-dialog.component.html
@@ -4,7 +4,12 @@
     <p>{{ data.message }}</p>
     <p [innerHTML]="t('to_report_issue_email', { issueEmailLink: issueEmailLink })"></p>
     <p *ngIf="browserUnsupported" class="unsupported-browser" [innerHTML]="t('unsupported_browser', browserLinks)"></p>
-    <p class="error-id">{{ t("error_id", { errorId: data.eventId }) }}</p>
+    <p>
+      <ng-container *ngFor="let portion of i18n.interpolateVariables('error.error_id', { errorId: data.eventId })">
+        <ng-container *ngIf="portion.id == null">{{ portion.text }}</ng-container>
+        <span *ngIf="portion.id === 'errorId'" class="error-id">{{ portion.text }}</span>
+      </ng-container>
+    </p>
 
     <ng-container *ngIf="data.stack">
       <a (click)="this.showDetails = !this.showDetails">{{ showDetails ? t("hide_details") : t("show_details") }}</a>
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-dialog/error-dialog.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-dialog/error-dialog.component.scss
@@ -15,3 +15,7 @@ mat-dialog-content {
     overflow-x: auto;
   }
 }
+
+.error-id {
+  font-family: monospace;
+}
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-dialog/error-dialog.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-dialog/error-dialog.component.spec.ts
@@ -4,13 +4,18 @@ import { NgModule } from '@angular/core';
 import { ComponentFixture, fakeAsync, flush, TestBed } from '@angular/core/testing';
 import { MatDialog, MatDialogConfig } from '@angular/material/dialog';
 import { NoopAnimationsModule } from '@angular/platform-browser/animations';
+import { mock } from 'ts-mockito';
+import { FeatureFlagService } from '../feature-flags/feature-flag.service';
 import { ChildViewContainerComponent, configureTestingModule, TestTranslocoModule } from '../test-utils';
 import { UICommonModule } from '../ui-common.module';
 import { ErrorAlertData, ErrorDialogComponent } from './error-dialog.component';
 
+const mockedFeatureFlagService = mock(FeatureFlagService);
+
 describe('ErrorDialogComponent', () => {
   configureTestingModule(() => ({
-    imports: [DialogTestModule]
+    imports: [DialogTestModule],
+    providers: [{ provide: FeatureFlagService, useMock: mockedFeatureFlagService }]
   }));
 
   let overlayContainer: OverlayContainer;
@@ -35,7 +40,7 @@ describe('ErrorDialogComponent', () => {
 
     expect(env.errorMessage.textContent).toBe(dialogData.message);
     expect(env.showDetails?.textContent).toBe('Show details');
-    expect(env.errorId.textContent).toBe(`Error ID: ${dialogData.eventId}`);
+    expect(env.errorId.textContent).toBe(dialogData.eventId);
     expect(env.stackTrace).toBeNull();
 
     env.showDetails?.click();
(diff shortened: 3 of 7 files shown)


pmachapman (changes requested):
Looks good (and runs great) - just one question below...

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2545#-:-O0HF_9Q8fZpI9WGbMLF:bvbb2dj)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 308 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2545#-O0HApm656xuGENjbADh:-O0HApm656xuGENjbADi:b-c7mvmo) ([raw file](https://github.com/sillsdev/web-xforge/blob/28ea5d93c3bf65b4022c970384d764344a686a30/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L308)):*
> ```XML
>     while ((match = regex.exec(translation)) !== null) {
>       matches.push(match);
>     }
> ```

As SF now targets ES2022, should we change this (and the similar code in `I18nService.interpolate()`)  to use `string.matchAll(regex)`?



_Code quote:_
```XML
    // ES2020 introduces string.matchAll(regex), but as of the time of writing SF is using ES2018
    let match: RegExpExecArray | null;
    while ((match = regex.exec(translation)) !== null) {
      matches.push(match);
    }
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2545)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 308 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2545#-O0HApm656xuGENjbADh:-O0HGMHYBR78CBB2CcL1:brhj4u) ([raw file](https://github.com/sillsdev/web-xforge/blob/28ea5d93c3bf65b4022c970384d764344a686a30/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L308)):*

As SF now targets ES2022, should we change this (and the similar code in `I18nService.interpolate()`)  to use `string.matchAll(regex)`?
</blockquote></details>

Yes, good call. I'll work on this tomorrow.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2545)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 308 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2545#-O0HApm656xuGENjbADh:-O0KNUxcC88083ipWd1E:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/28ea5d93c3bf65b4022c970384d764344a686a30/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L308)):*

Yes, good call. I'll work on this tomorrow.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 303 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2545#-O0KNeg06QLrQiQOpIJY:-O0KNeg06QLrQiQOpIJZ:b-fgbiz6) ([raw file](https://github.com/sillsdev/web-xforge/blob/6915b23c6275b78fe79c6f842ea3419449d98ff8/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L303)):*
> ```XML
>   interpolate(key: I18nKey, params?: HashMap): { text: string; id?: number }[] {
>     const hashKey = this.localeCode + ' ' + key;
>     if (this.interpolationCache[hashKey] != null) {
> ```

I got rid of this because it doesn't account for params changing.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 305 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2545#-O0KNZ9Z2SdjcH74way4:-O0KNZ9Z2SdjcH74way5:bnpd1at) ([raw file](https://github.com/sillsdev/web-xforge/blob/0575c1b665f32bd542bf7df7673e635a410fa34b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L305)):*
> ```XML
>     for (const match of translation.matchAll(regex)) {
>       // Docs clearly indicate that match.index is a number, but type definitions treat is as possibly undefined
>       if (typeof match.index !== 'number') throw new Error('match.index is not a number');
> ```

This is super ugly but I'm not sure what a better option would be.

pmachapman (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2545)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 305 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2545#-O0KNZ9Z2SdjcH74way4:-O0L4HEO5xq_A7WgDlq3:b-ofoyq9) ([raw file](https://github.com/sillsdev/web-xforge/blob/0575c1b665f32bd542bf7df7673e635a410fa34b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L305)):*

This is super ugly but I'm not sure what a better option would be.
</blockquote></details>

I think we may need to update the `lib` compiler option in `tsconfig.json` from `es2019` to fix this?

```json
    "target": "ES2022",
    "module": "es2020",
    "lib": ["es2019", "dom"],
```

See https://www.typescriptlang.org/tsconfig/#lib

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2545)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 305 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2545#-O0KNZ9Z2SdjcH74way4:-O0LGr0b1l8yXA4lXaQG:b-iohd2) ([raw file](https://github.com/sillsdev/web-xforge/blob/0575c1b665f32bd542bf7df7673e635a410fa34b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L305)):*

I think we may need to update the `lib` compiler option in `tsconfig.json` from `es2019` to fix this?

```json
    "target": "ES2022",
    "module": "es2020",
    "lib": ["es2019", "dom"],
```

See https://www.typescriptlang.org/tsconfig/#lib
</blockquote></details>

Unfortunately that does not fix it.

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2545#-:-O0M66Y32g_dpQqVsVc2:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2548, opened 2024-06-26
SF-2640 Warn when font unsupported

Nateowami:
Problem  | Project | Resource
---------|---------|---------
Fallback |  | 
Graphite |  | 

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2548)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/translate-overview/translate-overview.component.html
@@ -1,5 +1,26 @@
 <ng-container *transloco="let t; read: 'translate_overview'">
   <div class="title mat-headline-4">{{ t("translate_overview") }}</div>
+
+  <app-notice
+    *ngIf="isPTUser && !fontService.isFontFullySupported(selectedFont ?? '')"
+    type="warning"
+    mode="fill-dark"
+    class="font-warning"
+  >
+    {{ t("warn_font_unsupported", { selectedFont, fallbackFont: fontService.fontFallback(selectedFont ?? "") }) }}
+    {{ t(isPtProject ? "change_font_or_contact_for_help" : "contact_for_help") }}
+  </app-notice>
+
+  <app-notice
+    *ngIf="isPTUser && fontService.isGraphiteFont(selectedFont ?? '')"
+    type="warning"
+    mode="fill-dark"
+    class="font-warning"
+  >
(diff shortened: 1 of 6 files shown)


josephmyers:
image

If the selected font isn't present in the fallbacks, the message is messed up. I had just picked a font in PT at random. So we need a fallback for the fallbacks?

Nateowami (commented):
:facepalm: Thanks for catching that.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2548#-:-O0Mm3DdDjT5lACrOaFj:bjoxljm)* status: 0 of 12 files reviewed, all discussions resolved (waiting on @josephmyers)

Nateowami (commented):
Should be fixed now.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2548#-:-O0MmI2a2bhDjS3FUKuJ:bql6urd)* status: 0 of 12 files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers:
image

The other thing I'm seeing is the "graphite" message showing in Firefox. Are we intending that to be an informational message, or is something the matter?

josephmyers (changes requested):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2548)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami:
You can use dev tools to show what fonts were used for particular elements:


Additionally, using text that is in the script Awami Nastaliq is intended for would make it much more apparent. You can get example text with most (all?) SIL font downloads.

Nateowami:
Yes, it is intended to be shown in all browsers. I don't want the limitations to be hidden. If it works on the admin's machine, but not other users, we don't want to hide that fact.

josephmyers:
I see, good point. However, it does leave open the following questions. If you show that warning in the case above, users will wonder if there's something wrong, e.g. "Is the font actually working correctly?" "How do I know if my browser supports Graphite?" This is a lot to expect from them.

Now you _could_ show an additional message putting their concerns to rest, e.g. "Your browser fully supports this font," in the case where it's working and the warning is still shown.

Nateowami:
There is exactly one font we support that requires Graphite. And users of these writing systems tend to be acutely aware that Graphite is required to render their writing system, and that it's not supported in most systems.

josephmyers (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2548)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2551, opened 2024-06-27
Replace ID with class in history component to eliminate collisions

Nateowami:
IDs [are supposed to be unique within a page:](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id)
> The id [global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes) defines an identifier (ID) which must be unique in the whole document. Its purpose is to identify the element when linking (using a [fragment identifier](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Identifying_resources_on_the_Web#fragment)), scripting, or styling (with [CSS](https://developer.mozilla.org/en-US/docs/Glossary/CSS)).

Consequently, components that may appear multiple times on the same page ought not to use IDs. 

In practice, this is an extremely minor concern, and I'm not aware of anything that breaks due to duplicating IDs.

I found the duplicate IDs by running this code (and these were the only duplicates I was able to find):

``` ts
setInterval(() => checkForDuplicateIds(), 500);

function checkForDuplicateIds(): void {
  const ids = Array.from(document.querySelectorAll('[id]')).map(e => e.id);

  const repeats = ids.filter((id, index) => ids.indexOf(id) !== index);

  console.log(repeats);
}
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2551)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.html
@@ -1,6 +1,6 @@
 <ng-container *transloco="let t; read: 'history_chooser'">
-  <div id="toolbar" *ngIf="historyRevisions.length > 0">
-    <mat-form-field id="history-select" appearance="outline">
+  <div class="toolbar" *ngIf="historyRevisions.length > 0">
+    <mat-form-field class="history-select" appearance="outline">
       <mat-select
         [value]="selectedRevision"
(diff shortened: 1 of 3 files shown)


pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2551#-:-O0N-OA4A8x6WWZnyoLf:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2556, opened 2024-06-30
SF-2802 Make it possible to revert to the draft as it was originally created

pmachapman:
This PR adds support for reverting text documents to the points in time when either drafts were applied or the document was reverted to a point in time in history.

This is done by using the source feature of ShareDB, which although supported by the version of ShareDB we use, was not present in the type mappings. The source was also misused in `TextViewModel`, but has been corrected.

**This PR makes the following changes:**

 * Mark all TextDoc ops as coming from the editor, apply draft, or restore history.
 * Update the Revision History API to return the additional metadata.
   * Compatibility is maintained with the previous versions of the Angular frontend via two properties marked Obsolete.
 * Update the Angular frontend to utilize the new metadata and display an icon next to the revision to show its source.
 * Adds the API support for displaying a user's avatar or username next to their revision.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2556)


josephmyers (changes requested):
Great work on this so far. I think this feature will add a ton to the UX, though it may not appear so

> Reviewed 32 of 32 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-:-O0mO7dCCAst7FQe7KSO:b-npg25p)* status: all files reviewed, 6 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/paratext.service.ts` line 22 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0mHnP11Uy4neOiAG9F:-O0mHnP11Uy4neOiAG9G:b-2jcuy1) ([raw file](https://github.com/sillsdev/web-xforge/blob/436821ff63101ee27378bf10b115801a8a4c43eb/src/SIL.XForge.Scripture/ClientApp/src/app/core/paratext.service.ts#L22)):*
> ```XML
>    * This may be 'editor', 'history', 'draft', 'paratext', or null.
>    */
>   source?: string;
> ```

I am (wait for it) strongly in favor of using strong types when they exist. Can we not use `TextDocSource`, which you just made?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts` line 47 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0mFeudDwv7n3V6iqh_:-O0mFeudDwv7n3V6iqha:b-j5sb8w) ([raw file](https://github.com/sillsdev/web-xforge/blob/436821ff63101ee27378bf10b115801a8a4c43eb/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts#L47)):*
> ```XML
>       // We want to pass the source, as it will be used to mark an op as coming from draft or history
>       textDoc.submitSource = true;
>       await textDoc.submit(diff, source);
> ```

It's a little strange to me that we're now having to set a property on the textDoc before calling one of its methods. It'd be better for the textDoc itself to be able to determine this on its own. This will really help future usages.

I do see that doing this excludes the three other calls to `.submit` from submitting the source. Can you explain why they shouldn't be doing this? Wouldn't it be better to always set the metadata, with the 'restore' and 'auto-draft' options being distinguishable from the others, then having the front end only display icons given those two sources (which it's already doing). I believe this would posture our data to be more complete, future-ready, and helpful. What are your thoughts?

Or: maybe instead of this "flip on then off" style, which is pretty tribal-knowledge-y, we need a new method on the textDoc, e.g. `submitWithoutSource`. That would help highlight the oddness in behavior, i.e. `submit` accepting a source but not adding it to the metadata unless the right property is set beforehand.

Is any of this making sense? Thanks
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts` line 193 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0mIcew3lDyXc8XhiLN:-O0mIcew3lDyXc8XhiLO:bz4ro1a) ([raw file](https://github.com/sillsdev/web-xforge/blob/436821ff63101ee27378bf10b115801a8a4c43eb/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-draft/editor-draft.component.ts#L193)):*
> ```XML
>     }
> 
>     await this.textDocService.overwrite(this.textDocId!, this.draftDelta, 'draft');
> ```

Scary!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.html` line 12 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0mInqL1teIB8Sx3q9F:-O0mInqL1teIB8Sx3q9G:b-yynhcu) ([raw file](https://github.com/sillsdev/web-xforge/blob/436821ff63101ee27378bf10b115801a8a4c43eb/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.html#L12)):*
> ```HTML
>         <mat-option *ngFor="let r of historyRevisions" [value]="r">
>           <mat-icon>{{ r.source === "draft" ? "auto_awesome" : r.source === "history" ? "history" : "" }}</mat-icon>
>           {{ r | revisionFormat }}
> ```

Supporting icons and the check is a little weird to me. Additionally, I wasn't crazy about having the icons to the left, as, for one, they disalign the options with the dropdown value. So I spent a little time experimenting with what this would look like. Feel free to push back on this, but I think it's improved:



_Code snippet (i):_
```HTML
      <mat-select
        [value]="selectedRevision"
        (selectionChange)="onSelectionChanged($event)"
        panelWidth="''"
        [disabled]="isSelectDisabled$ | async"
        hideSingleSelectionIndicator
      >
        <mat-select-trigger>
          {{ selectedRevision ?? "" | revisionFormat }}
        </mat-select-trigger>
        <mat-option *ngFor="let r of historyRevisions" [value]="r">
          <div class="option-content">
            <span>{{ r | revisionFormat }}</span>
            <mat-icon>{{ r.source === "draft" ? "auto_awesome" : r.source === "history" ? "history" : "" }}</mat-icon>
          </div>
        </mat-option>
      </mat-select>
```



_Code snippet (ii):_
```HTML

::ng-deep .mdc-list-item__primary-text {
  width: 100%;
}

.mat-mdc-option.mdc-list-item--selected:not(.mdc-list-item--disabled) {
  --mat-option-selected-state-layer-color: rgba(0, 0, 0, 0.12);
}

.option-content {
  display: flex;
  align-items: center;
  gap: 8px;

  span {
    flex-grow: 1;
  }

  mat-icon {
    margin: 0;
  }
}
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.html` line 12 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0mJvP-DLOgPDEhFO6d:-O0mJvP05fRDa729-8mS:b7yw9b7) ([raw file](https://github.com/sillsdev/web-xforge/blob/436821ff63101ee27378bf10b115801a8a4c43eb/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.html#L12)):*
> ```HTML
>         <mat-option *ngFor="let r of historyRevisions" [value]="r">
>           <mat-icon>{{ r.source === "draft" ? "auto_awesome" : r.source === "history" ? "history" : "" }}</mat-icon>
>           {{ r | revisionFormat }}
> ```

Ends up like: 

___
*[`src/SIL.XForge.Scripture/Models/DocumentRevision.cs` line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0mKWX90af9kunxxHOa:-O0mKWX90af9kunxxHOb:bf839tv) ([raw file](https://github.com/sillsdev/web-xforge/blob/436821ff63101ee27378bf10b115801a8a4c43eb/src/SIL.XForge.Scripture/Models/DocumentRevision.cs#L27)):*
> ```Smalltalk
>     /// The source may be: 'editor', 'history', 'draft', 'paratext', or null.
>     /// </value>
>     public string? Source { get; init; }
> ```

Can we not make an enum to represent this type? `public const string Paratext` to represent one of the options is weird to me.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2556)* status: 16 of 36 files reviewed, 4 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/paratext.service.ts` line 22 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0mHnP11Uy4neOiAG9F:-O0pSsUuESVLRS--Gf70:b-1kupl8) ([raw file](https://github.com/sillsdev/web-xforge/blob/436821ff63101ee27378bf10b115801a8a4c43eb/src/SIL.XForge.Scripture/ClientApp/src/app/core/paratext.service.ts#L22)):*

I am (wait for it) strongly in favor of using strong types when they exist. Can we not use `TextDocSource`, which you just made?
</blockquote></details>

Done. My concern was unknown future values, but they are handled OK using this type.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts` line 47 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0mFeudDwv7n3V6iqh_:-O0pUVPc3iR5d5yrJtA_:bpzs5gv) ([raw file](https://github.com/sillsdev/web-xforge/blob/436821ff63101ee27378bf10b115801a8a4c43eb/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts#L47)):*

It's a little strange to me that we're now having to set a property on the textDoc before calling one of its methods. It'd be better for the textDoc itself to be able to determine this on its own. This will really help future usages.

I do see that doing this excludes the three other calls to `.submit` from submitting the source. Can you explain why they shouldn't be doing this? Wouldn't it be better to always set the metadata, with the 'restore' and 'auto-draft' options being distinguishable from the others, then having the front end only display icons given those two sources (which it's already doing). I believe this would posture our data to be more complete, future-ready, and helpful. What are your thoughts?

Or: maybe instead of this "flip on then off" style, which is pretty tribal-knowledge-y, we need a new method on the textDoc, e.g. `submitWithoutSource`. That would help highlight the oddness in behavior, i.e. `submit` accepting a source but not adding it to the metadata unless the right property is set beforehand.

Is any of this making sense? Thanks
</blockquote></details>

Done. I originally did this to match the behavior in ShareDB documentation for `source` and `submitSource` - I figured if I follow their practice, there it is at least a bit more transparency as to what is going on. See https://share.github.io/sharedb/api/doc#submitop

That said, on reflection, I have made it always submit the source, as I think that will be useful and it fixes the known issue (which I have removed from the PR description). Thanks!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.html` line 12 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0mInqL1teIB8Sx3q9F:-O0pXd4x8IDK2-X4uOcx:b-6l36ft) ([raw file](https://github.com/sillsdev/web-xforge/blob/436821ff63101ee27378bf10b115801a8a4c43eb/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.html#L12)):*

Supporting icons and the check is a little weird to me. Additionally, I wasn't crazy about having the icons to the left, as, for one, they disalign the options with the dropdown value. So I spent a little time experimenting with what this would look like. Feel free to push back on this, but I think it's improved:
</blockquote></details>

Done. It looks good. We could use the left hand side for the avatar icon, if an when we do that.

My only concern would be future versions of Angular possibly breaking the CSS?
___
*[`src/SIL.XForge.Scripture/Models/DocumentRevision.cs` line 27 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0mKWX90af9kunxxHOa:-O0pjmSS8ma9ug9wqOCp:b-ydonir) ([raw file](https://github.com/sillsdev/web-xforge/blob/436821ff63101ee27378bf10b115801a8a4c43eb/src/SIL.XForge.Scripture/Models/DocumentRevision.cs#L27)):*

Can we not make an enum to represent this type? `public const string Paratext` to represent one of the options is weird to me.
</blockquote></details>

Done. Good idea.

pmachapman (commented):
I have also added support for the backend to set the source, and done this for TextDoc ops from Paratext.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-:-O0q1Dhr6yMjMcR2xrbC:b4geynn)* status: 16 of 48 files reviewed, 4 unresolved discussions (waiting on @josephmyers)

josephmyers (changes requested):
> Reviewed 33 of 33 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2556)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/RealtimeServer/common/index.ts` line 252 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0qrRG74J7wSXepFumg:-O0qrRG74J7wSXepFumh:bmlrn22) ([raw file](https://github.com/sillsdev/web-xforge/blob/130044708b4be4597b582dbe904835c5ce5ab17d/src/RealtimeServer/common/index.ts#L252)):*
> ```XML
>     }
>     const options: any = {};
>     if (source != null) {
> ```

Do we need an else block to unset `submitSource` if the the source is empty, or is the only time we want to unset it when there's an error? This seems like a really obvious thing to consider, but I wanted to point it out, since it's different from the "flip on flip off" logic it replaces.
___
*[`src/SIL.XForge/Realtime/OpMetadata.cs` line 28 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0qnoOO7na7ePwfmEUu:-O0qnoOP14nUuEnAiNCh:b4pxlsu) ([raw file](https://github.com/sillsdev/web-xforge/blob/130044708b4be4597b582dbe904835c5ce5ab17d/src/SIL.XForge/Realtime/OpMetadata.cs#L28)):*
> ```Smalltalk
>     [OnError]
> #pragma warning disable CA1822 // Members that do not access instance data or call instance methods can be marked static
>     internal void OnError(StreamingContext context, ErrorContext errorContext) => errorContext.Handled = true;
> ```

`context` is unused. Actually, the whole method is unused? Is this meant for future consumers? Is it right for it to be internal then?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts` line 47 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0mFeudDwv7n3V6iqh_:-O0qpHG13J947fZaiOnb:b-pp2yf5) ([raw file](https://github.com/sillsdev/web-xforge/blob/436821ff63101ee27378bf10b115801a8a4c43eb/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts#L47)):*

Done. I originally did this to match the behavior in ShareDB documentation for `source` and `submitSource` - I figured if I follow their practice, there it is at least a bit more transparency as to what is going on. See https://share.github.io/sharedb/api/doc#submitop

That said, on reflection, I have made it always submit the source, as I think that will be useful and it fixes the known issue (which I have removed from the PR description). Thanks!
</blockquote></details>

Awesome. Glad to hear my suggestion helped, and I see your reasoning now. I do wonder, though, what their intended use case is for submitting with source while `submitSource` is false. I hinted at this in another comment. Does it just update locally and not send? I like us wrapping that up into our adapter layers. Thanks
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.html` line 12 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0mInqL1teIB8Sx3q9F:-O0quWaP-FIIgM1hn5GC:b3yj6y1) ([raw file](https://github.com/sillsdev/web-xforge/blob/436821ff63101ee27378bf10b115801a8a4c43eb/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor-history/history-chooser/history-chooser.component.html#L12)):*

Done. It looks good. We could use the left hand side for the avatar icon, if an when we do that.

My only concern would be future versions of Angular possibly breaking the CSS?
</blockquote></details>

Yeah, if they wanna restructure things again, we'll have to adapt. Most likely those first couple rules. I think both of them are worth it, though.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2556)* status: all files reviewed, 2 unresolved discussions (waiting on @josephmyers)

___
*[`src/RealtimeServer/common/index.ts` line 252 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0qrRG74J7wSXepFumg:-O0uCM9t3kzzJCSyKC0J:bacmlvu) ([raw file](https://github.com/sillsdev/web-xforge/blob/130044708b4be4597b582dbe904835c5ce5ab17d/src/RealtimeServer/common/index.ts#L252)):*
> Do we need an else block to unset `submitSource` if the the source is empty.

Hmmm it could help to set it to `false`. Thanks!

> or is the only time we want to unset it when there's an error?

This block isn't an error block, it is the callback which is executed on success and error. Only when an error occurs is the err object populated. See https://github.com/share/sharedb/blob/master/lib/client/doc.js#L854

I am setting the value to false whenever the submitOp completes, failure or success.
___
*[`src/SIL.XForge/Realtime/OpMetadata.cs` line 28 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0qnoOO7na7ePwfmEUu:-O0uGI0I3Kni0WEiBxwv:bi4dvph) ([raw file](https://github.com/sillsdev/web-xforge/blob/130044708b4be4597b582dbe904835c5ce5ab17d/src/SIL.XForge/Realtime/OpMetadata.cs#L28)):*

`context` is unused. Actually, the whole method is unused? Is this meant for future consumers? Is it right for it to be internal then?
</blockquote></details>

This method is required for error handling of cases where the string that source is set to is not in the enum.

Both the arguments are required, but I have changed the `StreamingContext` argument to an `_` discard variable, to highlight that it is not used.

See https://www.newtonsoft.com/json/help/html/SerializationErrorHandling.htm#OnErrorAttribute for details.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts` line 47 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0mFeudDwv7n3V6iqh_:-O0uW5Yq2eKd6FqVoZJY:bkrg6nu) ([raw file](https://github.com/sillsdev/web-xforge/blob/436821ff63101ee27378bf10b115801a8a4c43eb/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.ts#L47)):*

Awesome. Glad to hear my suggestion helped, and I see your reasoning now. I do wonder, though, what their intended use case is for submitting with source while `submitSource` is false. I hinted at this in another comment. Does it just update locally and not send? I like us wrapping that up into our adapter layers. Thanks
</blockquote></details>

ShareDB uses a `source` of false to determine if an op comes from the remote server, or `truthy` to determine if the op originates locally. In cases where the `source` is `true` or `false`, we do not want to send that to the server. For `string` values we specify, we do want to submit them to server.

See https://github.com/share/sharedb/blob/dcef5491ca400826389c823a9ae2c53e2b02beac/lib/client/doc.js#L597-L612 for where `source` is used by ShareDB, but is not submitted.

josephmyers (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2556)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/RealtimeServer/common/index.ts` line 252 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0qrRG74J7wSXepFumg:-O0v4SVvARzS2fjv8XTm:b-v1e0sr) ([raw file](https://github.com/sillsdev/web-xforge/blob/130044708b4be4597b582dbe904835c5ce5ab17d/src/RealtimeServer/common/index.ts#L252)):*

> Do we need an else block to unset `submitSource` if the the source is empty.

Hmmm it could help to set it to `false`. Thanks!

> or is the only time we want to unset it when there's an error?

This block isn't an error block, it is the callback which is executed on success and error. Only when an error occurs is the err object populated. See https://github.com/share/sharedb/blob/master/lib/client/doc.js#L854

I am setting the value to false whenever the submitOp completes, failure or success.
</blockquote></details>

Ah, thanks, I wasn't reading it super closely.
___
*[`src/SIL.XForge/Realtime/OpMetadata.cs` line 28 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2556#-O0qnoOO7na7ePwfmEUu:-O0v59a23VAWq5jFs1Ex:b-4hvi0z) ([raw file](https://github.com/sillsdev/web-xforge/blob/130044708b4be4597b582dbe904835c5ce5ab17d/src/SIL.XForge/Realtime/OpMetadata.cs#L28)):*

This method is required for error handling of cases where the string that source is set to is not in the enum.

Both the arguments are required, but I have changed the `StreamingContext` argument to an `_` discard variable, to highlight that it is not used.

See https://www.newtonsoft.com/json/help/html/SerializationErrorHandling.htm#OnErrorAttribute for details.
</blockquote></details>

Got it!

PR #2569, opened 2024-07-04
SF-2830 Optimise the syncing of notes to Paratext

pmachapman:
This PR optimizes the syncing of notes to Paratext by optimizing the code that writes each book's notes to Paratext to only run once per project (well twice if you include the writing of biblical terms notes).

This is possible because previous work done to SF made the use of the book number sent to `ParatextService.UpdateParatextCommentsAsync()` to be redundant.

This change results in a speed increase when syncing projects with a many books and many notes, especially if few notes are created or edited in Scripture Forge. In my test project that has a full canon and note for every verse, a 2.5 minute sync was reduced to 2 minutes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2569)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/IParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/IParatextService.cs
@@ -53,7 +53,6 @@ public interface IParatextService
     Task<SyncMetricInfo> UpdateParatextCommentsAsync(
         UserSecret userSecret,
         string paratextId,
-        int? bookNum,
         IEnumerable<IDocument<NoteThread>> noteThreadDocs,
         IReadOnlyDictionary<string, string> userIdsToUsernames,
         Dictionary<string, ParatextUserProfile> ptProjectUsers,
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -1315,7 +1315,6 @@ public class ParatextService : DisposableBase, IParatextService
     public async Task<SyncMetricInfo> UpdateParatextCommentsAsync(
         UserSecret userSecret,
         string paratextId,
-        int? bookNum,
         IEnumerable<IDocument<NoteThread>> noteThreadDocs,
         IReadOnlyDictionary<string, string> displayNames,
         Dictionary<string, ParatextUserProfile> ptProjectUsers,
--- a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
@@ -554,7 +554,7 @@ public class ParatextSyncRunner : IParatextSyncRunner
                 .ToList();
             if (!_paratextService.IsResource(paratextId))
             {
-                LogMetric("Updating Paratext notes");
+                LogMetric("Updating Paratext notes for questions");
                 if (questionDocsByBook[text.BookNum].Count > 0)
                 {
                     await UpdateParatextNotesAsync(text, questionDocsByBook[text.BookNum]);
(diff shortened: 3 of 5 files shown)


RaymondLuong3 (commented):
Was the speed increase because there were no notes where `PublishedToSF` was true?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2569#-:-O0zwvWO0dY0OETiTy-0:b8xoy9i)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 572 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2569#-O0zwKII6o7K_XmANMtG:-O0zwKII6o7K_XmANMtH:b9o6yrh) ([raw file](https://github.com/sillsdev/web-xforge/blob/95a8d3dea1fe56da8ecf256a932986e5a4d62341/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L572)):*
> ```Smalltalk
>         // Update the notes for all books if this is not a resource
>         LogMetric("Updating Paratext notes");
>         if (!_paratextService.IsResource(paratextId))
> ```

Oh, this is nice to check if it is a resource.



_Code quote:_
```Smalltalk
        if (!_paratextService.IsResource(paratextId))
```
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 575 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2569#-O0zvcZv00xdjWUkicxS:-O0zvcZv00xdjWUkicxT:bsfdfwd) ([raw file](https://github.com/sillsdev/web-xforge/blob/95a8d3dea1fe56da8ecf256a932986e5a4d62341/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L575)):*
> ```Smalltalk
>         {
>             // Only update the note tag if there are SF note threads in the project
>             if (noteDocs.Any(nt => nt.Data.PublishedToSF == true && nt.Data?.BiblicalTermId == null))
> ```

Am I correct in my understanding that the biggest change here is that projects with no editable notes can skip updating Paratext comments, whereas previously comments would be updated if there are editable biblical terms on the project?



_Code quote:_
```Smalltalk
            if (noteDocs.Any(nt => nt.Data.PublishedToSF == true && nt.Data?.BiblicalTermId == null))
```

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2569)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 575 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2569#-O0zvcZv00xdjWUkicxS:-O1DfYT9A_D5QwJvOEiV:br1f9wz) ([raw file](https://github.com/sillsdev/web-xforge/blob/95a8d3dea1fe56da8ecf256a932986e5a4d62341/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L575)):*
> Am I correct in my understanding that the biggest change here is that projects with no editable notes can skip updating Paratext comments,

Partially. This PR both skips updating notes for projects with no editable notes, but also for projects with editable notes, the notes are only updated twice per sync (once for in-book notes, the second time for BT notes)

> whereas previously comments would be updated if there are editable biblical terms on the project?

No, previously if a project had notes that were editable in two or more books, then we would load the notes XML files (an CPU and memory intensive process) for each book with notes, then update the notes just for that book. As the code to updated notes in PsratextService did not actually use the book number, this PR removes that argument, and the code is now executed outside of the loop over each book.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2569)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 575 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2569#-O0zvcZv00xdjWUkicxS:-O1ISQTd-a5mi8ORFONi:b-ctvcz2) ([raw file](https://github.com/sillsdev/web-xforge/blob/95a8d3dea1fe56da8ecf256a932986e5a4d62341/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L575)):*

> Am I correct in my understanding that the biggest change here is that projects with no editable notes can skip updating Paratext comments,

Partially. This PR both skips updating notes for projects with no editable notes, but also for projects with editable notes, the notes are only updated twice per sync (once for in-book notes, the second time for BT notes)

> whereas previously comments would be updated if there are editable biblical terms on the project?

No, previously if a project had notes that were editable in two or more books, then we would load the notes XML files (an CPU and memory intensive process) for each book with notes, then update the notes just for that book. As the code to updated notes in PsratextService did not actually use the book number, this PR removes that argument, and the code is now executed outside of the loop over each book.
</blockquote></details>

Hah, I completely missed the part where you close the for loop over all the books of the project and then you allow the code to update all the note threads at once. Nice!

PR #2573, opened 2024-07-05
SF-2828 Remove the delete user button on self on system admin page

kylebuss:
Prevent logged in admin user from deleting their user in the System Administration page.
Removed word-break css so username displays nicely

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2573)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/system-administration/sa-users.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/system-administration/sa-users.component.html
@@ -35,9 +35,11 @@
       </ng-container>
       <ng-container matColumnDef="remove">
         <td mat-cell *matCellDef="let userRow" class="remove-cell">
+          @if(currentUserId != userRow.id) {
           <button mat-icon-button class="remove-user" (click)="removeUser(userRow.id, userRow.user)">
             <mat-icon>close</mat-icon>
           </button>
+          }
         </td>
       </ng-container>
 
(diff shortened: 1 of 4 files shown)


RaymondLuong3 (changes requested):
You will want to rename the PR to be in the format
<Issue ID> <Description of the behaviour change>. Something like SF-2828 Remove the delete user button on self on system admin page.
That can be edited right on the github PR.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2573#-:-O13uWTMEjXCHAV0YVYH:b-y3p595)* status: all files reviewed, 1 unresolved discussion (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/system-administration/sa-users.component.html` line 38 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2573#-O13wVU89VdjqEPEDnp6:-O13wVU89VdjqEPEDnp7:b-sradnd) ([raw file](https://github.com/sillsdev/web-xforge/blob/ae8956198430aa68ec164cb96d201a01802e7aef/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/system-administration/sa-users.component.html#L38)):*
> ```HTML
>       <ng-container matColumnDef="remove">
>         <td mat-cell *matCellDef="let userRow" class="remove-cell">
>           @if(currentUserId != userRow.id) {
> ```

In this case I do not think there is a problem, but there is a difference between `!=` and `!==`. The second is a strict not equals, and that is what you want. You would use `!=` when you want to check if something is not null or undefined i.e. `object != null`



_Code quote:_
```HTML
          @if(currentUserId != userRow.id) {
```

PR #2575, opened 2024-07-08
SF-2711 Do not display sync complete before ops are written to Mongo

pmachapman:
This PR executes the ops that run outside of the transaction to run after the translation is committed or rolled back.

This ensures that the sync is displayed as complete only when it is complete. It also means that actions that modify the project data are not committed outside of the transaction.

As a part of this change, the backup of the Paratext repository now takes place before the transaction is committed. This is because the contents of the repo have been already synced to the Paratext Send/Receive server. If a failure (hypothetically) occurs during the `CommitTransactionAsync()`, then the next sync will just wind the repo back to the appropriate Mercurial commit id when the backup is restored.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2575)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
@@ -30,7 +30,7 @@ namespace SIL.XForge.Scripture.Services;
 /// Algorithm:
 /// 1. The text deltas from the real-time docs are converted to USX.
 /// 2. The local repo is updated using the converted USX.
-/// 3. A note changelist is computed by diffing the the real-time question docs and the notes in the local repo.
+/// 3. A note changelist is computed by diffing the real-time question docs and the notes in the local repo.
 /// 4. The local repo is updated using the note changelist.
 /// 5. PT send/receive is performed (remote and local repos are synced).
 /// 6. Docs associated with remotely deleted books are deleted.
@@ -1010,6 +1010,7 @@ public class ParatextSyncRunner : IParatextSyncRunner
         _conn.BeginTransaction();
         _conn.ExcludePropertyFromTransaction<SFProject>(op => op.Sync.QueuedCount);
         _conn.ExcludePropertyFromTransaction<SFProject>(op => op.Sync.DataInSync);
+        _conn.ExcludePropertyFromTransaction<SFProject>(op => op.Sync.LastSyncSuccessful);
         _projectDoc = await _conn.FetchAsync<SFProject>(projectSFId);
         if (!_projectDoc.IsLoaded)
         {
@@ -1700,11 +1701,8 @@ public class ParatextSyncRunner : IParatextSyncRunner
             }
         }
 
-        // NOTE: This is executed outside of the transaction because it modifies "Sync.QueuedCount"
         await _projectDoc.SubmitJson0OpAsync(op =>
         {
-            op.Set(pd => pd.Sync.LastSyncSuccessful, successful);
-
             // Get the latest shared revision of the local hg repo. On a failed synchronize attempt, the data
             // is known to be out of sync if the revision does not match the corresponding revision stored
             // on the project doc.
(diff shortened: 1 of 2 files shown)


marksvc (changes requested):
If I make the code throw an exception before `BackupRepository()`, I can confirm that it is not updating the SF DB, such as field `dateLastSuccessfulSync`.

However, I observe with interest that when the sync thus fails (and re-enters `CompleteSync()` , now with `successful: false`, which was also interesting to notice and consider the behaviour of), a message appears to the user at the bottom of the web page saying the sync was successful! I realize that `LastSyncSuccessful` is not getting updated if the sync failed. 

This is remedied with a change like the following. With it, the user is successfully told that the sync failed. 
```diff
@@ -1003,6 +1003,7 @@ public class ParatextSyncRunner : IParatextSyncRunner
         _conn.BeginTransaction();
         _conn.ExcludePropertyFromTransaction<SFProject>(op => op.Sync.QueuedCount);
         _conn.ExcludePropertyFromTransaction<SFProject>(op => op.Sync.DataInSync);
+        _conn.ExcludePropertyFromTransaction<SFProject>(op => op.Sync.LastSyncSuccessful);
         _projectDoc = await _conn.FetchAsync<SFProject>(projectSFId);
         if (!_projectDoc.IsLoaded)
         {
@@ -1695,8 +1696,6 @@ public class ParatextSyncRunner : IParatextSyncRunner
 
         await _projectDoc.SubmitJson0OpAsync(op =>
         {
-            op.Set(pd => pd.Sync.LastSyncSuccessful, successful);
-
             // Get the latest shared revision of the local hg repo. On a failed synchronize attempt, the data
             // is known to be out of sync if the revision does not match the corresponding revision stored
             // on the project doc.
@@ -1874,6 +1873,7 @@ public class ParatextSyncRunner : IParatextSyncRunner
             // Backup the repository
             if (!_paratextService.IsResource(_projectDoc.Data.ParatextId))
             {
+                throw new Exception("Simulate problem");
                 bool backupOutcome = _paratextService.BackupRepository(_userSecret, _projectDoc.Data.ParatextId);
                 if (_syncMetrics != null)
                 {
@@ -1899,6 +1899,7 @@ public class ParatextSyncRunner : IParatextSyncRunner
         // and dataInSync must record the real value if a transaction fails.
         await _projectDoc.SubmitJson0OpAsync(op =>
         {
+            op.Set(pd => pd.Sync.LastSyncSuccessful, successful);
             op.Set(pd => pd.Sync.DataInSync, dataInSync);
 
             // The frontend checks the queued count to determine if the sync is complete. The ShareDB client emits
```

(As long as nothing will go wrong between `CommitTransactionAsync()` and setting `LastSyncSuccessful`.)

So I think we should include something like that, if I'm thinking about this the right way.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2575#-:-O1IWrI78_O_kUGK0wJW:b-nsz2xn)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 1707 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2575#-O1IMscx0ZNAkoI50WLi:-O1IMscx0ZNAkoI50WLj:bn2ncbm) ([raw file](https://github.com/sillsdev/web-xforge/blob/65ea2320aa928878467929525cfd719aaaf328c4/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L1707)):*
> ```Smalltalk
>             if (successful)
>             {
>                 Log($"CompleteSync: Successfully synchronized to PT repo commit id '{repoVersion}'.");
> ```

I see that we will print this message even if later we don't commit the transaction. :-) :-/
Something to at least keep in mind when looking at the logs.
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 1902 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2575#-O1IORSAEozmqKisJ0Aa:-O1IORSAEozmqKisJ0Ab:bvk24q8) ([raw file](https://github.com/sillsdev/web-xforge/blob/65ea2320aa928878467929525cfd719aaaf328c4/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L1902)):*
> ```Smalltalk
>         await _projectDoc.SubmitJson0OpAsync(op =>
>         {
>             op.Set(pd => pd.Sync.DataInSync, dataInSync);
> ```

Hmm. Because at the beginning of a sync we `SetRepoToRevision()`, I wonder if we don't need to record if `DataInSync` anymore.  I started examining how we are using DataInSync and got a bit bogged down :). I see the Sync.cs and sync.ts models have some commentary. Maybe in the future.

marksvc:
Hmm, I didn't look to see if this was a problem before. Sorry if I'm asking you to take care of something that wasn't a problem you were introducing :)

pmachapman (commented):
> Hmm, I didn't look to see if this was a problem before. Sorry if I'm asking you to take care of something that wasn't a problem you were introducing :)

No, you are spot on. When a `SubmitJson0OpAsync` contains just one property that is to be excluded from the transaction, then all of the property updates are submitted. Documenting it via `_conn.ExcludePropertyFromTransaction<SFProject>(op => op.Sync.LastSyncSuccessful);` is not necessary, but it is helpful for clarity, so I will add it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2575#-:-O1JQA276kzS_kOYixze:b-4gzbsz)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @marksvc)

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2575)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

marksvc (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2575)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2584, opened 2024-07-11
SF-2831 Null check for remote changes

josephmyers:
The problem here was that the source GUI property hasn't been updated when the remote change arrives. So the hasSource check is true when the source is undefined. The block that was failing accounts for when the source changes from one source to another, but not for when the source doesn't exist. Ideally, we would have the source tab open after the source is set, which would prevent this, but I'm sure there's a reason we haven't done this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2584)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts
@@ -3956,6 +3956,18 @@ describe('EditorComponent', () => {
           env.dispose();
         });
       }));
+
+      it('should not throw exception on remote change when source is undefined', fakeAsync(() => {
+        const env = new TestEnvironment();
+        env.setProjectUserConfig();
+        env.wait();
+
+        env.component.source = undefined;
+
+        expect(() => env.updateFontSize('project01', 24)).not.toThrow();
+
+        env.dispose();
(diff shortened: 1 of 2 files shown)


RaymondLuong3 (approved):
Nice work. Let's get this to testing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2584#-:-O1YYcJ43zAP6P-4myoT:bai7qlh)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 727 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2584#-O1YLiE6723vLN5pEpsX:-O1YLiE6723vLN5pEpsY:bpx3exr) ([raw file](https://github.com/sillsdev/web-xforge/blob/8a9c941e3520f54734ee658ffb60ee1fa07be6d6/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L727)):*
> ```XML
>                 this.text.bookNum,
>                 this._chapter
>               );
> ```

Thanks for the fix. I can see from testing that if I do select a new source, the window appears after the sync finishes. That does seem like the ideal behaviour.

PR #2589, opened 2024-07-12
SF-2538 Update page when user role changes

RaymondLuong3:
When the user's role changes, it is possible that the page they are currently on is no longer accessible to the role that they have been given. If a Commenter's role is updated to Community Checker, they no longer have permission on the edit and review page.

This PR adds a subscription to listen to the project doc and update the current user's page if their role's permission to view a page has been revoked.

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -255,6 +255,19 @@ describe('AppComponent', () => {
     expect(env.location.path()).toEqual('/projects');
   }));
 
+  it('response to project role changed', fakeAsync(() => {
+    const env = new TestEnvironment();
+    env.navigate(['/projects', 'project01']);
+    env.init();
+    env.setCurrentUser('user04');
+
+    expect(env.selectedProjectId).toEqual('project01');
+    when(mockedLocationService.pathname).thenReturn('/projects/project01/translate');
+    env.changeUserRole('project01', 'user04', SFProjectRole.CommunityChecker);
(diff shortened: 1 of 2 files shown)


josephmyers (changes requested):
I really like that your changes are centralized and account for both Translate and Checking. My only comment is that the swap from Checking to Translate is working on master, so I'm thinking we may have some duplicate business logic in here somewhere. I remember seeing something related a long time ago but don't remember exactly where. Potentially `checking.utils.ts`?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2589#-:-O1pQ_dh2cX7eaS1rg1t:bmuxrh5)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
Yes, you are right. There was existing logic so that conversion from community checker to viewer or commenter worked. I have consolidated the logic so that the switching of roles is only handled in the app component. Thanks for pointing that out.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2589#-:-O21kseaB85Sb8wbhKri:baifoym)* status: 0 of 8 files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (approved):
Wow! I didn't know that would be so involved. Thanks for cleaning it up.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2589#-:-O236Bmo7pXWhW8aD13S:b-cpjt8m)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2586, opened 2024-07-11
SF-2847 Do not show source tab group if user does not have source access

RaymondLuong3:
Commenters and viewer users will not generally have access to view the source project. An error would be thrown at the moment when the source tab group is being instantiated because the user does not have permission to read the project data of the source. This PR adds a check if the user has source project access before trying to initialize the source tab group.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2586)

PR #2593, opened 2024-07-15
SF-2851 Update source id when tab groups consolidate

RaymondLuong3:
After doing a git bisect, I found that the commit that introduced the initial error when Commenters navigated to a project was on SF-2808. The id of the source text container was set in the html, causing the text component to try to access the source project, even when users did not have permission.
The change reverts some of the changes in PR #2586 that failed to fix the issue, and introduces a fix to only set the source id if that user has permission to see the source text.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2593)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
@@ -89,7 +89,7 @@
                 <div class="text-container">
                   <app-text
                     #source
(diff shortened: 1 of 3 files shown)


siltomato (approved):
Thanks for fixing this, Raymond. :lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2593#-:-O1s48jD8PvnvuoFYPVd:b-c0wyp0)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2595, opened 2024-07-17
SF-2779 Set 'Select all' to select text for current segment

siltomato:
Previously, selecting all text via "Ctrl+A" from a segment selected the content from the longest segment.  This is due to `TextViewModel.getSegmentRef()` choosing the segment with the biggest overlap with the selection range.

This PR changes `getSegmentRef()` to take a "preferred ref" argument that will be the current segment ref so that if the current segment is completely within the selection range, it will be returned regardless of whether it is the longest segment.  As a result, "Ctrl + A" (whose range encompasses all segments) will select the text from the current segment.


RaymondLuong3 (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2595#-:-O26kPkaEk_8pJmIliKC:bg7igkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

PR #2602, opened 2024-07-22
SF-2870 Improve layout on my projects page

Nateowami:
- Fixed layout of cards
- Reduced font weight of titles
- Reduced spacing of titles
- Fixed cards starting out narrow and then expanding wider roughly 1 second after page load
- Fixed long words in names not wrapping

To see the biggest difference in behavior, check out the "Layout lengths" story on a small screen.

### After

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2602)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html
@@ -1,144 +1,123 @@
 <ng-container *transloco="let t; read: 'my_projects'">
-  <div class="wrapper">
-    <h1>{{ t("my_projects") }}</h1>
-    <app-notice
-      id="message-no-pt-or-sf-projects"
-      icon="warning"
-      type="warning"
-      *ngIf="!userHasProjects && !initialLoadingSFProjects && !loadingPTProjects"
+  <h1>{{ t("my_projects") }}</h1>
+  <app-notice
+    id="message-no-pt-or-sf-projects"
+    icon="warning"
+    type="warning"
+    *ngIf="!userHasProjects && !initialLoadingSFProjects && !loadingPTProjects"
+  >
+    {{ t("no_projects") }} {{ t("access_another_project") }}
+  </app-notice>
+  <h2 id="header-connected-projects" *ngIf="userConnectedProjects.length > 0 || initialLoadingSFProjects">
+    {{ t("connected") }}
+  </h2>
+  <mat-card
+    *ngFor="let projectDoc of userConnectedProjects"
+    [class.active-project]="isLastSelectedProject(projectDoc)"
+    id="user-connected-project-card-{{ projectDoc.data?.paratextId }}"
+    attr.data-pt-project-id="{{ projectDoc.data?.paratextId }}"
+    data-project-type="user-connected-project"
+  >
+    <span class="project-name">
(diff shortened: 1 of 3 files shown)


marksvc (changes requested):
The display is looking good. I noticed that although long titles are displayed over the "Open" button in the "Connected" and "Digital Bible Library resources" lists, in the "Not connected" list, names are still shown to the left of the "Connect" or "Join" button (as viewable in the "Layout Lengths" story when you make the browser window narrow.

Such as seen in this screenshot:



The "Digital Bible Library resources" and "Not connected" headings look a bit close to the content above them. It looks like it should have more space above them. Such as seen in the "Layout Lengths" story. Tho be careful about its spacing as seen in story "SF Translator", as it seems to have a bit more space above it there.

Such as seen in this screenshot:



The SF and PT loading cards look good, in that they have the button in the lower-right, as seen in story "SF Loading".

> Fixed cards starting out narrow and then expanding wider roughly 1 second after page load

Good catch; I don't recall spotting that.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2602#-:-O2WBqAb3g_xhBVweFRu:b-x98d4q)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

marksvc:
@Nateowami Thank you. Here are some things to address.

Nateowami (commented):
Thanks. I've made some adjustments to the margin for the headers. What do you think?

Keeping the button on the same line as the title for non-connected projects was intentional. For connected projects, there are three main elements: title, description, and button. Since the title tends to be the longest and varies in length a lot, I thought it belonged on its own line, rather than competing with the button for space. The button then goes on the same line as the description.

However, for non-connected projects, there are only two elements, and so moving it to the bottom alters the layout a lot more, and in my opinion looks weird and uses too much space. Of course, the consequence is that on small viewports with long titles, it doesn't look great. I've changed it to use flexbox to make it dynamic. When the title and button can't fit on the same line it now wraps to the second line.

I've also added two stories to show the mobile layout, and to show how it looks when in RTL mode.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2602#-:-O2ZnU_O6iKBeKoaer44:boug1hk)* status: 1 of 3 files reviewed, all discussions resolved (waiting on @marksvc)

Nateowami:
I just made a change to make the project short names slightly bold. I think this makes it easier to skim the page.

marksvc (approved):
It is good. Thanks for adding RTL and mobile.
Now that short codes are bolded, the headers don't stand out very well, tho :)   Hmm, it looks like that's mostly when the window size is such that the window borders are next to the sides of the cards. Such as 800 px viewport width. Not terrible, but perhaps room for improvement.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2602#-:-O2akzle1ALLR9qyGbYg:bte9b7l)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami:
Agree it's not ideal. Hopefully we'll be able to continue to improve the page.

Are you planning to also review the Chromatic UI tests?

marksvc:
@Nateowami  Yes! Thank you. I think I may have done that for the first CSS changes, but not yet for the followup! Done.

PR #2610, opened 2024-07-23
SF-2741 Delete project on server before deleting serval project

kylebuss:
Relocated code for deleting directories and RealtimeServer data to be deleted before deleting serval data and project secrets on Project Deletion.
Added check for existing directory when Creating SF Project (Connecting) and error if directory exists.
Added Test case for directory check on creating project.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2610)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/SFProjectService.cs
+++ b/src/SIL.XForge.Scripture/Services/SFProjectService.cs
@@ -84,11 +84,16 @@ public class SFProjectService : ProjectService<SFProject, SFProjectSecret>, ISFP
         if (!userSecretAttempt.TryResult(out UserSecret userSecret))
             throw new DataNotFoundException("The user does not exist.");
 
+        string projectDir = Path.Combine(SiteOptions.Value.SiteDir, "sync", settings.ParatextId);
+        if (FileSystemService.DirectoryExists(projectDir))
+            throw new ApplicationException("The directory already exists.");
+
         IReadOnlyList<ParatextProject> ptProjects = await _paratextService.GetProjectsAsync(userSecret);
 
         ParatextProject ptProject =
             ptProjects.SingleOrDefault(p => p.ParatextId == settings.ParatextId)
             ?? throw new DataNotFoundException("The paratext project does not exist.");
+
         var project = new SFProject
         {
             IsRightToLeft = ptProject.IsRightToLeft,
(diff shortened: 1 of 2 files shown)


RaymondLuong3 (changes requested):
This looks good. You commit message should focus on the impact to users. In this case, something like "Delete project on server before deleting serval project". When we look at shipping and see the list of changes, we want to be able to tell what change happened so that we can include it in our release notes. Just one minor change to make, and then we can send it to testers.

On the Jira task, add a comment with acceptance tests. You can look at other issues for examples. In this case, since it is not a bug that you are fixing, you just want to explain the most common use case that should be working which includes reconnecting to the project a second time after deleting it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2610#-:-O2_jh9A1dh57HYWYggA:b-1oyxkn)* status: all files reviewed, 1 unresolved discussion (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 254 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2610#-O2_cFPfAprjPmaaYqX7:-O2_cFPgCCmHzNTt-xEl:bgo0hzx) ([raw file](https://github.com/sillsdev/web-xforge/blob/f031a2f1339371135048c199c51d7c9839503969/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L254)):*
> ```Smalltalk
> 
>             ptProjectId = projectDoc.Data.ParatextId;
>             // delete the project first, so that users get notified about the deletion
> ```

This comment can be removed.



_Code quote:_
```Smalltalk
            // delete the project first, so that users get notified about the deletion
```

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2610)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kylebuss)

src/SIL.XForge.Scripture/Services/SFProjectService.cs
```diff
@@ -84,11 +84,16 @@ public async Task<string> CreateProjectAsync(string curUserId, SFProjectCreateSe
         if (!userSecretAttempt.TryResult(out UserSecret userSecret))
             throw new DataNotFoundException("The user does not exist.");
 
+        string projectDir = Path.Combine(SiteOptions.Value.SiteDir, "sync", settings.ParatextId);
+        if (FileSystemService.DirectoryExists(projectDir))
+            throw new InvalidOperationException("A directory for this project already exists.");
```
kylebuss, line 89: @pmachapman Is using an InvalidOperationException here okay or do you believe we should use a different exception? Currently the InvalidOperationException is caught in the SFProjectsRpcController.cs to present a clear message to the front end.

pmachapman: Are you wanting the error to be displayed?

I don't think it is too out of place - contextually it makes sense compared to the other API endpoints to throw an `InvalidOperationException` for this type of problem.

You can always use a different exception code, convert it to a custom CommandErrorCode, then handle appropriately on the frontend, if you want to do something different.

RaymondLuong3 (approved):
I will approve this. Leave a comment in the Jira card to let them know to re-test this issue.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2610#-:-O2fj7tJ71r0p4ItMwpm:b-th1r1t)* status: all files reviewed, 1 unresolved discussion (waiting on @kylebuss)

src/SIL.XForge.Scripture/Services/SFProjectService.cs
kylebuss: Yes. We ran into an issue when this was being tested and displaying the error message will make it easier to verify the issue and help resolve it.

PR #2612, opened 2024-07-24
SF-2873 Show "Draft in progress" above "Preview last draft"

Nateowami:
The primary thing on the page is now first. Personally I think the "Draft in progress" section should be in a card too (or we should find something better than a card), but I'm trying to keep this change simple.

Before | After
-------|------
 | 

I felt the bold here was excessive and made it more subtle:

Before | After
-------|-------
 | 

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2612)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.html
@@ -5,7 +5,10 @@
   </h1>
 
   <section>
-    <div *ngIf="isDraftJobFetched && !isDraftInProgress(draftJob) && !hasAnyCompletedBuild">
+    <div
+      *ngIf="isDraftJobFetched && !isDraftInProgress(draftJob) && !hasAnyCompletedBuild"
+      class="drafting-instructions"
+    >
       <div *ngIf="isBackTranslationMode">
         <p>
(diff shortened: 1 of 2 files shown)


RaymondLuong3 (approved):
I like it. Nice simple redesign.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2612#-:-O2f2yV5AXQpN14jQOZX:b-zgrmag)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.scss` line 170 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2612#-O2f2eQZ-qqsMZcmI1NG:-O2f2eQZ-qqsMZcmI1NH:b-4mm50u) ([raw file](https://github.com/sillsdev/web-xforge/blob/fbfb70fd903c98d93ff775365e79d314f59cbd18/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation.component.scss#L170)):*
> ```SCSS
> .drafting-instructions b {
>   font-weight: 500;
> }
> ```

The contrast between the bold text and the normal text is not as noticeable, but I don't think it will be a big issue.



_Code quote:_
```SCSS
.drafting-instructions b {
  font-weight: 500;
}
```

PR #2613, opened 2024-07-24
SF-2859 Inform user when the share link problem is known

RaymondLuong3:
Error messages that are returned from the server are typically CommandErrors that have a message in the form "Error invoking ... project_link_invalid". We can use the error message to determine if a known error code is included in the message, and then show a message to the user to inform them of the known issue.
This was not previously working because we were assuming the message was in the format that was just the error code, but in reality we should have been checking whether the error code was contained in the error message.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2613)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.spec.ts
@@ -6,7 +6,7 @@ import { NoopAnimationsModule } from '@angular/platform-browser/animations';
 import { ActivatedRoute, Router } from '@angular/router';
 import { SFProjectRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-role';
 import { of } from 'rxjs';
-import { anything, capture, mock, verify, when } from 'ts-mockito';
+import { anything, capture, mock, resetCalls, verify, when } from 'ts-mockito';
 import { AnonymousService } from 'xforge-common/anonymous.service';
 import { AuthService } from 'xforge-common/auth.service';
 import { CommandError, CommandErrorCode } from 'xforge-common/command.service';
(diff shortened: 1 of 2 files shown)


pmachapman (changes requested):
Looks good - just a minor point of feedback.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2613#-:-O2uyMYXCpUAh2bkbQYo:b-j9dgev)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts` line 203 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2613#-O2uxC69BXqBpZIcNkxK:-O2uxC6AB6W0aPLxcGdv:bh7ic1b) ([raw file](https://github.com/sillsdev/web-xforge/blob/b8ad888b166e826a736539bdc5609c01af53c51a/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts#L203)):*
> ```XML
>     if (typeof code === 'string') {
>       for (const knownCode of KNOWN_ERROR_CODES) {
>         if (code.match(knownCode)) return knownCode;
> ```

Since `knownCode` is not a regular expression, and `match()` returns an array or null, should we use `includes()` instead? i.e.
```ts
if (code.includes(knownCode)) return knownCode;
```



_Code quote:_
```XML
if (code.match(knownCode)) return knownCode;
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2613)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts` line 203 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2613#-O2uxC69BXqBpZIcNkxK:-O3-Gh5C0FvWcqIUkqIF:b-x384q3) ([raw file](https://github.com/sillsdev/web-xforge/blob/b8ad888b166e826a736539bdc5609c01af53c51a/src/SIL.XForge.Scripture/ClientApp/src/app/join/join.component.ts#L203)):*

Since `knownCode` is not a regular expression, and `match()` returns an array or null, should we use `includes()` instead? i.e.
```ts
if (code.includes(knownCode)) return knownCode;
```
</blockquote></details>

Wow, I don't know how I didn't see that. Thanks!

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2613#-:-O3-OXDn3CUP1ULaqoor:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2615, opened 2024-07-25
Add query to measure sync success rate over time

Nateowami:
For the results of this query, see the doc "Sync failure stats" in the "Metrics" directory of the team drive. I suggest understanding the output before trying to understand the logic of this script, because it dictated an unusual approach for processing the sync metrics into the output.

The goal of this query is to measure not how many syncs are succeeding or failing, but how many projects are able to successfully sync. Each project then has a state it's in, which is the result of its most recent sync.

However, I wanted to be able to look at only active projects, so I added the ability to only look at projects within a certain time window, and if they don't try to sync for that time window, they stop being counted.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2615)

### DIFF
--- a/mongodb/SyncMetrics/SyncStatusOverTime.mongodb
+++ b/mongodb/SyncMetrics/SyncStatusOverTime.mongodb
@@ -1,0 +1,73 @@
+use('xforge')
+
+// A project is considered active for 30 days from when it attempts to sync
+const activePeriod = Infinity;
+
+const firstSyncMetricStartTime = db.sync_metrics.find({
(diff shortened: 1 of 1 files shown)


pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2615#-:-O2v84_-1woIZPddn7nO:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2622, opened 2024-07-29
SF-2882 Make project names clickable

Nateowami:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2622)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html
@@ -18,9 +18,9 @@
     attr.data-pt-project-id="{{ projectDoc.data?.paratextId }}"
     data-project-type="user-connected-project"
   >
-    <span class="project-name">
+    <a class="project-name" appRouterLink="/projects/{{ projectDoc.id }}">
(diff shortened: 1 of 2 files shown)


josephmyers (changes requested):
A concern with this is that it's not clear that the labels are clickable (especially mobile). We should design the clickable space to indicate that it is. A somewhat bigger concern is that it's now inconsistent with the unconnected projects. The fact that you can click some and not others is strange to me, though I'm sure it was an intentional decision.

@nigel-wells, do you have thoughts on either of these issues? We should discuss this at the UX meeting.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2622#-:-O2wPrguB-hSGf4GkABX:bbv61pd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

nigel-wells:
Yeah, let's discuss. The original thinking was around using the buttons as the primary way to interact with the projects so that it was clear across any card as to what you can do. I do get that you may want to click on a title. We could add (I've not compiled this branch yet) common  link indicators i.e. underline, different color, you can even add a small icon at the end to help indicate it is clickable. Can discuss more tomorrow.

PR #2626, opened 2024-07-29
SF-2884 Fix sync problem when a note has no content

RaymondLuong3:
A project has been having issues synchronizing with Paratext. It is still not clear how the user project got into the state where a note had no content and the owner of the note was a non-paratext users. The result is that the project cannot synchronize because of the expectation that notes with no content can only belong to Paratext users.
UPDATE: It turns out the user IS a paratext user, but their PT User Profile on the project is missing their user Id. This is normally saved the first time the project is created.

This change makes it possible for notes to have no content and not cause a crash. It is not necessary to restrict notes with no content to only belong to Paratext users.

I don't know if we should merge this Monday or Tuesday, and then releasing to live on Tuesday would overwrite this hotfix. Perhaps it is better to wait to Tuesday after releasing to live, then we can hotfix that commit.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2626)

### DIFF
--- a/src/SIL.XForge.Scripture/Models/Note.cs
+++ b/src/SIL.XForge.Scripture/Models/Note.cs
@@ -27,7 +27,7 @@ public class Note : Comment
     /// Content of note. Contains XML. Corresponds to `Contents` element, which is not always present in the
     /// Notes XML file.
     /// </summary>
-    public string Content { get; set; }
+    public string? Content { get; set; }
 
     /// <summary>
     /// Change difference accepted from a conflict, if any. Contains encoded XML. Not always present, in
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -2869,7 +2869,8 @@ public class ParatextService : DisposableBase, IParatextService
         out string xml
     )
     {
-        string equivalentCommentContent = GetCommentContentsFromNote(note, displayNames, ptProjectUsers);
+        string equivalentCommentContent =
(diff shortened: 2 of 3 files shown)


Nateowami:
@RaymondLuong3 Instead of targeting live, I would much prefer to target master, and then cherry-pick it as a hotfix to live. Given our shipping procedure has just changed, merging this won't even deploy it.

RaymondLuong3 (commented):
@Nateowami I can target master for this PR, and then cherry-pick it to live, but it will delay when we can add this hotfix and get the Gayo team unstuck. Or are you suggesting that we merge this into master without testing and hotfix live? I think this is a reasonable change to merge without testing since it only introduces a null check.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2626#-:-O33Hc7O6FdisUWZNOOC:b-pj9xez)* status: 0 of 3 files reviewed, all discussions resolved

Nateowami:
> but it will delay when we can add this hotfix and get the Gayo team unstuck. Or are you suggesting that we merge this into master without testing and hotfix live?

Yes. I don't like having things on a hotfix that aren't on master as well, because then we end up reverting a hotfix by shipping master.

pmachapman (approved):
:lgtm: I was able to recreate this locally on `master` (via hacking the SF code to write `undefined` as the note content), and can confirm that this PR fixes this issue.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2626#-:-O34bZ74CNgEL91i8KxP:b693q8h)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2629, opened 2024-07-30
SF-2890 Add German localization

Nateowami:
I have marked this as "Testing not required" because the locale is hidden using `"production": false` in `locales.json`. Eventually it will need to be tested, but we don't need to test it before getting it into the repo.

In my testing, the biggest issue I found was this on the login page:

I think there will probably be other Auth0 changes that are needed as well to make everything work.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2629)


RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2629)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/locales.json` line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2629#-O33etuH4oM_EhCSoWWa:-O33etuH4oM_EhCSoWWb:brux8z1) ([raw file](https://github.com/sillsdev/web-xforge/blob/032f98c6db2f9ceb14892c26cd0ab6c9bb838396/src/SIL.XForge.Scripture/locales.json#L92)):*
> ```JSON
>     "production": false,
>     "helps": "en"
>   }
> ```

I think German should eventually be moved up in the list when it is ready for production since we want languages to be in order of popularity.



_Code quote:_
```JSON
  {
    "localName": "Deutsch",
    "englishName": "German",
    "tags": ["de", "de-DE"],
    "production": false,
    "helps": "en"
  }
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2629)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/locales.json` line 92 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2629#-O33etuH4oM_EhCSoWWa:-O33fohw-FvhVmB5kFDn:b-cq69nj) ([raw file](https://github.com/sillsdev/web-xforge/blob/76594c8219bb4118ddfac1c683914a2f125a9352/src/SIL.XForge.Scripture/locales.json#L92)):*

I think German should eventually be moved up in the list when it is ready for production since we want languages to be in order of popularity.
</blockquote></details>

Agree. I've tried to sort by popularity (number of speakers), but it's probably easier to just use number of users as the sort order (and probably more useful).

PR #2441, opened 2024-05-01
(WIP) SF-2508b Use Connect Project without target project selector

marksvc:
The Connect Project component doesn't need to show a list of Paratext projects to connect to, now that at the My projects component a user selects a project to connect to. This PR is some work in progress to remove the Paratext target project selector from the Connect Project component. 

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2441)

PR #2635, opened 2024-07-30
SF-2516 Remove project selector from connect project page

RaymondLuong3:
This change removes the project selector from the connect project page since it is no longer expected that users will be updating which project they intend to connect to.
Connect project connect
Connect project join

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2635)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.html
@@ -11,16 +11,14 @@
     </div>
     <mat-card *ngSwitchCase="'connecting'" class="progress-container card-outline">
       <mat-card-content>
-        <div class="progress-label">{{ t("connecting_to_project", { projectName: connectProjectName }) }}</div>
+        <div class="progress-label">{{ t("connecting_to_project", { projectName: projectTitle }) }}</div>
         <app-sync-progress [projectDoc]="projectDoc" (inProgress)="updateStatus($event)"></app-sync-progress>
       </mat-card-content>
     </mat-card>
-    <ng-container *ngIf="projects.length > 0">
-      <div class="project-title">
-        <img src="/assets/images/logo-pt9.png" alt="Paratext Logo" class="paratext-logo" />
-        {{ projectTitle }}
-      </div>
-    </ng-container>
+    <div class="project-title" *ngIf="state !== 'connecting'">
+      <img src="/assets/images/logo-pt9.png" alt="Paratext Logo" class="paratext-logo" />
+      {{ projectTitle }}
+    </div>
     <form *ngSwitchDefault [formGroup]="connectProjectForm" (ngSubmit)="submit()">
       <mat-card id="settings-card" *ngIf="showSettings" class="card-outline" formGroupName="settings">
         <mat-card-title>{{ t("project_settings") }}</mat-card-title>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.html
@@ -1,106 +1,68 @@
 <ng-container *transloco="let t; read: 'connect_project'">
   <div class="title mat-headline-4">{{ t("connect_paratext_project") }}</div>
-  @if (!isAppOnline) {
-    <span class="offline-text"> {{ t("connect_network_to_connect_project") }} </span>
-  }
-  <div class="content">
-    @switch (state) {
-      @case ("login") {
-        <div class="paratext-login-container">
-          <div>{{ t("you_must_be_logged_in_to_paratext") }}</div>
-          <button mat-flat-button id="paratext-login-button" type="button" (click)="logInWithParatext()">
-            <mat-icon><img src="/assets/images/logo-pt9.png" alt="Paratext Logo" class="paratext-logo" /></mat-icon>
-            {{ t("log_in_with_paratext") }}
-          </button>
-        </div>
-      }
-      @case ("connecting") {
-        <mat-card class="progress-container card-outline">
-          <mat-card-content>
-            <div class="progress-label">{{ t("connecting_to_project", { projectName: connectProjectName }) }}</div>
-            <app-sync-progress [projectDoc]="projectDoc" (inProgress)="updateStatus($event)"></app-sync-progress>
-          </mat-card-content>
-        </mat-card>
-      }
-      @default {
-        <form [formGroup]="connectProjectForm" (ngSubmit)="submit()">
-          @if (projects.length > 0 || state === "offline") {
-            @if (hasNonAdministratorProject && !showSettings && isAppOnline) {
-              <p id="connect-non-admin-msg">
-                {{ t("only_paratext_admins_can_start") }}
-              </p>
-            }
-            <mat-form-field appearance="outline">
-              <mat-label>{{ t("paratext_project") }}</mat-label>
-              <mat-select #projectSelect id="project-select" disableOptionCentering formControlName="paratextId">
-                @for (project of projects; track project) {
-                  <mat-option [value]="project.paratextId" [disabled]="!project.isConnectable">
-                    {{ projectLabel(project) }}
-                    @if (!project.isConnectable && project.isConnected) {
-                      &nbsp;{{ t("already_connected") }}
-                    }
-                    @if (!project.isConnectable && !project.isConnected) {
-                      &nbsp;**
-                    }
-                  </mat-option>
-                }
-              </mat-select>
-            </mat-form-field>
-          } @else {
-            @if (state !== "loading") {
-              <p
-                id="no-projects-msg"
-                [innerHtml]="i18n.translateAndInsertTags('connect_project.no_connectable_projects')"
-              ></p>
-            }
-          }
-          @if (showSettings) {
-            <mat-card id="settings-card" class="card-outline" formGroupName="settings">
-              <mat-card-title>{{ t("project_settings") }}</mat-card-title>
-              <mat-card-subtitle>{{ t("you_can_change_later") }}</mat-card-subtitle>
-              <mat-card-content class="card-content">
-                <p>{{ t("select_project_or_resource") }}</p>
-                <p>{{ t("translation_suggestions_require_source_text") }}</p>
-                <app-project-select
-                  formControlName="sourceParatextId"
-                  placeholder="{{ t('source_text_placeholder') }}"
-                  [projects]="projects"
-                  [resources]="resources"
-                  [hiddenParatextIds]="[paratextIdControl.value]"
-                ></app-project-select>
-                @if (showResourcesLoadingFailedMessage) {
-                  <mat-error>{{ t("error_fetching_resources") }}</mat-error>
-                }
-                @if (isBasedOnProjectSet) {
-                  <div fxLayout="row" fxLayoutAlign="start center">
-                    <div fxLayout="column" fxFlex="grow">
-                      <mat-checkbox formControlName="translationSuggestions" id="translation-suggestions-checkbox">
-                        {{ t("enable_translation_suggestions") }}
-                      </mat-checkbox>
-                      <p
-                        class="helper-text"
-                        [innerHTML]="i18n.translateAndInsertTags('settings.translations_will_be_suggested')"
-                      ></p>
-                    </div>
-                  </div>
-                }
-              </mat-card-content>
-              <mat-divider></mat-divider>
-              <mat-card-content class="card-content">
-                <div>
-                  <mat-checkbox formControlName="checking" id="checking-checkbox">
-                    {{ t("enable_community_checking") }}
-                  </mat-checkbox>
-                </div>
-                <p class="helper-text">{{ t("engage_the_wider_community_to_check_scripture") }}</p>
-              </mat-card-content>
-            </mat-card>
-          }
-          <button mat-flat-button type="submit" id="connect-submit-button" [disabled]="submitDisabled">
-            {{ t("connect") }}
-          </button>
-        </form>
-      }
-    }
+  <span *ngIf="!isAppOnline" class="offline-text"> {{ t("connect_network_to_connect_project") }} </span>
+  <div class="content" [ngSwitch]="state">
+    <div *ngSwitchCase="'login'" class="paratext-login-container">
+      <div>{{ t("you_must_be_logged_in_to_paratext") }}</div>
+      <button mat-flat-button id="paratext-login-button" type="button" (click)="logInWithParatext()">
+        <mat-icon><img src="/assets/images/logo-pt9.png" alt="Paratext Logo" class="paratext-logo" /></mat-icon>
+        {{ t("log_in_with_paratext") }}
+      </button>
+    </div>
+    <mat-card *ngSwitchCase="'connecting'" class="progress-container card-outline">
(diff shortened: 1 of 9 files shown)


Nateowami:
Why does it still take some time for the page to load? Instead of passing a Paratext project id, can't you pass all the information needed about the project to connect from the my projects page, to this component, thereby skipping the delay waiting for the registry to respond?

RaymondLuong3:
It is possible that we can remove some logic for loading things from Paratext, but we still need to load projects and resources when users connect for the first time since these are needed so a user can set a source text.
Additionally, we need to hit the "projects" endpoint to access connectable paratext projects so we can get the full name of the project. This could be resolved by passing it in as an additional parameter. That wouldn't be difficult, I can do that.

Nateowami:
Yes, that's true; we do need the full list. However, on current master it doesn't wait for the list of resources to make the component mostly functional. It waits for the list of projects, but if the resources take longer, the page loads before they're fetched. Now that projects are only needed in the same context as resources, it could do that for both of them.

RaymondLuong3:
I have updated the PR to account for when there is no need to get additional projects and resources in the event that the user is just joining an existing project. The my projects page will pass in the project meta data so there is no reason for accessing projects and resources.

marksvc:
If helpful, a diff of another time working on removing the project selector from Connect Project component can be found at 
https://github.com/sillsdev/web-xforge/pull/2441/commits/c76fdf33bcf7ce8eeb020b85ddcac3f10a3a3960 and 
https://github.com/sillsdev/web-xforge/pull/2441/commits/2d16e64e19bc4b07fa3195bbb34ad564183b8460

Nateowami:
@RaymondLuong3 That's a great improvement I didn't even think of. However, I still think we can dramatically improve page load time by not waiting on the projects, just like we don't wait on the resources. See the `populateProjectList()` method.

RaymondLuong3:
Thanks for the suggestion. I have made it so the page is instantly available for both joining and connecting. In the case of connecting, the only thing that waits for the projects to load is the source text input field and the connect button. Apart from that, all settings inputs can be interacted with immediately.

nigel-wells (changes requested):
Always nice to remove so much code. I think we could improve the "Join" flow a bit more. I've I'm Joining then I get presented with a screen that seems a bit pointless along with a "Connect" button which is different from the connect button on the projects screen. Instead, if this user can't do anything other than click the connect button then how about we automate the whole thing and send them straight to the project without having to interact any further?


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-:-O3VC-Ef1T__6vTin8ZQ:b-baquqa)* status: all files reviewed, 7 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts` line 102 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-O3V8wfa0GSNFfs595AE:-O3V8wfa0GSNFfs595AF:b-k0wqu6) ([raw file](https://github.com/sillsdev/web-xforge/blob/af444afd8ff7f28421251c03607da1c9eeca1ef4/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts#L102)):*
> ```XML
> 
>   get projects(): ParatextProject[] {
>     return this.projectsFromParatext != null ? this.projectsFromParatext : [];
> ```

Could be simplified to `return this.projectsFromParatext ?? [];`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts` line 105 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-O3V7_252zaS-AqkkBGu:-O3V7_252zaS-AqkkBGv:b-gcb4mf) ([raw file](https://github.com/sillsdev/web-xforge/blob/af444afd8ff7f28421251c03607da1c9eeca1ef4/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts#L105)):*
> ```XML
>   }
> 
>   get translationSuggestionsEnabled(): boolean {
> ```

nit Can you remove this while you're here as it isn't used any more
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts` line 119 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-O3V9HgQ7jf0c3-3W1ym:-O3V9HgQ7jf0c3-3W1yn:b8s5zkp) ([raw file](https://github.com/sillsdev/web-xforge/blob/af444afd8ff7f28421251c03607da1c9eeca1ef4/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts#L119)):*
> ```XML
>   get projectTitle(): string {
>     if (this.projectMetadata == null) return '';
>     return `${this.projectMetadata.shortName} - ${this.projectMetadata.name}`;
> ```

Would be beneficial to still use `projectLabel` to keep the label consistent with project select autocomplete component otherwise the utility is only ever used with the project auto complete component - there may be some logic in that utility that you need to ensure is still accounted for here as well.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts` line 191 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-O3V7PKF4fKTrnFuuyYB:-O3V7PKF4fKTrnFuuyYC:b-gcb4mf) ([raw file](https://github.com/sillsdev/web-xforge/blob/af444afd8ff7f28421251c03607da1c9eeca1ef4/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts#L191)):*
> ```XML
>   }
> 
>   translateFromSettings(key: string): string {
> ```

nit Can you remove this while you're here as it isn't used any more
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.spec.ts` line 119 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-O3VCtz60o48nQTYav1R:-O3VCtz60o48nQTYav1S:b-v8zsj7) ([raw file](https://github.com/sillsdev/web-xforge/blob/af444afd8ff7f28421251c03607da1c9eeca1ef4/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.spec.ts#L119)):*
> ```XML
> 
>   it('should redirect to my projects if missing paratext id in params', fakeAsync(() => {
>     const env = new TestEnvironment({ useUndefinedParatextId: true });
> ```

Just a note that I found this variable name odd as by default `paratextId` is already considered undefined in `TestEnvironmentParams`. What if you set `paratextId` to `null` and then in this test you set it to `undefined`?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.spec.ts` line 467 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-O3VF0XrDLUWX7QFuwdV:-O3VF0XrDLUWX7QFuwdW:b-g900w0) ([raw file](https://github.com/sillsdev/web-xforge/blob/af444afd8ff7f28421251c03607da1c9eeca1ef4/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.spec.ts#L467)):*
> ```XML
>   }
> 
>   get sourceParatextIdControl(): AbstractControl {
> ```

nit  can be deleted
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.spec.ts` line 526 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-O3VF641CGJogO0X8EUO:-O3VF641CGJogO0X8EUP:b-g900w0) ([raw file](https://github.com/sillsdev/web-xforge/blob/af444afd8ff7f28421251c03607da1c9eeca1ef4/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.spec.ts#L526)):*
> ```XML
>   }
> 
>   getMenuItems(menu: DebugElement): DebugElement[] {
> ```

nit  can be deleted

Nateowami:
> how about we automate the whole thing and send them straight to the project without having to interact any further?

I agree, though I'm not sure the user should even get sent to this component. Perhaps the my projects component should call the endpoint to add the user to a project, and then redirect? I don't have a strong opinion either way; just wanting to bring up the possibility for consideration.

RaymondLuong3:
Hmm, I think that would be reasonable to go straight to the project without ever coming to the connect project page. It shouldn't be too much work to make that happen.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2635)* status: 2 of 8 files reviewed, 2 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts` line 102 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-O3V8wfa0GSNFfs595AE:-O3djthsD3B-tJF8DMrq:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/af444afd8ff7f28421251c03607da1c9eeca1ef4/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts#L102)):*

Could be simplified to `return this.projectsFromParatext ?? [];`
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts` line 119 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-O3V9HgQ7jf0c3-3W1ym:-O3djzpk7jePP1T-f-i-:by3n4uf) ([raw file](https://github.com/sillsdev/web-xforge/blob/af444afd8ff7f28421251c03607da1c9eeca1ef4/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.ts#L119)):*

Would be beneficial to still use `projectLabel` to keep the label consistent with project select autocomplete component otherwise the utility is only ever used with the project auto complete component - there may be some logic in that utility that you need to ensure is still accounted for here as well.
</blockquote></details>

Done. Good thinking!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.spec.ts` line 119 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-O3VCtz60o48nQTYav1R:-O3dk8Nl5lnO5jNE6MGw:b-4e8k5s) ([raw file](https://github.com/sillsdev/web-xforge/blob/af444afd8ff7f28421251c03607da1c9eeca1ef4/src/SIL.XForge.Scripture/ClientApp/src/app/connect-project/connect-project.component.spec.ts#L119)):*

Just a note that I found this variable name odd as by default `paratextId` is already considered undefined in `TestEnvironmentParams`. What if you set `paratextId` to `null` and then in this test you set it to `undefined`?
</blockquote></details>

Good thinking. that was confusing but I think it is more obvious now that we can pass undefined explicitly.

nigel-wells (changes requested):
That's much better thanks. 
I have observed that it can show the project in a not connected and connected state while joining now. It only lasts a few seconds but it is long enough to make me wonder what has happened as a user.


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-:-O3eRjdZD06en-2PY71v:bsc3au0)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html` line 107 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-O3eQPS1BqZxTDcHhryJ:-O3eQPS1BqZxTDcHhryK:b2a06ht) ([raw file](https://github.com/sillsdev/web-xforge/blob/a40e8941e42795710471f917b6f65b36e1a2a250/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html#L107)):*
> ```HTML
>         {{ "connect_project.connect" | transloco }}
>       </a>
>       <button
> ```

Any reason to make this one a `<button>` vs using the `<a>`?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.spec.ts` line 88 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-O3eRb8a2kLIXMGxBRbj:-O3eRb8a2kLIXMGxBRbk:b-vsknmy) ([raw file](https://github.com/sillsdev/web-xforge/blob/a40e8941e42795710471f917b6f65b36e1a2a250/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.spec.ts#L88)):*
> ```XML
>   }));
> 
>   it('click Join, passes PT project id and SF project id', fakeAsync(() => {
> ```

nit  Can you rename the test to reflect what is now happening

RaymondLuong3 (commented):
Thanks. I've added a check to make sure that does not happen.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-:-O3itV13EgcE99PEu2ZA:begsv3x)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html` line 107 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-O3eQPS1BqZxTDcHhryJ:-O3itYQg6wtsSoaISY2M:b1ql7so) ([raw file](https://github.com/sillsdev/web-xforge/blob/a40e8941e42795710471f917b6f65b36e1a2a250/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html#L107)):*

Any reason to make this one a `<button>` vs using the `<a>`?
</blockquote></details>

I made it a button instead of an `<a>` since it doesn't actually link to anything so I thought it would make the most sense as a button.

nigel-wells (approved):
Working as expected now. 
I've left one nit as part of the template syntax conversion - doesn't need to hold up testing though.
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-:-O3jkbZ55bWcOQoMdvxe:b-72xss4)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html` line 107 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-O3eQPS1BqZxTDcHhryJ:-O3jkZpq22ZsBnS20KUl:bqqapj4) ([raw file](https://github.com/sillsdev/web-xforge/blob/a40e8941e42795710471f917b6f65b36e1a2a250/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html#L107)):*

I made it a button instead of an `<a>` since it doesn't actually link to anything so I thought it would make the most sense as a button.
</blockquote></details>

Thanks - makes sense.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html` line 36 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2635#-O3jjLOg0SLl1pnusPSa:-O3jjLOg0SLl1pnusPSb:bglp2jr) ([raw file](https://github.com/sillsdev/web-xforge/blob/1678e07d089f1adbd7f77063834457390e5d60d4/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html#L36)):*
> ```HTML
>     </mat-card>
>   }
>   @if (initialLoadingSFProjects) {
> ```

nit  I realize you converted what was there but this if statement can be combined with the one above it.

PR #2637, opened 2024-07-31
SF-2892 Fix error when Serval admin enables drafting

pmachapman:
This PR fixes the "The project has been deleted or is no longer accessible." dialog appearing when a serval admin enables or disables pre-translation drafting for a project that the user does not have access to.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2637)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -82,7 +82,9 @@ const ROUTES: Route[] = [
   { path: 'projects/:projectId/checking', component: MockComponent },
   { path: 'projects', component: MockComponent },
   { path: 'my-account', component: MockComponent },
-  { path: 'connect-project', component: MockComponent }
+  { path: 'connect-project', component: MockComponent },
+  { path: 'serval-administration', component: MockComponent },
+  { path: 'serval-administration/:projectId', component: MockComponent }
 ];
 
 describe('AppComponent', () => {
@@ -291,6 +293,18 @@ describe('AppComponent', () => {
     // Get past setTimeout to navigation
     tick();
     expect(env.location.path()).toEqual('/projects');
+  }));
+
+  it('response to remote project change for serval admin', fakeAsync(() => {
+    const env = new TestEnvironment();
+    env.setCurrentUser('user05');
+    env.navigate(['/serval-administration', 'project01']);
+    when(mockedLocationService.pathname).thenReturn('/serval-administration/project01');
+    env.init();
+
+    expect(env.selectedProjectId).toEqual('project01');
+    env.updatePreTranslate('project01');
+    verify(mockedDialogService.message(anything())).never();
   }));
 
   it('response to Commenter project role changed', fakeAsync(() => {
@@ -553,6 +567,7 @@ class TestEnvironment {
     this.addUser('user02', 'User 02', 'auth0|user02');
     this.addUser('user03', 'User 03', 'sms|user03');
     this.addUser('user04', 'User 04', 'sms|user04');
+    this.addUser('user05', 'User 05', 'paratext|user05', SystemRole.ServalAdmin);
 
     this.addProject(
       'project01',
@@ -787,6 +802,12 @@ class TestEnvironment {
     this.wait();
   }
 
+  updatePreTranslate(projectId: string): void {
+    const projectDoc = this.realtimeService.get<SFProjectProfileDoc>(SFProjectProfileDoc.COLLECTION, projectId);
+    projectDoc.submitJson0Op(op => op.set<boolean>(p => p.translateConfig.preTranslate, true), false);
+    this.wait();
+  }
+
   addUserToProject(projectId: string): void {
     const projectDoc = this.realtimeService.get<SFProjectProfileDoc>(SFProjectProfileDoc.COLLECTION, projectId);
     projectDoc.submitJson0Op(op => op.set<string>(p => p.userRoles['user01'], SFProjectRole.CommunityChecker), false);
@@ -805,12 +826,13 @@ class TestEnvironment {
     this.wait();
   }
 
-  private addUser(userId: string, name: string, authId: string): void {
+  private addUser(userId: string, name: string, authId: string, systemRole: SystemRole = SystemRole.User): void {
     this.realtimeService.addSnapshot<User>(UserDoc.COLLECTION, {
       id: userId,
       data: createTestUser({
         name,
         authId,
+        roles: [systemRole],
         sites: {
           sf: {
             projects: ['project01', 'project02', 'project03']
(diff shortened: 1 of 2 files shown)


RaymondLuong3 (approved):
I tested this fix and it appears to resolve the issue. Moving to testers.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2637#-:-O38mdPf1PAmvXKpzBss:bbin63o)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 273 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2637#-O38cr5i1bMYbieG91D-:-O38cr5jEIj23WEzqSN7:b6mpq0y) ([raw file](https://github.com/sillsdev/web-xforge/blob/0656e2d56ac563b18260f6efb138df2ebbc2ce43/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L273)):*
> ```XML
> 
>       this.permissionsChangedSub?.unsubscribe();
>       this.permissionsChangedSub = this._selectedProjectDoc.remoteChanges$.subscribe(() => {
> ```

It looks like the `remoteChanges$` property on the project doc is not strictly remote changes, but also local changes too. Looking at the code, it appears that we should only filter for remote changes only.
I wonder if `remoteChanges$` is too specific and there can be multiple instances of the same project doc, and we happen to be subscribed to one that is not the instance that had the changes applied. That's possible.



_Code quote:_
```XML
      this.permissionsChangedSub = this._selectedProjectDoc.remoteChanges$.subscribe(() => {
```

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2637)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 273 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2637#-O38cr5i1bMYbieG91D-:-O39Fm3s4WhccRk6gFlt:buz4ix3) ([raw file](https://github.com/sillsdev/web-xforge/blob/0656e2d56ac563b18260f6efb138df2ebbc2ce43/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L273)):*

It looks like the `remoteChanges$` property on the project doc is not strictly remote changes, but also local changes too. Looking at the code, it appears that we should only filter for remote changes only.
I wonder if `remoteChanges$` is too specific and there can be multiple instances of the same project doc, and we happen to be subscribed to one that is not the instance that had the changes applied. That's possible.
</blockquote></details>

@RaymondLuong3 This property only emits remote changes when running outside of a unit test - unless you see code or behavior elsewhere that I have missed?

However, if the `source` is set to `false`, the `MemoryRealtimeDocAdapter` emits an op as a remote change (via `emitRemoteChange()`), pretending that an op came in remotely. This adapter is only used in unit tests.

Projects can only be changed on the server, so should only ever emit `remoteChange$`.

`changes$` is for getting local and remote changes.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2637)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 273 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2637#-O38cr5i1bMYbieG91D-:-O39KmI97NHXrHs7otpt:b-sj92v7) ([raw file](https://github.com/sillsdev/web-xforge/blob/0656e2d56ac563b18260f6efb138df2ebbc2ce43/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L273)):*

@RaymondLuong3 This property only emits remote changes when running outside of a unit test - unless you see code or behavior elsewhere that I have missed?

However, if the `source` is set to `false`, the `MemoryRealtimeDocAdapter` emits an op as a remote change (via `emitRemoteChange()`), pretending that an op came in remotely. This adapter is only used in unit tests.

Projects can only be changed on the server, so should only ever emit `remoteChange$`.

`changes$` is for getting local and remote changes.
</blockquote></details>

From what I understand, this problem exists because the serval admin is making a change on a particular project, and then `remoteChanges$` emits even though the change came from the current user, even though we would not expect `remoteChanges$` to emit. Is that not the case?

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2637)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 273 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2637#-O38cr5i1bMYbieG91D-:-O39LL5_9awRvE8ROi5n:b8ndueb) ([raw file](https://github.com/sillsdev/web-xforge/blob/0656e2d56ac563b18260f6efb138df2ebbc2ce43/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L273)):*

From what I understand, this problem exists because the serval admin is making a change on a particular project, and then `remoteChanges$` emits even though the change came from the current user, even though we would not expect `remoteChanges$` to emit. Is that not the case?
</blockquote></details>

When the user triggers the JSON-RPC call, the change is made by the backend server directly to the RealtimeServer, so from the web client's perspective, it is a remote change as the op is emitted from the realtime server to the client.

If the user wrote the op locally to the sf_project document in IndexedDB, then it would be a local change (ignoring the fact that sf_project permissions prevent this).

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2637)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 273 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2637#-O38cr5i1bMYbieG91D-:-O39RAlk4iAff8jN-fdO:buwko7z) ([raw file](https://github.com/sillsdev/web-xforge/blob/0656e2d56ac563b18260f6efb138df2ebbc2ce43/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L273)):*

When the user triggers the JSON-RPC call, the change is made by the backend server directly to the RealtimeServer, so from the web client's perspective, it is a remote change as the op is emitted from the realtime server to the client.

If the user wrote the op locally to the sf_project document in IndexedDB, then it would be a local change (ignoring the fact that sf_project permissions prevent this).
</blockquote></details>

Ah yes, I forgot that we actually send a request to the backend to update the project for us. I missed that. So it does make sense that the `remoteChanges$` observable is emitted. Thanks!

PR #2651, opened 2024-08-06
SF-2880 Note icon not appearing after navigating from Users page

pmachapman:
This PR fixes a bug where navigating to the translate area fails to show notes if multiple tabs are open in the target pane.

This bug occurs because the target ViewChild had not loaded when the code that retrieves the notes executes. The solution is to wait one tick for Angular's change detection, then retrieve the note.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2651)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts
@@ -9,8 +9,10 @@ import {
   Inject,
   OnDestroy,
   OnInit,
+  QueryList,
   TemplateRef,
-  ViewChild
+  ViewChild,
+  ViewChildren
 } from '@angular/core';
 import { takeUntilDestroyed } from '@angular/core/rxjs-interop';
 import { MediaObserver } from '@angular/flex-layout';
@@ -195,12 +197,13 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
   sourceSplitHeight: string = '';
   targetSplitHeight: string = '';
   multiCursorViewers: MultiCursorViewer[] = [];
+  target: TextComponent | undefined;
 
   @ViewChild('source') source?: TextComponent;
-  @ViewChild('target') target?: TextComponent;
   @ViewChild('fabButton', { read: ElementRef }) insertNoteFab?: ElementRef<HTMLElement>;
   @ViewChild('fabBottomSheet') TemplateBottomSheet?: TemplateRef<any>;
   @ViewChild('mobileNoteTextarea') mobileNoteTextarea?: ElementRef<HTMLTextAreaElement>;
+  @ViewChildren('target') targetTextComponent?: QueryList<TextComponent>;
 
   private sourceScrollContainer: Element | undefined;
   private targetScrollContainer: Element | undefined;
@@ -634,8 +637,12 @@ export class EditorComponent extends DataLoadingComponent implements OnDestroy, 
     });
 
     this.subscribe(
-      this.activatedRoute.params.pipe(filter(params => params['projectId'] != null && params['bookId'] != null)),
-      async params => {
+      combineLatest([
+        this.activatedRoute.params.pipe(filter(params => params['projectId'] != null && params['bookId'] != null)),
+        this.targetTextComponent!.changes
+      ]),
+      async ([params, components]) => {
+        this.target = components.first;
         this.showSuggestions = false;
         this.sourceLoaded = false;
         this.targetLoaded = false;


RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2651)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 708 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2651#-O3cvV5m0HXtya1umRps:-O3cvV5n5FkIxeiitJBh:b5xb085) ([raw file](https://github.com/sillsdev/web-xforge/blob/acab37d5862aeb35adc36a9556f30a87eeae8514/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L708)):*
> ```XML
>         this.updateVerseNumber();
> 
>         // Set chapter from route if provided
> ```

You'll want to update this comment location.



_Code quote:_
```XML
        // Set chapter from route if provided
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 709 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2651#-O3cuTRe00A5OMDzfqOO:-O3cuTRf4o2UI7RuVi4-:b8cgyvg) ([raw file](https://github.com/sillsdev/web-xforge/blob/acab37d5862aeb35adc36a9556f30a87eeae8514/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L709)):*
> ```XML
> 
>         // Set chapter from route if provided
>         if (this.target == null) {
> ```

I wonder if we should create an observable to emits when the target is available would be a better solution. That way we will always wait for the target to be available before loading the project config. At first glance, that will mean finding which event to listen to. We have an event handler called `onTextLoaded` and that could be helpful.

Alternatively, you can set this delay in the `loadProjectUserConfig` method and trigger the loading of note threads after `setTimeout(() => {})`



_Code quote:_
```XML
        if (this.target == null) {
```

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2651)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 708 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2651#-O3cvV5m0HXtya1umRps:-O3dmUSd9sCcz5ycvkr8:b-npy8yu) ([raw file](https://github.com/sillsdev/web-xforge/blob/acab37d5862aeb35adc36a9556f30a87eeae8514/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L708)):*

You'll want to update this comment location.
</blockquote></details>

Done. I have reworded the comment.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 709 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2651#-O3cuTRe00A5OMDzfqOO:-O3dUJzRBt2PwUGppVyz:byeawr3) ([raw file](https://github.com/sillsdev/web-xforge/blob/acab37d5862aeb35adc36a9556f30a87eeae8514/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L709)):*

I wonder if we should create an observable to emits when the target is available would be a better solution. That way we will always wait for the target to be available before loading the project config. At first glance, that will mean finding which event to listen to. We have an event handler called `onTextLoaded` and that could be helpful.

Alternatively, you can set this delay in the `loadProjectUserConfig` method and trigger the loading of note threads after `setTimeout(() => {})`
</blockquote></details>

I had the delay here, rather than in `loadProjectUserConfig`, as `loadProjectUserConfig` requires the target to be set to perform most of the tasks in its subroutines.

I considered `onTextLoaded`, but that is too late in the process for all of the logic in this function.

I have updated the logic to trigger when the target TextComponent ViewChild is created. Let me know if you think this is a better approach?

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2651)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 709 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2651#-O3cuTRe00A5OMDzfqOO:-O3dqk_JCqnYDaGjm-O2:b-x6g0c9) ([raw file](https://github.com/sillsdev/web-xforge/blob/acab37d5862aeb35adc36a9556f30a87eeae8514/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L709)):*

I had the delay here, rather than in `loadProjectUserConfig`, as `loadProjectUserConfig` requires the target to be set to perform most of the tasks in its subroutines.

I considered `onTextLoaded`, but that is too late in the process for all of the logic in this function.

I have updated the logic to trigger when the target TextComponent ViewChild is created. Let me know if you think this is a better approach?
</blockquote></details>

I think this should be better. Since for the majority of the time when `ngOnInit` is called we expect that the `target` is defined, I think this is a safe change.

PR #2653, opened 2024-08-06
SF-2867 Allow moving backwards via verse with SF note using left arrow

pmachapman:
This PR fixes moving to the previous verse with the arrow key, when the current verse has notes at the start of it (usually SF created notes).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2653)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -1043,7 +1043,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     if (this.editor == null || this.segment == null) {
       return false;
     }
-    const selection = this.editor.getSelection();
+    const selection: RangeStatic | null = this.editor.getSelection();
     if (selection == null) {
       return false;
     }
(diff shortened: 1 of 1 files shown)


RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2653)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 1042 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2653#-O3nG5FV41R-gF7JGA84:-O3nG5FV41R-gF7JGA85:baupdiu) ([raw file](https://github.com/sillsdev/web-xforge/blob/d3eb4787f0063fb22f1f581a9301522c14bf8249/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L1042)):*
> ```XML
>   }
> 
>   private isSelectionAtSegmentPosition(end: boolean): boolean {
> ```

If we strip out the embeds in the segments, we should be able to accurately determine if the selection is at the beginning or end of a segment. That is good.
Does footnotes and cross-references affect this?
Update: They look like it works good with footnotes and cross-references too.



_Code quote:_
```XML
  private isSelectionAtSegmentPosition(end: boolean): boolean {
```

PR #2657, opened 2024-08-06
SF-2866 Update the avatar icon after the user's name changes

pmachapman:
This PR ensures that the user's icon is updated if their name changes, and their icon is set to be their initials.

This was not taking place previously, because when the user's name did change, it was modifying a property in the `UserProfile` specified for the avatar, not changing the reference to the `UserProfile`. To resolve this, the `DoCheck` interface is implemented, which compares the two properties in the `UserProfile` for changes, and triggers the `OnChange` handler.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2657)


RaymondLuong3 (approved):
This is good. Alternatively we could accept the userDoc in this component and listen to changes$, but this solution does what is needed.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2657#-:-O3iw22oBzQS9dT170LL:b-316bp5)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2662, opened 2024-08-08
SF-2906 Fix crash when syncing and notes with duplicate ids are incoming

pmachapman:
This PR fixes a crash noticed in production syncs when notes were incoming that had duplicate ids

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2662)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -1263,7 +1263,9 @@ public class ParatextService : DisposableBase, IParatextService
             IEnumerable<string> newCommentIds = ptCommentIds.Except(matchedCommentIds);
             foreach (string commentId in newCommentIds)
             {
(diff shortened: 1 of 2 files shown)


RaymondLuong3 (approved):
As I was testing this out, I saw that I could create a duplicate note in Paratext, but if I edit on of the notes, the other one is unchanged. I would then have notes with duplicate IDs, but the content would be different. If I synced the notes to SF, the original note content would exist in SF. SF never created the second of the duplicate notes.
Overall, I think fixing the crash is the priority, and getting the duplicate note handling behaviour to be the same in PT and SF is not high priority.
I'll approve on the basis that this will fix the crash and will not cause unwanted side affect.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2662#-:-O3nUAFgEqCqMHhDmaNJ:b-jed5gx)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2666, opened 2024-08-09
Remove unneeded trailing padding from nav menu items

Nateowami:
The menu items in the main navigation have leading and trailing padding of 12px.

The leading padding is needed in order to give the items a consistent layout, but the trailing padding has no effect other than to cut off long menu items.

This is the padding I'm referring to:

This PR removes the trailing padding so that longer strings are not cut off.

Before | After
-------|------
 | 

I am marking this as "testing not required" since there is comprehensive test coverage of visual changes to this component.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2666)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/navigation/navigation.component.scss
@@ -3,7 +3,7 @@
 mat-nav-list .mdc-list-item {
   font-weight: 500;
   font-size: 14px;
-  padding-inline: 12px;
+  padding-inline: 12px 0;
 
   mat-icon {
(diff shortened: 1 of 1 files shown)


pmachapman (approved):
:lgtm: I experimented with other longer strings, and although perhaps an ellipsis could be a better option, I couldn't get that looking right.

I think the proposed solution is the best for now, although when other translations are updated for the "Sync with Paratext" string, such as for Russian, those will be even longer.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2666#-:-O424qIsBpvDhMiNXm-M:b-qwqk6x)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2669, opened 2024-08-12
SF-2908 Fix crash when syncing specific Paratext Resources

pmachapman:
This PR fixes a crash when connecting to or syncing older DBL Resources, such as the SBLGNT.

This crash occurs because the .SSF file does not contain the language identifier. Paratext usually works around this by running an migrator on the project, which only changes the project in memory.

To resolve the crash, I have reimplemented the migrator in a way that works with Scripture Forge.

If you wish to test this on your developer machine, use the resource "apw - The New Testament of our Lord and Saviour Jesus Christ" which suffers from the same bug.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2669)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -27,6 +27,7 @@ using Newtonsoft.Json.Linq;
 using Paratext.Data;
 using Paratext.Data.Languages;
 using Paratext.Data.ProjectComments;
+using Paratext.Data.ProjectFileAccess;
 using Paratext.Data.ProjectSettingsAccess;
 using Paratext.Data.RegistryServerAccess;
 using Paratext.Data.Repository;
@@ -2355,6 +2356,13 @@ public class ParatextService : DisposableBase, IParatextService
     {
         if (resource.InstallableResource != null)
         {
+            // Correct the language code for old resources
+            LanguageId? overrideLanguageId = null;
+            if (DetermineBestLanguageForResource(resource.InstallableResource.ExistingScrText))
+            {
+                overrideLanguageId = resource.InstallableResource.ExistingScrText?.Settings.LanguageID;
+            }
+
             // Install the resource if it is missing or out of date
             if (
                 !resource.IsInstalled
@@ -2364,6 +2372,12 @@ public class ParatextService : DisposableBase, IParatextService
             {
                 resource.InstallableResource.Install();
                 needsToBeCloned = true;
+
+                // On first install, we will now have an existing ScrText, so check the language is OK
+                if (DetermineBestLanguageForResource(resource.InstallableResource.ExistingScrText))
+                {
+                    overrideLanguageId = resource.InstallableResource.ExistingScrText?.Settings.LanguageID;
+                }
             }
 
             // Extract the resource to the source directory
@@ -2372,7 +2386,7 @@ public class ParatextService : DisposableBase, IParatextService
                 string path = LocalProjectDir(targetParatextId);
                 _fileSystemService.CreateDirectory(path);
                 resource.InstallableResource.ExtractToDirectory(path);
-                MigrateResourceIfRequired(username, targetParatextId);
+                MigrateResourceIfRequired(username, targetParatextId, overrideLanguageId);
             }
         }
         else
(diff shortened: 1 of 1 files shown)


josephmyers (changes requested):
So, I am able to successfully see the source (apw) showing up for NT books, as your steps indicate. However, is this message now expected? The language pretty clearly is "apache (apw)" in Paratext, though it sounds like there are problems getting that.



Also we need unit tests for this.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2669#-:-O4EEEyD2-HKIV8AmeD8:b-w3w2hk)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2397 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2669#-O4EG0dX4BSZnIQZzV-8:-O4EG0dX4BSZnIQZzV-9:b-x728xw) ([raw file](https://github.com/sillsdev/web-xforge/blob/08483573009b0a165fd077f63d159ad19135573a/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2397)):*
> ```Smalltalk
>     /// </para>
>     /// </remarks>
>     private static bool DetermineBestLanguageForResource(ScrText? scrText)
> ```

Determine* is a bit of an odd word choice here -- how about Set*?
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2447 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2669#-O4EGya7E7vrhmwVFU9W:-O4EGya7E7vrhmwVFU9X:blpn2n6) ([raw file](https://github.com/sillsdev/web-xforge/blob/08483573009b0a165fd077f63d159ad19135573a/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2447)):*
> ```Smalltalk
>         if (overrideLanguage is not null)
>         {
>             scrText.Settings.LanguageID = overrideLanguage;
> ```

Is it strange that we're setting the LanguageID again, when we just did in the method you added? Am I reading this wrong?

pmachapman (commented):
> So, I am able to successfully see the source (apw) showing up for NT books, as your steps indicate. However, is this message now expected? The language pretty clearly is "apache (apw)" in Paratext, though it sounds like there are problems getting that.

I split that into a separate issue, https://jira.sil.org/browse/SF-2912.

>Also we need unit tests for this.

Hmmm will need to think how we can unit test this as it is heavy on internal ParatextData logic and DBL API calls, and there are no existing tests in the code that loads resources.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2669#-:-O4HKF8c3zSQ7CgxxDHA:bi5u36c)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2397 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2669#-O4EG0dX4BSZnIQZzV-8:-O4HKbgJ87MRv2id2L9f:b4airlo) ([raw file](https://github.com/sillsdev/web-xforge/blob/08483573009b0a165fd077f63d159ad19135573a/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2397)):*

Determine* is a bit of an odd word choice here -- how about Set*?
</blockquote></details>

I called it determine as that is the name of the Paratext method that does the same (see the Paratext source code for the parent implementation).

I am not too keen on renaming it for that reason, but if you think it is important I can.
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2447 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2669#-O4EGya7E7vrhmwVFU9W:-O4HKphb1wEF34oEiZSX:b-vsh767) ([raw file](https://github.com/sillsdev/web-xforge/blob/08483573009b0a165fd077f63d159ad19135573a/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2447)):*

Is it strange that we're setting the LanguageID again, when we just did in the method you added? Am I reading this wrong?
</blockquote></details>

This is another quirk of the internal logic for Paratext. The resource in this case has actually changed into a project in the eyes of Paratext data, and so the value has reset in the edge cases that cause this bug to occur. Another quirk of `ScrText` for projects with incomplete projects.

josephmyers (changes requested):
Got it, thanks.

In my mind, having this bugfix covered would be useful. Obviously the covering here is against regressions. I think it's even more important for us to have, in general, some tests for loading resources, right? That hasn't been done because of the difficulty?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2669#-:-O4IOoBQ4o5RiAX8yZSh:b4z7e7f)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2397 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2669#-O4EG0dX4BSZnIQZzV-8:-O4IYgdNBwQ4sbuagz3y:bj1tf4y) ([raw file](https://github.com/sillsdev/web-xforge/blob/08483573009b0a165fd077f63d159ad19135573a/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2397)):*

I called it determine as that is the name of the Paratext method that does the same (see the Paratext source code for the parent implementation).

I am not too keen on renaming it for that reason, but if you think it is important I can.
</blockquote></details>

Bummer, dude! I do think it's important, but not that important. It's just not obvious (at all) that this method is modifying the input. Its return type is just as bad...
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2447 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2669#-O4EGya7E7vrhmwVFU9W:-O4J5hkg3aytjPlSSS_m:b-oh4quc) ([raw file](https://github.com/sillsdev/web-xforge/blob/08483573009b0a165fd077f63d159ad19135573a/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2447)):*

This is another quirk of the internal logic for Paratext. The resource in this case has actually changed into a project in the eyes of Paratext data, and so the value has reset in the edge cases that cause this bug to occur. Another quirk of `ScrText` for projects with incomplete projects.
</blockquote></details>

Ah, ok. I think this is worth noting as a comment here. Quirks like this could understandably get refactored away in the future, and having test coverage would prevent this.

pmachapman (commented):
Yes, it is in the quite hard bucket. When I originally wrote the feature we weren't able to mock some aspects of ParatextData, but now we can so it may be possible now.

I can look at it next week in a separate PR if you don't want to block this one? I think it will be 1-2 days work to write these.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2669#-:-O4JE4-MEw6l7JjQyPxB:b-1odoxj)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2447 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2669#-O4EGya7E7vrhmwVFU9W:-O4JE_ke4-8JppsbtRQL:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/08483573009b0a165fd077f63d159ad19135573a/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2447)):*

Ah, ok. I think this is worth noting as a comment here. Quirks like this could understandably get refactored away in the future, and having test coverage would prevent this.
</blockquote></details>

Done.

josephmyers (approved):
Thanks. I've created SF-2929 for this 
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2669#-:-O4cFjcqAKYAuf1wqkP5:be0pmvc)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

PR #2670, opened 2024-08-12
SF-2901 Display completion status via new ProgressService

josephmyers:
Split from translate-overview, this class calculates the filled segments for the entire project. It extends the DataLoadingComponent and expands it to include an isLoaded event, distinct from isLoading and necessary in order to wait for loading to finish, as opposed to simply "not loading." ProgressService needs to have an initialize method to do this setup, even though it's a service, since the required projectId is given as a route parameter.

As the calculation is a costly operation, I added an initialize() call to the draft-generation-steps, in addition to the chip selection. The latter may be redundant in the current usage, but it will be necessary for reusing its component. The former starts the loading as soon as the page starts loading, as opposed to whenever the chips are loading in (a difference of a few seconds due to the animations).

As for the front end, I diverged a little from the original Balsamiq design. The original RYG coloring scheme didn't work due to low contrast in cases where the chips were selected. Plus, all the added colors were jarring and made the page a little overwhelming. Finally, it wasn't clear what the colors were actually indicating. The current implementation is clearer, though it may yet be valuable to add further clarification, e.g. in the form of a hover item.

Note that the below UI has gone through a number of iterations and is somewhat in a state of flux. Feel free to recommend changes or to just say you don't like it :D

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2670)


pmachapman (changes requested):
Just a couple of nits.

I'm not sure about using the bottom border to show progress...it feels a little unintuitive? Is it not possible to have a progress donut in the chip?

> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2670#-:-O4CLGrIA6asltDOXH4n:b-uyk6cl)* status: all files reviewed, 2 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.ts` line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2670#-O4CH1rxBVc0BmIdReZ0:-O4CH1rxBVc0BmIdReZ1:b-yyfhgt) ([raw file](https://github.com/sillsdev/web-xforge/blob/8abf36b6856b2d99e240d1b876d6380c43bef61f/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.ts#L8)):*
> ```XML
> import { filter, firstValueFrom, map } from 'rxjs';
> import { UICommonModule } from 'xforge-common/ui-common.module';
> import { SubscriptionDisposable } from '../../../xforge-common/subscription-disposable';
> ```

NIT: Remove "../../../"



_Code quote:_
```XML
import { SubscriptionDisposable } from '../../../xforge-common/subscription-disposable';
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/progress-service/progress-service.ts` line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2670#-O4CHcpC9iKyHGQSSVoi:-O4CHcpC9iKyHGQSSVoj:b-yyfhgt) ([raw file](https://github.com/sillsdev/web-xforge/blob/8abf36b6856b2d99e240d1b876d6380c43bef61f/src/SIL.XForge.Scripture/ClientApp/src/app/shared/progress-service/progress-service.ts#L8)):*
> ```XML
> import { DataLoadingComponent } from '../../../xforge-common/data-loading-component';
> import { NoticeService } from '../../../xforge-common/notice.service';
> import { OnlineStatusService } from '../../../xforge-common/online-status.service';
> ```

NIT: Remove "../../../"



_Code quote:_
```XML
import { DataLoadingComponent } from '../../../xforge-common/data-loading-component';
import { NoticeService } from '../../../xforge-common/notice.service';
import { OnlineStatusService } from '../../../xforge-common/online-status.service';
```

josephmyers (commented):
This isn't something we had considered as a design, at least not that was voiced. I'll look into it. In general, it seems like lots of people have complaints with the line. Maybe it was better in its idea form than implementation? I do think a donut would be as clear or clearer. We won't be able to use the SF green, due to contrast with the blue. A further potential conflict is the train/draft by chapter feature, whose menu button would be where I'd also put the donut. Then again, this may all get reworked with mixed sources anyway.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2670#-:-O4EQwdLBOrOU4HVmFXc:bhiun2n)* status: 10 of 12 files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (approved):
No worries. I will approve this PR and leave it up to you on whether you redesign in this PR, leave as-is, or choose another path.

:lgtm: The concept and code looks good.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2670#-:-O4HIUifB4w1QjP48Zuu:b5w9t0y)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers:
image

Here's what it looks like with the donut. I don't think it's bad, but I do like the line under better aesthetically. Unless you're of the mind that the donut is significantly better, for whatever reason, I'd say we move forward with the line. It's easy to make changes like this, if we want to in the future.

pmachapman (approved):
Thank you for taking the time to do the demonstration. I think it looks better, but go forward with the line for now. We can always change it if users are confused by the line.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2670#-:-O4cqrIT10J_GelzdjnZ:b-dr04kf)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

PR #2671, opened 2024-08-12
SF-2909 Biblical Terms Enhancements

pmachapman:
This PR enhances Biblical Terms to include support for user requested features from Paratext.

These include:

 * Moving Biblical Terms to a tab
 * Adding filtering by category for Biblical Terms
 * Adding filtering by verse, category, and book for Biblical Terms.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671)


josephmyers (changes requested):
Hey, great work so far. I think this way is more discoverable than toggling it from the settings cog. Here are a couple areas where we could use some padding:



I am also wondering if we should move the "transliterate" toggle into this tab somewhere. As it is, it's pretty hard to find. Have you considered this? Thanks

> Reviewed 32 of 32 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-:-O4nmSH2C4TXF2KaCjRm:b-5467tc)* status: all files reviewed, 4 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-header/tab-header.component.scss` line 77 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4naQ9EExGOz8PvkcMb:-O4naQ9EExGOz8PvkcMc:bsuezbz) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-header/tab-header.component.scss#L77)):*
> ```SCSS
>     align-items: center;
>     justify-content: center;
>     vertical-align: middle;
> ```

Of these four, I am only seeing vertical-align have an effect
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 65 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4nn0aZD5DCU5hcO17L:-O4nn0aZD5DCU5hcO17M:b63512h) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L65)):*
> ```HTML
>       <tr mat-footer-row *matFooterRowDef="['id']; sticky: true"></tr>
>       <tr mat-row *matNoDataRow>
>         <td class="mat-cell" [attr.colspan]="columnsToDisplay.length"></td>
> ```

We should put some text here, explaining that there are no results for the selection.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 60 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4nnVaf2zlxMm6NhWxg:-O4nnVah4aCq3OyEQwuL:bxtgnz7) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L60)):*
> ```XML
> const defaultLocaleCode = I18nService.defaultLocale.canonicalTag;
> 
> type ViewFilter = 'current_verse' | 'current_chapter' | 'current_book';
> ```

I'd rather call this (and related properties) something like RangeFilter, since Category is also a way of filtering.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.spec.ts` line 184 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4noP8h3opj_rvMjiRe:-O4noP8h3opj_rvMjiRf:bildvdg) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.spec.ts#L184)):*
> ```XML
>   }));
> 
>   it('should update the categories when the language changes', fakeAsync(() => {
> ```

This is a good test; however, I notice in production that changing the language resets the category and range without actually updating the list. Actually, it is clearing the list.

It'd be safer and more robust to convert the filter and category properties, which are currently just simple strings, to have setters, within which we refresh the display. Similar with `selectedCategory` and `categories`, which I believe are a source of the problem. I'm not a huge fan of "dumb" simple types driving business logic, as what is essentially state logic is bug prone.

The bug occurs because, after we've updated `rows`, we reset the `selectedCategory` if the current selection is not present in the category list. This will always be the case when we change language, due to us not keeping our properties up to date.

This would have been a tiny use case, but we should fix it anyway. And please add a unit test for it!

pmachapman (commented):
> Here are a couple areas where we could use some padding:

Done. Thank you!

> I am also wondering if we should move the "transliterate" toggle into this tab somewhere. As it is, it's pretty hard to find. Have you considered this?

I am going to think on this over the weekend. I will see if there is a nice way to include it in the UI without it being buried in settings (like you, I don't like how it is currently buried).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-:-O4rwvPqE19jxBleNtTF:brz7zyl)* status: 27 of 32 files reviewed, 4 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-header/tab-header.component.scss` line 77 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4naQ9EExGOz8PvkcMb:-O4rty1oEnPSjHC1a9O9:b-ioxi27) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-header/tab-header.component.scss#L77)):*

Of these four, I am only seeing vertical-align have an effect
</blockquote></details>

Done. Thank you!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 65 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4nn0aZD5DCU5hcO17L:-O4ru3JsA6tFxlACGFC3:by2gpkg) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L65)):*

We should put some text here, explaining that there are no results for the selection.
</blockquote></details>

It displays while loading, so I was hesitant to put a message here. I only added the row so during the loading phase, the bottom row would not be jammed up against the header row.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 60 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4nnVaf2zlxMm6NhWxg:-O4ruGly9jJ8siTtU_Pt:b-8fwqts) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L60)):*

I'd rather call this (and related properties) something like RangeFilter, since Category is also a way of filtering.
</blockquote></details>

Done. Thank you for the naming suggestion - I struggled to name this one!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.spec.ts` line 184 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4noP8h3opj_rvMjiRe:-O4rwk358AhIMSFYuZco:bawzoth) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.spec.ts#L184)):*

This is a good test; however, I notice in production that changing the language resets the category and range without actually updating the list. Actually, it is clearing the list.

It'd be safer and more robust to convert the filter and category properties, which are currently just simple strings, to have setters, within which we refresh the display. Similar with `selectedCategory` and `categories`, which I believe are a source of the problem. I'm not a huge fan of "dumb" simple types driving business logic, as what is essentially state logic is bug prone.

The bug occurs because, after we've updated `rows`, we reset the `selectedCategory` if the current selection is not present in the category list. This will always be the case when we change language, due to us not keeping our properties up to date.

This would have been a tiny use case, but we should fix it anyway. And please add a unit test for it!
</blockquote></details>

I'm not sure I understand - can we meet early next week to discuss this?

josephmyers (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671)* status: all files reviewed, 3 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 65 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4nn0aZD5DCU5hcO17L:-O4sFaN5B3i1RVruOF69:b1dndpu) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L65)):*

It displays while loading, so I was hesitant to put a message here. I only added the row so during the loading phase, the bottom row would not be jammed up against the header row.
</blockquote></details>

Actually, I just added an `isLoaded` property to `DataLoadingComponent`, which helps us determine the difference between first-time loading and dynamic loading. I think you could harness that.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.scss` line 58 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4sPHc-4a6Khp44cRnC:-O4sPHc-4a6Khp44cRnD:b-e4bn1m) ([raw file](https://github.com/sillsdev/web-xforge/blob/d6fc82f2b8e4868a6d93d5a16bc9e7e418a021a1/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.scss#L58)):*
> ```SCSS
> .mat-mdc-header-cell:first-child,
> .mat-mdc-cell:first-child {
>   padding-left: 1.25em;
> ```

I haven't looked, but does work in RTL situations? We typically prefer `padding-inline-start`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.spec.ts` line 184 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4noP8h3opj_rvMjiRe:-O4sPUku1VBVloQHJUrz:b-hbl0z2) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.spec.ts#L184)):*

I'm not sure I understand - can we meet early next week to discuss this?
</blockquote></details>

Sure, sorry for not being clear. Let's talk next week!

pmachapman (commented):
> I am also wondering if we should move the "transliterate" toggle into this tab somewhere. As it is, it's pretty hard to find. Have you considered this?

I have added a settings popup menu at the bottom of the tab. Lett me know if this works better for you!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-:-O5AIUDU1F3q3_8kxE4D:b-e1v99j)* status: all files reviewed, 3 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 65 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4nn0aZD5DCU5hcO17L:-O5AMOFb1PW2TkyZYY8Y:b-35rxtl) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L65)):*

Actually, I just added an `isLoaded` property to `DataLoadingComponent`, which helps us determine the difference between first-time loading and dynamic loading. I think you could harness that.
</blockquote></details>

Thank you! I have added support for this, and slightly extended your implementation.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.scss` line 58 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4sPHc-4a6Khp44cRnC:-O5AMgFW6f0ghzKUDP1O:b-r21xql) ([raw file](https://github.com/sillsdev/web-xforge/blob/d6fc82f2b8e4868a6d93d5a16bc9e7e418a021a1/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.scss#L58)):*

I haven't looked, but does work in RTL situations? We typically prefer `padding-inline-start`
</blockquote></details>

Fixed - thank you for catching that!

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671)* status: 22 of 33 files reviewed, 3 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.spec.ts` line 184 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4noP8h3opj_rvMjiRe:-O5BUWEoAFYT3xE2cyIP:b-leg6uq) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.spec.ts#L184)):*

Sure, sorry for not being clear. Let's talk next week!
</blockquote></details>

I have fixed the bug, and updated the unit test.

josephmyers (commented):
Can you add tooltips to the Actions buttons?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-:-O5CYxDPAoKMzO1bV3gI:b-3omlom)* status: 22 of 33 files reviewed, 3 unresolved discussions

pmachapman (commented):
> Can you add tooltips to the Actions buttons?

I have added very brief tooltips, as the functionality for the actions varies a lot based on whether the user has edit permission, or add note permission for the project.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-:-O5EWT80E0oqYaSl3CiD:b-2bqz1c)* status: 22 of 33 files reviewed, 3 unresolved discussions (waiting on @josephmyers)

josephmyers (changes requested):
I do notice that if the BT tab is open and you disable BT in the Settings, the BT tab stays open upon returning to the editor. Small issue, but we should fix it.

> Reviewed 2 of 6 files at r3, 6 of 7 files at r4, 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-:-O5HPX8REwjljsQ6BCPs:b-ujqt4b)* status: all files reviewed, 4 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 65 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4nn0aZD5DCU5hcO17L:-O5HO-qp8C4zovs5M6CN:b-t2th7z) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L65)):*

Thank you! I have added support for this, and slightly extended your implementation.
</blockquote></details>

I notice that while the list is loading, especially page load, we get the "No Biblical Terms match your criteria" message. Is there a way to avoid this?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 31 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5HOjv2Exg4EHnS1jkb:-O5HOjv2Exg4EHnS1jkc:bomjhbu) ([raw file](https://github.com/sillsdev/web-xforge/blob/f1c62ece1f50dc8bb885e0e906989494ab8f1c69/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L31)):*
> ```HTML
>             <mat-icon [style.transform]="row.transformNotesIcon">{{ row.notesIcon }}</mat-icon>
>           </button>
>           <button mat-icon-button (click)="editRendering(row.id)" [matTooltip]="t('renderings')">
> ```

Can this say "Edit Renderings"?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 36 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5HU6Nc-uYML9GJYg_K:-O5HU6Nc-uYML9GJYg_L:by56tjx) ([raw file](https://github.com/sillsdev/web-xforge/blob/f1c62ece1f50dc8bb885e0e906989494ab8f1c69/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L36)):*
> ```HTML
>         </td>
>         <td *matFooterCellDef [attr.colspan]="columnsToDisplay.length">
>           <div class="toolbar">
> ```

I don't want to be too picky here, but why is the toolbar at the bottom of this tab? Was this intentional? All our other tabs place the toolbar above the content, and it's weird that this one is backwards. Do we gain something from putting it at the bottom?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 66 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5HThHZ8NPtGRZO-s4C:-O5HThHZ8NPtGRZO-s4D:b9ahx0i) ([raw file](https://github.com/sillsdev/web-xforge/blob/f1c62ece1f50dc8bb885e0e906989494ab8f1c69/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L66)):*
> ```HTML
>               [matMenuTriggerFor]="menu"
>               class="mat-menu-popup-button"
>               [matTooltip]="t('settings')"
> ```

I think this tooltip should just be "Settings" -- simpler
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.spec.ts` line 184 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4noP8h3opj_rvMjiRe:-O5HeVgv4e62SUSiWba7:b-qmfkiv) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.spec.ts#L184)):*

I have fixed the bug, and updated the unit test.
</blockquote></details>

Thanks. You did fix the bug, but you've introduced some slightly odd behavior to cope with the persistent "properties are out of date" issue in this class. By that I mean that it is weird and sad that we have to reset their category and range when the language changes. We're doing that to cope with the fact that the category, which gets updated, isn't found in the list of categories, because they're not updated til later. It's a code smell.

I'm sorry for belaboring the point, so I'm resolving this.

pmachapman (commented):
> I do notice that if the BT tab is open and you disable BT in the Settings, the BT tab stays open upon returning to the editor. Small issue, but we should fix it.

Good spotting. Thank you!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-:-O5KGzKv5ifisXBDsCMf:barn5ka)* status: 28 of 33 files reviewed, 4 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 65 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4nn0aZD5DCU5hcO17L:-O5LNdcF8aqonmgCK5da:b4dcewm) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L65)):*

I notice that while the list is loading, especially page load, we get the "No Biblical Terms match your criteria" message. Is there a way to avoid this?
</blockquote></details>

Done. This was harder than I thought, but the net result is much better for the observables and subscriptions, resulting in much less computation occuring.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 31 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5HOjv2Exg4EHnS1jkb:-O5L2Jgq1JJ2HjYITxqE:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/f1c62ece1f50dc8bb885e0e906989494ab8f1c69/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L31)):*

Can this say "Edit Renderings"?
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 36 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5HU6Nc-uYML9GJYg_K:-O5LA5T_56sPlrQKCeOT:b-wofev7) ([raw file](https://github.com/sillsdev/web-xforge/blob/f1c62ece1f50dc8bb885e0e906989494ab8f1c69/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L36)):*

I don't want to be too picky here, but why is the toolbar at the bottom of this tab? Was this intentional? All our other tabs place the toolbar above the content, and it's weird that this one is backwards. Do we gain something from putting it at the bottom?
</blockquote></details>

I was kind of forced. The column headings are sticky to the top, and having another sticky row at the top did not work in my experiments. I would really, really, prefer them to be at the top, but my limited CSS skills are failing me.

If you have a way to have both the toolbar and the column headings sticky at the top, that would be amazing, and I really would be thankful!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 66 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5HThHZ8NPtGRZO-s4C:-O5LAZCo3dnDL0Xfv-t6:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/f1c62ece1f50dc8bb885e0e906989494ab8f1c69/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L66)):*

I think this tooltip should just be "Settings" -- simpler
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.spec.ts` line 184 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4noP8h3opj_rvMjiRe:-O5LAhFqCnt31S9l5cJO:b-xn3wxh) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.spec.ts#L184)):*

Thanks. You did fix the bug, but you've introduced some slightly odd behavior to cope with the persistent "properties are out of date" issue in this class. By that I mean that it is weird and sad that we have to reset their category and range when the language changes. We're doing that to cope with the fact that the category, which gets updated, isn't found in the list of categories, because they're not updated til later. It's a code smell.

I'm sorry for belaboring the point, so I'm resolving this.
</blockquote></details>

Yes, it is a code smell. Or perhaps a data smell. :-(

josephmyers (changes requested):
Cool, that's fixed. I am also seeing, though I'm not sure how long it's been here, a small graphical bug with the filter on page load. This happens when the Category is set to something other than All. Here's a brief video: 
[Recording 2024-08-28 082834.mp4](https://files.reviewableusercontent.io/images/reviewable/43733878/UVDToxZFF1Y2UR_CYe4Jl4tG/Recording%202024-08-28%20082834.mp4)

I may have found another bug. Unless we're wanting to allow multiple instances of the tab?


I think both the above issues should have test coverage, though I could give the former a pass if it's difficult.

I'm also seeing this tab take a noticeable amount of time to close. It's about a second, and I wasn't seeing this yesterday. Any ideas?

> Reviewed 5 of 5 files at r6, 4 of 4 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-:-O5LaLyvCm1TG7RkstPb:bjfuryq)* status: all files reviewed, 8 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 65 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4nn0aZD5DCU5hcO17L:-O5M1dyc-hAFAjUJiXLp:b-mbde5k) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L65)):*

Done. This was harder than I thought, but the net result is much better for the observables and subscriptions, resulting in much less computation occuring.
</blockquote></details>

Looks so much better! Thanks for doing this. Can you add two tests that verify the message is not shown when loading and shown when done?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 36 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5HU6Nc-uYML9GJYg_K:-O5MQ4mb0h56-6e4JHmD:b-qf9y3n) ([raw file](https://github.com/sillsdev/web-xforge/blob/f1c62ece1f50dc8bb885e0e906989494ab8f1c69/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L36)):*

I was kind of forced. The column headings are sticky to the top, and having another sticky row at the top did not work in my experiments. I would really, really, prefer them to be at the top, but my limited CSS skills are failing me.

If you have a way to have both the toolbar and the column headings sticky at the top, that would be amazing, and I really would be thankful!
</blockquote></details>

It's not CSS -- it's HTML. Just move the `<div class\="toolbar"\>` above the `table`


___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 183 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5MGgMC7GpKA_iwpAs4:-O5MGgMC7GpKA_iwpAs5:bhs457m) ([raw file](https://github.com/sillsdev/web-xforge/blob/fb6afc2d90a59bd084383135838ff2566c60cd5d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L183)):*
> ```XML
>   categories: string[] = [];
>   columnsToDisplay = ['term', 'category', 'gloss', 'renderings', 'id'];
>   rangeFilters: RangeFilter[] = ['current_verse', 'current_chapter', 'current_book'];
> ```

Can this and the one above be readonly?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 294 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5MIp6L7_ooPfk7FKhh:-O5MIp6L7_ooPfk7FKhi:b-g5oob3) ([raw file](https://github.com/sillsdev/web-xforge/blob/fb6afc2d90a59bd084383135838ff2566c60cd5d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L294)):*
> ```XML
>       this.projectDoc = await this.projectService.getProfile(projectId);
>       this.projectUserConfigDoc = await this.projectService.getUserConfig(
>         this.activatedProjectService.projectId ?? projectId,
> ```

Why do we need both here? Is there a chance that `projectId` is wrong? Whichever we need, it feels like the other one would cause problems.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 303 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5MJbEJ6CgRHOyETEHU:-O5MJbEJ6CgRHOyETEHV:bpeeeb8) ([raw file](https://github.com/sillsdev/web-xforge/blob/fb6afc2d90a59bd084383135838ff2566c60cd5d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L303)):*
> ```XML
>       // Subscribe to any project, book, chapter, verse, locale, biblical term, or note changes
>       this.loadingStarted();
>       let biblicalTermsAndNotesChanges$: Observable<any> = await this.getBiblicalTermsAndNotesChanges(projectId);
> ```

Make const, please
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 310 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5MJymB8yD4TrRBg_os:-O5MJymC8yNKXp_4TbzR:b-trdhgz) ([raw file](https://github.com/sillsdev/web-xforge/blob/fb6afc2d90a59bd084383135838ff2566c60cd5d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L310)):*
> ```XML
>           this.chapter$,
>           this.verse$,
>           this.i18n.locale$,
> ```

Obviously there is some inefficiency in our data structure for BT, but conceptually, why does changing the locale affect the filtering? That doesn't seem right.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1187 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5MNgCZ-KXyRgcbayKR:-O5MNgCZ-KXyRgcbayKS:b-bx28yk) ([raw file](https://github.com/sillsdev/web-xforge/blob/fb6afc2d90a59bd084383135838ff2566c60cd5d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1187)):*
> ```XML
>             let projectDoc: SFProjectProfileDoc | undefined = undefined;
>             if (tabData.projectId != null) {
>               projectDoc = await this.projectService.getProfile(tabData.projectId);
> ```

So, looking at this if block, wouldn't this include BT tabs whose `projectId` is populated? If it has a `projectId`, a BT tab will use the `projectService`, but if it doesn't, it will use the `activatedProject`. That seems to contradict the intent of your comment below. Maybe further comments are necessary? Or flip these two conditionals?

I see the tests you added for this, but we should probably cover this specific use case, as well.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.ts` line 25 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5MR-mW0nsi9sAIVDEA:-O5MR-mXEg9lBJIOQFxm:blee6zk) ([raw file](https://github.com/sillsdev/web-xforge/blob/99c77960faedc654dc189e52cb546339b1f5a6d3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.ts#L25)):*
> ```XML
>             svgIcon: 'biblical_terms',
>             headerText: await firstValueFrom(
>               this.i18n.translate('editor_tab_factory.default_biblical_terms_tab_header')
> ```



The tab title is weird when you reload the page. Is that coming from here?

pmachapman (commented):
> I am also seeing, though I'm not sure how long it's been here, a small graphical bug with the filter on page load. This happens when the Category is set to something other than All. Here's a brief video:

I think I have found the cause of that (the `selectedCategory` being set too early) - please check that I have fixed it for you.

> I may have found another bug. Unless we're wanting to allow multiple instances of the tab?

I think this is OK, as the user may want one tab in each pane, or have one tab filtering by chapter, the other by verse, etc. I think we should allow the UI to be as flexible as possible, so long as there are no technical limitations.

> I'm also seeing this tab take a noticeable amount of time to close. It's about a second, and I wasn't seeing this yesterday. Any ideas?

I don't have that happen to me?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-:-O5PrZ9W2sAZUkJv3Bvi:b1a3fz3)* status: 29 of 33 files reviewed, 8 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 65 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O4nn0aZD5DCU5hcO17L:-O5QQ5hv6mrSzeqprXPe:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/18868b81e9175f9b21a32a62472b5f3b7a978b2b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L65)):*

Looks so much better! Thanks for doing this. Can you add two tests that verify the message is not shown when loading and shown when done?
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html` line 36 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5HU6Nc-uYML9GJYg_K:-O5PrHLr2YBUnOx9IKCz:b1jy3f6) ([raw file](https://github.com/sillsdev/web-xforge/blob/f1c62ece1f50dc8bb885e0e906989494ab8f1c69/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.html#L36)):*

It's not CSS -- it's HTML. Just move the `<div class\="toolbar"\>` above the `table`


</blockquote></details>

I did, but the `sticky` CSS that the `mat-table` uses for the table headers makes the header cover the toolbar when scrolling:
[Recording 2024-08-29 092048.mp4](https://files.reviewableusercontent.io/images/reviewable/8665431/hGRYXSjLyv4JsK0tBGMTHspM/Recording%202024-08-29%20092048.mp4)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 183 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5MGgMC7GpKA_iwpAs4:-O5Pxilx2adfwBoEVT6q:bugs9as) ([raw file](https://github.com/sillsdev/web-xforge/blob/fb6afc2d90a59bd084383135838ff2566c60cd5d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L183)):*

Can this and the one above be readonly?
</blockquote></details>

Done. Good spotting.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 294 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5MIp6L7_ooPfk7FKhh:-O5Pxs_3B1RrFwHyanjV:b-gurfdz) ([raw file](https://github.com/sillsdev/web-xforge/blob/fb6afc2d90a59bd084383135838ff2566c60cd5d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L294)):*

Why do we need both here? Is there a chance that `projectId` is wrong? Whichever we need, it feels like the other one would cause problems.
</blockquote></details>

We don't need it anymore. Removed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 303 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5MJbEJ6CgRHOyETEHU:-O5PybXDAxLBrsrmIyL4:boyexb3) ([raw file](https://github.com/sillsdev/web-xforge/blob/fb6afc2d90a59bd084383135838ff2566c60cd5d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L303)):*

Make const, please
</blockquote></details>

Done. Thanks!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 310 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5MJymB8yD4TrRBg_os:-O5Pydsf-vjjGZeBTi2U:bt34e9g) ([raw file](https://github.com/sillsdev/web-xforge/blob/fb6afc2d90a59bd084383135838ff2566c60cd5d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L310)):*

Obviously there is some inefficiency in our data structure for BT, but conceptually, why does changing the locale affect the filtering? That doesn't seem right.
</blockquote></details>

Because the categories will change based on locale, and then require a re-filtering.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 1187 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5MNgCZ-KXyRgcbayKR:-O5QQaAo2UQdigkvQDel:bhau392) ([raw file](https://github.com/sillsdev/web-xforge/blob/fb6afc2d90a59bd084383135838ff2566c60cd5d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L1187)):*

So, looking at this if block, wouldn't this include BT tabs whose `projectId` is populated? If it has a `projectId`, a BT tab will use the `projectService`, but if it doesn't, it will use the `activatedProject`. That seems to contradict the intent of your comment below. Maybe further comments are necessary? Or flip these two conditionals?

I see the tests you added for this, but we should probably cover this specific use case, as well.
</blockquote></details>

I have removed the code below you referenced, as I was misusing the `projectId` and `projectRef` properties of the tab.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.ts` line 25 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5MR-mW0nsi9sAIVDEA:-O5QHqtE0wx-Y6rTDu_h:b-8lnaaq) ([raw file](https://github.com/sillsdev/web-xforge/blob/99c77960faedc654dc189e52cb546339b1f5a6d3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.ts#L25)):*



The tab title is weird when you reload the page. Is that coming from here?
</blockquote></details>

Done. That was an unintended side effect of another change. I have fixed in.

josephmyers:
I don't have time to get through all this today, but I do believe the toolbar is worth persevering on. Can you experiment with replacing the built-in `sticky` setting for the header row?

pmachapman:
> I don't have time to get through all this today, but I do believe the toolbar is worth persevering on. Can you experiment with replacing the built-in sticky setting for the header row?

Alright. I think I've figured it out. I had to resort to `::ng-deep`:
image

josephmyers (changes requested):
This is looking better. Thanks for your perseverance!

> Reviewed 3 of 4 files at r8, 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-:-O5ktU12DEdip0jTwbLT:b37s4o7)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 250 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5kEjVP6Mc1vYYZTtr1:-O5kEjVP6Mc1vYYZTtr2:b-vqw5qz) ([raw file](https://github.com/sillsdev/web-xforge/blob/a66a059440a07f821ff0db7107f56724bf179132/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L250)):*
> ```XML
> 
>   set transliterateBiblicalTerms(value: boolean) {
>     this.projectUserConfigDoc?.submitJson0Op(op => op.set<boolean>(puc => puc.transliterateBiblicalTerms, value));
> ```

This is well-designed, because it links directly to the doc. Gets and sets access the source itself.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 312 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5kGk9j3posT2tNBmYu:-O5kGk9j3posT2tNBmYv:bqtu0c8) ([raw file](https://github.com/sillsdev/web-xforge/blob/a66a059440a07f821ff0db7107f56724bf179132/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L312)):*
> ```XML
>           if (!selectedCategoryLoaded) {
>             selectedCategoryLoaded = true;
>             this.selectedCategory = this.projectUserConfigDoc?.data?.selectedBiblicalTermsCategory ?? 'show_all';
> ```

I'm guessing we're preventing sets after the first specifically because doing so when the locale changes would cause problems? The other observables seem to be fine. But there's a bug in the db:


If you kept the property updated, you would fix this bug and wouldn't need to avoid this set. If you do this, you guarantee the property is never wrong:



_Code snippet:_
```XML
  get selectedCategory(): string {
    return this.projectUserConfigDoc?.data?.selectedBiblicalTermsCategory ?? 'show_all';
  }

  set selectedCategory(value: string) {
    if (this.selectedCategory !== value) {
      this.projectUserConfigDoc?.submitJson0Op(op => op.set(puc => puc.selectedBiblicalTermsCategory, value));
      this.filterBiblicalTerms(this._bookNum ?? 0, this._chapter ?? 0, this._verse);
    }
  }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 313 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5kshbrARYmo18p_gqa:-O5kshbs2McUjpgYIDh9:b-g85lrx) ([raw file](https://github.com/sillsdev/web-xforge/blob/a66a059440a07f821ff0db7107f56724bf179132/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L313)):*
> ```XML
>             selectedCategoryLoaded = true;
>             this.selectedCategory = this.projectUserConfigDoc?.data?.selectedBiblicalTermsCategory ?? 'show_all';
>           }
> ```

I recommend making the above change for `selectedRangeFilter`, as well, which would allow you to completely remove `onSelectionChanged()`

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671)* status: 30 of 33 files reviewed, 2 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 312 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5kGk9j3posT2tNBmYu:-O5o6KeE-IxIII1aPqCm:bb0vwzp) ([raw file](https://github.com/sillsdev/web-xforge/blob/a66a059440a07f821ff0db7107f56724bf179132/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L312)):*

I'm guessing we're preventing sets after the first specifically because doing so when the locale changes would cause problems? The other observables seem to be fine. But there's a bug in the db:


If you kept the property updated, you would fix this bug and wouldn't need to avoid this set. If you do this, you guarantee the property is never wrong:
</blockquote></details>

Done. There was a little bit of wrangling to stop the display glitch you documented earlier (the `categoriesLoading` variable is now at class level).
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 313 at r11](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5kshbrARYmo18p_gqa:-O5oGlfIAdwT3bSgMiOO:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/a66a059440a07f821ff0db7107f56724bf179132/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L313)):*

I recommend making the above change for `selectedRangeFilter`, as well, which would allow you to completely remove `onSelectionChanged()`
</blockquote></details>

Done.

josephmyers (changes requested):
Just one last question -- thank you!

> Reviewed 3 of 3 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-:-O5pKf6j7cRyCf85E8dh:b6e4fbz)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 255 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5pJS7E0uKibEufIbFR:-O5pJS7E0uKibEufIbFS:bh57g7d) ([raw file](https://github.com/sillsdev/web-xforge/blob/038622a74a456b8f6cc77ebf04704839e73da832/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L255)):*
> ```XML
>     if (this.selectedCategory !== value) {
>       // Store the default dropdown value of show_all as undefined in the database
>       if (value === '' || value === 'show_all') value = undefined;
> ```

Why aren't we sending in the "All" and "Current Verse" selections? Was the old `onSelectionChanged()` behaving this way? What's the reason for the change?

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 255 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5pJS7E0uKibEufIbFR:-O5pa8i59q1qyUJNE-Op:bqcvoj3) ([raw file](https://github.com/sillsdev/web-xforge/blob/038622a74a456b8f6cc77ebf04704839e73da832/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L255)):*

Why aren't we sending in the "All" and "Current Verse" selections? Was the old `onSelectionChanged()` behaving this way? What's the reason for the change?
</blockquote></details>

It isn't that we aren't sending it, instead we are clearing it as they are being set to the default values for the dropdown. It is done mostly for state consistency. If you open Biblical Terms for the first time, the two properties are `undefined`, and so correspond to the default values.

I can undo it if you really want, although I have a strong preference for the `show_all` category to be recorded as an `undefined` value in case we have to alter the value for it for some reason in the future.

josephmyers (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-:-O5qmw0K4RwZ3emyeOg-:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 255 at r12](https://reviewable.io/reviews/sillsdev/web-xforge/2671#-O5pJS7E0uKibEufIbFR:-O5qmqg0BwaezNZ8dixi:b-9uvnqt) ([raw file](https://github.com/sillsdev/web-xforge/blob/038622a74a456b8f6cc77ebf04704839e73da832/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L255)):*

It isn't that we aren't sending it, instead we are clearing it as they are being set to the default values for the dropdown. It is done mostly for state consistency. If you open Biblical Terms for the first time, the two properties are `undefined`, and so correspond to the default values.

I can undo it if you really want, although I have a strong preference for the `show_all` category to be recorded as an `undefined` value in case we have to alter the value for it for some reason in the future.
</blockquote></details>

Nope, your reasoning makes sense to me. Thanks

PR #2365, opened 2024-03-06
SF-2594 Allow saving comments with the bottomsheet with XML symbols

RaymondLuong3:
The content for a note needs to have any xml related symbols such as `>` and `<` encoded to be stored in mongo. This was previously done when editing or adding a note to a thread in the note dialog, but adding notes through the bottom sheet was missed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2365)

PR #2675, opened 2024-08-16
SF-2927 Fix incorrect escaping of XML characters in notes

Nateowami:
This error was caused by attempting to parse `1&2` as XML (when the `&` should have been escaped).

It is not clear to me exactly how the situation was created, but what I did observe is:
1. A Paratext user commented on a key term
2. That comment ended up having the user's name added to the comment *as thought the user was a reviewer, even though the user was not*. Perhaps joining/leaving the project, or changing names, somehow confused SF. I'm not sure.

The partially redacted XML looked like this:

``` XML
  <Comment Thread="BT_Δαυίδ" User="John Doe" VerseRef="MAT 1:1" Language="" Date="2024-08-05T04:00:02.6700000+00:00">
    <SelectedText />
    <StartPosition>0</StartPosition>
    <ContextBefore />
    <ContextAfter />
    <Status>todo</Status>
    <Type></Type>
    <ConflictType>unknownConflictType</ConflictType>
    <Verse />
    <ExtraHeadingInfo>Δαυίδ￼greek￼David David</ExtraHeadingInfo>
    <HideInTextWindow>false</HideInTextWindow>
    <Field Name="versionNumber">1</Field>
    <Contents>
      <p sf-user-label="true">[John Doe - Scripture Forge]</p>
      <p>there was some text here 1&amp;2 more text
</p>
    </Contents>
    <BiblicalTermId>Δαυίδ</BiblicalTermId>
    <TagAdded>-3</TagAdded>
  </Comment>
```

As you can see, the ampersand is escaped, as expected. However, the SF UI showed this error when trying to view the thread:

This is because in Mongo the ampersand was not escaped.

Testing showed that if you add a comment as a reviewer in SF and include an ampersand, it is stored in the XML escaped twice, and shows in the PT UI escaped once.

The rest of the explanation of the problem will be on the changed lines themselves.

I have not written tests, though I think they should be added.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2675)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.spec.ts
@@ -38,6 +38,7 @@ import { SFProjectProfileDoc } from '../../core/models/sf-project-profile-doc';
 import { SFProjectUserConfigDoc } from '../../core/models/sf-project-user-config-doc';
 import { SF_TYPE_REGISTRY } from '../../core/models/sf-type-registry';
 import { SFProjectService } from '../../core/sf-project.service';
+import { XmlUtils } from '../../shared/utils';
 import { MockNoteDialogRef } from '../editor/editor.component.spec';
 import { NoteDialogComponent } from '../editor/note-dialog/note-dialog.component';
 import { BiblicalTermDialogIcon, BiblicalTermNoteIcon, BiblicalTermsComponent } from './biblical-terms.component';
@@ -215,7 +216,7 @@ describe('BiblicalTermsComponent', () => {
     env.biblicalTermsNotesButton.click();
     env.wait();
 
-    const noteContent: string = 'content in the thread';
+    const noteContent: string = 'content in the thread & with an ampersand';
     env.mockNoteDialogRef.close({ noteContent });
     env.wait();
 
@@ -232,7 +233,7 @@ describe('BiblicalTermsComponent', () => {
     expect(noteThread.originalSelectedText).toEqual('');
     expect(noteThread.publishedToSF).toBe(true);
     expect(noteThread.notes[0].ownerRef).toEqual('user01');
-    expect(noteThread.notes[0].content).toEqual(noteContent);
+    expect(noteThread.notes[0].content).toEqual(XmlUtils.encodeForXml(noteContent));
     expect(noteThread.notes[0].tagId).toEqual(BIBLICAL_TERM_TAG_ID);
     const projectUserConfigDoc: SFProjectUserConfigDoc = env.getProjectUserConfigDoc('project01', 'user01');
     expect(projectUserConfigDoc.data!.noteRefsRead).not.toContain(noteThread.notes[0].dataId);
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts
@@ -12,7 +12,7 @@ import {
   NoteThread,
   NoteType
 } from 'realtime-server/lib/esm/scriptureforge/models/note-thread';
-import { SFProjectDomain, SF_PROJECT_RIGHTS } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-rights';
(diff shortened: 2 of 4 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2675)* status: 0 of 1 files reviewed, 2 unresolved discussions

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2914 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2675#-O4Rc9gnAcaLfJNcZwNi:-O4Rc9gnAcaLfJNcZwNj:b-r9q9q9) ([raw file](https://github.com/sillsdev/web-xforge/blob/b16fbc42c9a546b596927e72d8acc579a52944e7/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2914)):*
> ```Smalltalk
>         {
>             // add the note content in a paragraph tag
>             XElement commentNode = new XElement("p", note.Content);
> ```

The `note` in this method is taken from our database. `note.Content` is already XML, but it's being passed to this constructor as something it should take as a literal value, rather than something it should parse to create an XML document.

The `else` clause below parses `note.Content` as XML. The `if` clause has just been updated to treat it the same.
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2975 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2675#-O4Rcw0g7YIYjOHAiyjF:-O4Rcw0g7YIYjOHAiyjG:bu1sck6) ([raw file](https://github.com/sillsdev/web-xforge/blob/8beb285e4b0d6b91287120a2ea87c3afe9f854d0/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2975)):*
> ```Smalltalk
>             // If there is only one paragraph node other than the SF user label then omit the paragraph tags
>             if (isReviewer && paragraphNodeCount <= 2)
>                 sb.Append(element.GetInnerXml());
> ```

In the context of this bug, the user was not a reviewer, but for some reason the `sf-user-label` was added to the note, and therefore `isReviewer` is true in this method.

The `else` clause appends the node to the `StringBuilder`, which I *think* will later call `node.ToString()`, which I *think* will return the node as XML, not the inner content of the node.

However, the `if` clause is getting `element.Value`, which gets the text contents, not an XML representation of the element.

I'm not sure if `GetInnerXml()` from `SIL.Xml` is the right thing to use here. It seems to be the only place we use this method.

pmachapman (commented):
In my multiple attempts to recreate this, manually and via manipulating variables in memory in sync to match the conditions, I was unable to confirm that PR fixes the bug. I think this is because the line `XDocument commentDoc = XDocument.Parse($"<root>{note.Content}</root>");` does not allow the corrupt string, e.g. "See 1&2 John.", to be set

What I have found out is that the bug was introduced by #2365.

I have concerns that because of the logic introduced in SF-2594, this PR will cause other issues.

I think the best solution is to expand SF-2594 to work with Biblical Terms, and to manually fix the record for this project (I have created a branch - https://github.com/sillsdev/web-xforge/compare/fix/SF-2927 - to do this).

@nateowami What do you think?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2675#-:-O4b7qAI9oQ2DU1xE-iF:bs1n3a3)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

pmachapman:
@Nateowami I have added my fix for the incoming data from Biblical Terms, fixed a bug with the C# implementation found while writing the unit tests, and have updated the unit tests to throw when the entities are double encoded or de-encoded. The unit tests I changed will fail on master, if whoever reviews it wishes to check out the bug in action.

@RaymondLuong3 or @Nateowami if you want to review my changes?

Nateowami (commented):
:lgtm:

> Reviewed 3 of 4 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2675#-:-O4qyw8TEWtHiSuSh43k:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

pmachapman (approved):
:lgtm_strong: 

> Reviewed 2 of 4 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2675#-:-O4qz1pWDGqyBPksb5xH:b-uwkowd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2621, opened 2024-07-26
SF-2860 Users notified when offline on My Projects page

kylebuss:
My Projects page will not download all the necessary data for projects to be accessed offline (texts, notes, questions).
"Offline Access" will display on the Projects card once data is accessible in the IndexedDB.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2621)

PR #2677, opened 2024-08-20
SF-2919 Tweak UI for My Projects cards

josephmyers:
- removed the buttons for connected items
- made connected items' cards clickable
- narrowed the cards
- removed the card containers from unconnected items (including the loading cards) and gave them a bottom line

image
and
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2677)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.module.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.module.ts
@@ -19,6 +19,7 @@ import { InAppRootOverlayContainer } from 'xforge-common/overlay-container';
 import { SupportedBrowsersDialogComponent } from 'xforge-common/supported-browsers-dialog/supported-browsers-dialog.component';
 import { UICommonModule } from 'xforge-common/ui-common.module';
 import { XForgeCommonModule } from 'xforge-common/xforge-common.module';
+import { MatRipple } from '@angular/material/core';
 import { environment } from '../environments/environment';
 import { AppRoutingModule } from './app-routing.module';
 import { AppComponent } from './app.component';
@@ -73,7 +74,8 @@ import { UsersModule } from './users/users.module';
     TranslocoModule,
     AppRoutingModule,
     SharedModule,
-    AvatarComponent
+    AvatarComponent,
+    MatRipple
   ],
   providers: [
     CookieService,
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html
@@ -1,149 +1,156 @@
 <ng-container *transloco="let t; read: 'my_projects'">
-  <h1>{{ t("my_projects") }}</h1>
-  @if (!userHasProjects && !initialLoadingSFProjects && !loadingPTProjects) {
-    <app-notice id="message-no-pt-or-sf-projects" icon="warning" type="warning">
-      {{ t("no_projects") }} {{ t("access_another_project") }}
-    </app-notice>
-  }
-  @if (userConnectedProjects.length > 0 || initialLoadingSFProjects) {
-    <h2 id="header-connected-projects">
-      {{ t("connected") }}
-    </h2>
-  }
-  @for (projectDoc of userConnectedProjects; track projectDoc) {
-    <mat-card
-      [class.active-project]="isLastSelectedProject(projectDoc)"
-      id="user-connected-project-card-{{ projectDoc.data?.paratextId }}"
-      attr.data-pt-project-id="{{ projectDoc.data?.paratextId }}"
-      data-project-type="user-connected-project"
-    >
-      <span class="project-name">
-        <b>{{ projectDoc.data?.shortName }}</b> - {{ projectDoc.data?.name }}
-      </span>
-      <span class="project-description">
-        {{ projectTypeDescription(projectDoc) }}
-      </span>
-      <a appRouterLink="/projects/{{ projectDoc.id }}" mat-flat-button color="primary">{{ t("open") }}</a>
-    </mat-card>
-  }
-  @if (initialLoadingSFProjects) {
-    <mat-card id="sf-loading-card" class="loading-card">
-      <div class="loading-project-name"></div>
-      <div class="loading-project-description"></div>
-      <div class="loading-action"></div>
-    </mat-card>
-    <mat-card class="loading-card">
-      <div class="loading-project-name"></div>
-      <div class="loading-project-description"></div>
-      <div class="loading-action"></div>
-    </mat-card>
-  }
-  @if (userHasProjects && !userIsPTUser) {
-    <mat-expansion-panel id="message-looking-for-another-project" hideToggle>
-      <mat-expansion-panel-header>{{ t("looking_for_another_project") }}</mat-expansion-panel-header>
-      <p>
-        {{ t("access_another_project") }}
-      </p>
-    </mat-expansion-panel>
-  }
+  <div class="content">
+    <h1>{{ t("my_projects") }}</h1>
+    @if (!userHasProjects && !initialLoadingSFProjects && !loadingPTProjects) {
+      <app-notice id="message-no-pt-or-sf-projects" icon="warning" type="warning">
+        {{ t("no_projects") }} {{ t("access_another_project") }}
+      </app-notice>
+    }
+    @if (userConnectedProjects.length > 0 || initialLoadingSFProjects) {
+      <h2 id="header-connected-projects">
+        {{ t("connected") }}
+      </h2>
+    }
+    @for (projectDoc of userConnectedProjects; track projectDoc) {
+      <mat-card
+        [class.active-project]="isLastSelectedProject(projectDoc)"
+        id="user-connected-project-card-{{ projectDoc.data?.paratextId }}"
+        attr.data-pt-project-id="{{ projectDoc.data?.paratextId }}"
+        data-project-type="user-connected-project"
+        appRouterLink="/projects/{{ projectDoc.id }}"
+      >
+        <span class="project-name">
+          <b>{{ projectDoc.data?.shortName }}</b> - {{ projectDoc.data?.name }}
+        </span>
+        <span class="project-description">
+          {{ projectTypeDescription(projectDoc) }}
+        </span>
+      </mat-card>
+    }
+    @if (initialLoadingSFProjects) {
+      <mat-card id="sf-loading-card" class="loading-card">
+        <div class="loading-project-name"></div>
+        <div class="loading-project-description"></div>
+        <div class="loading-action"></div>
+      </mat-card>
+      <mat-card class="loading-card">
+        <div class="loading-project-name"></div>
+        <div class="loading-project-description"></div>
+        <div class="loading-action"></div>
+      </mat-card>
+    }
+    @if (userHasProjects && !userIsPTUser) {
+      <mat-expansion-panel id="message-looking-for-another-project" hideToggle>
+        <mat-expansion-panel-header>{{ t("looking_for_another_project") }}</mat-expansion-panel-header>
+        <p>
+          {{ t("access_another_project") }}
+        </p>
+      </mat-expansion-panel>
+    }
 
-  @if (userConnectedResources.length > 0) {
-    <h2>{{ t("dbl_resources") }}</h2>
-  }
-  @for (projectDoc of userConnectedResources; track projectDoc) {
-    <mat-card
-      [class.active-project]="isLastSelectedProject(projectDoc)"
-      attr.data-pt-project-id="{{ projectDoc.data?.paratextId }}"
-      data-project-type="user-connected-resource"
-    >
-      <span class="project-name">
-        <b>{{ projectDoc.data?.shortName }}</b> - {{ projectDoc.data?.name }}
-      </span>
-      <span class="project-description">{{ t("dbl_resource") }}</span>
-      <a appRouterLink="/projects/{{ projectDoc.id }}" mat-flat-button color="primary">{{ t("open") }}</a>
-    </mat-card>
-  }
+    @if (userConnectedResources.length > 0) {
+      <h2>{{ t("dbl_resources") }}</h2>
+    }
+    @for (projectDoc of userConnectedResources; track projectDoc) {
+      <mat-card
+        [class.active-project]="isLastSelectedProject(projectDoc)"
+        attr.data-pt-project-id="{{ projectDoc.data?.paratextId }}"
+        data-project-type="user-connected-resource"
+        appRouterLink="/projects/{{ projectDoc.id }}"
+      >
+        <span class="project-name">
+          <b>{{ projectDoc.data?.shortName }}</b> - {{ projectDoc.data?.name }}
+        </span>
+        <span class="project-description">{{ t("dbl_resource") }}</span>
+      </mat-card>
+    }
 
-  @if (userIsPTUser) {
-    <div class="pt-content">
-      @if (
-        userUnconnectedParatextProjects.length > 0 ||
-        loadingPTProjects ||
-        problemGettingPTProjects ||
-        (isOnline | async) === false
-      ) {
-        <h2 id="header-not-connected-projects">
-          {{
-            loadingPTProjects
-              ? t("loading_more_pt_projects")
-              : userConnectedProjects.length === 0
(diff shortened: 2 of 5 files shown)


Nateowami:
Thanks for your work on this. There are a number of things that I don't like with this layout:
1. The width of the items, being the only thing on the page, feels odd. I think being wider was better.
2. A card is not generally something you click, and I don't think we should give that association to users. I think it would make more sense to use a card for the non-connected items, rather than the connected items, since they contain a clickable item, rather than being a clickable item.
3. There's no hover effect, so the cards don't look clickable.
4. I don't really like the divider line as a way of separating the non-connected items.

josephmyers:
As I process and work through these items, I'm wondering if @nigel-wells would like to contribute, as all of these comments other than #3 seem to be directed at the design (though the implementation may be able to be improved)

josephmyers:
> A card is not generally something you click, and I don't think we should give that association to users. I think it would make more sense to use a card for the non-connected items, rather than the connected items, since they contain a clickable item, rather than being a clickable item.
There's no hover effect, so the cards don't look clickable.

I've tried to address these two. Does it look/work any better? It sounds like with 1, 2, and 4 you have some fundamental problems with the design in Balsamiq. Unless that's not the case, maybe we should reconvene at the UX meeting next week to address them?

josephmyers:
> The width of the items, being the only thing on the page, feels odd. I think being wider was better.

There were a number of things I was keeping in mind when making and viewing this change, which justified it to me. First, we're already predisposed to the larger style, since we've been seeing it so regularly. I doubt many users are carrying around a strong bias against change. Second, the emptiness of the screen will be reduced as we add other features, e.g. filtering, to this page, so we should expect things to look somewhat so. Third, reducing the size of the cards was in my impression fundamental to redesigning them to look more like buttons. The current "board" shape doesn't look like a button, and itself looks quite empty without the button on the right.

All that said, we could toy with the layout some more. It's not like the "board" style is the only thing that works. For example, what if we allowed, say, three items per row?

> I think it would make more sense to use a card for the non-connected items, rather than the connected items, since they contain a clickable item, rather than being a clickable item.

I am honestly having a hard time understanding what you're proposing. Can you explain a little more?

nigel-wells:
Yes, it may pay to have another discussion around taking the wireframe to the design.
This is probably closer to what I had in mind which also includes an option to not have divider lines.
I increased the width a little bit more but it does help us keep more true to mobile first.
I also adjusted some of the other padding and margin around the place.
I pushed through my changes to `improvement/SF-2919b` - it was very quick and rough and I'm sure the CSS and div tags I was adding at will could be done more efficient.
image

Nateowami:
I've pushed an example of a concept I like to `feature/project-page-example`

Here's a screenshot:



It doesn't look balanced because there's nothing on the right side, but we plan to change that soon.

kylebuss:
Just a heads up, there are changes from SF-2860 (PR https://github.com/sillsdev/web-xforge/pull/2621) that will cause merge conflicts with these changes. On the card itself there will be additional text added on the Connected projects cards but the width looks like it should be fine.

josephmyers:
@nigel-wells I like what you did with the unconnected projects. I like how it subtly communicates a difference between them and the connected items.

I'd be fine pulling these changes in, but I don't believe that my version is different enough to be unacceptable. I actually believe having the lines balances out and brings valuable association to the titles and buttons. With your screenshot, it's a _little_ harder to visually trace over to the corresponding button. The line helps with that, though there are doubtless other design options. It may be that the increased/restored row width contributes to the difficulty. But I think we could tweak this and get it looking quite nice.

nigel-wells:
Where did we get to with this? I'm happy to review but I'm not sure we're all agreed on the final design yet are we?

josephmyers:
We didn't discuss it (at the UX meeting) without you. The latest is my last comment here. I can absolutely pull in your changes, but it looks too spaced out without any sort of grouping for each row. We don't have to solve it with a line, but we do need to do something about it

nigel-wells (changes requested):
Just noted down the points we discussed yesterday.
It would be great if this last one could be addressed where they background gray would extend to the bottom of the window rather than it leaving a large white gap like it currently does. 


> Reviewed 1 of 3 files at r3, 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-:-O5uhvU732wLh_fe66Be:b-7p2hia)* status: all files reviewed, 3 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 22 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-O5uh7pk6kVA9FtramVw:-O5uh7pk6kVA9FtramVx:b-nzmpdm) ([raw file](https://github.com/sillsdev/web-xforge/blob/05c7c6419f807e16ae053fafbf1a604fc5186ec5/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L22)):*
> ```SCSS
> }
> 
> .pt-content-container {
> ```

Needs another style specific to the `h2` to make the bottom margin appear similar to the headings above the cards to replicate the flex gap used there.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 25 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-O5uhc4L1u48V84MpAjE:-O5uhc4L1u48V84MpAjF:b-jcy5a7) ([raw file](https://github.com/sillsdev/web-xforge/blob/05c7c6419f807e16ae053fafbf1a604fc5186ec5/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L25)):*
> ```SCSS
>   background: #f4f4f4;
>   margin-top: 30px;
>   padding: 20px 0;
> ```

Change this to `padding: 20px;` to compensate for the change in margin mentioned below.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 30 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-O5ugubB9yzUwg9OzeLN:-O5ugubB9yzUwg9OzeLO:b-llh8mr) ([raw file](https://github.com/sillsdev/web-xforge/blob/05c7c6419f807e16ae053fafbf1a604fc5186ec5/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L30)):*
> ```SCSS
>     flex-direction: column;
> 
>     margin: 0 -20px;
> ```

This needs to be moved to `.pt-content-container` and merged in with `margin-top` i.e. `margin: 30px -20px 0;`

josephmyers (commented):
Ok, thanks. This should be done.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-:-O5zj5lS8VZO1FcH2w37:b-c9zes6)* status: all files reviewed, 3 unresolved discussions (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 22 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-O5uh7pk6kVA9FtramVw:-O5zlhOhCTBMDXRN88Bj:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/05c7c6419f807e16ae053fafbf1a604fc5186ec5/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L22)):*

Needs another style specific to the `h2` to make the bottom margin appear similar to the headings above the cards to replicate the flex gap used there.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 25 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-O5uhc4L1u48V84MpAjE:-O5zl2xc5m5qRUdCL0aC:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/05c7c6419f807e16ae053fafbf1a604fc5186ec5/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L25)):*

Change this to `padding: 20px;` to compensate for the change in margin mentioned below.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 30 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-O5ugubB9yzUwg9OzeLN:-O5zjWDl110xbqSlJqKu:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/05c7c6419f807e16ae053fafbf1a604fc5186ec5/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L30)):*

This needs to be moved to `.pt-content-container` and merged in with `margin-top` i.e. `margin: 30px -20px 0;`
</blockquote></details>

Done.

nigel-wells (changes requested):
Can both the connected and non-connected projects be the same width - feels a bit strange that the top layout is narrower than bottom?


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-:-O6-5J5nEENuGjo5dqhs:b-av714z)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 25 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-O5uhc4L1u48V84MpAjE:-O6-4gXD2q6_d5cvHc-d:bn9dl4m) ([raw file](https://github.com/sillsdev/web-xforge/blob/05c7c6419f807e16ae053fafbf1a604fc5186ec5/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L25)):*

Done.
</blockquote></details>

Looks like you missed this one as the left/right padding is still zero. It looks fine on Desktop but smaller devices looks like this:

Nateowami:
Here's another proposed style: `feature/my-projects-nate-proposal-2`

I still think we should not be making something look like a card but act like a button.

josephmyers (commented):
Good find, Nigel. This is a smaller use case, as it will only appear when you don't have an item that fills up the full width of the cards.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-:-O6JW8dzDPjmiAkNJN-H:b-a1rz3u)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 25 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-O5uhc4L1u48V84MpAjE:-O6J_zNk-dyrpFm6Eyl3:b-7iqvca) ([raw file](https://github.com/sillsdev/web-xforge/blob/05c7c6419f807e16ae053fafbf1a604fc5186ec5/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L25)):*

Looks like you missed this one as the left/right padding is still zero. It looks fine on Desktop but smaller devices looks like this:

</blockquote></details>

Thanks. I think I changed it the dev tools :D

josephmyers:
> Here's another proposed style: `feature/my-projects-nate-proposal-2`
> 
> I still think we should not be making something look like a card but act like a button.

I think this branch looks okay, other than one fairly glaring issue: the top items now look exactly like the bottom items (less the button), but they function fundamentally differently. So although these rows look exactly the same, one of them is interactive and one is not. This isn't good implicit communication on our part, and users will get confused. The top items, once they've seen a "Connect" button, look disabled. The top section needs to look different enough that users are drawn to those projects first and then that they understand those are different from the ones below.

image

Maybe you have thoughts on how future features on this page could alleviate this? But we'd want to wait until we implemented those.

nigel-wells:
I agree Joseph - they look too much like the bottom section which isn't clickable
I've chatted over some ideas with our design and she's come up with a middle ground for us to consider.
Suggestions:
* Keep the cards
* Reduce the heavy drop shadow in favor of the softer default one
* Add a background color to each card to help enforce it is clickable - as a side note, when she first saw the cards her first instinct was that it was already clickable
* Don't add alternating background colors to the cards as it starts to bring into doubt to what you're looking at i.e. is this a list, table, something else?

Nathaniel just a note regarding cards, it is in Material's own guidelines that cards themselves can be actionable i.e. clickable, draggable, swipeable (is that a word?). I've also used many apps where cards are clickable and it feels very natural.

image

nigel-wells (approved):
Happy with this subject to us agreeing on the final design :)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-:-O6JpWIp3u3JaXuvRZg0:b-dz5v25)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers:
Nigel, glad to get some outside feedback on the matter. I'm a little unclear about your approval -- are you wanting to implement the suggestions now or later or not at all?

nigel-wells:
More just awaiting feedback from @Nateowami before deciding if we want any further changes or not.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2677)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 71 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-O6RNnwR2zZFi7gt18Jg:-O6RNnwR2zZFi7gt18Jh:bqz7kci) ([raw file](https://github.com/sillsdev/web-xforge/blob/b882977debe3a3987ba2e531019f0d6bacdcca29/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L71)):*
> ```SCSS
>   grid-template-areas:
>     'project-name project-name'
>     'description button';
> ```

If you're removing the  buttons, you should be updating the CSS that refers to the buttons. Here and elsewhere. I suggest leaving it as a CSS grid, since we plan to add more to the cards, and it will probably make future changes easier.

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2677)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @Nateowami and @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 71 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-O6RNnwR2zZFi7gt18Jg:-O6Yl0NV-U3win8q4yUl:b-68nccp) ([raw file](https://github.com/sillsdev/web-xforge/blob/b882977debe3a3987ba2e531019f0d6bacdcca29/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L71)):*

If you're removing the  buttons, you should be updating the CSS that refers to the buttons. Here and elsewhere. I suggest leaving it as a CSS grid, since we plan to add more to the cards, and it will probably make future changes easier.
</blockquote></details>

Yes, thanks, I've kept the grid

nigel-wells (commented):
Your changes are fine but I did notice that after your rebase (which reviewable insisted I reviewed) that we now have two noticed about being online which ultimately stack against each other. Looks like they came in via SF-2934 earlier today. Doesn't need to hold this PR up from going to testing unless you wanted to quickly resolve it


> Reviewed 2 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-:-O6ZYaXlCnGxC_zsLuJ_:b-94lig8)* status: 4 of 5 files reviewed, all discussions resolved (waiting on @josephmyers)

nigel-wells (approved):
> Reviewed 1 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2677)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (commented):
Thanks! That's a rebase error. I thought I was preserving the changes to master.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2677#-:-O6_2a_lEJbPnAqp5i1l:b-eg3uv3)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

PR #2680, opened 2024-08-21
Create improved book toggling component

Nateowami:
This is a proposed component for toggling a book when selecting books (an alternative to mat chips).

I'm opening it as a draft PR so Storybook will get published to Chromatic.

I'm proposing a design, and I think this branch can serve as a basis if others want to propose a different design, by using this as a starting point.

For option A (what this PR originally had), see this storybook:
https://644bffe327c540a62f0fd260-wvwluxllwf.chromatic.com/?path=/story/translate-togglebook--selected

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2680)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.html
@@ -32,17 +32,12 @@
   }
   <div class="flex-row">
     @for (book of bookOptions; track book) {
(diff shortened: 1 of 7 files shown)


Nateowami:
See `feature/book-toggle-no-border` for another option:

kylebuss:
My vote would be the book toggle with no border. I really like the info display of book completion percentage when hovering over the book!

josephmyers:
@Nateowami, do you want to push forward with this at all? Would be a good candidate for our UX meetings.

PR #2684, opened 2024-08-23
SF-2844 Notify user resource failed to load due to being offline

kylebuss:
When a user attempts to load a resource but losses internet connection they could receive a variety of Error Dialog popups as a result.

One minor change has to do with background loading the resource if it already exists in the IndexedDB. The dialog will stay open can close once the resource is sync.

Added additional error messages to be displayed in `editor-tab-add-resource-dialog`:
- User will be notified if loading resource failed due to being offline.
- User will be notified if loading project failed due to not being authorized.

Added offline checks to `realtime-service`, this may not be what we want in the long run. But this is an issue when we attempt to subscribe to anything with the realtime server that has not been loaded for offline access (Line 264 in the `realtime-doc.ts` appears to be the culprit). This also seems to be one of the reasons why we see the infinite loading bar at the top of the screen when offline.

Updated `exception-handling-service` to not show errors when user is offline. I'm not sure this how we want to handle this but in testing this issue I would sometimes see 2-3 error dialogs appear for both server http and angular errors. These will still be logged to the console but as the user is offline they would not be reported to bugsnag.

EDIT: Looks like I forgot to run all tests and have a large amount failing due adding the OnlineStatusService in the `exception-handling-service`. Converted this to a draft for feedback on that portion of the pull request.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2684)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/project-notification.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/project-notification.service.ts
@@ -1,6 +1,7 @@
 import { Injectable } from '@angular/core';
 import { AbortError, HubConnection, HubConnectionBuilder, IHttpConnectionOptions } from '@microsoft/signalr';
 import { AuthService } from 'xforge-common/auth.service';
+import { OnlineStatusService } from 'xforge-common/online-status.service';
 
 @Injectable({
   providedIn: 'root'
@@ -11,8 +12,15 @@ export class ProjectNotificationService {
     accessTokenFactory: async () => (await this.authService.getAccessToken()) ?? ''
   };
 
-  constructor(private authService: AuthService) {
+  constructor(
+    private authService: AuthService,
+    private readonly onlineService: OnlineStatusService
+  ) {
     this.connection = new HubConnectionBuilder().withUrl('/project-notifications', this.options).build();
+  }
+
+  get appOnline(): boolean {
+    return this.onlineService.isOnline && this.onlineService.isBrowserOnline;
   }
 
   setNotifySyncProgressHandler(handler: any): void {
@@ -22,8 +30,9 @@ export class ProjectNotificationService {
   async start(): Promise<void> {
     await this.connection.start().catch(err => {
       // Suppress AbortErrors, as they are not caused by server error, but the SignalR connection state
-      // These will be thrown if a user navigates away quickly after starting the sync
-      if (err instanceof AbortError) {
+      // These will be thrown if a user navigates away quickly after
+      // starting the sync or the app loses internet connection
+      if (err instanceof AbortError || !this.appOnline) {
         return;
       } else {
         throw err;
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.spec.ts
@@ -57,6 +57,7 @@ describe('SyncProgressComponent', () => {
 
   it('does not initialize if app is offline', fakeAsync(async () => {
     const env = new TestEnvironment({ userId: 'user01' });
+    env.setupProjectDoc();
     env.onlineStatus = false;
     verify(mockedProjectService.get('sourceProject02')).never();
     expect(await env.getMode()).toBe('indeterminate');
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.spec.ts
@@ -1,6 +1,7 @@
 import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { Component, ViewChild } from '@angular/core';
 import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { provideAnimations } from '@angular/platform-browser/animations';
 import { SFProject } from 'realtime-server/lib/esm/scriptureforge/models/sf-project';
 import { SFProjectRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-role';
 import { createTestProject } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-test-data';
@@ -10,6 +11,9 @@ import { ErrorReportingService } from 'xforge-common/error-reporting.service';
 import { NoticeService } from 'xforge-common/notice.service';
 import { TestRealtimeModule } from 'xforge-common/test-realtime.module';
 import { TestRealtimeService } from 'xforge-common/test-realtime.service';
+import { OnlineStatusService } from 'xforge-common/online-status.service';
+import { TestOnlineStatusModule } from 'xforge-common/test-online-status.module';
+import { TestOnlineStatusService } from 'xforge-common/test-online-status.service';
 import { configureTestingModule, TestTranslocoModule } from 'xforge-common/test-utils';
 import { UICommonModule } from 'xforge-common/ui-common.module';
 import { SFProjectDoc } from '../../core/models/sf-project-doc';
@@ -28,22 +32,32 @@ describe('SyncProgressComponent', () => {
   configureTestingModule(() => ({
     declarations: [HostComponent, SyncProgressComponent],
     imports: [
+      TestOnlineStatusModule.forRoot(),
       UICommonModule,
       TestTranslocoModule,
       TestRealtimeModule.forRoot(SF_TYPE_REGISTRY),
       HttpClientTestingModule
     ],
     providers: [
+      provideAnimations(),
       { provide: NoticeService, useMock: mockedNoticeService },
       { provide: ProjectNotificationService, useMock: mockedProjectNotificationService },
       { provide: SFProjectService, useMock: mockedProjectService },
-      { provide: ErrorReportingService, useMock: mockedErrorReportingService }
+      { provide: ErrorReportingService, useMock: mockedErrorReportingService },
+      { provide: OnlineStatusService, useClass: TestOnlineStatusService }
     ]
   }));
 
   it('does not initialize if projectDoc is undefined', fakeAsync(async () => {
     const env = new TestEnvironment({ userId: 'user01' });
     expect(env.host.projectDoc).toBeUndefined();
+    verify(mockedProjectService.get('sourceProject02')).never();
+    expect(await env.getMode()).toBe('indeterminate');
+  }));
+
+  it('does not initialize if app is offline', fakeAsync(async () => {
+    const env = new TestEnvironment({ userId: 'user01' });
+    env.onlineStatus = false;
     verify(mockedProjectService.get('sourceProject02')).never();
     expect(await env.getMode()).toBe('indeterminate');
   }));
@@ -144,6 +158,9 @@ interface TestEnvArgs {
 }
 
 class TestEnvironment {
+  readonly testOnlineStatusService: TestOnlineStatusService = TestBed.inject(
+    OnlineStatusService
+  ) as TestOnlineStatusService;
   readonly fixture: ComponentFixture<HostComponent>;
   readonly host: HostComponent;
 
@@ -209,6 +226,12 @@ class TestEnvironment {
 
   get progressBar(): HTMLElement | null {
     return this.fixture.nativeElement.querySelector('mat-progress-bar');
+  }
+
+  set onlineStatus(isOnline: boolean) {
+    this.testOnlineStatusService.setIsOnline(isOnline);
+    tick();
+    this.fixture.detectChanges();
   }
 
   updateSyncProgress(percentCompleted: number, projectId: string): void {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts
@@ -5,6 +5,7 @@ import { isParatextRole } from 'realtime-server/lib/esm/scriptureforge/models/sf
 import { BehaviorSubject, Observable, map, merge } from 'rxjs';
 import { ErrorReportingService } from 'xforge-common/error-reporting.service';
 import { FeatureFlagService } from 'xforge-common/feature-flags/feature-flag.service';
+import { OnlineStatusService } from 'xforge-common/online-status.service';
 import { SubscriptionDisposable } from 'xforge-common/subscription-disposable';
 import { SFProjectDoc } from '../../core/models/sf-project-doc';
 import { ProjectNotificationService } from '../../core/project-notification.service';
@@ -46,13 +47,18 @@ export class SyncProgressComponent extends SubscriptionDisposable {
     private readonly projectService: SFProjectService,
     private readonly projectNotificationService: ProjectNotificationService,
     private readonly featureFlags: FeatureFlagService,
-    private readonly errorReportingService: ErrorReportingService
+    private readonly errorReportingService: ErrorReportingService,
+    private readonly onlineStatus: OnlineStatusService
   ) {
     super();
 
     this.projectNotificationService.setNotifySyncProgressHandler((projectId: string, progressState: ProgressState) => {
       this.updateProgressState(projectId, progressState);
     });
+  }
+
+  get appOnline(): boolean {
+    return this.onlineStatus.isOnline && this.onlineStatus.isBrowserOnline;
   }
 
   @Input() set projectDoc(doc: SFProjectDoc | undefined) {
@@ -64,7 +70,7 @@ export class SyncProgressComponent extends SubscriptionDisposable {
   }
 
   async initialize(): Promise<void> {
-    if (this._projectDoc?.data == null) {
+    if (this._projectDoc?.data == null || !this.appOnline) {
       return;
     }
     await this.projectNotificationService.start();
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html
@@ -31,20 +31,26 @@
     }
 
     @if (projectLoadingFailed && resourceLoadingFailed) {
-      <mat-error>
+      <mat-error id="project-resources-error">
         {{ t("error_fetching_projects_resources") }}
       </mat-error>
     } @else if (projectLoadingFailed && !resourceLoadingFailed) {
-      <mat-error>
+      <mat-error id="project-error">
         {{ t("error_fetching_projects") }}
       </mat-error>
     } @else if (resourceLoadingFailed && !projectLoadingFailed) {
-      <mat-error>
+      <mat-error id="resource-error">
         {{ t("error_fetching_resources") }}
       </mat-error>
     }
+    @if (!appOnline) {
+      <mat-error id="offline-error">
+        {{ t("error_offline") }}
+      </mat-error>
+    }
+
     @if (projectFetchFailed || syncFailed) {
-      <mat-error>
+      <mat-error id="fetch-sync-error">
         {{ t("error_loading_resource") }}
       </mat-error>
     }
@@ -55,7 +61,7 @@
     <button
       mat-flat-button
       color="primary"
-      [disabled]="!form.valid || isLoading || isSyncActive || projectFetchFailed || syncFailed"
+      [disabled]="!form.valid || isLoading || isSyncActive || !appOnline"
       (click)="confirmSelection()"
     >
       {{ t("select") }}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.spec.ts
@@ -1,5 +1,7 @@
 import { ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
+import { provideAnimations } from '@angular/platform-browser/animations';
+
 import { SFProject } from 'realtime-server/lib/esm/scriptureforge/models/sf-project';
 import { createTestProject } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-test-data';
 import { anything, mock, verify, when } from 'ts-mockito';
@@ -23,11 +25,12 @@ describe('EditorTabAddResourceDialogComponent', () => {
   configureTestingModule(() => ({
     imports: [TestOnlineStatusModule.forRoot(), TestTranslocoModule],
     providers: [
+      provideAnimations(),
       { provide: SFProjectService, useMock: mockSFProjectService },
       { provide: ParatextService, useMock: mockParatextService },
       { provide: PermissionsService, useMock: mockPermissionsService },
-      { provide: OnlineStatusService, useClass: TestOnlineStatusService },
       { provide: MatDialogRef, useMock: mockMatDialogRef },
+      { provide: OnlineStatusService, useClass: TestOnlineStatusService },
       { provide: MAT_DIALOG_DATA, useValue: {} }
     ]
   }));
@@ -201,9 +204,24 @@ describe('EditorTabAddResourceDialogComponent', () => {
     expect(env.component.projectFetchFailed).toBe(false);
     expect(env.component.syncFailed).toBe(false);
   });
+
+  it('should return without calling sync if offline and show that sync failed', fakeAsync(() => {
+    const env = new TestEnvironment();
+    env.onlineStatus = false;
+    env.component.confirmSelection();
+    verify(mockSFProjectService.onlineAddCurrentUser(env.projectId)).never();
+    verify(mockSFProjectService.onlineCreateResourceProject(env.paratextId)).never();
+    verify(mockPermissionsService.canSync(anything())).never();
+    verify(mockSFProjectService.onlineSync(anything())).never();
+    verify(mockMatDialogRef.close(anything())).never();
+    expect(env.component.syncFailed).toBe(true);
+  }));
 });
 
 class TestEnvironment {
+  readonly testOnlineStatusService: TestOnlineStatusService = TestBed.inject(
+    OnlineStatusService
+  ) as TestOnlineStatusService;
   component: EditorTabAddResourceDialogComponent;
   fixture: ComponentFixture<EditorTabAddResourceDialogComponent>;
 
@@ -229,6 +247,12 @@ class TestEnvironment {
     this.component = this.fixture.componentInstance;
 
     this.component.form.setValue({ sourceParatextId: this.paratextId });
+  }
+
+  set onlineStatus(isOnline: boolean) {
+    this.testOnlineStatusService.setIsOnline(isOnline);
+    tick();
+    this.fixture.detectChanges();
   }
 
   setupProject(overrides?: Partial<SFProject>): void {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts
@@ -83,7 +83,7 @@ export class EditorTabAddResourceDialogComponent implements OnInit {
         }
         if (project?.projectId != null) {
           // Add the user to the project if they are not already connected to it
-          if (!project.isConnected && this.appOnline) {
+          if (!project.isConnected) {
             await this.projectService.onlineAddCurrentUser(project.projectId);
           }
           this.selectedProjectDoc =
@@ -99,11 +99,13 @@ export class EditorTabAddResourceDialogComponent implements OnInit {
 
         if (this.selectedProjectDoc != null) {
           if (this.permissionsService.canSync(this.selectedProjectDoc)) {
-            if (!this.selectedProjectDoc?.data?.texts?.length) {
+            if (!this.selectedProjectDoc.data?.texts.length) {
               this.isSyncActive = true;
               await this.projectService.onlineSync(this.selectedProjectDoc.id);
             } else {
-              this.projectService.onlineSync(this.selectedProjectDoc.id);
+              this.projectService
+                .onlineSync(this.selectedProjectDoc.id)
+                .catch(_ => console.warn('Syncing to resource project failed'));
               this.dialogRef.close(this.selectedProjectDoc);
             }
           } else {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts
@@ -56,6 +56,10 @@ export class EditorTabAddResourceDialogComponent implements OnInit {
     @Inject(MAT_DIALOG_DATA) readonly dialogData: EditorTabAddResourceDialogData
   ) {}
 
+  get appOnline(): boolean {
+    return this.onlineStatus.isOnline && this.onlineStatus.isBrowserOnline;
+  }
+
   async ngOnInit(): Promise<void> {
     await this.getProjectsAndResources();
   }
@@ -65,6 +69,7 @@ export class EditorTabAddResourceDialogComponent implements OnInit {
   }
 
   async confirmSelection(): Promise<void> {
+    this.resetErrors();
     const paratextId: string | null | undefined = this.form.value.sourceParatextId;
 
     try {
@@ -73,27 +78,32 @@ export class EditorTabAddResourceDialogComponent implements OnInit {
 
         // If the Paratext project has a SF project id, add the user to that project if they are not already
         const project = this.projects?.find(p => p.paratextId === paratextId);
+        if (!this.appOnline) {
+          return;
+        }
         if (project?.projectId != null) {
           // Add the user to the project if they are not already connected to it
-          if (!project.isConnected) {
+          if (!project.isConnected && this.appOnline) {
             await this.projectService.onlineAddCurrentUser(project.projectId);
           }
-          this.selectedProjectDoc = await this.projectService.get(project.projectId);
+          this.selectedProjectDoc =
+            project?.projectId != null ? await this.projectService.get(project.projectId) : undefined;
         } else {
           // Load the project or resource, creating it if it is not present
-          const projectId: string | undefined = await this.projectService.onlineCreateResourceProject(paratextId);
-          this.selectedProjectDoc = projectId != null ? await this.projectService.get(projectId) : undefined;
+          const projectId: string | undefined = this.appOnline
+            ? await this.projectService.onlineCreateResourceProject(paratextId)
+            : undefined;
+          this.selectedProjectDoc =
+            projectId != null && this.appOnline ? await this.projectService.get(projectId) : undefined;
         }
 
         if (this.selectedProjectDoc != null) {
           if (this.permissionsService.canSync(this.selectedProjectDoc)) {
-            // Wait for sync if no texts
-            if (!this.selectedProjectDoc.data?.texts.length) {
+            if (!this.selectedProjectDoc?.data?.texts?.length) {
               this.isSyncActive = true;
-              await this.syncProject(this.selectedProjectDoc.id);
+              await this.projectService.onlineSync(this.selectedProjectDoc.id);
             } else {
-              // Otherwise, start a sync in the background and close dialog
-              this.syncProject(this.selectedProjectDoc.id);
+              this.projectService.onlineSync(this.selectedProjectDoc.id);
               this.dialogRef.close(this.selectedProjectDoc);
             }
           } else {
@@ -104,11 +114,19 @@ export class EditorTabAddResourceDialogComponent implements OnInit {
         }
       }
     } catch (e) {
-      this.syncFailed = true;
       try {
-        this.cancelSync();
+        if (this.appOnline) {
+          this.isSyncActive = false;
+          this.syncFailed = true;
+          this.cancelSync();
+        }
       } catch {}
     } finally {
+      if (!this.appOnline) {
+        this.syncFailed = true;
+        this.isSyncActive = false;
+      }
+
       this.isLoading = false;
     }
   }
@@ -124,7 +142,7 @@ export class EditorTabAddResourceDialogComponent implements OnInit {
     this.isSyncActive = isActive;
 
     // Wait for sync to complete before closing dialog
-    if (!isActive) {
+    if (!isActive && this.selectedProjectDoc?.data?.texts?.length) {
       this.dialogRef.close(this.selectedProjectDoc);
     }
   }
@@ -165,10 +183,6 @@ export class EditorTabAddResourceDialogComponent implements OnInit {
     this.resourceLoadingFailed = false;
     this.projectFetchFailed = false;
     this.syncFailed = false;
-  }
-
-  private async syncProject(projectId: string): Promise<void> {
-    await this.projectService.onlineSync(projectId);
   }
 
   /**
--- a/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json
+++ b/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json
@@ -492,6 +492,7 @@
     "error_fetching_projects": "There was an error fetching projects. You will only be able to select resources from Digital Bible Library.",
     "error_fetching_resources": "There was an error fetching the Digital Bible Library resources. You will only be able to select projects.",
     "error_loading_resource": "There was an error loading the selected resource.",
+    "error_offline": "You are offline. Please connect to the internet to load this resource.",
     "loading": "Loading",
     "placeholder_loading": "{{ editor_add_tab_resource_dialog.loading }}...",
     "placeholder_ready": "Select paratext project or DBL resource",
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts
@@ -5,6 +5,7 @@ import { v1 as uuidv1 } from 'uuid';
 import { hasNumberProp, hasObjectProp, hasStringProp } from '../type-utils';
 import { BugsnagService } from './bugsnag.service';
 import { COMMAND_API_NAMESPACE } from './url-constants';
+import { OnlineStatusService } from './online-status.service';
 
 /** See also C# enum EdjCase.JsonRpc.Common.RpcErrorCode, which this somewhat matches. */
 export enum CommandErrorCode {
@@ -61,8 +62,13 @@ export class CommandError extends Error {
 export class CommandService {
   constructor(
     private readonly http: HttpClient,
-    private readonly bugsnagService: BugsnagService
+    private readonly bugsnagService: BugsnagService,
+    private readonly onlineStatus: OnlineStatusService
   ) {}
+
+  get appOnline(): boolean {
+    return this.onlineStatus.isOnline && this.onlineStatus.isBrowserOnline;
+  }
 
   async onlineInvoke<T>(url: string, method: string, params: any = {}): Promise<T | undefined> {
     url = `${COMMAND_API_NAMESPACE}/${url}`;
@@ -83,13 +89,17 @@ export class CommandService {
       'request'
     );
     try {
-      const response = await lastValueFrom(
-        this.http.post<JsonRpcResponse<T>>(url, request, { headers: { 'Content-Type': 'application/json' } })
-      );
-      if (response.error != null) {
-        throw response.error;
+      if (!this.appOnline) {
+        return Promise.resolve(undefined);
+      } else {
+        const response = await lastValueFrom(
+          this.http.post<JsonRpcResponse<T>>(url, request, { headers: { 'Content-Type': 'application/json' } })
+        );
+        if (response.error != null) {
+          throw response.error;
+        }
+        return response.result;
       }
-      return response.result;
     } catch (error) {
       // Transform the various kinds of errors into a CommandError.
 
@@ -119,7 +129,11 @@ export class CommandService {
         moreInformation = `Unexpected error type: ${error}`;
       }
       const message = `Error invoking ${method}: ${moreInformation}`;
-      throw new CommandError(code, message, data);
+      if (this.appOnline) {
+        throw new CommandError(code, message, data);
+      }
+      console.error(`${code} ${message} ${data}`);
+      return undefined;
     }
   }
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts
@@ -5,7 +5,6 @@ import { v1 as uuidv1 } from 'uuid';
 import { hasNumberProp, hasObjectProp, hasStringProp } from '../type-utils';
 import { BugsnagService } from './bugsnag.service';
 import { COMMAND_API_NAMESPACE } from './url-constants';
-import { OnlineStatusService } from './online-status.service';
 
 /** See also C# enum EdjCase.JsonRpc.Common.RpcErrorCode, which this somewhat matches. */
 export enum CommandErrorCode {
@@ -62,13 +61,8 @@ export class CommandError extends Error {
 export class CommandService {
   constructor(
     private readonly http: HttpClient,
-    private readonly bugsnagService: BugsnagService,
-    private readonly onlineStatus: OnlineStatusService
+    private readonly bugsnagService: BugsnagService
   ) {}
-
-  get appOnline(): boolean {
-    return this.onlineStatus.isOnline && this.onlineStatus.isBrowserOnline;
-  }
 
   async onlineInvoke<T>(url: string, method: string, params: any = {}): Promise<T | undefined> {
     url = `${COMMAND_API_NAMESPACE}/${url}`;
@@ -89,17 +83,13 @@ export class CommandService {
       'request'
     );
     try {
-      if (!this.appOnline) {
-        return Promise.resolve(undefined);
-      } else {
-        const response = await lastValueFrom(
-          this.http.post<JsonRpcResponse<T>>(url, request, { headers: { 'Content-Type': 'application/json' } })
-        );
-        if (response.error != null) {
-          throw response.error;
-        }
-        return response.result;
+      const response = await lastValueFrom(
+        this.http.post<JsonRpcResponse<T>>(url, request, { headers: { 'Content-Type': 'application/json' } })
+      );
+      if (response.error != null) {
+        throw response.error;
       }
+      return response.result;
     } catch (error) {
       // Transform the various kinds of errors into a CommandError.
 
@@ -129,11 +119,7 @@ export class CommandService {
         moreInformation = `Unexpected error type: ${error}`;
       }
       const message = `Error invoking ${method}: ${moreInformation}`;
-      if (this.appOnline) {
-        throw new CommandError(code, message, data);
-      }
-      console.error(`${code} ${message} ${data}`);
-      return undefined;
+      throw new CommandError(code, message, data);
     }
   }
 }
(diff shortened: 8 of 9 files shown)


kylebuss (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2684)* status: 0 of 6 files reviewed, 2 unresolved discussions

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts` line 243 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5-m6PgCv4y3RwEq7J2:-O5-m6PgCv4y3RwEq7J3:b-bs0dgw) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L243)):*
> ```XML
>       try {
>         // Don't show a dialog if app is offline or this is a silent error or that we just want sent to Bugsnag
>         if (!silently || this.appOnline) {
> ```

This is possibly not how we want to handle our errors as there might be cases that we would want the Error Dialog to open when the user is offline - but for loading a resource the user might see multiple error dialogs if they lose internet connection when attempting to load a new resource.



_Code quote:_
```XML
if (!silently || this.appOnline) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts` line 89 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5-mqpa44WULjKBcEff:-O5-mqpa44WULjKBcEfg:bvu7kr9) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts#L89)):*
> ```XML
>     if (this.appOnline) {
>       await doc.subscribe();
>     }
> ```

This might be a crude way of handling this but calling `await doc.subscribe()` when offline never resolves/returns from the realtime server which is at least partly responsible for the infinite loading bar.



_Code quote:_
```XML
    if (this.appOnline) {
      await doc.subscribe();
    }
```

RaymondLuong3 (commented):
Thanks for working on this. Let me know if you want to join a call and we can discuss this further. I had lots of questions, and I might not be fully comprehending the task or problem.

From looking at SF-2844, my understanding is that we should be more proactive to prevent a user from trying to use the dialog when they are offline. My suggestion is that we redesign the dialog to warn users when they are offline, rather than try to retroactively detect when an error occurred while they were offline (or are having network troubles).

The reason this is better is because if the user is truly offline, we can know beforehand if something will work or not (such as adding a user to a project/syncing a project). So why not notify them before they hit an error.

I also highly recommend we do not change the `exception-handling-service` but rather change the component's handling of errors.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-:-O5EflE32ZcQ7ezq9mH3:b-whoozd)* status: all files reviewed, 11 unresolved discussions (waiting on @kylebuss and @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html` line 50 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5EcALc3MRiU5JSUsB7:-O5EcALc3MRiU5JSUsB8:b91ugwr) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html#L50)):*
> ```HTML
>         {{ t("error_user_not_permitted") }}
>       </mat-error>
>     }
> ```

This should be removed.



_Code quote:_
```HTML
      <mat-error>
        {{ t("error_user_not_permitted") }}
      </mat-error>
    }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html` line 51 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5EehaLCjS_ShZtKtEG:-O5EehaLCjS_ShZtKtEH:bnytbjb) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html#L51)):*
> ```HTML
>       </mat-error>
>     }
>     @if (offlineFailure && !appOnline) {
> ```

This message can show in the event that the dialog is open and the user is offline.



_Code quote:_
```HTML
    @if (offlineFailure && !appOnline) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html` line 69 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5ESpWW32PupszomfaE:-O5ESpWX-MLrZxRdiARB:bkj9yyf) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html#L69)):*
> ```HTML
>       color="primary"
>       [disabled]="
>         !form.valid || isLoading || isSyncActive || projectFetchFailed || syncFailed || (offlineFailure && !appOnline)
> ```

The select button could probably disabled if the dialog is offline.



_Code quote:_
```HTML
 (offlineFailure && !appOnline)
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 85 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5EccMC64TyWIfBDRO9:-O5EccMC64TyWIfBDROA:b-jmqy6r) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L85)):*
> ```XML
>           // Add the user to the project if they are not already connected to it
>           if (!project.isConnected) {
>             await this.projectService.onlineAddCurrentUser(project.projectId);
> ```

Before calling something with 'online' in the name, we should probably check that the app is actually online.



_Code quote:_
```XML
           await this.projectService.onlineAddCurrentUser(project.projectId);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 87 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5EeoxUD2APbGwWZdbh:-O5EeoxUD2APbGwWZdbi:bousdfn) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L87)):*
> ```XML
>             await this.projectService.onlineAddCurrentUser(project.projectId);
>           }
>           this.selectedProjectDoc = await this.projectService.get(project.projectId);
> ```

You might want to do this as a final clean-up task:
We do not need to get access to the full project, just the project profile (i.e. `this.projectService.getProfile(...)`. The profile is more likely to be available in indexedDB. This small change will require additional changes for the tests to work.



_Code quote:_
```XML
          this.selectedProjectDoc = await this.projectService.get(project.projectId);
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 98 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5EZgrE4ACCQ0ZbLDY0:-O5EZgrE4ACCQ0ZbLDY1:bgqd1c) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L98)):*
> ```XML
>             await this.syncProject(this.selectedProjectDoc.id);
>           } else {
>             this.userNotPermitted = true;
> ```

I don't think this is needed. If a user is not permitted to sync the project, it does not mean they should not be able to see the resource if the project already exists in SF.



_Code quote:_
```XML
            this.userNotPermitted = true;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 117 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5Ed0mi5XwgMQquLc-d:-O5Ed0mi5XwgMQquLc-e:bf4umz0) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L117)):*
> ```XML
>         this.projectFetchFailed = false;
>         this.offlineFailure = true;
>       }
> ```

Instead of setting `offlineFailure` retroactively, it might be best to just show a warning message when the dialog is open and the user is offline.
There is always the possibility that the instant the user clicks and we try to connect to the project there is a network failure. But in that case, the dialog could remain open, and when internet is restored, the connect button can be enabled again.



_Code quote:_
```XML
      if (!this.appOnline) {
        this.isSyncActive = false;
        this.syncFailed = false;
        this.projectFetchFailed = false;
        this.offlineFailure = true;
      }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 178 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5EYkKmDhQVODALXqcN:-O5EYkKmDhQVODALXqcO:b2nhodu) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L178)):*
> ```XML
> 
>   private async syncProject(projectId: string): Promise<void> {
>     this.isSyncActive = true;
> ```

If this method depends on the app to be online, then you can save yourself extra indentations by returning early if the app is not online. `if (!this.appOnline) return`
Also, isSyncActive should only be set to true if the app is online if I understand correctly.



_Code quote:_
```XML
    this.isSyncActive = true;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts` line 243 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5-m6PgCv4y3RwEq7J2:-O5EVTzN2xEjPPQcB3ud:b-1dweyn) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.ts#L243)):*

This is possibly not how we want to handle our errors as there might be cases that we would want the Error Dialog to open when the user is offline - but for loading a resource the user might see multiple error dialogs if they lose internet connection when attempting to load a new resource.
</blockquote></details>

This is probably not what we want to do. Many errors can happen while a user is offline, and we want the error dialogs to show. Let's revert the changes in this file.
If a component is not behaving while offline, we should fix it in the component as best as we can and let the error handler freedom to report the errors it receives.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.spec.ts` line 31 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5EVt7C4SW381d4QLBx:-O5EVt7C4SW381d4QLBy:btz7ks2) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.spec.ts#L31)):*
> ```XML
> const mockedNoticeService = mock(NoticeService);
> const mockedPwaService = mock(PwaService);
> const mockedCookieService = mock(CookieService);
> ```

You can revert the changes to this file also.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts` line 89 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5-mqpa44WULjKBcEff:-O5EW6Ra492vRjYMbcLo:bitvw71) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts#L89)):*

This might be a crude way of handling this but calling `await doc.subscribe()` when offline never resolves/returns from the realtime server which is at least partly responsible for the infinite loading bar.
</blockquote></details>

Would it be better if we detect in the components whether the app is offline before trying to get and subscribe to the real-time doc? Say that the user is offline, and we try to call `SFProjectService.get()` and we get the project doc, but it turns out that we are not subscribed to changes. This could cause bugs when the user is back online since we had forgotten in offline mode, docs are not subscribed to.

Put it more simply, we don't want to be overly creative. If the user is offline and they want to access their project, explain that it is not possible.

kylebuss (commented):
I've made the changes, tests are still failing so I have not done anything with those yet as I hope I have identified the root of the error dialogs and we probably need to discuss if my changes are how we want to handle them.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-:-O5JTGo01KcHZrUbHWgD:b-4krwde)* status: all files reviewed, 9 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html` line 50 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5EcALc3MRiU5JSUsB7:-O5JK0yP6nEFkKWQz69h:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html#L50)):*

This should be removed.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html` line 51 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5EehaLCjS_ShZtKtEG:-O5JK1p_4DrqTvDIQ0rV:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html#L51)):*

This message can show in the event that the dialog is open and the user is offline.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html` line 69 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5ESpWW32PupszomfaE:-O5JK2vY86fMJQx5arIp:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html#L69)):*

The select button could probably disabled if the dialog is offline.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 85 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5EccMC64TyWIfBDRO9:-O5JK5Qy4Uwtbx_AyBx4:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L85)):*

Before calling something with 'online' in the name, we should probably check that the app is actually online.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 87 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5EeoxUD2APbGwWZdbh:-O5JMafU-uS9k7lHIgY5:b-hdh93q) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L87)):*

You might want to do this as a final clean-up task:
We do not need to get access to the full project, just the project profile (i.e. `this.projectService.getProfile(...)`. The profile is more likely to be available in indexedDB. This small change will require additional changes for the tests to work.
</blockquote></details>

I have added some additional checking. But I think the original code before and after this line address the concern between loading the project vs. the project profile. Let me know if you still think we should update this to load the profile.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 98 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5EZgrE4ACCQ0ZbLDY0:-O5JLnW7BUUqqdEM399Q:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L98)):*

I don't think this is needed. If a user is not permitted to sync the project, it does not mean they should not be able to see the resource if the project already exists in SF.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 117 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5Ed0mi5XwgMQquLc-d:-O5JLtxL4ATWDOQFZmdb:b273k58) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L117)):*

Instead of setting `offlineFailure` retroactively, it might be best to just show a warning message when the dialog is open and the user is offline.
There is always the possibility that the instant the user clicks and we try to connect to the project there is a network failure. But in that case, the dialog could remain open, and when internet is restored, the connect button can be enabled again.
</blockquote></details>

Changes made to show offline message and disable select button only partially solve the issue if the user lost connection when trying to load the resource. Keeping this code to ensure the error message accurately displays to the user that the load failure was due to losing internet connection.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 178 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5EYkKmDhQVODALXqcN:-O5JN2uRD8RFJtZsz5hb:b-73jb4s) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L178)):*

If this method depends on the app to be online, then you can save yourself extra indentations by returning early if the app is not online. `if (!this.appOnline) return`
Also, isSyncActive should only be set to true if the app is online if I understand correctly.
</blockquote></details>

Removed this method completely after seeing Nathaniel's comment in Slack on not awaiting an async call. I also think it is better for the user to keep the dialog open and show the offline notification in the dialog rather than closing it and displaying it in the editor.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.spec.ts` line 31 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5EVt7C4SW381d4QLBx:-O5JRoQS7WkxWSqtayer:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/exception-handling-service.spec.ts#L31)):*

You can revert the changes to this file also.
</blockquote></details>

Done.

kylebuss (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2684)* status: 1 of 8 files reviewed, 11 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/project-notification.service.ts` line 35 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JVzzc-UT8pw0em_VU:-O5JVzzc-UT8pw0em_VV:b-lzwit9) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/core/project-notification.service.ts#L35)):*
> ```XML
>       // These will be thrown if a user navigates away quickly after
>       // starting the sync or the app loses internet connection
>       if (err instanceof AbortError || !this.appOnline) {
> ```

This was the other tweak to catch an invalid offline error from displaying the error dialog. I think this one is okay as it falls under an error that is not caused by a server error. Let me know your thoughts.



_Code quote:_
```XML
      // starting the sync or the app loses internet connection
      if (err instanceof AbortError || !this.appOnline) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts` line 133 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JVHS-6_9v2RMncKl8:-O5JVHS-6_9v2RMncKl9:b-s5oe8o) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L133)):*
> ```XML
>       console.error(`${code} ${message} ${data}`);
>       return undefined;
>     }
> ```

This is one of the changes I'm uncertain if we want to handle it this way. We may additionally want to check the response code is not one of the expected JsonRPC error codes?



_Code quote:_
```XML
      if (this.appOnline) {
        throw new CommandError(code, message, data);
      }
      console.error(`${code} ${message} ${data}`);
      return undefined;
    }
```

RaymondLuong3 (commented):
Thanks for the changes. This is much closer to a good way to handle network failures for this component. You should add a test that when offline, the offline message appears. And a second test showing that `offlineFailure` is set correctly when there is no internet connection after the user clicks the select button.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-:-O5JlfbT2wqTtmWn0UvR:byg0ef9)* status: all files reviewed, 8 unresolved discussions (waiting on @kylebuss and @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/project-notification.service.ts` line 35 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JVzzc-UT8pw0em_VU:-O5JhZ9F3IwyIykMub7A:b-nyhbi5) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/core/project-notification.service.ts#L35)):*

This was the other tweak to catch an invalid offline error from displaying the error dialog. I think this one is okay as it falls under an error that is not caused by a server error. Let me know your thoughts.
</blockquote></details>

This seems to be a more appropriate way to handle errors from the sync progress. I'm a little skittish hiding errors, but in this case I think this will help suppress the error we don't want to see and other sync errors will still appear.
FYI, brief network problems may not result in `this.appOnline` to be false.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html` line 46 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5Jf6ig3a1C7-Mfamrr:-O5Jf6ig3a1C7-Mfamrs:bx203ux) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.html#L46)):*
> ```HTML
>       </mat-error>
>     }
>     @if (!appOnline || this.offlineFailure) {
> ```

Nit: You can take the `this` off of the variable.
Hey, what do you think of this, you can move the `offlineFailure` check into the `else if`. That way, the message for errors loading resources shows if there are errors, and the "please connect to the internet" message shows only when the app is offline, then it goes away if connection is restored.



_Code quote:_
```HTML
 this.offlineFailure) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 117 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5Ed0mi5XwgMQquLc-d:-O5Jkj1FEGz3NVDAxnFX:b-il6u08) ([raw file](https://github.com/sillsdev/web-xforge/blob/b0c239836b20386fa62e14dd41a266f2a7a9bf1e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L117)):*

Changes made to show offline message and disable select button only partially solve the issue if the user lost connection when trying to load the resource. Keeping this code to ensure the error message accurately displays to the user that the load failure was due to losing internet connection.
</blockquote></details>

With the changes you have made, it makes sense only set the offline failure message to true and the other messages to be false.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 78 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5Jh5vVDC1Xmr7aHlNq:-O5Jh5vVDC1Xmr7aHlNr:bbzlg1q) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L78)):*
> ```XML
>       if (paratextId != null) {
>         this.isLoading = true;
> ```

At the top of `confirmSelection` you can add a few lines of code that will help you clean up the remainder of the method. For starters, you can just return immediately if the user is offline since this method depends on the user being online.



_Code snippet:_
```XML
if (!this.appOnline) {
  this.offlineFailure = true;
  return;
}
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 83 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JizLfAqkaVpmVW5-2:-O5JizLfAqkaVpmVW5-3:b-vcct27) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L83)):*
> ```XML
>         if (project?.projectId != null) {
>           // Add the user to the project if they are not already connected to it
>           if (!project.isConnected && this.appOnline) {
> ```

Then these checks can be removed.



_Code quote:_
```XML
this.appOnline
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 101 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JmV8e6W8MQLmMI9PJ:-O5JmV8e6W8MQLmMI9PK:bfgyce8) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L101)):*
> ```XML
>             if (this.appOnline) {
>               this.isSyncActive = true;
>               await this.projectService.onlineSync(this.selectedProjectDoc.id);
> ```

Awaiting the sync is what ensures that the try/catch block actually catches any errors before the dialog is closed, is that correct?
I have a recommendation. Revert this change so we only await the online sync when the text is not available, then when it is available, you can avoid awaiting online sync to finish, but add this line.



_Code snippet:_
```XML
if (!this.selectedProjectDoc.data?.texts.length) {
  this.isSyncActive = true;
  await this.projectService.onlineSync(this.selectedProjectDoc.id);
} else {
  // catch the sync error and close the dialog
  this.projectService.onlineSync(this.selectedProjectDoc.id).catch(() => this.dialogRef.close(..)
  this.dialogRef.close(...)
}
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 103 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JjBEQ5797x_8RkiiK:-O5JjBEQ5797x_8RkiiL:b-55xdah) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L103)):*
> ```XML
>               await this.projectService.onlineSync(this.selectedProjectDoc.id);
>             } else {
>               this.offlineFailure = true;
> ```

This can be removed also.



_Code quote:_
```XML
              this.offlineFailure = true;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 482 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JlLXy6XoEEROhVJdM:-O5JlLXy6XoEEROhVJdN:b-1qn7z6) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L482)):*
> ```JSON
>     "error_loading_resource": "There was an error loading the selected resource.",
>     "error_offline": "You are offline. Please connect to the internet to load this resource.",
>     "error_user_not_permitted": "You do not have permission to access this project. Please contact your project administrator.",
> ```

This can be removed.



_Code quote:_
```JSON
 "error_user_not_permitted": "You do not have permission to access this project. Please contact your project administrator.",
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts` line 133 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JVHS-6_9v2RMncKl8:-O5JnFsP1JSG1mutIafX:b-b0uiv1) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L133)):*

This is one of the changes I'm uncertain if we want to handle it this way. We may additionally want to check the response code is not one of the expected JsonRPC error codes?
</blockquote></details>

Hmm, I would rather not suppress errors like this. The user is not notified that there was a problem, and they would just expect what they tried to do succeeded. It is better to catch the error in the component and explain to the user what happened.

kylebuss (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2684)* status: all files reviewed, 8 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 78 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5Jh5vVDC1Xmr7aHlNq:-O5KgpPkApHaVLxmZ41n:b-i05qi8) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L78)):*

At the top of `confirmSelection` you can add a few lines of code that will help you clean up the remainder of the method. For starters, you can just return immediately if the user is offline since this method depends on the user being online.
</blockquote></details>

The problem is there is about 3 seconds from clicking select to the when the sync begins. If we only perform one check there is a chance that we can pass that check but end up offline before we attempt the sync.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 101 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JmV8e6W8MQLmMI9PJ:-O5KhdB_Ca6zcYDGTXIq:bwxquzt) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L101)):*

Awaiting the sync is what ensures that the try/catch block actually catches any errors before the dialog is closed, is that correct?
I have a recommendation. Revert this change so we only await the online sync when the text is not available, then when it is available, you can avoid awaiting online sync to finish, but add this line.
</blockquote></details>

I was going back and forth on this one. I can go both ways, I wonder if we keep the dialog open so we display any potential failures to sync and maybe we don't disable the Select button if we encounter a failure?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 482 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JlLXy6XoEEROhVJdM:-O5KQFDh6lnfCaf2R9Sm:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L482)):*

This can be removed.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts` line 133 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JVHS-6_9v2RMncKl8:-O5KaCsq6HciQXx6btoO:bnsgt48) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L133)):*

Hmm, I would rather not suppress errors like this. The user is not notified that there was a problem, and they would just expect what they tried to do succeeded. It is better to catch the error in the component and explain to the user what happened.
</blockquote></details>

What are your thoughts on moving the online check to the top of the onlineInvoke method? This way we check before sending the request to the backend and won't be messing with the error handling at all. I'm still doing some testing of this as I'm getting varying results.



_Code snippet:_
```XML
const response = this.appOnline ? await lastValueFrom(
        this.http.post<JsonRpcResponse<T>>(url, request, { headers: { 'Content-Type': 'application/json' } })
      ) : undefined;
      if (response?.error != null) {
        throw response.error;
      }
      return response?.result;
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2684)* status: all files reviewed, 8 unresolved discussions (waiting on @kylebuss and @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 78 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5Jh5vVDC1Xmr7aHlNq:-O5KsQi151OpqGq0zWFF:b-5zmvcj) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L78)):*

The problem is there is about 3 seconds from clicking select to the when the sync begins. If we only perform one check there is a chance that we can pass that check but end up offline before we attempt the sync.
</blockquote></details>

Right, I think that is ok. The one check at the top keeps things simple, and in the case of an intermittent network failure, there will be an error that should be caught.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 101 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JmV8e6W8MQLmMI9PJ:-O5KstPu3-VKSby2cY82:b6concf) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L101)):*

I was going back and forth on this one. I can go both ways, I wonder if we keep the dialog open so we display any potential failures to sync and maybe we don't disable the Select button if we encounter a failure?
</blockquote></details>

There isn't much advantage with keeping the dialog open and waiting for the sync to finish before allowing the user to see the resource. The only advantage is that they would see that the sync failed, but that is not necessarily helpful and will just get them confused. It would be better just close the dialog and let the user see the out-of-date resource.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts` line 133 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JVHS-6_9v2RMncKl8:-O5Ktcou6FlAaMCOLVe1:b5gxjhz) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L133)):*

What are your thoughts on moving the online check to the top of the onlineInvoke method? This way we check before sending the request to the backend and won't be messing with the error handling at all. I'm still doing some testing of this as I'm getting varying results.
</blockquote></details>

Right... that would work. We can just throw an error immediately if the user is not online. Let's give that a try.

kylebuss (commented):
Updated changes and added tests.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-:-O5VBrB8DmaJ0O-K5jLR:bg7ucgt)* status: 3 of 48 files reviewed, 5 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 78 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5Jh5vVDC1Xmr7aHlNq:-O5VAQ5j004hwZuxiNjb:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L78)):*

Right, I think that is ok. The one check at the top keeps things simple, and in the case of an intermittent network failure, there will be an error that should be caught.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 83 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JizLfAqkaVpmVW5-2:-O5VANHW8WqWI5GxKKYW:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L83)):*

Then these checks can be removed.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 101 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JmV8e6W8MQLmMI9PJ:-O5UTBpJCuCAForkbVO8:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L101)):*

There isn't much advantage with keeping the dialog open and waiting for the sync to finish before allowing the user to see the resource. The only advantage is that they would see that the sync failed, but that is not necessarily helpful and will just get them confused. It would be better just close the dialog and let the user see the out-of-date resource.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 103 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JjBEQ5797x_8RkiiK:-O5UTCvaB2TqFUwFRWL1:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L103)):*

This can be removed also.
</blockquote></details>

Done.

RaymondLuong3 (changes requested):
> Reviewed 8 of 47 files at r3, 39 of 39 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2684)* status: all files reviewed, 6 unresolved discussions (waiting on @kylebuss and @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.spec.ts` line 59 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5Zm2dL0JAaEp9Oga72:-O5Zm2dL0JAaEp9Oga73:brvh175) ([raw file](https://github.com/sillsdev/web-xforge/blob/faf72028fe03d4c54ea3d1527a247e1f4d2f8a2c/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.spec.ts#L59)):*
> ```XML
> 
>   it('does not initialize if app is offline', fakeAsync(async () => {
>     const env = new TestEnvironment({ userId: 'user01' });
> ```

You will have to call env.setupProjectDoc() to accurately test that the `initialize()` method returns early if the user is offline.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 83 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JizLfAqkaVpmVW5-2:-O5ZmYyZCwf7lYIQnxGW:b46lvrf) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L83)):*

Done.
</blockquote></details>

We might have mistakenly re-added the online check here. Remove again.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 102 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5ZppiWBAMzTH-R89T_:-O5ZppiWBAMzTH-R89Ta:bcqk1nt) ([raw file](https://github.com/sillsdev/web-xforge/blob/faf72028fe03d4c54ea3d1527a247e1f4d2f8a2c/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L102)):*
> ```XML
>         if (this.selectedProjectDoc != null) {
>           if (this.permissionsService.canSync(this.selectedProjectDoc)) {
>             if (!this.selectedProjectDoc?.data?.texts?.length) {
> ```

You can remove the optional chaining on the doc and the text



_Code quote:_
```XML
  if (!this.selectedProjectDoc?.data?.texts?.length)
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 106 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5Zn6E97Jp_AMt7H-K8:-O5Zn6E97Jp_AMt7H-K9:b-1bxs7) ([raw file](https://github.com/sillsdev/web-xforge/blob/faf72028fe03d4c54ea3d1527a247e1f4d2f8a2c/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L106)):*
> ```XML
>               await this.projectService.onlineSync(this.selectedProjectDoc.id);
>             } else {
>               this.projectService.onlineSync(this.selectedProjectDoc.id);
> ```

Since we are calling this and never awaiting it, an error can occur after this block completes and the error won't be caught. You can change this to catch the error when the promise rejects.



_Code snippet:_
```XML
this.projectService.onlineSync(this.selectedProjectDoc.id).catch(_ => console.warn('Syncing resource failed');
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts` line 94 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5ZoQUo-xUyxGkUqMf7:-O5ZoQUo-xUyxGkUqMf8:bj0yrll) ([raw file](https://github.com/sillsdev/web-xforge/blob/faf72028fe03d4c54ea3d1527a247e1f4d2f8a2c/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L94)):*
> ```XML
>       if (!this.appOnline) {
>         return Promise.resolve(undefined);
>       } else {
> ```

Nit: You don't need an else when the if statement returns.
With my suggestions, can you remove this change and see if there is still an error. I think we are going to regret suppressing errors in this method because we are silently failing to do something. When this throws, it helps us recognize that the caller made a request, and it failed. By failing silently here, we are not doing ourselves any favours. We should revert this if at all possible.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts` line 131 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5Zo9pNB6JlJ5Qs8vAk:-O5Zo9pNB6JlJ5Qs8vAl:b6bl407) ([raw file](https://github.com/sillsdev/web-xforge/blob/faf72028fe03d4c54ea3d1527a247e1f4d2f8a2c/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L131)):*
> ```XML
>         moreInformation = `Unexpected error type: ${error}`;
>       }
>       const message = `Error invoking ${method}: ${moreInformation}`;
> ```

Let's remove this. Adding the check if the app is online is already enough handling.

kylebuss (commented):
I removed the offline checks in `command.service.ts` and was not seeing the error dialogs. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-:-O5Zy_e7BJ2Jab2TRPwS:bnf15mn)* status: all files reviewed, 5 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.spec.ts` line 59 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5Zm2dL0JAaEp9Oga72:-O5Zy90SA82pn0FjG4eI:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/faf72028fe03d4c54ea3d1527a247e1f4d2f8a2c/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.spec.ts#L59)):*

You will have to call env.setupProjectDoc() to accurately test that the `initialize()` method returns early if the user is offline.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 83 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5JizLfAqkaVpmVW5-2:-O5ZyAUc29hIvaKVXMMA:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/6421eba7736a845661031a1821569ef88958fcb3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L83)):*

We might have mistakenly re-added the online check here. Remove again.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 102 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5ZppiWBAMzTH-R89T_:-O5ZyBVH7UeZpaVqj4d-:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/faf72028fe03d4c54ea3d1527a247e1f4d2f8a2c/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L102)):*

You can remove the optional chaining on the doc and the text
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts` line 106 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5Zn6E97Jp_AMt7H-K8:-O5ZyCqeCHZ3JWvFcPV9:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/faf72028fe03d4c54ea3d1527a247e1f4d2f8a2c/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-add-resource-dialog/editor-tab-add-resource-dialog.component.ts#L106)):*

Since we are calling this and never awaiting it, an error can occur after this block completes and the error won't be caught. You can change this to catch the error when the promise rejects.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts` line 131 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-O5Zo9pNB6JlJ5Qs8vAk:-O5ZyJLp8mFNCawJR74L:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/faf72028fe03d4c54ea3d1527a247e1f4d2f8a2c/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/command.service.ts#L131)):*

Let's remove this. Adding the check if the app is online is already enough handling.
</blockquote></details>

Done.

RaymondLuong3 (approved):
I think we finally figured it out! Nice work. :lgtm: 

> Reviewed 3 of 3 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2684#-:-O5_Es0-4LYHKKHdWWTI:brep0y3)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2700, opened 2024-09-04
Fix @for track expression in error dialog

Nateowami:
If you select the error ID in the error dialog and press Ctrl, it clears the selection (there's probably a bunch of other ways to trigger this bug, but this is one). You can see this even in Storybook.

I think the reason is because it would trigger change detection, and the `track` option was wrong. The object it specifies changes each time, and so it would result in always creating new elements, which results in clearing the user's selection. Tracking by the text ensures that Angular understands nothing actually changed, and it doesn't need to re-render the elements.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2700)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-dialog/error-dialog.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/error-dialog/error-dialog.component.html
@@ -7,7 +7,7 @@
       <p class="unsupported-browser" [innerHTML]="t('unsupported_browser', browserLinks)"></p>
     }
     <p>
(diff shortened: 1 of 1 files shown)


RaymondLuong3 (approved):
Nice one finding the fix for this. I've tested it and it does appear to work as expected.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2700#-:-O5y45Gb8yfR0O6dNc7_:b-b1dsiv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2707, opened 2024-09-05
Enable de and npi locales in production

Nateowami:
@nigel-wells I've updated the JSON files in S3. I thought we needed to add the locales in the login page itself, but that appears to not be the case based on my testing. Does that mean we can enable this in production straight away?

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2707)


RaymondLuong3:
I just updated the PR to update the Nepali strings on the pages in Auth0. That can be merged after this is merged.

nigel-wells:
The Auth0 login page template does need to be updated to have the new language codes added - npi is only available on `sil-appbuilder` at the moment. Raymond's [PR ](https://github.com/sillsdev/auth0-configs/pull/147)contains this update so we can look to get that merged into QA to further test this. 
image

Nateowami:
@nigel-wells Well, `de` is not on `sil-appbuilder`, but it works. Are you sure it needs to be added to the list?

nigel-wells:
Looks like the `availableLangaugeOptions` only relates to when a language picker appears and doesn't restrict the language code passed. The code to fetch the JSON will attempt whatever language is set which will explain why it is working.

I'm not actually sure who uses the language picker any more on the login page - do other Auth0 apps use it? SF used it a along time ago but not any more as far as I'm aware.

Nateowami:
@nigel-wells So can we merge this in then?

nigel-wells:
Yes, so long as the Auth0 PR moves at the same pace as it is provides translation strings for the rest of Auth0. 
Do we have another PR for DE for the rest of Auth0?

Nateowami:
@nigel-wells Why do we have to do that first? Can't it just fall back to English for the emails?

nigel-wells:
Doesn't have to be first - technically it doesn't even have to happen. I figured if we were going to announce, "We now support Napali" then it would be best to have it rolled out. Your call.

nigel-wells (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2707#-:-O6Y31jzDg3SCLBSYQoH:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2708, opened 2024-09-05
SF-2934 Notify user when joining a project fails

RaymondLuong3:
Users cannot join a project while they are offline because it requires hitting an endpoint on the server. This change will allow the error to be caught while the user is not online and notify the user that they failed to join the project.

My projects failed to join

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2708)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html
@@ -87,6 +87,11 @@
           {{ t(errorMessage) }}
         </app-notice>
       }
+      @if ((isOnline | async) === false) {
+        <app-notice id="message-offline" icon="cloud_off" [type]="userHasProjects ? 'info' : 'error'">
+          {{ t("offline") }}
+        </app-notice>
+      }
       @for (ptProject of userUnconnectedParatextProjects; track ptProject) {
         <mat-card
           attr.data-pt-project-id="{{ ptProject.paratextId }}"
@@ -120,7 +125,7 @@
               mat-flat-button
               color="primary"
               (click)="joinProject(ptProject.projectId)"
-              [disabled]="joiningProjects.includes(ptProject.projectId)"
+              [disabled]="!(isOnline | async) || joiningProjects.includes(ptProject.projectId)"
             >
               {{ t("join") }}
             </button>
@@ -138,11 +143,6 @@
           <div class="loading-project-description"></div>
           <div class="loading-action"></div>
         </mat-card>
-      }
-      @if ((isOnline | async) === false) {
-        <app-notice id="message-offline" icon="cloud_off" [type]="userHasProjects ? 'info' : 'error'">
-          {{ t("offline") }}
-        </app-notice>
       }
     </div>
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html
@@ -87,7 +87,7 @@
           {{ t(errorMessage) }}
         </app-notice>
       }
-      @if ((isOnline | async) === false) {
+      @if (!(isOnline | async)) {
         <app-notice id="message-offline" icon="cloud_off" [type]="userHasProjects ? 'info' : 'error'">
           {{ t("offline") }}
         </app-notice>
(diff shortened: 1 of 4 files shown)


kylebuss (commented):
Thanks for picking this one up. Not sure how I missed Joseph's comments on the Jira ticket. I just had a couple of comments for you to look at.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2708#-:-O689g5qEDGpB1eLecn_:b695dal)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html` line 90 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2708#-O685XQyBL70r91AqUhf:-O685XQyBL70r91AqUhg:bomht7w) ([raw file](https://github.com/sillsdev/web-xforge/blob/d27b5c5f96b68ee371db53b63798632be3e04849/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html#L90)):*
> ```HTML
>         </app-notice>
>       }
>       @if ((isOnline | async) === false) {
> ```

nit This could be switched this to a logical NOT to match the code added for disabling the button.



_Code quote:_
```HTML
(isOnline | async) === false
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts` line 127 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2708#-O687GAUElB5btPxQmEY:-O687GAUElB5btPxQmEZ:b-9ljr6y) ([raw file](https://github.com/sillsdev/web-xforge/blob/d27b5c5f96b68ee371db53b63798632be3e04849/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts#L127)):*
> ```XML
>       this.noticeService.loadingFinished();
>       this.joiningProjects.pop();
>     }
> ```

I like how this gracefully handles the error of going offline in the middle of trying to join a project. When I tested this out locally, I noticed that the message did not display until my internet connection resumed. This is good if the internet is spotty as resuming connection won't necessarily break the sync we've begun. 

I wonder if there needs to be a timeout after 1 minute that will display the message in cases where the internet does not resume quickly?



_Code quote:_
```XML
    try {
      this.noticeService.loadingStarted();
      this.joiningProjects.push(projectId);
      await this.projectService.onlineAddCurrentUser(projectId);
      this.router.navigate(['projects', projectId]);
    } catch (err) {
      this.noticeService.show(translate('my_projects.failed_to_join_project'));
    } finally {
      this.noticeService.loadingFinished();
      this.joiningProjects.pop();
    }
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2708)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html` line 90 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2708#-O685XQyBL70r91AqUhf:-O6MejhV5G3xmKyuKGIG:b-7rl0jd) ([raw file](https://github.com/sillsdev/web-xforge/blob/d27b5c5f96b68ee371db53b63798632be3e04849/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html#L90)):*

nit This could be switched this to a logical NOT to match the code added for disabling the button.
</blockquote></details>

done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts` line 127 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2708#-O687GAUElB5btPxQmEY:-O6MelYI73k98a4-tOl4:bvpiwja) ([raw file](https://github.com/sillsdev/web-xforge/blob/d27b5c5f96b68ee371db53b63798632be3e04849/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts#L127)):*

I like how this gracefully handles the error of going offline in the middle of trying to join a project. When I tested this out locally, I noticed that the message did not display until my internet connection resumed. This is good if the internet is spotty as resuming connection won't necessarily break the sync we've begun. 

I wonder if there needs to be a timeout after 1 minute that will display the message in cases where the internet does not resume quickly?
</blockquote></details>

Thanks for checking this out. I think this should be OK, since if the user ever gets to this point, they recently were connected to the internet since the join button is only enabled if they are connected.

kylebuss (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2708)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

kylebuss (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2708)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

kylebuss:
:lgtm:

PR #2710, opened 2024-09-06
WIP SF-2911 Fix sync failures caused by Paratext username changes

kylebuss:
Updated `existingUsers` to check the Projects Paratext users by `SFUserId` rather than just on `username`.
Added conditional to compare on `Username` and update `existingUser` if it has changed.

Updated tests to include optional `SFUserId` for user name changes or other future tests when setting the SF Data.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2710)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/note-dialog/note-dialog.component.ts
@@ -432,11 +432,14 @@ export class NoteDialogComponent implements OnInit {
 
     // If the user is not a PT user, or the note was created in SF, or the user created the note
     if (
-      syncUser == null || // There is no sync user, i.e. the note is not synced yet or the current user is not a PT user
-      note.editable || // Only notes created in SF are editable, so display the SF owner, falling back to the sync user
-      syncUser.sfUserId === ownerDoc.id // The note is not editable, but the sync user is the owner, so use the SF owner
+      // There is no sync user, i.e. the note is not synced yet or the current user is not a PT user
+      syncUser == null ||
+      // Only notes created in SF are editable, so display the SF owner, falling back to the sync user
+      note.editable ||
+      // The note is not editable, but the sync user is the owner, so use the SF owner
+      syncUser.sfUserId === note.ownerRef
     ) {
-      return this.userService.currentUserId === ownerDoc.id
+      return this.userService.currentUserId === note.ownerRef
         ? translate('checking.me') // "Me", i.e. the current user
         : (ownerDoc.data?.displayName ?? // Another user
             syncUser?.username ?? // Fallback to the sync user
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -2823,8 +2823,7 @@ public class ParatextService : DisposableBase, IParatextService
 
                     // We can only update a note if the comment and note have the same version number.
                     // Or if the note is authored by a commenter, set the comment content to be the SF note content
-                    bool isCommenterNote =
-                        ptProjectUsers.Values.SingleOrDefault(p => p.SFUserId == note.OwnerRef) == null;
+                    bool isCommenterNote = !ptProjectUsers.Values.Any(p => p.SFUserId == note.OwnerRef);
                     if (note.VersionNumber == comment.VersionNumber || isCommenterNote)
                     {
                         bool commentUpdated = false;
@@ -2986,10 +2985,11 @@ public class ParatextService : DisposableBase, IParatextService
     {
         if (note.Content == null)
             return null;
+        string syncUser = note.SyncUserRef;
         string ownerId = note.OwnerRef;
         // if the user is a paratext user, keep the note content as is
         if (
-            ptProjectUsers.Values.SingleOrDefault(u => u.SFUserId == ownerId) != null
+            ptProjectUsers.Values.Any(u => u.SFUserId == ownerId)
             || !displayNames.TryGetValue(ownerId, out string displayName)
         )
             return note.Content;
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -2823,7 +2823,7 @@ public class ParatextService : DisposableBase, IParatextService
 
                     // We can only update a note if the comment and note have the same version number.
                     // Or if the note is authored by a commenter, set the comment content to be the SF note content
-                    bool isCommenterNote = !ptProjectUsers.Values.Any(p => p.SFUserId == note.OwnerRef);
+                    bool isCommenterNote = !ptProjectUsers.Values.Any(u => u.SFUserId == note.OwnerRef);
                     if (note.VersionNumber == comment.VersionNumber || isCommenterNote)
                     {
                         bool commentUpdated = false;
@@ -2985,13 +2985,10 @@ public class ParatextService : DisposableBase, IParatextService
     {
         if (note.Content == null)
             return null;
-        string syncUser = note.SyncUserRef;
-        string ownerId = note.OwnerRef;
+
         // if the user is a paratext user, keep the note content as is
-        if (
-            ptProjectUsers.Values.Any(u => u.SFUserId == ownerId)
-            || !displayNames.TryGetValue(ownerId, out string displayName)
-        )
+        bool isParatextUser = ptProjectUsers.Any(u => u.Value.SFUserId == note.OwnerRef);
+        if (isParatextUser || !displayNames.TryGetValue(note.OwnerRef, out string displayName))
             return note.Content;
 
         var contentElem = new XElement("Contents");
--- a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs
@@ -1837,11 +1837,10 @@ public class ParatextSyncRunner : IParatextSyncRunner
                 {
                     ParatextUserProfile existingUser = _projectDoc.Data.ParatextUsers.SingleOrDefault(u =>
                         u.Username == activePtSyncUser.Username
-                        || (!string.IsNullOrEmpty(u.SFUserId) && string.Equals(u.SFUserId, activePtSyncUser.SFUserId))
                     );
                     if (existingUser == null)
                         op.Add(pd => pd.ParatextUsers, activePtSyncUser);
-                    else if (existingUser.SFUserId == null)
+                    else if (string.IsNullOrWhiteSpace(existingUser.SFUserId))
                     {
                         int index = _projectDoc.Data.ParatextUsers.FindIndex(u =>
                             u.Username == activePtSyncUser.Username
@@ -1850,10 +1849,13 @@ public class ParatextSyncRunner : IParatextSyncRunner
                         if (!string.IsNullOrEmpty(userId))
                             op.Set(pd => pd.ParatextUsers[index].SFUserId, userId);
                     }
-                    else if (!string.Equals(existingUser.Username, activePtSyncUser.Username))
+                    else if (_currentPtSyncUsers.Count(u => u.Value.SFUserId == existingUser.SFUserId) > 1)
                     {
-                        int index = _projectDoc.Data.ParatextUsers.FindIndex(u => u.SFUserId == existingUser.SFUserId);
-                        op.Set(pd => pd.ParatextUsers[index].Username, activePtSyncUser.Username);
+                        // username changed update old username with empty string
+                        int index = _projectDoc.Data.ParatextUsers.FindIndex(u =>
+                            u.Username == activePtSyncUser.Username
+                        );
+                        op.Set(pd => pd.ParatextUsers[index].SFUserId, string.Empty);
                     }
                 }
             });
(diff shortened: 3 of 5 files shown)


RaymondLuong3 (changes requested):
I was able to get my project into the bad state by changing the PT name for a user in my project and using that user to create a note on the project. Then I synchronized the project and could see 2 users in `SFProject.ParatextUsers` with the same `SFUserId`.  When I switched to this branch, and synchronized the project again, it fails since 2 users now exist with the same `SFUserId`. I've suggested some changes that should get this to satisfy the requirements for SF-2911.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2710#-:-O6N-Z1r8ofI9DMxddax:bmfv29x)* status: all files reviewed, 3 unresolved discussions (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 1838 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2710#-O6MiKYi6H6iB-UXjtcy:-O6MiKYi6H6iB-UXjtcz:bwvpioi) ([raw file](https://github.com/sillsdev/web-xforge/blob/eb8afeb028331f47f86dbcbff8f67345bdf0550a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L1838)):*
> ```Smalltalk
>                 foreach (ParatextUserProfile activePtSyncUser in _currentPtSyncUsers.Values)
>                 {
>                     ParatextUserProfile existingUser = _projectDoc.Data.ParatextUsers.SingleOrDefault(u =>
> ```

I don't think that this change will fix the problem described in SF-2911. The problem is that a project can get into a bad state if a user changes their name. The result is that 2 `ParatextProjectUsers` will have the same `SFUserId`. What I expect to see is that we create a new `ParatextProjectUser`, we check if the `SFUserId` is already in use. If so, we 'transfer' the id to the new ParatextProjectUser and assume the old user is no longer in use.

To find out the current PT users for a project based on the data in the registry, the `_paratextUser` property contains that info. So when a user's name is changed, you can distinguish between the old and new based on the record in the registry.

Instead of updating the username of the old `ParatextUserProfile`, you will want to create a new record and update the old record's `SFUserId` with the empty string.



_Code quote:_
```Smalltalk
                    ParatextUserProfile existingUser = _projectDoc.Data.ParatextUsers.SingleOrDefault(u =>
```
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 1840 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2710#-O6MgfmUESxXqUKb-rRY:-O6MgfmUESxXqUKb-rRZ:b-8m4x3h) ([raw file](https://github.com/sillsdev/web-xforge/blob/eb8afeb028331f47f86dbcbff8f67345bdf0550a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L1840)):*
> ```Smalltalk
>                     ParatextUserProfile existingUser = _projectDoc.Data.ParatextUsers.SingleOrDefault(u =>
>                         u.Username == activePtSyncUser.Username
>                         || (!string.IsNullOrEmpty(u.SFUserId) && string.Equals(u.SFUserId, activePtSyncUser.SFUserId))
> ```

Any reason for using `string.Equals` over just the `==` operator? I like the `==` because it is easier to read, but not a big deal.



_Code quote:_
```Smalltalk
|| (!string.IsNullOrEmpty(u.SFUserId) && string.Equals(u.SFUserId, activePtSyncUser.SFUserId))
```

kylebuss (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2710)* status: all files reviewed, 3 unresolved discussions (waiting on @github-advanced-security[bot] and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 1838 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2710#-O6MiKYi6H6iB-UXjtcy:-O6NNjyG1eTEdtoLMD1H:b-9ray1v) ([raw file](https://github.com/sillsdev/web-xforge/blob/eb8afeb028331f47f86dbcbff8f67345bdf0550a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L1838)):*

I don't think that this change will fix the problem described in SF-2911. The problem is that a project can get into a bad state if a user changes their name. The result is that 2 `ParatextProjectUsers` will have the same `SFUserId`. What I expect to see is that we create a new `ParatextProjectUser`, we check if the `SFUserId` is already in use. If so, we 'transfer' the id to the new ParatextProjectUser and assume the old user is no longer in use.

To find out the current PT users for a project based on the data in the registry, the `_paratextUser` property contains that info. So when a user's name is changed, you can distinguish between the old and new based on the record in the registry.

Instead of updating the username of the old `ParatextUserProfile`, you will want to create a new record and update the old record's `SFUserId` with the empty string.
</blockquote></details>

You bring up a good point. I'll do some investigating on this.
___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 1840 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2710#-O6MgfmUESxXqUKb-rRY:-O6NS8HC-ncwlBEhQoy4:b-js9my7) ([raw file](https://github.com/sillsdev/web-xforge/blob/eb8afeb028331f47f86dbcbff8f67345bdf0550a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L1840)):*

Any reason for using `string.Equals` over just the `==` operator? I like the `==` because it is easier to read, but not a big deal.
</blockquote></details>

Apparently, I wrongly assumed a possible `null` string would throw an error when using the `==` operator. I'd say it's mostly out of habit because that was the preferred standard where I worked previously (we seemed to encounter more issues comparing strings if we used `==` rather than `string.Equals`).
___
*[`test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs` line 616 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2710#gh-1747694168:-O6N3Vfz-e-VPRHQGGaL:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/7a74f76a977193ca794013b78fccdaa7ffa346f5/test/SIL.XForge.Scripture.Tests/Services/ParatextSyncRunnerTests.cs#L616)):*
<details><summary><i>Previously, github-advanced-security[bot] wrote…</i></summary><blockquote>

## Useless assignment to local variable

This assignment to [project](1) is useless, since its value is never read.

[Show more details](https://github.com/sillsdev/web-xforge/security/code-scanning/855)
</blockquote></details>

Done.

RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2710)* status: all files reviewed, 2 unresolved discussions (waiting on @github-advanced-security[bot] and @kylebuss)

___
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 1838 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2710#-O6MiKYi6H6iB-UXjtcy:-O6XYopp321oN2LhDi0Y:b-ez1wuj) ([raw file](https://github.com/sillsdev/web-xforge/blob/eb8afeb028331f47f86dbcbff8f67345bdf0550a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L1838)):*

You bring up a good point. I'll do some investigating on this.
</blockquote></details>

I think this solution of updating the ParatextUsers entry is probably a necessary thing to do, but we need to also consider that when a user changes their Paratext Username, we need to decide if we want the user's previous `opaqueUserId` to stay the same on old notes. We probably do want it to stay the same, and still link to the same user.

Look at the `FindOrCreateParatextUser` method in `ParatextService`. I think you will need to update how this method matches a `ParatextUserProfile`. The key should now be the `SFUserId` if it exists, and default to the username if the `SFUserId` does not exist. The way it currently works is it assumes the `Username` is the key,

RaymondLuong3:
Where would the ParatextUserName property get stored? We probably don't want that name stored on the note, but you probably mean storing the property on the user. How is that different from what we already have in the ParatextUserProfile?
I am looking at the code for the logic on how to display the Paratext User of a note, and it looks like it will try to show the sync user's display name and fall back to the sync user's paratext username which is what I expected. Can you describe the problem since I cannot see it?

kylebuss:
The issue was that the note/user did not display correctly when opening the note dialog if the `sfUserId` was empty and the users name had changed.

For example:
- Translator has a note on Genesis 1:1 with Paratext Username = Wilma Jones
- Translator changes name in Paratext registry to Wilma Franklin
- SF has users for both Paratext Usernames and only Wilma Franklin has an SFUserId
- We update and sync the project with Paratext.
- Viewing the notes in SF that were made with Wilma Jones as the username no longer display the message or user data.

RaymondLuong3:
`    return this.userService.currentUserId === syncUserProfile?.id
      ? translate('checking.me') // "Me", i.e. the current user
      : (syncUserProfile?.data?.displayName ?? syncUser.username); // Another user, or fallback to the sync user`

From what I can tell, it will try to get the sync user's PT username since it will not be able to get the profile for the user without the sfUserId.

What does it show instead? I wouldn't worry too much about trying to display the old name if it cannot. I'm happy if it just tries to display a name, or fallback to unknown author.

kylebuss:
To clarify, the entire note is missing not just the username. We have the icon showing there is a note but when the dialog opens you just see the "Your Comment" text area field.

RaymondLuong3:
Ok, then there is an issue here. My preference is to find a way to show the note without the username at all. We are trying to support the situation where the user name changed without dedicating too much code to it since there is no urgent need for it to work well.

kylebuss:
I'll take a look into front-end issue further. I think it might have more to do with the previous `userName` having an empty string for the `sfUserId`.

kylebuss:
*[`src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs` line 1838 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2710#-O6MiKYi6H6iB-UXjtcy:-O6wZvms3-FVPsS57T63:b-okvscg) ([raw file](https://github.com/sillsdev/web-xforge/blob/eb8afeb028331f47f86dbcbff8f67345bdf0550a/src/SIL.XForge.Scripture/Services/ParatextSyncRunner.cs#L1838)):*

I think this solution of updating the ParatextUsers entry is probably a necessary thing to do, but we need to also consider that when a user changes their Paratext Username, we need to decide if we want the user's previous `opaqueUserId` to stay the same on old notes. We probably do want it to stay the same, and still link to the same user.

Look at the `FindOrCreateParatextUser` method in `ParatextService`. I think you will need to update how this method matches a `ParatextUserProfile`. The key should now be the `SFUserId` if it exists, and default to the username if the `SFUserId` does not exist. The way it currently works is it assumes the `Username` is the key,
</blockquote></details>

I realized that we want to use `.Any()` here rather than expect only a single result to be returned. The reason for this is that before the projects `paratextUsers` is synced/saved to db we have two records with `sfUserIds` that is needed to update any existing notes made by the old username before the name change when we run the sync.

We are still clearing the `sfUserId` for the old username when we call `CompleteSync()` to avoid the creation of duplicate notes.

kylebuss (commented):
Changing to WIP as I've identified an issue when changing username, adding a note, and then syncing with paratext.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2710#-:-O6wbhhA8KTy8o6iCZcS:b-m421sw)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @github-advanced-security[bot] and @RaymondLuong3)

kylebuss:
@RaymondLuong3 ,

I tested the following scenario out which resulted in the note being saved.

1) Changed name in Paratext Registry
2) Log in and Sync with Paratext
3) Create new note
4) Sync with Paratext

Without syncing first, the `CommentManager` is not returning the notes created under the new username. Aside from adding a check when the user logs in to alert them to "sync before continuing" if their Paratext Username has changed, I'm not sure there is anything more we can do on this. It seems to be on the Paratext side with handling the username change in the `CommentManager`.

kylebuss (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2710)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @github-advanced-security[bot] and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2988 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2710#gh-1761946895:-O79KQ_r93pmWP1mZkPM:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/281ed6500677288a810da2db987ea59bcfbb2664/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2988)):*
<details><summary><i>Previously, github-advanced-security[bot] wrote…</i></summary><blockquote>

## Useless assignment to local variable

This assignment to [syncUser](1) is useless, since its value is never read.

[Show more details](https://github.com/sillsdev/web-xforge/security/code-scanning/856)
</blockquote></details>

Done.

RaymondLuong3:
@kylebuss Thanks for your efforts on this problem. I am happy for you to put it aside for the time being and step back into the sunshine. I will pick up the task and see what can be done to make it possible for SF not to crash when a sure changes the PT username, even if it has to be done in a less-than-optimal way

kylebuss:
Should we close this PR since there is a newer one from @RaymondLuong3?

RaymondLuong3:
Yes, let's close this.

PR #2712, opened 2024-09-09
SF-2951 Fix ProgressService not subscribing to chapter updates

josephmyers:
The former may have been a carryover from the previous code. It didn't appear to be behaving as intended, or at least not as valuable as intended, given that changes to the project (e.g. adding a chapter) were the only things being listened to. It certainly wasn't picking up text changes, and it's unlikely we were ever updating progress without refreshing the page (or some other similar action).

This implementation merges the changes$ for all chapters into one subscription. It's still throttled to once a second.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2712)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/progress-service/progress-service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/progress-service/progress-service.spec.ts
@@ -1,8 +1,9 @@
 import { NgZone } from '@angular/core';
-import { fakeAsync, TestBed, tick } from '@angular/core/testing';
+import { discardPeriodicTasks, fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { createTestProjectProfile } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-test-data';
+import { TextData } from 'realtime-server/lib/esm/scriptureforge/models/text-data';
 import { TextInfo } from 'realtime-server/lib/esm/scriptureforge/models/text-info';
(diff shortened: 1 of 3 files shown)


RaymondLuong3 (approved):
Nice work improving the subscription to changes on the texts of the whole book for the progress service. That will be much for useful then listening to the changes to the number of chapters on a book.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2712#-:-O6NVb0d3aEMP14M734O:b-wjpgiu)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/progress-service/progress-service.ts` line 70 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2712#-O6NTNNi6iKZ9H74wGdA:-O6NTNNi6iKZ9H74wGdB:b-k1ggz4) ([raw file](https://github.com/sillsdev/web-xforge/blob/b1670feee85381e7540fdcaf11b3d903d5b6cc1b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/progress-service/progress-service.ts#L70)):*
> ```XML
>       });
> 
>       const chapterObservables = [];
> ```

Nit: add a type here.



_Code quote:_
```XML
      const chapterObservables = [];
```

RaymondLuong3:
I recommend updating the title of the PR to mention the progress service. The title is not clear enough of what the change actually affects.

josephmyers:
Done. Thanks for the review!

PR #2722, opened 2024-09-16
SF-2910 Add script to duplicate text audio docs in mongodb

RaymondLuong3:
The previous doc ids for text audio docs incorrectly used the book number instead of the book ID. This PR updates how we calculate the text audio IDs and provides a script to update the mongoDB with docs with the updated text audio IDs by duplicating the docs.
Currently there are roughly 15 documents on live to update the doc ID for. It is important to update the mongo DB before shipping this code.

Developer testing will be sufficient. Once this gets to QA, regression tests will catch any errors.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2722)

### DIFF
--- a/scripts/db_tools/duplicate-doc.js
+++ b/scripts/db_tools/duplicate-doc.js
@@ -1,0 +1,35 @@
+#!/usr/bin/env node
+
+// This file can be used to duplicate documents in a collection with new IDs. This can be
+// useful for migrating documents to have a new ID format.
+const utils = require('./utils.js');
+const ShareDB = require('sharedb/lib/client');
+const Scr = require('@sillsdev/scripture');
+const OTJson0 = require('ot-json0');
+
+// Edit these settings to specify which docs to duplicate
+const connectionConfig = utils.devConfig;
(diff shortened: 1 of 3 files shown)


Nateowami:
Can this not be run as a migration that would automatically run on startup?

RaymondLuong3:
The realtime server does provide a way for us to migrate documents at start up, but creating new documents is not something that has been done before, only altering existing documents. I am hesitant to add a migration that creates new documents since a mistake could result in more documents being created than we intend.
I don't really want to put much more work into this since changing the IDs is not really meaningful, and this would be a quick one-of run of this script.
What did you have in mind?

kylebuss (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2722#-:-O7ArAZP4L5bkUPPrNdj:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2724, opened 2024-09-18
SF-2805 Only add source tabs if user can access the source

josephmyers:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2724)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.spec.ts
@@ -125,6 +125,16 @@ describe('PermissionsService', () => {
     expect(await env.service.canAccessText(cloneDeep(textDoc) as TextDocId)).toBe(false);
   }));
 
+  it('checks current user doc to determine if user is on project', fakeAsync(async () => {
+    const env = new TestEnvironment();
+
+    expect(await env.service.isUserOnProject('project01')).toBe(true);
+
+    env.setCurrentUser('other');
+
+    expect(await env.service.isUserOnProject('project01')).toBe(false);
+  }));
+
   describe('canSync', () => {
     it('returns false when projectDoc.data is undefined', fakeAsync(() => {
       const env = new TestEnvironment();
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.ts
@@ -36,14 +36,19 @@ export class PermissionsService {
     return role != null && roleCanAccessTranslate(role as SFProjectRole);
   }
 
+  async isUserOnProject(projectId: string): Promise<boolean> {
+    const currentUserDoc = await this.userService.getCurrentUser();
+    const userOnProject: boolean =
+      currentUserDoc?.data?.sites[environment.siteId].projects.includes(projectId) ?? false;
+    return userOnProject;
+  }
+
   async canAccessText(textDocId: TextDocId): Promise<boolean> {
     // Get the project doc, if the user is on that project
     let projectDoc: SFProjectProfileDoc | undefined;
     if (textDocId.projectId != null) {
-      const currentUserDoc = await this.userService.getCurrentUser();
-      const userOnProject: boolean =
-        currentUserDoc?.data?.sites[environment.siteId].projects.includes(textDocId.projectId) ?? false;
-      projectDoc = userOnProject ? await this.projectService.getProfile(textDocId.projectId) : undefined;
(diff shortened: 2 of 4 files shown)


RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2724)* status: all files reviewed, 2 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.ts` line 39 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2724#-O769KamCEPjq36LuY--:-O769KamCEPjq36LuY-0:byqqcex) ([raw file](https://github.com/sillsdev/web-xforge/blob/b277444196c10c8710ec20d11cdba94bf38230b9/src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.ts#L39)):*
> ```XML
>   }
> 
>   async isUserOnProject(projectId: string): Promise<boolean> {
> ```

This could be a private method, but I can see how it would be useful for the future.
Oh, I do see that we are using this. I imagine we are just interested if the user is on the project and not if the user has permission to see the text. Wait, what would happen if the user does not have access to see the text? Would the source panel disappear. Then would this bug be possible again?



_Code quote:_
```XML
  async isUserOnProject(projectId: string): Promise<boolean> {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.ts` line 41 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2724#-O769WKC4B242AlXwaEB:-O769WKC4B242AlXwaEC:b-qcjc0e) ([raw file](https://github.com/sillsdev/web-xforge/blob/b277444196c10c8710ec20d11cdba94bf38230b9/src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.ts#L41)):*
> ```XML
>   async isUserOnProject(projectId: string): Promise<boolean> {
>     const currentUserDoc = await this.userService.getCurrentUser();
>     const userOnProject: boolean =
> ```

Nit: You could just return this without assigning it.



_Code quote:_
```XML
    const userOnProject: boolean =
```

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2724)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.ts` line 39 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2724#-O769KamCEPjq36LuY--:-O76ei_6D8k1FTHM7HAi:b-ptxsxa) ([raw file](https://github.com/sillsdev/web-xforge/blob/b277444196c10c8710ec20d11cdba94bf38230b9/src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.ts#L39)):*

This could be a private method, but I can see how it would be useful for the future.
Oh, I do see that we are using this. I imagine we are just interested if the user is on the project and not if the user has permission to see the text. Wait, what would happen if the user does not have access to see the text? Would the source panel disappear. Then would this bug be possible again?
</blockquote></details>

Is it possible to be on a project and not able to view text? I'm not aware of a way to do this, but maybe I'm missing something?

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2724)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.ts` line 39 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2724#-O769KamCEPjq36LuY--:-O7ACWkoBsG3laRXNZ7O:bdvcuq8) ([raw file](https://github.com/sillsdev/web-xforge/blob/b277444196c10c8710ec20d11cdba94bf38230b9/src/SIL.XForge.Scripture/ClientApp/src/app/core/permissions.service.ts#L39)):*

Is it possible to be on a project and not able to view text? I'm not aware of a way to do this, but maybe I'm missing something?
</blockquote></details>

When a source text does not exist for a book, then the source panel disappears. So I tested this out, but I did not see any issues. It looks like it is working well.

Nateowami:
Your use of the plural "tabs" in the title has me somewhat concerned. Does this mean the entire tab group will not be available? Or does it mean don't show the source tab, but do provide the tab group (so the draft and other things can still be opened)?

josephmyers:
The group is still added, for sure, but I'm seeing now that the unit test names are misleading (not my introduction, but I didn't make it better). They indicate the groups, but they're testing the tabs. I'll make a followup PR. Thanks

PR #2711, opened 2024-09-06
SF-2958 Remove dependency on MediaObserver

RaymondLuong3:
This PR removes the dependency for the MediaObserver from the angular/flex-layout library and replaces it with the BreakpointObserver.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2711)

PR #2726, opened 2024-09-20
Fix checking app cut off on medium-sized viewport

RaymondLuong3:
I noticed that when I moved the screen size close to the medium value (768px) the scripture viewer is cut off. I've updated the hiding of the side bar to get pinned only when the viewport size is large (992px) to better use of the screen on medium sized devices.

Before
checking app cut off before

After
Checking app cut off after

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2726)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
@@ -85,7 +85,7 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
   ) {
     super(noticeService);
     this.subscribe(
-      this.breakpointObserver.observe(this.breakpointService.width('>', Breakpoint.MD)),
+      this.breakpointObserver.observe(this.breakpointService.width('>', Breakpoint.LG)),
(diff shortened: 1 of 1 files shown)


Nateowami:
@RaymondLuong3 The before and after images are identical.

RaymondLuong3:
I was just fixing the images :)

Nateowami:
@RaymondLuong3 This was broken in #2711. Can you review changes in that PR to make sure this wasn't the only place this mistake was made?

RaymondLuong3:
I just double-checked the scenarios that were updated, and this was the only piece I could find that was not working properly. Previously the side bar would get pinned when the viewport size was `lt-lg` and I updated it to be (> Md) which would be the same thing, but clearly it is not.

Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2726#-:-O7FgOP14nR9uSrZHgrt:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2739, opened 2024-09-24
Allow Serval admins to download resources from search page

Nateowami:
Serval admins have full access to projects, so being able to access DBL resources is the least of our concern. If we ever open up the role to more users, we'll have to create a permissions system to limit their access to only certain projects.


RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2739)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2740, opened 2024-09-24
SF-2998 Revert Biblical Terms migration

pmachapman:
A bug was discovered on QA when a user had Biblical Terms enabled for their project and visible, the following error displays if the user has not yet updated SF to the latest version.

This PR fixes this by removing the migration, and restoring the `biblicalTermsEnabled` property in `SFProjectUserConfig` and making it nullable.

The downside of this PR is that users who previously had the Biblical Terms panel open will no longer have it open when the update their Scripture Forge PWA.

**Question**

Should we have Biblical Terms auto appear, like we do for Auto Draft if Biblical Terms is enabled for the project?

**Implementation Notes**

- The `biblicalTermsEnabled` property was restored and marked nullable because older clients will be able to set this value via their older version of the Translation Settings dialog.
- This PR is labelled as testing not required, as it is only testable on QA.
- I have kept the `SFProjectUserConfigMigration7` class and cleared its contents because this migration has already been run on all developer workstations, and deployed to QA. Any future migrations for ``SFProjectUserConfig` should begin with version 8.

**Error**

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2740)

### DIFF
--- a/src/RealtimeServer/scriptureforge/models/sf-project-user-config.ts
+++ b/src/RealtimeServer/scriptureforge/models/sf-project-user-config.ts
@@ -17,6 +17,7 @@ export interface SFProjectUserConfig extends ProjectData {
   selectedBiblicalTermsFilter?: string;
   isTargetTextRight: boolean;
   confidenceThreshold: number;
+  biblicalTermsEnabled?: boolean;
   transliterateBiblicalTerms: boolean;
   translationSuggestionsEnabled: boolean;
   numSuggestions: number;
--- a/src/RealtimeServer/scriptureforge/services/sf-project-user-config-migrations.spec.ts
+++ b/src/RealtimeServer/scriptureforge/services/sf-project-user-config-migrations.spec.ts
@@ -103,107 +103,19 @@ describe('SFProjectUserConfigMigrations', () => {
   });
 
   describe('version 7', () => {
-    it('creates the tab in the source if there are source tabs', async () => {
+    it('does not modify the document', async () => {
       const env = new TestEnvironment(6);
       const conn = env.server.connect();
       await createDoc(conn, SF_PROJECT_USER_CONFIGS_COLLECTION, 'project01:user01', {
-        biblicalTermsEnabled: true,
-        editorTabsOpen: [{ groupId: 'source' }]
+        biblicalTermsEnabled: true
       });
       let userConfigDoc = await fetchDoc(conn, SF_PROJECT_USER_CONFIGS_COLLECTION, 'project01:user01');
-      expect(userConfigDoc.data.editorTabsOpen.length).toBe(1);
-      expect(userConfigDoc.data.biblicalTermsEnabled).toBeDefined();
-
-      await env.server.migrateIfNecessary();
-
-      userConfigDoc = await fetchDoc(conn, SF_PROJECT_USER_CONFIGS_COLLECTION, 'project01:user01');
-      expect(userConfigDoc.data.biblicalTermsEnabled).not.toBeDefined();
-      expect(userConfigDoc.data.editorTabsOpen.length).toBe(2);
-      expect(userConfigDoc.data.editorTabsOpen[1]).toEqual({
-        tabType: 'biblical-terms',
-        groupId: 'source',
-        isSelected: false
-      });
-    });
-
-    it('creates the tab in the target if there are no source tabs', async () => {
-      const env = new TestEnvironment(6);
-      const conn = env.server.connect();
-      await createDoc(conn, SF_PROJECT_USER_CONFIGS_COLLECTION, 'project01:user01', {
-        biblicalTermsEnabled: true,
-        editorTabsOpen: []
-      });
-      let userConfigDoc = await fetchDoc(conn, SF_PROJECT_USER_CONFIGS_COLLECTION, 'project01:user01');
-      expect(userConfigDoc.data.editorTabsOpen.length).toBe(0);
-      expect(userConfigDoc.data.biblicalTermsEnabled).toBeDefined();
-
-      await env.server.migrateIfNecessary();
-
-      userConfigDoc = await fetchDoc(conn, SF_PROJECT_USER_CONFIGS_COLLECTION, 'project01:user01');
-      expect(userConfigDoc.data.biblicalTermsEnabled).not.toBeDefined();
-      expect(userConfigDoc.data.editorTabsOpen.length).toBe(1);
-      expect(userConfigDoc.data.editorTabsOpen[0]).toEqual({
-        tabType: 'biblical-terms',
-        groupId: 'target',
-        isSelected: false
-      });
-    });
-
-    it('does not create a tab if one already exists', async () => {
-      const env = new TestEnvironment(6);
-      const conn = env.server.connect();
-      await createDoc(conn, SF_PROJECT_USER_CONFIGS_COLLECTION, 'project01:user01', {
-        biblicalTermsEnabled: true,
-        editorTabsOpen: [
-          {
-            tabType: 'biblical-terms',
-            groupId: 'target',
-            isSelected: false
-          }
-        ]
-      });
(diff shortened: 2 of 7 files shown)


nigel-wells (changes requested):
Just one query Peter. 
Just to confirm, this issue only relates to testers on QA and it hasn't made its way onto live yet?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2740#-:-O7bO-3qAkBAnS-5N_Um:byukt9p)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/RealtimeServer/scriptureforge/services/sf-project-user-config-service.ts` line 53 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2740#-O7bNm5DDMah2POjvaVc:-O7bNm5DDMah2POjvaVd:bhewv0b) ([raw file](https://github.com/sillsdev/web-xforge/blob/6b3e4cc02df0e0181dd2ae6488ff2e7c67ec441c/src/RealtimeServer/scriptureforge/services/sf-project-user-config-service.ts#L53)):*
> ```XML
>       },
>       biblicalTermsEnabled: {
>         bsonType: 'bool'
> ```

Should this be `['null', 'bool']`?

josephmyers:
@pmachapman had mentioned to me a front-end migration, instead of just removing the back-end one, where we auto open the BT tab. I think this would be better than having the panel gone with the previous method of opening it gone and the user confused as to if BT has totally disappeared. We would do that in this PR, right?

pmachapman (commented):
> Just to confirm, this issue only relates to testers on QA and it hasn't made its way onto live yet?

Correct. This has not been deployed to live yet, or we would be swamped with errors.

> [@pmachapman](https://github.com/pmachapman) had mentioned to me a front-end migration, instead of just removing the back-end one, where we auto open the BT tab. I think this would be better than having the panel gone with the previous method of opening it gone and the user confused as to if BT has totally disappeared. We would do that in this PR, right?

Yes. I am going to investigate, and discuss with @Nateowami options regarding this.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2740#-:-O7ec0SB81BCwwvNA0-4:bs0f472)* status: all files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

___
*[`src/RealtimeServer/scriptureforge/services/sf-project-user-config-service.ts` line 53 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2740#-O7bNm5DDMah2POjvaVc:-O7ecFpK6LCSH1Z45Ygx:b-2vub79) ([raw file](https://github.com/sillsdev/web-xforge/blob/6b3e4cc02df0e0181dd2ae6488ff2e7c67ec441c/src/RealtimeServer/scriptureforge/services/sf-project-user-config-service.ts#L53)):*

Should this be `['null', 'bool']`?
</blockquote></details>

No, because it will be omitted, not stored as null if the field is set to null.

pmachapman (commented):
> [@pmachapman](https://github.com/pmachapman) had mentioned to me a front-end migration, instead of just removing the back-end one, where we auto open the BT tab. I think this would be better than having the panel gone with the previous method of opening it gone and the user confused as to if BT has totally disappeared. We would do that in this PR, right?

@josephmyers @Nateowami @nigel-wells I have added code to the editor to display the biblical terms tab, if the user had it enabled previously, and then "migrate" that old setting by unsetting it.

I am really open to feedback on how I did this. Another possible way would to be always add a Biblical Terms tab if it is enabled for the project, but that might get annoying if you don't want it to be auto-added.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2740#-:-O7f35Rr2n72WIgjWPgJ:bonfe1j)* status: 5 of 7 files reviewed, 1 unresolved discussion (waiting on @nigel-wells)

Nateowami (approved):

nigel-wells (approved):
This works but it does mean we will have this code sitting in the repo forever? Is there a script we can run to keep an eye on who is in this boat based on the config of the project and users?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2740#-:-O7zo0VP8jYREJKsNANh:b-q1y6pj)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3951 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2740#-O7znt1-9KLxQHGLCbob:-O7znt1045YJr-5WJ8cm:bto8ipf) ([raw file](https://github.com/sillsdev/web-xforge/blob/9a3edb5f7a42555c2b911cd475df6726d9ea4241/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3951)):*
> ```XML
>         env.wait();
> 
>         const tabGroup = env.component.tabState.getTabGroup('source');
> ```

nit  rename to `sourceTabGroup` for consistency
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3994 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2740#-O7znm2sDs5xXkwXMi0p:-O7znm2sDs5xXkwXMi0q:b-1soumq) ([raw file](https://github.com/sillsdev/web-xforge/blob/9a3edb5f7a42555c2b911cd475df6726d9ea4241/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3994)):*
> ```XML
>       }));
> 
>       it('should not add the biblical terms tab when if the user had biblical terms disabled', fakeAsync(() => {
> ```

nit  remove "when"

pmachapman (commented):
I think we could remove this code when we remove the `BiblicalTermsEnabled` property I marked with the `[Obsolete]` attribute. I'm not sure what our policy is for the period of time before removing obsolete attributes, but I imagine it may be after a year or so?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2740#-:-O7zqSTAA0e2jZBNeTWy:be3or88)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3951 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2740#-O7znt1-9KLxQHGLCbob:-O7zqvyVCtul91GuSiSH:boyexb3) ([raw file](https://github.com/sillsdev/web-xforge/blob/9a3edb5f7a42555c2b911cd475df6726d9ea4241/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3951)):*

nit  rename to `sourceTabGroup` for consistency
</blockquote></details>

Done. Thanks!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts` line 3994 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2740#-O7znm2sDs5xXkwXMi0p:-O7zqxgpBv4pSZCgCrJL:b-ioxi27) ([raw file](https://github.com/sillsdev/web-xforge/blob/9a3edb5f7a42555c2b911cd475df6726d9ea4241/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.spec.ts#L3994)):*

nit  remove "when"
</blockquote></details>

Done. Thank you!

nigel-wells (approved):
I'm not sure if we have a policy - they probably just remain until someone asks the question :)

We can discuss that later. Merge it in.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2740#-:-O7zsRTkB2H-n75wpj2c:b-91ghzb)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2746, opened 2024-09-25
Add tracking of components using notice service

Nateowami:
This is intended to help us track down components that never call `loadingFinish`, or that call `loadingFinished` without first calling `loadingStarted`

This change makes it quite clear that `CollaboratorsComponent` is currently broken, as it consistently calls `loadingFinished` without a corresponding `loadingStarted` call.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2746)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/notice.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/notice.service.ts
@@ -4,13 +4,12 @@ import { firstValueFrom } from 'rxjs';
 import { I18nService } from './i18n.service';
 
 /** Manages and provides access to notices shown to user on the web site. */
-@Injectable({
-  providedIn: 'root'
-})
+@Injectable({ providedIn: 'root' })
 export class NoticeService {
   private _isAppLoading: boolean = false;
-  private loadingCount: number = 0;
   private messageOnDisplay?: string;
+
+  private loadingCountsByCallerId: { [callerId: string]: number } = {};
 
   constructor(
     private readonly snackBar: MatSnackBar,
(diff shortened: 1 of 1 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2746)* status: all files reviewed, 1 unresolved discussion (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/notice.service.ts` line 89 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2746#-O7iNNML8PNZoyjEdwG0:-O7ojyHE4BK7NDE2Vddg:b2sfj16) ([raw file](https://github.com/sillsdev/web-xforge/blob/5ad33d2b3a7a28fdc3236147fb8a7979795d6c5c/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/notice.service.ts#L89)):*

I think this is fine, but I wonder if there is another way we can attempt to get the caller class name here?

Perhaps it's never going to be an issue (Error stack is guaranteed to find a match), but I could see this having the potential to log `unknown` in the console and not give much insight where the call came from.

Probably for a different pull request but it could be useful to have an optional parameter on the `loadingStarted` and `loadingFinished` functions and only use the `getCallerClassName` as a backup.
</blockquote></details>

I don't think that's at all likely to happen, and if it does, we can address it. There wont' be any impact on the user.

kylebuss (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2746)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2756, opened 2024-09-30
SF-3009 Fix error reading projectRef in SFUserProjectsService

Nateowami:
The biggest change is reverting user-projects.service.ts to the state it was before #2621. Everything else is deleting references to what is then removed. Testing not required, because we have no repro steps for the bug.

This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2756)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.spec.ts
@@ -1,7 +1,7 @@
 import { HttpErrorResponse, HttpStatusCode } from '@angular/common/http';
 import { HttpClientTestingModule } from '@angular/common/http/testing';
 import { Component, DebugElement } from '@angular/core';
-import { ComponentFixture, fakeAsync, flush, TestBed, tick } from '@angular/core/testing';
+import { ComponentFixture, TestBed, fakeAsync, flush, tick } from '@angular/core/testing';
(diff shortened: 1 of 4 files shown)


pmachapman (approved):
:lgtm: The code revert looks good to me, however did we isolate the cause of this issue? This is the same bug I documented in SF-3001.

I noticed this bug only happens to me if I open SF for the first time in a little while, and the doc ops are streaming in. This makes me think this bug occurs because the `async` method `updateProjectList` is called without an `await`, and then while it is running, it is being triggered one or more times by `remoteChange$` (also without an await). This is just me surmising though, as I have not properly isolated the issue.

I will leave it to you whether you want to merge this, or if you want to (or you want me to) investigate the root bug further?



> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2756#-:-O83auJK8U_IaScEHJEN:bluv4ek)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami:
@pmachapman I'm quite confident this will fix the error, since we're deleting the code that causes it.

I think the cause of the bug was an incorrect assumption that if you have subscribed to a doc, and you got back a reference to the doc, that it guarantees that the doc data has loaded. The type on the document seems to refute this idea (it's nullable), and experience refutes this idea. Perhaps it would be helpful for methods like ProjectService.getProject to not resolve until the project data has actually loaded, so that the caller can safely ignore the possibility of the data not having loaded yet.

PR #2760, opened 2024-10-02
SF-2973 Switch draft generation to horizontal stepper

josephmyers:
Note that this is just the work to implement the horizontal stepper. It does not include any work for the individual pages, other than the big titles.

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2760)


Nateowami:
Thanks for your work on this. A few things I notice:

### Next & back buttons

I like that you got rid of the back button above the stepper. The buttons for the steps seem a little inconsistent now. Maybe you did this intentionally since the first "back" button isn't really like the others, since it leaves the stepper? Personally I don't think that distinction means as much for users, and it would be ideal for them to all have next/previous icons. The last step really is different though (hence the different button text), and I think it would make sense to add the `auto_awesome` icon to it to indicate this will actually generate a draft.

First step | Middle steps | Last step
-----------|--------------|----------
 |  | 

### Tab focus state

I'm not sure why this is, but when you click next, the focus state gets set on the stepper itself. I would expect this if the user actually clicked the step, but it's weird to click on something at the bottom of the page, and have a focus set on a different element at the top.

*Actually, upon further review, this is already the case on master; it's just not as noticeable. If you want to fix this great; but it doesn't have to be part of this PR since it's not actually a new issue*



### Optional steps

I don't think any of the steps should be labeled optional.

- It's weird that in order to start a draft, I *have* to go to the last step, even though it's supposedly "optional".
- More importantly, none of the steps are actually optional; they just have defaults that are acceptable. Even when source and target languages are both NLLB languages, the user still has to select the training data. "No training data" is a valid selection, but selecting no data is not the same as not selecting data. We do not want anyone to ever skip any step.

pmachapman (changes requested):
The code looks good. I agree with Nathaniel's comments on optional steps, and the next/back buttons.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2760#-:-O8Df3Di1R7CACvdXgOd:b-l9dmp)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (commented):
Thanks. Yes, I actually left the hover state because it was in master to begin with, but I also think it's kind of silly.

I can remove the optional labels, but I disagree that none of the steps are optional. I mean, we do have a flag called "isTrainingOptional," and the Additional Training and Advanced steps can clearly be skipped without input. To the user, that's optional.

The purpose optionality serves for us in this case is help explain why some steps can be skipped (and why they look different if incomplete). Have you seen this scenario?


If the user doesn't have training data where it's required, that step will be colored slightly differently than an optional step like Advanced. It'd be confusing if we don't indicate why they're different colors. And I do think the color distinction is useful. What are your thoughts? Or we could do away with this altogether and force them to visit each step, even if no input is required. This is different than what's on Live.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2760#-:-O8EkolP0rNKlKEAekBF:b-tze8yl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (commented):
Here's Live looking weird to me on multiple levels:


Why are the icons pencils? Why is there is a 2? Why is it a different color? Hopefully I'm addressing most of these.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2760#-:-O8EzmEu9sXZ11MQmGqR:bzd7xjm)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

Nateowami:
@josephmyers All the steps should look the same, and all the steps should have to be visited, in order. That can be fixed in another PR.

pmachapman (approved):
:lgtm: Works well in my testing. I'll approve, on the condition that @Nateowami is happy with the buttons.

> Reviewed 3 of 3 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2760#-:-O8cSMBZ9izFebvySad3:bthbiyw)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami:
I'm OK with it as-is.

josephmyers:
@Nateowami I removed the initial back arrow, but I didn't touch the Start button. We could do that when we rework that page. Actually I see now that you originally recommended we add icons to everything, where I just removed the outlying icon. Since you're okay with things as is, let's just move this on.

Nateowami:
:+1:

PR #2769, opened 2024-10-04
Deprecate ProjectRights.roleHasRight

Nateowami:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2769)


Nateowami:
@RaymondLuong3 Nothing has to be changed (they all work as expected, I think), but if we ever make more permissions capable of being customized on a per-project basis, each place it's used will have incorrect behavior.

`hasRight` takes a project, and looks at settings on the project to make its determination. `roleHasRight` is just checking whether a right is associated with a role (without considering whether the project may have customized the rights). We don't allow customizing most rights, so in most cases there is no difference. But we may want to in the future (this is actually likely).

PR #2783, opened 2024-10-08
Touching up the layout of SF-2986

josephmyers:
Created for visibility.

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2783)


Nateowami:
I like the layout, but I'm not convinced removing the description text is a good idea.

josephmyers:
Yeah I'm not hard-set against them. If we want descriptions, you could easily put some back in. I think it makes less sense to describe the individual pieces, since they're not configuring any of the pieces, and more sense to describe the two main components, i.e. "make sure the overall picture looks good." I had two descriptions for this I had made a while back, but you don't have to use them, of course:
image

Nateowami:
@josephmyers I'm closing this because I've integrated it into my branch (which required resolving conflicts, and deleting styles that were no longer meaningful after you made this change)

PR #2791, opened 2024-10-09
Make diagnostics overlay persist between page loads

Nateowami:
I think there are a number of things we're likely to add, so I've taken cues from the Chromium dev tools to try to shrink things down while remaining readable. 

It also remembers whether it's open, and whether it's collapsed, between page loads.

Before | After
-------|------
 |

PR #2792, opened 2024-10-09
Add subscriber and query counts to diagnostics overlay

Nateowami:
This is showing signs of memory leaks worse than I'd even guessed.

A known issue is that it triggers a scrollbar that breaks the layout. However, that will be fixed when #2791 is merged.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2792)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/diagnostic-overlay/diagnostic-overlay.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/diagnostic-overlay/diagnostic-overlay.component.html
@@ -17,14 +17,16 @@
         <thead>
           <tr>
             <th>Collection</th>
-            <th>Count</th>
+            <th>Documents</th>
+            <th>Subscribers</th>
           </tr>
         </thead>
         <tbody>
           @for (docType of docCountsByCollection | keyvalue; track docType.key) {
             <tr>
               <td>{{ docType.key }}</td>
-              <td>{{ docType.value | l10nNumber }}</td>
+              <td>{{ docType.value.docCount | l10nNumber }}</td>
+              <td>{{ docType.value.subscriberCount | l10nNumber }}</td>
             </tr>
           }
         </tbody>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/diagnostic-overlay/diagnostic-overlay.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/diagnostic-overlay/diagnostic-overlay.component.html
@@ -19,14 +19,16 @@
             <th>Collection</th>
             <th>Documents</th>
             <th>Subscribers</th>
+            <th>Queries</th>
           </tr>
         </thead>
         <tbody>
           @for (docType of docCountsByCollection | keyvalue; track docType.key) {
             <tr>
               <td>{{ docType.key }}</td>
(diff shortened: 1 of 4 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2792)* status: 0 of 4 files reviewed, 1 unresolved discussion

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts` line 51 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2792#-O8mQqMq-YWT0wTI241F:-O8mQqMq-YWT0wTI241G:b-k5z6qf) ([raw file](https://github.com/sillsdev/web-xforge/blob/172760667b7a8dd525470e380b3aad2918c1d91e/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts#L51)):*
> ```XML
>   get docsCountByCollection(): { [key: string]: number } {
>     const docCollections: { [key: string]: number } = {};
>     this.docs.forEach((value, key) => {
> ```

This previous logic was buggy. It would:
- Create an empty object
- Return the object
- Add properties to the object over time after returning it

Because the only thing we use the object for is to show it on the page, we still got the behavior we wanted, but it's more of a happy accident than anything else.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2792)* status: 0 of 4 files reviewed, 1 unresolved discussion

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts` line 51 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2792#-O8mQqMq-YWT0wTI241F:-O8mT3fs1gE-m4gM3_Sn:b-7gwxqn) ([raw file](https://github.com/sillsdev/web-xforge/blob/172760667b7a8dd525470e380b3aad2918c1d91e/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/realtime.service.ts#L51)):*

This previous logic was buggy. It would:
- Create an empty object
- Return the object
- Add properties to the object over time after returning it

Because the only thing we use the object for is to show it on the page, we still got the behavior we wanted, but it's more of a happy accident than anything else.
</blockquote></details>

@kylebuss Do you understand this explanation?

Nateowami:
@kylebuss `forEach(fn)` basically says "finish doing whatever you're doing, and everything you have already planned, and then run this function once for each of these items". The functions get put to the back of the queue.

kylebuss (commented):
Thanks, that helps clarify it. The `forEach(fn)` was executing after the object was returned whereas your change executes the `for` loop before the object returns.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2792#-:-O8mezTk8m8Qzk7K8w6E:bka6x7v)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @kylebuss)

kylebuss (approved):
:lgtm:

> Reviewed 1 of 4 files at r1, 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2792#-:-O8mhV5d1mDtVvkFARRh:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2802, opened 2024-10-14
Fix language picker font not being used

Nateowami:
The font picker was not actually being used because Material has a rule that directly applied to `.mat-mdc-menu-item-text`, and our rule only applied it to `.mat-mdc-menu-item` (the parent element).

You can tell what font is being used by going to the "computed" tab of the element inspector and scrolling to the bottom. You can also look at the network tab and filter for fonts, and see that the language picker font wasn't even being loaded.

I also removed ` (W. Kayah Li)` because the text was originally added in https://github.com/sillsdev/web-xforge/pull/594 for font reasons.

Please note that not all locale names are being rendered with the language picker font, because we haven't been updating the language picker font. However, it does have the Western Kayah Li font, so I feel comfortable removing the English.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2802)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.scss
@@ -147,7 +147,7 @@ header {
       visibility: hidden;
     }
 
-    .mat-mdc-menu-item {
+    .mat-mdc-menu-item .mat-mdc-menu-item-text {
       font-family: variables.$languageFonts;
     }
   }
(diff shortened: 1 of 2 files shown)


pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2802#-:-O9BcE1I4RFXMlHlPoKN:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2805, opened 2024-10-14
Add logging and notification of Registry Server clock drift

pmachapman:
This PR adds logging of the approximate time difference (clock drift) between the Scripture Forge server and the Paratext Registry server. When this value gets too large, an exception is reported in Bugsnag.


Nateowami:
I have some concern with this approach that I'd like to discuss when we get a chance.

pmachapman:
> I have some concern with this approach that I'd like to discuss when we get a chance.

@Nateowami I have updated based on your feedback.

PR #2809, opened 2024-10-16
Fix diagnostic overlay position strategy

Nateowami:
If you open the diagnostics panel on master and look at the console, this error is present:

If you use the element inspector to look at the position of the `.cdk-overlay-pane`, it's at position `0, 0` on the page. The only reason the panel shows up in the correct place is because an element within it has `position: fixed;`.

We're not really using any of the built-in positioning logic, and instead have a overlay with zero size in the top left corner, that contains a positioned element.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2809)


kylebuss (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2809#-:-O9KrL4O32T9U-uX3VSS:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2811, opened 2024-10-16
SF-3047 Fix a bug stopping additional training data files from being uploaded to Serval

pmachapman:
Additional Training Data files are not being uploaded to Serval, due to their URL being stored with a query string in MongoDB.

For example:

```
"fileUrl": "/6625d2d7d636353d0686b856/5f67d5137d39f012009c8a5a_668ef52e87faed28d74749c9.csv?t=133651185101585771"
```

This PR resolves that by stripping the query string if present, and updating the tests to check for this condition.

I have marked this as testing not required, as it is a straightforward change, and really only testable by a developer.


Nateowami:
How is the query string being added in the first place?

pmachapman:
> How is the query string being added in the first place?

@Nateowami Both the audio urls and training data urls have a timestamp query string appended to them to stop old versions getting stuck in cache. In this case it is done in `TrainingDataService.SaveTrainingDataAsync()`.

Nateowami:
@pmachapman Makes sense; I'm just surprised we do it with the training data files, since I didn't think we were serving those to the client, so I would think HTTP caching wouldn't be an issue.

pmachapman:
> I'm just surprised we do it with the training data files, since I didn't think we were serving those to the client, so I would think HTTP caching wouldn't be an issue.

I used the same code as Audio expecting a form of CSV file management interface coming in the future. I think one will be coming in the new design in the Settings screen?

RaymondLuong3 (approved):
Nice! I tested this as per the acceptance tests and it works as expected.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2811#-:-O9QiYL_Bwgvj_ETz0vv:buir5kv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2596, opened 2024-07-18
SF-2749 Refactor scrolling to active verse

josephmyers:
- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2596)

PR #2815, opened 2024-10-17
SF-3007 Add audio comments

RaymondLuong3:
This PR allows users to add audio to their comments. This re-uses the TextAndAudio component and replaces the old comment input with this component. Audio can be played back to users in the form of the single button audio player.

Audio comment full screen

Audio comment action on new line

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815)


RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815)* status: 15 of 18 files reviewed, 1 unresolved discussion (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 17 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-O9V5V-U1mS0_kz46kRl:-O9Vqxow2Qrh0gG50Xe2:by8d68b) ([raw file](https://github.com/sillsdev/web-xforge/blob/e409a7cea803f8b1a89d64618bb711021e50ed39/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L17)):*

I can't decide if I like the single button player or if it should use the `app-checking-audio-player`. If we use the single button player and no text is entered I think we should have a default "Listen to the comment" similar what we do for a question.
</blockquote></details>

I like the suggestion for a placeholder text. How about this? 

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.ts` line 198 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-O9V6ucL72R8hV3qXTMA:-O9Vr40i36NPyYvMe8Hs:bojsrp1) ([raw file](https://github.com/sillsdev/web-xforge/blob/e409a7cea803f8b1a89d64618bb711021e50ed39/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.ts#L198)):*

See note above about adding a default comment when no text is present.
</blockquote></details>

Right, I added a placeholder text in the template. We don't want to add data directly to the objects.

kylebuss (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-:-O9VrlS99wK5rR9E8HyK:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

kylebuss (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

josephmyers (changes requested):
Good job adding the base feature. But there are a number of things missing from the wireframe, and we should implement those here, unless you have a specific for not doing so. I'll try to list them all here, but be sure to look there for yourself. Keep in mind the biggest consideration for Community Checking has to be checkers on mobile. Yes, admins on desktop are important, too, but we should be fitting their additional features around the core checking experience.

- For answers, Edit/Delete should be wrapped up into icons and placed on the same row as the avatar and date. For admin buttons, we could include them on that row, too, probably as icons. We could also only use icons on mobile (when space is limited) -- but it'd be simpler to just use icons (with tooltips).
- For comments, the name and date should display on the right (two rows), with Edit/Delete appearing to the left of them. For audio only, we can display all this on the same line as the audio player. When there's text, we should pop all this down to one row, below text and any audio.
- As part of the above work, I also recommend removing the placeholder text. That space is too valuable, and I think it would be confusing.
- Make sure the content is vertically centered. Currently, when the name and date take up two rows and the content is only one, the content is at the top, and it looks a little weird.
- Consider replacing the divider lines with alternating backgrounds. This would help us continue to move toward a more modern UX, if we could get it to look good.
- Any reason for not adding the comment icon from the design?


Also
- How difficult would it be to add a loading symbol while the audio is being uploaded? The interim between submission and display is noticeable, and it looks like a bug until it appears.
- You can't stop the audio once it starts playing. Tapping the stop button does nothing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-:-O9gYoy2AOuuoLOS1OSO:b-wxu594)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3:
Here are the updated images

Comment no actions
Audio comment no actions

Audio comment with actions
Audio comment mobile

Audio comment full screen
Audio comment desktop

RaymondLuong3 (commented):
As for adding a loading symbol when comments are being uploaded, I think we should create a new issue for this since it affects questions, answers and comments. If we add a loading symbol it would make sense to add them in all 3 scenarios.

Thanks for the feedback, I didn't realize the balsamiq mock-up was a larger redesign and just implemented the audio comment part, but I really like the redesign and the update to using icons.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-:-O9kro9eBZWO1CUGecVX:b-jcwsqh)* status: 14 of 22 files reviewed, all discussions resolved (waiting on @kylebuss)

josephmyers:
Man, I am getting a bizarre Auth0 exception on this branch, preventing me from launching. I've confirmed I can use master and other branches, but not this one. I highly doubt this is a problem with your branch, but I can't proceed until I can start the app. Will keep trying today
[exception_snippet.txt](https://github.com/user-attachments/files/17470305/exception_snippet.txt)

josephmyers (changes requested):
Cool, thanks for that turnaround time! I'm going to get a little picky with some of these comments, but it's important for this with how limited the space is.

- The Add Comment button hover is touching the divider

- Did you try removing the dividers in favor of the alternating (very light) background? Did it look bad?
- The Edit and Delete icons should be smaller. They should appear smaller than the audio player
- The Export and Archive buttons on the answers should probably be icons on mobile, at least. Right now, they're on their own row, and it looks a little strange.
- The author and date should be on top of one another, like they are for answers. They should all align on the right, which will help with the clarity of this section.
- Somewhat related, I don't think the author should be styled the exact same as the content itself. We could do better to break up the sameness
- When the text is small, Edit and Delete appear on the same line as it. The only time these two should not be in the bottom left is for audio only.


This is all I have seen so far. Thanks for hanging in there

> Reviewed 1 of 8 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-:-O9mIvfRCVO6IdHAg_s2:bro020x)* status: 15 of 22 files reviewed, 1 unresolved discussion (waiting on @kylebuss and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 58 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-O9mJaa16E1N8gliN6tF:-O9mJaa16E1N8gliN6tG:b-gevblk) ([raw file](https://github.com/sillsdev/web-xforge/blob/b323a580a2b094265f70b3443e08066e466689c8/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L58)):*
> ```HTML
>       }
>       @if (commentFormVisible && activeComment?.dataId === comment.dataId) {
>         <app-checking-comment-form
> ```

Nathaniel made the good point that this component (form) is really similar to the answers form. We should consider reusing this form for the answers. Obviously there will be some minor differences, but we can pass those variables in as parameters. Some things can just align, e.g. "Save comment" can become "Save"

Here are some things I notice that will need to be parameterized:
- Attach verses, enabled/disabled
- Input box height
- Input box default text

I'm open to discussion on this at a conceptual level. There is a risk that there will be enough differences in the future that we overcomplicate things by trying to share this code. However, I lean more toward sharing, as I don't foresee many differences upcoming.

RaymondLuong3 (commented):
Thanks for the feedback. I believe I have addressed most of the comments. I added the alternating background between comments, I'm not entirely sold but it works.
The flag for review and resolve buttons I left as is. Since this is only visible to admins, I didn't want to do too much. Their role is fundamentally different from the edit/delete button and I thought it would warrant being text buttons rather than icon buttons. Plus, there background colour when they are selected doesn't look very good if converted to icon buttons.
I didn't update the author styling. We can do that in a follow up if desired.

I also took the time to update the answer form to use the comment form. I like that we can pull combine the behaviour of the forms into one form. I like the result but we should be careful to do testing to ensure all features continue to work. I'll comment on the Jira card to look for regressions.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-:-OA4fDiw8o57d6gte4em:b-qbsu7e)* status: 9 of 29 files reviewed, 1 unresolved discussion (waiting on @josephmyers and @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 58 at r6](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-O9mJaa16E1N8gliN6tF:-OA4g_5qCGLdccE6eE2f:b-hr7w1g) ([raw file](https://github.com/sillsdev/web-xforge/blob/b323a580a2b094265f70b3443e08066e466689c8/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L58)):*

Nathaniel made the good point that this component (form) is really similar to the answers form. We should consider reusing this form for the answers. Obviously there will be some minor differences, but we can pass those variables in as parameters. Some things can just align, e.g. "Save comment" can become "Save"

Here are some things I notice that will need to be parameterized:
- Attach verses, enabled/disabled
- Input box height
- Input box default text

I'm open to discussion on this at a conceptual level. There is a risk that there will be enough differences in the future that we overcomplicate things by trying to share this code. However, I lean more toward sharing, as I don't foresee many differences upcoming.
</blockquote></details>

Yep, it was definitely not trivial, but in the long term I think combining the forms was worth the effort.

josephmyers (changes requested):
Cool, man, thanks for your work on this. It's getting there! Some general comments:

- We really should be styling the author name for this. Just take the author's font size down a smidge. Try using the same font size as the date, but not changing anything else.
- Author name should be right aligned.
- Rename the "comment form" component + files to not specify comments anymore.
- Sometimes the alternating background doesn't appear. Really weird

Screenshot of mobile as an observer:


> Reviewed 2 of 19 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-:-OAGF2z88qTmF4jRvdaK:b-yzft4)* status: 11 of 29 files reviewed, 11 unresolved discussions (waiting on @kylebuss and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-global-vars.scss` line 15 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGKYvaAxrKqwYvjwhB:-OAGKYvaAxrKqwYvjwhC:b428q7c) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-global-vars.scss#L15)):*
> ```SCSS
> $borderColor: rgba(0, 0, 0, 0.1);
> $notifyUser: $errorColor;
> $shadedBackgroundColor: $borderColor;
> ```

The alternating background should not be the same color as the border. It needs to be lighter, aesthetically and for the theoretical use case where it's used in conjunction with borders. How about alpha 0.05?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html` line 52 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGAA-nEJ5C46aEReDC:-OAGAA-nEJ5C46aEReDD:b-ee1pzs) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html#L52)):*
> ```HTML
>       } @else {
>         <div id="answer-form">
>           <app-checking-comment-form
> ```

I get this when tapping Add Answer (I'm on Dev Tools mobile)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.scss` line 1 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGBxC996JKUWM-ePbQ:-OAGBxC996JKUWM-ePbR:badydu7) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.scss#L1)):*
> ```SCSS
> @use 'src/variables';
> ```

In this file is a border-bottom rule which adds a divider between the answer and any comments. We should remove this. Then the only place that uses dividers will be between answers
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comment-form/checking-comment-form.component.html` line 7 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGAjVkCknWERzW5epK:-OAGAjVkCknWERzW5epL:b4qupvt) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comment-form/checking-comment-form.component.html#L7)):*
> ```HTML
>       appAutofocus
>       [input]="textAndAudioInput"
>       [textLabel]="t('your_comment')"
> ```

Says Your Comment for answers. Needs to be parameterized
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comment-form/checking-comment-form.component.ts` line 75 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGO-9S4Mh990OPNM5m:-OAGO-9S4Mh990OPNM5n:b-wn4mw5) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comment-form/checking-comment-form.component.ts#L75)):*
> ```XML
>     this.selectedText = value?.scriptureText;
>     this.verseRef = value?.verseRef;
>     this.compact = false;
> ```

This property should be an Input, and it should be specified/set in the HTML, not another property's setter
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comment-form/checking-comment-form.component.ts` line 80 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGPg5dD67oLRAVfiyG:-OAGPg5dD67oLRAVfiyH:b-rjdqrl) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comment-form/checking-comment-form.component.ts#L80)):*
> ```XML
> 
>   @Input() set comment(value: Comment | undefined) {
>     this.textAndAudioInput = value;
> ```

It would be much better to be combine the comment and answer Inputs, if possible. This should be, given that they share a type. Could call it "input"

Similarly, the selectedText and verseRef are totally unlinked from their parent property, textSelectionEnabled. When you combine the two Inputs, there should be some check to confirm that selectedText and verseRef are present if the textSelectedEnabled is true.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comment-form/checking-comment-form.component.ts` line 93 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGQXc06VbJqVM-0zo1:-OAGQXc06VbJqVM-0zo2:b-jmeoik) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comment-form/checking-comment-form.component.ts#L93)):*
> ```XML
> 
>   selectScripture(): void {
>     if (this.textsByBookId == null || this._questionDoc?.data == null) return;
> ```

This method should likely be gated behind the textSelectionEnabled you made. Same for the rest of the file
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 8 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGRd-b2rRKqfWAXcbu:-OAGRd-b2rRKqfWAXcbv:b-kyef0n) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L8)):*
> ```HTML
>         (i + 1 < maxCommentsToShow || commentCount === maxCommentsToShow || showAllComments)
>       ) {
>         <div class="comment" [ngClass]="{ 'comment-unread': !hasUserReadComment(comment), shaded: i % 2 !== 0 }">
> ```

This is great, but can we experiment with adding the condition that we only shade when there are >2 comments shown? Specifically shown, since we also have a behavior to collapse when there are many comments
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 20 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGEemS1aWsxKBZpwoJ:-OAGEemS1aWsxKBZpwoK:b-betbbq) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L20)):*
> ```HTML
>           }
>           @if (comment.text != null) {
>             <span class="comment-text">{{ comment.text }}</span>
> ```

At some point since I last looked, when there are both text and audio, text is now put on the next line. They should both be on the same line, but keep in mind that the text can wrap.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss` line 40 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGI-2d8mXcVQWbDYEf:-OAGI-2d8mXcVQWbDYEg:b-xbjib9) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss#L40)):*
> ```SCSS
> 
>     .mat-icon {
>       transform: scale(0.8);
> ```

I recommend not using scale here. It makes it harder to control and know what the actual value is. You should be able to use `line-height` or `font-size` (or both) to set the icon size. I believe the smallest icons we're using are 18px.

Similarly, please make the icons for the answer the same size as the comments. Right now they're different
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss` line 41 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGIjgj5xOfKnTaJryo:-OAGIjgj5xOfKnTaJryp:bufgosh) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss#L41)):*
> ```SCSS
>     .mat-icon {
>       transform: scale(0.8);
>     }
> ```



Another problem with scale is that it makes the layout harder to set. In the pic above, the spacing is all weird, especially for the audio only comment. The spacing between the two icons should be smaller than between them and the audio player. This will be critical in helping users grasp that there's a difference between the icons and the audio.

On a related note, it would be nice, though optional, if the Edit icon for the previous comment were centered over the audio player. That's a little picky, though

josephmyers (changes requested):
Looking really good. Here are some minor comments:

- The icons are very black and bold, and they really catch the eye (mostly Admin view). The default alpha there currently is 0.87 -- can you tweak this down to 0.70 and see what it looks like?
- The author is not fully right aligned for comments. It doesn't match the date below or the answer above.



> Reviewed 6 of 19 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-:-OAQ86JT2JoEVmEBDpb5:b3t8i7e)* status: 13 of 30 files reviewed, 6 unresolved discussions (waiting on @kylebuss and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html` line 52 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGAA-nEJ5C46aEReDC:-OAQGwh72GepJF1X8-4S:buasgh9) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html#L52)):*

This was a tricky one that I tracked down to the breakpointObserver. I couldn't figure out how to prevent it other than by delaying the breakpoint subscriber by a clock cycle.
</blockquote></details>

Just move the whole `.subscribe()` into the constructor. No need for AfterViewInit
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.scss` line 1 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGBxC996JKUWM-ePbQ:-OAPxDlB8-vfHwG_zFYd:b60vtp5) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.scss#L1)):*

Done.
</blockquote></details>

Looks great, thanks
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 8 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGRd-b2rRKqfWAXcbu:-OAQ9tX68o5-SE8VLWhs:bguhoug) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L8)):*

Yes, that is possible. I have updated it accordingly to behave as you described
</blockquote></details>

Thanks for this. I think the problem was that the first comment always looked weird. I've been playing around with `shaded: i % 2 === 0` and I think it looks a lot better. No alternateBackground property needed
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss` line 40 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGI-2d8mXcVQWbDYEf:-OAPzNUI5_9LvJ73RhZ1:bvf9znm) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss#L40)):*

I've added a mixin that will get the icon to appear smaller.
</blockquote></details>

Great idea. Any reason you didn't use 18px? The Add Comment icon you added is 18px, but these are 20px. The former has an applied style from being in a button, but the other icons should match this when possible.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss` line 41 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGIjgj5xOfKnTaJryo:-OAQ3JbKDcP5DRxXB9v1:b-px1ukf) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss#L41)):*

I have increased the spacing between the play button and the action buttons to 8px. Let me know if you think more is needed.
</blockquote></details>

It's good to increase that, thanks. But the core problem is still there, since the buttons that contain the icons are the same size as before. The Edit icon is still closer to the audio than to the Trash icon.

I'm assuming you didn't change the buttons themselves because of the Material guidelines for clickable areas. 40px is the lowest we typically go. I don't like the idea of increasing the Audio<->Edit spacing even more, but we could consider that.

I think you've exposed a small area where the Balsamiq doesn't quite work. Another design to try is to bump Edit/Delete down to the second row always. (This will result in the top row being pretty sparse for audio only, but the icons won't conflict for spacing, and it'll be more uniform.) Let's try this and see if it works
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comment-form/checking-comment-form.component.ts` line 75 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGO-9S4Mh990OPNM5m:-OAQsAh29nNmzzmc8Cev:b4zm1k3) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comment-form/checking-comment-form.component.ts#L75)):*

I've removed this and just styled in from the parent `CheckingCommentComponent`. Unfortunately something isn't working when there is no existing comment, so on new comments the input opens up at a larger height. I have run out of ideas on what to try.
</blockquote></details>

I'm not sure why you removed the property altogether, instead of making it an Input. Maybe I wasn't clear enough.

If making `compact` an Input doesn't work for you, you could just base the style off of the `label` you just made. Setting the `compact` class on the `app-text-and-audio`, like you had before, but something like this:



_Code snippet:_
```XML
[ngClass]="{ compact: label === 'comment' }"
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815)* status: 13 of 30 files reviewed, 6 unresolved discussions (waiting on @josephmyers and @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html` line 52 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGAA-nEJ5C46aEReDC:-OATx1Wb6JpSKEBg26P8:b-7dzf9p) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-answers.component.html#L52)):*

Just move the whole `.subscribe()` into the constructor. No need for AfterViewInit
</blockquote></details>

That's it! Thanks.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 8 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGRd-b2rRKqfWAXcbu:-OATx9eJ6mUJ39niXgyy:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L8)):*

Thanks for this. I think the problem was that the first comment always looked weird. I've been playing around with `shaded: i % 2 === 0` and I think it looks a lot better. No alternateBackground property needed
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss` line 40 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGI-2d8mXcVQWbDYEf:-OATxB0ECwYxuztQCJbT:b-rqky27) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss#L40)):*

Great idea. Any reason you didn't use 18px? The Add Comment icon you added is 18px, but these are 20px. The former has an applied style from being in a button, but the other icons should match this when possible.
</blockquote></details>

I moved it down to 18px and it is starting to look tiny on mobile, but it will be easy to adjust bigger if it is needed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss` line 41 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGIjgj5xOfKnTaJryo:-OATxR9qDcl83zMNkXVz:bptbi9d) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss#L41)):*

It's good to increase that, thanks. But the core problem is still there, since the buttons that contain the icons are the same size as before. The Edit icon is still closer to the audio than to the Trash icon.

I'm assuming you didn't change the buttons themselves because of the Material guidelines for clickable areas. 40px is the lowest we typically go. I don't like the idea of increasing the Audio<->Edit spacing even more, but we could consider that.

I think you've exposed a small area where the Balsamiq doesn't quite work. Another design to try is to bump Edit/Delete down to the second row always. (This will result in the top row being pretty sparse for audio only, but the icons won't conflict for spacing, and it'll be more uniform.) Let's try this and see if it works
</blockquote></details>

I think moving the action buttons onto a new like is the right way to go. All of the delete buttons will be aligned when admins are looking at it. For community checkers, the icon alignment is not as big a deal unless they have left lots of comments in the thread.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comment-form/checking-comment-form.component.ts` line 75 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGO-9S4Mh990OPNM5m:-OATxsE_6bFIRwi6freo:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comment-form/checking-comment-form.component.ts#L75)):*

I'm not sure why you removed the property altogether, instead of making it an Input. Maybe I wasn't clear enough.

If making `compact` an Input doesn't work for you, you could just base the style off of the `label` you just made. Setting the `compact` class on the `app-text-and-audio`, like you had before, but something like this:
</blockquote></details>

Done.

josephmyers (changes requested):
Few more small comments. But, man, this is starting to look really good, especially for community checkers on mobile

> Reviewed 1 of 8 files at r5, 3 of 19 files at r7, 6 of 19 files at r8, 10 of 10 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-:-OAVjtK8-4hfKjArC_zO:bdrbaqi)* status: all files reviewed, 11 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 1244 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVb9JO1nIdkDBeN3yZ:-OAVb9JP221TeBCi8WVf:b-d62sqg) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L1244)):*
> ```XML
>       }
>     }
>     return true;
> ```

This is a little weird to me. If `audio == null`, we return true?

I'm assuming this is to account for audio simply not being present on submission. If that's the case, this method and its usages would be much, much clearer if we extracted the wrapping prerequisite checks, e.g. `audio == null`, out to the callers. In effect, we wouldn't call this unless the submission has audio.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 8 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGRd-b2rRKqfWAXcbu:-OAVEF3n-kdBTeCB7HWO:b3fcsd2) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L8)):*

Done.
</blockquote></details>

Your code says `!==`. It should be `===`, unless you specifically intended this. In my opinion, the first comment looks pretty slick being gray, whereas having the second gray looks off.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 8 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVjIpiAkQuXZFa6uxf:-OAVjIpiAkQuXZFa6uxg:b93n38f) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L8)):*
> ```HTML
>         (i + 1 < maxCommentsToShow || commentCount === maxCommentsToShow || showAllComments)
>       ) {
>         <div class="comment" [ngClass]="{ 'comment-unread': !hasUserReadComment(comment), shaded: i % 2 !== 0 }">
> ```



I am still seeing a tiny bit of flakiness with the alternating background. It's really weird
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 56 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVhO8j11PqCcMf8T6E:-OAVhO8j11PqCcMf8T6F:blq6y8c) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L56)):*
> ```HTML
>       }
>       @if (commentFormVisible && activeComment?.dataId === comment.dataId) {
>         <app-checking-input-form
> ```



We need a tiny bit of top margin/padding for when the textarea pushes the placeholder up. Around 4px. And check the other usages of `app-checking-input-form`, too. Thanks
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss` line 40 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGI-2d8mXcVQWbDYEf:-OAVgSv5AdnN9uYmliZ0:b-g1k4h6) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss#L40)):*

I moved it down to 18px and it is starting to look tiny on mobile, but it will be easy to adjust bigger if it is needed.
</blockquote></details>

Sounds good, thanks for being flexible. I'm fine either way, I just want to be consistent
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss` line 41 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGIjgj5xOfKnTaJryo:-OAVjXkS7rqO5jb1BDhM:b-29cl92) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.scss#L41)):*

I think moving the action buttons onto a new like is the right way to go. All of the delete buttons will be aligned when admins are looking at it. For community checkers, the icon alignment is not as big a deal unless they have left lots of comments in the thread.
</blockquote></details>

Yeah, agreed. This is a good change
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.spec.ts` line 64 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVHWH99RuJI9B58PE3:-OAVHWH99RuJI9B58PE4:b7xsx6e) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.spec.ts#L64)):*
> ```XML
>   it('can record', async () => {
>     const env = new TestEnvironment();
>     env.waitForRecorder(300);
> ```

Is this something we still need?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.html` line 4 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVSXa--jWlLgYMi9le:-OAVSXa--jWlLgYMi9lf:b-z3ekwg) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.html#L4)):*
> ```HTML
>   <form autocomplete="off" appScrollIntoView>
>     <app-text-and-audio
>       #commentInput
> ```

Still a few places in this file whose names mention "comment" when they shouldn't
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.html` line 12 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVgAOoARa7HnpLcPwm:-OAVgAOoARa7HnpLcPwn:bgfcvzy) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.html#L12)):*
> ```HTML
>     ></app-text-and-audio>
>     <div class="form-actions">
>       <div class="flex-row">
> ```



Now, the Record and Attach buttons have a gap between them. I believe it's coming from the flex-row
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.scss` line 7 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVUYnvA_IDrszyKzIj:-OAVUYnvA_IDrszyKzIk:b-yrumkl) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.scss#L7)):*
> ```SCSS
>   align-items: center;
> 
>   .button-actions {
> ```

Unused
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.ts` line 55 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVYPuvEKLraIk_7ogU:-OAVYPuvEKLraIk_7ogV:byssz7l) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.ts#L55)):*
> ```XML
>   isScreenSmall: boolean = false;
>   submittingResponse: boolean = false;
>   verseRef?: VerseRefData;
> ```

Make this private, and use it directly with `this.` in the method that needs it
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.ts` line 56 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVVu06AHQMp1DjpjFk:-OAVVu06AHQMp1DjpjFl:bcvk83u) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.ts#L56)):*
> ```XML
>   submittingResponse: boolean = false;
>   verseRef?: VerseRefData;
>   textAndAudioInput?: { text?: string; audioUrl?: string };
> ```

Please make this private, with a public or protected access method/getter to expose it

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815)* status: 23 of 30 files reviewed, 11 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 1244 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVb9JO1nIdkDBeN3yZ:-OAYcQ6C36SsNUOl4mCO:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L1244)):*

This is a little weird to me. If `audio == null`, we return true?

I'm assuming this is to account for audio simply not being present on submission. If that's the case, this method and its usages would be much, much clearer if we extracted the wrapping prerequisite checks, e.g. `audio == null`, out to the callers. In effect, we wouldn't call this unless the submission has audio.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 8 at r7](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAGRd-b2rRKqfWAXcbu:-OAYcSnNA3xybyQzxvN8:b-n8hsji) ([raw file](https://github.com/sillsdev/web-xforge/blob/8ec1e4138f82d108892c52072e53cb99487d979a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L8)):*

Your code says `!==`. It should be `===`, unless you specifically intended this. In my opinion, the first comment looks pretty slick being gray, whereas having the second gray looks off.
</blockquote></details>

Oops, I see what you mean now. Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 8 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVjIpiAkQuXZFa6uxf:-OAYcVGt4SoAeQqZFb8F:be5ju1r) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L8)):*



I am still seeing a tiny bit of flakiness with the alternating background. It's really weird
</blockquote></details>

I'm not able to reproduce this. Perhaps we can open another issue to investigate this?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 56 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVhO8j11PqCcMf8T6E:-OAYcaVLDrrTy05oXOC6:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L56)):*



We need a tiny bit of top margin/padding for when the textarea pushes the placeholder up. Around 4px. And check the other usages of `app-checking-input-form`, too. Thanks
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.spec.ts` line 64 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVHWH99RuJI9B58PE3:-OAYcjIlDGYo-AQ_4WbJ:blxl9h0) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.spec.ts#L64)):*

Is this something we still need?
</blockquote></details>

It doesn't seem so. Removed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.html` line 4 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVSXa--jWlLgYMi9le:-OAYcdST4UMa5Y9n-xdK:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.html#L4)):*

Still a few places in this file whose names mention "comment" when they shouldn't
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.html` line 12 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVgAOoARa7HnpLcPwm:-OAYceei1KYuioiYjAjP:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.html#L12)):*



Now, the Record and Attach buttons have a gap between them. I believe it's coming from the flex-row
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.scss` line 7 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVUYnvA_IDrszyKzIj:-OAYcfa09HTHBKqgM4ca:bobclp9) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.scss#L7)):*

Unused
</blockquote></details>

Good catch. Removed.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.ts` line 55 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVYPuvEKLraIk_7ogU:-OAYciDDDHV7QyKnyNy5:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.ts#L55)):*

Make this private, and use it directly with `this.` in the method that needs it
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.ts` line 56 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVVu06AHQMp1DjpjFk:-OAYciTTDCb3jF_ckWc4:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.ts#L56)):*

Please make this private, with a public or protected access method/getter to expose it
</blockquote></details>

Done.

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815)* status: 23 of 30 files reviewed, 11 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.spec.ts` line 64 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVHWH99RuJI9B58PE3:-OAYlu6Q9vJ84_l0UMws:b-riksnf) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.spec.ts#L64)):*

It doesn't seem so. Removed.
</blockquote></details>

Oh, seems like this test is flaky on the CI build. Let's exclude it and come back to it.

josephmyers (changes requested):
One minor comment, that doesn't block sending to test. Great job!

> Reviewed 5 of 7 files at r10, 2 of 2 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-:-OAotShJ7YAJeBA17dxK:b87a1l)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 8 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVjIpiAkQuXZFa6uxf:-OAomAF5Dy-9xDdSbVX4:b-o7j5dx) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L8)):*

I'm not able to reproduce this. Perhaps we can open another issue to investigate this?
</blockquote></details>

Sure, just be aware of it. I've seen it a couple times
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html` line 56 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVhO8j11PqCcMf8T6E:-OAonSZ65-dH4M7my5Tj:blei6w5) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-comments/checking-comments.component.html#L56)):*

Done.
</blockquote></details>

At first, I really didn't like this. Hard spacing below items is not a good way to add spacing intended to spread them out. You typically want to use `gap` properties on the container. But in this case, I think it's fine. The spacing here helps everything that can be below the last comment to breath better. It's not super clear, but it does work as is.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.spec.ts` line 64 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVHWH99RuJI9B58PE3:-OAosf812VSrUeFYRaRY:b-sqbkng) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.spec.ts#L64)):*

Oh, seems like this test is flaky on the CI build. Let's exclude it and come back to it.
</blockquote></details>

If these are the two options, I'd prefer we have the 300ms wait and keep it running. It's better to run slowly than not at all
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.html` line 12 at r9](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-OAVgAOoARa7HnpLcPwm:-OAos_WS9QtWbtD5JmWf:b-4so2ez) ([raw file](https://github.com/sillsdev/web-xforge/blob/27677b508d3a8f82e19577fe1c0fe7cfa6288dc6/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.html#L12)):*

Done.
</blockquote></details>

Great, thanks

josephmyers:
@RaymondLuong3 I thought about holding this up until you've rebased, but maybe we can get away without it. You'll want to do this soon though. It'll be cool to see this new code with the new recording code.

josephmyers (commented):
On this branch, I'm seeing the "scroll into view" behavior broken for comments, particularly ones toward the bottom of the screen. This would be a regression, though I'm not quite sure how this caused it. I originally thought you were missing the fix I made a week or two ago, but that doesn't appear to be the case. You may be interested to investigate a little more as to why it's being weird now.

Given that the "scroll into view" works initially -- if you watch closely -- before the splitter resizes and scrolls it offscreen, I think there's a conflict between "scroll into view" and the splitter resize, `calculateScriptureSliderPosition`. I'm not entirely sure why your branch breaks it, but it almost looks like "scroll into view" is being called earlier now. Because we've messed with the component/markdown structure? One option I've found to work reliably is to delay the `ScrollIntoViewDirective` directive, i.e. increase the `setTimeout` delay from 0 to 400. But it'd be far better to fix the root problem.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2815#-:-OAzgja10Sw8f9BqNVq6:b-7w59ri)* status: all files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

RaymondLuong3:
@josephmyers I have rebased these changes onto master. I'm not sure which PR you mentioned for fixing the scroll into view. Was it PR #2596 ? I can sometimes see the comments not displaying completely, but the verse for a question always scrolls into view which is what that PR was supposed to fix. Maybe we could schedule a call to chat about this. Can you test this branch as it currently is to see if it still behaves oddly?

PR #2820, opened 2024-10-23
SF-3052 Add interactive-widget setting to viewport meta tag

josephmyers:
Chrome 108 (Nov '22) added support for this new tag, which resizes the visual viewport to the visible space. It does not appear after brief investigation that Safari supports this, and Firefox has been documented to resize with or without this setting.

Note that if we choose to hide the bottom bar in the future (which is currently the case), it may not be quite necessary to use "resizes-content" as compared to the other accepted values. It may be that, given that bar will go away, the value "overlays-content" may also suffice. However, the default (resizes-visual for Chrome) likely would cause buggy behavior, given that it currently hides the top bar.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2820)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/index.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/index.html
@@ -17,7 +17,7 @@
     <title>Scripture Forge</title>
     <base href="/" />
 
-    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="viewport" content="width=device-width, initial-scale=1, interactive-widget=resizes-content" />
     <link rel="icon" type="image/x-icon" href="favicon.ico" />
     <link rel="apple-touch-icon" href="assets/icons/sf-192x192.png" />
     <link rel="manifest" href="manifest.json" />


josephmyers:
I'm aware that this is not expected to be supported on Safari. My current stance is that improving the experience for some browsers without harming others is better than what is currently in Live.

kylebuss (approved):
Good work Joseph! I think `resizes-content` makes most sense, even down the road. I briefly read up on this and looks like Chrome/Safari on iOS already does this natively and the changes in Chrome 108 was to bring Android up to par. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2820#-:-O9uw6GgEiy3ce1YaHEE:br5k3b1)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers:
Interesting, Kyle, as I came to a different conclusion about Safari. It seems like, based on https://github.com/bramus/viewport-resize-behavior/blob/main/explainer.md#the-visual-viewport, that Safari overlays its keyboard, which would result in bottom-fixed content being hidden. It'd be valuable to test this on an actual device, though, though I don't think it should hold up this improvement. Most of what I was seeing on the net indicated that custom visual viewport calculation would be required to resize bottom content for Safari. Maybe these sources are out of date now?

PR #2828, opened 2024-10-29
SF-3042 Show no-book message for source

marksvc:
When viewing a tab with the source text, while navigated to a text
location that is not in the source, it was showing "Loading...".

This re-works how and when we decide on placeholder messages, with the
result that the source text will say "book not found" rather than
"Loading..." in the situation above.

The TextComponent Input placeholder is now being used to allow
overriding the message shown when there is no content to show.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2828)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-text/checking-text.component.spec.ts
@@ -277,6 +277,12 @@ class TestEnvironment {
             ],
             hasSource: true,
             permissions: {}
+          },
+          {
+            bookNum: 41,
+            chapters: [{ number: 1, lastVerse: 3, isValid: true, permissions: {} }],
+            hasSource: true,
+            permissions: {}
           }
         ]
       })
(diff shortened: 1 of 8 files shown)


marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2828)* status: 4 of 8 files reviewed, 3 unresolved discussions (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 317 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2828#-OAxQ0Cb2rfdXf8WN3QH:-OAyIHn66832lVLK3zic:b-d9n7pb) ([raw file](https://github.com/sillsdev/web-xforge/blob/58bd513ad3a1453b08fc5e3bb3f6544a84a9d726/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L317)):*

I think we should have a message if there was a 'problem' loading the text. Something like "We were unable to load the text, please refresh your browser and if the problem persists contact support"?
</blockquote></details>

I see that I ended up with only one place where loadingState was set to 'problem', which was when `this.isDestroyed`. This is only after `ngOnDestroy()`. I removed the option and setting of 'problem'; the text component should be on its way out if this is happening. Oh, I like the idea of using loadingState  'unloaded' if ngOnDestroy() was called.

What do you think of this next version?

kylebuss (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2828#-:-OB0fH4bDQUa6k000slt:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @marksvc)

kylebuss (approved):
No worries, I was a bit confused but think we are good now!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2828#-:-OB1Eu4p53BDgbTe1u6g:byvbxj7)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @marksvc)

PR #2838, opened 2024-11-05
SF-3069 Fix audio visualization error in Safari iOS

kylebuss:
This PR fixes the error that occurs when recording on iOS Safari version < 17.3. The canvas context `reset()` throws an error that prevents audio from being recorded on older mobile iOS devices in Safari. The `resetTransform()` function is compatible with older mobile iOS devices and resets the canvas to the initial identity matrix.

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.ts
@@ -191,7 +191,7 @@ export class AudioRecorderDialogComponent
       return;
     }
     this.refreshWaveformSub?.unsubscribe();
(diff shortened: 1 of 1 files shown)


RaymondLuong3 (approved):
I am surprised that `reset()` does not have better support on older devices. `resetTransform()` seems to have much better support and it works in my testing. Nice work!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2838#-:-OAxSk_b1O3cSTxlX4UN:bhik6nz)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kylebuss)

kylebuss:
The research indicated that `reset()` is a relatively new function and really only supported on newer browsers. I think the intent is for it to replace `resetTransform()` but either needs older browsers to start supporting it or become obsolete so `resetTransform()` can be depreciated.

PR #2843, opened 2024-11-07
SF-3072 Fix list of books not loading on draft page

Nateowami:
Here's what this branch looks like. Not amazing, but way better than staring at a page that hasn't loaded and shows no sign of life.

As an example of how bad it is on live, I had this page for 1 minute 50 seconds before the books loaded:

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2843)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.html
@@ -51,4 +51,10 @@
       </mat-chip-listbox>
     }
   </div>
+  @if (!loaded) {
+    <div class="loading-message">
+      <mat-spinner [diameter]="15"></mat-spinner>
+      {{ t("loading_message") }}
+    </div>
+  }
 </ng-container>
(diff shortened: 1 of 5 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2843)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.ts` line 32 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2843#-OBBkKboB041fMhggsjV:-OBQqiFPBacsQyjph3kk:b-6omnh1) ([raw file](https://github.com/sillsdev/web-xforge/blob/17d645c9b19c43f8ceb7e36a862a1c59965d5182/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.ts#L32)):*

Nit: I would make this a field that is private with a public getter since we would never want to have another component modify it.
</blockquote></details>

That seems like a lot of extra overhead for something so trivial. I've made it [protected](https://www.typescriptlang.org/docs/handbook/2/classes.html#protected).

RaymondLuong3 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2843#-:-OBYLUdlD3Fd0xwLPZ2W:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2860, opened 2024-11-20
SF-3091 Load text docs in progress service in parallel

Nateowami:
In my testing this makes the draft page books load much more quickly.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2860)


pmachapman (approved):
:lgtm: I haven't marked ready to test, as it still requires an acceptance test in JIRA. Is it possible to test this on a lower spec machine? I think this may be faster than the current method on low end machines, but it would be nice to confirm that there is no degradation of performance.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2860#-:-OC9eWC41f9mpShGB7Ot:b-dvy7ot)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2871, opened 2024-11-26
SF-3078 Prompt user for microphone access before recorder countdown

RaymondLuong3:
This PR prompts the user for microphone access before the countdown begins on the audio recorder dialog. I also removed the dependency on the SubscriptionDisposible class.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2871)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.spec.ts
@@ -6,7 +6,6 @@ import { firstValueFrom } from 'rxjs';
 import { anything, mock, verify, when } from 'ts-mockito';
 import { NAVIGATOR } from 'xforge-common/browser-globals';
 import { DialogService } from 'xforge-common/dialog.service';
-import { I18nService } from 'xforge-common/i18n.service';
 import { MockConsole } from 'xforge-common/mock-console';
 import { UserDoc } from 'xforge-common/models/user-doc';
 import { NoticeService } from 'xforge-common/notice.service';
@@ -30,7 +29,6 @@ import {
 const mockedNoticeService = mock(NoticeService);
 const mockedNavigator = mock(Navigator);
 const mockedDialog = mock(DialogService);
-const mockedI18nService = mock(I18nService);
 const mockedConsole: MockConsole = MockConsole.install();
 
 describe('AudioRecorderDialogComponent', () => {
@@ -44,9 +42,8 @@ describe('AudioRecorderDialogComponent', () => {
     providers: [
       { provide: NoticeService, useMock: mockedNoticeService },
       { provide: NAVIGATOR, useMock: mockedNavigator },
-      { provide: OnlineStatusService, useclass: TestOnlineStatusService },
-      { provide: DialogService, useMock: mockedDialog },
-      { provide: I18nService, useMock: mockedI18nService }
+      { provide: OnlineStatusService, useClass: TestOnlineStatusService },
+      { provide: DialogService, useMock: mockedDialog }
     ]
   }));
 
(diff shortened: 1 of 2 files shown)


kylebuss (changes requested):
Looks good Raymond! One minor UX suggestion.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2871#-:-OChwgSk6GiocwNpffxD:bg70n0h)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.ts` line 99 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2871#-OChuUxYB-2S4J6xiEz0:-OChuUxYB-2S4J6xiEz1:b-1mdhkj) ([raw file](https://github.com/sillsdev/web-xforge/blob/ecd13c26cb990b48de9c4cb5c9fb7736ead1e4a0/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.ts#L99)):*
> ```XML
>       this.navigator.mediaDevices
>         .getUserMedia({ audio: true })
>         .then(this.startCountdown.bind(this), this.errorCallback.bind(this));
> ```

Would it be difficult to move this prompt when the user clicks the button to record before we open the dialog? If the user for some reason chooses not to allow audio access it leaves an empty dialog open on the screen. Anther option would be to automatically close the dialog in that case.



_Code quote:_
```XML
      this.navigator.mediaDevices
        .getUserMedia({ audio: true })
        .then(this.startCountdown.bind(this), this.errorCallback.bind(this));
```

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2871)* status: all files reviewed, 1 unresolved discussion (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.ts` line 99 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2871#-OChuUxYB-2S4J6xiEz0:-OCio6MsBeYINrI69BHr:b-ebwq) ([raw file](https://github.com/sillsdev/web-xforge/blob/ecd13c26cb990b48de9c4cb5c9fb7736ead1e4a0/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.ts#L99)):*

Would it be difficult to move this prompt when the user clicks the button to record before we open the dialog? If the user for some reason chooses not to allow audio access it leaves an empty dialog open on the screen. Anther option would be to automatically close the dialog in that case.
</blockquote></details>

Initially that sounds like a good idea, but then I thought a little more, there are multiple places where a user can open the dialog, and we don't want to make it the responsibility of the caller to detect if the user has permission for the microphone in every place.
I also noticed the blank dialog, but since it is easy to close and not too likely for a user to deny microphone permission, I decided not to put any effort into that since it breaks tests to automatically close the dialog. If you think there is a good reason to improve that behaviour, I can do that.

kylebuss (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2871#-:-OCj6KRu8QV4K4abc1M3:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.ts` line 99 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2871#-OChuUxYB-2S4J6xiEz0:-OCj58FO0tIm94F2cW4l:br0agdo) ([raw file](https://github.com/sillsdev/web-xforge/blob/ecd13c26cb990b48de9c4cb5c9fb7736ead1e4a0/src/SIL.XForge.Scripture/ClientApp/src/app/shared/audio-recorder-dialog/audio-recorder-dialog.component.ts#L99)):*

Initially that sounds like a good idea, but then I thought a little more, there are multiple places where a user can open the dialog, and we don't want to make it the responsibility of the caller to detect if the user has permission for the microphone in every place.
I also noticed the blank dialog, but since it is easy to close and not too likely for a user to deny microphone permission, I decided not to put any effort into that since it breaks tests to automatically close the dialog. If you think there is a good reason to improve that behaviour, I can do that.
</blockquote></details>

That's okay, this sounds reasonable to leave as is.

PR #2889, opened 2024-12-04
Update query for active translate projects

Nateowami:
This query didn't work as I wanted to write it due to the older version of Mongo. It can now be updated since that's no longer a problem.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2889)


josephmyers (approved):
Looks good. I'm seeing this return the same count as previous on Live.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2889#-:-ODJkv_h2NyuYfUiyBqr:br6jzqi)* status: 0 of 1 files reviewed, all discussions resolved

PR #2890, opened 2024-12-05
Remove unused private members from TypeScript files

Nateowami:
These were found by temporarily adding `"noUnusedLocals": true` to `src/SIL.XForge.Scripture/ClientApp/tsconfig.json`.

I did not persist the config change because the FeatureFlagService has private properties that are accessed via `Object.values(this)`, but appear to not be used when only statically analyzed.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2890)


RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2890)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2891, opened 2024-12-08
Fix broken unit tests

pmachapman:
This PR fixes front end unit tests that are not behaving correctly:

 * Time zone in date unit test fails in New Zealand and time zones that are not whole hours (i.e. India)
 * Error in `AppComponent`:
```
ERROR: 'NG0304: 'app-global-notices' is not a known element (used in the 'AppComponent' component template):
1. If 'app-global-notices' is an Angular component, then verify that it is a part of an @NgModule where this component is declared.
2. If 'app-global-notices' is a Web Component then add 'CUSTOM_ELEMENTS_SCHEMA' to the '@NgModule.schemas' of this component to suppress this message.', [[[...]]]
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2891)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.spec.ts
@@ -46,6 +46,7 @@ import { SF_TYPE_REGISTRY } from './core/models/sf-type-registry';
 import { PermissionsService } from './core/permissions.service';
 import { SFProjectService } from './core/sf-project.service';
 import { NavigationComponent } from './navigation/navigation.component';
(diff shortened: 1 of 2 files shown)


Nateowami (approved):
Thanks for catching this @pmachapman.  The regex was a simple mistake on my part, though I realized I likely couldn't think of every combination that might happen. The "not a known element" error happens all the time because it's so easy to forget. If there's a way to make it fail the tests that would really help us remember.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2891#-:-ODg2FvS2AawMxiG_oPE:bti4a8j)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #2787, opened 2024-10-09
SF-3028 Add final confirmation step when generating drafts

josephmyers:
This is once again ready for review. I took the time to refactor the steps component to be a good bit clearer, at the same time fixing a few minor-medium bugs on master. If you're interested in more details about the bugs, feel free to take a look at my (later) commits and commit messages. I've listed the ones I remember here:

- Restoring last draft selections for all training sources
- Only showing a translated book if it's present in at least one source
- Added a warning for when the user has no access to the training source

I did change the book select logic to only select the first training source found when selecting a translated book, as opposed to selecting everything. I did this so as not to presume that the user wants everything included. I found the "select everything" behavior a little annoying, but I understand if we want to keep it.

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2787)

PR #2893, opened 2024-12-09
SF-3108 Do not allow generating drafts in draft wizard when offline

RaymondLuong3:
The draft wizard cannot be opened offline, but if the user goes offline while the draft wizard is open, then the user can proceed and generate a draft that will fail. This PR prevents the user from advancing steps if the internet connection drops and shows a message to the user.

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2893)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.spec.ts
@@ -149,7 +149,7 @@ describe('DraftGenerationStepsComponent', () => {
       );
       when(mockTrainingDataService.queryTrainingDataAsync(anything())).thenResolve(instance(mockTrainingDataQuery));
       when(mockTrainingDataQuery.docs).thenReturn([]);
-      when(mockFeatureFlagService.allowFastTraining).thenReturn(createTestFeatureFlag(true));
+      when(mockFeatureFlagService.allowFastTraining).thenReturn(createTestFeatureFlag(false));
 
       fixture = TestBed.createComponent(DraftGenerationStepsComponent);
       component = fixture.componentInstance;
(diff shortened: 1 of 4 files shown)


Nateowami:
There's a near infinite number of ways the connection can connect and disconnect. I'm not sure it's necessary to prevent going to the next step. Maybe just prevent the final "generate draft" button on the final step?

RaymondLuong3 (commented):
That is a good point. I was thinking that it would be more robust by letter the user know before advancing steps if they are offline, but the stepper seems to work find offline so it is just the last step that needs an internet connection to work. I've made the update you suggested.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2893#-:-ODhBI-jD_wUxclIyg7b:b-67dlzb)* status: 0 of 2 files reviewed, all discussions resolved

Nateowami:
In almost all cases where functionality is not available offline, we opt to inform the user and disable associated buttons. Here are some examples:







I can't think of any good reason to diverge from that in this case. We do have at least a couple cases where we use a toast, such as User Menu > Edit name icon. However, in that case it's because there isn't a good place to put a persistent notice.

Nateowami:
I think just above the generate button would make the most sense.

RaymondLuong3:
That might not be a good idea since it needs to be duplicated on every step, and if we add steps, you'll need to remember to offline message.

Nateowami:
> Unless you meant just on the final step.

That is what I was thinking.

> That also requires putting it on multiple steps since there are 3 possibilities of the final step.

I didn't realize that was still the case. I think that's changing as soon as #2787 is merged. Maybe put this on hold until that's done?

Nateowami:
Or feel free to pick a different temporary solution.

RaymondLuong3:
I think this PR doesn't have to be put on hold. Once #2787 is merged we can update this message.

Nateowami:
Can you at least drop it down below the stepper so it's close to the button it most pertains to? Right now it's rather far away and easy to miss.

RaymondLuong3:
It doesn't work nicely below the stepper. Since the stepper also includes the buttons, the offline message goes below the buttons and is out of view on some pages. If the user scrolls, they will see it.
image

Nateowami:
Isn't it just as easy to lose it above the fold though?

Nateowami:
Maybe just merge? I don't think it's even worth testing.

RaymondLuong3:
I think that is true, the offline message could be out of view above or below the stepper. I do like that the text is closer to the buttons below the stepper. I've updated it accordingly.
Yes, we can merge it without testing.

Nateowami (approved):
:lgtm:

> Reviewed 1 of 2 files at r2, 2 of 3 files at r3, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2893#-:-ODx0VflDVSBWhNLy4SG:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2900, opened 2024-12-10
SF-3068 Improve Biblical Terms Tab when offline

kylebuss:
This PR removes Biblical Terms option from the Editor Tabs when the user is offline. If the tab was previously open it will persist as the terms should load even if the user is offline. Hiding the option from the tab is inline with Drafts and Resources while offline and the previous behavior for Biblical Terms (when showing terms was in the Editor Settings).

Note: There is an underlying issue with how some docs are loaded vs. others with the RealtimeService when offline. I decided not to dig too far into it because I observed similar behavior if the user is offline on the My Projects page and tries to open a project that has not been stored in IndexedDB. We may need to discuss how we want to go about improving the overall offline functionality.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2900)


Nateowami:
As a rule, we try to never hide things, only state that they're not available and why. I think the right option would be to allow opening the tab, but the tab itself says they can't be loaded offline.

kylebuss:
Okay, I'll look into this. The reason I went this route was this is the current behavior for Auto Drafts and Resource when trying to add a new tab. Showing a message to connect to the internet does seem like a better solution as it's more informative.

pmachapman:
Also, Biblical Terms is meant to fully work offline if the Biblical Terms are downloaded to your PC. Perhaps this PR should only check if there are no Biblical Terms available AND the user is offline?

kylebuss (commented):
Updated!

When offline`biblicalTermsAndNotesChanges$` never emits which ends up keeping the loading icon on the screen. Added a check for both online status and biblical terms/notes have loaded.

@pmachapman I didn't get into the weeds with the Realtime Server as this is not the only component that has issues when offline. I'm open to a further discussion on an approach to address how we want to improve offline support.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2900#-:-ODre9es-YpimOjyD0E3:byetjfr)* status: 0 of 5 files reviewed, all discussions resolved

josephmyers:
We have a number of outstanding bugs surrounding offline support (this one, 3084, 3087, 3089, and the reopened 2860). Has something changed architecturally that's caused our offline support to be less stable? I agree with Kyle that it may be valuable to get some of the relevant heads together and on the same page.

Nateowami:
@josephmyers No, I don't believe so. I think we have just been expanding without focusing on offline.

pmachapman (changes requested):
> I'm open to a further discussion on an approach to address how we want to improve offline support.

To get Biblical Terms and Questions working offline again, I think we will need to implement an offline friendly version of `SharedbRealtimeRemoteStore.createQueryAdapter(collection, parameters)`, which is probably out of the scope of this PR.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2900#-:-OEQkH1d5xo6Wt_1R_ub:byx913r)* status: all files reviewed, 1 unresolved discussion (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts` line 246 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2900#-OEQemYL2X0hDIuWmPl-:-OEQemYL2X0hDIuWmPl0:b-h5wlcp) ([raw file](https://github.com/sillsdev/web-xforge/blob/2ff3dbb62886ab5e225373bd6e301c0bb0bc788f/src/SIL.XForge.Scripture/ClientApp/src/app/translate/biblical-terms/biblical-terms.component.ts#L246)):*
> ```XML
>   }
> 
>   get isOnline(): boolean {
> ```

NIT: Please rename to `appOnline` for consistency with other components this same logic is used.



_Code quote:_
```XML
get isOnline(): boolean {
```

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2900#-:-OEV2Cbw3MlA8MGfo4iK:bnfp4nl)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on @kylebuss)

PR #2901, opened 2024-12-11
SF-3026 Force consistent date format for Nepali across platforms

josephmyers:
By default, Chrome on Windows defaults to standard English formatting for Nepali dates, whereas Chrome on Android uses Nepali script and formatting (to an extent). This change supplies a custom date format that all platforms will use. It gets as close to the appropriate date/time format as possible with what's built into the standard API.

The one difference/shortcoming of which I'm aware is that Nepal uses the Vikram Samvat calendar, which is not supported with the 'calendar' option. We could use the 'indian' calendar, but given that Android defaults to Nepali script and Gregorian calendar, it should be acceptable to do so here, as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2901)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts
@@ -70,7 +70,7 @@ const defaultLocale = locales.find(locale => locale.tags.some(canonicalTag => ca
 export class I18nService {
   static readonly locales: Locale[] = locales;
 
-  static dateFormats: { [key: string]: DateFormat } = {
+  static customDateFormats: { [key: string]: DateFormat } = {
     en: { month: 'short' },
     'en-GB': { month: 'short', hour12: true },
     // Chrome formats az dates as en-US. This manual override is the format Firefox uses for az
@@ -80,6 +80,29 @@ export class I18nService {
         s += ` ${I18nService.getHumanReadableTimeZoneOffset('az', d)}`;
       }
       return s;
+    },
+    npi: (d: Date, options) => {
+      const o = {
+        year: 'numeric',
+        month: '2-digit',
+        day: '2-digit',
+        hour: '2-digit',
+        minute: '2-digit',
+        hour12: false,
+        numberingSystem: 'deva',
+        ...options
+      };
+
+      const parts = new Intl.DateTimeFormat('npi', o as any).formatToParts(d);
+
+      // Build custom YYYY-MM-DD, HH:MM format
+      const year = parts.find(p => p.type === 'year')?.value;
+      const month = parts.find(p => p.type === 'month')?.value;
+      const day = parts.find(p => p.type === 'day')?.value;
+      const hour = parts.find(p => p.type === 'hour')?.value;
+      const minute = parts.find(p => p.type === 'minute')?.value;
+
+      return `${year}-${month}-${day}, ${hour}:${minute}`;
     },
     [PseudoLocalization.locale.canonicalTag]: PseudoLocalization.dateFormat
   };
@@ -263,7 +286,7 @@ export class I18nService {
 
   formatDate(date: Date, options: { showTimeZone?: boolean } = {}): string {
     // fall back to en in the event the language code isn't valid
-    const format = I18nService.dateFormats[this.localeCode] || {};
+    const format = I18nService.customDateFormats[this.localeCode] || {};
     return typeof format === 'function'
       ? format(date, options)
       : date.toLocaleString(
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts
@@ -91,9 +91,9 @@ export class I18nService {
         hour12: false,
         numberingSystem: 'deva',
         ...options
-      };
+      } as const;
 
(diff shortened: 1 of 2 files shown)


src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts
```diff
@@ -81,6 +81,29 @@ export class I18nService {
...
+      const o = {
+        year: 'numeric',
+        month: '2-digit',
+        day: '2-digit',
+        hour: '2-digit',
+        minute: '2-digit',
+        hour12: false,
+        numberingSystem: 'deva',
+        ...options
+      };
```
Nateowami, line 94: Change this:
``` suggestion
      } as const;
 ```
So the type will be inferred as this:
``` ts
const o: {
    readonly showTimeZone?: boolean;
    readonly year: "numeric";
    readonly month: "2-digit";
    readonly day: "2-digit";
    readonly hour: "2-digit";
    readonly minute: "2-digit";
    readonly hour12: false;
    readonly numberingSystem: "deva";
}
```
Instead of this:
``` ts
const o: {
    showTimeZone?: boolean;
    year: string;
    month: string;
    day: string;
    hour: string;
    minute: string;
    hour12: boolean;
    numberingSystem: string;
}
```
Then the `as any` isn't needed a couple lines later.

src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts
```diff
@@ -70,7 +70,7 @@ const defaultLocale = locales.find(locale => locale.tags.some(canonicalTag => ca
 export class I18nService {
   static readonly locales: Locale[] = locales;
 
-  static dateFormats: { [key: string]: DateFormat } = {
+  static customDateFormats: { [key: string]: DateFormat } = {
```
Nateowami, line 73: This is a good name change. Thanks.

src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts
josephmyers: Done. This is really interesting! Using an inline object avoids all this, but when you extract to a variable, suddenly it can't infer the type anymore?

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2901)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts
Nateowami: It's not that it can't, it's that it almost always shouldn't, because it would be very inconvenient.

Imagine you wrote this:
``` ts
let o = { year: 'numeric' }; // infers type { year: string; } (type widening)
// If it inferred { year: "numeric"; } the next line would fail:
o.year = '2-digit';
```

Adding `as const` tells it not to widen the type.

What would actually be better than `as const` though would be specifying the type:
```
const o: Intl.DateTimeFormatOptions = {
```

If you get a type wrong after specifying the type, you get a helpful error on the same line as the mistake:
``` ts
const o: Intl.DateTimeFormatOptions = {
  year: 'tumeric', // Type '"tumeric"' is not assignable to type '"numeric" | "2-digit"'. Did you mean '"numeric"'?
  month: '2-digit',
  day: '2-digit',
  hour: '2-digit',
  minute: '2-digit',
  hour12: false,
  numberingSystem: 'deva',
  ...options
};
```

On the other hand, if you use `as const`, the problem isn't nearly as apparent until later, and the error is much less helpful:
``` ts
const o = {
  year: 'tumeric',
  month: '2-digit',
  day: '2-digit',
  hour: '2-digit',
  minute: '2-digit',
  hour12: false,
  numberingSystem: 'deva',
  ...options
} as const;

const parts = new Intl.DateTimeFormat('npi', o).formatToParts(d);
// No overload matches this call.
//   Overload 1 of 2, '(locales?: LocalesArgument, options?: DateTimeFormatOptions): DateTimeFormat', gave the following error.
//     Argument of type '{ readonly showTimeZone?: boolean; readonly year: "tumeric"; readonly month: "2-digit"; readonly day: "2-digit"; readonly hour: "2-digit"; readonly minute: "2-digit"; readonly hour12: false; readonly numberingSystem: "deva"; }' is not assignable to parameter of type 'DateTimeFormatOptions'.
//       Types of property 'year' are incompatible.
//         Type '"tumeric"' is not assignable to type '"numeric" | "2-digit"'. Did you mean '"numeric"'?
//   Overload 2 of 2, '(locales?: string | string[], options?: DateTimeFormatOptions): DateTimeFormat', gave the following error.
//     Argument of type '{ readonly showTimeZone?: boolean; readonly year: "tumeric"; readonly month: "2-digit"; readonly day: "2-digit"; readonly hour: "2-digit"; readonly minute: "2-digit"; readonly hour12: false; readonly numberingSystem: "deva"; }' is not assignable to parameter of type 'DateTimeFormatOptions'.ts(2769)

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2901#-:-OEQGyN-ER54laTYi0Hn:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @josephmyers and @Nateowami)

PR #2903, opened 2024-12-11
SF-3103 Warn user when project does not contain all draft chapters

RaymondLuong3:
When a book is essentially blank on the project selected to apply a draft to, our mongoDB only has the first chapter for text documents. One option is to create the additional text documents and then apply the drafts, but this appears to go against our normal approach to have users format and create books in Paratext.
This PR gives the user a warning if they are attempting to add a draft to a book where chapters from the draft are not all present.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2903)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sfvalidators.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sfvalidators.ts
@@ -16,6 +16,7 @@ export enum CustomValidatorState {
   InvalidProject,
   BookNotFound,
   NoWritePermissions,
+  MissingChapters,
   None
 }
 
@@ -87,6 +88,8 @@ export class SFValidators {
           return { bookNotFound: true };
         case CustomValidatorState.NoWritePermissions:
           return { noWritePermissions: true };
+        case CustomValidatorState.MissingChapters:
+          return { missingChapters: true };
         default:
           return null;
       }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-apply-dialog/draft-apply-dialog.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-apply-dialog/draft-apply-dialog.component.html
@@ -21,7 +21,11 @@
         <div class="target-project-content">
           @if (targetChapters$ | async; as chapters) {
             <app-notice icon="warning" type="warning"
-              >{{ t("project_has_text_in_chapters", { bookName, numChapters: chapters, projectName: project.name }) }}
+              >{{
+                chapters > 1
+                  ? t("project_has_text_in_chapters", { bookName, numChapters: chapters, projectName: project.name })
+                  : t("project_has_text_in_one_chapter", { bookName, projectName: project.name })
+              }}
             </app-notice>
           } @else {
             <app-notice [icon]="'verified'">{{
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-apply-dialog/draft-apply-dialog.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-apply-dialog/draft-apply-dialog.component.html
@@ -22,7 +22,7 @@
           @if (targetChapters$ | async; as chapters) {
             <app-notice icon="warning" type="warning"
               >{{
-                chapters > 1
+                i18n.getPluralRule(chapters) !== "one"
                   ? t("project_has_text_in_chapters", { bookName, numChapters: chapters, projectName: project.name })
                   : t("project_has_text_in_one_chapter", { bookName, projectName: project.name })
               }}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-apply-dialog/draft-apply-dialog.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-apply-dialog/draft-apply-dialog.component.spec.ts
@@ -63,7 +63,7 @@ describe('DraftApplyDialogComponent', () => {
       { provide: I18nService, useMock: mockedI18nService },
       { provide: OnlineStatusService, useClass: TestOnlineStatusService },
       { provide: MatDialogRef, useMock: mockedDialogRef },
-      { provide: MAT_DIALOG_DATA, useValue: { bookNum: 1 } }
+      { provide: MAT_DIALOG_DATA, useValue: { bookNum: 1, chapters: [1, 2] } }
     ]
   }));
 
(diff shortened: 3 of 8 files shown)


pmachapman (changes requested):
The code looks good, just some feedback on how this works.

I created an empty book in Paratext, and this was the warning I got applying the draft:



I then applied the draft, then went to reapply it to the project and got this:



Can we have the singular/plural chapter(s) working correctly? Also, should this replace the book is empty message (or should we make the book is empty warning yellow like this?)

Finally, these warnings are only in the add to another project dialog. Should they be in the add to project dialog too (or maybe merge the two dialogs)? This may need to be a separate issue to this PR.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2903#-:-ODsFlNiC2oCGA-C6ge0:b-4v10ph)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

Nateowami:
> but this appears to go against our normal approach to have users format and create books in Paratext.

That's mostly (or entirely?) because we expect the user to set up formatting ahead of time, since it isn't editable in SF.

In this case we're completely overwriting everything, so what formatting does/doesn't exist is irrelevant. Can you look into the feasibility of just creating all the book content? If we can't easily do it that's one thing, but I don't think there's any reason we shouldn't if we can. Unless there's something I'm not considering.

RaymondLuong3 (commented):
Here is what the dialog should look like when there are missing chapters. The error text in the message box should appear.


> Should they be in the add to project dialog too (or maybe merge the two dialogs)? This may need to be a separate issue to this PR.

Good point. I am willing to reuse the apply to alternate project dialog for this. But maybe @josephmyers has thoughts on that.

> That's mostly (or entirely?) because we expect the user to set up formatting ahead of time, since it isn't editable in SF

That is a good point. I looked into creating text docs while applying a draft and I think this is not a good idea. We would have to guess what permissions to set for these chapters. We could probably just use the book level permission on each chapter, but I feel kind of icky making that decision.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2903#-:-ODxGemFA7uItWoY1uSm:b-osqko0)* status: 2 of 7 files reviewed, all discussions resolved (waiting on @pmachapman)

josephmyers (commented):
I'm always down for UX consistency and code reuse. I like the idea of using the "add to any project" dialog across the board, and when they click "add to project" it would autofill the project field with their current project. We would need to reassess and potentially touch up some of the text (e.g. dialog title) to account for this.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2903#-:-OECH0cABynpjho9d2cc:b-50oyi0)* status: 2 of 7 files reviewed, all discussions resolved (waiting on @pmachapman)

RaymondLuong3 (commented):
I opened up SF-3126 to track this issue.
> Can we have the singular/plural chapter(s) working correctly? Also, should this replace the book is empty message

I have updated the PR to handle singular and plural. Note that this is not perfect because many languages have multiple forms of a noun for different numbers such as 0, 1, 2, few, many. But this is a start.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2903#-:-OEFPXEf0oPwdaJnQt4q:bwdkh65)* status: 2 of 8 files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (changes requested):
> Here is what the dialog should look like when there are missing chapters. The error text in the message box should appear.

I don't get that message - I still get the blue "{book} in {project} is empty" message reported above.

My SFM for Joshua is simply:
```
\id JOS - Paipera Tapu Back Translation


```

I can share this project with you, if you like?

> I have updated the PR to handle singular and plural. Note that this is not perfect because many languages have multiple forms of a noun for different numbers such as 0, 1, 2, few, many. But this is a start.

Cool! This maybe useful if you think 0/1/2/few/many will affect our currently supported languages? https://jsverse.github.io/transloco/docs/plugins/message-format. If 1 or plural is all we need for now, I am happy with how you have done it.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2903#-:-OEFk7kmALR1OI7Mr0Dh:b-vtju70)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 401 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2903#-OEFnSyDACOR-iiRiw7o:-OEFnSyDACOR-iiRiw7p:b9tptww) ([raw file](https://github.com/sillsdev/web-xforge/blob/eb159e21c954b301a3255d7d97a02184f0902545/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L401)):*
> ```XML
>    * Possible values include 'zero', 'one', 'two', 'few', 'many', and 'other'.
>    */
>   getPluralRule(number: number): string {
> ```

NIT: If you change the return type to `Intl.LDMLPluralRule`, it will enforce type checking on any comparisons.



_Code quote:_
```XML
string
```

RaymondLuong3 (commented):
Hmm, regarding the blue message, I wonder if the draft you generated has multiple chapters. If the draft only contains a single chapter, then it would understand that as all chapters are available to in Paipera Tapu Back Translation. Or there might be something else going on. Yes, please add me to that project so I can investigate

> This maybe useful if you think 0/1/2/few/many will affect our currently supported languages?

That is neat. I think supporting that right now is not worth the effort, but good to know in the future!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2903#-:-OEG0H4JA5WvR-Wq6Hqy:boty3xy)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts` line 401 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/2903#-OEFnSyDACOR-iiRiw7o:-OEG10xV4L1bMOd7qNAy:b-v6rzh8) ([raw file](https://github.com/sillsdev/web-xforge/blob/eb159e21c954b301a3255d7d97a02184f0902545/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/i18n.service.ts#L401)):*

NIT: If you change the return type to `Intl.LDMLPluralRule`, it will enforce type checking on any comparisons.
</blockquote></details>

Nice! Thanks.

pmachapman (commented):
@RaymondLuong3 Were you able to recreate the blue "{book} in {project} is empty" message with the PTBT project I added you to?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2903#-:-OEQGTLF4Cl_X-EcExKc:b-4qwytd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3:
@pmachapman Thanks for the reminder. Let me look at that again

RaymondLuong3:
Ok, I see that if I try to generate a draft of Joshua from and train on books from the PT1868 resource, I don't see Joshua available as a draft book option... Maybe there was an update to serval so that empty source books do not get a draft. I suspect that the issue you are seeing is because Joshua cannot be drafted because there is no content if using PT1868 as a source. If you tried generating a draft with a different source and add to the empty Joshua book in PT1868, I wonder what you would see.

RaymondLuong3:
@pmachapman I figured it out. You most likely created a draft using the PTBT project as a target project. Since PTBT only has the empty 1st chapter in Joshua, when you try to apply the draft to PTBT, the dialog gets notified that there is only 1 chapter in Joshua because the original target project has 1 chapter. It seems that we just discovered a related issue. The original issue I was trying to fix was if the alternate target project had only an empty book. I think I can update this PR to handle both cases of the drafting target and the alternate target having an empty book.

pmachapman (approved):
:lgtm:

> Reviewed 9 of 9 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2903#-:-OGaJcjj-ZxqhKsQVlV3:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-apply-dialog/draft-apply-dialog.component.ts` line 140 at r5](https://reviewable.io/reviews/sillsdev/web-xforge/2903#-OGaJ2Ik7TQbjubpsXT0:-OGaJ2Ik7TQbjubpsXT1:b-7lrhu6) ([raw file](https://github.com/sillsdev/web-xforge/blob/90c2506f4b593fc99e2ca02e7fe55139a28857a5/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-apply-dialog/draft-apply-dialog.component.ts#L140)):*
> ```XML
>           setTimeout(() => {
>             // give the project select component a whole cycle to load the projects
>             this.projectSelect.value = this.data.paratextId;
> ```

Hmmm I don't like when we have to resort to `setTimeout`, but I can't find any other way to do it short of rearchitecting the project select control.



_Code quote:_
```XML
          setTimeout(() => {
            // give the project select component a whole cycle to load the projects
            this.projectSelect.value = this.data.paratextId;
```

RaymondLuong3:
Previously I included a commit to merge the apply to a different project dialog with the add to project method. That is being worked in SF-3126. I've removed that commit

PR #2905, opened 2024-12-12
SF-3114 Make chapter audio only hide bottom bar if scripture text is shown

josephmyers:
This fixes an issue where "hide text" projects couldn't navigate questions on mobile, since the bottom bar is now being hidden.

So now, we don't hide the bottom bar if "hide text" is enabled. We will instead only hide the bar if the "hide text" is disabled and the chapter audio is shown. Effectively, this is when "the user opens the chapter audio."

The second case where we hide the bar remains. I did a slight refactoring to make all this clear and readable.

The original reason for the hiding behavior is still intact, that being to free up space on mobile and keep users focused on their intended task.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2905)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.html
@@ -199,7 +199,7 @@
         </div>
       </div>
       @if (projectDoc && totalVisibleQuestions() > 0) {
-        <div id="question-nav" [ngClass]="{ hide: (textHasFocus && isScreenSmall) || showScriptureAudioPlayer }">
+        <div id="question-nav" [ngClass]="{ hide: textboxIsShownMobile || userOpenedChapterAudio }">
           @if (!isQuestionListPermanent) {
             <button mat-button type="button" (click)="setQuestionsOverlayVisibility(true)">
               {{ t("view_questions") }}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts
@@ -252,6 +252,20 @@ export class CheckingComponent extends DataLoadingComponent implements OnInit, A
     return this.projectDoc?.data?.checkingConfig.hideCommunityCheckingText ?? false;
   }
 
+  get userOpenedChapterAudio(): boolean {
+    return this.showScriptureAudioPlayer && !this.hideChapterText;
+  }
+
+  get textboxIsShown(): boolean {
+    const isAnswering = this.answersPanel?.answerInput != null;
+    const isCommenting = this.answersPanel?.allComments?.find(c => c.inputComponent != null) != null;
+    return isAnswering || isCommenting;
+  }
+
+  get textboxIsShownMobile(): boolean {
+    return this.textboxIsShown && this.isScreenSmall;
+  }
+
   get isRightToLeft(): boolean {
     if (this.projectDoc?.data?.isRightToLeft != null) {
       return this.projectDoc.data.isRightToLeft;
@@ -1080,12 +1094,6 @@ export class CheckingComponent extends DataLoadingComponent implements OnInit, A
       // Reset filter, but don't update visible questions yet if navigating
       this.activeQuestionFilter = QuestionFilter.None;
     }
-  }
-
-  get textHasFocus(): boolean {
-    const isAnswering = this.answersPanel?.answerInput != null;
-    const isCommenting = this.answersPanel?.allComments?.find(c => c.inputComponent != null) != null;
-    return isAnswering || isCommenting;
   }
 
   /**


Nateowami:
@josephmyers Please update the PR title to conform to the guidelines in [CONTRIBUTING.md](https://github.com/sillsdev/web-xforge/blob/master/CONTRIBUTING.md).

I know the meaning of the title might seem obvious to you, but very often there are two completely contradictory interpretations that can't be differentiated without knowing what the original problem was.

> Chapter audio only hides bottom bar if scripture text is shown

Upon reading this title, my first thought was that the bottom bar wasn't being hidden in a situation where it should, but it sounds like it's the exact opposite. Sometimes people create PRs that describe the problem (e.g. #1539), but you can't tell whether a title is describing the problem or the solution without context. On the other hand, if you always use imperative mood, there isn't nearly as much room for ambiguity, and it's easier to read a list of commit messages that have the same grammatical structure.

RaymondLuong3 (changes requested):
I agree, the commit message would be better if it describe what was done to the repo. Something like "Fix bottom nav hidden if hide scripture text enabled" describes the problem that was resolved without being ambiguous. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2905#-:-ODvuOte8AbG7di6xFHc:b-y3df1m)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 259 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2905#-ODvtfiaD6ntfKMZtx8t:-ODvtfiaD6ntfKMZtx8u:b-m4d79b) ([raw file](https://github.com/sillsdev/web-xforge/blob/e81ecf76c1ae1f43bff12dc5fd3be6eaca863cca/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L259)):*
> ```XML
>   }
> 
>   get textboxIsShown(): boolean {
> ```

Nit: This could be private.



_Code quote:_
```XML
  get textboxIsShown(): boolean {
```

josephmyers (commented):
To be fair, the commit message does describe what was done to the repo. Nathaniel's complaint is that you could interpret that message differently, I guess as if my checkin were saying that is the problem.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2905#-:-OECJPDY1jKH5eudzRcn:banp1wr)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers)

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2905)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 259 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2905#-ODvtfiaD6ntfKMZtx8t:-OECLUCz8Scjz6I0cXgK:bbm9wp8) ([raw file](https://github.com/sillsdev/web-xforge/blob/e81ecf76c1ae1f43bff12dc5fd3be6eaca863cca/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L259)):*

Nit: This could be private.
</blockquote></details>

You could, but I don't see a need to. The only thing I could think of is maybe API clutter, but we're way beyond that for this class.

Readonly properties like this are almost always valuable (or neutral) to expose. It's the modifiable fields, e.g. simple boolean flags, that can be harmful when exposed. It's impossible for this property to cause a bug (unless of course there's a bug in the logic here that I don't see).

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2905)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 259 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2905#-ODvtfiaD6ntfKMZtx8t:-OEFIB4mCaQraTdQ7xpP:b-pwo0ir) ([raw file](https://github.com/sillsdev/web-xforge/blob/e81ecf76c1ae1f43bff12dc5fd3be6eaca863cca/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L259)):*

You could, but I don't see a need to. The only thing I could think of is maybe API clutter, but we're way beyond that for this class.

Readonly properties like this are almost always valuable (or neutral) to expose. It's the modifiable fields, e.g. simple boolean flags, that can be harmful when exposed. It's impossible for this property to cause a bug (unless of course there's a bug in the logic here that I don't see).
</blockquote></details>

That is a far point. Would it not be possible to assign the getter to be something else like `() => true`? and then the getter would not work? I assume not but typescript lets you do unexpected things sometimes =P

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2905)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts` line 259 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2905#-ODvtfiaD6ntfKMZtx8t:-OEGGR1u1DYA7AEwwnfm:b82kxat) ([raw file](https://github.com/sillsdev/web-xforge/blob/e81ecf76c1ae1f43bff12dc5fd3be6eaca863cca/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking.component.ts#L259)):*

That is a far point. Would it not be possible to assign the getter to be something else like `() => true`? and then the getter would not work? I assume not but typescript lets you do unexpected things sometimes =P
</blockquote></details>

Yes, that would be really weird. Maybe too weird even for TS!

PR #2911, opened 2024-12-16
SF-3070 Refresh the selected verse if the states are wrong

josephmyers:
This fixes a bug where the selected verse would be doubly toggled in some cases, especially after a history restore. The selected verse is tracked via state in the editor with a simple field, and it can become wrong. With the design in place, it is tricky to keep this field in sync.

The core of this comes down to the difficulty in distinguishing between different types of local changes, some coming from the editor and some from elsewhere. Local changes from elsewhere, like a history restore, will overwrite the text doc completely, including clearing its selected verse. To fix the bug, we check if the selection has been cleared and the editor thinks it hasn't, then refresh the editor.

I'm not sure about a unit test here. This is a really niche use case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2911)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -743,7 +743,7 @@ export class TextComponent extends SubscriptionDisposable implements AfterViewIn
     for (const segment of this.viewModel.segments) {
       const range = segment[1];
       const formats = getAttributesAtPosition(this.editor, range.index);
-      if (formats['commenter-selection'] != null && formats['commenter-selection'] === true) {
+      if (formats['commenter-selection'] === true) {
         ret.push(range);
       }
     }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts
@@ -18,7 +18,7 @@ import QuillCursors from 'quill-cursors';
 import { AuthType, getAuthType } from 'realtime-server/lib/esm/common/models/user';
 import { SFProjectRole } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-role';
 import { TextAnchor } from 'realtime-server/lib/esm/scriptureforge/models/text-anchor';
-import { Subject, Subscription, fromEvent, timer } from 'rxjs';
+import { fromEvent, Subject, Subscription, timer } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
 import { LocalPresence, Presence } from 'sharedb/lib/sharedb';
 import tinyColor from 'tinycolor2';
@@ -36,11 +36,11 @@ import { SFProjectService } from '../../core/sf-project.service';
 import { TextDocService } from '../../core/text-doc.service';
 import { MultiCursorViewer } from '../../translate/editor/multi-viewer/multi-viewer.component';
 import {
-  VERSE_REGEX,
   attributeFromMouseEvent,
   getBaseVerse,
   getVerseRefFromSegmentRef,
-  getVerseStrFromSegmentRef
+  getVerseStrFromSegmentRef,
+  VERSE_REGEX
 } from '../utils';
 import { getAttributesAtPosition, registerScripture } from './quill-scripture';
 import { Segment } from './segment';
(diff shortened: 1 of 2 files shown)


pmachapman (changes requested):
Just a couple of nits.

> I'm not sure about a unit test here. This is a really niche use case.

A form of test is important I think (especially for niche/edge cases), so we don't get a regression, BUT... I spent 30 minutes and couldn't create a useful unit test. My only concern is when we move to Quill 2.0 a regression may occur. It should be OK just to document this though in this PR's JIRA ticket.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2911#-:-OEQcXlw7dR9pqo7euwv:b-cd13vx)* status: all files reviewed, 2 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 926 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2911#-OEQMeRtAUs5UL55or1z:-OEQMeRtAUs5UL55or2-:bfsovov) ([raw file](https://github.com/sillsdev/web-xforge/blob/dd6fd6a78285c503c831e26c2a1760d1e43e98b3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L926)):*
> ```XML
>     }
> 
>     if (this.target.commenterSelection.length === 0 && this.commenterSelectedVerseRef != null) {
> ```

NIT: It is good practice in an `if (... && ...)` to put the expression that requires the least computation first (when possible), as JavaScript uses short-circuit evaluation:
```ts
if (this.commenterSelectedVerseRef != null && this.target.commenterSelection.length === 0) {
```



_Code quote:_
```XML
if (this.target.commenterSelection.length === 0 && this.commenterSelectedVerseRef != null) {
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 746 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2911#-OEQMIrB0pa9_IrL4Yzq:-OEQMIrCCfb3Om6a6Qlw:b-u1350g) ([raw file](https://github.com/sillsdev/web-xforge/blob/dd6fd6a78285c503c831e26c2a1760d1e43e98b3/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L746)):*
> ```XML
>       const range = segment[1];
>       const formats = getAttributesAtPosition(this.editor, range.index);
>       if (formats['commenter-selection'] != null && formats['commenter-selection'] === true) {
> ```

NIT: This can be simplified to:
```ts
if (formats['commenter-selection'] === true) {
```



_Code quote:_
```XML
if (formats['commenter-selection'] != null && formats['commenter-selection'] === true) {
```

josephmyers (commented):
Agreed. I looked at it and determined that it was just not worth it. Most of these core/low-level components are complex, oversized, and highly dependent on the view, making unit testing prohibitively difficult. This use case is a good example of how tech debt (via complex components and test environments) negatively impacts development. A 5000-line test file is ridiculous. But I digress

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2911#-:-OERNMhF2jQ1p1x5MlHq:b2psxdj)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts` line 746 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2911#-OEQMIrB0pa9_IrL4Yzq:-OERSYg3BB89KkDRdmIo:bb6un1j) ([raw file](https://github.com/sillsdev/web-xforge/blob/dd6fd6a78285c503c831e26c2a1760d1e43e98b3/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/text.component.ts#L746)):*

NIT: This can be simplified to:
```ts
if (formats['commenter-selection'] === true) {
```
</blockquote></details>

Done
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts` line 926 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2911#-OEQMeRtAUs5UL55or1z:-OERStdA9JAgzjBFNseO:b-z4gcqs) ([raw file](https://github.com/sillsdev/web-xforge/blob/dd6fd6a78285c503c831e26c2a1760d1e43e98b3/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.ts#L926)):*

NIT: It is good practice in an `if (... && ...)` to put the expression that requires the least computation first (when possible), as JavaScript uses short-circuit evaluation:
```ts
if (this.commenterSelectedVerseRef != null && this.target.commenterSelection.length === 0) {
```
</blockquote></details>

Yep, thanks for catching this

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2911#-:-OERsYqH-1p07AVXYDZP:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

PR #2916, opened 2024-12-17
Verify migrations have monotonically increasing version numbers

Nateowami:
I'm fairly certain there's a way to do this at compile time with the type system, but doing a runtime check is much easier to implement.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2916)

### DIFF
--- a/src/RealtimeServer/common/migration.ts
+++ b/src/RealtimeServer/common/migration.ts
@@ -33,3 +33,21 @@ export abstract class DocMigration implements Migration {
     // do nothing
   }
 }
+
+/**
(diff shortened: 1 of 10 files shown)


RaymondLuong3 (approved):
Thanks for adding this. I can confirm that an error appears when I list the migrations and they are not in increasing order. I'll note that this will not catch an issue if a migration class is missing.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2916#-:-OEKmHTR1KBxo6Yhy5Hj:bq5xf1l)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #2922, opened 2024-12-19
Move localization button into the user menu

josephmyers:
This helps clean up the top bar. The localization button won't really be used after setting it once, possibly on the home page, before even seeing this button.

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2922)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.html
@@ -32,23 +32,6 @@
           </span>
         }
         <span class="toolbar-spacer"></span>
-        @if (isAppOnline) {
-          <button mat-icon-button [title]="t('language')" [matMenuTriggerFor]="localeMenu">
-            <mat-icon>translate</mat-icon>
-          </button>
-        }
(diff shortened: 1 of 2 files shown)


nigel-wells:
The code looks fine. The only thing that bugs me is the arrow pointing to the right and then the sub-menu appears to the left - can we hide that or change it to something else?

Nateowami:
Definitely going to want to talk about this one.

josephmyers:
> The code looks fine. The only thing that bugs me is the arrow pointing to the right and then the sub-menu appears to the left - can we hide that or change it to something else?

I absolutely looked into this. It appears that any programmatic fix has been labeled as "won't fix" -- https://github.com/angular/components/issues/8402 (but I didn't look into it too much further)

josephmyers:
There's nothing really holding this up, after meeting with Nathaniel and Nigel; however, it's also been suggested to just wait to do this until we have a more holistic design for the user menu that includes the impending Notification feature (and any other upcoming features that should go there).

Nateowami:
@josephmyers I'm going to go ahead and close this because I'm guessing when we actually do it we may implement it differently.

PR #2927, opened 2024-12-23
SF-3141 Fix selected reference books persisting between steps

kylebuss:
This PR fixes the resource books selected for training not persisting between steps. It adds a variable to store the books that will be used for training.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2927)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.spec.ts
@@ -198,6 +198,7 @@ describe('DraftGenerationStepsComponent', () => {
       fixture.detectChanges();
       component.userSelectedTranslateBooks = [1];
       component.userSelectedTrainingBooks = [2, 3];
+      component.userSelectedSourceTrainingBooks = [2, 3];
       fixture.detectChanges();
       // Go to training books
       component.tryAdvanceStep();
@@ -219,6 +220,7 @@ describe('DraftGenerationStepsComponent', () => {
       const trainingDataFiles: string[] = [];
       const translationBooks = [2];
 
+      component.userSelectedSourceTrainingBooks = trainingBooks;
       component.userSelectedTrainingBooks = trainingBooks;
       component.userSelectedTranslateBooks = translationBooks;
       component.selectedTrainingDataIds = trainingDataFiles;
@@ -249,6 +251,7 @@ describe('DraftGenerationStepsComponent', () => {
       component.userSelectedTrainingBooks = trainingBooks;
       component.userSelectedTranslateBooks = translationBooks;
       component.selectedTrainingDataIds = trainingDataFiles;
+      component.userSelectedSourceTrainingBooks = trainingBooks;
       component['draftSourceProjectIds'] = {
         draftingSourceId: 'sourceProject',
         trainingSourceId: 'sourceProject',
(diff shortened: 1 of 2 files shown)


Nateowami:
This is better, but it still only properly handles a single reference project. If I select a second reference project, it has the same problem as before.

Nateowami:
@kylebuss What if the user goes back to a previous step? Is it really about "enabling fast translating", or is it about going to a different step before generating the draft? Users can go back to previous steps.

kylebuss:
Good point. I've made the additional training source selected books persistent through changing steps.

josephmyers (changes requested):
Thanks for picking this up. Please add a unit test to verify the properties in question after the step changes.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2927#-:-OF4rmYr7p61FuwCLpFV:b-h9pd00)* status: all files reviewed, 2 unresolved discussions (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.ts` line 79 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2927#-OF4XQwT3xoPLHrg9TlS:-OF4XQwT3xoPLHrg9TlT:bp2jwo2) ([raw file](https://github.com/sillsdev/web-xforge/blob/b43d91d0b13bd61ba5d617cf2bd1ac73d8f03204/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.ts#L79)):*
> ```XML
>   userSelectedAdditionalSourceTrainingBooks: number[] = [];
> 
>   refBooksSelectedTrainingBooks: number[] = [];
> ```

What is this variable name supposed to mean? Can we make this any better/clearer?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.ts` line 375 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2927#-OF4_enpA1dxXjjdoPoi:-OF4_enpA1dxXjjdoPoj:b-k5iph3) ([raw file](https://github.com/sillsdev/web-xforge/blob/931e3a42eeabcac7ca76ce4827d394f9f6dc78d8/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.ts#L375)):*
> ```XML
>       newSelectedTrainingBooks.includes(b)
>     );
>     this.userSelectedAdditionalSourceTrainingBooks = this.selectableAdditionalSourceTrainingBooks.filter(b =>
> ```

It's not really your doing, but this class is turning into a massive state-driven mess. Why are we not wanting to update this field now? This is a red flag, from a maintainability standpoint. If we can't update a "training books" variable in the "update training books" method, we've got a problem.

States are not bad in small number, but this file has no less than 32 mindless state variables to maintain (L 60-102). Yikes. Any way we can reduce this will also reduce future bugs.

The root bug here is that the "update training books" method is overwriting the user-selected training books. (The bug has nothing to do with the additional source books, near as I can tell.) This method is called every time the step changes.

From a design perspective, why is the user selection being overwritten whenever the step changes? This is the real problem. The only time we should blow away a user's choice is if it's no longer valid, e.g. if the user goes back and removes a book selected in a future step. Before, it was sufficient for us to simply reset everything on step change; now, that's obviously not good enough. We could be smarter about when we overwrite the user selection, e.g. only when a selected book is invalidated.

Wouldn't it make more sense to only overwrite the user selection initially (done in `setInitialTrainingBooks`) and when a preceding selection invalidates a succeeding one?

To get away from all these states would be a little more work, and I don't want to put that on you. But let's try to fix the states we have and avoid adding in more, if possible.

josephmyers (approved):
Thanks for fixing this up. I do think it'd be useful to have a unit test for this, but in interest of time, I'll approve this. (I'm also reworking some things in this component in my final step branch, so the test would likely change anyway)


:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2927#-:-OFdsnTs7LqtHqJx-Qgy:bl4anhy)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.ts` line 375 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/2927#-OF4_enpA1dxXjjdoPoi:-OFdsiL4E49EW8coUidH:b-wclv0o) ([raw file](https://github.com/sillsdev/web-xforge/blob/931e3a42eeabcac7ca76ce4827d394f9f6dc78d8/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.ts#L375)):*

> The root bug here is that the "update training books" method is overwriting the user-selected training books. (The bug has nothing to do with the additional source books, near as I can tell.) This method is called every time the step changes.

Both the user selected source and user selected alternate source were being overwritten in the method. The state I added was doing the same thing that could be done with `userSelectedSourceTrainingBooks` so I've removed my initial changes and updated the method to correctly update source and additional source books.
</blockquote></details>

Fantastic. Thanks for the second look.

PR #2944, opened 2025-01-12
SF-3113 Fix realtime resource leaks

siltomato:
Some issues were resolved by calling `dispose()` in `ngOnDestroy()` for the realtime query, but the sneakier issue was when queries were assigned to the result of an `await`.  If the component is destroyed after the query is requested but before the promise resolves, then the query does not get cleaned up in `ngOnDstroy()` because `ngOnDstroy()` has already run by that point.

A solution to this is to use the component `DestroyRef` callback to clean up.  This PR creates a helper method `manageQuery()` to do this.  This may need to be added in other places the app, but the current additions fix the queries that show in developer diagnostics.

Additionally, this PR makes `RealtimeQuery.dispose()` idempotent, as it may be called twice: by `ngOnDestroy()` and by the `manageQuery()` method.

**Unresolved Findings**
A potential issue I ran across in the `RealtimeQuery.dispose()` method is that `adapter.ready` is false when the app is offline, so `onRemovedFromSubscribeQuery()` doesn't get called on the docs when the app is offline.  I'm not sure how to handle this.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2944)


pmachapman (changes requested):
Your synopsis of the problem looks correct to me.

Would it make sense to pass the destroyRef to methods like `CheckingQuestionsService.queryQuestions()` and have that method call `manageQuery`? It might make the code a little bit cleaner, and if all methods that called `return this.realtimeService.subscribeQuery(...)` were the ones that executed `manageQuery`, then no one is likely to forget to call `manageQuery` and cause future leaks.

i.e. we change `CheckingQuestionsService.queryQuestions()` (and similar methods) to look like:
```ts
  queryQuestions(
    projectId: string,
    destroyRef: DestroyRef,
    options: { bookNum?: number; chapterNum?: number; activeOnly?: boolean; sort?: boolean } = {}
  ): Promise<RealtimeQuery<QuestionDoc>> {
    const queryParams: QueryParameters = {
      [obj<Question>().pathStr(q => q.projectRef)]: projectId
    };

    ...

    return manageQuery(this.realtimeService.subscribeQuery(QuestionDoc.COLLECTION, queryParams), destroyRef);
  }
```

If you don't think this is practical, I can approve the PR as-is.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2944#-:-OGRrvQK4e9WVga1zoxN:bawl53d)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

siltomato (commented):
I considered that and even having `RealtimeService.subscribeQuery()`, itself, take a `DestroyRef` parameter to handle disposing the subscription.  When I asked myself, "whose responsibility is it to dispose of the query?" it seemed right that it should be the component.  Although, perhaps the component passing a 'destroy' callback to the service satisfies that thought.

Do you think having an optional `DestroyRef` param in `RealtimeService.subscribeQuery()`would be a good way to handle this?  Then any component or service subscribing to a query could pass `DestroyRef` if it wanted the query to be 'managed'.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2944#-:-OGVy02x90tZNVUWTLaW:bwt3r8p)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

pmachapman (approved):
> Do you think having an optional `DestroyRef` param in `RealtimeService.subscribeQuery()`would be a good way to handle this? Then any component or service subscribing to a query could pass `DestroyRef` if it wanted the query to be 'managed'.

I am torn.... your comment on it being the component that calls that disposes the query is architecturally correct.

Looking into it more, from what I understand the DestroyRef should make its way to the component? For example in `RealtimeQuery`, an unsubscribe$ subject is used, so the constructor contains:

```ts
this.adapter.ready$.pipe(takeUntil(this.unsubscribe$)).subscribe(() => this.onReady());
```
when perhaps it should be something along the lines of:

```ts
this.adapter.ready$.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(() => this.onReady());
```
With us passing the DestroyRef to the constructor of this class.

This kind of rework feels out of scope to your PR, however?

I'll approve and leave it up to you, as to whether you want to think on it further, or if we need to discuss this further (perhaps at the meeting this evening?), or whether we press on with your PR as-is, as it does fix the issue.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2944#-:-OGW2Iuh8WtSk-AtKoob:b-jdak62)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

siltomato (commented):
- Moved `manageQuery()` to be a private method on `RealtimeService`
- Added `DestroyRef` param to `RealtimeService.subscribeQuery()`
- Refactored call chain leading to `RealtimeService.subscribeQuery()`
- Adjusted unit tests for additional `DestroyRef` param

I ran into one case where a query is created without a component `DestroyRef` in the `ResumeCheckingService`, as this query subscription is supposed to be created once in the singleton service and not disposed in components that inject it.  I created a 'no-op' `DestroyRef` to use in this case and in unit tests as well.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2944#-:-OGYDvqY7Dgtb1DqIf1x:bsvxwpu)* status: 1 of 28 files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (approved):
:lgtm: This is awesome!!!!

> Reviewed 27 of 27 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2944#-:-OGaKH8Z6Fz1V6cCx_nY:b-s3ng8w)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

PR #2956, opened 2025-01-17
SF-3156 Warn user when mismatch in translated and training books

RaymondLuong3:
The user should see warnings on the training step for two scenarios:

- The user's training source does not contain all of the translated books (books with more than 10 translated segments). This can indicate the user is using a source other than the one they are actually translating from.
- The user selected translated books to use for training, but did not select matching books an any of the reference projects.

Warnings will appear for the user, but they can still proceed to generate a draft.

Draft generation training warning source
Draft generation training warning books

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2956)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.html
@@ -138,6 +138,16 @@
             {{ t("choose_books_for_training_error") }}
           </app-notice>
         }
+        @if (!translatedBooksSelectedInTrainingSources) {
+          <app-notice class="warn-translated-books-unselected" type="warning" icon="warning">
+            {{ t("translated_book_selected_no_training_pair") }}
+          </app-notice>
+        }
+        @if (translatedBooksWithNoSource.length > 0) {
+          <app-notice class="warn-source-books-missing" type="warning" icon="warning">
+            {{ t("source_books_missing") }}
+          </app-notice>
+        }
         <div class="button-strip">
           <button mat-stroked-button matStepperPrevious>{{ t("back") }}</button>
           <button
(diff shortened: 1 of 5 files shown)


kylebuss (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2956#-:-OGq4K7g7JeQf7j4qbWS:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

josephmyers:
Shouldn't we prevent them from proceeding to the next step in this case? I can't think of any good reason why they'd actually want this.

RaymondLuong3:
I could go either way. If we prevented them from proceeding when not all translated books selected have a matching reference book selected to train on, the user might realize they had made a mistake. That would be a positive. So perhaps that is worth doing. For the second scenario when the user has chosen a source without books matching the target project, I think the warning is good enough. If a project decides to train on only the back translation (maybe that will happen) then they would get just the warning rather than forced not to use the BT because its translation is behind the target project.

kylebuss:
What if we remove the book from the selected books list if no reference sources have it selected?

josephmyers:
@kylebuss, currently only source-target pairs are included for training. So I think we're already doing that, unless you're meaning something else by "selected books list"

@RaymondLuong3, yes, good clarification. I'm only recommending we prevent proceeding for the first scenario.

kylebuss:
@josephmyers, I was mainly addressing the second scenario when the target has selected books but not all of those books were selected in the source. Rather than displaying a warning we could remove the missing book from the target list. I hope this clarifies what I'm meaning.

As for the first scenario, there might be an edge case where we want the user to acknowledge before proceeding like we do with the language codes on the first step. One thought that comes to mind is if we down the road allow selected chapters to be include in training rather than the whole book.

I don't know how likely of a scenario it would be for a project to use one source for NT and another for OT (like Hebrew -> English OT and Greek -> English NT), but in that case, if they have translated books in both OT and NT but the source only has OT we would want to allow them to proceed.

Nateowami:
@RaymondLuong3 I thought the wording could be improved and asked @bhartmoore who came up with these suggested changes:

> 1. Draft generation training warning source:
       **current**: "The source for this project does not contain all the translated books. Are you sure you are using the correct source?"
       **suggested**: "The reference text is missing one or more of the translated books in your project. Consider choosing a reference text that contains a match for all the translated books."
(I'm thinking users might not equate "source for this project" and "correct source" with something they've selected in SF settings, displayed on the current page as "reference books." This aims to make that more explicit so users understand what's being suggested.)
> 2. Draft generation training warning books
       **current**: "You selected translated books but did not select matching reference books for training. Please choose matching reference books."
       **suggested**: "Some of the training books you selected have no matching reference books selected. Please select matching reference books."
(Clarifying that the mismatch may only apply to some of the books.)

RaymondLuong3:
Thanks. I've updated the wording on the warning messages. I opened SF-3173 to add the require user to fix warning task. I am going to merge this as-is since it has already gone through testing and wasn't part of the original issue.

PR #2959, opened 2025-01-17
SF-3084 Fix checking overview page not showing questions correctly

RaymondLuong3:
On the overview page, the user was wrongly being shown that questions were loading when they were offline. This is because the questionQuery.ready property never returns true if offline. However, the query results will still load the offline docs. This change adds checking for the online status and only shows the loading message if the user is online.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2959)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.spec.ts
@@ -111,6 +111,16 @@ describe('CheckingOverviewComponent', () => {
       expect(env.noQuestionsLabel).not.toBeNull();
     }));
 
+    it('should not display loading if user is offline', fakeAsync(() => {
+      const env = new TestEnvironment();
+      env.testOnlineStatusService.setIsOnline(false);
+      tick();
+      env.fixture.detectChanges();
+      expect(env.loadingQuestionsLabel).toBeNull();
+      expect(env.noQuestionsLabel).not.toBeNull();
+      env.waitForQuestions();
+    }));
+
     it('should not display "Add question" button for community checker', fakeAsync(() => {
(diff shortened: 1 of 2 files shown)


pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2959#-:-OGzvZ0fCRZiR4MkVggy:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2959)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.ts` line 195 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2959#-OHDbcHf8Xdyo_ljsZ8P:-OHDbcHf8Xdyo_ljsZ8Q:b-w1mw5z) ([raw file](https://github.com/sillsdev/web-xforge/blob/7f5ba6cf0cc447c5bb5ba124271adff0e6bdb63f/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking-overview/checking-overview.component.ts#L195)):*
> the query will still return the offline docs

Is it immediate though? It seems like there ought to be a way to know when the query for offline documents is complete/ready.

RaymondLuong3:
I just did a test for loading question docs offline. I had a project with 700+ questions, and after going offline and navigating to the questions overview page, the docs all loaded immediately. The load time is almost negligible. Since it is only accessing IndexedDB and does not have network latency issues, I would say there is no need to show loading for docs while offline.

Nateowami:
Thanks for that explanation @RaymondLuong3

PR #2960, opened 2025-01-17
SF-3111 Fix question dialog verse reference error message overlap

RaymondLuong3:
The error message in the text input has the `position: absolute` rule applied. So it is expected to show over top of anything that comes below. In this case, this rule makes it hard to know how much space below the input. I've given it the minimum so that it shows nicely. FWIW, I would not mind if we just left the error message overlap as by design.

Before
Question dialog overlap

After
Overlapping text add question


pmachapman (approved):
:lgtm: + bonus points for removing `SubscriptionDisposable`!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2960#-:-OGzwC-m80kDi5QPWg83:bb22f4e)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #2966, opened 2025-01-21
SF-2860 Offline project add question error 

siltomato:
Fixes issue where adding a question with empty xforge indexeddb database would throw an error (see Jira for video capture).

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/question-dialog/question-dialog.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/question-dialog/question-dialog.component.ts
@@ -127,7 +127,7 @@ export class QuestionDialogComponent extends SubscriptionDisposable implements O
         this.scriptureEnd.setValue(endVerseRef.toString());
       }
       this.updateSelection();
-    } else if (this.data.defaultVerse != null) {
+    } else if (this.data.defaultVerse != null && this.data.defaultVerse.bookNum > 0) {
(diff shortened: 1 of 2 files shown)


Nateowami:
Thanks @siltomato

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2966#-:-OH9vHGK-pyl0D2cAJSG:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/question-dialog/question-dialog.service.ts` line 108 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2966#-OH5Ob1f74_zPaCUX6zv:-OH9uvds3FI5gxBGWLXm:b-55ca8p) ([raw file](https://github.com/sillsdev/web-xforge/blob/188fddf1292fc2a605780cbbe9f52d0ee6c535d9/src/SIL.XForge.Scripture/ClientApp/src/app/checking/question-dialog/question-dialog.service.ts#L108)):*

I'm not sure about this change.  This seemed to work for retrieving a project profile doc, where `get(projectId)` to get a project doc would not when offline.  Could this be because the project profile docs were loaded before I went offline on the project connect page?
</blockquote></details>

Yes, that will be it. Usually we load the project profile rather than the project, as you only really need the project doc if you want the paratextUsers property.

PR #2972, opened 2025-01-22
deploy: specify OS version, note a need on bump

marksvc:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2972)


marksvc:
This change doesn't _fix_ anything, but tries to prevent unexpected breaking in the future.

I made this change so we can break the deploy process manually next time rather than unexpectedly.

PR #2975, opened 2025-01-23
Add heap snapshot generation to the realtime server

pmachapman:
This PR adds support for generating heap snapshots of the Realtime Server. This will be useful to diagnose why the Out of Memory errors are occurring every 1-3 days, which result in the Realtime Server restarting.

**To Use:**
1. Get the process id of node: `ps aux | grep node`
2. Send the USR2 signal to the node process: `kill -USR2 {process_id_here}`

If developing, the console output will tell you where it was written, for example:
```
Heap snapshot written to /home/peter/src/web-xforge/src/SIL.XForge.Scripture/bin/Debug/net8.0/scriptureforge.heapsnapshot
```
If on QA or Production, look in the journal to see where the file was written.

**Note:** As we are already using USR2, I was unable to use the Node command line argument [--heapsnapshot-signalsignal](https://nodejs.org/api/cli.html#--heapsnapshot-signalsignal).

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2975)

### DIFF
--- a/src/RealtimeServer/common/diagnostics.ts
+++ b/src/RealtimeServer/common/diagnostics.ts
@@ -1,5 +1,6 @@
 import fs from 'fs';
 import inspector from 'inspector';
+import { writeHeapSnapshot } from 'node:v8';
 import path from 'path';
 
 const secondsToProfile = 30;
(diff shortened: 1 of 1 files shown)


Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2975#-:-OHJSqSGA5XDNhQlOpCU:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

___
*[`src/RealtimeServer/common/diagnostics.ts` line 12 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/2975#-OHJSYe671DMMx105dIr:-OHJSYe70wPF3zyArlpB:barwvxt) ([raw file](https://github.com/sillsdev/web-xforge/blob/0f4e770327e55efbf8e02fb9045238a745509db9/src/RealtimeServer/common/diagnostics.ts#L12)):*
> ```XML
> process.on('SIGUSR2', () => {
>   console.log('Signal SIGUSR2 received; writing heap snapshot');
>   const heapPath = path.join(process.cwd(), 'scriptureforge.heapsnapshot');
> ```

It seems it would be ideal for the file name to include a timestamp so we can take multiple without them overwriting each other, but I'll merge as-is for now.

PR #2999, opened 2025-02-10
SF-3201 Fix broken editor jump links

siltomato:
This PR fixes an error in some resources since upgrading to Quill v2: `[Parchment] Unable to create link blot`.

Quill v1 didn't render the links at all.  These "jump links" seem to be intending to link to a project/book/chapter/verse.  The quick fix is to render the link contents in a span.  In the future, we may try to support these links.

Here is an example value being sent to a 'link' blot:
```json
{
  "style": "jmp",
  "link-href": "prj:TNN.XXB.1.0",
  "contents": {
    "ops": [
      {
        "insert": "Guidelines for Using the Translator's Notes Series"
      }
    ]
  }
}
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2999)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-scripture.ts
@@ -10,6 +10,7 @@ import QuillTextBlot from 'quill/blots/text';
 import QuillClipboard from 'quill/modules/clipboard';
 import QuillHistory, { StackItem } from 'quill/modules/history';
 import { DeltaOperation, StringMap } from 'rich-text';
+import { isString } from '../../../type-utils';
 import { DragAndDrop } from './drag-and-drop';
 import { TextComponent } from './text.component';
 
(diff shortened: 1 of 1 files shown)


Nateowami:
@siltomato Any way to prevent this from being a game of wack-a-mole? It seems like there could be a whole array of things we don't support? See `usfm.sty` for a (possibly outdated) list attributes elements can have. Even if we support all of them, how can we be sure more won't be added in the future? Is there a way we can gracefully handle arbitrary blots? As an example, it looks like cross-references can have links as well.

RaymondLuong3 (commented):
Thanks for the fix for this. I see that this fixes the problem in the issue and that link embeds can be rendered (but not clicked on).
We probably want to add a test to show that these kind of text docs can be loaded into the text component without causing the error.
Perhaps what Nathaniel suggested is worth opening a different card for.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2999#-:-OIm8kDeE4Dz8XPkRNIz:bqjbo6w)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

Nateowami (approved):
:lgtm: We're merging for now. We can address other possible scenarios after getting this fix out.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2999#-:-OIrFrZR0ZayfUT3ITIg:b-cagois)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

PR #3008, opened 2025-02-13
Prevent front-end tests from passing when warnings are printed

Nateowami:
We frequently end up merging tests that pass but cause warnings to be printed. This updates the workflow to try to stop this. Initially, this branch will fail *on purpose* to prove the change works. Then it will be updated to fix the warnings.

Actually, not sure if that strategy will work, because maybe it will run the workflow as it already exists on master.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3008)

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -84,8 +84,14 @@ jobs:
             -p:CollectCoverage=true \
             -e:CoverletOutputFormat=opencover \
             -e:Exclude=\"[NUnit3.TestAdapter]*,[SIL.XForge.*.Views]*,[SIL.XForge.Tests]*\"
+
+      # Run the tests while making sure none of the common/known warnings are printed
       - name: "Test: Frontend"
-        run: cd src/SIL.XForge.Scripture/ClientApp && npm run test:gha
+        run: |
+          set -e
+          cd src/SIL.XForge.Scripture/ClientApp
+          npm run test:gha | tee output.log
+          ! grep NG0303 output.log # Can't bind to 'some_property since it isn't a known property of 'some_element'
 
       - name: "Coverage: Backend"
         run: |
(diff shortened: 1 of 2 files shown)


RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3008)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`.github/workflows/build-and-test.yml` line 94 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3008#-OJ-FJGT4XF3ELDbI_Ty:-OJ-FJGT4XF3ELDbI_Tz:b-qhal4m) ([raw file](https://github.com/sillsdev/web-xforge/blob/7cb4c6c818f19e5a4c6b8183a7ea13a64a5a1e52/.github/workflows/build-and-test.yml#L94)):*
> ```YAML
>           cd src/SIL.XForge.Scripture/ClientApp
>           npm run test:gha | tee output.log
>           ! grep NG0303 output.log # Can't bind to 'some_property since it isn't a known property of 'some_element'
> ```

I like that this fixes an issue in our test that should have been failing. I wonder if we could make this grep more general than the specific NG0303.



_Code quote:_
```YAML
          ! grep NG0303 output.log # Can't bind to 'some_property since it isn't a known property of 'some_element'
```

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3008)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`.github/workflows/build-and-test.yml` line 94 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3008#-OJ-FJGT4XF3ELDbI_Ty:-OJ0HYOiEKDE2r7An2eE:but39yw) ([raw file](https://github.com/sillsdev/web-xforge/blob/7cb4c6c818f19e5a4c6b8183a7ea13a64a5a1e52/.github/workflows/build-and-test.yml#L94)):*

@RaymondLuong3 My idea was that each time we find a different warning we think should result in failure, we add another line to grep `output.log`. I don't want to make it too general.
</blockquote></details>

Ok, so this needs to be updated as we find more issues. That is probably safer than being too general and catching issues that aren't real issues.

PR #3009, opened 2025-02-13
SF-3199 CacheService disposes of older subscriptions

josephmyers:
The CacheService was subscribing to all chapters for each project visited during the lifetime of the app. This was causing memory issues for the server, though this hasn't always been a problem. I'm not exactly sure what changed -- maybe the space for each chapter is growing, or we're getting more concurrent users. This changes the CacheService to uncache its subscriptions, when the project changes, before subscribing to the new chapters. I verified with the Node Dev Tools in Chrome that, while the number of ClientSessions still grows, the memory usage does not grow when switching projects.

I also took the opportunity to refactor away the service's dependence on some other class to tell it when to cache. It now listens to the activated project and cues its own operations. You simply need to reference it somewhere (currently in app.component).

I of course updated the tests and, as part of this, reduced the total number of chapters being used in the tests. I had to reduce this number, because the 20000 mock creations (or something) were crashing the test client. I'm sad to say it couldn't even handle 2000, but I believe the tests are just as effective at 200. I also reworked one of the more interesting tests to be more explicit/dependable.

Screenshot 2025-02-12 150419

Above, you can see the memory usage as I navigate to different projects with the new code (green line) and the old code (orange line), with the first snapshot of each "section" being taken from the My Projects page (no project loaded). For the third "section" I navigated to six different projects before taking that second and final snapshot.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3009)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
@@ -39,6 +39,7 @@ import { SFProjectProfileDoc } from './core/models/sf-project-profile-doc';
 import { roleCanAccessTranslate } from './core/models/sf-project-role-info';
 import { SFProjectUserConfigDoc } from './core/models/sf-project-user-config-doc';
 import { SFProjectService } from './core/sf-project.service';
+import { CacheService } from './shared/cache-service/cache.service';
 import { checkAppAccess } from './shared/utils';
 
 declare function gtag(...args: any): void;
@@ -77,6 +78,7 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
     private readonly fileService: FileService,
     private readonly reportingService: ErrorReportingService,
     private readonly activatedProjectService: ActivatedProjectService,
+    private readonly cacheService: CacheService,
     private readonly locationService: LocationService,
     private readonly breakpointObserver: BreakpointObserver,
     private readonly breakpointService: MediaBreakpointService,
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts
@@ -78,7 +78,6 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
     private readonly fileService: FileService,
     private readonly reportingService: ErrorReportingService,
     private readonly activatedProjectService: ActivatedProjectService,
-    private readonly cacheService: CacheService,
     private readonly locationService: LocationService,
     private readonly breakpointObserver: BreakpointObserver,
     private readonly breakpointService: MediaBreakpointService,
@@ -88,6 +87,7 @@ export class AppComponent extends DataLoadingComponent implements OnInit, OnDest
     readonly urls: ExternalUrlService,
     readonly featureFlags: FeatureFlagService,
     private readonly pwaService: PwaService,
+    _cacheService: CacheService, //don't remove, as this needs to be used to be constructed
     onlineStatusService: OnlineStatusService
   ) {
     super(noticeService);
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.spec.ts
@@ -2,30 +2,42 @@ import { NgZone } from '@angular/core';
 import { fakeAsync, flush, TestBed, tick } from '@angular/core/testing';
 import { createTestProjectProfile } from 'realtime-server/lib/esm/scriptureforge/models/sf-project-test-data';
 import { Chapter, TextInfo } from 'realtime-server/lib/esm/scriptureforge/models/text-info';
+import { BehaviorSubject } from 'rxjs';
 import { anything, deepEqual, instance, mock, verify, when } from 'ts-mockito';
+import { ActivatedProjectService } from 'xforge-common/activated-project.service';
 import { configureTestingModule } from 'xforge-common/test-utils';
+import { MemoryRealtimeDocAdapter } from '../../../xforge-common/memory-realtime-remote-store';
+import { RealtimeDocAdapter } from '../../../xforge-common/realtime-remote-store';
 import { SFProjectProfileDoc } from '../../core/models/sf-project-profile-doc';
-import { TextDocId } from '../../core/models/text-doc';
+import { TextDoc, TextDocId } from '../../core/models/text-doc';
 import { PermissionsService } from '../../core/permissions.service';
 import { SFProjectService } from '../../core/sf-project.service';
 import { CacheService } from './cache.service';
 
 const mockedProjectService = mock(SFProjectService);
-const mockedProjectDoc = mock(SFProjectProfileDoc);
 const mockedPermissionService = mock(PermissionsService);
+const mockedActivatedProject = mock(ActivatedProjectService);
+const projectId$ = new BehaviorSubject<string>('');
 
 describe('cache service', () => {
   configureTestingModule(() => ({
     providers: [
       { provide: SFProjectService, useMock: mockedProjectService },
-      { provide: PermissionsService, useMock: mockedPermissionService }
+      { provide: PermissionsService, useMock: mockedPermissionService },
+      { provide: ActivatedProjectService, useMock: mockedActivatedProject }
     ]
   }));
+
   describe('load all texts', () => {
+    beforeEach(() => {
+      const text = { adapter: instance(mock<RealtimeDocAdapter>()) } as TextDoc;
+      when(mockedProjectService.getText(anything())).thenResolve(text);
+      when(mockedActivatedProject.projectId$).thenReturn(projectId$);
+    });
+
     it('does not get texts from project service if no permission', fakeAsync(async () => {
-      const env = new TestEnvironment();
       when(mockedPermissionService.canAccessText(anything())).thenResolve(false);
-      await env.service.cache(env.projectDoc);
+      const env = new TestEnvironment();
       env.wait();
 
       verify(mockedProjectService.getText(anything())).times(0);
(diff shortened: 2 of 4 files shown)


josephmyers:
[The Jira task](https://jira.sil.org/browse/SF-3199) mentions two other solutions, which may produce better code in the longterm. I didn't investigate those much. I believe one would require a good bit of rework, and for the other I would need more discussion to fully how it would solve the original problem (offline support). I can look into these if desired.

RaymondLuong3 (changes requested):
This looks like a good direction to take for this time. It should make an improvement to the current behaviour. We can always explore the other options as the need arises.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3009#-:-OJ-RkyT805wSkKW5Vsm:b-ocbe6l)* status: all files reviewed, 4 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts` line 18 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3009#-OJ-GfLQ4Wu8vtQrnDSc:-OJ-GfLQ4Wu8vtQrnDSd:b-pz02l8) ([raw file](https://github.com/sillsdev/web-xforge/blob/c62760b852947b4cd0e284b4bc38c2b9e2e8fd5a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts#L18)):*
> ```XML
>     private readonly permissionsService: PermissionsService,
>     currentProject: ActivatedProjectService,
>     destroyRef: DestroyRef
> ```

I'm pretty sure these can be private.



_Code quote:_
```XML
    currentProject: ActivatedProjectService,
    destroyRef: DestroyRef
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts` line 31 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3009#-OJ-Huk3-p17UauPaD0l:-OJ-Huk3-p17UauPaD0m:b8ifd4v) ([raw file](https://github.com/sillsdev/web-xforge/blob/c62760b852947b4cd0e284b4bc38c2b9e2e8fd5a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts#L31)):*
> ```XML
>   private uncache(): void {
>     this.subscribedTexts.forEach(t => t.adapter.destroy());
>     this.subscribedTexts.length = 0;
> ```

I have never seen assigning a length to an array before. Why not just set the `subscribedTexts` value to an empty array?



_Code quote:_
```XML
    this.subscribedTexts.length = 0;
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts` line 41 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3009#-OJ-HO6o6JxYpUTdHuj1:-OJ-HO6o6JxYpUTdHuj2:b-8jv9wu) ([raw file](https://github.com/sillsdev/web-xforge/blob/c62760b852947b4cd0e284b4bc38c2b9e2e8fd5a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts#L41)):*
> ```XML
>   private async loadAllChapters(project: SFProjectProfileDoc): Promise<void> {
>     var abort = false;
>     const sub = this.abortCurrent.subscribe(() => (abort = true));
> ```

We probably want to unsubscribe to `abortCurrent`at the end of this function since the subscription is meaningless at that point.



_Code quote:_
```XML
    const sub = this.abortCurrent.subscribe(() => (abort = true));
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.spec.ts` line 12 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3009#-OJ-NXFXAywAe7FhXc60:-OJ-NXFXAywAe7FhXc61:b-rgjpth) ([raw file](https://github.com/sillsdev/web-xforge/blob/c62760b852947b4cd0e284b4bc38c2b9e2e8fd5a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.spec.ts#L12)):*
> ```XML
> import { TestRealtimeModule } from 'xforge-common/test-realtime.module';
> import { configureTestingModule, TestTranslocoModule } from 'xforge-common/test-utils';
> import { UICommonModule } from 'xforge-common/ui-common.module';
> ```

Nit: Can use absolute path for `xforge-common`

Nateowami (commented):
> Reviewed 2 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3009)* status: all files reviewed, 5 unresolved discussions (waiting on @josephmyers and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts` line 31 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3009#-OJ-Huk3-p17UauPaD0l:-OJ0fD25Bqx-a_IOGVM6:bjlrezs) ([raw file](https://github.com/sillsdev/web-xforge/blob/c62760b852947b4cd0e284b4bc38c2b9e2e8fd5a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts#L31)):*

I have never seen assigning a length to an array before. Why not just set the `subscribedTexts` value to an empty array?
</blockquote></details>

Is there a reason to not just set it to a new array (`x = []`) instead of clearing it?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts` line 30 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3009#-OJ0ekKl3a6JFiNNoCQN:-OJ0ekKl3a6JFiNNoCQO:bxtqbf8) ([raw file](https://github.com/sillsdev/web-xforge/blob/c62760b852947b4cd0e284b4bc38c2b9e2e8fd5a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts#L30)):*
> ```XML
> 
>   private uncache(): void {
>     this.subscribedTexts.forEach(t => t.adapter.destroy());
> ```

It's generally preferable to use `for...of` instead of `forEach`. I think it's more readable, but it's also supposed to be more performant on large arrays.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3009)* status: all files reviewed, 6 unresolved discussions (waiting on @josephmyers and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts` line 48 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3009#-OJ4Koeb4jeKfmoH8-WL:-OJ4Koeb4jeKfmoH8-WM:b-8lu241) ([raw file](https://github.com/sillsdev/web-xforge/blob/c62760b852947b4cd0e284b4bc38c2b9e2e8fd5a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts#L48)):*
> ```XML
>       for (const text of project.data.texts) {
>         for (const chapter of text.chapters) {
>           if (abort) {
> ```

Instead of setting up an abortCurrent event which then needs a subscription managed, why not change this to:
``` ts
if (this.activatedProject.projectId != null && this.activatedProject.projectId !== project.id) return;
```

Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3009)* status: all files reviewed, 7 unresolved discussions (waiting on @josephmyers and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts` line 81 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3009#-OJ4ii8U1W7fdPBdd5rv:-OJ4ii8U1W7fdPBdd5rw:b5cqaxz) ([raw file](https://github.com/sillsdev/web-xforge/blob/c62760b852947b4cd0e284b4bc38c2b9e2e8fd5a/src/SIL.XForge.Scripture/ClientApp/src/app/app.component.ts#L81)):*
> ```XML
>     private readonly reportingService: ErrorReportingService,
>     private readonly activatedProjectService: ActivatedProjectService,
>     private readonly cacheService: CacheService,
> ```

This shouldn't be a member, and it should have a comment stating why it is need (i.e. so it is active in the application). This is kind of an abuse of the AppComponent, but it's the best I'm aware of for now. Otherwise someone could very reasonably remove it.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3009)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts` line 18 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3009#-OJ-GfLQ4Wu8vtQrnDSc:-OJP4cDO8FETnhMm5jBh:b-dj7s06) ([raw file](https://github.com/sillsdev/web-xforge/blob/46580a27e15466479e76e33df71338e85066405e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts#L18)):*

They're not used outside the constructor, so I don't see any reason to save them. You're thinking these are getting saved as public fields? This would happen if I added `readonly` to them
</blockquote></details>

That's right. I forgot that bit.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts` line 31 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3009#-OJ-Huk3-p17UauPaD0l:-OJPBLtU2Tz22E-t6iyy:bfd8xca) ([raw file](https://github.com/sillsdev/web-xforge/blob/46580a27e15466479e76e33df71338e85066405e/src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts#L31)):*

Yes, it's readonly
</blockquote></details>

Ok, according to the docs this will delete the entries from the array when setting the length to 0. That seems reasonable.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3009)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3009)* status: all files reviewed, 3 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts` line 30 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3009#-OJQSGgH8gnmhrXYocay:-OJQSGgIEWKpw6LaruOD:biqgsqk) ([raw file](https://github.com/sillsdev/web-xforge/blob/d1e95e2987afee2bae8638cbf12e198b7bdfc68d/src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts#L30)):*
> ```XML
>   private uncache(): void {
>     for (const t of this.subscribedTexts) {
>       t.adapter.destroy();
> ```

I see two potential issues with this:
1. This *doesn't* actually clean up the document (see the diagnostics panel). To do that, you need `doc.dispose()`
2. I'm not sure you *can* safely dispose of a document in this context. How can we be certain nothing is using it?

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3009)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @Nateowami and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts` line 30 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3009#-OJQSGgH8gnmhrXYocay:-OJqwHQFA7Czo1xXR-Jb:b-rfeles) ([raw file](https://github.com/sillsdev/web-xforge/blob/d1e95e2987afee2bae8638cbf12e198b7bdfc68d/src/SIL.XForge.Scripture/ClientApp/src/app/shared/cache-service/cache.service.ts#L30)):*

I see two potential issues with this:
1. This *doesn't* actually clean up the document (see the diagnostics panel). To do that, you need `doc.dispose()`
2. I'm not sure you *can* safely dispose of a document in this context. How can we be certain nothing is using it?
</blockquote></details>

I've reimplemented the caching to use queries, instead of pulling each doc, as suggested in the JIRA task. This is certainly cleaner, but I'm not sure it fully addresses this comment. At the same time, I'm not sure it's this class's responsibility to manage the site's memory. I would think its responsibility is to ensure it properly unsubscribes. Then, if a document has no subscribers (unreachable), it can be disposed of by garbage collection when appropriate.

This is somewhat pulling from a C# understanding of garbage collection.

josephmyers:
Yes, Raymond, there's still some more work to be done on this. I've sent a message out to Nathaniel and Peter yesterday. In fact, the second implementation actually doesn't fix the memory leak at all. I should have tested it better, like I did for destroying the adapters. More to come on this

josephmyers:
Were you seeing this latest change fix the memory leak? Maybe I missed something in my testing. Otherwise, we should pull it out from Ready to Test

josephmyers:
> I have a solution in the works. And I think disposing documents is better than disposing the adapter (I think the adapter is disposed when the document is disposed)

From Nathaniel. I've pulled this task back into Code Review, and it sounds like we may even just can this guy. I'll keep this PR updated as I learn more.

josephmyers:
Nathaniel mentioned last week that he has a solution in the works

PR #3012, opened 2025-02-13
SF-2829 Allow paratext roles that have changed to be updated

kylebuss:
This PR addresses the issue where users roles have been updated in Paratext but not reflected in Scripture Forge. Primarily when a user is changed to an Paratext Administrator they do not have administrator access in SF until another admin runs a sync.

My Projects page UI change:

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3012)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/paratext-project.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/paratext-project.ts
@@ -7,4 +7,5 @@ export interface ParatextProject {
   projectId?: string;
   isConnectable: boolean;
   isConnected: boolean;
+  hasUserRoleChanged?: boolean;
 }
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html
@@ -6,6 +6,38 @@
         {{ t("no_projects") }} {{ t("access_another_project") }}
       </app-notice>
     }
+    @if (userUpdateParatextProjects.length > 0) {
+      <h2 id="header-update-projects">
+        {{ t("updates") }}
+      </h2>
+      @if ((isOnline | async) === false) {
+        <app-notice id="message-offline" icon="cloud_off" [type]="userHasProjects ? 'info' : 'error'">
+          {{ t("offline_update") }}
+        </app-notice>
+      }
+    }
+    @for (ptProject of userUpdateParatextProjects; track ptProject) {
+      <mat-card
+        id="user-update-project-card-{{ ptProject.projectId }}"
+        [attr.data-pt-project-id]="ptProject.projectId"
+        class="user-update-project mat-elevation-z4"
+      >
+        <span class="project-name">
+          <b>{{ ptProject.shortName }}</b> - {{ ptProject.name }}
+        </span>
+        <span class="project-description">
+          {{ t("update_paratext_role") }}
+        </span>
+        <button
+          class="user-update-project-btn"
+          mat-flat-button
+          [disabled]="(isOnline | async) === false"
+          (click)="syncUserRole(ptProject.projectId!)"
+        >
+          {{ t("update") }}
+        </button>
+      </mat-card>
+    }
     @if (userConnectedProjects.length > 0 || initialLoadingSFProjects) {
       <h2 id="header-connected-projects">
         {{ t("connected") }}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.html
@@ -20,7 +20,7 @@
       <mat-card
         id="user-update-project-card-{{ ptProject.projectId }}"
         [attr.data-pt-project-id]="ptProject.projectId"
-        class="user-update-project mat-elevation-z4"
+        class="user-update-project"
       >
         <span class="project-name">
           <b>{{ ptProject.shortName }}</b> - {{ ptProject.name }}
@@ -31,6 +31,7 @@
         <button
           class="user-update-project-btn"
           mat-flat-button
+          color="primary"
           [disabled]="(isOnline | async) === false"
           (click)="syncUserRole(ptProject.projectId!)"
         >
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss
@@ -96,6 +96,22 @@ mat-card:active {
   flex-direction: column;
 }
 
+.user-update-project {
+  display: grid;
+  grid-template-areas:
+    'project-name project-name'
+    'description empty';
+  background-color: #e8fafd;
+}
+
+.user-update-project-btn {
+  background-color: #51d6f0 !important;
+}
+
+.user-update-project-btn:hover {
+  background-color: white !important;
+}
+
 .loading-card {
   display: grid;
   grid-template-areas:
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss
@@ -54,10 +54,16 @@ mat-card {
   grid-template-columns: 1fr auto;
   cursor: pointer;
   transition: all 0.1s;
+  &.user-update-project {
+    cursor: unset;
+  }
 }
 
 mat-card:hover {
   filter: brightness(90%);
+  &.user-update-project {
+    filter: none;
+  }
 }
 
 mat-card:active {
@@ -101,15 +107,11 @@ mat-card:active {
   grid-template-areas:
     'project-name project-name'
     'description empty';
-  background-color: #e8fafd;
+  background-color: lighten(variables.$info-color, 40%);
 }
 
-.user-update-project-btn {
-  background-color: #51d6f0 !important;
-}
-
-.user-update-project-btn:hover {
-  background-color: white !important;
+.user-update-project-btn:not(:hover):not(:disabled) {
+  background-color: variables.$info-color !important;
 }
 
 .loading-card {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.spec.ts
@@ -444,6 +444,63 @@ describe('MyProjectsComponent', () => {
     // We won't expect getProjects() to have been called.
     verify(mockedParatextService.getProjects()).never();
   }));
+
+  it('shows project in updates section when user role has changed and is removed when role is updated successfully', fakeAsync(() => {
+    // Suppose a users role on Paratext has been changed, but the role on SF has not been updated.
+    // We should display the project in the updates section, and show a button to update the role.
+    const env = new TestEnvironment({ isKnownPTUser: true });
+    when(mockedParatextService.getProjects()).thenResolve([
+      {
+        paratextId: 'pt-project-id-update-role',
+        name: 'PT Project Update Role',
+        projectId: 'testProject1',
+        shortName: 'PTUR',
+        isConnectable: true,
+        isConnected: true,
+        hasUserRoleChanged: true
+      }
+    ] as ParatextProject[]);
+    env.waitUntilLoaded();
+    expect(env.component.userUpdateParatextProjects.length).toEqual(1);
+    expect(
+      env.component.userUpdateParatextProjects.find(proj => proj.projectId === 'testProject1')?.hasUserRoleChanged
+    ).toBeTrue();
+    expect(env.userUpdateParatextProjects).not.toBeNull();
+
+    env.click(env.buttonForUpdateUserRole('testProject1'));
+    verify(mockedNoticeService.loadingStarted(anything())).once();
+    verify(mockedSFProjectService.onlineSyncUserRole('testProject1')).once();
+    expect(env.component.userUpdateParatextProjects.length).toEqual(0);
+  }));
+
+  it('shows project in updates section when user role has changed in paratext and remains when role is not updated', fakeAsync(() => {
+    const env = new TestEnvironment({ isKnownPTUser: true });
+    when(mockedParatextService.getProjects()).thenResolve([
+      {
+        paratextId: 'pt-project-id-update-role',
+        name: 'PT Project Update Role',
+        projectId: 'testProject1',
+        shortName: 'PTUR',
+        isConnectable: true,
+        isConnected: true,
+        hasUserRoleChanged: true
+      }
+    ] as ParatextProject[]);
+    env.waitUntilLoaded();
+    expect(env.component.userUpdateParatextProjects.length).toEqual(1);
+    expect(
+      env.component.userUpdateParatextProjects.find(proj => proj.projectId === 'testProject1')?.hasUserRoleChanged
+    ).toBeTrue();
+    expect(env.userUpdateParatextProjects).not.toBeNull();
+
+    when(mockedSFProjectService.onlineSyncUserRole('testProject1')).thenThrow(new Error('test error'));
+    env.click(env.buttonForUpdateUserRole('testProject1'));
+
+    expect(env.component.userUpdateParatextProjects.length).toEqual(1);
+    expect(
+      env.component.userUpdateParatextProjects.find(proj => proj.projectId === 'testProject1')?.hasUserRoleChanged
+    ).toBeTrue();
+  }));
 });
 
 class TestEnvironment {
@@ -624,11 +681,19 @@ class TestEnvironment {
     return this.getElement(`div[data-pt-project-id=${ptProjectId}].user-unconnected-project`);
   }
 
+  userUpdateParatextProjects(ptProjectId: string): DebugElement {
+    return this.getElement(`mat-card[data-pt-project-id=${ptProjectId}].user-update-project`);
+  }
+
   buttonForUnconnectedProject(ptProjectId: string): DebugElement {
     return (
       this.getElement(`div[data-pt-project-id=${ptProjectId}] > button`) ??
       this.getElement(`div[data-pt-project-id=${ptProjectId}] > a`)
     );
+  }
+
+  buttonForUpdateUserRole(ptProjectId: string): DebugElement {
+    return this.getElement(`mat-card[data-pt-project-id=${ptProjectId}].user-update-project > button`);
   }
 
   waitUntilLoaded(): void {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts
@@ -28,7 +28,7 @@ import { SFProjectService } from '../core/sf-project.service';
   styleUrls: ['./my-projects.component.scss']
 })
 export class MyProjectsComponent extends SubscriptionDisposable implements OnInit {
-  /** PT projects that the user can access that they are not connected to on SF. */
+  /** PT projects that the user can access. */
   private userParatextProjects: ParatextProject[] = [];
 
   /** SF projects that the current user is on at SF. */
@@ -119,13 +119,6 @@ export class MyProjectsComponent extends SubscriptionDisposable implements OnIni
     return `${drafting}${checking}`;
   }
 
-  userRoleNeedsUpdated(projectId: string): boolean {
-    return (
-      this.userIsPTUser &&
-      (this.userParatextProjects?.find(project => project.projectId === projectId)?.hasUserRoleChanged ?? false)
-    );
-  }
-
   async joinProject(projectId: string): Promise<void> {
     try {
       this.noticeService.loadingStarted(this.constructor.name);
@@ -163,12 +156,12 @@ export class MyProjectsComponent extends SubscriptionDisposable implements OnIni
     this.loadingPTProjects = true;
     this.problemGettingPTProjects = false;
     try {
-      const getPtProjects = await this.paratextService.getProjects();
-      if (getPtProjects == null) {
+      const userPtProjects = await this.paratextService.getProjects();
+      if (userPtProjects == null) {
         this.problemGettingPTProjects = true;
         return;
       }
-      this.userParatextProjects = getPtProjects;
+      this.userParatextProjects = userPtProjects;
     } catch (err) {
       if (err instanceof HttpErrorResponse && err.status === HttpStatusCode.ServiceUnavailable) {
         this.errorMessage = 'failed_to_connect_to_pt_server';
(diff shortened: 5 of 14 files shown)


josephmyers:
Your permissions on this project are out of date. [Update]

josephmyers (changes requested):
I appreciate your work so far on this. There are a couple minor problems with the general UX. I apologize for not laying out better what needed to be done, as I wasn't expecting this fix to involve UI changes. Case in point, you could make an argument for updating the role automatically, with no user input; however, we're instead going in a different direction, and I'll lay out some specific requirements to help you along:

-   Show a notice on the My Projects page, at the top (above the list of projects), in a new section titled “Update permissions” or "Updates"
    
-   Inform the user if the role update fails
    
-   On success, do not redirect the user
    
-   The project card should still be visible, i.e. not replaced
    
-   Hide notice immediately upon success (and section if no “updates” are left)

Also, I did a brief mockup in Balsamiq of a different way to style the notice. This spawned from Nigel's suggestion to keep consistency between this new item and the existing project cards. Most of this is optional, and it will as always need to be tailored to fit the actual site (not the wireframe), but at the least we will want to adopt the wording and move the button to the righthand side.

https://balsamiq.cloud/sghq53/pejrlz8/r4D30

Once we're pretty good with the UX, I'll do a better code review for you. Thanks for your patience and work!

> Reviewed 1 of 7 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-:-OJMKYwQ7NUOUcoP-uEl:ba203pk)* status: 1 of 9 files reviewed, 3 unresolved discussions

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts` line 70 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJMJoL26Sfou5I0X9tP:-OJMJoL26Sfou5I0X9tQ:b-tloa1u) ([raw file](https://github.com/sillsdev/web-xforge/blob/c76158dc1c8be8721dc5a42a780b460981689528/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts#L70)):*
> ```XML
> 
>   get userUnconnectedParatextProjects(): ParatextProject[] {
>     return this.userParatextProjects == null ? [] : this.userParatextProjects.filter(project => !project.isConnected);
> ```

I like this change, but we should be able to make `userParatextProjects` private now.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts` line 159 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJMIJXy7hImkrIoypR2:-OJMIJXy7hImkrIoypR3:b-palpi3) ([raw file](https://github.com/sillsdev/web-xforge/blob/c76158dc1c8be8721dc5a42a780b460981689528/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts#L159)):*
> ```XML
>     this.problemGettingPTProjects = false;
>     try {
>       this.userParatextProjects = await this.paratextService.getProjects();
> ```

There's a problem with this small change, in that now the public property can become null/undefined. Beforehand, we would return before clearing, maintaining the integrity of the class. We could preserve this easily by introducing a temporary variable here and bumping the field assignment back to where it was.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/models/paratext-project.ts` line 11 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJM0amy5XN3iwd9t74S:-OJM0amy5XN3iwd9t74T:b-ppj5wv) ([raw file](https://github.com/sillsdev/web-xforge/blob/c76158dc1c8be8721dc5a42a780b460981689528/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/paratext-project.ts#L11)):*
> ```XML
>   isConnectable: boolean;
>   isConnected: boolean;
>   userRoleNeedsUpdated?: boolean;
> ```

This messed up grammar bugs me, even though this is code speak. It'd be simpler to say "update" instead of "updated"

kylebuss (commented):
Thanks for the UI/UX requirements. This might still need some fine tuning on the UI but everything requested should be complete. There was also a change to the backend that I made because the original remove user then add user would create issues with a new user id not matching existing notes/questions.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-:-OJTnbAn05bVNe0OYBKY:b-nj9ytl)* status: 0 of 15 files reviewed, 3 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/core/models/paratext-project.ts` line 11 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJM0amy5XN3iwd9t74S:-OJOCvB5DZW_OK68VIDi:bs7wumx) ([raw file](https://github.com/sillsdev/web-xforge/blob/c76158dc1c8be8721dc5a42a780b460981689528/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/paratext-project.ts#L11)):*

This messed up grammar bugs me, even though this is code speak. It'd be simpler to say "update" instead of "updated"
</blockquote></details>

Done. 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts` line 70 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJMJoL26Sfou5I0X9tP:-OJODFUu8g7YnchJHaYV:b-oz8o77) ([raw file](https://github.com/sillsdev/web-xforge/blob/c76158dc1c8be8721dc5a42a780b460981689528/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts#L70)):*

I like this change, but we should be able to make `userParatextProjects` private now.
</blockquote></details>

Good catch
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts` line 159 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJMIJXy7hImkrIoypR2:-OJOEnDREGndLSY6DMMf:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/c76158dc1c8be8721dc5a42a780b460981689528/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts#L159)):*

There's a problem with this small change, in that now the public property can become null/undefined. Beforehand, we would return before clearing, maintaining the integrity of the class. We could preserve this easily by introducing a temporary variable here and bumping the field assignment back to where it was.
</blockquote></details>

Done.

josephmyers (changes requested):
This is looking pretty good. I have a couple smaller comments about the UX:

- These update cards look like buttons, even though they don't route to anything. Start with preventing the cursor change and removing the hover and click behaviors, and overall just make sure it doesn't act like a button that goes nowhere.
- Nit: The hover color for the Update button is a tiny bit weird. Maybe it will be fixed when you change the hover for the surrounding card?

> Reviewed 6 of 14 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-:-OJrDPHC5MMFW66tmifn:b4phn3c)* status: 6 of 15 files reviewed, 5 unresolved discussions (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts` line 159 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJMIJXy7hImkrIoypR2:-OJrGdM3DZiSp2BDf5Zp:b-98se7s) ([raw file](https://github.com/sillsdev/web-xforge/blob/c76158dc1c8be8721dc5a42a780b460981689528/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts#L159)):*

Done.
</blockquote></details>

Can you reset this name back to userPTProjects? It made a little more sense
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 104 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJrFNhN2S-cS3FuNjQR:-OJrFNhN2S-cS3FuNjQS:b-v9up9f) ([raw file](https://github.com/sillsdev/web-xforge/blob/6c2c499070dba7aa962208b506be2014ffc3471d/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L104)):*
> ```SCSS
>     'project-name project-name'
>     'description empty';
>   background-color: #e8fafd;
> ```

Did you look into reusing any of the themed colors we already have? E.g., did you experiment with `$info-color` at all?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts` line 31 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJrFv8K1SCv6KfCXCUz:-OJrFv8K1SCv6KfCXCV-:bfkjg7s) ([raw file](https://github.com/sillsdev/web-xforge/blob/6c2c499070dba7aa962208b506be2014ffc3471d/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts#L31)):*
> ```XML
> })
> export class MyProjectsComponent extends SubscriptionDisposable implements OnInit {
>   /** PT projects that the user can access that they are not connected to on SF. */
> ```

Is this comment correct? It sounds like you're describing `userUnconnectedParatextProjects`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts` line 122 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJrGBHQ4qEy5v56vAxf:-OJrGBHQ4qEy5v56vAxg:b-rpad72) ([raw file](https://github.com/sillsdev/web-xforge/blob/6c2c499070dba7aa962208b506be2014ffc3471d/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts#L122)):*
> ```XML
>   }
> 
>   userRoleNeedsUpdated(projectId: string): boolean {
> ```

Grammar here, too
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts` line 151 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJrH-E8CM17nj1C7bFJ:-OJrH-E8CM17nj1C7bFK:b-al445y) ([raw file](https://github.com/sillsdev/web-xforge/blob/6c2c499070dba7aa962208b506be2014ffc3471d/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts#L151)):*
> ```XML
>       if (project != null) project.hasUserRoleChanged = false;
>     } catch {
>       this.noticeService.showError(translate('my_projects.failed_to_update_user_role'));
> ```

Do these errors get sent to Bugsnag? Make sure we're not suppressing exceptions. See also `ErrorReportingService.silentError`

kylebuss (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3012)* status: 3 of 16 files reviewed, 5 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 104 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJrFNhN2S-cS3FuNjQR:-OJsI2Br4CY0fe5-Cvl5:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/6c2c499070dba7aa962208b506be2014ffc3471d/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L104)):*

Did you look into reusing any of the themed colors we already have? E.g., did you experiment with `$info-color` at all?
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts` line 159 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJMIJXy7hImkrIoypR2:-OJsSFkW3kKFeyChGj6G:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/c76158dc1c8be8721dc5a42a780b460981689528/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts#L159)):*

Can you reset this name back to userPTProjects? It made a little more sense
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts` line 31 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJrFv8K1SCv6KfCXCUz:-OJsIS1W-VmZqyF4NjyJ:bs7wumx) ([raw file](https://github.com/sillsdev/web-xforge/blob/6c2c499070dba7aa962208b506be2014ffc3471d/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts#L31)):*

Is this comment correct? It sounds like you're describing `userUnconnectedParatextProjects`
</blockquote></details>

Done. 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts` line 122 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJrGBHQ4qEy5v56vAxf:-OJsISTd1uluwmXfr2TZ:bs7wumx) ([raw file](https://github.com/sillsdev/web-xforge/blob/6c2c499070dba7aa962208b506be2014ffc3471d/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts#L122)):*

Grammar here, too
</blockquote></details>

Done. 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts` line 151 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJrH-E8CM17nj1C7bFJ:-OJsJ3PX0A6aSg_qzz1F:betyg78) ([raw file](https://github.com/sillsdev/web-xforge/blob/6c2c499070dba7aa962208b506be2014ffc3471d/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.ts#L151)):*

Do these errors get sent to Bugsnag? Make sure we're not suppressing exceptions. See also `ErrorReportingService.silentError`
</blockquote></details>

Yes, this should be catching if the API had an exception and report it to bugsnag through the `command.service.ts`. The `ErrorReportingService`, as I understand it, should be reporting any Angular errors to bugsnag.

josephmyers (changes requested):
A few more simple UI comments. And a minor deficiency in this task is that updates don't appear dynamically -- you must navigate away and back or refresh. I wouldn't spend too much time on this, as it's not going to be common to just sit on the My Projects page. But all that aside, it's looking really good!

> Reviewed 1 of 7 files at r2, 8 of 14 files at r3, 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-:-OJw3jpR2TCvdeEg1vem:b-v5nypc)* status: all files reviewed, 5 unresolved discussions (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 104 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJrFNhN2S-cS3FuNjQR:-OJvkQrf3iRZXfhGUcBD:b-2bow4u) ([raw file](https://github.com/sillsdev/web-xforge/blob/6c2c499070dba7aa962208b506be2014ffc3471d/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L104)):*

Done.
</blockquote></details>

We should stick with the original #547ec8 for `$info-color`, unless you have a strong preference against it. I was suggesting you change your styling, not change the saved variables we already have. When appropriate, we want to match the colors we've already defined. Try this:

`background-color: lighten($info-color, 40%);`

in conjunction with below:
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 58 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJvkqAmEb2xNZqCvRzR:-OJvkqAmEb2xNZqCvRzS:bws5wok) ([raw file](https://github.com/sillsdev/web-xforge/blob/667fe24a6d9d2dea8fc76e435a5648268185820b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L58)):*
> ```SCSS
>   transition: all 0.1s;
>   &.user-update-project {
>     cursor: default;
> ```

One small problem I see with this is that the cursor doesn't change to the I-beam when over text. You can fix this easily with `cursor: unset;`
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 114 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJvlpJl37qiyMj8GrX_:-OJvlpJl37qiyMj8GrXa:b-a2sh4v) ([raw file](https://github.com/sillsdev/web-xforge/blob/667fe24a6d9d2dea8fc76e435a5648268185820b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L114)):*
> ```SCSS
> 
> .user-update-project-btn {
>   background-color: #51d6f0 !important;
> ```

This. You'll also see I account for offline mode, which beforehand didn't have great contrast. Only applying our custom styling in specific cases allows us (when it looks acceptable) to use the pre-existing behavior.



_Code snippet:_
```SCSS
.user-update-project-btn:not(:hover):not(:disabled) {
  background-color: $info-color !important;
}
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 118 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJvnLYQAG4qiNGMo4aw:-OJvnLYQAG4qiNGMo4ax:bc1r0h8) ([raw file](https://github.com/sillsdev/web-xforge/blob/667fe24a6d9d2dea8fc76e435a5648268185820b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L118)):*
> ```SCSS
> 
> .user-update-project-btn:hover {
>   background-color: white !important;
> ```

You should also remove this and add `color='primary'` to these buttons' HTML. We get all the benefits of the primary color, which is what we want, while only overriding the background color in one scenario (first glance).

Do all of these and you'll get this:

___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2274 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJvyFAyB8nSuwf_6w7C:-OJvyFAyB8nSuwf_6w7D:bcug2uj) ([raw file](https://github.com/sillsdev/web-xforge/blob/667fe24a6d9d2dea8fc76e435a5648268185820b/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2274)):*
> ```Smalltalk
>                 sfProjectExists
>                 && sfUserIsOnSfProject
>                 && ConvertFromUserRole(ptUserRole) != correspondingSfProject.UserRoles[userSecret.Id];
> ```

Good, readable code. Nice!

kylebuss (commented):
Thanks for the feedback on this, css is not my strong suit. I'm not sure I follow what you mean that the updates don't appear dynamically?

I think the Jira issue may need updated to reflect how we decided to implement this. Unless you are referring to a bug you've observed with this implementation?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-:-OJxMsqfAZFBFat7UXjP:b-uprjzw)* status: 11 of 16 files reviewed, 4 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 104 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJrFNhN2S-cS3FuNjQR:-OJxMd61BoHoai8Vf7Ig:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/6c2c499070dba7aa962208b506be2014ffc3471d/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L104)):*

We should stick with the original #547ec8 for `$info-color`, unless you have a strong preference against it. I was suggesting you change your styling, not change the saved variables we already have. When appropriate, we want to match the colors we've already defined. Try this:

`background-color: lighten($info-color, 40%);`

in conjunction with below:
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 58 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJvkqAmEb2xNZqCvRzR:-OJxJnrT5DqSuxIGyTe7:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/667fe24a6d9d2dea8fc76e435a5648268185820b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L58)):*

One small problem I see with this is that the cursor doesn't change to the I-beam when over text. You can fix this easily with `cursor: unset;`
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 114 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJvlpJl37qiyMj8GrX_:-OJxMdoa3XFJ8lLZGrsy:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/667fe24a6d9d2dea8fc76e435a5648268185820b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L114)):*

This. You'll also see I account for offline mode, which beforehand didn't have great contrast. Only applying our custom styling in specific cases allows us (when it looks acceptable) to use the pre-existing behavior.
</blockquote></details>

Done.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss` line 118 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-OJvnLYQAG4qiNGMo4aw:-OJxMf5jDKfEjjzjOhDf:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/667fe24a6d9d2dea8fc76e435a5648268185820b/src/SIL.XForge.Scripture/ClientApp/src/app/my-projects/my-projects.component.scss#L118)):*

You should also remove this and add `color='primary'` to these buttons' HTML. We get all the benefits of the primary color, which is what we want, while only overriding the background color in one scenario (first glance).

Do all of these and you'll get this:

</blockquote></details>

Done.

josephmyers (changes requested):
Yes, it's a small bug. If the user is on the My Projects page and his role is updated, the "Updates" don't appear (until he navigates away and comes back).

> Reviewed 5 of 5 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-:-OJzVeKZ8FgUtNxPGK9i:bu6ok9s)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kylebuss)

kylebuss:
I think this is acceptable for now. We currently query the Paratext Registry once when the My Project page loads - this is when we determine if the user role in Paratext has changed. 

It may be worth a backlog issue to further improve this by updating the project user roles model to reflect when a changed is detected so we can subscribe to changes in the realtime server?

kylebuss (commented):
Thanks again for reviewing this! 

It's probably not worth the effort to try and make this more dynamic now. My understanding is this is something of an edge case so letting users know they can navigate to or refresh the My Projects page to see if there are any Updates for a given project should be sufficient.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3012#-:-OK6baeRE8o-cKOgMHZq:b9boh9v)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kylebuss)

josephmyers:
@kylebuss, just to be clear, the bug I reported is an edge case, but the core bug, which you've fixed with this PR, is not an edge case. It happens pretty regularly in my experience, but I think you've made it much less confusing for folks. Good job

PR #3019, opened 2025-02-15
Fix common audio test failure

Nateowami:
I put this test in a loop and ran it 100 times. With an 1100ms delay, it failed 56 times. With a 1200ms delay, it didn't fail once. I don't think in practice this means this test will completely stop failing, but it should be a lot less frequent.

Note that if the delay is set to 1500ms, it starts failing with another error. This is a very short audio file, and we're trying to pause it after 1s and before it finishes playing, so we can see a paused state. If I log the duration, it's `1.296`.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3019)


pmachapman (approved):
:lgtm: I appreciate your PR comment, as it answered all of my questions!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3019#-:-OJF41_O8kXuoUoVzK1g:b-322286)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #3035, opened 2025-02-21
SF-3188 Upgrade to ShareDB 5.1.1 on server side

RaymondLuong3:
This PR upgrades sharedb to version 5.1.1 on the realtime server. This may be a safer way of upgrading because we can have confidence that having sharedb at version 5.1.1 is still compatible with clients on the previous sharedb 1.9.2. Then we can upgrade sharedb on the client app.

From running the realtime server tests, I found some deprecated messages that needed to be addressed

- updated realtime server constructor initializes ShareDB with `doNotForwardSendPresenceErrorsToClient`
- updated to mongodb 4.17.2 to satisfy the punycode deprecation warning (but it seems we should upgrade to further to 6.x)

Things I checked:

- the breaking changes from shareDB 1.x to 5.1.1 do not affect us (most breaking changes are dropping support for older versions of node).
- our sharedb extended classes overriding private methods (i.e. MigrationAgent._handleMessage()) do not affect how we override the method
- manual real-time testing for behaviour (online and offline)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3035)

### DIFF
--- a/src/RealtimeServer/common/realtime-server.ts
+++ b/src/RealtimeServer/common/realtime-server.ts
@@ -152,7 +152,8 @@ export class RealtimeServer extends ShareDB {
       milestoneDb,
       presence: true,
       disableDocAction: true,
-      disableSpaceDelimitedActions: true
+      disableSpaceDelimitedActions: true,
+      doNotForwardSendPresenceErrorsToClient: true
     });
     shareDBAccess(this);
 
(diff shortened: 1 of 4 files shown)


pmachapman (approved):
:lgtm: This looks ready to send to the test team. I couldn't find any issues in my testing and reviewing of the changes to sharedb and the other updated libraries.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3035#-:-OKSn62-5gUHNFBj7MGV:b-csq2n5)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #3037, opened 2025-02-24
SF-2917 Log when a project is deleted

pmachapman:
This PR adds logging to the event metrics collection when a project is deleted.

Because the acceptance test requires looking at the database, it can be developer tested.

### DIFF
--- a/src/SIL.XForge.Scripture/Services/SFProjectService.cs
+++ b/src/SIL.XForge.Scripture/Services/SFProjectService.cs
@@ -255,14 +255,44 @@ public class SFProjectService : ProjectService<SFProject, SFProjectSecret>, ISFP
             if (projectDoc.Data.Sync.QueuedCount > 0)
                 throw new InvalidOperationException("A project cannot be deleted while it is syncing.");
 
+            // Get the user information who is deleting the project
+            IDocument<User> userDoc = await conn.FetchAsync<User>(curUserId);
+            if (!userDoc.IsLoaded)
+                throw new ForbiddenException();
+
+            // Log this event to the event metrics
+            // We do this here to record additional information about the user and project
             string ptProjectId = projectDoc.Data.ParatextId;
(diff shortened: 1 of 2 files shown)


Nateowami (approved):
Tested and :lgtm:. Merging.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3037#-:-OJsUO294oPx34L3f7Tz:b-o04k46)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/SFProjectService.cs` line 275 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3037#-OJsVm1yBqPZ8BFSbORH:-OJsVm1yBqPZ8BFSbORI:bxp5jk4) ([raw file](https://github.com/sillsdev/web-xforge/blob/3de6f6e1c932c34ca94cc1d6bc7663386d4cd0e1/src/SIL.XForge.Scripture/Services/SFProjectService.cs#L275)):*
> ```Smalltalk
>             await _eventMetricService.SaveEventMetricAsync(
>                 projectId,
>                 curUserId,
> ```

The only thing I don't like about this is that it would be easy to accidentally swap the first two argument since they both have the same type, and it wouldn't be very obvious that it had happened. I see the same issue on the TypeScript side, where I'd like there to be an error if we ever try to assign a SFProjectId to a PTProjectId, or directly compare them (even though they're both strings, there's a sense in which they are a completely different type). 

One way you can avoid this is to pass an argument with properties, rather than multiple arguments. In TypeScript it's as simple as something like the below, though I don't think C# has as simple of a syntax:
``` ts
saveEventmetricAsync({ projectId, curUserId }, ...)
```
...and it becomes difficult to accidentally swap the values.

I'm not asking for any changes; just pointing out the one thing that concerns me, especially when at some point in the future a different developer is adding logging.

PR #3045, opened 2025-02-26
ci: rely on called workflow environment secrets

marksvc:
release-qa.yml job deploy can't access secrets without an
`environment`. But it can't use an `environment` without a `runs-on`.
But if it has a `runs-on` it can't have a `uses`.

In investigating solutions for this, I observe that release.yml job
deploy should be able to access secrets anyway from its specified
environment.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3045)

### DIFF
--- a/.github/workflows/release-live.yml
+++ b/.github/workflows/release-live.yml
@@ -91,21 +91,6 @@ jobs:
   deploy:
     name: "Deploy to Live"
     needs: [determine_build_commit, determine_version]
-    secrets:
-      deploy_known_hosts: ${{ secrets.deploy_known_hosts }}
-      deploy_ssh_private_key: ${{ secrets.deploy_ssh_private_key }}
-      deploy_server: ${{ secrets.deploy_server }}
-      alternate_domain: ${{ secrets.alternate_domain }}
-      paratext_client_id: ${{ secrets.paratext_client_id }}
-      serval_client_id: ${{ secrets.serval_client_id }}
-      auth_backend_secret: ${{ secrets.auth_backend_secret }}
-      auth_webhook_password: ${{ secrets.auth_webhook_password }}
-      auth_health_check_api_key: ${{ secrets.auth_health_check_api_key }}
-      paratext_api_token: ${{ secrets.paratext_api_token }}
-      paratext_resource_password_base64: ${{ secrets.paratext_resource_password_base64 }}
-      paratext_resource_password_hash: ${{ secrets.paratext_resource_password_hash }}
-      serval_client_secret: ${{ secrets.serval_client_secret }}
-      serval_webhook_secret: ${{ secrets.serval_webhook_secret }}
     uses: ./.github/workflows/release.yml
     with:
       environment: "production"
--- a/.github/workflows/release-qa.yml
+++ b/.github/workflows/release-qa.yml
@@ -112,21 +112,6 @@ jobs:
   deploy:
     name: "Deploy to QA"
     needs: [determine_build_commit, determine_version]
-    secrets:
-      deploy_known_hosts: ${{ secrets.deploy_known_hosts }}
-      deploy_ssh_private_key: ${{ secrets.deploy_ssh_private_key }}
-      deploy_server: ${{ secrets.deploy_server }}
-      alternate_domain: ${{ secrets.alternate_domain }}
-      paratext_client_id: ${{ secrets.paratext_client_id }}
-      serval_client_id: ${{ secrets.serval_client_id }}
-      auth_backend_secret: ${{ secrets.auth_backend_secret }}
-      auth_webhook_password: ${{ secrets.auth_webhook_password }}
-      auth_health_check_api_key: ${{ secrets.auth_health_check_api_key }}
-      paratext_api_token: ${{ secrets.paratext_api_token }}
-      paratext_resource_password_base64: ${{ secrets.paratext_resource_password_base64 }}
-      paratext_resource_password_hash: ${{ secrets.paratext_resource_password_hash }}
-      serval_client_secret: ${{ secrets.serval_client_secret }}
-      serval_webhook_secret: ${{ secrets.serval_webhook_secret }}
     uses: ./.github/workflows/release.yml
     with:
       environment: "qa_deploy"


RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3045)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`.github/workflows/release-live.yml` line 97 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3045#-OK2Y3Ch4laS8p_PNoJ7:-OK2Y3Ch4laS8p_PNoJ8:b-mijliq) ([raw file](https://github.com/sillsdev/web-xforge/blob/32a97ff3641d4309cf1369de88014aa7461224fa/.github/workflows/release-live.yml#L97)):*
> ```YAML
>       deploy_known_hosts: ${{ secrets.deploy_known_hosts }}
>       deploy_ssh_private_key: ${{ secrets.deploy_ssh_private_key }}
>       deploy_server: ${{ secrets.deploy_server }}
> ```

In this case, we are relying on the `release.yml` file already having access to these secrets, is that correct? I don't see these secrets in the `release.yml` file. Should these be added there?



_Code quote:_
```YAML
      deploy_known_hosts: ${{ secrets.deploy_known_hosts }}
      deploy_ssh_private_key: ${{ secrets.deploy_ssh_private_key }}
      deploy_server: ${{ secrets.deploy_server }}
```

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3045)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`.github/workflows/release-live.yml` line 97 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3045#-OK2Y3Ch4laS8p_PNoJ7:-OK2ZCU_AibdR0y6cZFm:bacj7he) ([raw file](https://github.com/sillsdev/web-xforge/blob/32a97ff3641d4309cf1369de88014aa7461224fa/.github/workflows/release-live.yml#L97)):*

In this case, we are relying on the `release.yml` file already having access to these secrets, is that correct? I don't see these secrets in the `release.yml` file. Should these be added there?
</blockquote></details>

Right, good question! So in `release.yml` job `deploy`, it has setting 
```
environment: "${{inputs.environment}}"
```
The `inputs.environment` will resolve to something like "qa_deploy" (since in `release-qa.yml` job `deploy` we have
```
    with:
      environment: "qa_deploy"
```
). My understanding is that `release.yml` job `deploy` will therefore be able to access the secrets in its specified environment. 

The current/prior design was unnecessarily over-specifying what the secrets values were. I think while designing these workflow files, I may have been handling the need to pass on secrets from _different secret names_. (Like "qa_secret_key" and "live_secret_key".) But eventually that collapsed to being the same secret name for both QA and Live (Like just "secret_key"). The need for different secret names at the time may have been influential in getting the secrets during the first workflow file (eg release-qa.yml) and passing them to the called workflow file (release.yml). But then that need went away.

But in the end I don't think that was even necessary, since the called workflow file  (release.yml) specifies the environment and so should be able to query the secrets from it. 

I may be easily proved wrong when this gets run, but that is my current thinking.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3045)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @marksvc)

___
*[`.github/workflows/release-live.yml` line 97 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3045#-OK2Y3Ch4laS8p_PNoJ7:-OK2aVz-6XgjHh-mq1R0:b7791zm) ([raw file](https://github.com/sillsdev/web-xforge/blob/32a97ff3641d4309cf1369de88014aa7461224fa/.github/workflows/release-live.yml#L97)):*

Right, good question! So in `release.yml` job `deploy`, it has setting 
```
environment: "${{inputs.environment}}"
```
The `inputs.environment` will resolve to something like "qa_deploy" (since in `release-qa.yml` job `deploy` we have
```
    with:
      environment: "qa_deploy"
```
). My understanding is that `release.yml` job `deploy` will therefore be able to access the secrets in its specified environment. 

The current/prior design was unnecessarily over-specifying what the secrets values were. I think while designing these workflow files, I may have been handling the need to pass on secrets from _different secret names_. (Like "qa_secret_key" and "live_secret_key".) But eventually that collapsed to being the same secret name for both QA and Live (Like just "secret_key"). The need for different secret names at the time may have been influential in getting the secrets during the first workflow file (eg release-qa.yml) and passing them to the called workflow file (release.yml). But then that need went away.

But in the end I don't think that was even necessary, since the called workflow file  (release.yml) specifies the environment and so should be able to query the secrets from it. 

I may be easily proved wrong when this gets run, but that is my current thinking.
</blockquote></details>

Ok, I see. Let's merge it and give it a try so we can quickly see if it does what we want.

marksvc (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3045)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`.github/workflows/release-live.yml` line 97 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3045#-OK2Y3Ch4laS8p_PNoJ7:-OK2a7O26qWpahCP9R1j:b23tmda) ([raw file](https://github.com/sillsdev/web-xforge/blob/32a97ff3641d4309cf1369de88014aa7461224fa/.github/workflows/release-live.yml#L97)):*

Ok, I see. Let's merge it and give it a try so we can quickly see if it does what we want.
</blockquote></details>

In the workflow file currently on master, the `release-qa.yml` job `deploy` fetches the secret from the environment, and then passes it _as a secret_ to the called workflow job. 
So the syntax of
```
secrets:
  my_secret: ${{ secrets.my_secret }}
```
doesn't make the `my_secret` available to `release-qa.yml` job `deploy`, but rather _passes it_ as a secret to the called job. The intention was that it would make it available as a secret to `release.yml` called job `deploy`. And it presumably does. But the called job `deploy` has its own means to access the secrets because of its `environment` setting.

marksvc:
My solution was wrong. It is addressed by a follow up PR #3048.

PR #3047, opened 2025-02-26
Fix editor NG0955 warning

siltomato:
This PR fixes NG0955 warnings due to loop track expression targeting `tab.type`, since there can be multiple tabs of the same type, such as multiple resource tabs.  You can reproduce this warning by adding multiple resource tabs.

Added an `id` prop to `TabInfo` for angular to track instead.  The `id` prop is assigned a UUID upon tab creation.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3047)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.html
@@ -11,7 +11,7 @@
   (tabMove)="moveTab($event)"
 ></app-tab-group-header>
 
-@for (tab of tabs; track tab; let i = $index) {
+@for (tab of tabs; track tab.id; let i = $index) {
   <app-tab-body [active]="selectedIndex === i">
     <ng-container *ngTemplateOutlet="tab.contentTemplate"></ng-container>
   </app-tab-body>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.component.spec.ts
@@ -1,13 +1,14 @@
 import { QueryList } from '@angular/core';
 import { ComponentFixture, TestBed } from '@angular/core/testing';
 import { of } from 'rxjs';
+import { v4 as uuid } from 'uuid';
 import { TestTranslocoModule } from 'xforge-common/test-utils';
 import { TabMenuService } from '../../shared/sf-tab-group';
 import { TabAddRequestService } from './base-services/tab-add-request.service';
 import { TabFactoryService } from './base-services/tab-factory.service';
 import { SFTabsModule } from './sf-tabs.module';
 import { TabGroupComponent } from './tab-group.component';
-import { TabStateService } from './tab-state/tab-state.service';
+import { TabInfo, TabStateService } from './tab-state/tab-state.service';
 import { TabComponent } from './tab/tab.component';
 
 describe('TabGroupComponent', () => {
@@ -59,7 +60,8 @@ describe('TabGroupComponent', () => {
 
   it('should add tab using TabFactory and TabStateService when addTab is called', async () => {
     const newTabType = 'test';
-    const tab = {
+    const tab: TabInfo<string> = {
+      id: uuid(),
       type: 'test',
       headerText: 'Tab Header',
       closeable: false,
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.stories.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/sf-tab-group/tab-group.stories.ts
@@ -1,6 +1,7 @@
 import { Component, Input, OnChanges } from '@angular/core';
 import { Meta, moduleMetadata, StoryObj } from '@storybook/angular';
 import { Observable, of } from 'rxjs';
+import { v4 as uuid } from 'uuid';
 import {
   SFTabsModule,
   TabFactoryService,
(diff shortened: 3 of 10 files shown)


kylebuss (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3047#-:-OK2z0ZT9ICBIcuiRSQH:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

RaymondLuong3:
There is a bunch of warnings in the build log for the Build and test action. I can't tell if that is what is failing the build. We probably want to sort out those warnings

siltomato (commented):
Yeah, I looked into that, and it looks like angular doesn't like track by object ref.  I'm not sure why the warnings are only showing in the tests.

Since `TabInfo` doesn't have an id prop, I created a composite key to track by.  I changed the loop to track by `tab.type + tab.headerText`.  That should satisfy both the NG0955 and NG0956 warnings.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3047#-:-OK378r8-ew1ezmFh2oO:b-dnih72)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @kylebuss)

kylebuss (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3047)* status: all files reviewed, 1 unresolved discussion (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html` line 63 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3047#-OK3G_j0-LkD-XZqL5mR:-OK3G_j10kxkjfeBibeX:b-6bsb0m) ([raw file](https://github.com/sillsdev/web-xforge/blob/aa61b09148114e68d40dba7eeb55ef5ec1117b72/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html#L63)):*
> ```HTML
>         [connectedTo]="(tabState.groupIds$ | async) ?? []"
>       >
>         @for (tab of tabGroup.value.tabs; track tab.type + tab.headerText) {
> ```

I'm recalling that object level tracking was changed to require a unique key when we updated Angular versions. Let's use `track $index` instead of the proposed object key. This should prevent the NG warnings we are getting, as you can still observe the warning when opening multiple history tabs with the same date even with the `tab.type + tab.headerText` change.



_Code quote:_
```HTML
track tab.type + tab.headerText
```

siltomato (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3047)* status: all files reviewed, 1 unresolved discussion (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html` line 63 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3047#-OK3G_j0-LkD-XZqL5mR:-OK3LsB802Vq3ioCU_xk:bxfcfr9) ([raw file](https://github.com/sillsdev/web-xforge/blob/aa61b09148114e68d40dba7eeb55ef5ec1117b72/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html#L63)):*

I'm recalling that object level tracking was changed to require a unique key when we updated Angular versions. Let's use `track $index` instead of the proposed object key. This should prevent the NG warnings we are getting, as you can still observe the warning when opening multiple history tabs with the same date even with the `tab.type + tab.headerText` change.
</blockquote></details>

Ah, I see what you mean with the identical history tabs.

One problem with `track $index`, though, is that tab order can change.  Reordered tabs will get re-rendered when tab position changes in the group if we track index.  Aside from performance issues, this will reset any unpersisted tab state, such as history tab state.

It feels like we need an id prop.  I'll take a look at what that would entail...

siltomato (commented):
After considering @kylebuss comments, I added an `id` prop to `TabInfo`.  We can `track tab.id` now, which should be ideal.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3047#-:-OK3XzsvA1SCVzX5k1Ab:bjo8qdl)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @kylebuss)

siltomato (commented):
Also added an `id` prop to `TabComponent` as running tests were showing NG0956 warnings for the loop in the tab group component template.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3047#-:-OK3mW42C5vEz4rhotW4:b-slx8ff)* status: 0 of 10 files reviewed, 1 unresolved discussion (waiting on @kylebuss)

kylebuss (approved):
:lgtm:

> Reviewed 8 of 8 files at r3, 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3047#-:-OK6b04K0UFpOvsUJrN8:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.spec.ts` line 24 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/3047#-OK6aNgs3itq0Cr9qHFQ:-OK6aNgs3itq0Cr9qHFR:b-62byzi) ([raw file](https://github.com/sillsdev/web-xforge/blob/f74100fdca1fb0760f30261cf5c14f8a5a996900/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.spec.ts#L24)):*
> ```XML
>   it('should create a "biblical terms" tab', async () => {
>     const tab = await service.createTab('biblical-terms');
>     expect(tab.id?.length).toBeTruthy();
> ```

nit This works, but wouldn't `expect(tab.id).not.toBeNull()` also have worked here?



_Code quote:_
```XML
expect(tab.id?.length).toBeTruthy();
```

siltomato (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3047)* status: all files reviewed, 1 unresolved discussion (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.spec.ts` line 24 at r4](https://reviewable.io/reviews/sillsdev/web-xforge/3047#-OK6aNgs3itq0Cr9qHFQ:-OK6iL811ptcIksjOHlO:bhk4btw) ([raw file](https://github.com/sillsdev/web-xforge/blob/f74100fdca1fb0760f30261cf5c14f8a5a996900/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/tabs/editor-tab-factory.service.spec.ts#L24)):*

nit This works, but wouldn't `expect(tab.id).not.toBeNull()` also have worked here?
</blockquote></details>

I wanted to also check that it wasn't an empty string, but 'truthy' probably isn't the best way.  I changed it to be more explicit.

kylebuss (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3047)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

PR #3052, opened 2025-03-03
SF-3212 Create 'unknown' blot as fallback blot

siltomato:
This PR creates a custom `ScrollBlot` to override Quill's scroll blot and an `UnknownBlot` to render when an unknown blot type is attempted.

If an unknown blot type creation is attempted, the custom `ScrollBlot` catches the error and creates the `UnknownBlot`.  After the editor renders (via `setTimeout()`), an error is thrown displaying the name of the failed attempted blot.

`UnknownBlot` uses `tagName = 'sf-unknown'`, ('span' was causing it to be incorrectly matched from a quill registry query and throw an `after.appendChild not a function` error).

You can test this blot by adding TNN as a resource.

Example:
image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3052)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/_text.scss
@@ -288,3 +288,14 @@ usx-segment {
     font-family: Roboto, sans-serif;
   }
 }
+
+// Embed blot for unknown attributes
+sf-unknown {
+  display: inline-block;
+  color: #999;
+  background: #f0f0f0;
+  padding: 0 4px;
+  border-radius: 4px;
+  font-style: italic;
+  user-select: none;
+}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-editor-registration/quill-formats/quill-blots.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/text/quill-editor-registration/quill-formats/quill-blots.spec.ts
@@ -15,6 +15,7 @@ import {
   RefInline,
   SegmentInline,
   TextAnchorInline,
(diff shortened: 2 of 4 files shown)


josephmyers:
I haven't done a full review yet, but I think it's less than ideal to utilize the generic exception dialog for this, since that's typically reserved for errors that aren't handled, as a last resort. Per the JIRA description, we want to be notified of new blots that we need to support; however, we can be a little more elegant than the last-resort exception dialog. You could:

- Log an error to the console for the least intrusion
- Show an error "notice" as a middle ground

I'd think the former, since a dialog or even an error notice showing every single time would get pretty old pretty fast to users. What do you think?

siltomato (commented):
Yes, I think you are right.  The error dialog does get old fast.  The idea was to alert (still rendering the text) so that the missing blot could be remedied, but perhaps the visual presence in the text is sufficient (along with logging to the console).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3052#-:-OKRF7Wa4tcN6nxgjJDq:bhl4fad)* status: 0 of 4 files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (approved):
:lgtm: Thanks. I've tested this with resources and projects, and it seems to handle things as expected. Awesome!

> Reviewed 3 of 4 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3052#-:-OKUQTdY5VjTxZkOEVbE:bkle2pd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @siltomato)

PR #3075, opened 2025-03-10
SF-3209 Enable "manage audio" for users without questions permission

josephmyers:
The software supports a bit of an odd state, in my opinion, where Paratext users can be allowed to manage audio without being able to manage questions, and in some cases even add answers.

The direction we're going to head is to move the audio management wholly into the Chapter Audio dialog and out of the Overview page.

This fixes the bug where non-Admin PT users who are given permission to manage audio aren't able to delete audio.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3075)


kylebuss (changes requested):
I'm on the fence with this solution. It solves the problem, but creates other concerns. The `checking-overview.component.html` will display either "Manage Questions" or "My Progress" based on the user having add/edit questions permission. This solution adds a third and fourth option to that component "Manage Audio" (only audio permissions) and "Manage Questions & Audio" (both questions and audio permissions). 

I guess the question is whether we should fix this issue with your solution and update the `checking-overview` and sidebar navigation to handle each case or make a general improvement to managing scripture audio and keep it separate from the questions/progress?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-:-OL-MlcN1U5J-0EDi8Zj:bk64hv8)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (commented):
Yeah, thanks for the review. The bigger problem, that managing audio is hand-jammed into the Overview page, is not addressed here. But I spoke with Nathaniel about this, and he recommended we go ahead and fix it now. So I'll look into allowing management to happen from the dialog, at each chapter, and just removing the hand-jammed audio management from the Overview page. That should result in a better workflow all-around

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-:-OL2DIHWBNgSFsnD9GaB:brr6517)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers:
Keep an eye out for a bug where the chapter audio never loads. I'm wondering if this is an actual bug or just local machine weirdness with audio

josephmyers:
Balsamiq design: https://balsamiq.cloud/sghq53/pejrlz8/r8C9B/bAF36

kylebuss (commented):
Awesome job on this! There were a couple of things that weren't immediately intuitive but I was able to figure them out. I did not run into an issue with the Chapter audio not loading and I used both Chrome and Firefox.

> Reviewed 1 of 2 files at r2, 8 of 8 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-:-OMgJHkG77-Ad6wOWZT-:b2qp1i8)* status: all files reviewed, 6 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.html` line 31 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-OMgAYvt9D9n9IyfCx9i:-OMgAYvt9D9n9IyfCx9j:b-lu7z62) ([raw file](https://github.com/sillsdev/web-xforge/blob/55c6dee66731a9a14ba0d72e1c13aa0867763693/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.html#L31)):*
> ```HTML
>       <span class="verse-label">{{ t("segment") }} {{ currentSegmentDisplay }}/{{ numberOfSegments }}</span>
>     }
>   </div>
> ```

I think this would be valuable to show when playing the audio and not only when managing it. Is it safe to assume most projects will segment there audio timing based on chapter verse? If that is the case, could we us "verse" instead of "segment"?

I also wonder if we should have a "delete" icon that allows admins and manage audio users to delete existing audio/timing data as well?



_Code quote:_
```HTML
  <div class="footer">
    <span class="verse-label">{{ verseLabel }}</span>
    @if (_showSegment) {
      <span class="verse-label">{{ t("segment") }} {{ currentSegmentDisplay }}/{{ numberOfSegments }}</span>
    }
  </div>
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.html` line 93 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-OMgG65Y5t3HqfmzReCw:-OMgG65Y5t3HqfmzReCx:b-vc28hi) ([raw file](https://github.com/sillsdev/web-xforge/blob/55c6dee66731a9a14ba0d72e1c13aa0867763693/src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.html#L93)):*
> ```HTML
>         <mat-spinner diameter="24"></mat-spinner>
>       </div>
>     }
> ```

When I first tested this the uploading message displayed for a while so you might need to test on a slower internet connection. I wonder if the position/spacing of this needs tweaking.





_Code quote:_
```HTML
    @if (isLoadingAudio) {
      <div class="uploading">
        <span>{{ t("uploading") }}</span>
        <mat-spinner diameter="24"></mat-spinner>
      </div>
    }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.html` line 99 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-OMg8z-i5aGUV50k9oi3:-OMg8z-i5aGUV50k9oi4:b3mhk9v) ([raw file](https://github.com/sillsdev/web-xforge/blob/55c6dee66731a9a14ba0d72e1c13aa0867763693/src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.html#L99)):*
> ```HTML
>     <button mat-flat-button color="primary" type="submit" (click)="save()" id="audio-save-btn" [disabled]="!isOnline">
>       {{ t("save") }}
>     </button>
> ```

I think we should only show a Save button in the `allFieldsValid` screen. I would suggest making this a "Delete Existing Audio" button if `_selectionHasAudioAlready`, since that is the behavior from this screen. Below are additional comments on validating when files are missing.



_Code quote:_
```HTML
    <button mat-flat-button color="primary" type="submit" (click)="save()" id="audio-save-btn" [disabled]="!isOnline">
      {{ t("save") }}
    </button>
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.ts` line 269 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-OMgAMWQ9mmx6OBuufH7:-OMgAMWQ9mmx6OBuufH8:bpzq10h) ([raw file](https://github.com/sillsdev/web-xforge/blob/55c6dee66731a9a14ba0d72e1c13aa0867763693/src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.ts#L269)):*
> ```XML
>       !this.hasTimingDataError &&
>       this.book != null &&
>       this.chapter !== null
> ```

Do we need a strict `!== null` check here?



_Code quote:_
```XML
this.chapter !== null
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.ts` line 288 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-OMg9h5NA1EopWqob7iE:-OMg9h5NA1EopWqob7iF:b5be4ia) ([raw file](https://github.com/sillsdev/web-xforge/blob/55c6dee66731a9a14ba0d72e1c13aa0867763693/src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.ts#L288)):*
> ```XML
>     if (!this.isAudioUploaded && !canSave) {
>       this._audioErrorKey = 'no_audio_file_uploaded';
>     }
> ```

Could we validate this on file upload and remove the save option from the upload screen and only show it on the `allFieldsValid` screen? An example would be the user uploads audio file but not timing then we would display the "No timing data uploaded" message. 



_Code quote:_
```XML
    if (!this.isTimingUploaded && !canSave) {
      this._timingErrorKey = 'no_timing_data_uploaded';
    }
    if (!this.isAudioUploaded && !canSave) {
      this._audioErrorKey = 'no_audio_file_uploaded';
    }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json` line 259 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-OMgEc4W88LH8YnWGNHd:-OMgEc4X1KVHf0QMc649:bnd7jq) ([raw file](https://github.com/sillsdev/web-xforge/blob/55c6dee66731a9a14ba0d72e1c13aa0867763693/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json#L259)):*
> ```JSON
>     "audio_cannot_be_played": "This audio cannot be played while you are offline because it has not been downloaded. Connect to the internet to play.",
>     "audio_cannot_be_previewed": "The audio is in a format that cannot be previewed. Save any changes and connect to the internet to play.",
>     "segment": "Segment"
> ```

See above comment about if we can use "verse" instead of "segment".



_Code quote:_
```JSON
"segment": "Segment"
```

kylebuss (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3075)* status: all files reviewed, 4 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.html` line 93 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-OMgG65Y5t3HqfmzReCw:-OMlEm4Z9nll16NA50Wf:b-rlob0q) ([raw file](https://github.com/sillsdev/web-xforge/blob/55c6dee66731a9a14ba0d72e1c13aa0867763693/src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.html#L93)):*

Yeah, I agree, and it's not great on master, either. I improved it, I think
</blockquote></details>

I like this!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.html` line 99 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-OMg8z-i5aGUV50k9oi3:-OMlCRLe03uwsqvsHmYj:b-kjdmfi) ([raw file](https://github.com/sillsdev/web-xforge/blob/55c6dee66731a9a14ba0d72e1c13aa0867763693/src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.html#L99)):*

I'm wondering what problem this is intended to solve. Maybe just that the Save button doesn't save until their fields are filled? If we remove the Save button initially, users will be wondering how to save, or if they even need to. It's typically better UX to allow the user to see the primary buttons from the start, potentially interacting with them and receiving further guidance as necessary.

Hopefully I'm understanding you. You're also suggesting to rename Save to Delete Existing Audio, if we launch with audio present, instead of having both. I think we could do this, with some rework, but there are a number of scenarios to think through (e.g. deleting and then adding different audio). I don't like the idea of making Delete Existing Audio a primary button like Save. Delete isn't a great primary action. Further, I think the consistency, of having a Save, between the two "modes" will be helpful, in that at any time the user knows they can save or discard their work.

If your primary motivation is to remove Save from this second screen (for when they already have audio), we could change it to Close or something similar. But I think the header and footer of the dialog help achieve familiarity for users swapping audio back and forth.
</blockquote></details>

Primary motivation was changing the name on the first screen if there was existing audio and keeping it as Save on the second page.

Scenarios I was thinking:

No audio - First page displays Save but doesn't provide that function. If you click it when no files are attached it shows the errors mentioned in my related comment. Once both files are attached you're taken to the second page where you can actually Save the files to the project.

Existing Audio - When you open the manage audio dialog you are taken to the second page. If you delete both files you can Save from the first page but the function is actually deleting the audio from the project.

I agree with your points, the second scenario is the more problematic one. It's not straight forward that clicking save in this state will delete the existing audio from the project. 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.ts` line 288 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-OMg9h5NA1EopWqob7iE:-OMlBcY6D6wY1Um_J2Qm:b-hundwm) ([raw file](https://github.com/sillsdev/web-xforge/blob/55c6dee66731a9a14ba0d72e1c13aa0867763693/src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.ts#L288)):*

I'm not entirely sure what you mean by "validate" or "file upload." Are you suggesting you want to validate the files or that both files are present? And you want to do this when they select the file(s), or when they click Save?

We could do a lot of things. I'm operating under the assumption that the dialog was previously working pretty well and that we shouldn't change it unless we have a good reason. I think the current behavior works well for the example you gave.
</blockquote></details>

This was related to the above comments on the save button on the first page of the dialog.

It would be after the user selected the files we would check if both were uploaded and show any errors rather than not displaying the error until the user clicks save.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.html` line 31 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-OMgAYvt9D9n9IyfCx9i:-OMlAeI62_hxx-EPrDjR:bugkfkb) ([raw file](https://github.com/sillsdev/web-xforge/blob/55c6dee66731a9a14ba0d72e1c13aa0867763693/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.html#L31)):*

The average user does not care about the segment. They care about the verse, which we're displaying already. I'm a little confused about your thinking as to why they would want to see this. This is pretty decidedly administrative information. Or are you suggesting we show the segment on the chapter audio player (in the main page) for Admin's and anyone with audio management, in addition to this dialog?
</blockquote></details>

Thanks for clarify this for me. I'm not seeing the verse change but it's possible it has to do with my timing file? It only displays verse 1 when I tried segmenting verse 1 and 2.

kylebuss (commented):
Hopefully I've clarified a couple of my comments. Changes look good, I did notice an issue with the chapter display when the audio plays but that might just be a me issue.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-:-OMlFsO64jC0k9lmdkLv:br4r6pc)* status: all files reviewed, 4 unresolved discussions (waiting on @josephmyers)

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3075)* status: 8 of 12 files reviewed, 4 unresolved discussions (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.html` line 93 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-OMgG65Y5t3HqfmzReCw:-OMoCsNPEB5X-ig7BRlD:b-txi5ci) ([raw file](https://github.com/sillsdev/web-xforge/blob/55c6dee66731a9a14ba0d72e1c13aa0867763693/src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.html#L93)):*

I like this!
</blockquote></details>

Thanks, me too
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.html` line 99 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-OMg8z-i5aGUV50k9oi3:-OMoYvcC63izFGv3Gzj7:b-l2ki85) ([raw file](https://github.com/sillsdev/web-xforge/blob/55c6dee66731a9a14ba0d72e1c13aa0867763693/src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.html#L99)):*

Primary motivation was changing the name on the first screen if there was existing audio and keeping it as Save on the second page.

Scenarios I was thinking:

No audio - First page displays Save but doesn't provide that function. If you click it when no files are attached it shows the errors mentioned in my related comment. Once both files are attached you're taken to the second page where you can actually Save the files to the project.

Existing Audio - When you open the manage audio dialog you are taken to the second page. If you delete both files you can Save from the first page but the function is actually deleting the audio from the project.

I agree with your points, the second scenario is the more problematic one. It's not straight forward that clicking save in this state will delete the existing audio from the project.
</blockquote></details>

Think of it like how the other audio portions of our app function, e.g. editing questions and also editing answers. The deletion needs to be a cancel-able and save-able action. For an answer with audio, the user can delete the audio, but then cancel that action.

Here's a video of this happening for answers:
[Recording 2025-04-02 103939.mp4](https://files.reviewableusercontent.io/images/reviewable/43733878/9vGOqVRNdPibTgK0dl8amejh/Recording%202025-04-02%20103939.mp4)

And here's editing a question:
[Recording 2025-04-02 104547.mp4](https://files.reviewableusercontent.io/images/reviewable/43733878/3oZe47mfEfMYVUSm5pqHYC0x/Recording%202025-04-02%20104547.mp4)

Maybe the fact that the dialog sizes changes is confusing and making you think that these are separate pages? This is all one dialog, with one Save and Cancel. I've updated the dialog to stay the same size during deletion. See if this helps
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.ts` line 288 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-OMg9h5NA1EopWqob7iE:-OMoeM-G-BWXjPvhj00L:bxbna6) ([raw file](https://github.com/sillsdev/web-xforge/blob/55c6dee66731a9a14ba0d72e1c13aa0867763693/src/SIL.XForge.Scripture/ClientApp/src/app/checking/chapter-audio-dialog/chapter-audio-dialog.component.ts#L288)):*

This was related to the above comments on the save button on the first page of the dialog.

It would be after the user selected the files we would check if both were uploaded and show any errors rather than not displaying the error until the user clicks save.
</blockquote></details>

The reason we didn't do this in the past was to avoid showing an error if the user wanted to upload one at a time. This is especially important for drag-and-drop, as well as for users who don't know that you can select multiple files with a single "select file" dialog (most of them)
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.html` line 31 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-OMgAYvt9D9n9IyfCx9i:-OMoeuiQ4Xsf5rmJRWxj:bpyo78b) ([raw file](https://github.com/sillsdev/web-xforge/blob/55c6dee66731a9a14ba0d72e1c13aa0867763693/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.html#L31)):*

Thanks for clarify this for me. I'm not seeing the verse change but it's possible it has to do with my timing file? It only displays verse 1 when I tried segmenting verse 1 and 2.
</blockquote></details>


Make sure your timing file is properly formatted. The verse display is kinda already working (I haven't changed it). I'm going to upload a legit timing file I received back when we initially worked on all this

josephmyers:
[MRK_004_Label_Format.txt](https://github.com/user-attachments/files/19561200/MRK_004_Label_Format.txt)

josephmyers:
It's possible there's an issue with verse display. Can you check master to see if it's behaving the same way?

josephmyers (commented):
I've implemented the recommendation to show the chapter player beneath the inputs, instead of replacing them. Take a look and let me know what you think

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-:-ONmt5KaD5CkqHnoSNaA:bxm0vj1)* status: 3 of 12 files reviewed, 4 unresolved discussions (waiting on @kylebuss)

kylebuss (approved):
:lgtm: One small nit would be if we could add a clearer way to delete both files if that is the action the user wants rather than having to click the delete for each of the selected files.

> Reviewed 1 of 2 files at r6, 9 of 9 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-:-ONoaZKiEWMvPkDhjHRb:b-jkm74d)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.html` line 31 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3075#-OMgAYvt9D9n9IyfCx9i:-OMp__mi9LFSEKGIvhwV:b49u28x) ([raw file](https://github.com/sillsdev/web-xforge/blob/55c6dee66731a9a14ba0d72e1c13aa0867763693/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-scripture-audio-player/checking-scripture-audio-player.component.html#L31)):*

Make sure your timing file is properly formatted. The verse display is kinda already working (I haven't changed it). I'm going to upload a legit timing file I received back when we initially worked on all this
</blockquote></details>

Thanks for the audio file, this format is working to change the verse. I'm not sure why the format I used works to upload segments but doesn't change the verse? Anyways, this looks good.

PR #3077, opened 2025-03-10
Remove a number of unneeded non-null assertions

Nateowami:
I applied the Grit pattern, then checked out all files that had build failures. That means there are *lots* of unneeded non-null assertions in files that also have non-null assertions that _are_ needed, and can't be automatically removed as easily.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3077)

### DIFF
--- a/.grit/patterns/remove_non_null_assertion.md
+++ b/.grit/patterns/remove_non_null_assertion.md
@@ -1,0 +1,13 @@
+# Remove non-null assertion
+
+Removes all non-null assertions, without regard to whether they're needed or not. One use for this is to remove all
+non-null assertions from certain files and then only add them back in where they're needed.
+
+```grit
+language js
+
+non_null_expression() as $nonNullAssertion where {
+  $nonNullAssertion <: contains or { member_expression(), identifier() } as $expression,
+  $nonNullAssertion => $expression
+}
+```
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-input-form/checking-input-form.component.ts
@@ -95,8 +95,8 @@ export class CheckingInputFormComponent {
     const verseRef = this._questionDoc.data.verseRef;
 
     const dialogData: TextChooserDialogData = {
-      bookNum: (this.verseRef && this.verseRef.bookNum) || verseRef!.bookNum,
-      chapterNum: (this.verseRef && this.verseRef.chapterNum) || verseRef!.chapterNum,
+      bookNum: (this.verseRef && this.verseRef.bookNum) || verseRef.bookNum,
+      chapterNum: (this.verseRef && this.verseRef.chapterNum) || verseRef.chapterNum,
       textsByBookId: this.textsByBookId,
       projectId: this._questionDoc.data.projectRef,
       isRightToLeft: this.project?.isRightToLeft,
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.spec.ts
@@ -101,7 +101,7 @@ describe('CheckingQuestionComponent', () => {
         audioUrl: 'test-audio-player-b.webm',
         projectRef: 'project01',
         text: 'another question',
-        verseRef: env.component.questionDoc.data!.verseRef!
+        verseRef: env.component.questionDoc.data!.verseRef
       }
     } as QuestionDoc;
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/question-dialog/question-dialog.service.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/question-dialog/question-dialog.service.ts
@@ -74,7 +74,7 @@ export class QuestionDialogService {
       await questionDoc.submitJson0Op(op =>
         op
           .set(q => q.verseRef, verseRefData)
-          .set(q => q.text!, text)
+          .set(q => q.text, text)
           .set(q => q.audioUrl, audioUrl)
           .set(q => q.dateModified, currentDate)
       );
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/models/question-doc.ts
@@ -70,12 +70,12 @@ export class QuestionDoc extends ProjectDataDoc<Question> {
 
     if (this.data.dataId === dataId) {
       // The file belongs to the question
-      return obj<Question>().path(q => q.audioUrl!);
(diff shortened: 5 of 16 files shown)


pmachapman (changes requested):
Just one question!

> Reviewed 15 of 15 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3077#-:-OL0QtRJ12UbKBQGGMTY:b-w58e4u)* status: all files reviewed, 1 unresolved discussion (waiting on @Nateowami)

___
*[`.grit/patterns/remove_non_null_assertion.md` line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3077#-OL0Q5FK8-MgjP6QW460:-OL0Q5FK8-MgjP6QW461:bto6ojo) ([raw file](https://github.com/sillsdev/web-xforge/blob/da64023833e8e94d5bf2ab5966af059d6f44190f/.grit/patterns/remove_non_null_assertion.md#L8)):*
> ```Markdown
> ```grit
> function removeExclamation($expression) js {
>   return $expression.text.slice(0, $expression.text.length - 1);
> ```

This seems a bit broad, like it will always chop the last character off? Should you check that the last character is an exclamation mark before slicing?



_Code quote:_
```Markdown
return $expression.text.slice(0, $expression.text.length - 1);
```

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3077#-:-OL0UrroBFuQBf72OFAh:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

pmachapman (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3077)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3077)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`.grit/patterns/remove_non_null_assertion.md` line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3077#-OL0Q5FK8-MgjP6QW460:-OL0U48O6eASEhpfeoYL:by5wna4) ([raw file](https://github.com/sillsdev/web-xforge/blob/da64023833e8e94d5bf2ab5966af059d6f44190f/.grit/patterns/remove_non_null_assertion.md#L8)):*

This seems a bit broad, like it will always chop the last character off? Should you check that the last character is an exclamation mark before slicing?
</blockquote></details>

Given this pattern it intended to be run manually, and the results checked, I would be more concerned about silently skipping things it should have matched (and having no way of realizing it) than having it incorrectly rewrite things in a way that is obviously wrong.

Here's a more idiomatic way of writing it (though it does introduce a different probably much worse, now that I think about it, assumption: That every non_null_expression contains either a member_expression (`a.b`) or identifier (`a`), as the nearest descendant.

I think if we had something like:
``` ts
someFunction(a.b)!.c
```
it might now rewrite it as:
``` ts
a.b.c
```
(Actually upon testing it rewrites it to `someFunction.c`, because `someFunction` is an identifier)

I think "the last character in a non-null assertion is !" is actually one of the most robust assumptions that can be made.

Nateowami:
@pmachapman Can you look at my comments regarding why the new solution is actually much worse?

pmachapman (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3077)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`.grit/patterns/remove_non_null_assertion.md` line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3077#-OL0Q5FK8-MgjP6QW460:-OL0YYf01PXoAMm_YoNe:bxp5i3g) ([raw file](https://github.com/sillsdev/web-xforge/blob/da64023833e8e94d5bf2ab5966af059d6f44190f/.grit/patterns/remove_non_null_assertion.md#L8)):*

Given this pattern it intended to be run manually, and the results checked, I would be more concerned about silently skipping things it should have matched (and having no way of realizing it) than having it incorrectly rewrite things in a way that is obviously wrong.

Here's a more idiomatic way of writing it (though it does introduce a different probably much worse, now that I think about it, assumption: That every non_null_expression contains either a member_expression (`a.b`) or identifier (`a`), as the nearest descendant.

I think if we had something like:
``` ts
someFunction(a.b)!.c
```
it might now rewrite it as:
``` ts
a.b.c
```
(Actually upon testing it rewrites it to `someFunction.c`, because `someFunction` is an identifier)

I think "the last character in a non-null assertion is !" is actually one of the most robust assumptions that can be made.
</blockquote></details>

OK no problem.

My only thought when I read your original code is that we should probably check that `$expression.text.substr($expression.text.length - 1) === '!'` before running `$expression.text.slice(0, $expression.text.length - 1)`. If you don't think that's necessary, than :lgtm_strong:

Nateowami (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3077#-:-OL0_0N10f26B0kDuJJr:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`.grit/patterns/remove_non_null_assertion.md` line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3077#-OL0Q5FK8-MgjP6QW460:-OL0ZbbB3Y7Ddf9bnxf-:b-2jrvt7) ([raw file](https://github.com/sillsdev/web-xforge/blob/da64023833e8e94d5bf2ab5966af059d6f44190f/.grit/patterns/remove_non_null_assertion.md#L8)):*

OK no problem.

My only thought when I read your original code is that we should probably check that `$expression.text.substr($expression.text.length - 1) === '!'` before running `$expression.text.slice(0, $expression.text.length - 1)`. If you don't think that's necessary, than :lgtm_strong:
</blockquote></details>

We can add a check, but then what do we do? Silently do nothing?

pmachapman (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3077)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

___
*[`.grit/patterns/remove_non_null_assertion.md` line 8 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3077#-OL0Q5FK8-MgjP6QW460:-OL0dGhfBq7Y6LBHD81b:bsv5w59) ([raw file](https://github.com/sillsdev/web-xforge/blob/da64023833e8e94d5bf2ab5966af059d6f44190f/.grit/patterns/remove_non_null_assertion.md#L8)):*

We can add a check, but then what do we do? Silently do nothing?
</blockquote></details>

I don't know - I only raised it in case it was perhaps something missed.

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3077)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #3080, opened 2025-03-11
Fix use of QuietDestroyRef

Nateowami:
Having QuietDestroyRef inject DestroyRef does not produce a DestroyRef that is tied to the component injecting QuietDestroyRef. Instead, the DestroyRef needs to be injected in the context of the component or service actually needing the QuietDestroyRef.

### DIFF
--- a/.grit/patterns/migrate_quiet_destroy_ref.md
+++ b/.grit/patterns/migrate_quiet_destroy_ref.md
@@ -102,7 +102,7 @@ export class EditorTabAddResourceDialogService {
   constructor(
     private readonly paratextService: ParatextService,
     private readonly userProjectsService: SFUserProjectsService,
-    private readonly destroyRef: QuietDestroyRef
+    private destroyRef = getQuietDestroyRef();
   ) {
     this.userProjectsService.projectDocs$.pipe(takeUntilDestroyed(this.destroyRef)).subscribe(async projects => {
       if (projects == null) return;
--- a/.grit/patterns/take_until_destroyed.md
+++ b/.grit/patterns/take_until_destroyed.md
@@ -49,7 +49,7 @@ pattern removeExtends() {
 }
 
 pattern addProperty() {
-  maybe `constructor( $args ){  $body }` => `constructor( $args, private destroyRef: QuietDestroyRef ){$body}` where {
+  maybe `constructor( $args ){  $body }` => `constructor( $args, private destroyRef = getQuietDestroyRef(); ){$body}` where {
     $body <: within classSubscriptionDisposable(),
     $args <: not contains `QuietDestroyRef`
   }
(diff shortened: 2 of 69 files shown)


siltomato (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3080)* status: 0 of 69 files reviewed, 9 unresolved discussions (waiting on @github-advanced-security[bot] and @Nateowami)

___
*[a discussion](https://reviewable.io/reviews/sillsdev/web-xforge/3080#-OL4yHpq69t7Q1ENbYip:-OL4yHpq69t7Q1ENbYiq:b69szuo) (no related file):*
I was looking through the comment thread for the NG0911 issue in the angular repo, and I ran across a workaround that replaces `takeUntilDestroyed` instead of replacing `DestroyRef`: https://github.com/angular/angular/issues/54527#issuecomment-2098254508

Modified a little for logging, that might look like:
```ts
export function safeTakeUntilDestroyed<T>(destroyRef?: DestroyRef): MonoTypeOperatorFunction<T> {
  const originalStack = new Error().stack;

  return pipe(
    takeUntil(
      NEVER.pipe(
        takeUntilDestroyed(destroyRef),
        catchError(error => {
          if (hasStringProp(error, 'message') && error.message.includes('NG0911')) {
            console.warn('NG0911 error caught and ignored. Original stack: ', originalStack);
          } else {
            // Re-throw if it's not an NG0911 error
            throw error;
          }
          return EMPTY;
        }),
        defaultIfEmpty(null)
      )
    )
  );
}
```

One downside is that we wouldn't have access to `ErrorReportingService` unless we passed it in, which would be yucky. 

Although, we could workaround that by capturing an instance of ErrorReportingService in a static field so that if it has been created by angular DI, we could get the instance.

Something like:
```TS
@Injectable({
  providedIn: 'root'
})
export class ErrorReportingService {
  private static instance: ErrorReportingService;

  constructor() {
    // Grab the instance when constructed by angular
    ErrorReportingService.instance = this;
  }

  static getInstance(): ErrorReportingService | undefined {
    return ErrorReportingService.instance;
  }
  
  // ... rest
}
```

Then, we could add that to the previous function like:
```TS
export function safeTakeUntilDestroyed<T>(destroyRef?: DestroyRef): MonoTypeOperatorFunction<T> {
  const originalStack = new Error().stack;
  const errorReportingService = ErrorReportingService.getInstance();

  return pipe(
    takeUntil(
      NEVER.pipe(
        takeUntilDestroyed(destroyRef),
        catchError(error => {
          if (hasStringProp(error, 'message') && error.message.includes('NG0911')) {
            console.warn('NG0911 error caught and ignored. Original stack: ', originalStack);
            errorReportingService?.silentError('NG0911 error caught and ignored', { originalStack });
          } else {
            // Re-throw if it's not an NG0911 error
            throw error;
          }
          return EMPTY;
        }),
        defaultIfEmpty(null)
      )
    )
  );
}
```

Would this be a better approach than `QuietDestroyRef`?

Nateowami:
@siltomato Replacing `takeUntiDestroyed` was actually my original approach, and after working on this more today, I started going in that direction, even before seeing your comment. I think I thought injecting a different class was more elegant, but what I forgot we were losing was the ability to store the stack trace to the exact place `takeUntiDestroyed` was called. I'm creating a separate branch and new PR.

PR #3085, opened 2025-03-12
SF-3245 Fix quill editor scroll and hidden app notice

kylebuss:
This PR addresses two bugs related to the same issue of overflow within the editor container. The quill editor overflowed below the text container causing the last few lines to be hidden in cases the target chapter had a lot of text. This also created an issue where if the user scrolled down and selected a verse at the bottom when refreshing or navigating back to the editor it would hide any app notices above the text container.

This change updates the notice to remain at the top of the editor and only scroll the quill editor. One test needed updated as the width of the editor was set extremely narrow (beyond real-world cases) and failed with this change. By setting the width and heigh to a more reasonable size the testing performs as expected. Updating the app notices to `mode="fill-dark"` removes the border and creates a cleaner look as they now remain fixed at the top of the editor.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3085)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.html
@@ -92,7 +92,7 @@
               @case ("project-source") {
                 <div class="project-tab-container">
                   @if (hasSourceCopyrightBanner) {
-                    <app-notice icon="copyright" type="warning" class="copyright-banner">
+                    <app-notice mode="fill-dark" icon="copyright" type="warning" class="copyright-banner">
                       <div>
                         {{ sourceCopyrightBanner }}
                         <span class="copyright-more-info" (click)="showCopyrightNotice('source')">{{
@@ -119,12 +119,12 @@
               @case ("project-target") {
                 <div class="project-tab-container">
                   @if (writingSystemWarningBanner) {
-                    <app-notice icon="warning" type="warning" class="writing-system-warning-banner">
+                    <app-notice mode="fill-dark" icon="warning" type="warning" class="writing-system-warning-banner">
                       <span [innerHtml]="writingSystemWarningMessage"></span>
                     </app-notice>
                   }
                   @if (hasTargetCopyrightBanner) {
-                    <app-notice icon="copyright" type="warning" class="copyright-banner">
+                    <app-notice mode="fill-dark" icon="copyright" type="warning" class="copyright-banner">
                       <div>
                         {{ targetCopyrightBanner }}
                         <span class="copyright-more-info" (click)="showCopyrightNotice('target')">{{
@@ -134,35 +134,35 @@
                     </app-notice>
                   }
                   @if (!isUsfmValid && hasEditRight && invalidTags().length === 0) {
-                    <app-notice type="warning" icon="warning" class="formatting-invalid-warning">
+                    <app-notice mode="fill-dark" type="warning" icon="warning" class="formatting-invalid-warning">
                       {{ t("cannot_edit_chapter_formatting_invalid") }}
                     </app-notice>
                   }
                   @if (!isUsfmValid && hasEditRight && invalidTags().length > 0) {
-                    <app-notice type="warning" icon="warning" class="formatting-invalid-warning">
+                    <app-notice mode="fill-dark" type="warning" icon="warning" class="formatting-invalid-warning">
                       <span
                         [innerHTML]="t('cannot_edit_chapter_unsupported_usfm', { tags: invalidTags().join(', ') })"
                       ></span>
                     </app-notice>
                   }
                   @if (!dataInSync && hasEditRight) {
-                    <app-notice type="warning" icon="warning" class="out-of-sync-warning">
+                    <app-notice mode="fill-dark" type="warning" icon="warning" class="out-of-sync-warning">
                       {{ t("project_data_out_of_sync") }}
                     </app-notice>
                   }
                   @if (target.areOpsCorrupted && hasEditRight) {
-                    <app-notice type="error" icon="error" class="doc-corrupted-warning">
+                    <app-notice mode="fill-dark" type="error" icon="error" class="doc-corrupted-warning">
                       {{ t("text_doc_corrupted") }}
                       <span [innerHTML]="t('to_report_issue_email', { issueEmailLink: issueEmailLink })"></span>
                     </app-notice>
                   }
                   @if (projectTextNotEditable && hasEditRight) {
-                    <app-notice type="info" icon="info" class="project-text-not-editable">
+                    <app-notice mode="fill-dark" type="info" icon="info" class="project-text-not-editable">
                       {{ t("project_text_not_editable") }}
                     </app-notice>
                   }
                   @if (showNoEditPermissionMessage) {
-                    <app-notice type="info" icon="info" class="no-edit-permission-message">
+                    <app-notice mode="fill-dark" type="info" icon="info" class="no-edit-permission-message">
                       {{ t("no_permission_edit_chapter", { userRole: userRoleStr }) }}
                     </app-notice>
                   }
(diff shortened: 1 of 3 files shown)


RaymondLuong3 (changes requested):
The code looks good, but I was able to get into a state where we get a double scroll bar. The steps I used to reproduce this are:
1. Navigate to a text, scroll down, and click on a verse at the bottom of the screen.
2. Click the translate overview navigation button.
3. Go back to the edit and review page.
4. Click on the same verse. Then the double scroll bar appears.



> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3085#-:-OLKTZh32LiuwTF5aOOj:blqlmgk)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kylebuss)

siltomato (changes requested):
I noticed an issue where the note fab icon is not aligned with the selected verse.  To reproduce, click on a verse at the start of a long chapter, and then scroll down enough to make the fab sticky to the top, and click on another verse (like the last verse).  The note fab should be aligned with the verse.  I posted a suggested fix in the css file comments.



> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3085#-:-OLZwGflA-8_0mx297AD:b-txumuu)* status: all files reviewed, 2 unresolved discussions (waiting on @kylebuss)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss` line 164 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3085#-OLZwWYLA2z8SqXKOfsg:-OLZwWYLA2z8SqXKOfsh:bw7pd4s) ([raw file](https://github.com/sillsdev/web-xforge/blob/0d5ce843c83f598037c147ff8b9fead2a997ec6d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss#L164)):*
> ```SCSS
>   flex-direction: column;
>   height: 100%;
>   overflow: hidden;
> ```

If you keep this `overflow: hidden` and add it to text-container instead of `auto`, it seems to fix the issues.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss` line 170 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3085#-OLZwlq-2MYCyb8TG1K-:-OLZwlq-2MYCyb8TG1K0:b-l8ws0g) ([raw file](https://github.com/sillsdev/web-xforge/blob/995879f151ce0344425254fda823c76cdc4bb843/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss#L170)):*
> ```SCSS
>   position: relative;
>   height: 100%;
>   overflow: auto;
> ```

This seems to work with `overflow: hidden` as mentioned above.

kylebuss (commented):
Thanks Raymond and Colin. These should be fixed. 

Colin, I put the overflow hidden on the `quill-editor` in the `text.component.scss`, I wasn't able to reproduce the issue but based on your suggestion I'd say this is the safer place to put it. The `overflow: hidden;` in the editor component was causing the issue of the app notices being hidden.

Raymond, the change I made for this was shrinking the quill editor's height to 99%. This component is nested in multiple containers all with a 100% height. I'm assuming that how quill builds the component, it creates a bit of overflow that even when hidden creates the problem you observed.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3085#-:-OL_lR-y9Pn2NKo8FJ55:b-abl25z)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3 and @siltomato)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss` line 164 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3085#-OLZwWYLA2z8SqXKOfsg:-OL_no0mA6Dg1d0CepSE:bk4hqcv) ([raw file](https://github.com/sillsdev/web-xforge/blob/0d5ce843c83f598037c147ff8b9fead2a997ec6d/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss#L164)):*

If you keep this `overflow: hidden` and add it to text-container instead of `auto`, it seems to fix the issues.
</blockquote></details>

When Nathaniel and I looked at this together we saw unexpected behaviors with it set to hidden in both css elements. Setting it to `auto` in the text-container was the best solution. I moved this suggestion to the `quill-editor`, let me know if you are still seeing the bug.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss` line 170 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3085#-OLZwlq-2MYCyb8TG1K-:-OL_o_DR8JLC0Zj4hrMM:bckp6w3) ([raw file](https://github.com/sillsdev/web-xforge/blob/995879f151ce0344425254fda823c76cdc4bb843/src/SIL.XForge.Scripture/ClientApp/src/app/translate/editor/editor.component.scss#L170)):*

This seems to work with `overflow: hidden` as mentioned above.
</blockquote></details>

See previous comments.

siltomato (commented):
Hi Kyle, did you try putting `overflow: hidden` in both `.project-tab-container { }` and `.text-container { }`?  When I do this, the app notices are shown, the bottom text is not cut off, and I do not see the note fab issue.  I did not need to modify `text.component.scss`.

@RaymondLuong3, do you still get the double scrollbar issue if you modify editor.component.scss as I described above?

Here's an animated gif to help with reproducing the note fab issue:


> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3085#-:-OLaEPQlDJBKrmz0x3Ln:bk0xfq2)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3)

kylebuss:
I will look into this further. At first glance it does appear to be resolving both issues.

kylebuss (commented):
After further testing, Colin's suggestion of setting `overflow: hidden;` in both the project-tab-container and text-container is fixing both issues. 

It appears the default behavior of quill is to fit into its parent element, when all the project-tab-container elements would first load it displays the notice and hides the bottom of the quill editor outside of the text-container. When you would select the bottom most text and navigate away and then return, quill would "notice" there was more text and adjust itself to fit - forcing the notice to hide above the project-tab-container. By setting the text-container to hide the quill overflow this prevents the buggy behavior as it sets quills parent container overflow to the text-container and not the project-tab-container.

The bug Raymond observed was also likely due to quill's default behavior (`overflow: auto;`). Similar to the above, having the text-container set to auto creates the first scroll bar and quill loads itself within the element, but when you reach the bottom of quill, it again believes it should resize and needs to add its own scroll bar to fit.

I believe the bug Colin found with the note fab is similar to Raymond's observation it's just that the extra scroll bar is not visible but quill is behaving as if it is there which "adjusts" it's width and miscalculates where the fab should be displayed on it's vertical axis. By removing the project-container scroll bar we allow quill to fit itself within the container and not bug out the way we've observed.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3085#-:-OLdJUO9Bsdt0sAFolEX:bejwkt3)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @RaymondLuong3 and @siltomato)

siltomato (approved):
Glad we're both seeing the same thing!  UI bugs can be pesky to reproduce.

:lgtm: 

> Reviewed 2 of 3 files at r1, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3085#-:-OLdbMJh6szodjT0IljD:b-6e3kob)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kylebuss)

RaymondLuong3 (approved):
Ok, I would like to see the warning be dismissible since it can take up a good amount of vertical space on mobile. But this is an improvement and we can make it dismissible when we get complaints.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3085#-:-OLe6IL9Ax7r6WvbxUpl:b-bq0gf3)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kylebuss)

kylebuss:
That's a fair point. Maybe collapsible rather than dismissible if viewport width is less than a certain amount.

PR #3090, opened 2025-03-13
Update link on draft generation steps to config sources

kylebuss:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3090)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/language-codes-confirmation/language-codes-confirmation.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/language-codes-confirmation/language-codes-confirmation.component.html
@@ -11,8 +11,8 @@
         <p>
           @if (isProjectAdmin) {
             <transloco
-              key="language_codes_confirmation.change_source_projects_on_settings_page"
-              [params]="{ projectSettingsUrl: { route: projectSettingsUrl } }"
+              key="language_codes_confirmation.change_source_projects_on_draft_sources_configuration_page"
+              [params]="{ configSourcesUrl: { route: configSourcesUrl } }"
             ></transloco>
           } @else {
             {{ t("contact_project_administrator_to_change_settings") }}
@@ -40,8 +40,8 @@
         <p>
           @if (isProjectAdmin) {
             <transloco
-              key="language_codes_confirmation.change_source_projects_on_settings_page"
-              [params]="{ projectSettingsUrl: { route: projectSettingsUrl } }"
+              key="language_codes_confirmation.change_source_projects_on_draft_sources_configuration_page"
+              [params]="{ configSourcesUrl: { route: configSourcesUrl } }"
             ></transloco>
           } @else {
             {{ t("contact_project_administrator_to_change_settings") }}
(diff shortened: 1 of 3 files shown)


Nateowami (changes requested):
> Reviewed 1 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3090)* status: 1 of 3 files reviewed, 2 unresolved discussions

___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 446 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3090#-OLFneqC-hbNfVHvkXcC:-OLFneqC-hbNfVHvkXcD:b-2j7cyv) ([raw file](https://github.com/sillsdev/web-xforge/blob/898165548c398f017a7f8242e8e13d70f0e131fa/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L446)):*
> ```JSON
>   "language_codes_confirmation": {
>     "all_sources_should_be_same_language": "All source and reference projects should be in the same language",
>     "change_source_projects_on_settings_page": "Source projects can be changed on the [link:configSourcesUrl]configure sources page[/link]",
> ```

This does not work well when switching languages. When you add a new key in English, and it doesn't exist in the other files, it falls back to the English file and behaves how you would want. However, when you already have a localization in all the other files, it means the variables won't match if it's not in English. 

I'm guessing you were trying to keep the string so we wouldn't be losing already translated material, but that strategy doesn't actually work. When we upload the files to Crowdin we'll have to tell it to throw away existing translations because the variable name has changed.

Either we should have a completely new string, or we should keep the variable name the same.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/language-codes-confirmation/language-codes-confirmation.component.html` line 15 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3090#-OLFpCTx0pXyNmkyfZ4_:-OLFpCTx0pXyNmkyfZ4a:bkitu70) ([raw file](https://github.com/sillsdev/web-xforge/blob/898165548c398f017a7f8242e8e13d70f0e131fa/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/language-codes-confirmation/language-codes-confirmation.component.html#L15)):*
> ```HTML
>             <transloco
>               key="language_codes_confirmation.change_source_projects_on_settings_page"
>               [params]="{ configSourcesUrl: { route: configSourcesUrl } }"
> ```

This is not the only instance of this in this file. That's what's causing the build failure.

kylebuss (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3090)* status: 0 of 3 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/language-codes-confirmation/language-codes-confirmation.component.html` line 15 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3090#-OLFpCTx0pXyNmkyfZ4_:-OLFpgBo9Z2KwUGoP-L1:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/898165548c398f017a7f8242e8e13d70f0e131fa/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/language-codes-confirmation/language-codes-confirmation.component.html#L15)):*

This is not the only instance of this in this file. That's what's causing the build failure.
</blockquote></details>

Done.

Nateowami (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3090#-:-OLFyhJW7th089wewOv9:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kylebuss)

PR #3091, opened 2025-03-13
Show error state when failing to save draft sources config

Nateowami:
**NOTE:** This is a PR against `task/sf-3238`, not `master`

To test this, try saving both as an admin (which does have permission), and as a translator (which does not have permission).

Stopping translators from getting to this page will be a separate task.

The implementation here feels quite bad to me, and it doesn't feel like we're handling all the save/sync states completely properly, but it seems to work for the scenarios I'm testing. We will likely need to revisit this later.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3091)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.html
@@ -190,15 +190,15 @@
         <mat-card-title>{{ t("saving_draft_sources") }}</mat-card-title>
       </mat-card-header>
       <mat-card-content>
-        @if (getControlState("projectSettings") === ElementState.Submitting) {
-          <div class="saving-indicator">
+        <div class="saving-indicator">
+          @if (getControlState("projectSettings") === ElementState.Submitting) {
             <mat-spinner [diameter]="24" color="primary"></mat-spinner> {{ t("saving") }}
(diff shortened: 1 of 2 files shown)


RaymondLuong3 (approved):
Hiding this from translators will be the bigger piece to prevent translators from getting to this state. In my testing, it looks like the user will get an obvious error. I assume this was intended. 

Do we want to add tests to this? It might not be too hard. If this was core functionality I would say we should do it.



> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3091#-:-OLG1DOQ1pMpNuf9iRkj:bdfz632)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami:
(And yes an obvious answer was intentional)

RaymondLuong3 (approved):
I was thinking the most obvious one would be that the user sees an error icon when updating project settings fails. The other would be that they can close the panel if there is an error.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3091#-:-OLGH5oB4T8rsi7yEQfR:b-qq8ox9)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

Nateowami:
I think it would be good to have a test, but I have run out of time to add one today, and the entire saving area needs work, so I think I'll merge as-is.

PR #3097, opened 2025-03-14
SF-3252 Add endpoint to get draft sources

RaymondLuong3:
This is the first step in preparation to update our data model to store draft sources as arrays. This PR moves the logic to get the sources of a project to the backend and creates an endpoint to access the draft sources arrays.
Since draft sources are used on pages that require an internet connection (Serval admin, configure drafting, draft wizard) this does not break any existing functionality even though it requires an internet connection.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3097)


pmachapman (changes requested):
Just a few minor comments

> Reviewed 22 of 23 files at r1, 1 of 1 files at r2, all commit messages.
> Dismissed @github-advanced-security[bot] from 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3097#-:-OLW-pWY-XuywEi-oHmT:b-ac3huz)* status: all files reviewed, 5 unresolved discussions (waiting on @RaymondLuong3)

___
*[`test/SIL.XForge.Scripture.Tests/Models/DraftSourcesArraysTests.cs` line 7 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3097#-OLVtBJv58IBzCyp-D8k:-OLVtBJv58IBzCyp-D8l:blidifg) ([raw file](https://github.com/sillsdev/web-xforge/blob/4104dca616bb56a86c0bdb7611ab7ae50dd30d41/test/SIL.XForge.Scripture.Tests/Models/DraftSourcesArraysTests.cs#L7)):*
> ```Smalltalk
> namespace SIL.XForge.Scripture.Models;
> 
> struct ProjectSourceArgs
> ```

How come you use a `struct` not a `class` or `record? usually we only use structs if they are less than 16 bytes, as they are put on the stack. https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/choosing-between-class-and-struct



_Code quote:_
```Smalltalk
struct
```
___
*[`src/SIL.XForge.Scripture/Controllers/SFProjectsRpcController.cs` line 671 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3097#-OLVs-2BC5AsPGUqg6vn:-OLVs-2BC5AsPGUqg6vo:b-rxspur) ([raw file](https://github.com/sillsdev/web-xforge/blob/4104dca616bb56a86c0bdb7611ab7ae50dd30d41/src/SIL.XForge.Scripture/Controllers/SFProjectsRpcController.cs#L671)):*
> ```Smalltalk
>         {
>             _exceptionHandler.RecordEndpointInfoForException(
>                 new Dictionary<string, string> { { "method", "GetDraftConfig" }, { "projectId", projectId } }
> ```

NIT: This should be `"GetDraftSources"`.



_Code quote:_
```Smalltalk
"GetDraftConfig"
```
___
*[`src/SIL.XForge.Scripture/Models/DraftSourcesArrays.cs` line 9 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3097#-OLVvj1l5KKP9eyCWOqY:-OLVvj1l5KKP9eyCWOqZ:blt04kz) ([raw file](https://github.com/sillsdev/web-xforge/blob/4104dca616bb56a86c0bdb7611ab7ae50dd30d41/src/SIL.XForge.Scripture/Models/DraftSourcesArrays.cs#L9)):*
> ```Smalltalk
>     public TranslateSource[] DraftingSources { get; set; }
>     public TranslateSource[] TrainingSources { get; set; }
>     public SFProject[] TrainingTargets { get; set; }
> ```

You should initialize these properties to default values:
```cs
    public TranslateSource[] DraftingSources { get; set; } = [];
    public TranslateSource[] TrainingSources { get; set; } = [];
    public SFProject[] TrainingTargets { get; set; } = [];
```



_Code quote:_
```Smalltalk
    public TranslateSource[] DraftingSources { get; set; }
    public TranslateSource[] TrainingSources { get; set; }
    public SFProject[] TrainingTargets { get; set; }
```
___
*[`src/SIL.XForge.Scripture/Models/DraftSourcesArrays.cs` line 11 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3097#-OLVup6V8t3JFbfmP9he:-OLVup6V8t3JFbfmP9hf:b-mnwxpr) ([raw file](https://github.com/sillsdev/web-xforge/blob/4104dca616bb56a86c0bdb7611ab7ae50dd30d41/src/SIL.XForge.Scripture/Models/DraftSourcesArrays.cs#L11)):*
> ```Smalltalk
>     public SFProject[] TrainingTargets { get; set; }
> 
>     internal static DraftSourcesArrays ProjectToDraftSources(SFProject project)
> ```

I think things get murky when we include logic in data classes (even though this is static, so the logic does not affect the instance.

Have you considered creating an extension method for SFProject, by creating a class in Services that looks something like:

```cs
public static class SFProjectExtensions
{
    public static DraftSourcesArrays ToDraftSources(this SFProject project)
    {
    }
}
```

Then we can use:
```cs
project.ToDraftSources();
```



_Code quote:_
```Smalltalk
internal static DraftSourcesArrays ProjectToDraftSources(SFProject project)
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/serval-administration/serval-projects.component.ts` line 134 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3097#-OLVz5XiE0XhYUuS8JGd:-OLVz5XiE0XhYUuS8JGe:bg51kz9) ([raw file](https://github.com/sillsdev/web-xforge/blob/4104dca616bb56a86c0bdb7611ab7ae50dd30d41/src/SIL.XForge.Scripture/ClientApp/src/app/serval-administration/serval-projects.component.ts#L134)):*
> ```XML
>     const rows: Row[] = [];
>     for (const projectDoc of this.projectDocs) {
>       // FIXME: Do this in parallel
> ```

Are you planning to do this in this PR?



_Code quote:_
```XML
// FIXME: Do this in paralle
```

Nateowami (changes requested):
I have some serious concerns with this approach. I'm going to be gone most of the afternoon, so I don't have time to leave more comments now, but I plan to get back to this later.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3097#-:-OL_Rd5S7Y2BWwp10ypK:b-c15z75)* status: all files reviewed, 5 unresolved discussions (waiting on @RaymondLuong3)

RaymondLuong3:
Sounds good, I'll put a pause on this issue until we can have a chat about it.

RaymondLuong3:
@Nateowami How would you like me to proceed with this change?

Nateowami:
@RaymondLuong3 I'm closing this for two reasons:
1. This isn't actually the approach I was proposing.
2. I'm not sure the approach I was proposing is even the right way forward at this time. I think we have a document somewhere where we discussed possible approaches. We'll need to pick up that conversation again.

PR #3098, opened 2025-03-14
SF-3260 Show error when draft fails because source is not specified

RaymondLuong3:
When a project without a source starts a pretranslation build, the build never starts because there is a silent error. This change reports the error and shows the draft as failed.
I have marked this as testing not required since developer testing should be sufficient.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3098)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/MachineProjectService.cs
+++ b/src/SIL.XForge.Scripture/Services/MachineProjectService.cs
@@ -519,6 +519,7 @@ public class MachineProjectService(
     /// <param name="cancellationToken">The cancellation token.</param>
     /// <returns>An asynchronous task.</returns>
     /// <exception cref="DataNotFoundException">The project or project secret could not be found.</exception>
+    /// <exception cref="InvalidDataException">The language of the source project was not specified.</exception>
     /// <remarks>This can be mocked in unit tests.</remarks>
     protected internal virtual async Task BuildProjectAsync(
         string curUserId,
@@ -944,7 +945,7 @@ public class MachineProjectService(
                 // This error can occur if the project source is cleared while the build is running
                 if (projectDoc.Data.TranslateConfig.Source is null)
                 {
-                    throw new DataNotFoundException("The project source is not specified.");
+                    throw new InvalidDataException("The project source is not specified.");
                 }
 
                 // Update the source writing system tag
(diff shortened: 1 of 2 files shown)


pmachapman:
@RaymondLuong3 I have concerns about this fix, that it will break the original use case this code was designed to resolve.

This block was originally to trap errors that occur when a project is deleted while a build is running (this was happening on QA quite a lot). This change will push the exception to the `catch (Exception ex)` block, which will throw a hideous error as it attempts to write to the [at this stage deleted] project secret, and then cause the Hangfire job to run again and again until Hangfire gives up.

Based on the discussion in the Slack thread, and I right in understanding that the following block in `MachineProjectService.EnsureTranslationEngineExistsAsync()` is the cause of this PR?
```cs
// This error can occur if the project source is cleared while the build is running
if (projectDoc.Data.TranslateConfig.Source is null)
{
    throw new DataNotFoundException("The project source is not specified.");
}
```
If so, it might be better to throw a different exception there instead of `DataNotFoundException`, and document it in the method code doc? (it looks like this was an oversight in my original refactoring of this class when multiple sources were implemented)

RaymondLuong3 (commented):
That is a good idea. I wasn't aware of the history of why the DataNotFound exception was being used and logged as a warning. I've added new behaviour so `BuildProjectAsync` will now throw a `InvalidDataException` if the project's source cannot be found and that exception will be reported to bugsnag. It is definitely the safer option.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3098#-:-OL_MZmn63q4_kmu6iPL:bxzkvt3)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman (approved):
:lgtm: The PR worked as expected in dev testing.

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3098#-:-OL_SH1S2Q4TVmsBGUCR:b3c32re)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #3109, opened 2025-03-20
Add language code normalization logic to draft sources component

Nateowami:
This PR changes the language code comparison logic to allow different but equivalent language codes. For the purpose of the comparison, "equivalent" means "refer to the exact same language", *not* "refer to roughly the same language" or "map to the same NLLB language when Serval maps language codes".

This logic is used to prevent saving a draft config where the source and reference projects are not all in the same language.

There are several reasons for this decision:
1. **Easier implementation.** The logic Serval uses requires .NET libraries and would have to run server-side. This introduces network requests for basic UI logic.
2. **Possibly more correct.** The mapping Serval does is an attempt to find the closest language, for some definition of closest, in order to find the appropriate language in the model. What we're trying to determine is whether they're the same language. There might be two dialects that Serval maps to the same NLLB language, but we should still prevent those different dialects from both being used as a source.
3. **Different/simpler problem.** We don't actually have to know what Serval maps a language code to. If we send Serval projects with language codes `abc` an `abc`, we don't need to know that Serval maps it to `xyz` to know that Serval will map them both to the same language code.

### Bundle sizes
Adding data about language codes is of course going to increase the bundle size, and this trade-off must be considered. However, my testing (`npm run build:stats`) shows that the estimated over-the-wire transfer increases by only 451 bytes, which is pretty impressive on the part of the compression algorithms. However, my testing also shows that Brotli isn't currently in use, so our current actual over-the-wire transfer size is 913kB, about 26% more.

#### Before change
```
Initial chunk files           | Names         |  Raw size | Estimated transfer size
main.48cb3e66b2c2781b.js      | main          |   3.77 MB |               692.63 kB
styles.72ad50f8a98b3712.css   | styles        | 130.81 kB |                13.80 kB
polyfills.2a2c7847c69b8586.js | polyfills     |  60.35 kB |                18.49 kB
runtime.fd5dad9cd506830b.js   | runtime       |   1.40 kB |               739 bytes

                              | Initial total |   3.96 MB |               725.66 kB
```
#### After change
```
Initial chunk files           | Names         |  Raw size | Estimated transfer size
main.ba4c01ff9d8c804b.js      | main          |   3.77 MB |               693.07 kB
styles.72ad50f8a98b3712.css   | styles        | 130.81 kB |                13.80 kB
polyfills.2a2c7847c69b8586.js | polyfills     |  60.35 kB |                18.49 kB
runtime.fd5dad9cd506830b.js   | runtime       |   1.40 kB |               739 bytes

                              | Initial total |   3.96 MB |               726.10 kB
```

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3109)


Nateowami:
> Will there be an action setup to update the language_code_mapping.json?

Yes, I have been planning to add that as a followup. In the meantime, I doubt these are updated very often. It's not like the codes can change meaning; they can only add new codes.

> We may also want to consider allowing Serval Admins the ability to add mappings at some point if we start to see false positives that are preventing teams from drafting.

It may be necessary to revisit at some point. However, at that point it may be a linguistic and/or NLP question regarding which codes are compatible. This PR only considers codes that have the exact same meaning according to the standard.

PR #3112, opened 2025-03-20
Use nullish coalescing assignment when possible

Nateowami:
- Migrated `if ($expression == null) { $expression = $value; }` => `$expression ??= $value;`
- Updated all Grit lint rules to error when they fail (I recently created a PR that broke one of the rules, and couldn't see it without deliberately going to the checks tab of the pull request.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3112)

### DIFF
--- a/.grit/patterns/import_from_lodash_es.md
+++ b/.grit/patterns/import_from_lodash_es.md
@@ -1,5 +1,5 @@
 ---
-level: warn
+level: error
 ---
 
 Imports can safely be made directly from `lodash-es` instead of `lodash-es/...`.
--- a/.grit/patterns/missed_nullish_assignment_opportunity.md
+++ b/.grit/patterns/missed_nullish_assignment_opportunity.md
@@ -1,0 +1,13 @@
+---
+level: error
+---
+
+# Missed opportunity to use nullish assignment
+
+```grit
+language js
+
+`if ($expression == null) {
+    $expression = $value;
+}` => `$expression ??= $value;`
+```
--- a/.grit/patterns/xforge_common_absolute_only.md
+++ b/.grit/patterns/xforge_common_absolute_only.md
@@ -1,6 +1,6 @@
 ---
 tags: [precommit]
-level: warn
+level: error
 ---
 
 # Import from xforge-common, not from ../xforge-common
--- a/scripts/update_font_list.mts
+++ b/scripts/update_font_list.mts
@@ -36,9 +36,7 @@ for (const entry of Object.values(fonts)) {
 
   const defaultFileUrl = entry.files[defaultFileName]?.flourl;
 
-  if (filesByFamily[entry.family] == null) {
-    filesByFamily[entry.family] = [];
-  }
+  filesByFamily[entry.family] ??= [];
   filesByFamily[entry.family].push({
     fileName: defaultFileName,
     fileUrl: defaultFileUrl
--- a/src/RealtimeServer/common/metadata-db.ts
+++ b/src/RealtimeServer/common/metadata-db.ts
@@ -15,9 +15,7 @@ export function MetadataDB<T extends DBConstructor>(Base: T): T {
       options: any,
       callback: (...args: any[]) => any
     ): void {
-      if (options == null) {
-        options = {};
-      }
+      options ??= {};
       options.metadata = true;
       super.getOpsToSnapshot(collection, id, from, snapshot, options, callback);
     }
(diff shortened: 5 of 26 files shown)


Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3112)* status: 0 of 23 files reviewed, 1 unresolved discussion

___
*[a discussion](https://reviewable.io/reviews/sillsdev/web-xforge/3112#-OLpBsP-2rdibDRp4YRJ:-OLpBsP-2rdibDRp4YRK:b-g58dqj) (no related file):*
-51 lines of code, and remarkably, the bundle size drops by 387 bytes. Which is negligible, but a lot more than I expected.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts` line 35 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3112#-OLpApJF9wi8HpzF6O8b:-OLpApJF9wi8HpzF6O8c:b-jjgmeq) ([raw file](https://github.com/sillsdev/web-xforge/blob/74fc733f01efff04684df62b2243a43f4771da51/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/utils.ts#L35)):*
> ```XML
>     chromium: '>=87',
>     edge: '>=87',
>     firefox: '>=79',
> ```

I bumped the required Firefox version to 79, because the operator isn't available until that version: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment#browser_compatibility

However, Angular only supports the last two major versions of each browser anyway, so while we supposedly support these older versions, there's a good chance they're broken anyway. So I don't consider bumping the version to really be a trade-off.

RaymondLuong3 (changes requested):
> Reviewed 22 of 23 files at r1, 1 of 1 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3112)* status: all files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[a discussion](https://reviewable.io/reviews/sillsdev/web-xforge/3112#-OLpBsP-2rdibDRp4YRJ:-OLtzlpB8LGXZamSXGVw:b-ez5t0d) (no related file):*

-51 lines of code, and remarkably, the bundle size drops by 387 bytes. Which is negligible, but a lot more than I expected.
</blockquote></details>

Neat. I like the new operator and how it looks like it clean up code. Just one Minor request.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 660 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3112#-OLtypUm0uc-Q1u6rvbq:-OLtypUm0uc-Q1u6rvbr:b631weu) ([raw file](https://github.com/sillsdev/web-xforge/blob/3fab34017f624e66282dbe8922711bab19c3c005/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L660)):*
> ```XML
>       this.checkSessionPromise = undefined;
>     });
>     return this.checkSessionPromise;
> ```

Looks like this should be awaited. Otherwise there is no need for this to be an async function.



_Code quote:_
```XML
    return this.checkSessionPromise;
```

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3112)* status: 25 of 26 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts` line 660 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3112#-OLtypUm0uc-Q1u6rvbq:-OLuH47C3_bZG_tAEa8m:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/3fab34017f624e66282dbe8922711bab19c3c005/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.ts#L660)):*

Looks like this should be awaited. Otherwise there is no need for this to be an async function.
</blockquote></details>

Done.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3112)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #3114, opened 2025-03-21
SF-3279 Replace SMS login with basic options on login with share links

RaymondLuong3:
This PR updates the configuration on the Auth0 login page that comes up when using a share link to join a project. Uses will now be prompted to login with google, facebook, or email/password rather than being prompted to use SMS as the default option. The Auth0 changes are on the development tenant so you can checkout this branch and try it.
image

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/xforge-common/auth.service.spec.ts
@@ -381,24 +381,7 @@ describe('AuthService', () => {
     }
   }));
 
-  it('should login with passwordless enabled', fakeAsync(() => {
-    const env = new TestEnvironment();
-    const returnUrl = 'test-returnUrl';
-
-    env.service.logIn({ returnUrl });
-
-    verify(mockedWebAuth.loginWithRedirect(anything())).once();
-    const authOptions: RedirectLoginOptions | undefined = capture<RedirectLoginOptions | undefined>(
-      mockedWebAuth.loginWithRedirect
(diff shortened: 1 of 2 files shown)


josephmyers (changes requested):
So, the landing page, shown in your screenshot above, is really, really similar to the page you reach by clicking "log in with a different method." The only difference I can see is the latter has the Paratext login option. Is there any reason not to just show the Paratext login on the landing page? Then remove the "log in with a different method" button? Thanks for your work so far.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3114#-:-OM6ITN24JYLMQ9B9jx2:b-7m9p43)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3:
@Nateowami Any thoughts? We did intentionally hide the log in with Paratext option since these share links will typically go out to non-paratext users. I think we should keep it as is, but you might know more.

josephmyers:
I'm not strongly opinionated, but keep in mind the share link takes you to a page where you are prompted to simply enter your user name so that we create an account for you. The only way to get to the screen above is by clicking the option to log in instead with an existing account.

josephmyers (approved):
Thanks for bringing this up at the planning meeting. I am fine with keeping it as is, too 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3114#-:-OMADmb49kgqJuREPTPC:b-x2kh2d)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #3120, opened 2025-03-26
SF-3290 Fix error when user tries to open resource tab

RaymondLuong3:
Now that we show the source panel even if the user does not have permission, there is an interesting scenario where the user gets an error if they try to open a resource. This was because of some logic to access the source's paratext Id. This PR updates the logic to not try to get the project that the user does not have access to.


pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3120#-:-OMJ1owY6STzltU_hFpT:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

PR #3124, opened 2025-03-27
SF-3189 Fix project deleted dialog appearing more than necessary

RaymondLuong3:
This PR fixes an issue where the project deleted or user no longer has access dialog appears multiple times when a user is removed from a project. This is because both the project component and app component had logic to show the dialog when the project doc is updated. This PR removes the logic from the project component so that the app component can handle showing the dialog.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3124)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/project/project.component.ts
@@ -66,9 +66,6 @@ export class ProjectComponent extends DataLoadingComponent implements OnInit {
     navigateToProject$.pipe(quietTakeUntilDestroyed(this.destroyRef)).subscribe(async projectId => {
       if (userDoc.data?.sites[environment.siteId].projects?.includes(projectId)) {
         this.navigateToProject(projectId);
-      } else {
-        await this.dialogService.message('app.project_has_been_deleted');
-        this.router.navigateByUrl('/projects', { replaceUrl: true });
       }
     });
   }


Nateowami:
I've seen it show up like 100 times at the same time (the entire background turns black due to so many dialog backdrops. I'm guessing that's an unrelated bug?

kylebuss (commented):
This looks good!

I observed an interesting behavior when I tested this locally and not sure how we want to handle it?

User joins the project and is logged in.
Admin removes users from project.
User gets dialog and remains logged in (should we log user out if they were only on the one project and removed?)
User gets new link and pastes in browser and is automatically rejoined without the Login to set user name.

When I first went through and pasted the link in Firefox I expected the Login screen to appear but since I was still logged in anonymously it automatically joined me to the project. I guess the main question is if this should be the expected behavior or if we should log the user out if they are removed from the only project they are on?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3124#-:-OMRqK9E23tg349iKgVp:b-prii4a)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
> I've seen it show up like 100 times at the same time

Yes, that sounds like what is happening in SF-3204 and shouldn't be related to this issue.

> User gets new link and pastes in browser and is automatically rejoined without the Login to set user name.

That is the expected behaviour. When a user logs in anonymously, the have a valid SF account and it is also valid in auth0. Once they log out, there is no way for them to regain access to the account, but while they are logged in, everything should work that same as if they were logged with google or email.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3124#-:-OMSJDYtESRi8b6CcdV5:b-1m3exm)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

kylebuss (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3124#-:-OMSKR7I6R9Z0o0ZCXeV:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

Nateowami:
@RaymondLuong3 The change you've made here does not seem to correspond to the description in the issue:
> The user should not see the "project deleted" message for "test2" since the user is still part of that project

The description seems to be saying that a user is shown this message when actually still on the project. This change seems to be about how many times a dialog is shown, not about it being shown when it shouldn't be.

PR #3133, opened 2025-03-31
SF-3268 Fix clearing first training source

marksvc:
Clearing the first training source would incorrectly set it to the
source project.

The bug fix is in draft-sources.component.ts where we now pass
`definedReferences` and `definedSources` instead of
`this.trainingSources` and `this.draftingSources`.

draft-sources.component.html:
- I didn't see that the `loading ? undefined : ..` was working as
intended and removed it.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3133)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts
@@ -19,8 +19,8 @@ import { NoticeService } from 'xforge-common/notice.service';
 import { SFUserProjectsService } from 'xforge-common/user-projects.service';
 import { quietTakeUntilDestroyed } from 'xforge-common/util/rxjs-util';
 import { XForgeCommonModule } from 'xforge-common/xforge-common.module';
+import { hasData, isInstantiated } from '../../../../type-utils';
 import { SFProjectProfileDoc } from '../../../core/models/sf-project-profile-doc';
-import { SFProjectSettings } from '../../../core/models/sf-project-settings';
 import { ParatextService, SelectableProject, SelectableProjectWithLanguageCode } from '../../../core/paratext.service';
 import { SFProjectService } from '../../../core/sf-project.service';
 import { projectLabel } from '../../../shared/utils';
@@ -116,8 +116,8 @@ export class DraftSourcesComponent extends DataLoadingComponent {
         this.draftingSources = draftingSources.map(translateSourceToSelectableProjectWithLanguageTag);
 
         this.nonSelectableProjects = [
-          ...this.trainingSources.filter(s => s != null),
-          ...this.draftingSources.filter(s => s != null)
+          ...this.trainingSources.filter(isInstantiated),
+          ...this.draftingSources.filter(isInstantiated)
         ];
 
         if (this.draftingSources.length < 1) this.draftingSources.push(undefined);
@@ -140,13 +140,13 @@ export class DraftSourcesComponent extends DataLoadingComponent {
   }
 
   get referenceLanguageDisplayName(): string {
-    const uniqueTags = Array.from(new Set(this.trainingSources.filter(s => s != null).map(p => p.languageTag)));
+    const uniqueTags = Array.from(new Set(this.trainingSources.filter(isInstantiated).map(p => p.languageTag)));
     const displayNames = uniqueTags.map(tag => this.i18n.getLanguageDisplayName(tag) ?? tag);
     return this.i18n.enumerateList(displayNames);
   }
 
   get sourceLanguageDisplayName(): string | undefined {
-    const definedSources = this.draftingSources.filter(s => s != null);
+    const definedSources = this.draftingSources.filter(isInstantiated);
 
     if (definedSources.length > 1) throw new Error('Multiple drafting sources not supported');
     else if (definedSources.length < 1) return undefined;
@@ -164,15 +164,15 @@ export class DraftSourcesComponent extends DataLoadingComponent {
   }
 
   get sourceSubtitle(): string {
-    return this.i18n.enumerateList(this.draftingSources.filter(s => s != null).map(s => s.shortName) ?? []);
+    return this.i18n.enumerateList(this.draftingSources.filter(isInstantiated).map(s => s.shortName) ?? []);
   }
 
   get referencesSubtitle(): string {
-    return this.i18n.enumerateList(this.trainingSources.filter(s => s != null).map(r => r.shortName) ?? []);
+    return this.i18n.enumerateList(this.trainingSources.filter(isInstantiated).map(r => r.shortName) ?? []);
   }
 
   get targetSubtitle(): string {
-    return this.i18n.enumerateList(this.trainingTargets.filter(s => s != null).map(t => t.shortName) ?? []);
+    return this.i18n.enumerateList(this.trainingTargets.filter(isInstantiated).map(t => t.shortName) ?? []);
   }
 
   parentheses(value?: string): string {
@@ -194,10 +194,10 @@ export class DraftSourcesComponent extends DataLoadingComponent {
 
   get draftSourcesAsArray(): DraftSourcesAsSelectableProjectArrays {
     return {
-      draftingSources: this.draftingSources.filter(s => s != null),
-      trainingSources: this.trainingSources.filter(s => s != null),
+      draftingSources: this.draftingSources.filter(isInstantiated),
+      trainingSources: this.trainingSources.filter(isInstantiated),
       trainingTargets: this.trainingTargets
-        .filter(s => s != null)
+        .filter(isInstantiated)
         .map(t => translateSourceToSelectableProjectWithLanguageTag(t))
     };
   }
@@ -216,9 +216,6 @@ export class DraftSourcesComponent extends DataLoadingComponent {
     index: number,
     paratextId: string | undefined
   ): void {
-    // When still loading projects, the project selectors will temporarily set the value to null
-    if (!this.isLoaded) return;
-
     const selectedProject: SelectableProject | null =
       this.projects?.find(p => p.paratextId === paratextId) ??
       this.resources?.find(r => r.paratextId === paratextId) ??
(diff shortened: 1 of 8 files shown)


marksvc:
FWIW, I'm having trouble interpreting CodeCov to understand where the introduced && untested code lines are.

pmachapman (changes requested):
Are you able to rebase then remove the changes to packages.lock.json from this PR? I need to look into why it was recreated and modified to remove ICU4C on your workstation.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3133#-:-OMhzH9B3AMF-qAvNeTd:bn2phpw)* status: all files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts` line 52 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3133#-OMi-s5rDRIPsYBB8Wl8:-OMi-s5rDRIPsYBB8Wl9:bgwzutd) ([raw file](https://github.com/sillsdev/web-xforge/blob/6d8c50046414e3b9b11418cb2f9c8f1e662f816e/src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts#L52)):*
I find the naming to be confusing, as there are no real parallels our code or JS/TS to the naming `isInstantiated`. Could we name it `isNotNull` or similar?

I also think that `value != null` is clearer than `value !== null && value !== undefined`.

According to the docs (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Data_structures#undefined_type):
> Conceptually, `undefined` indicates the absence of a _value_, while `null` indicates the absence of an _object_ (which could also make up an excuse for [`typeof null === "object"`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#typeof_null)).

Technically only `!== null` would mean not instantiated, as a value cannot be instantiated, only objects can (my understanding is instantiate is an OOP term).



_Code quote:_
```XML
isInstantiated
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.spec.ts` line 31 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3133#-OMi2CBl1VGgyPR55bRa:-OMi2CBl1VGgyPR55bRb:b-eb7dl) ([raw file](https://github.com/sillsdev/web-xforge/blob/6d8c50046414e3b9b11418cb2f9c8f1e662f816e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.spec.ts#L31)):*
> ```XML
> import { DraftSourcesComponent, sourceArraysToSettingsChange } from './draft-sources.component';
> 
> interface UltimateProjectDescription {
> ```

I think the naming of this interface is unclear. Maybe `TestProjectConfig`, `TestDraftProject`, or something similar, would be better?



_Code quote:_
```XML
UltimateProjectDescription
```

marksvc (commented):
packages.lock.json - Done; thank you!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3133#-:-OMsdZkS9D1EkcmGNH-P:bup9n5v)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts` line 52 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3133#-OMi-s5rDRIPsYBB8Wl8:-OMsee6CAWSpR_-vG2Rb:bxmakj7) ([raw file](https://github.com/sillsdev/web-xforge/blob/6d8c50046414e3b9b11418cb2f9c8f1e662f816e/src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts#L52)):*

I find the naming to be confusing, as there are no real parallels our code or JS/TS to the naming `isInstantiated`. Could we name it `isNotNull` or similar?

I also think that `value != null` is clearer than `value !== null && value !== undefined`.

According to the docs (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Data_structures#undefined_type):
> Conceptually, `undefined` indicates the absence of a _value_, while `null` indicates the absence of an _object_ (which could also make up an excuse for [`typeof null === "object"`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof#typeof_null)).

Technically only `!== null` would mean not instantiated, as a value cannot be instantiated, only objects can (my understanding is instantiate is an OOP term).
</blockquote></details>

Thank you. How about `notNull()`?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.spec.ts` line 31 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3133#-OMi2CBl1VGgyPR55bRa:-OMsfBgBAnajDzi1Aic-:b-k02xlu) ([raw file](https://github.com/sillsdev/web-xforge/blob/6d8c50046414e3b9b11418cb2f9c8f1e662f816e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.spec.ts#L31)):*

I think the naming of this interface is unclear. Maybe `TestProjectConfig`, `TestDraftProject`, or something similar, would be better?
</blockquote></details>

The draft-sources.component works with projects using a variety of types. So if I make an array of `SFProject` objects, I might need to transform them to `ParatextProject[]` to be sent via paratextService.getProjects(), or to `TranslateSource` to be set into `sfProject0.translateConfig.draftConfig.alternateSource`. In an attempt to address this in the spec in a slightly saner way than some alternatives, I made a type that lets me define a project using _all four types_ at once. Then later if I want that project as a `SelectableProjectWithLanguageCode` or as a `ParatextProject`, I can get it as that type. 

For this comprehensive project description :), I gave 'UltimateProjectDescription`, which perhaps makes more sense with the backstory :). I wonder if `MultiTypeProjectDescription` might communicate what the interface is providing? Ah ha, how about this code comment.

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3133#-:-OMt3Sh14IhYs0G759FV:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts` line 52 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3133#-OMi-s5rDRIPsYBB8Wl8:-OMt2wiz77sq8nWm7abw:brxp9sq) ([raw file](https://github.com/sillsdev/web-xforge/blob/6d8c50046414e3b9b11418cb2f9c8f1e662f816e/src/SIL.XForge.Scripture/ClientApp/src/type-utils.ts#L52)):*

Thank you. How about `notNull()`?
</blockquote></details>

Sounds good.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.spec.ts` line 31 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3133#-OMi2CBl1VGgyPR55bRa:-OMt3DkHBRAT0MxKz5Rw:bdqzdm7) ([raw file](https://github.com/sillsdev/web-xforge/blob/6d8c50046414e3b9b11418cb2f9c8f1e662f816e/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.spec.ts#L31)):*

The draft-sources.component works with projects using a variety of types. So if I make an array of `SFProject` objects, I might need to transform them to `ParatextProject[]` to be sent via paratextService.getProjects(), or to `TranslateSource` to be set into `sfProject0.translateConfig.draftConfig.alternateSource`. In an attempt to address this in the spec in a slightly saner way than some alternatives, I made a type that lets me define a project using _all four types_ at once. Then later if I want that project as a `SelectableProjectWithLanguageCode` or as a `ParatextProject`, I can get it as that type. 

For this comprehensive project description :), I gave 'UltimateProjectDescription`, which perhaps makes more sense with the backstory :). I wonder if `MultiTypeProjectDescription` might communicate what the interface is providing? Ah ha, how about this code comment.
</blockquote></details>

Yes, The backstory did help! I think they rename and comment is fine.

src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts
```diff
@@ -216,9 +213,6 @@ export class DraftSourcesComponent extends DataLoadingComponent {
     index: number,
     paratextId: string | undefined
   ): void {
-    // When still loading projects, the project selectors will temporarily set the value to null
-    if (!this.isLoaded) return;
```
Nateowami, line 220: @marksvc Do you remember why you deleted these two lines? This PR broke the logic for determining whether to warn the user about unsaved changes when leaving the page, because several lines farther down `this.changesMade = true;` is run when it shouldn't.

marksvc: Yeah. My interpretation of observations I saw when debugging the code running the program and running in unit tests was that our loading, processing, and subscriptions had a race situation that always happened to be won by the right party when running the program, but not by the right party when run in tests. So I changed some of the processing to be more resilient to the order that things are set. I think I could have given a better description back when I was in the midst of it. 

One observation I had was that draft-sources.component.html was designed with particular behaviour regarding time during loading, with `[value]="loading ? undefined : source?.paratextId"`. In one step during working on SF-3268 I found that setting value=undefined seemed to be part of the trouble. Later on I'm not sure how functionally important it was to make a change there, except that _when running the program, this code didn't seem to be at all working as designed_, and so I took it out to add clarity.

The lines in question were removed because (1) they were observed as party to a race situation, and (2) they were seemed related to trouble getting down to a line in the method of `array[index] = undefined;` which was causing problems. 

So that's why they were removed. But with that context, I think it would be fine to put back in whatever checks etc as long as the tests and storybook and program still run correctly. 

Let me know if you'd like me to work on something here.

PR #3139, opened 2025-04-01
SF-3303 Allow selection of the echo translation engine

pmachapman:
This PR changes the echo translation engine from being a feature flag configured on the server via user secrets, to a settings that can be set per build, like fast training.

**Notes**
If the "Allow Echo for Pre-Translation Drafting" feature flag is disabled, edit your user secrets and ensure that this secret is not present:
```
"FeatureManagement:UseEchoForPreTranslation": "false",
```

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.html
@@ -202,17 +202,27 @@
           </div>
         </mat-step>
       }
-      @if (featureFlags.allowFastTraining.enabled) {
+      @if (featureFlags.allowFastTraining.enabled || featureFlags.useEchoForPreTranslation.enabled) {
         <mat-step [completed]="true">
           <ng-template matStepLabel>
             {{ t("configure_advanced_settings_header") }}
           </ng-template>
           <h1 class="mat-headline-4">{{ t("configure_advanced_settings_title") }}</h1>
-          <div>
-            <mat-checkbox class="fast-training" [(ngModel)]="fastTraining">{{ t("fast_training") }}</mat-checkbox>
-          </div>
-          @if (fastTraining) {
-            <app-notice icon="warning" type="warning" class="warning">{{ t("fast_training_warning") }}</app-notice>
+          @if (featureFlags.allowFastTraining.enabled) {
+            <div>
+              <mat-checkbox class="fast-training" [(ngModel)]="fastTraining">{{ t("fast_training") }}</mat-checkbox>
+            </div>
+            @if (fastTraining) {
+              <app-notice icon="warning" type="warning" class="warning">{{ t("fast_training_warning") }}</app-notice>
+            }
+          }
+          @if (featureFlags.useEchoForPreTranslation.enabled) {
+            <div>
+              <mat-checkbox class="use-echo" [(ngModel)]="useEcho">{{ t("use_echo") }}</mat-checkbox>
+            </div>
+            @if (useEcho) {
+              <app-notice icon="warning" type="warning" class="warning">{{ t("use_echo_warning") }}</app-notice>
+            }
           }
           <div class="button-strip">
             <button mat-stroked-button matStepperPrevious>
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-generation-steps/draft-generation-steps.component.spec.ts
@@ -137,6 +137,7 @@ describe('DraftGenerationStepsComponent', () => {
       );
       when(mockTrainingDataQuery.docs).thenReturn([]);
       when(mockFeatureFlagService.allowFastTraining).thenReturn(createTestFeatureFlag(false));
+      when(mockFeatureFlagService.useEchoForPreTranslation).thenReturn(createTestFeatureFlag(false));
 
       fixture = TestBed.createComponent(DraftGenerationStepsComponent);
       component = fixture.componentInstance;
@@ -202,6 +203,7 @@ describe('DraftGenerationStepsComponent', () => {
       );
       when(mockTrainingDataQuery.docs).thenReturn([]);
       when(mockFeatureFlagService.allowFastTraining).thenReturn(createTestFeatureFlag(false));
+      when(mockFeatureFlagService.useEchoForPreTranslation).thenReturn(createTestFeatureFlag(false));
 
       fixture = TestBed.createComponent(DraftGenerationStepsComponent);
       component = fixture.componentInstance;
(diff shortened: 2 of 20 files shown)


Nateowami:
In my testing, it sometimes works, and sometimes says it's in progress but never completes.

pmachapman:
> In my testing, it sometimes works, and sometimes says it's in progress but never completes.

If you refreshed the page, did it sort itself? I wonder if the echo engine is too fast, that our polling misses the window of opportunity of the build finishing?

Nateowami:
> If you refreshed the page, did it sort itself?

I'm almost certain I tried that.

> I wonder if the echo engine is too fast, that our polling misses the window of opportunity of the build finishing?

If there's a limited window of opportunity to detect that a build finished, I would consider that a bug we should fix. I wouldn't have really thought that was possible though, since IIRC the front end calls the back end asking for the information on the most recent draft, which is then either one in progress or complete.

marksvc (approved):
In my own testing so far, it has worked. I have just done it a couple times.

> Reviewed 9 of 9 files at r1, 11 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3139#-:-ON1Wyed8hbxhmbf7hBS:beop89d)* status: all files reviewed, 2 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 696 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3139#-OMxeVWF4HhhcKfy_6V8:-OMxeVWF4HhhcKfy_6V9:blp0tqx) ([raw file](https://github.com/sillsdev/web-xforge/blob/40449e09984c14e0053775d1844b8d6b80578c4c/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L696)):*
> ```Smalltalk
>     /// <returns>The translation engine id.</returns>
>     /// <exception cref="DataNotFoundException">The translation engine could not be created.</exception>
>     /// <remarks>This can be mocked in unit tests.</remarks>
> ```

What is the purpose of this remark? Is this meaning to communicate that when testing `MachineProjectService`, in the tests for `MachineProjectService` this method itself can be mocked? 

I wouldn't think it means that when testing _other classes_ that use a `MachineProjectService`, this method can be mocked, because I would expect that to normally be so.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 1258 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3139#-OMxgVd23JcsmP_Ku1Fp:-OMxgVd23JcsmP_Ku1Fq:b48f5kn) ([raw file](https://github.com/sillsdev/web-xforge/blob/40449e09984c14e0053775d1844b8d6b80578c4c/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L1258)):*
> ```Smalltalk
>             }
> 
>             // See if the source language has changed
> ```

I am glad for the comments sprinkled through the code, saying what is intended.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 1868 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3139#-OMxjr2WDD3aNU0id4Rh:-OMxjr2WDD3aNU0id4Ri:b-hhyo4t) ([raw file](https://github.com/sillsdev/web-xforge/blob/40449e09984c14e0053775d1844b8d6b80578c4c/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L1868)):*
> ```Smalltalk
>     /// <returns><c>true</c> if the translation engine exists; otherwise <c>false</c>.</returns>
>     /// <remarks>This can be mocked in unit tests.</remarks>
>     protected internal virtual async Task<bool> TranslationEngineExistsAsync(
> ```

This seem strange. I think what this is essentially doing is that it is checking if a translation engine exists
1. For a given SF project.
2. With a given translation engine id.
3. That does or does not use NMT, as specified by the `preTranslate` argument.
4. And tolerates various formatting of the engine type.

To confirm my own understanding, would it be accurate to say that the lines
```cs
bool useEcho = ...
string type = ...
return translationEngineName == ...
```
would work the same if you instead used the following lines?
```cs
bool useEcho = translationEngine.Type == Echo;
if (translationEngine.Name == projectId && useEcho) return true;
string type = GetTranslationEngineType(preTranslae, useEcho);
return translationEngine.Name == ...
    && string.Equals(...
```

I'm not saying you should change to that; just asking to confirm my understanding of the code here.

marksvc:
Okay, actually, I'm not sure now that it did really work :). Or at least, not as I expected it to. The text is quite similar to the configured source. But it is different. For example, a word might be added, or missing. So it's not a perfect echoing of the configured source. But it did.. quickly give me something. And maybe this is what the echo is expected to look like.

pmachapman (commented):
> If there's a limited window of opportunity to detect that a build finished, I would consider that a bug we should fix. I wouldn't have really thought that was possible though, since IIRC the front end calls the back end asking for the information on the most recent draft, which is then either one in progress or complete.

@Nateowami I can't recreate the issue you were seeing - if it happens again, please let me know the translation engine id, and the target + source, so I can look into it.

> Okay, actually, I'm not sure now that it did really work :). Or at least, not as I expected it to. The text is quite similar to the configured source. But it is different. For example, a word might be added, or missing. So it's not a perfect echoing of the configured source. But it did.. quickly give me something. And maybe this is what the echo is expected to look like.

@marksvc What source and target did you use (please add me to both if you are able), and I will see if I can recreate this and investigate further.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3139#-:-ONGf22n6elH85FoXJSr:bd6l48e)* status: 19 of 20 files reviewed, 2 unresolved discussions (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 696 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3139#-OMxeVWF4HhhcKfy_6V8:-ONGfe6J-5C0iSk99qPj:b-ldd0xb) ([raw file](https://github.com/sillsdev/web-xforge/blob/40449e09984c14e0053775d1844b8d6b80578c4c/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L696)):*
> What is the purpose of this remark?

It is a comment to try to explain the `protected internal virtual async`, and why it is there.

> Is this meaning to communicate that when testing `MachineProjectService`, in the tests for `MachineProjectService` this method itself can be mocked?

Yes. If you can think of better wording, please let me know.
___
*[`src/SIL.XForge.Scripture/Services/MachineProjectService.cs` line 1868 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3139#-OMxjr2WDD3aNU0id4Rh:-ONGhNZCAAuWQ1AxEhwR:bhrzjfy) ([raw file](https://github.com/sillsdev/web-xforge/blob/40449e09984c14e0053775d1844b8d6b80578c4c/src/SIL.XForge.Scripture/Services/MachineProjectService.cs#L1868)):*

This seem strange. I think what this is essentially doing is that it is checking if a translation engine exists
1. For a given SF project.
2. With a given translation engine id.
3. That does or does not use NMT, as specified by the `preTranslate` argument.
4. And tolerates various formatting of the engine type.

To confirm my own understanding, would it be accurate to say that the lines
```cs
bool useEcho = ...
string type = ...
return translationEngineName == ...
```
would work the same if you instead used the following lines?
```cs
bool useEcho = translationEngine.Type == Echo;
if (translationEngine.Name == projectId && useEcho) return true;
string type = GetTranslationEngineType(preTranslae, useEcho);
return translationEngine.Name == ...
    && string.Equals(...
```

I'm not saying you should change to that; just asking to confirm my understanding of the code here.
</blockquote></details>

Not quite. `if (translationEngine.Name \== projectId && useEcho) return true;` is incorrect. Echo can only be used for pre-translation. `GetTranslationEngineType` could be expressed as the following truth table:
```
+--------------+---------+--------------+
| preTranslate | useEcho |   returns    |
+--------------+---------+--------------+
| true         | true    | echo         |
| true         | false   | nmt          |
| false        | false   | smt-transfer |
| false        | true    | smt-transfer |
+--------------+---------+--------------+

```

Reference:
```cs
    protected internal virtual string GetTranslationEngineType(bool preTranslate, bool useEcho) =>
        preTranslate switch
        {
            true when useEcho => Echo,
            true => Nmt,
            false => SmtTransfer,
        };
```

PR #3141, opened 2025-04-02
WIP: SF-3263 Improve build event logging and display builds coherently

pmachapman:
This PR is a work in progress.

In Progress:

 * Backend
 
To Do:

 * Frontend
 * Feature flag for frontend?


RaymondLuong3 (changes requested):
I was able to use the swagger endpoint and get the build config for the most recent draft for a project and theoretically it also shows previously started builds also.

> Reviewed 10 of 10 files at r1, 12 of 12 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3141#-:-ON1by869CvXb_rqfGnp:b1nmt3g)* status: all files reviewed, 4 unresolved discussions (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 1555 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3141#-ON17yz5EgOvaqUkOntf:-ON17yz5EgOvaqUkOntg:b-ydgmkp) ([raw file](https://github.com/sillsdev/web-xforge/blob/97b57c7c5a5d6a0550e1c512c2e92576e5981a9d/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L1555)):*
> ```Smalltalk
>                 new ProjectScriptureRange { ScriptureRange = buildConfig.TrainingScriptureRange }
>             );
>         }
> ```

I wonder if we should make the `TrainingScriptureRange` and `TranslationScriptureRange` properties obsolete in the model. It is confusing to have both range and ranges. That doesn't need to hold up this PR, but if we did a migration sooner it would make updates like this easier in the future.



_Code quote:_
```Smalltalk
        if (!string.IsNullOrWhiteSpace(buildConfig.TrainingScriptureRange))
        {
            buildDto.AdditionalInfo.TrainingScriptureRanges.Add(
                new ProjectScriptureRange { ScriptureRange = buildConfig.TrainingScriptureRange }
            );
        }
```
___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 1569 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3141#-ON1bSpm1qS2V2r_b3fy:-ON1bSpm1qS2V2r_b3fz:b-r6ijtx) ([raw file](https://github.com/sillsdev/web-xforge/blob/97b57c7c5a5d6a0550e1c512c2e92576e5981a9d/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L1569)):*
> ```Smalltalk
>                 new ProjectScriptureRange
>                 {
>                     ProjectId = buildConfig.ProjectId,
> ```

I don't think the build config projectId is correct. It should be the drafting source project Id. I'm not sure where to get that. I think we should put the correct projectId from the drafting source if we can.



_Code quote:_
```Smalltalk
                    ProjectId = buildConfig.ProjectId,
```
___
*[`src/SIL.XForge/Models/QueryResults.cs` line 18 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3141#-ON12jmi2co-UPs59Qlk:-ON12jmj9wIlSCybqEZ7:bj3sowm) ([raw file](https://github.com/sillsdev/web-xforge/blob/97b57c7c5a5d6a0550e1c512c2e92576e5981a9d/src/SIL.XForge/Models/QueryResults.cs#L18)):*
> ```Smalltalk
>     /// An empty query results set.
>     /// </summary>
>     public static QueryResults<T> Empty { get; } = new QueryResults<T> { Results = new List<T>(), UnpagedCount = 0L };
> ```

What does this `OL` mean for the unpagedCount?



_Code quote:_
```Smalltalk
    public static QueryResults<T> Empty { get; } = new QueryResults<T> { Results = new List<T>(), UnpagedCount = 0L };
```
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineApiServiceTests.cs` line 1051 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3141#-ON1CsUz3UVF-kgPa_Nr:-ON1CsUz3UVF-kgPa_Ns:b-nsedsq) ([raw file](https://github.com/sillsdev/web-xforge/blob/97b57c7c5a5d6a0550e1c512c2e92576e5981a9d/test/SIL.XForge.Scripture.Tests/Services/MachineApiServiceTests.cs#L1051)):*
> ```Smalltalk
>         };
>         env.TranslationEnginesClient.GetAllBuildsAsync(TranslationEngine01, CancellationToken.None)
>             .Returns(Task.FromResult<IList<TranslationBuild>>([translationBuild]));
> ```

Multiple tests use this line to configure the TranslationEnginesClient to treturn the translationBuild. I think if we could create a helper method to configure the TranslationEnginesClient with a translationBuild and also configure the asserts. That way is would be much easier to identify what is different between the tests. Maybe the function can be called ConfigureEnginesClientAndGetBuild(). 



_Code quote:_
```Smalltalk
        env.TranslationEnginesClient.GetAllBuildsAsync(TranslationEngine01, CancellationToken.None)
            .Returns(Task.FromResult<IList<TranslationBuild>>([translationBuild]));
```

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3141)* status: 20 of 22 files reviewed, 4 unresolved discussions (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge/Models/QueryResults.cs` line 18 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3141#-ON12jmi2co-UPs59Qlk:-ONGjpsx0FML9_hh9acM:b63x09f) ([raw file](https://github.com/sillsdev/web-xforge/blob/97b57c7c5a5d6a0550e1c512c2e92576e5981a9d/src/SIL.XForge/Models/QueryResults.cs#L18)):*

What does this `OL` mean for the unpagedCount?
</blockquote></details>

0 as a 64-bit long. I could have just written 0 (i.e. 0 as a 32-bit integer), but as the property is a long, I try to match the type of a magic number to the type of the underlying property/variable.

It does mean that if `UnpagedCount` changes to a 32-bit integer in future, this line will cause a compiler error (which can be fixed by removing the `L`).

Reference:
```cs
  public required long UnpagedCount { get; init; }
```
___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 1555 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3141#-ON17yz5EgOvaqUkOntf:-ONGkq8G0DI_7n2Q9_08:bgef8o0) ([raw file](https://github.com/sillsdev/web-xforge/blob/97b57c7c5a5d6a0550e1c512c2e92576e5981a9d/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L1555)):*

I wonder if we should make the `TrainingScriptureRange` and `TranslationScriptureRange` properties obsolete in the model. It is confusing to have both range and ranges. That doesn't need to hold up this PR, but if we did a migration sooner it would make updates like this easier in the future.
</blockquote></details>

I considered doing that in this PR, but noticed that the frontend uses the `TranslationScriptureRange` string, not the `TranslationScriptureRanges` array.

I think when we change the frontend to use `TranslationScriptureRanges`, then we can mark both `TrainingScriptureRange` and `TranslationScriptureRange` obsolete.
___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 1569 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3141#-ON1bSpm1qS2V2r_b3fy:-ONGltfuCPgTGt5g-j3O:b3zm14x) ([raw file](https://github.com/sillsdev/web-xforge/blob/97b57c7c5a5d6a0550e1c512c2e92576e5981a9d/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L1569)):*

I don't think the build config projectId is correct. It should be the drafting source project Id. I'm not sure where to get that. I think we should put the correct projectId from the drafting source if we can.
</blockquote></details>

Using the project id for the `TranslationScriptureRange` is correct, as this is the target, not the source, and the target for builds will always be the current project (at least until we support multiple targets).

I didn't set the source project if for the `TrainingScriptureRange`, as there is no reliable way to get this, as the project source config may have changed since the build was run (and these will be older builds). In this case I figure it is best not to set it so the frontend can display "Unknown" or similar.
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineApiServiceTests.cs` line 1051 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3141#-ON1CsUz3UVF-kgPa_Nr:-ONGzVl_3gWfrvQvl9Io:br7s375) ([raw file](https://github.com/sillsdev/web-xforge/blob/97b57c7c5a5d6a0550e1c512c2e92576e5981a9d/test/SIL.XForge.Scripture.Tests/Services/MachineApiServiceTests.cs#L1051)):*

Multiple tests use this line to configure the TranslationEnginesClient to treturn the translationBuild. I think if we could create a helper method to configure the TranslationEnginesClient with a translationBuild and also configure the asserts. That way is would be much easier to identify what is different between the tests. Maybe the function can be called ConfigureEnginesClientAndGetBuild().
</blockquote></details>

I have added two new methods to the test environment, and (hopefully) cleaned up the build code. Please let me know if this is what you were thinking, or if there is further work I can do on this.

RaymondLuong3 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3141)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 1555 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3141#-ON17yz5EgOvaqUkOntf:-ONHT0O-1TkddeR4X-gf:b-q0k36q) ([raw file](https://github.com/sillsdev/web-xforge/blob/97b57c7c5a5d6a0550e1c512c2e92576e5981a9d/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L1555)):*

I considered doing that in this PR, but noticed that the frontend uses the `TranslationScriptureRange` string, not the `TranslationScriptureRanges` array.

I think when we change the frontend to use `TranslationScriptureRanges`, then we can mark both `TrainingScriptureRange` and `TranslationScriptureRange` obsolete.
</blockquote></details>

Sounds good. I remember I was the one who decided to use the `TranslationScriptureRange` rather than ranges, so I assume the blame for the messyness that has resulted.
___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 1569 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3141#-ON1bSpm1qS2V2r_b3fy:-ONHUISK8gGGIo1MoCuE:bnhojmv) ([raw file](https://github.com/sillsdev/web-xforge/blob/97b57c7c5a5d6a0550e1c512c2e92576e5981a9d/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L1569)):*

Using the project id for the `TranslationScriptureRange` is correct, as this is the target, not the source, and the target for builds will always be the current project (at least until we support multiple targets).

I didn't set the source project if for the `TrainingScriptureRange`, as there is no reliable way to get this, as the project source config may have changed since the build was run (and these will be older builds). In this case I figure it is best not to set it so the frontend can display "Unknown" or similar.
</blockquote></details>

Ok, so what you are saying is that `TranslationScriptureRange` refers to the books that are being drafted for the target project. But my understanding of `TranslationScriptureRange` refer to the books in the source that are being translated from. The reasoning is that one day we may have multiple drafting sources and then we may draft books from 2 different sources.
I don't see an issue at the moment for this implementation, but I would think it is better to assign the projectId of the source for this, and not the projectId of the target.
Maybe I'm not quite understanding something.
___
*[`test/SIL.XForge.Scripture.Tests/Services/MachineApiServiceTests.cs` line 1051 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3141#-ON1CsUz3UVF-kgPa_Nr:-ONHTej09_vGBLBTNNoG:bg3h6vf) ([raw file](https://github.com/sillsdev/web-xforge/blob/97b57c7c5a5d6a0550e1c512c2e92576e5981a9d/test/SIL.XForge.Scripture.Tests/Services/MachineApiServiceTests.cs#L1051)):*

I have added two new methods to the test environment, and (hopefully) cleaned up the build code. Please let me know if this is what you were thinking, or if there is further work I can do on this.
</blockquote></details>

Yes, this is a lot easier to read and see differences between tests. Thanks!

pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3141)* status: 20 of 23 files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/Services/MachineApiService.cs` line 1569 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3141#-ON1bSpm1qS2V2r_b3fy:-ONHlCjH02dgAu-uo_hb:bo57i92) ([raw file](https://github.com/sillsdev/web-xforge/blob/97b57c7c5a5d6a0550e1c512c2e92576e5981a9d/src/SIL.XForge.Scripture/Services/MachineApiService.cs#L1569)):*

Ok, so what you are saying is that `TranslationScriptureRange` refers to the books that are being drafted for the target project. But my understanding of `TranslationScriptureRange` refer to the books in the source that are being translated from. The reasoning is that one day we may have multiple drafting sources and then we may draft books from 2 different sources.
I don't see an issue at the moment for this implementation, but I would think it is better to assign the projectId of the source for this, and not the projectId of the target.
Maybe I'm not quite understanding something.
</blockquote></details>

You are right - I was confused. Thank you for bouncing back on this.

Like I did for the `TrainingScriptureRange`, I will not include the `ProjectId`, as we cannot accurately tell what this is.

RaymondLuong3 (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3141#-:-ONKrk609kPgM8X-TgVn:bg7igkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #3147, opened 2025-04-08
SF-3198 Fix memory leak in RealtimeService from CheckIfRunning()

pmachapman:
This PR fixes a memory leak when the Realtime Service checks if the Realtime Server is running.

Also, following the guidance on https://docs.hangfire.io/en/latest/background-methods/using-ioc-containers.html, I have updated the background job logic to execute jobs by their interface, and have included Hangfire.Autofac so that Hangfire uses a tagged lifetime scope, meaning the objects are correctly disposed within the object lifecycle.

This PR is marked testing not required, as a debugger with breakpoint is required to run the acceptance test.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3147)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/IMachineApiService.cs
+++ b/src/SIL.XForge.Scripture/Services/IMachineApiService.cs
@@ -117,6 +117,8 @@ public interface IMachineApiService
         CancellationToken cancellationToken
     );
     Task<LanguageDto> IsLanguageSupportedAsync(string languageCode, CancellationToken cancellationToken);
+
+    [Mutex]
     Task RetrievePreTranslationStatusAsync(string sfProjectId, CancellationToken cancellationToken);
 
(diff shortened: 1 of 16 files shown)


josephmyers (approved):
:lgtm: I verified that, on master, the breakpoint is hit regularly, and here it's not. Great.

From a code perspective, I'm wondering about what was actually going on before, versus with this change. If I'm understanding, calling `CheckIfRunning` on the retrieved/supplied instance from the job manager was required? Or was the job manager creating the new instance every time, because we weren't specific enough in calling `AddOrUpdate` with the generic?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3147#-:-ONxafJK5u2uaZKdo9ON:b-gedmpi)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #3150, opened 2025-04-08
Change Path.Combine to Path.Join to fix CodeQL warnings

pmachapman:
This PR updates the use of `Path.Combine()` to `Path.Join`, which fixes these CodeQL warnings: https://github.com/sillsdev/web-xforge/security/code-scanning?query=is%3Aopen+branch%3Amaster+System.IO.Path.Combine

Using `Path.Combine` can be problematic if path parts after the first part start with a "/", as the path will then start from that part of the path.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3150)

### DIFF
--- a/src/Help/UpdateHelp/Program.cs
+++ b/src/Help/UpdateHelp/Program.cs
@@ -50,15 +50,15 @@ class Program
         bool doWrite = ((args.Length >= 2 ? args[1] : "") == "write");
 
         string userDir = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
-        string helpDir = Path.Combine(userDir, "src", "web-sf-helps", "src");
+        string helpDir = Path.Join(userDir, "src", "web-sf-helps", "src");
         string source = "en";
-        string sourceDir = Path.Combine(helpDir, source);
-        string sourceWhxDir = Path.Combine(sourceDir, "whxdata");
+        string sourceDir = Path.Join(helpDir, source);
+        string sourceWhxDir = Path.Join(sourceDir, "whxdata");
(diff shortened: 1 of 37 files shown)


RaymondLuong3 (approved):
I checked out the branch and was able to connect to a new project and see the sync succeed. I approve.

> Reviewed 37 of 37 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3150#-:-ONqMEhk3AT0acwWz8aa:bqwkxtm)* status: all files reviewed, 1 unresolved discussion (waiting on @pmachapman)

___
*[`src/Migrations/PTDDCloneAll/CloneAllService.cs` line 75 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3150#-ONqGJrf2eYxROPs9SZQ:-ONqGJrf2eYxROPs9SZR:b5fzpoc) ([raw file](https://github.com/sillsdev/web-xforge/blob/29386d835f971541393d921fc5ae1778291baf62/src/Migrations/PTDDCloneAll/CloneAllService.cs#L75)):*
> ```Smalltalk
>         public async Task CloneSFProjects(string mode, IEnumerable<SFProject> projectsToClone)
>         {
>             string syncDir = Path.Combine(_siteOptions.Value.SiteDir, "sync");
> ```

I want to say we probably do not need to update these since they are checked into the repo for historical purposes, but in reality there is no harm making this change.



_Code quote:_
```Smalltalk
            string syncDir = Path.Combine(_siteOptions.Value.SiteDir, "sync");
```

PR #3154, opened 2025-04-10
Improve theme support for drafting

nigel-wells:
This touches most community draft generation related components where needed.

There are some colors left which need to be addressed separately i.e. the event log icon colors.

Dark mode improvements have been made.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3154)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/_book-multi-select-theme.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/_book-multi-select-theme.scss
@@ -1,0 +1,13 @@
+@use '@angular/material' as mat;
+
+@mixin color($theme) {
+  $is-dark: mat.get-theme-type($theme) == dark;
+
+  --sf-book-border-fill: #{mat.get-theme-color($theme, primary, if($is-dark, 90, 10))};
+}
+
+@mixin theme($theme) {
+  @if mat.theme-has($theme, color) {
+    @include color($theme);
+  }
+}
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.scss
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.scss
@@ -20,7 +20,7 @@
     bottom: 0;
     left: 0;
     height: 3px;
-    background: black;
+    background: var(--sf-book-border-fill);
   }
 }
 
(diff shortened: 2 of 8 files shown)


Nateowami:
Checking this out locally I think it looks good. Thanks for working on it.

Nateowami (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3154)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nigel-wells)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.scss` line 58 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3154#-ON_EWOx-NP3vGJpRFlT:-ON_EWOx-NP3vGJpRFlU:b-nrzrtt) ([raw file](https://github.com/sillsdev/web-xforge/blob/d92f473fc6bff5f2cad93f0e0783c33e2913f1e1/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.scss#L58)):*
> ```SCSS
> .step-subtitle {
>   font-size: 0.9em;
>   opacity: 0.6;
> ```

In this situation I don't think it's going to cause problems, but we should avoid using opacity for text. This seems like a neutral color we could define once for the entire theme and then use throughout the application.

nigel-wells (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3154)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.scss` line 58 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3154#-ON_EWOx-NP3vGJpRFlT:-ONm90hNBECLWUMVhV7u:b-j2jpw8) ([raw file](https://github.com/sillsdev/web-xforge/blob/a28884b4b5ee5b6cf7c8f6a670fa23e70acbdb11/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.scss#L58)):*

In this situation I don't think it's going to cause problems, but we should avoid using opacity for text. This seems like a neutral color we could define once for the entire theme and then use throughout the application.
</blockquote></details>

FYI opacity on an element is ok and doesn't interfere with rendering characters. Transparency on the color itself is where it runs into weird overlay issues. Here is an example in Arabic - the left is using `opacity: 0.3` on the element and the right is a transparency `color: #00000033`


Yes, an actual color could be defined. Opacity can often be a nice trick to give the appearance of a secondary color across multiple themes without being explicit in each theme.

PR #3167, opened 2025-04-16
SF-3281: Show navigation button when questions list is empty

josephmyers:
This adds a button to the questions list when it is empty and there are unanswered questions. It will take the user to their first unanswered question.

This required a query to the questions service in the questions list, since the questions component is simply given its questions, based on the filter-range that lives in the checking mega-class. Although most of the questions component is written to update imperatively, this query needs to keep itself up to date to support remote changes like archiving.

When the button is clicked, we activate the question, accompanied by the list change flag set to false, which closes the list flyout on mobile. When there are no unanswered questions for the user, we simply display the previous message (tailored to be clearer).

I also tweaked a few things that were less than ideal around the range-filter:
- Hiding the "Show All" text when there is no filter applied, since this looked more like a command to do something, and it clears up unnecessary bloat text.
- Changed the padding and margins on the headers and list content when there are no questions. This helps the page flow better and look more visually pleasing.


marksvc (changes requested):
:wrench: I get an error dialog if I do the following.
- Open a window to the questions and answers page.
- Make sure all questions are answered. Be at some book, RUT.
- Make sure the filter is set to only show questions from the Current book.
- Open another window to the Manage questions page.
- Add a new questions for some other book, NUM.
- The window at the questions and answers page shows the "Go to questions" button, but also an error dialog appears, saying 
> An error has occurred
> 
> Document already exists
> 
> ...
> ShareDBError: Document already exists
>     at push.9838.Doc.create (http://localhost:5000/vendor.js:51514:15)

I tried on branch master and found that the error did not occur there.
See attached video showing causing the problem.
[Screencast from 2025-04-16 11-36-13.webm](https://files.reviewableusercontent.io/images/reviewable/7265309/O0dUgoH0qS_SxTKK0fmTy7iP/Screencast%20from%202025-04-16%2011-36-13.webm)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-:-ONzeAnLBszRV5vrCbla:b-9qc8v1)* status: all files reviewed, 4 unresolved discussions (waiting on @josephmyers)

___
*[`-- commits` line 9 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONzY03m1PKoKWxBizbp:-ONzY03n5eu80wY4gxXg:buyxi74):*
:speech_balloon: I appreciated the explanations in the commit message, that I could first read and learn what is being sought in the PR.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json` line 298 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONzYmqo2qP07lqRqXda:-ONzYmqo2qP07lqRqXdb:bzien58) ([raw file](https://github.com/sillsdev/web-xforge/blob/c542e780fb7dd09e451f913d7790a40fea95e399/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json#L298)):*
> ```JSON
>     "listen_to_question": "Listen to the question for {{ referenceForDisplay }}",
>     "loading": "Loading...",
>     "no_questions_found": "No questions found for the settings above.",
> ```

I like this improvement of "for the settings above" from "filter".
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 146 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONzgdpZ-MGis0r3aTOz:-ONzgdpZ-MGis0r3aTP-:b-ero5ef) ([raw file](https://github.com/sillsdev/web-xforge/blob/c542e780fb7dd09e451f913d7790a40fea95e399/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L146)):*
> ```XML
> 
>   constructor(
>     private readonly questionsService: CheckingQuestionsService,
> ```

:speech_balloon: BTW, I'm not saying you should change to this, but I just wanted to mention, it occurred to me that a way to get similar behaviour but without hooking checking-questions.component to `CheckingQuestionsService` (if that would be desirable)  would be to have on checking-questions.component.ts something like
```ts
  @Input() thereAreUnansweredQuestionsIHaventToldYouAbout boolean;
  @Output() theUserWantsToGoToAnUnansweredQuestionYouHaventToldMeAbout = new EventEmitter();

  get hasUnansweredQuestion(): boolean {
    return this.thereAreUnansweredQuestionsIHaventToldYouAbout;
  }

  protected emitNewQuestion(): void {
    this.theUserWantsToGoToAnUnansweredQuestionYouHaventToldMeAbout.emit();
  }
```
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 217 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONz_kdrCjxhLYaBPutF:-ONz_kdrCjxhLYaBPutG:b-u73dxy) ([raw file](https://github.com/sillsdev/web-xforge/blob/c542e780fb7dd09e451f913d7790a40fea95e399/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L217)):*
> ```XML
>   }
> 
>   protected emitNewQuestion(): void {
> ```

:art: You might alternatively consider a method name of `goToUnsaweredQuestion` or `emitUnsaweredQuestion`.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 219 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONza9JAC_pWDRwfpfrn:-ONza9JAC_pWDRwfpfro:b41d721) ([raw file](https://github.com/sillsdev/web-xforge/blob/c542e780fb7dd09e451f913d7790a40fea95e399/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L219)):*
> ```XML
>   protected emitNewQuestion(): void {
>     if (this._firstUnansweredQuestion === undefined) return;
>     this.activateQuestion(this._firstUnansweredQuestion.docs[0], { isQuestionListChange: false });
> ```

:wrench: When I see `[0]`, I think "Could  be undefined!" :)
What do you think about writing something like this instead?
```ts
  protected emitNewQuestion(): void {
    const firstDoc: QuestionDoc | undefined = this._firstUnansweredQuestion?.docs[0];
    if (firstDoc == null) return;
    this.activateQuestion(firstDoc, { isQuestionListChange: false });
  }
```

josephmyers (commented):
Hm, I'm not able to reproduce this on my side, and this error smells funny. Maybe we could talk about it after (or before) the retrospective today? You can likely walk me through what I'm doing wrong.

Edit: it's a tiny bit odd that you have two windows open with the same user account. I almost never do this. Is this a normal, valid user workflow? I may need to improve my developer testing

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-:-OO098qh3AII41VLEO4V:bde3ufx)* status: all files reviewed, 4 unresolved discussions (waiting on @josephmyers)

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3167)* status: 4 of 7 files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 146 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONzgdpZ-MGis0r3aTOz:-OO0ZpjVEZ8F9APsw05S:bl8si5z) ([raw file](https://github.com/sillsdev/web-xforge/blob/c542e780fb7dd09e451f913d7790a40fea95e399/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L146)):*

:speech_balloon: BTW, I'm not saying you should change to this, but I just wanted to mention, it occurred to me that a way to get similar behaviour but without hooking checking-questions.component to `CheckingQuestionsService` (if that would be desirable)  would be to have on checking-questions.component.ts something like
```ts
  @Input() thereAreUnansweredQuestionsIHaventToldYouAbout boolean;
  @Output() theUserWantsToGoToAnUnansweredQuestionYouHaventToldMeAbout = new EventEmitter();

  get hasUnansweredQuestion(): boolean {
    return this.thereAreUnansweredQuestionsIHaventToldYouAbout;
  }

  protected emitNewQuestion(): void {
    this.theUserWantsToGoToAnUnansweredQuestionYouHaventToldMeAbout.emit();
  }
```
</blockquote></details>

Yes, but this seems to go against OOP encapsulation. It's better for the class to be able to take care of itself, instead of requiring every caller to tell it how to do its job. Unless I'm not understanding fully
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 217 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONz_kdrCjxhLYaBPutF:-OO0gcdN-2UttA7V4-6N:bsmgqgl) ([raw file](https://github.com/sillsdev/web-xforge/blob/c542e780fb7dd09e451f913d7790a40fea95e399/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L217)):*

:art: You might alternatively consider a method name of `goToUnsaweredQuestion` or `emitUnsaweredQuestion`.
</blockquote></details>

Done. Thanks, I forgot to fix this before checking in!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 219 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONza9JAC_pWDRwfpfrn:-OO0hd745rt3qAj0gkrJ:b-ku17z3) ([raw file](https://github.com/sillsdev/web-xforge/blob/c542e780fb7dd09e451f913d7790a40fea95e399/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L219)):*

:wrench: When I see `[0]`, I think "Could  be undefined!" :)
What do you think about writing something like this instead?
```ts
  protected emitNewQuestion(): void {
    const firstDoc: QuestionDoc | undefined = this._firstUnansweredQuestion?.docs[0];
    if (firstDoc == null) return;
    this.activateQuestion(firstDoc, { isQuestionListChange: false });
  }
```
</blockquote></details>

Done. How about using the check I made just above?
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json` line 295 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONzSjtpB8DhSgEWg9Wm:-OO0hpQOB-vZGuBemlhY:bkfktfh) ([raw file](https://github.com/sillsdev/web-xforge/blob/c542e780fb7dd09e451f913d7790a40fea95e399/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json#L295)):*

Hmm. I might have thought it generally not to be a desirable practice to say things like "Click here to" in the UI. Mightn't it be preferable to instead have action labels, like `<a>Go to the first unanswered questions</a>` or `<a>First unanswered questions</a>`?
I am still learning.

Because there is a big purple button saying "Go to questions", I would be inclined to omit the "Click here to go to the first question you haven't answered:" label. Without it, I see 
```
<i>No questions found for the settings above</i>

<big button>Go to questions</big button>
```
, which seems pretty obvious and easy to get to the questions if I don't see them in the list. 

What are your thoughts on removing the "Click here" label and the button being sufficient for guiding the user?
</blockquote></details>

I don't think you could say "Click here to" is always bad UI -- that's likely a preference. Good UI is simple and clear (and accurate), and there could be cases where saying that meets those criteria.

I personally like striking the label. My biggest concern, however, is that it's not super clear what questions we're referring to without the explanation. (I also don't like the messages you suggested which would create a big button, and I didn't like any of mine either.)

What the software is actually doing is kind of hard to explain concisely, so we can either:
- explain with an additional label
- not explain it

In the past, I've gotten docked for not being clear enough, or oversimplifying to the detriment of precision, so I was avoiding that here. But if you agree that it's not necessary to specify which questions we're taking them to, I can happily get on board with simply saying "Go to questions," without increasing the precision of that wording at all.

marksvc (changes requested):
(We were not able to reproduce the error.)

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-:-OO4q7pK9VMw3lUQfioR:bmiw6f9)* status: all files reviewed, 1 unresolved discussion (waiting on @josephmyers and @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 146 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONzgdpZ-MGis0r3aTOz:-OO4k53Q-LcIvHwR4Lgr:b-fhf4rw) ([raw file](https://github.com/sillsdev/web-xforge/blob/c542e780fb7dd09e451f913d7790a40fea95e399/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L146)):*

Yes, but this seems to go against OOP encapsulation. It's better for the class to be able to take care of itself, instead of requiring every caller to tell it how to do its job. Unless I'm not understanding fully
</blockquote></details>

I remember a talk at a department conference many years ago where the virtues were promoted of keeping different parts of a system from knowing too much about each other. In this case, an advantage would be that fewer leaf pieces in the program have access back into the system data and need to be able to work with its API. In this way, it might increase reusability of the leaf components in other contexts, and make it easier to change APIs for working with system data because less parts of the program are directly touching them. 

There are different approaches. There are advantages to passing in a `CheckingQuestionsService`.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 219 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONza9JAC_pWDRwfpfrn:-OO4mOC4Ef3BltMPFZNE:b-c7bdva) ([raw file](https://github.com/sillsdev/web-xforge/blob/c542e780fb7dd09e451f913d7790a40fea95e399/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L219)):*

Done. How about using the check I made just above?
</blockquote></details>

I feel more comfortable with a check closer to the actual properties being used rather than a sort of anecdotal evidence check. But I can still approve `if (!this.hasUnansweredQuestion) return;`. Hmm, though, it's vulnerable to future changes to the `hasUnsaweredeQuestion()` method. :/

That said, I see a recent change is introducing a `!` non-null assertion operator. With all the errors on QA and Live from non-null assertion operators, I don't think we should put this in. 

I think we will at least still want to change 
```ts
if (!this.hasUnansweredQuestion) return;
this.activateQuestion(this._firstUnansweredQuestion!.docs[0], { isQuestionListChange: false });
```
to something like
```ts
if (!this.hasUnansweredQuestion || this._firstUnansweredQuestion == null) return;
this.activateQuestion(this._firstUnansweredQuestion.docs[0], { isQuestionListChange: false });
```
to remove the `!`.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json` line 295 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONzSjtpB8DhSgEWg9Wm:-OO4pQjz5I14bIlF6Wd3:b7qaj3o) ([raw file](https://github.com/sillsdev/web-xforge/blob/c542e780fb7dd09e451f913d7790a40fea95e399/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json#L295)):*

I don't think you could say "Click here to" is always bad UI -- that's likely a preference. Good UI is simple and clear (and accurate), and there could be cases where saying that meets those criteria.

I personally like striking the label. My biggest concern, however, is that it's not super clear what questions we're referring to without the explanation. (I also don't like the messages you suggested which would create a big button, and I didn't like any of mine either.)

What the software is actually doing is kind of hard to explain concisely, so we can either:
- explain with an additional label
- not explain it

In the past, I've gotten docked for not being clear enough, or oversimplifying to the detriment of precision, so I was avoiding that here. But if you agree that it's not necessary to specify which questions we're taking them to, I can happily get on board with simply saying "Go to questions," without increasing the precision of that wording at all.
</blockquote></details>

Okay. I'm okay with it like it is. BTW some ideas for different presentations are:

- Button saying "Unanswered questions"
- Colored link text saying "Unanswered questions"

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3167)* status: all files reviewed, 1 unresolved discussion (waiting on @marksvc)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 146 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONzgdpZ-MGis0r3aTOz:-OOKhk8e-17pXoJD4o7J:b-rhp9yz) ([raw file](https://github.com/sillsdev/web-xforge/blob/c542e780fb7dd09e451f913d7790a40fea95e399/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L146)):*

I remember a talk at a department conference many years ago where the virtues were promoted of keeping different parts of a system from knowing too much about each other. In this case, an advantage would be that fewer leaf pieces in the program have access back into the system data and need to be able to work with its API. In this way, it might increase reusability of the leaf components in other contexts, and make it easier to change APIs for working with system data because less parts of the program are directly touching them. 

There are different approaches. There are advantages to passing in a `CheckingQuestionsService`.
</blockquote></details>

Ok, we're in the conceptual world here. The problem with this design would be that you have to reimplement the "leaf code" in every place you use the leaf component, instead of just once. That would get cumbersome (and buggy) very quickly, not to mention greater test burden.

However, if you never end up reusing the leaf component, I can see your strategy being potentially easier. But the limitation is that you have to centralize all creations of the leaf components, given that you're trying to minimize the usages of your internal services. If you don't, you're still going to have the same refactor burden, and you might as well have just put that code in the leaf component to start with.

It seems like, from what you've said, that the main benefit of this alternate way would be easier component reuse due to using data types as input instead of services, since you can then freely change internal API without refactor, but I'd argue that this way could be marginally easier if done very specifically but would be a good bit more cumbersome and buggy in every other case.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts` line 219 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONza9JAC_pWDRwfpfrn:-OOKu0ej52S_8lHM5eeG:bpxwblk) ([raw file](https://github.com/sillsdev/web-xforge/blob/c542e780fb7dd09e451f913d7790a40fea95e399/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-questions/checking-questions.component.ts#L219)):*

I feel more comfortable with a check closer to the actual properties being used rather than a sort of anecdotal evidence check. But I can still approve `if (!this.hasUnansweredQuestion) return;`. Hmm, though, it's vulnerable to future changes to the `hasUnsaweredeQuestion()` method. :/

That said, I see a recent change is introducing a `!` non-null assertion operator. With all the errors on QA and Live from non-null assertion operators, I don't think we should put this in. 

I think we will at least still want to change 
```ts
if (!this.hasUnansweredQuestion) return;
this.activateQuestion(this._firstUnansweredQuestion!.docs[0], { isQuestionListChange: false });
```
to something like
```ts
if (!this.hasUnansweredQuestion || this._firstUnansweredQuestion == null) return;
this.activateQuestion(this._firstUnansweredQuestion.docs[0], { isQuestionListChange: false });
```
to remove the `!`.
</blockquote></details>

Done. We're never setting this member to null, but I am happy to avoid the assertion.

marksvc (approved):
Thanks for your work on this!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-:-OOTXf0D5JcNEKmUck5B:b7tt6lp)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

jellomoosie:
```
[src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json line 298 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONzYmqo2qP07lqRqXda:-ONzYmqo2qP07lqRqXdb:bzien58) ([raw file](https://github.com/sillsdev/web-xforge/blob/c542e780fb7dd09e451f913d7790a40fea95e399/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json#L298)):

    "listen_to_question": "Listen to the question for {{ referenceForDisplay }}",
    "loading": "Loading...",
    "no_questions_found": "No questions found for the settings above.",
I like this improvement of "for the settings above" from "filter".
```

I found this wording change to be confusing. "Settings above" is more general and more likely to require changes if the UI layout changes in the future, while "filter" is more specific and matches the icon. Is there a compelling reason to make the change, which will also have a cascading effect on the localized versions of this page?

marksvc:
> ```
> [src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json line 298 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3167#-ONzYmqo2qP07lqRqXda:-ONzYmqo2qP07lqRqXdb:bzien58) ([raw file](https://github.com/sillsdev/web-xforge/blob/c542e780fb7dd09e451f913d7790a40fea95e399/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/checking_en.json#L298)):
> 
>     "listen_to_question": "Listen to the question for {{ referenceForDisplay }}",
>     "loading": "Loading...",
>     "no_questions_found": "No questions found for the settings above.",
> I like this improvement of "for the settings above" from "filter".
> ```
> 
> I found this wording change to be confusing. "Settings above" is more general and more likely to require changes if the UI layout changes in the future, while "filter" is more specific and matches the icon. Is there a compelling reason to make the change, which will also have a cascading effect on the localized versions of this page?

I wonder if "filter" will be an unfamiliar term for users. Or maybe they won't really understand that the funnel and text above is what is being referred to by "filter". Maybe it will be clear.  I am okay with it changing if you prefer.

jellomoosie:
Maybe "Selections" would be better than "settings" or "filter"

PR #3169, opened 2025-04-17
Fix border radius on draft sources stepper

Nateowami:
See Chromatic for the before/after.

For a while I was confused why this mat-card didn't have the border radius applied properly, but the other cards on the page did, and there didn't appear to be any CSS difference. Eventually I concluded:
- If you set a background on an element, and a border-radius, that background follows the border radius
- If you have an element with a border radius, and inside another element with a background (but no border radius), you need to set `overflow: hidden` for the parent to constrain the background of the child elements.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3169)


RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3169)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #3184, opened 2025-05-01
SF-3349 Prompt user when leaving draft sources with unsaved changes

RaymondLuong3:
This change introduces a confirmation when a user tries to navigate away from the draft sources page when there are unsaved changes. It uses the window.confirm api to show the prompt. This shows the prompt when a user clicks back/forward on the browser or selects a page to navigate to so that the user will not mistakenly leave the page without saving their changes.j

Unfortunately this does not behave the same way if a user edits the url or refreshes the page, but since that is not as likely I think this is sufficient for the task.

image

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3184)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/project-router.guard.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/project-router.guard.spec.ts
@@ -1,0 +1,55 @@
+import { TestBed } from '@angular/core/testing';
+import { mock } from 'ts-mockito';
+import { AuthGuard } from 'xforge-common/auth.guard';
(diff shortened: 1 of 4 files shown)


pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3184#-:-OPRz1KGCXmLYjRwRKj0:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @RaymondLuong3)

josephmyers:
@RaymondLuong3 was this an accidental merge?

PR #3191, opened 2025-05-07
SF-3349b Only prompt user about unsaved source changes

RaymondLuong3:
This is a follow-up PR for SF-3349. If the user has saved their changes, then they will be able to navigate away from the configure sources page without the prompt about unsaved changes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3191)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/project-router.guard.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/project-router.guard.spec.ts
@@ -16,34 +16,16 @@ describe('DraftNavigationAuthGuard', () => {
     ]
   }));
 
-  it('can navigate away when no changes', () => {
+  it('can navigate away when no changes', async () => {
     // navigate away
     const env = new DraftNavigationTestEnvironment();
-    spyOn(window, 'confirm');
-    expect(
-      env.service.canDeactivate({ deactivationPrompt: 'unsaved changed', promptUserToDeactivate: () => false })
-    ).toBe(true);
-    expect(window.confirm).not.toHaveBeenCalled();
+    expect(await env.service.canDeactivate({ confirmLeave: () => Promise.resolve(true) })).toBe(true);
   });
 
-  it('can shows prompt and navigate away', () => {
+  it('can shows prompt and stay on page', async () => {
     // navigate away
     const env = new DraftNavigationTestEnvironment();
-    spyOn(window, 'confirm').and.returnValue(true);
-    expect(
-      env.service.canDeactivate({ deactivationPrompt: 'unsaved changed', promptUserToDeactivate: () => true })
-    ).toBe(true);
-    expect(window.confirm).toHaveBeenCalled();
-  });
-
-  it('can shows prompt and stay on page', () => {
-    // navigate away
-    const env = new DraftNavigationTestEnvironment();
-    spyOn(window, 'confirm').and.returnValue(false);
-    expect(
-      env.service.canDeactivate({ deactivationPrompt: 'unsaved changed', promptUserToDeactivate: () => true })
-    ).toBe(false);
-    expect(window.confirm).toHaveBeenCalled();
+    expect(await env.service.canDeactivate({ confirmLeave: () => Promise.resolve(false) })).toBe(false);
   });
 });
 
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/project-router.guard.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/project-router.guard.ts
@@ -169,23 +169,20 @@ export class TranslateAuthGuard extends RouterGuard {
   }
 }
 
-export interface DeactivateAllowed {
-  deactivationPrompt: string;
-
-  promptUserToDeactivate(): boolean;
+export interface ConfirmOnLeave {
+  confirmLeave(): Promise<boolean>;
 }
 
 @Injectable({
   providedIn: 'root'
 })
-export class DraftNavigationAuthGuard extends RouterGuard implements CanDeactivate<DeactivateAllowed> {
+export class DraftNavigationAuthGuard extends RouterGuard implements CanDeactivate<ConfirmOnLeave> {
   constructor(authGuard: AuthGuard, projectService: SFProjectService) {
     super(authGuard, projectService);
   }
 
-  canDeactivate(component: DeactivateAllowed): boolean {
-    if (!component.promptUserToDeactivate()) return true;
-    return confirm(component.deactivationPrompt);
+  async canDeactivate(component: ConfirmOnLeave): Promise<boolean> {
+    return await component.confirmLeave();
   }
 
   check(_: SFProjectProfileDoc): boolean {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.spec.ts
@@ -96,6 +96,7 @@ describe('DraftSourcesComponent', () => {
       tick();
       env.fixture.detectChanges();
       env.clickLanguageCodesConfirmationCheckbox();
+      expect(env.component.promptUserToDeactivate()).toBe(true);
 
       // Suppose the user loads up the sources configuration page, changes no projects, and clicks Save. The settings
       // change request will just correspond to what the project already has for its settings.
@@ -119,6 +120,7 @@ describe('DraftSourcesComponent', () => {
         mockedSFProjectService.onlineUpdateSettings
       ).last()[1];
       expect(actualSettingsChangeRequest).toEqual(expectedSettingsChangeRequest);
+      expect(env.component.promptUserToDeactivate()).toBe(false);
     }));
 
     it('clearing second training source works', fakeAsync(() => {
@@ -126,6 +128,7 @@ describe('DraftSourcesComponent', () => {
       tick();
       env.fixture.detectChanges();
       env.clickLanguageCodesConfirmationCheckbox();
+      expect(env.component.promptUserToDeactivate()).toBe(true);
 
       // Suppose the user loads up the page, clears the second training/reference project box, and clicks Save. The
       // settings change request will show a requested change for unsetting the additional-training-source.
@@ -156,6 +159,7 @@ describe('DraftSourcesComponent', () => {
         mockedSFProjectService.onlineUpdateSettings
       ).last()[1];
       expect(actualSettingsChangeRequest).toEqual(expectedSettingsChangeRequest);
+      expect(env.component.promptUserToDeactivate()).toBe(false);
     }));
 
     it('clearing first training source works', fakeAsync(() => {
@@ -163,6 +167,7 @@ describe('DraftSourcesComponent', () => {
       tick();
       env.fixture.detectChanges();
       env.clickLanguageCodesConfirmationCheckbox();
+      expect(env.component.promptUserToDeactivate()).toBe(true);
 
       // Suppose the user comes to the page, leaves the second reference/training project selection alone, and clears
       // the first reference/training project selection. Let's respond by clearing the additional-training-source, and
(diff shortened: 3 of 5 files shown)


josephmyers (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3191)* status: 0 of 2 files reviewed, 5 unresolved discussions

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.spec.ts` line 99 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3191#-OPiLZs_61lP7Fyj2Uhy:-OPiLZs_61lP7Fyj2Uhz:bnwvc6u) ([raw file](https://github.com/sillsdev/web-xforge/blob/4599de5be3f1085be22d21af0979ed76c08793fc/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.spec.ts#L99)):*
> ```XML
>       env.fixture.detectChanges();
>       env.clickLanguageCodesConfirmationCheckbox();
>       expect(env.component.promptUserToDeactivate()).toBe(true);
> ```

It's a bit of a concern, conceptually, that throughout this file starting up `TestEnvironment` trips the prompt flag. What is actually causing this? I bring this up not to be picky but because it'd be valuable to add a test that verifies we can initialize and verify that the prompt flag is `false` before any changes are made.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts` line 64 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3191#-OPi9RxNAnLeeJyiS4pB:-OPi9RxO2EUWVTDxnqRe:b-d8m175) ([raw file](https://github.com/sillsdev/web-xforge/blob/4599de5be3f1085be22d21af0979ed76c08793fc/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts#L64)):*
> ```XML
>   styleUrl: './draft-sources.component.scss'
> })
> export class DraftSourcesComponent extends DataLoadingComponent implements DeactivateAllowed {
> ```

`DeactivateAllowed` is confusing terminology to me. What does deactivate mean? If it's the same as leaving the page, why not call it that? What does it mean to be allowed or not? It's not that it's not allowed, it's just that we're prompting them beforehand.

I recommend something like `PromptOnLeave` or `ConfirmOnLeave`.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts` line 85 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3191#-OPiAKtM2Ej_CyI6437I:-OPiAKtM2Ej_CyI6437J:b-disv5f) ([raw file](https://github.com/sillsdev/web-xforge/blob/4599de5be3f1085be22d21af0979ed76c08793fc/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts#L85)):*
> ```XML
>   languageCodeConfirmationMessageIfUserTriesToContinue: I18nKeyForComponent<'draft_sources'> | null = null;
>   clearLanguageCodeConfirmationCheckbox = new EventEmitter<void>();
>   changesMade = false;
> ```

This should be protected.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts` line 242 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3191#-OPiJJUr6d2Kq6lSs72F:-OPiJJUr6d2Kq6lSs72G:b-p4ibj9) ([raw file](https://github.com/sillsdev/web-xforge/blob/4599de5be3f1085be22d21af0979ed76c08793fc/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts#L242)):*
> ```XML
>     if (selectedProject?.paratextId === array[index]?.paratextId) return;
> 
>     this.changesMade = true;
> ```

I know you didn't design this, but the downside of doing it this way is that if you actually undo your changes, we still show the prompt. Couldn't you pretty easily change `changesMade` to a getter that checks `draftingSources` and `trainingSources` against what's been saved to the project?

It's not absolutely critical, but it does elevate our user experience from a little silly to somewhat intelligent.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts` line 283 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3191#-OPiD9Kb0q6q04eKbIxz:-OPiD9KcEB_tBVeGSPnl:baz52kg) ([raw file](https://github.com/sillsdev/web-xforge/blob/4599de5be3f1085be22d21af0979ed76c08793fc/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts#L283)):*
> ```XML
>     const leavePage =
>       !this.changesMade ||
>       (await this.dialogService.confirm(
> ```

It's strange that we now have two different dialogs that show, dependent on how the user leaves the page. It looks like Nathaniel was already trying to account for confirming discarding changes, he just only checked the Cancel button. I think your solution is better, because it covers many more use cases, including this one. So you can remove all this code and just leave the call to `navigateToDrafting`. (I do like his usage of dialogService for a better dialog, but what you have works, too.)

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3191)* status: 0 of 2 files reviewed, 6 unresolved discussions (waiting on @josephmyers and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts` line 242 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3191#-OPiJJUr6d2Kq6lSs72F:-OPkIHpxBNh8DQxW_NQD:bm8wdaf) ([raw file](https://github.com/sillsdev/web-xforge/blob/4599de5be3f1085be22d21af0979ed76c08793fc/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts#L242)):*

I know you didn't design this, but the downside of doing it this way is that if you actually undo your changes, we still show the prompt. Couldn't you pretty easily change `changesMade` to a getter that checks `draftingSources` and `trainingSources` against what's been saved to the project?

It's not absolutely critical, but it does elevate our user experience from a little silly to somewhat intelligent.
</blockquote></details>

This is more work than it's worth.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts` line 283 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3191#-OPiD9Kb0q6q04eKbIxz:-OPkGiDdBEvFf_I0C81r:bih1nbs) ([raw file](https://github.com/sillsdev/web-xforge/blob/4599de5be3f1085be22d21af0979ed76c08793fc/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts#L283)):*

It's strange that we now have two different dialogs that show, dependent on how the user leaves the page. It looks like Nathaniel was already trying to account for confirming discarding changes, he just only checked the Cancel button. I think your solution is better, because it covers many more use cases, including this one. So you can remove all this code and just leave the call to `navigateToDrafting`. (I do like his usage of dialogService for a better dialog, but what you have works, too.)
</blockquote></details>

The dialog is consistent with the rest of our application, localizeable, and easier to integrate e2e tests with. 
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts` line 294 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3191#-OPkHwiVAb-tuCp5WXzE:-OPkHwiVAb-tuCp5WXzF:bxsnlb6) ([raw file](https://github.com/sillsdev/web-xforge/blob/4599de5be3f1085be22d21af0979ed76c08793fc/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts#L294)):*
> ```XML
> 
>   promptUserToDeactivate(): boolean {
>     return this.changesMade && !this.allProjectsSavedAndSynced;
> ```

So if it fails to save, and tells the user it failed to save, and then the user tries to leave, we prompt the user? If I'm reading this correctly? That seems like it would be annoying. 

I'm not sure we need to force/encourage the user to stay on this page until the syncing is complete. Those are background jobs that *can* fail, but doesn't have to have the user watch them. It's also fairly obvious the page is doing something, so the user is unlikely to mistakenly leave the page. What we really need to prevent is the user not realizing there is a need to click the save button.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3191)* status: 0 of 2 files reviewed, 7 unresolved discussions (waiting on @josephmyers and @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts` line 293 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3191#-OPkJIc86QkmX0OyEJWO:-OPkJIc86QkmX0OyEJWP:b5ym222) ([raw file](https://github.com/sillsdev/web-xforge/blob/4599de5be3f1085be22d21af0979ed76c08793fc/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts#L293)):*
> ```XML
>   }
> 
>   promptUserToDeactivate(): boolean {
> ```

This isn't named correctly.

Nateowami (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3191)* status: 0 of 5 files reviewed, 5 unresolved discussions (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts` line 283 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3191#-OPiD9Kb0q6q04eKbIxz:-OPlnyqNCejkvOGwG6h2:b4exoqa) ([raw file](https://github.com/sillsdev/web-xforge/blob/4599de5be3f1085be22d21af0979ed76c08793fc/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts#L283)):*

I agree that the dialog is nice because we can theme it to look consistent with the app. Initially I wanted to reuse the dialog, but the 'canDeactivate` API needs a synchronous method that we get when using `window.confirm`. I've removed this dialog since I wasn't able to reuse it.
</blockquote></details>

Did you check CanDeactivateFn?

josephmyers (commented):
> Reviewed 4 of 5 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3191)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

RaymondLuong3 (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3191)* status: 1 of 6 files reviewed, 1 unresolved discussion (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts` line 283 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3191#-OPiD9Kb0q6q04eKbIxz:-OPqG1UjEbrjfHv_w8PU:b6w0979) ([raw file](https://github.com/sillsdev/web-xforge/blob/4599de5be3f1085be22d21af0979ed76c08793fc/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-sources/draft-sources.component.ts#L283)):*

Did you check CanDeactivateFn?
</blockquote></details>

I did some more testing, and it looks like we can provide an asynchronous function for the canDeactivate route guard. This means we can reuse the dialog to confirm if the user wants to leave the page! I think this will suit our needs.

josephmyers (approved):
Thanks, Raymond, this looks good to me 
:lgtm:

> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3191#-:-OQ1Z3Gt5CxXmbxrb7wl:by176i4)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

josephmyers:
I'm not sure if Nathaniel wants further input on this, but if not, feel free to move this on to the test team!

PR #2780, opened 2024-10-07
SF-2970 Fix scripture ref playback when question updated

RaymondLuong3:
The audio source URL was not being updated when a question's verse ref is updated. This meant that in the case where a question's reference is changed, the audio url did not and the wrong scripture is played back. This PR adds an observable to the CheckingQuestionComponent that emits a URL based on the question's current chapter and verse.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2780)

PR #3192, opened 2025-05-08
SF-3334 Fix bug with updating scripture audio

josephmyers:
This change undoes a part of #2780, which introduced a bug causing checking-question to not update correctly when chapter audio is uploaded. The core of that fix, refreshing audio when the question doc changes, is still intact.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3192)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/checking/checking/checking-answers/checking-question/checking-question.component.html
@@ -1,13 +1,13 @@
 <ng-container *transloco="let t; read: 'checking_questions'">
   <div>
-    @if (scriptureAudioUrl$ | async; as audioSource) {
+    @if (scriptureAudioUrl != null) {
       <div class="scripture-audio" [class]="focusedText">
         <span class="spacer"></span>
         <app-single-button-audio-player
           #scriptureAudio
           id="scriptureAudio"
           [class]="focusedText"
-          [source]="audioSource"
+          [source]="scriptureAudioUrl"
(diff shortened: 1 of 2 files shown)


RaymondLuong3 (changes requested):
I see this fixes the problem where the audio that is uploaded for the chapter is available to be played for the active question, but I see that the question's text is focused and not the play question verse audio. I think it would make more sense to focus the question verse audio as if the user was navigating to the question for the first time.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3192#-:-OPlqFh57IcvJwrCoOM9:b-x4ch3f)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (commented):
Sure! I went back and found that the old code was preserving the question text selection. But I think your recommendation is good, and the code appears to be in place for it to happen. The reason it's not happening is because when it updates the focus, the audio hasn't finished uploading yet. So I loosened the focus logic, under a hopefully safe assumption.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3192#-:-OPmcUGS5iosOakFsAos:b-gw47ct)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

RaymondLuong3 (approved):
:lgtm: 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3192#-:-OPp_dRTCX-NwitX8PMC:bg7igkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

PR #2965, opened 2025-01-20
SF-3166 Fix crash while retrieving history during sync

pmachapman:
When a user is retrieving the history for a chapter, and the project is being updated on disk by Paratext sync in another thread, the  get history request can fail with the following error:

```
System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
```

This PR fixes that issue, which was recreated via a unit test. I also took the opportunity to unit test the get snapshot and get revisions code that interacts with the Paratext project on disk via Mercurial.

I have marked this as not requiring testing, as this crash is really difficult to recreate, and since the unit test recreates the environment of the crash, all that is necessary is Developer testing to ensure that history revisions can be viewed and the snapshot retrieved via the frontend.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/2965)

PR #3196, opened 2025-05-11
SF-3166b Avoid using cache when generating revision summaries

pmachapman:
Although SF-3166 (#2965) greatly reduced the occurrence of the following error:

```
System.InvalidOperationException: Operations that change non-concurrent collections must have exclusive access. A concurrent update was performed on this collection and corrupted its state. The collection's state is no longer correct.
```

I notice that these still happen on bugsnag (https://app.bugsnag.com/sil-international/scripture-forge-v2-plus/errors/681c30c6d383ca457857e620). As I just had this error occur to me on dev, I was able to investigate the program state when this crash has occurred, and have determined it was caused by an internal non-thread safe cache that ParatextData's `HgRevisionCollection` uses.

As our code does not use the cache (due to only reading each revision summary once), I have updated `ParatextService.GetRevisionHistoryAsync()` to instantiate `RevisionChangeInfo` itself, and so avoid the problematic cache.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3196)

### DIFF
--- a/src/SIL.XForge.Scripture/Services/ParatextService.cs
+++ b/src/SIL.XForge.Scripture/Services/ParatextService.cs
@@ -2121,7 +2121,9 @@ public class ParatextService : DisposableBase, IParatextService
         foreach (HgRevision revision in revisionCollection)
         {
             // Get the revision summary to see if the book and chapter has changed
-            RevisionChangeInfo revisionSummary = revisionCollection.GetSummaryFor(revision);
+            // Note: revisionCollection.GetSummaryFor(revision) maintains a non-thread safe cache of the revision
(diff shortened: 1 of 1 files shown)


josephmyers (approved):
This makes sense. Thanks for the explanatory comment 
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3196#-:-OQ1pWwt959u--Mki3dR:b-2os0xd)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #3198, opened 2025-05-12
Fix realtime server crash when querying for notes for an invalid project

pmachapman:
This PR fixes a crash that occurs when a query specifies an invalid project id. This crash results in the entire real time server stopping until it is restarted by Hangfire.

You can recreate the crash and verify the fix by running the following code on the frontend, in a component that references the RealtimeService:
```ts
    const parameters: QueryParameters = {
      [obj<NoteThread>().pathStr(t => t.projectRef)]: { $ne: null }
    };
    await this.realtimeService.subscribeQuery(NoteThreadDoc.COLLECTION, parameters, this.destroyRef);
```

As this PR is very straightforward (adding a `.catch()` to the `.then()`), no testing is necessary.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3198)

### DIFF
--- a/src/RealtimeServer/scriptureforge/realtime-server.ts
+++ b/src/RealtimeServer/scriptureforge/realtime-server.ts
@@ -61,16 +61,18 @@ export default class SFRealtimeServer extends RealtimeServer {
           return;
         }
         const userId: string = context.agent.connectSession.userId;
-        this.getProject(context.query.projectRef).then(p => {
-          if (
-            p != null &&
-            !SF_PROJECT_RIGHTS.hasRight(p, userId, SFProjectDomain.PTNoteThreads, Operation.View) &&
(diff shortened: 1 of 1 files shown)


josephmyers (approved):
With your fix, and the code that produces the crash without it, I'm getting an exception; however, it doesn't crash, which I'm guessing is the point. Just wanted to make that clear. Thanks

:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3198#-:-OQ74-si8jgEbX751pT3:b-tmtnji)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

pmachapman:
> I'm getting an exception; however, it doesn't crash, which I'm guessing is the point.

Yes, that is the expected behavior.

PR #3200, opened 2025-05-14
SF-3360 Allow Restore for chapters with invalid USFM

josephmyers:
I removed the requirement that a chapter have valid USFM for the history revert feature to work. This allows users to restore revisions to get themselves out of unsupported USFM situations. I also verified that restoring a revision with unsupported USFM properly removes the ability to edit.

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.spec.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/core/text-doc.service.spec.ts
@@ -93,6 +93,96 @@ describe('TextDocService', () => {
     });
   });
 
+  describe('canRestore', () => {
+    it('should return false if the project is undefined', () => {
+      const env = new TestEnvironment();
+
+      // SUT
+      const actual: boolean = env.textDocService.canRestore(undefined, 1, 1);
+      expect(actual).toBe(false);
+    });
+
+    it('should return false if user does not have general edit right', () => {
+      const env = new TestEnvironment();
+      const project = createTestProjectProfile({
+        editable: true,
+        sync: { dataInSync: true },
+        texts: [
+          { bookNum: 1, chapters: [{ number: 1, isValid: true, permissions: { user01: TextInfoPermission.Write } }] }
+        ],
+        userRoles: { user01: SFProjectRole.ParatextObserver }
+      });
+
+      // SUT
+      const actual: boolean = env.textDocService.canRestore(project, 1, 1);
+      expect(actual).toBe(false);
+    });
+
+    it('should return false if user does not have chapter edit permission', () => {
+      const env = new TestEnvironment();
+      const project = createTestProjectProfile({
+        editable: true,
+        sync: { dataInSync: true },
+        texts: [
+          { bookNum: 1, chapters: [{ number: 1, isValid: true, permissions: { user01: TextInfoPermission.Read } }] }
+        ],
+        userRoles: { user01: SFProjectRole.ParatextAdministrator }
+      });
+
+      // SUT
+      const actual: boolean = env.textDocService.canRestore(project, 1, 1);
+      expect(actual).toBe(false);
+    });
+
+    it('should return false if data is not in sync', () => {
+      const env = new TestEnvironment();
+      const project = createTestProjectProfile({
+        editable: true,
+        sync: { dataInSync: false },
+        texts: [
+          { bookNum: 1, chapters: [{ number: 1, isValid: true, permissions: { user01: TextInfoPermission.Write } }] }
+        ],
+        userRoles: { user01: SFProjectRole.ParatextAdministrator }
+      });
+
+      // SUT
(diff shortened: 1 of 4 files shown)


RaymondLuong3 (changes requested):
Thanks for working on this. I can confirm that the restore button is available when the formatting is invalid. However, when I try to restore the version, the invalid error is still visible. I expect this is likely because we are restoring the text but not updating the textInfo for the particular chapter in the project document.

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3200#-:-OQFNQnQ1hm84aRqGnO8:b-89dyi7)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (commented):
Hm, this sounds weird, as I just tested this yesterday. Tell me more about what you're seeing, as you may have tried a use case I missed. Is the text editable? I'm guessing your "invalid error" is the "unsupported USFM warning"? What USFM did you use to cause this?

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3200#-:-OQGqGCJAMdGFnXrCzTz:bj93oj6)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers (commented):
So something is actually rejecting the restore (even though it says it successfully restored). For these revisions, the warning says to fix the error in Paratext. So you Sync to Paratext. Now back in your editor the previous revision is present, and Paratext S/R finds no changes. A fun problem for Monday!

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3200#-:-OQHIlM49AeKU2A5QjIj:be7thmy)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

josephmyers:
Got a chance to spend a little more time on this today. Something is quite wrong with the isValid field on some of our snapshots. Many, but not all, of the snapshots I'm seeing from Paratext with invalid markers are considered valid in SF. I see there's been a good bit of flux to the DeltaUsxMapper in the recent months -- that's where I'll start tomorrow.

josephmyers:
My changes exposed/introduced a bug which puts the project in a bad state if you restore an invalid revision not from Paratext. Be sure to delete any corrupted projects when testing.

josephmyers:
Technically, you can still see some weird behavior with SF revisions that are nearby PT revisions. Those SF revisions will inherit the validity of the nearest PT revision. I think. Will investigate more tomorrow.

josephmyers (commented):
Thanks for finding that Raymond. It uncovered a bug in the ParatextService, where we weren't validating the correct data for SF-generated revisions. I've pushed a fix and feel pretty good about it. I also disabled the ability to restore invalid revisions, but we can revert that commit, as I see it as optional (though I haven't tested reverting it).

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3200#-:-OQlhI7I9_lH3T3cADXB:b1hhr9s)* status: 3 of 10 files reviewed, all discussions resolved (waiting on @RaymondLuong3)

RaymondLuong3 (approved):
Let's finally get this to testers!

> Reviewed 1 of 2 files at r4, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3200#-:-ORD_VyG0YLPE4XreLXG:b-njhi17)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/Services/DeltaUsxMapper.cs` line 605 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3200#-ORD41ZN-qwMFf1Ll1oS:-ORD_2h_1ZYkoTpONBL0:bbcd1ls) ([raw file](https://github.com/sillsdev/web-xforge/blob/0d4db6ffbce4ba47ec512d728c66bd27e53c0d47/src/SIL.XForge.Scripture/Services/DeltaUsxMapper.cs#L605)):*

It would be 1.0, which I believe is what we'd want. You can test it out here: https://www.mobzystems.com/online/format-tester/
</blockquote></details>

Ok, I was completely off in my understanding of string.Format. Thanks for the helpful link!
___
*[`src/SIL.XForge.Scripture/Services/ParatextService.cs` line 2005 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3200#-ORD6YriCax3YuzNLnVC:-ORD_QXr1ZN3T1dVZ4aw:b-jadneb) ([raw file](https://github.com/sillsdev/web-xforge/blob/0d4db6ffbce4ba47ec512d728c66bd27e53c0d47/src/SIL.XForge.Scripture/Services/ParatextService.cs#L2005)):*

This is a good point. I had thought they maybe were interrelated, but I see that is wrong. I think for the purposes of validity, we can maybe fudge to 1.0 for now?
</blockquote></details>

That works for me.

PR #3201, opened 2025-05-14
SF-3331 Fix training source checkbox runoff for small screens

josephmyers:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3201)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.html
@@ -1,41 +1,43 @@
 <ng-container *transloco="let t; read: 'book_select'">
   @if (availableBooks.length > 0 && !readonly && !basicMode) {
-    <div class="scope-selection">
+    <div class="scope-selection" style="flex-wrap: wrap">
       @if (projectName != null) {
         <span class="project-name">{{ projectName }}</span>
       }
-      <mat-checkbox
-        class="ot-checkbox"
-        value="OT"
-        [disabled]="!isOldTestamentAvailable()"
-        [checked]="selectedAllOT"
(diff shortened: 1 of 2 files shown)


Nateowami (changes requested):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3201)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.html` line 3 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3201#-OQEYTBBEcoSsCfqJkEN:-OQEYTBBEcoSsCfqJkEO:bfw7yi1) ([raw file](https://github.com/sillsdev/web-xforge/blob/57b991d1663f02a084901dc9329034c028f6b9a2/src/SIL.XForge.Scripture/ClientApp/src/app/shared/book-multi-select/book-multi-select.component.html#L3)):*
> ```HTML
> <ng-container *transloco="let t; read: 'book_select'">
>   @if (availableBooks.length > 0 && !readonly && !basicMode) {
>     <div class="scope-selection" style="flex-wrap: wrap">
> ```

CSS should be in the .SCSS file. Looking at this in isolation, you can't even tell if this rule would have any effect, because it's not obvious that it's even a flexbox. On the other hand, when put in the SCSS file the rules will be right next to each other.

Otherwise LGTM.

pmachapman (approved):
:lgtm:

> Dismissed @Nateowami from a discussion.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3201#-:-OQeIRXRDMO4OTOjXKmk:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #3210, opened 2025-05-19
SF-3354 Update the new draft UI via SignalR

pmachapman:
This PR updates the new draft UI to use SignalR to update the frontend whenever Serval calls a webhook or when the pre-translations retrieval from Serval finishes when a build completes.

This also updates the SignalR client to the latest version.

I added a buildState object to the SignalR notifier, even though I don't yet need it, as it is a [breaking change](https://learn.microsoft.com/en-au/aspnet/core/signalr/api-design?view=aspnetcore-8.0#use-custom-object-parameters-to-ensure-backwards-compatibility) to add an object like this to a notification channel's signature after the fact, and the other issues dependent on this will likely need the info in this `ServalBuildState` object.

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/package-lock.json
+++ b/src/SIL.XForge.Scripture/ClientApp/package-lock.json
@@ -23,7 +23,7 @@
         "@angular/service-worker": "^18.2.8",
         "@auth0/auth0-spa-js": "^2.1.3",
         "@bugsnag/js": "^7.23.0",
-        "@microsoft/signalr": "^7.0.0",
+        "@microsoft/signalr": "^8.0.7",
         "@ngneat/transloco": "^4.3.0",
         "@ngneat/transloco-locale": "^4.1.0",
         "@sillsdev/lynx": "^0.3.4",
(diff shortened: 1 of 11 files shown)


pmachapman (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3210)* status: all files reviewed, 1 unresolved discussion (waiting on @RaymondLuong3)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-history-list/draft-history-list.component.ts` line 85 at r2](https://reviewable.io/reviews/sillsdev/web-xforge/3210#-ORDHw8m-L7CWfvnTg1c:-ORDrvPUD6cl9925fFae:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/bedab1c3094bab375f105a81930e79f666483b05/src/SIL.XForge.Scripture/ClientApp/src/app/translate/draft-generation/draft-history-list/draft-history-list.component.ts#L85)):*

You probably want to also pipe this through `take(1)` since we are only interested when `getBuildHistory` emits the first time.
</blockquote></details>

Done.

RaymondLuong3 (approved):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3210)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @pmachapman)

PR #3211, opened 2025-05-20
Fix CountBooksDrafted Mongo query

Nateowami:
Before this change this query was producing results that were clearly incorrect (I would run the query weeks later, and numbers were going down instead of up, even though it was querying a collection that only has documents added to it).

I don't know what was wrong with it. Probably reading a bunch of Mongo documentation and carefully looking at it would reveal the problem, or I can just rewrite it to be short and simple, which is the approach I took.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3211)

### DIFF
--- a/mongodb/EventMetrics/CountBooksDrafted.mongodb
+++ b/mongodb/EventMetrics/CountBooksDrafted.mongodb
@@ -94,41 +94,14 @@ const booksByProjectsDraftingIt = db.event_metrics.aggregate([
       books: { $addToSet: '$books' }
     }
   },
-  {
-    $project: {
-      books: {
-        $reduce: {
-          input: '$books',
-          initialValue: [],
-          in: {
-            $setUnion: ['$$value', '$$this']
-          }
-        }
-      },
-    }
-  },
-  // We now have a list of books by each project
-  // Unwinding this will allow us to count books by how many projects drafted each
-  { $unwind: '$books' },
-  {
-    $group: {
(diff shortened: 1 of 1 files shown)


mongodb/EventMetrics/CountBooksDrafted.mongodb
```diff
@@ -94,41 +94,14 @@ const booksByProjectsDraftingIt = db.event_metrics.aggregate([
...
-  const bIndex = bookIds.indexOf(b._id);
-  return aIndex - bIndex;
-});
-
-for (const record of booksByProjectsDraftingIt) {
-  print(record._id + '\t' + record.count);
+for (const bookId of bookIds) {
+  let count = 0;
+  for (const record of booksByProjectsDraftingIt) {
+    if (record.books != null && record.books.some(list => list != null && list.includes(bookId))) {
```
Nateowami, line 102: You may be looking at this and thinking it's a missed opportunity for optional chaining (`record?.books`), but no; unfortunately this is Mongo and we can't have nice things.

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3211#-:-OQj9FGN2ziIGYu6TdKy:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Nateowami)

PR #3213, opened 2025-05-22
SF-3382 Only fire changes to Lynx when the feature flag is enabled

pmachapman:
This PR provides an initial workaround for the crash reported in SF-3382, by only firing changes to Lynx if the Lynx feature flag is enabled.

The two Lynx bugs that this issue works around are:

 * sillsdev/lynx#35
 * sillsdev/lynx#36

Enabling the feature flag will cause the crash to occur, but disabling it (the default state for most users) will prevent the crash from occurring.

If you wish to recreate the crash, with apply a draft generated from DHH94 to a chapter, or (more easily) apply any draft twice to a chapter.

I have marked this as testing not required, as it is a workaround until the bugs are fixed in Lynx. When those a fixed, a follow up PR with this JIRA issue number should be created which will require testing.


Nateowami:
Thanks for the quick fix @pmachapman. I made some changes to the title when merging, to conform with our commit message guidelines:

> Explain what changed, not how it changed
The summary should explain what happens from a user's perspective, not what you did to fix a bug

Since I reviewed the change, I know what is meant by "Only fire changes to Lynx when the feature flag is enabled", but it's a description of what the solution/implementation was. If someone else is trying to write release notes it's not going to be very obvious what kind of change this was.

PR #3218, opened 2025-05-26
SF-3386 Don't initialize translator factory if suggestions are disabled

josephmyers:
This fixes a bug where getWordGraph was still called (and throwing errors) if the user had suggestions disabled via their user config. Now, showSuggestions should also be false in this case, which makes sense. (Having showSuggestions be true when the user disables it doesn't make sense to me, but maybe I'm missing a use case?)

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3218)


pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3218#-:-ORCyHuf5ENi1gRy6Vi4:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

PR #3220, opened 2025-05-27
SF-3335 Display additional text when syncing source project

josephmyers:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3220)

### DIFF
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.html
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.html
@@ -3,9 +3,9 @@
     <div *transloco="let t; read: 'sync'">
       <div class="sync-header centered-content">
         @if (isActiveProjectSource()) {
-          <div>{{ t("synchronize_reference_project") }}</div>
+          <div>{{ t("synchronize_source_project") }}</div>
         }
-        <h3>{{ activeSyncProjectName }}</h3>
+        <h3>{{ activeSyncProjectDoc?.data?.name }}</h3>
       </div>
       <div class="sync-status">
         @if (syncPhaseMessage.includes("_percent")) {
--- a/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts
+++ b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts
@@ -43,7 +43,7 @@ export class SyncProgressComponent {
   syncPhase?: SyncPhase;
   phasePercentage: number = 0;
   syncProgress: number = 0;
-  activeSyncProjectName: string = '';
+  activeSyncProjectDoc?: SFProjectDoc;
   private progressPercent$ = new BehaviorSubject<number>(0);
 
   /** The progress as a percent between 0 and 100 for the target project and the source project, if one exists. */
@@ -136,7 +136,7 @@ export class SyncProgressComponent {
   }
 
   public isActiveProjectSource(): boolean {
-    return this.activeSyncProjectName !== undefined && this.activeSyncProjectName === this.sourceProjectDoc?.data?.name;
+    return this.activeSyncProjectDoc !== undefined && this.activeSyncProjectDoc.id === this.sourceProjectDoc?.id;
   }
 
   protected updateProgressState(projectId: string, progressState: ProgressState): void {
@@ -146,12 +146,12 @@ export class SyncProgressComponent {
     this.phasePercentage =
       progressState.syncProgress != null ? Math.round((progressState.syncProgress - this.syncProgress) * 100) : 0;
     if (projectId === this._projectDoc?.id) {
-      this.activeSyncProjectName = this._projectDoc?.data?.name ?? '';
+      this.activeSyncProjectDoc = this._projectDoc;
       this.progressPercent$.next(
         hasSourceProject ? 0.5 + progressState.progressValue * 0.5 : progressState.progressValue
       );
     } else if (hasSourceProject && projectId === this.sourceProjectDoc?.id) {
-      this.activeSyncProjectName = this.sourceProjectDoc?.data?.name ?? '';
+      this.activeSyncProjectDoc = this.sourceProjectDoc;
       this.progressPercent$.next(progressState.progressValue * 0.5);
     }
   }
--- a/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json
+++ b/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json
@@ -720,7 +720,7 @@
     "successfully_synchronized_with_paratext": "Successfully synchronized {{ projectName }} with Paratext.",
     "synchronize": "Sync with Paratext",
     "synchronize_project": "Synchronize {{ projectName }} with Paratext",
-    "synchronize_reference_project": "Synchronizing source text:",
+    "synchronize_source_project": "Synchronizing source text:",
     "your_project_is_being_synchronized": "Your project is being synchronized...",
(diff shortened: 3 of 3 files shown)


josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3220)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @Nateowami)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts` line 139 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3220#-ORI6yF02fgzNgVXEm9g:-ORKL_3M4nYxj2-Dpbfw:b-i9seyy) ([raw file](https://github.com/sillsdev/web-xforge/blob/0c5695d407758643edf119fff709712754f6103b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts#L139)):*

Why are we using the project name as a identifier? Are we assuming the name can never change?
</blockquote></details>

Short answer: no, the code in this class is attempting to keep it up to date as the sync changes.

The field is only ever set from the sourceProjectDoc or projectDoc, and it's resynced each time the sync progress changes. If the project name changes on the project doc as part of a sync then the field will update accordingly and the logic will continue to work. Even in the case where the name is out of date on the project doc, we're only ever checking this field, which is sourced from the project doc, against what came from the project doc.

If the project doc name changed outside of a sync, then technically this method could produce wrong results outside of a sync (though we're only using it during sync), but at that point the whole field is wrong, and likely other parts of this component. As soon as you sync, the field will fix itself.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.spec.ts` line 246 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3220#-ORI6D2QCmyP1Qt-1w2J:-ORKBOtM5iAmSg-OPzTG:b-8ki6gw) ([raw file](https://github.com/sillsdev/web-xforge/blob/0c5695d407758643edf119fff709712754f6103b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.spec.ts#L246)):*

Marking this protected and then cheating the type system seems to me to be counterproductive. If this is ever renamed, or someone tries to find all references, it won't work properly. Much better in my opinion to just leave it public.
</blockquote></details>

Both renaming and Find All References work perfectly with this, at least in VS Code. It doesn't seem right having this public. We shouldn't be, and aren't, relying on class consumers to update its properties for it (and we shouldn't make API concessions just for tests).
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 709 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3220#-ORI6RF30gG_dQ-0CPyf:-ORKDU9YCFr5wbUPj4q1:bb6un1j) ([raw file](https://github.com/sillsdev/web-xforge/blob/0c5695d407758643edf119fff709712754f6103b/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L709)):*

Nowhere else in the UI do we call this a "reference project". The only place we use the term "reference project" is on the draft sources page. Maybe we should change our wording, but we definitely shouldn't add a description here that's inconsistent with the rest of the application.
</blockquote></details>

Done

josephmyers (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3220)* status: 2 of 5 files reviewed, 2 unresolved discussions (waiting on @Nateowami and @pmachapman)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts` line 139 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3220#-ORI6yF02fgzNgVXEm9g:-ORPo6kw1dl4yorslKTj:b-3uymci) ([raw file](https://github.com/sillsdev/web-xforge/blob/0c5695d407758643edf119fff709712754f6103b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.ts#L139)):*

There is at least one project with the same name as its back translation source on live. Can we use a different differentiator (i.e. `id`) to determine if the source project is syncing?
</blockquote></details>

Done. Thanks, I didn't know that was possible!
___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.spec.ts` line 246 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3220#-ORI6D2QCmyP1Qt-1w2J:-ORPhNiC2ihRWkT3UNFV:b-bl2ifm) ([raw file](https://github.com/sillsdev/web-xforge/blob/0c5695d407758643edf119fff709712754f6103b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.spec.ts#L246)):*

I agree with Nathaniel on this one - if we were exporting our TypeScript module to another system/codebase/npm/etc, then yes, it would make sense to keep it `protected`, but as we don't, and TypeScript doesn't have an `internal` modifier, I think that functions called from unit tests should be `public`, unless they are particularly obscure.

@Nateowami I can't find any docs on coding standards in TypeScript about public/private and tests (unlike C# which has a lot of material on this question). Maybe we should look at a form of TypeScript coding standards?
</blockquote></details>

I'm confused. Both of you guys are recommending against pretty basic OOP principles, even when we have an alternative we're using in many other places.

The issue here at least is less about coding standards and more about ease of testing. The only reason this test needs this method is because the notification service doesn't have an actual, i.e. usable in tests, way to notify its users. You'll notice it has no tests. It's a direct line to SignalR, and so both end up really difficult to test. Which is exactly why bypassing the access makes sense here.

If bypassing access in tests is cheating, then that's really good for me to know. But we should probably fix all the other tests that are using it, too.
___
*[`src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json` line 723 at r3](https://reviewable.io/reviews/sillsdev/web-xforge/3220#-ORNRL7LEROL58Mt1vof:-ORPpur83kSMCivXry6B:b-896fix) ([raw file](https://github.com/sillsdev/web-xforge/blob/964162ec8a601b1d3c34dc131f69c97179156e1c/src/SIL.XForge.Scripture/ClientApp/src/assets/i18n/non_checking_en.json#L723)):*

Can we rename this key to `synchronizing_source_project` or `synchronize_source_project` or similar? We don't use the phrase reference project anywhere else to refer to the source project.
</blockquote></details>

Done.

pmachapman (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/sillsdev/web-xforge/3220#-:-ORm2iS97USv7sR-eiv9:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @josephmyers)

___
*[`src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.spec.ts` line 246 at r1](https://reviewable.io/reviews/sillsdev/web-xforge/3220#-ORI6D2QCmyP1Qt-1w2J:-ORm24fb5L2obHdVMo7P:biztwvl) ([raw file](https://github.com/sillsdev/web-xforge/blob/0c5695d407758643edf119fff709712754f6103b/src/SIL.XForge.Scripture/ClientApp/src/app/sync/sync-progress/sync-progress.component.spec.ts#L246)):*

I'm confused. Both of you guys are recommending against pretty basic OOP principles, even when we have an alternative we're using in many other places.

The issue here at least is less about coding standards and more about ease of testing. The only reason this test needs this method is because the notification service doesn't have an actual, i.e. usable in tests, way to notify its users. You'll notice it has no tests. It's a direct line to SignalR, and so both end up really difficult to test. Which is exactly why bypassing the access makes sense here.

If bypassing access in tests is cheating, then that's really good for me to know. But we should probably fix all the other tests that are using it, too.
</blockquote></details>

IMO, TypeScript/JavaScript's lack of real encapsulation is the problem.

It is no biggie for this PR. I don't think this as a blocker.

